PR #65, opened 2022-12-05
Code reviewe

kos-for-juspay:
Thorough code review


jarnura:
We did most of the changes list here. There are some things pending which we take that later in the roadmap. Thanks for the deep review.

PR #211, opened 2022-12-21
PoC: Hexagonal Architecture Shift

kos-for-juspay:
## Description
* Hexagonal architecture is a software design pattern that helps to separate the business logic from the input/output logic. It does this by defining interfaces for each external interaction, which can then be injected into services that use them. This allows for testing with mocks and interchangeable implementations.

* Ports and Adapters is an architectural pattern that helps to separate the business logic from technical implementation details. It does this by creating ports, which are the boundaries of an application, and adapters, which are used to translate between the ports and the application model. This allows us to be more flexible and respond quickly to changes in business or technology.

## Motivation and Context
* Moving adapters and core from the router will dramatically improve health and decrease the time to recompile.
* App does not have unit tests, but it does have acceptance tests. Unit tests are tests that are written to test individual components of a program or system. Acceptance tests are tests written to determine if the program or system meets the requirements of the customer or user. Acceptance tests are usually done after unit tests have been completed to make sure that the requirements have been met.

## Type of Change

- [x] Requested demonstration of Ports and Adapters Architecture


jarnura:
Thanks kos,

We can't a achieve the single core flow abstraction with this approach, so that we end on diverging flows on how we processing payments for a connector and other auto generating of connector integration will be affected by this.

So closing this pull request.

PR #282, opened 2023-01-04
fix(router): make customer id mandatory when address or setup future usage given

vspecky:
## Type of Change

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates

## Description
Make customer ID mandatory when addresses or setup_future_usage are given.

### Additional Changes
Nil

## Motivation and Context
Customer id is required when addresses are given for GDPR compliance. It's also required when setup future usage is given since the payment method needs to be saved and associated with the given customer.

## How did you test it?
Stripe payments.

- **PaymentCreate**: when confirm = true and setup_future_usage is given but not customer id.
image

- **PaymentUpdate**: confirm = true, but customer id not given. Note that neither address nor setup future usage is given in the request. However it was given in the PaymentCreate request, i.e. verification needs to be done against values present in the database as well.
image

- **PaymentConfirm**: similar to **PaymentUpdate**
image

## Checklist

### DIFF
--- a/crates/router/src/core/payments/operations/payment_confirm.rs
+++ b/crates/router/src/core/payments/operations/payment_confirm.rs
@@ -111,7 +111,9 @@ impl<F: Send + Clone> GetTracker<F, PaymentData<F>, api::PaymentsRequest> for Pa
         helpers::validate_customer_id_mandatory_cases_storage(
             &shipping_address,
             &billing_address,
-            &payment_intent.setup_future_usage,
+            &payment_intent
+                .setup_future_usage
+                .or_else(|| request.setup_future_usage.map(ForeignInto::foreign_into)),
             &payment_intent
                 .customer_id
                 .clone()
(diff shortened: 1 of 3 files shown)


crates/router/src/core/payments/operations/payment_confirm.rs
```diff
@@ -87,24 +87,10 @@ impl<F: Send + Clone> GetTracker<F, PaymentData<F>, api::PaymentsRequest> for Pa
...
-                &payment_attempt.merchant_id,
-                &payment_attempt.attempt_id,
-                storage_scheme,
-            )
-            .await
-            .map_err(|error| {
-                error.to_not_found_response(errors::ApiErrorResponse::PaymentNotFound)
-            })?;
-
         let shipping_address = helpers::get_address_for_payment_request(
```
lsampras, line 94: We need to verify that the customer_id exists before checking for saved addresses.

lsampras: Also the function `get_address_for_payment_request` needs to accept a customer_id instead of Option<customer_id>

PR #324, opened 2023-01-09
feat(session_token): serialize with camelCase for gpay token

Narayanbhat166:
## Type of Change

- [x] Refactoring

## Description

Send the google pay session token ( taken from metadata ), in camelCase since google pay is expecting it in camelCase.

## Motivation and Context

Refactoring request from the frontend team

## How did you test it?

Screenshot 2023-01-09 at 3 26 36 PM

## Checklist

### DIFF
--- a/crates/api_models/src/payments.rs
+++ b/crates/api_models/src/payments.rs
@@ -744,18 +744,21 @@ pub struct PaymentsSessionRequest {
 }
 
 #[derive(Debug, Clone, serde::Serialize, serde::Deserialize)]
+#[serde(rename_all(serialize = "camelCase"))]
 pub struct GpayAllowedMethodsParameters {
     pub allowed_auth_methods: Vec<String>,
     pub allowed_card_networks: Vec<String>,
 }
 
 #[derive(Debug, Clone, serde::Serialize, serde::Deserialize)]
+#[serde(rename_all(serialize = "camelCase"))]
 pub struct GpayTokenParameters {
     pub gateway: String,
     pub gateway_merchant_id: String,
 }
 
 #[derive(Debug, Clone, serde::Serialize, serde::Deserialize)]
+#[serde(rename_all(serialize = "camelCase"))]
 pub struct GpayTokenizationSpecification {
     #[serde(rename = "type")]
     pub token_specification_type: String,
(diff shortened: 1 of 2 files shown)


crates/api_models/src/payments.rs
```diff
@@ -744,25 +744,29 @@ pub struct PaymentsSessionRequest {
 }
 
 #[derive(Debug, Clone, serde::Serialize, serde::Deserialize)]
+#[serde(rename_all(serialize = "camelCase"))]
```
maverox, line 747: **🟡 Important - Serialization Asymmetry**

This creates an asymmetric serialization/deserialization behavior:
- **Serialization**: Uses camelCase (via the new attribute)
- **Deserialization**: Uses snake_case (Rust default)

**Issue**: If this data is ever round-tripped (deserialized then serialized), the field names won't match.

**Suggested Fix**: Either apply the rename to both serialize and deserialize:
```rust
```

Or if only serialization needs camelCase, add a comment explaining why deserialization should remain snake_case.

crates/api_models/src/payments.rs
```diff
@@ -799,7 +806,9 @@ pub struct GpaySessionTokenData {
 #[serde(rename_all = "lowercase")]
 pub enum SessionToken {
     Gpay {
-        allowed_payment_methods: Vec<GpayAllowedPaymentMethods>,
+        #[serde(flatten)]
+        gpay_token: GpayMetadata,
+        #[serde(rename(serialize = "transactionInfo"))]
```
maverox, line 811: **🟢 Suggestion - Pattern Consistency**

Good refactoring to use `flatten` to include all GpayMetadata fields. However, I notice:

1. You're using `#[serde(rename(serialize = "camelCase"))]` on `transactionInfo` 
2. But the parent `GpayMetadata` struct already has `#[serde(rename_all(serialize = "camelCase"))]`

**Question**: Is the explicit `rename` on line 811 necessary? If `transaction_info` is already being serialized with camelCase from the struct definition, this might be redundant.

**Suggested Investigation**: Test if removing this explicit rename still produces the correct camelCase output for `transactionInfo`.

crates/api_models/src/payments.rs
```diff
@@ -799,7 +806,9 @@ pub struct GpaySessionTokenData {
 #[serde(rename_all = "lowercase")]
 pub enum SessionToken {
     Gpay {
-        allowed_payment_methods: Vec<GpayAllowedPaymentMethods>,
+        #[serde(flatten)]
```
maverox, line 809: **✅ Good Practice - Struct Flattening**

Nice refactoring! This change:
- ✅ Removes code duplication (`allowed_payment_methods` was redundant)
- ✅ Makes the structure cleaner by flattening `GpayMetadata` 
- ✅ Keeps `transaction_info` separate as it needs different handling

This follows the DRY principle and makes the enum variant more maintainable.

PR #337, opened 2023-01-10
feat(core): add access token connector auth type

Narayanbhat166:
## Type of Change

- [x] New feature

## Description

This PR will add support for refreshing the access token for supported connectors. Connector integration should be provided for supported connectors.

## Motivation and Context

For some connectors there is a requirement to get a time constrained access token which must be used for each request. This access token must be refreshed when it expires. This PR provides an interface for this feature. 
A pseudo code of the interface
```rust
let access_token = get_access_token_from_redis()
if let Some(token) = access_token {
    token = refresh_access_token()
    store_access_token_in_db_with_expiry(token)
}
```
Access token connector auth type is 
```rust
AccessToken {
        api_key: String,
        id: String,
        #[serde(skip_deserializing)]
        access_token: Option<String>,
    }
```

## How did you test it?

Not testable, will test by integrating with one of the connectors

## Checklist

### DIFF
--- a/crates/common_utils/src/lib.rs
+++ b/crates/common_utils/src/lib.rs
@@ -27,6 +27,11 @@ pub mod date_time {
     pub fn convert_to_pdt(offset_time: OffsetDateTime) -> PrimitiveDateTime {
         PrimitiveDateTime::new(offset_time.date(), offset_time.time())
     }
+
+    /// Get the current time in unix timestamp format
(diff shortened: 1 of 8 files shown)


Narayanbhat166:
Since there can be many connectors which have different types of auth ( i.e the number of keys ), it's better to use the same connector auth type and use it to convert to accestokenauthtyperequest internally. Closing this now since the suggested implementation has been already done in #407

PR #341, opened 2023-01-10
feat(core): implement refresh token functionality

Narayanbhat166:
## Type of Change

- [x] New feature

## Description

This PR will provide the implementation to refresh the access token which was added in #337. Refresh access token if not found in redis and store the access token in redis with expiry ( in seconds ).

## Motivation and Context

Add an access token type 
```rust
pub struct AccessToken {
    pub token: String,
    pub expires: i64, // expiration time in seconds
}
```
Provide implementation for `fn refresh_connector_access_token()`

## How did you test it?

Not able to test it, since no connector integration is done.

## Checklist


Narayanbhat166:
There is a separate implementation provided in #407

PR #345, opened 2023-01-11
fix(braintree): map settling status to charged

Narayanbhat166:
## Type of Change

- [x] Bugfix

## Description

This PR will fix the payment status of paypal payments being`Processing`. 

## Motivation and Context

When creating paypal payments, submitted for settlement is set to true so that automatic settlments happen. But braintree settles the amount after 90 minutes, so the status sent is `Settling` which is mapped to attempt status of `Pending` which is mapped to intent status of `Processing`. Looking at the transaction lifecycle, if the status is`Settling` then it has already passed `Authorized`, so it is safe to map`Settling` to `Charged`

Screenshot 2023-01-11 at 12 22 41 PM

## How did you test it?

Screenshot 2023-01-11 at 12 36 29 PM

## Checklist


bernard-eugine (approved):

SanchithHegde (approved):
Looks good to me!

jarnura (approved):

PR #348, opened 2023-01-11
feat: separate olap oltp

Sangamesh26:
## Type of Change

- [x] New feature

## Description

1. Instead of replica_db(Now , we are just using another user of the same master db  with just read access) and after getting replica_db, we can change the configurations.
2. Separating all OLAP and OLTP routes

> Run the application(with default OLTP) : cargo r
> Run the application for OLAP: cargo r --features olap

Another user(from master db) with only read access:
user name: rep
password: rep

### Additional Changes

## Motivation and Context

OLAP is for historical data with mostly read access
OLTP is for the normal operational transactions

## How did you test it?

Manual

## Checklist


crates/router/src/olap_lib.rs
```diff
@@ -0,0 +1,79 @@
+#![forbid(unsafe_code)]
```
jarnura, line 1: keep one lib.rs file only, add these functions there.

crates/router/src/routes/app.rs
```diff
@@ -207,18 +231,25 @@ impl EphemeralKey {
...
+            .app_data(web::Data::new(config))
             .service(web::resource("/revoke/{id}").route(web::post().to(revoke_mandate)))
     }
 }
 
 pub struct Webhooks;
 
 impl Webhooks {
-    pub fn server(config: AppState) -> Scope {
+    // pub fn olap_server(state: AppState) -> Scope {
```
jarnura, line 249: remove commented codes

crates/router/src/services.rs
```diff
@@ -29,7 +28,6 @@ impl Store {
     pub async fn new(config: &crate::configs::settings::Settings, test_transaction: bool) -> Self {
         Self {
             master_pool: diesel_make_pg_pool(&config.master_database, test_transaction).await,
-            #[cfg(feature = "olap")]
```
jarnura, line 32: we need to create feature flag to create connections, oltp cases the connections need not to be established to the databases.

jarnura (changes requested):

jarnura (changes requested):

jarnura (approved):

config/Development.toml
```diff
@@ -18,6 +18,14 @@ port = 5432
 dbname = "hyperswitch_db"
 pool_size = 5
 
+[replica_database]
+username = "rep"
+password = "rep"
+host = "localhost"
+port = 5432
+dbname = "hyperswitch_db"
+pool_size = 5
```
SanchithHegde, line 27: Not exactly a blocker, but could you update the username and password to be slightly more descriptive, `replica_user` and `replica_pass` maybe? And the database name, could you make it `replica_db` instead?

If you're making this change, please make this change in `config.example.toml` as well.

crates/router/src/db/payment_intent.rs
```diff
@@ -307,13 +300,15 @@ mod storage {
...
+        #[cfg(feature = "olap")]
         async fn filter_payment_intent_by_constraints(
             &self,
             merchant_id: &str,
-            pc: &api::PaymentListConstraints,
+            pc: &crate::types::api::PaymentListConstraints,
             _storage_scheme: enums::MerchantStorageScheme,
         ) -> CustomResult<Vec<PaymentIntent>, errors::StorageError> {
-            let conn = pg_connection(&self.master_pool).await;
+            #[cfg(feature = "olap")]
```
SanchithHegde, line 310: This annotation shouldn't be required since the whole function has the annotation, isn't it?

crates/router/src/db/refund.rs
```diff
@@ -172,14 +173,16 @@ mod storage {
...
+        #[cfg(feature = "olap")]
         async fn filter_refund_by_constraints(
             &self,
             merchant_id: &str,
             refund_details: &api_models::refunds::RefundListRequest,
             _storage_scheme: enums::MerchantStorageScheme,
             limit: i64,
         ) -> CustomResult<Vec<storage_models::refund::Refund>, errors::StorageError> {
-            let conn = pg_connection(&self.master_pool).await;
+            #[cfg(feature = "olap")]
```
SanchithHegde, line 184: Same here, I believe this isn't required.

crates/router/src/lib.rs
```diff
@@ -88,36 +89,100 @@ pub fn mk_app(
...
+            errors::error_handlers::custom_error_handlers,
+        ))
+        .wrap(cors::cors())
+        .service(routes::Payments::oltp_server(state.clone()))
+        .service(routes::Customers::oltp_server(state.clone()))
+        .service(routes::Refunds::oltp_server(state.clone()))
+        .service(routes::Payouts::oltp_server(state.clone()))
+        .service(routes::PaymentMethods::oltp_server(state.clone()))
+        .service(routes::EphemeralKey::oltp_server(state.clone()))
+        .service(routes::Webhooks::oltp_server(state.clone()));
```
SanchithHegde, line 164: There seems to be a significant amount of shared code between `mk_olap_app()` and `mk_oltp_app()`, specifically till before services are registered. Can we move that to a separate function that returns the application builder?

SanchithHegde (changes requested):

config/Development.toml
Sangamesh26: I have changed the username and pasword, as we are using the same db but different user, so I have kept hyperswitch_db only.

crates/router/src/lib.rs
Sangamesh26: Done

crates/router/src/lib.rs
```diff
@@ -124,9 +115,8 @@ pub async fn start_olap_server(conf: settings::Settings) -> ApplicationResult<(S
     Ok((server, app_state))
 }
 
-pub fn mk_oltp_app(
-    state: AppState,
-    request_body_limit: usize,
+pub fn get_application_builder(
+    json_cfg: JsonConfig,
```
SanchithHegde, line 119: I think we can construct the `JsonConfig` instance within this function itself, instead of providing it as a parameter, considering that the instance is the same for both the apps.

SanchithHegde (changes requested):

crates/router/src/lib.rs
```diff
@@ -75,7 +77,56 @@ pub fn mk_app(
...
+        server_app = server_app.service(routes::StripeApis::server(state.clone()));
+    }
+    server_app = server_app.service(routes::Health::olap_server(state));
+    server_app
+}
+
+#[allow(clippy::expect_used, clippy::unwrap_used)]
+/// # Panics
+///
+///  Unwrap used because without the value we can't start the server
```
NishantJoshi00, line 101: Can we have a bit more detailed comments

NishantJoshi00 (changes requested):

crates/router/src/lib.rs
Sangamesh26: Done

NishantJoshi00 (dismissed):

crates/router/src/lib.rs
```diff
@@ -99,6 +92,7 @@ pub fn mk_olap_app(
 /// # Panics
 ///
 ///  Unwrap used because without the value we can't start the server
+/// Starts the OLAP server with only OLAP services
```
SanchithHegde, line 95: This description should be included in the beginning of the function documentation.

```rust
/// Description
///
/// # Errors/Panics/Safety/Examples (one or more sections)
///
/// ...
```

Also, move the `#[allow(...)]` annotations after the documentation comment, just before the function signature.

crates/router/src/lib.rs
```diff
@@ -182,6 +176,7 @@ pub fn mk_oltp_app(
 /// # Panics
 ///
 ///  Unwrap used because without the value we can't start the server
+/// Starts the OLTP server with only OLTP services
 pub async fn start_oltp_server(conf: settings::Settings) -> ApplicationResult<(Server, AppState)> {
```
SanchithHegde, line 180: Same here.

SanchithHegde (changes requested):

jarnura (dismissed):

crates/router/src/lib.rs
```diff
@@ -68,14 +69,64 @@ pub fn mk_app(
...
+    #[cfg(feature = "stripe")]
+    {
+        server_app = server_app.service(routes::StripeApis::server(state.clone()));
+    }
+    server_app = server_app.service(routes::Health::olap_server(state));
+    server_app
+}
+
+/// Starts the OLAP server with only OLAP services
+/// # Panics
```
SanchithHegde, line 92:
```suggestion
///
/// # Panics
```

crates/router/src/lib.rs
```diff
@@ -88,36 +139,52 @@ pub fn mk_app(
...
         server_app = server_app.service(routes::StripeApis::server(state.clone()));
     }
-    server_app = server_app.service(routes::Health::server(state));
+    server_app = server_app.service(routes::Health::oltp_server(state));
     server_app
 }
 
-#[allow(clippy::expect_used, clippy::unwrap_used)]
+/// Starts the OLTP server with only OLTP services
 /// # Panics
```
SanchithHegde, line 176:
```suggestion
///
/// # Panics
```

SanchithHegde (dismissed):

SanchithHegde (approved):

jarnura (approved):

PR #352, opened 2023-01-11
refactor: move config defaults from TOML files to `Default` trait

SanchithHegde:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [x] Refactoring
- [ ] Dependency updates

## Description

This PR moves default values for `router` and `drainer` binaries' configuration from the `defaults.toml` files to the implementation of the `Default` trait. Since we add a `#[serde(default)]` to the `Settings` struct, it will use the `Default` trait implementation to provide the default values.

## Motivation and Context

This should help catch errors in the default configuration values during compilation, and help reduce binary size.

## How did you test it?

Ran the `router` and `drainer` binaries, verified that they both work with the current defaults provided.

## Checklist


jarnura (commented):
There are some mandatory field needs to configured in the ENV's, for those we can't provide default values. This approach will fail for those cases.

We need to check in compile time only for field where we can assign defaults and not all.

Example: Database username is mandatory field needs to passed and we can't assign a default to it.

SanchithHegde:
> There are some mandatory field needs to configured in the ENV's, for those we can't provide default values. This approach will fail for those cases.
> 
> We need to check in compile time only for field where we can assign defaults and not all.
> 
> Example: Database username is mandatory field needs to passed and we can't assign a default to it.

Yes, I did think of this problem when making this change, and intentionally used `String::new()` for such fields.

One solution that I can think of is that we can perform checks on startup to verify that these required fields are present and are not empty. What do you think?

jarnura:
Assign defaults to all fields and in runtime evaluate the values are not equal to deaults

crates/drainer/src/settings.rs
```diff
@@ -37,35 +40,99 @@ pub struct Database {
...
+            Err(errors::DrainerError::ConfigParsingError(
+                "database name must not be empty".into(),
+            ))
+        })
+    }
+}
+
+impl DrainerSettings {
+    fn validate(&self) -> Result<(), errors::DrainerError> {
+        common_utils::fp_utils::when(self.stream_name.is_default_or_empty(), || {
```
jarnura, line 105: use directly when, relative path not required.

crates/router/src/configs/validations.rs
```diff
@@ -0,0 +1,186 @@
...
+                ))
+            },
+        )
+    }
+}
+
+impl super::settings::Jwekey {
+    pub(crate) fn validate(&self) -> Result<(), ApplicationError> {
+        #[cfg(feature = "kms")]
+        common_utils::fp_utils::when(self.aws_key_id.is_default_or_empty(), || {
```
jarnura, line 47: use when directly.

jarnura: Same in other places too

jarnura (changes requested):

jarnura (approved):
LGTM :)

vspecky (approved):
LGTM

PR #375, opened 2023-01-13
feat(storage_models): impl generic filter order and add function in payment_attempt to get latest based on created_at

Abhicodes-crypto:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates

## Description

In order to have multiple payment_attempts for a single payment_intent we need a function to fetch the latest payment attempt. 

### Additional Changes

## How did you test it?

Checked if it compiles.

## Checklist

### DIFF
--- a/crates/storage_models/src/query/generics.rs
+++ b/crates/storage_models/src/query/generics.rs
@@ -359,6 +359,37 @@ where
     .attach_printable_lazy(|| "Error filtering records by predicate")
 }
 
+#[instrument(level = "DEBUG", skip_all)]
+pub(super) async fn generic_filter_order<T, P, R, Expr>(
+    conn: &PgPooledConn,
+    predicate: P,
+    limit: Option<i64>,
+    expr: Expr,
+) -> StorageResult<Vec<R>>
+where
+    Expr: diesel::Expression,
+    T: FilterDsl<P> + HasTable<Table = T> + Table + 'static,
+    <T as FilterDsl<P>>::Output: OrderDsl<Expr> + Send + 'static,
+    <<T as FilterDsl<P>>::Output as OrderDsl<Expr>>::Output: LimitDsl + Send + 'static,
+    <<<T as FilterDsl<P>>::Output as OrderDsl<Expr>>::Output as LimitDsl>::Output:
+        LoadQuery<'static, PgConnection, R> + QueryFragment<Pg> + Send,
+    R: Send + 'static,
+{
+    let query = <T as HasTable>::table()
+        .filter(predicate)
+        .order(expr)
+        .limit(limit.unwrap_or(100));
+
+    logger::debug!(query = %debug_query::<Pg, _>(&query).to_string());
+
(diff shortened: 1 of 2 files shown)


crates/storage_models/src/query/generics.rs
```diff
@@ -359,6 +359,37 @@ where
...
+    <T as FilterDsl<P>>::Output: OrderDsl<Expr> + Send + 'static,
+    <<T as FilterDsl<P>>::Output as OrderDsl<Expr>>::Output: LimitDsl + Send + 'static,
+    <<<T as FilterDsl<P>>::Output as OrderDsl<Expr>>::Output as LimitDsl>::Output:
+        LoadQuery<'static, PgConnection, R> + QueryFragment<Pg> + Send,
+    R: Send + 'static,
+{
+    let query = <T as HasTable>::table()
+        .filter(predicate)
+        .order(expr)
+        .limit(limit.unwrap_or(100));
```
Abhicodes-crypto, line 381: Let me know if the default value has to be added as a const in the beginning of the file or as a config.

ashokkjag: Since limit is passed by caller, You can just move it to CONST.

crates/storage_models/src/query/generics.rs
```diff
@@ -359,6 +359,37 @@ where
...
+
+    logger::debug!(query = %debug_query::<Pg, _>(&query).to_string());
+
+    query
+        .get_results_async(conn)
+        .await
+        .into_report()
+        .change_context(errors::DatabaseError::NotFound)
+        .attach_printable_lazy(|| "Error filtering records by predicate and order")
```
ashokkjag, line 390: Wordings imply expected to find a record, but not found need not be an error?

Abhicodes-crypto: Got it. It will return a zero sized vector if nothing is found. And any other error should be propagated. I copied this part from the above generic_filter function. Should I change it there as well.

Abhicodes-crypto: There are implementations which rely on the NotFound Error, for generic_filter. I'll take that up in another PR.

crates/storage_models/src/query/generics.rs
```diff
@@ -359,6 +361,37 @@ where
     .attach_printable_lazy(|| "Error filtering records by predicate")
 }
 
+#[instrument(level = "DEBUG", skip_all)]
+pub(super) async fn generic_filter_order<T, P, R, Expr>(
+    conn: &PgPooledConn,
+    predicate: P,
+    limit: Option<i64>,
+    expr: Expr,
```
SanchithHegde, line 369: Please name this parameter as `order_by` or `order_by_clause` for easier understanding.

SanchithHegde (changes requested):

PR #388, opened 2023-01-16
feat: token flow for wallet

Sangamesh26:
## Type of Change

- [x] New feature

## Description

> When payment token is present, fetch payment method data from the locker.
> Update the payment method data in locker, if wallet data is present.

## Motivation and Context

Implement token flow for wallet

## How did you test it?

Manual

## Checklist


SanchithHegde (changes requested):

SanchithHegde (approved):

jarnura (approved):

PR #391, opened 2023-01-17
refactor(router): move `fp_utils` module to `common_utils` crate

SanchithHegde:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [x] Refactoring
- [ ] Dependency updates

## Description

This PR moves the `fp_utils` module to the `common_utils` crate.

## Motivation and Context

This change would allow using the `when()` utility function in crates other than the `router` crate.

## How did you test it?

N/A

## Checklist


crates/common_utils/src/fp_utils.rs
```diff
@@ -18,7 +25,7 @@ impl<R, E> Applicative<R> for Result<R, E> {
     }
 }
 
-// This function allows lazy evaluation of the `f` argument
+/// This function allows lazy evaluation of the `f` argument
```
NishantJoshi00, line 28: can we provide a little more context about this function

NishantJoshi00 (changes requested):

crates/common_utils/src/fp_utils.rs
NishantJoshi00:
```suggestion
/// This function wraps the evaluated result of `f` into current context,
/// based on the condition provided into the `predicate`
```

jarnura (approved):

NishantJoshi00 (approved):

PR #404, opened 2023-01-18
feat: Add redirect url to pm list

dracarys18:
## Type of Change

- [x] Enhancement

## Description

List payment methods needs to have redirect url in it for sdk to fallback incase of any failure.

## Motivation and Context

N/A

## How did you test it?

Manual

## Checklist

### DIFF
--- a/crates/api_models/src/payment_methods.rs
+++ b/crates/api_models/src/payment_methods.rs
@@ -168,8 +168,14 @@ fn set_or_reject_duplicate<T, E: de::Error>(
     }
 }
 
-#[derive(Eq, PartialEq, Hash, Debug, serde::Deserialize)]
+#[derive(Debug, serde::Serialize)]
 pub struct ListPaymentMethodResponse {
+    pub redirect_url: Option<String>,
+    pub payment_methods: collections::HashSet<ListPaymentMethod>,
+}
+
+#[derive(Eq, PartialEq, Hash, Debug, serde::Deserialize)]
+pub struct ListPaymentMethod {
     pub payment_method: api_enums::PaymentMethodType,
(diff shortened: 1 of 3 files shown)


crates/api_models/src/payment_methods.rs
```diff
@@ -168,8 +168,14 @@ fn set_or_reject_duplicate<T, E: de::Error>(
     }
 }
 
-#[derive(Eq, PartialEq, Hash, Debug, serde::Deserialize)]
+#[derive(Debug, serde::Serialize)]
 pub struct ListPaymentMethodResponse {
+    pub redirect_url: Option<String>,
+    pub payment_methods: collections::HashSet<ListPaymentMethod>,
```
SanchithHegde, line 174: You can import `HashSet` as is, no need to qualify it as `collections::HashSet`, since it's a standard library item.

crates/router/src/core/payment_methods/cards.rs
```diff
@@ -383,8 +383,7 @@ pub async fn list_payment_methods(
             error.to_not_found_response(errors::ApiErrorResponse::MerchantAccountNotFound)
         })?;
 
-    let mut response: collections::HashSet<api::ListPaymentMethodResponse> =
-        collections::HashSet::new();
+    let mut response: collections::HashSet<api::ListPaymentMethod> = collections::HashSet::new();
```
SanchithHegde, line 386: Same here, use it as `HashSet` instead.

SanchithHegde (changes requested):

jarnura (approved):

SanchithHegde (approved):

Issue #185, opened 2022-12-20
feat: Get a Token with expiry for any connector payment flow


zereraz:
`GlobalPayments` and `Paypal` are to connectors that require tokens to make any call to their Api's.
These tokens have an expiry as well.

Need feature to get the token for any of the connector flows to proceed with the payment.

PR #313, opened 2023-01-06
fix: fixing payment method information for refunds

NishantJoshi00:
## Type of Change

- [x] Bugfix

## Description

This PR solves the following bugs
1. Refund not working for normal payments
2. Refund not working for recurring payments
3. While also maintaining support for `authorizedotnet` refunds

### Additional Changes

- [ ] This PR modifies the database schema
- [ ] This PR modifies application configuration/environment variables

## Motivation and Context
N/A

## How did you test it?
Screenshot 2023-01-06 at 5 25 32 PM

## Checklist

- [x] I formatted the code `cargo +nightly fmt`
- [x] I addressed lints thrown by `cargo clippy`
- [x] I reviewed submitted code

PR #407, opened 2023-01-18
feat(access_token): support creating access token for connectors

Narayanbhat166:
## Type of Change

- [x] New feature

## Description

This PR will add the support for creating the access token for connectors which need it. 

## Motivation and Context

Once the routerdata is constructed, there is a next step which will check if the connector needs access token and then either get the access token ( which is currently stored in redis, with expiry ) or refresh it by calling the connector. A separate flow has been introduced to support this feature called `UpdateAuth`.

## How did you test it?

Manual, Compiler guided.

## Checklist


crates/router/src/services/refresh_token.rs
```diff
@@ -0,0 +1,149 @@
...
+    Res: Debug + Clone + 'static,
+>(
+    state: &AppState,
+    connector: &api_types::ConnectorData,
+    merchant_account: &storage::MerchantAccount,
+    router_data: &types::RouterData<F, Req, Res>,
+) -> RouterResult<(
+    Result<Option<types::AccessToken>, types::ErrorResponse>,
+    bool,
+)> {
```
SanchithHegde, line 64: It would have been great if you could make this a struct instead. When the return type is a tuple, we have no information on what the bool in the second field signifies.

crates/router/src/services/refresh_token.rs
```diff
@@ -0,0 +1,149 @@
...
+                    // the access token
+                    let _ = db
+                        .set_access_token(
+                            merchant_id,
+                            connector.connector.id(),
+                            access_token.clone(),
+                        )
+                        .await
+                        .change_context(errors::ApiErrorResponse::InternalServerError)
+                        .attach_printable("DB error when setting the access token");
```
SanchithHegde, line 108: What would happen if an error occurs when persisting the access token in database?

crates/router/src/types.rs
```diff
@@ -299,6 +309,29 @@ impl ErrorResponse {
...
+            ConnectorAuthType::BodyKey { api_key, key1 } => Self {
+                app_id: api_key,
+                id: Some(key1),
+            },
+            ConnectorAuthType::SignatureKey { api_key, key1, .. } => Self {
+                app_id: api_key,
+                id: Some(key1),
+            },
+            ConnectorAuthType::NoKey => Self {
+                app_id: "".to_string(),
```
SanchithHegde, line 328: Use `String::new()` instead.

SanchithHegde (changes requested):

crates/router/src/services/refresh_token.rs
Narayanbhat166: It should not affect the payments since we already have the access token, so this error should be discarded

crates/router/src/core/refunds.rs
```diff
@@ -125,23 +125,38 @@ pub async fn trigger_refund_to_gateway(
...
+        services::execute_connector_processing_step(
+            state,
+            connector_integration,
+            &router_data,
+            payments::CallConnectorAction::Trigger,
+        )
+        .await
+        .map_err(|error| error.to_refund_failed_response())?
+    } else {
+        router_data.clone()
```
SanchithHegde, line 156: We can get rid of the clone.

crates/router/src/core/refunds.rs
```diff
@@ -273,22 +288,38 @@ pub async fn sync_refund_with_gateway(
...
```
SanchithHegde, line 319: Same here, we can get rid of the clone.

SanchithHegde (changes requested):

SanchithHegde (dismissed):

crates/router/src/db/merchant_connector_account.rs
```diff
@@ -1,13 +1,99 @@
...
+impl ConnectorAccessToken for Store {
+    async fn get_access_token(
+        &self,
+        merchant_id: &str,
+        connector_name: &str,
+    ) -> CustomResult<Option<types::AccessToken>, errors::StorageError> {
+        //TODO: Handle race condition
+        // This function should acquire a global lock on some resource, if access token is already
+        // being refreshed by other request then wait till it finishes and use the same access token
+        let key = format!("{merchant_id}_{connector_name}");
```
jarnura, line 39: we need to add feature prefix while fetching from config map

crates/router/src/db/merchant_connector_account.rs
```diff
@@ -1,13 +1,99 @@
...
+        Ok(access_token)
+    }
+
+    async fn set_access_token(
+        &self,
+        merchant_id: &str,
+        connector_name: &str,
+        access_token: types::AccessToken,
+    ) -> CustomResult<(), errors::StorageError> {
+        let conn = self.redis_conn.clone();
```
jarnura, line 62: can you try calling `set_key_with_expiry` without cloning

crates/router/src/services/refresh_token.rs
```diff
@@ -0,0 +1,151 @@
...
+    Res: Debug + Clone + 'static,
+>(
+    state: &AppState,
+    connector: &api_types::ConnectorData,
+    merchant_account: &storage::MerchantAccount,
+    router_data: &types::RouterData<F, Req, Res>,
+) -> RouterResult<types::AddAccessTokenResult> {
+    if connector.connector_name.supports_access_token() {
+        let merchant_id = &merchant_account.merchant_id;
+        let db = &*state.store;
```
jarnura, line 63: use `store` as variable name instead of `db`

crates/router/src/services/refresh_token.rs
```diff
@@ -0,0 +1,151 @@
...
+                    merchant_account,
+                    &refresh_token_router_data,
+                )
+                .await?
+                .async_map(|access_token| async {
+                    //Store the access token in db
+                    let db = &*state.store;
+                    // This error should not be propagated, we don't want payments to fail once we have
+                    // the access token, the next request will create new access token
+                    let _ = db
```
jarnura, line 96: let _ is not required

crates/router/src/services/refresh_token.rs
```diff
@@ -0,0 +1,151 @@
...
+                refresh_connector_auth(
+                    state,
+                    connector,
+                    merchant_account,
+                    &refresh_token_router_data,
+                )
+                .await?
+                .async_map(|access_token| async {
+                    //Store the access token in db
+                    let db = &*state.store;
```
jarnura, line 93: use `store` as variable name instead db

crates/router/src/types.rs
```diff
@@ -299,6 +314,29 @@ impl ErrorResponse {
...
+            },
+            ConnectorAuthType::BodyKey { api_key, key1 } => Self {
+                app_id: api_key,
+                id: Some(key1),
+            },
+            ConnectorAuthType::SignatureKey { api_key, key1, .. } => Self {
+                app_id: api_key,
+                id: Some(key1),
+            },
+            ConnectorAuthType::NoKey => Self {
```
jarnura, line 332: when this case will happen?

jarnura (changes requested):

crates/router/src/db/merchant_connector_account.rs
Narayanbhat166: This is reading directly from redis. Do we need that here?

crates/router/src/services/refresh_token.rs
Narayanbhat166: let _ is required or else cargo will complain about result not being used

crates/router/src/types.rs
Narayanbhat166: This case will not happen and should not happen, should this be a try_from instead of from and raise an error?

SanchithHegde (approved):

crates/router/src/db/merchant_connector_account.rs
jarnura: yes, for eg: mid123_stripe, without feature, any feature can override this value and we will runtime bug

jarnura (changes requested):

SanchithHegde (approved):

jarnura (approved):

crates/router/src/types.rs
jarnura: Handle as a error

jarnura (changes requested):

SanchithHegde (approved):

jarnura (approved):

crates/router/src/core/refunds.rs
```diff
@@ -125,23 +126,38 @@ pub async fn trigger_refund_to_gateway(
...
-    > = connector.connector.get_connector_integration();
-    let router_data = services::execute_connector_processing_step(
-        state,
-        connector_integration,
-        &router_data,
-        payments::CallConnectorAction::Trigger,
-    )
-    .await
-    .map_err(|error| error.to_refund_failed_response())?;
+    let add_access_token_result =
```
NishantJoshi00, line 129: Hey @jarnura, 
This method calls the connector for access token, in this scenario, won't the discussion we have for #313 apply. 
Because, here we are conditionally calling the connector which will add a delay that is expected during a `execute_connector_processing_step`.

During the discussion we came to the conclusion that, we shouldn't add extra delay for certain connectors as it will affect the overall time of refunds API?
Can you please give me further insights about the decision that was made for this PR.

jagan-jaya:
Closes #185

PR #408, opened 2023-01-18
feat(globalpay): implement access token creation

Narayanbhat166:
## Type of Change

- [x] New feature

## Description

- This PR will add support for creating access tokens for globalpay connector.
- Add SHA512 digest generation

## Motivation and Context

Globalpay connector requires access token to be sent in each request. 

## How did you test it?

- Create a merchant account with incorrect credentials, and try to create payment. Creating access token fails and payments creation is bypassed.
Screenshot 2023-01-20 at 5 38 38 PM

- Create a payment with correct credentials. status is succeeded.
Screenshot 2023-01-18 at 5 42 53 PM

Access token being stored in redis
`"{\"token\":\"oxefIdN70QcBPRy75Yu0b3jnLFji\",\"expires\":86399}"`

## Checklist


crates/router/src/connector/globalpay.rs
```diff
@@ -116,7 +183,19 @@ impl ConnectorIntegration<api::Void, types::PaymentsCancelData, types::PaymentsR
...
+            .clone()
+            .ok_or(errors::ConnectorError::FailedToObtainAuthType)?;
+
+        let access_token_header = (
+            headers::AUTHORIZATION.to_string(),
+            format!("Bearer {}", access_token.token),
+        );
+
+        headers.push(access_token_header);
+        Ok(headers)
```
SanchithHegde, line 198: This snippet seems to be repetitive. We could extract this to a function instead.

crates/router/src/connector/globalpay/transformers.rs
```diff
@@ -72,21 +79,70 @@ impl TryFrom<&types::PaymentsCancelRouterData> for GlobalpayPaymentsRequest {
...
+            .attach_printable("Could not convert connector_auth to globalpay_auth")?;
+
+        let current_time = date_time::now().to_string();
+        let nonce_with_api_key = format!("{}{}", current_time, globalpay_auth.key);
+        let secret_vec = crypto::Sha512
+            .generate_digest(nonce_with_api_key.as_bytes())
+            .change_context(errors::ConnectorError::RequestEncodingFailed)
+            .attach_printable("error creating request nonce")?;
+
+        let secret = hex::encode(secret_vec);
```
SanchithHegde, line 135: Have they specified that the nonce must be the current time, and that the format/contents of the secret doesn't matter?

Preferably, if you need random data, you should be using one of the RNGs from the `rand` crate.

SanchithHegde (changes requested):

crates/router/src/connector/globalpay/transformers.rs
```diff
@@ -69,21 +77,70 @@ impl TryFrom<&types::PaymentsCancelRouterData> for GlobalpayPaymentsRequest {
...
 
 impl TryFrom<&types::ConnectorAuthType> for GlobalpayAuthType {
     type Error = error_stack::Report<errors::ConnectorError>;
     fn try_from(auth_type: &types::ConnectorAuthType) -> Result<Self, Self::Error> {
         match auth_type {
-            types::ConnectorAuthType::HeaderKey { api_key } => Ok(Self {
-                api_key: api_key.to_string(),
+            types::ConnectorAuthType::BodyKey { api_key, key1 } => Ok(Self {
+                app_id: api_key.to_string(),
+                key: key1.to_string(),
```
jagan-jaya, line 90: map app_id with key1 and api_key with key

jagan-jaya (changes requested):

crates/router/src/connector/globalpay/transformers.rs
Narayanbhat166: No they haven't specified that, but isn't using current time faster than creating random numbers?

SanchithHegde: If I remember correctly, there was a discussion about this involving @vspecky. If you need random numbers, then use random numbers. Don't substitute them with timestamp.

vspecky: Yes, using system time can have two disadvantages :-
- It can be slower than your RNG since it needs to make a syscall (usually insignificant)
- The system time on the server does not move forward at a constant speed. When the server is synchronizing time with a reference server using NTP, the system time may speed up, slow down or even jump around, so it's not guaranteed to be completely random. I don't know if in this case it could be a problem but generally with cryptography, it's better if we go with an RNG since it will generally be "more random" than system time.

jagan-jaya (dismissed):

SanchithHegde (approved):

crates/common_utils/src/crypto.rs
```diff
@@ -215,6 +215,23 @@ impl DecodeMessage for GcmAes256 {
...
 }
 
+/// Secure Hash Algorithm 512
+#[derive(Debug)]
+pub struct Sha512;
+
+/// Trait for generating a digest for SHA
+pub trait GenerateDigest {
+    /// takes a message and creates a digest for it
+    fn generate_digest(&self, message: &[u8]) -> CustomResult<Vec<u8>, errors::CryptoError>;
```
NishantJoshi00, line 225: can we have the output of this function as `&[u8]`?

NishantJoshi00 (changes requested):
Other than that, looks good ⚡

crates/common_utils/src/crypto.rs
Narayanbhat166: Cannot return a reference of owned value from a function!

NishantJoshi00 (approved):

SanchithHegde (approved):

PR #411, opened 2023-01-18
fix: update test_mode and payment_methods_enabled in payments connector

Sangamesh26:
## Type of Change

- [x] Bugfix

## Description

1> Test mode was not getting updated in payments connector update, it is handled now
2>  when Payments_method_enabled was not passed in update call, it used to not show up the earlier payments_method_enabled, it is handled now
3> If payments_method_enabled is passed, then in the retrieve call, it used to show up the both earlier payments_method_enabled as well as present payments_method_enabled instead of just showing up the updated payments_method_enabled, it is handled now.

## Motivation and Context

Payments connector update and retrieve call proper working

## How did you test it?

Manual

## Checklist


crates/router/src/core/admin.rs
```diff
@@ -393,6 +393,7 @@ pub async fn update_payment_connector(
 
     let payment_methods_enabled = match req.payment_methods_enabled.clone() {
         Some(val) => {
+            vec.clear();
```
jarnura, line 396: why we need to clear the vec?

jarnura (dismissed):

dracarys18 (dismissed):

dracarys18 (approved):

jarnura (approved):

SanchithHegde (approved):

PR #413, opened 2023-01-18
feat(router): reuse reqwest clients for requests made to connectors

lsampras:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [x] Refactoring
- [ ] Dependency updates

## Description

Reuse reqwest clients for performance benefits
Have added caching for non-certificate authenticated clients using once cell
Moved request timeout parameters to request builder instead of client

### Additional Changes

## Motivation and Context

Creating a client has a significant overhead, reusing the clients should increase perf

## Checklist

### DIFF
--- a/crates/router/src/services/api/client.rs
+++ b/crates/router/src/services/api/client.rs
@@ -36,32 +36,33 @@ impl ProxyType {
 }
 
 fn create_base_client(
-   proxy: Option<(ProxyType, String)> 
+    proxy: Option<(ProxyType, String)>,
 ) -> CustomResult<reqwest::Client, errors::ApiClientError> {
-
-   Ok(match proxy {
+    Ok(match proxy {
         None => &PLAIN_CLIENT,
         Some((ProxyType::Http, _)) => &HTTP_PROXY_CLIENT,
         Some((ProxyType::Https, _)) => &HTTPS_PROXY_CLIENT,
-    }.get_or_try_init(|| {
-        let mut cb = reqwest::Client::builder()
-            .redirect(reqwest::redirect::Policy::none());
+    }
+    .get_or_try_init(|| {
+        let mut cb = reqwest::Client::builder().redirect(reqwest::redirect::Policy::none());
         cb = match proxy {
             None => cb,
             Some((proxy_type, url)) => cb.proxy(
-                    match proxy_type {
-                        ProxyType::Http => reqwest::Proxy::http(url),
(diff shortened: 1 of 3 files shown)


crates/router/src/services/api/client.rs
```diff
@@ -32,15 +35,59 @@ impl ProxyType {
...
 
+fn create_base_client(
+    proxy: Option<(ProxyType, String)>,
+) -> CustomResult<reqwest::Client, errors::ApiClientError> {
+    Ok(match proxy {
+        None => &PLAIN_CLIENT,
+        Some((ProxyType::Http, _)) => &HTTP_PROXY_CLIENT,
+        Some((ProxyType::Https, _)) => &HTTPS_PROXY_CLIENT,
+    }
+    .get_or_try_init(|| {
```
aravindgopall, line 46: By default reqwest library maintains a connection pool. Have you checked with traffic how many connections are we really opening or are they being reused. I am guessing Just clone would do the work and no need of onecell. Can you confirm this once?

aravindgopall (approved):

crates/router/src/services/api/client.rs
lsampras: Yeah cloning would work since `reqwest::client` internally uses an Arc,

Using OnceCell to have the initial client reference & keep the client alive ,

The alternative could be that we have clients defined at the start of the application and pass them along with app state,

Using OnceCell for now to reduce overall changes in the code, If this works without any bugs we can extend it maybe


Lazy_CLIENT (static reference to client) -> Clone (reuse the client since it internally uses an Arc) -> use the cloned client for individual requests...

I'll get more insights into the  connections (I believe we also need to enable [tcp_keepalive](https://docs.rs/reqwest/latest/reqwest/struct.ClientBuilder.html#method.tcp_keepalive) for reusing connections).

For now the main purpose of this PR was to prevent any CPU overhead from creating new clients,
will address the system resource optimizations later

aravindgopall: > I'll get more insights into the connections (I believe we also need to enable [tcp_keepalive](https://docs.rs/reqwest/latest/reqwest/struct.ClientBuilder.html#method.tcp_keepalive) for reusing connections).
Cool. Let me know, We can have a comment added to code and this PR can be merged

lsampras: created an [issue](https://github.com/juspay/hyperswitch/issues/424) for this.

PR #415, opened 2023-01-18
feat: add connector error

NishantJoshi00:
## Type of Change

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates

## Description
This PR introduces `ExternalConnectorError` into `ApiErrorResponse` which allows for status code propagation and providing error when a payment is failed

If a payment is failed from the connector side, an error in the standard error format will be displayed with 
Screenshot 2023-01-19 at 12 41 17 AM

Here, the error is in the format
```rust
errors::ApiErrorResponse::ExternalConnectorError {
 	message: "card details verification failed".to_string(),
    status_code: 500,
}
```

### Additional Changes

## Checklist

### DIFF
--- a/crates/router_derive/src/macros/api_error/helpers.rs
+++ b/crates/router_derive/src/macros/api_error/helpers.rs
@@ -14,6 +14,7 @@ mod keyword {
     custom_keyword!(error_type);
     custom_keyword!(code);
     custom_keyword!(message);
+    custom_keyword!(ignore);
 }
 
 enum EnumMeta {
@@ -107,6 +108,10 @@ enum VariantMeta {
         keyword: keyword::message,
         value: LitStr,
     },
+    Ignore {
+        keyword: keyword::ignore,
+        value: LitStr,
+    },
 }
 
 impl Parse for VariantMeta {
@@ -127,6 +132,11 @@ impl Parse for VariantMeta {
             let _: Token![=] = input.parse()?;
             let value = input.parse()?;
             Ok(Self::Message { keyword, value })
+        } else if lookahead.peek(keyword::ignore) {
+            let keyword = input.parse()?;
+            let _: Token![=] = input.parse()?;
+            let value = input.parse()?;
+            Ok(Self::Ignore { keyword, value })
         } else {
             Err(lookahead.error())
         }
@@ -139,6 +149,7 @@ impl Spanned for VariantMeta {
             Self::ErrorType { keyword, .. } => keyword.span,
             Self::Code { keyword, .. } => keyword.span,
             Self::Message { keyword, .. } => keyword.span,
+            Self::Ignore { keyword, .. } => keyword.span,
         }
     }
 }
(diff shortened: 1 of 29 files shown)


crates/router_derive/src/macros/api_error/helpers.rs
```diff
@@ -227,11 +252,25 @@ pub(super) fn check_missing_attributes(
...
         syn::Fields::Unnamed(_) => Vec::new(),
-        syn::Fields::Named(fields) => fields.named.iter().cloned().collect(),
+        syn::Fields::Named(fields) => fields
+            .named
+            .iter()
+            .filter(|field| {
+                field
+                    .ident
+                    .as_ref()
+                    .map_or(true, |idn| ignore.contains(&idn.to_string()))
```
SanchithHegde, line 270: If it makes it any simpler, you can also update the filtering condition (after the match block) to be:

```rust
!(message.contains(&field_name) && ignore.contains(&field_name))
```

crates/router_derive/src/macros/api_error/helpers.rs
```diff
@@ -245,3 +284,8 @@ pub(super) fn get_unused_fields(fields: &Fields, message: &str) -> Vec<Field> {
         .cloned()
         .collect()
 }
+
+// fn allow_field(ident: &Ident) -> bool {
+//     // This will allow fields with underscore at the end to be hidden during serialization
+//     !ident.to_string().ends_with("_")
+// }
```
SanchithHegde, line 291: And yeah, remove this commented function.

SanchithHegde (changes requested):
Apart from that, looks good to me!

crates/router_derive/src/macros/api_error/helpers.rs
```diff
@@ -260,32 +260,16 @@ pub(super) fn get_unused_fields(
...
     };
-
     fields
         .iter()
         .filter(|&field| {
             // Safety: Named fields are guaranteed to have an identifier.
             #[allow(clippy::unwrap_used)]
             let field_name = format!("{}", field.ident.as_ref().unwrap());
-            !message.contains(&field_name)
+            !message.contains(&field_name) && !ignore.contains(&field_name)
```
SanchithHegde, line 271: This should be either:

```rust
!(message.contains(&field_name) && ignore.contains(&field_name))
```

```rust
!message.contains(&field_name) || !ignore.contains(&field_name)
```

NishantJoshi00: let's consider the example, field `data` exists in message, then it should be `false`. If it doesn't exist in the message, then we check if it exists in ignore, if yes, then `false`. Which translates to `(!message.contains(&field_name))` this is the first condition, if this doesn't short circuit, then evaluate the second condition `(!ignore.contains(&field_name))`

This translates to:
either
```rust
!message.contains(&field_name) && !ignore.contains(&field_name)
```
or
```rust
!(message.contains(&field_name) || ignore.contains(&field_name))

SanchithHegde: My bad, you're right.

SanchithHegde (approved):

jarnura (approved):

PR #428, opened 2023-01-19
fix: add graceful shutdown for consumer & router

NishantJoshi00:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring
- [ ] Dependency updates

## Description

This PR adds graceful shutdown to consumer and ands configurability to the shutdown timeout for actix web server.

### Additional Changes

This change adds an extra configuration to `[server]`, it has a default value of `30` seconds. The configuration looks like:
> Example
> ```toml
> [server]
> shutdown_timeout = 30
> ```

## Motivation and Context
N/A

## Checklist


crates/router/src/scheduler/utils.rs
```diff
@@ -355,3 +358,23 @@ where
     };
     result
 }
+
+pub(crate) async fn signal_handler(
+    mut sig: signal_hook_tokio::Signals,
+    sender: oneshot::Sender<()>,
+) {
+    if let Some(signal) = sig.next().await {
+        logger::debug!("Requested a force shutdown");
```
lsampras, line 367: Move this to Ok() match clause & 
Log received sigterm/sigint here...

crates/router/src/scheduler/consumer.rs
```diff
@@ -62,6 +77,13 @@ pub async fn start_consumer(
                 },
                 sync::Arc::clone(&consumer_operation_counter),
             ));
+
+            match rx.try_recv() {
```
lsampras, line 81: We could use this as the root match instead of if is_ready

Also add separate loggers for Ok() and receiver closed error

lsampras (approved):

lsampras (dismissed):

lsampras (dismissed):

jarnura (dismissed):

SanchithHegde (approved):

SanchithHegde (changes requested):
My bad, I meant to request changes.

crates/router/src/scheduler/utils.rs
```diff
@@ -355,3 +358,28 @@ where
...
+) {
+    if let Some(signal) = sig.next().await {
+        logger::debug!(
+            "Received signal: {:?}",
+            signal_hook::low_level::signal_name(signal)
+        );
+        match signal {
+            signal_hook::consts::SIGTERM | signal_hook::consts::SIGINT => match sender.send(()) {
+                Ok(_) => {
+                    logger::debug!("Request for force shutdown received")
```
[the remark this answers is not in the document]
NishantJoshi00: sounds good 👍

crates/router/src/scheduler/consumer.rs
```diff
@@ -234,9 +234,9 @@ pub async fn some_error_handler<E: fmt::Display>(
     process: storage::ProcessTracker,
     error: E,
 ) -> CustomResult<(), errors::ProcessTrackerError> {
-    error!(%process.id, "Failed while executing workflow");
-    error!(%error);
-    error!(
+    logger::error!(%process.id, "Failed while executing workflow");
+    logger::error!(%error);
+    logger::error!(
```
SanchithHegde, line 239: Could you update this block to be logging a single log event, and use one of the error messages? And maybe also rename the function to remove the `some` I added when I wrote it 😬 ?
```rust
logger::error!(pt.name = ?process.name, pt.id = %process.id, %error, "Failed while executing workflow");
```

NishantJoshi00: Sure, your suggestion looks more cleaner in this case

SanchithHegde (approved):

jarnura (approved):

lsampras (approved):

PR #431, opened 2023-01-19
refactor(core): use previously used connector

Narayanbhat166:
## Type of Change

- [x] Refactoring

## Description

This PR will refactor the `get_connector` method to use previously used connector whenever necessary.

## Motivation and Context

For any of the later flows after creating the payment, the connector used should be the same from which the payment was made instead of using the routing algorithm. For confirm flow you can choose to override the connector that was used in payment. 

## How did you test it?

- Create a payment which requires capture action
Screenshot 2023-01-19 at 7 38 33 PM
- Capture the payment.
Screenshot 2023-01-19 at 7 39 44 PM

## Checklist


crates/router/src/core/payments/operations/payment_confirm.rs
```diff
@@ -253,8 +253,16 @@ impl<F: Clone + Send> Domain<F, api::PaymentsRequest> for PaymentConfirm {
...
     ) -> CustomResult<api::ConnectorCallType, errors::ApiErrorResponse> {
-        helpers::get_connector_default(state, request.connector).await
+        // Use a new connector in the confirm call or use the same one which was passed when
+        // creating the payment or if none is passed then use the routing algorithm
+        let request_connector = request.connector.map(|connector| connector.to_string());
+        helpers::get_connector_default(
+            state,
+            request_connector.as_ref().or(previously_used_connector),
+        )
+        .await
```
lsampras, line 265: Didn't understand this part why do we get an owned object using to_string() just to use as_ref() and convert it back to a immutable ref

crates/router/src/core/payments/operations/payment_create.rs
```diff
@@ -257,8 +257,10 @@ impl<F: Clone + Send> Domain<F, api::PaymentsRequest> for PaymentCreate {
         _merchant_account: &storage::MerchantAccount,
         state: &AppState,
         request: &api::PaymentsRequest,
+        _previously_used_connector: Option<&String>,
     ) -> CustomResult<api::ConnectorCallType, errors::ApiErrorResponse> {
-        helpers::get_connector_default(state, request.connector).await
+        let request_connector = request.connector.map(|connector| connector.to_string());
+        helpers::get_connector_default(state, request_connector.as_ref()).await
```
lsampras, line 263: Why do we get an owned object using to_string() just to use as_ref() and convert it back to a immutable ref

lsampras (changes requested):

crates/router/src/core/payments/operations/payment_confirm.rs
Narayanbhat166: Connector is an enum, to get the string representation of it `.to_string()` is being used

jarnura (approved):

jarnura (approved):

SanchithHegde (approved):

lsampras (approved):

PR #435, opened 2023-01-20
Rapyd webhook integration

manoj-juspay:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates

## Description
Integrate payment webhook for rapyd (only payment webhook)
Integrated Psync

### Additional Changes

## How did you test it?
Configuring the webhook url on rapyd dashboard to our local url

## Checklist


crates/router/src/connector/rapyd.rs
```diff
@@ -633,27 +634,108 @@ impl services::ConnectorIntegration<api::RSync, types::RefundsData, types::Refun
...
+                api::IncomingWebhookEvent::PaymentIntentSuccess
+            }
+            transformers::RapydWebhookObjectEventType::PaymentCaptured => {
+                api::IncomingWebhookEvent::PaymentIntentSuccess
+            }
+            transformers::RapydWebhookObjectEventType::PaymentFailed => {
+                api::IncomingWebhookEvent::PaymentIntentFailure
+            }
+            _ => Err(errors::ConnectorError::WebhookEventTypeNotFound).into_report()?,
+        })
```
SanchithHegde, line 720: Could you move this to a `From` implementation?

SanchithHegde (changes requested):

crates/router/src/connector/rapyd/transformers.rs
```diff
@@ -572,9 +572,51 @@ pub enum RapydWebhookObjectEventType {
...
+                error_code: "".to_owned(),
+                status: "SUCCESS".to_owned(),
+                message: None,
+                response_code: None,
+                operation_id: None,
+            },
+            data: Some(value),
+        }
+    }
+}
```
SanchithHegde, line 679: Where are these 2 implementations being used?

crates/router/src/connector/rapyd.rs
```diff
@@ -331,33 +332,83 @@ impl
...
+        let signature = self.generate_signature(&auth, "get", &url_path, "", &timestamp, &salt)?;
+
+        let headers = vec![
+            ("access_key".to_string(), auth.access_key),
+            ("salt".to_string(), salt),
+            ("timestamp".to_string(), timestamp.to_string()),
+            ("signature".to_string(), signature),
+        ];
+        let request = services::RequestBuilder::new()
+            .method(services::Method::Delete)
```
SanchithHegde, line 372: Do you need the `DELETE` request method to be used for payments sync?

crates/router/src/connector/rapyd/transformers.rs
```diff
@@ -550,6 +550,63 @@ impl TryFrom<types::PaymentsCancelResponseRouterData<RapydPaymentsResponse>>
...
 
+impl TryFrom<types::PaymentsSyncResponseRouterData<RapydPaymentsResponse>>
+    for types::PaymentsSyncRouterData
+{
+    type Error = error_stack::Report<errors::ParsingError>;
+    fn try_from(
+        item: types::PaymentsSyncResponseRouterData<RapydPaymentsResponse>,
+    ) -> Result<Self, Self::Error> {
+        let (status, response) = match item.response.status.status.as_str() {
+            "SUCCESS" => match item.response.data {
```
SanchithHegde, line 561: Can we use enums instead of strings for statuses?

crates/router/src/connector/rapyd/transformers.rs
manoj-juspay: https://github.com/juspay/hyperswitch/pull/435/files#diff-86fe3cf6ada14791544fffd23177175052f093813e48a2457e3ae6931810cb20R825

crates/router/src/connector/rapyd/transformers.rs
manoj-juspay: this response format is from postman couldn't find in rapyd reference, we discussed in previous pr for rapyd

crates/router/src/connector/rapyd/transformers.rs
manoj-juspay: will use this when implementing refund webhook

manoj-juspay: https://github.com/juspay/hyperswitch/pull/435/files#diff-86fe3cf6ada14791544fffd23177175052f093813e48a2457e3ae6931810cb20R825

SanchithHegde (dismissed):

crates/router/src/connector/utils.rs
```diff
@@ -175,3 +176,20 @@ impl AddressDetailsData for api::AddressDetails {
             .ok_or_else(missing_field_err("address.country"))
     }
 }
+
+pub fn get_header_key_value(
+    key: &str,
+    headers: &actix_web::http::header::HeaderMap,
+) -> CustomResult<String, errors::ConnectorError> {
+    headers
+        .get(key)
```
jarnura, line 185:
```
	header_value
		.to_str()
		.map(String::from)
		    -- we already getting reference of HeaderMap converting to string is not required.
        
        .map_err(|_| errors::ConnectorError::WebhookSignatureNotFound)
        	-- here we are discarding the error we, should not do that.
        	-- here the error is WebhookSourceVerificationFailed or parsing failed.
        	
         .into_report() 
         	-- we need to get to_str() as report and we need to change the context.
        })
        .ok_or(errors::ConnectorError::WebhookSourceVerificationFailed) 
        	-- this is WebhookSignatureNotFound
        	-- attach printable too
        	
        .into_report()? 
        	-- we need to change the error context , not generating one here
```

crates/router/src/connector/utils.rs
```diff
@@ -175,3 +176,20 @@ impl AddressDetailsData for api::AddressDetails {
             .ok_or_else(missing_field_err("address.country"))
     }
 }
+
+pub fn get_header_key_value(
+    key: &str,
+    headers: &actix_web::http::header::HeaderMap,
+) -> CustomResult<String, errors::ConnectorError> {
+    headers
```
jarnura, line 184:
```suggestion
    headers
        .get(key)
        .map(|header_value| {
            header_value
                .to_str()
                .into_report()
                .change_context(errors::ConnectorError::WebhookSignatureNotFound)
        })
        .ok_or(report!(errors::ConnectorError::WebhookSourceVerificationFailed))?
```

crates/router/src/connector/rapyd/transformers.rs
```diff
@@ -548,3 +616,74 @@ impl TryFrom<types::PaymentsCancelResponseRouterData<RapydPaymentsResponse>>
...
+            RapydWebhookObjectEventType::PaymentCompleted => Ok(Self::PaymentIntentSuccess),
+            RapydWebhookObjectEventType::PaymentCaptured => Ok(Self::PaymentIntentSuccess),
+            RapydWebhookObjectEventType::PaymentFailed => Ok(Self::PaymentIntentFailure),
+            _ => Err(errors::ConnectorError::WebhookEventTypeNotFound).into_report()?,
+        }
+    }
+}
+
+#[derive(Debug, Deserialize)]
+#[serde(untagged)]
```
jarnura, line 655: try to remove the `untagged` Deserialize

crates/router/src/connector/rapyd/transformers.rs
```diff
@@ -548,3 +616,74 @@ impl TryFrom<types::PaymentsCancelResponseRouterData<RapydPaymentsResponse>>
...
+    PaymentData(ResponseData),
+    RefundData(RefundResponseData),
+}
+
+impl From<ResponseData> for RapydPaymentsResponse {
+    fn from(value: ResponseData) -> Self {
+        Self {
+            status: Status {
+                error_code: "".to_owned(),
+                status: "SUCCESS".to_owned(),
```
jarnura, line 666: why we are hard coding `SUCCESS`?

Status of connector can't be any string. There will be limited set of status right?

crates/router/src/connector/rapyd/transformers.rs
```diff
@@ -548,3 +616,74 @@ impl TryFrom<types::PaymentsCancelResponseRouterData<RapydPaymentsResponse>>
...
+pub enum WebhookData {
+    PaymentData(ResponseData),
+    RefundData(RefundResponseData),
+}
+
+impl From<ResponseData> for RapydPaymentsResponse {
+    fn from(value: ResponseData) -> Self {
+        Self {
+            status: Status {
+                error_code: "".to_owned(),
```
jarnura, line 665: error code can't be empty we having for error_code and error_message, check `src/consts.rs` file.

crates/router/src/connector/rapyd/transformers.rs
```diff
@@ -548,3 +616,74 @@ impl TryFrom<types::PaymentsCancelResponseRouterData<RapydPaymentsResponse>>
...
+        }
+    }
+}
+
+impl From<RefundResponseData> for RefundResponse {
+    fn from(value: RefundResponseData) -> Self {
+        Self {
+            status: Status {
+                error_code: "".to_owned(),
+                status: "SUCCESS".to_owned(),
```
jarnura, line 681: why we are hard coding as `SUCCESS`?

crates/router/src/connector/rapyd/transformers.rs
```diff
@@ -442,44 +447,46 @@ impl TryFrom<types::PaymentsCaptureResponseRouterData<RapydPaymentsResponse>>
...
-                    Err(types::ErrorResponse {
-                        code: item.response.status.error_code,
-                        message: item.response.status.status,
-                        reason: item.response.status.message,
-                        status_code: item.http_code,
-                    }),
-                ),
-            },
-            "ERROR" => (
+        let (status, response) = match &item.response.data {
```
jarnura, line 450: `RapydPaymentsResponse` is same for mostly of the api's the handling also not differs, write a common function and use it.

crates/router/src/connector/rapyd.rs
```diff
@@ -642,25 +693,145 @@ impl services::ConnectorIntegration<api::RSync, types::RefundsData, types::Refun
...
+            .change_context(errors::ConnectorError::WebhookSourceVerificationFailed)?;
+        Ok(signature)
+    }
+
+    async fn get_webhook_source_verification_merchant_secret(
+        &self,
+        db: &dyn StorageInterface,
+        merchant_id: &str,
+    ) -> CustomResult<Vec<u8>, errors::ConnectorError> {
+        let key = format!("whsec_verification_{}_{}", self.id(), merchant_id);
```
jarnura, line 722: from prefix we need to get the whole picture about what it does, so prefer something like
`wh_msec_verification_` or `wh_mer_sec_verification_` or something like this.

crates/router/src/connector/rapyd.rs
```diff
@@ -642,25 +693,145 @@ impl services::ConnectorIntegration<api::RSync, types::RefundsData, types::Refun
...
+            .change_context(errors::ConnectorError::WebhookSourceVerificationFailed)?;
+        let message = self
+            .get_webhook_source_verification_message(headers, body, merchant_id, &secret)
+            .change_context(errors::ConnectorError::WebhookSourceVerificationFailed)?;
+
+        let stringify_auth = String::from_utf8(secret.to_vec())
+            .into_report()
+            .change_context(errors::ConnectorError::WebhookSourceVerificationFailed)
+            .attach_printable("Could not convert secret to UTF-8")?;
+        let auth: transformers::RapydAuthType = stringify_auth
```
jarnura, line 783: We are parsing this twice, we can have some kind of state pattern so that we can eliminate, we can take this in different PR also.

crates/router/src/connector/rapyd.rs
```diff
@@ -642,25 +693,145 @@ impl services::ConnectorIntegration<api::RSync, types::RefundsData, types::Refun
...
+            .parse_struct("RapydIncomingWebhook")
+            .change_context(errors::ConnectorError::WebhookEventTypeNotFound)?;
+        let response = match webhook.data {
+            transformers::WebhookData::PaymentData(payment_data) => {
+                let rapyd_response: transformers::RapydPaymentsResponse = payment_data.into();
+                Ok(rapyd_response)
+            }
+            _ => Err(errors::ConnectorError::WebhookEventTypeNotFound),
+        }?;
+        let res_json = serde_json::to_value(response)
```
jarnura, line 830: check in common_utils crate, ext_traits.rs
check any common utility function which does the same.

jarnura (changes requested):

crates/router/src/connector/rapyd.rs
```diff
@@ -329,33 +332,83 @@ impl
...
+        let response_id = req.request.connector_transaction_id.clone();
+        let url_path = format!(
+            "/v1/payments/{}",
+            response_id
+                .get_connector_transaction_id()
+                .change_context(errors::ConnectorError::MissingConnectorTransactionID)?
+        );
+        let signature = self.generate_signature(&auth, "get", &url_path, "", &timestamp, &salt)?;
+
+        let headers = vec![
```
jagan-jaya, line 365: move this header building part to common place as it is exactly being used in many places
refer https://github.com/juspay/hyperswitch/blob/b5720f1e3801b0b9965de8d330284f5d2daec9e7/crates/router/src/connector/cybersource.rs#L108

crates/router/src/connector/rapyd.rs
```diff
@@ -329,33 +332,83 @@ impl
...
+        let request = services::RequestBuilder::new()
+            .method(services::Method::Get)
+            .url(&types::PaymentsSyncType::get_url(self, req, connectors)?)
+            .headers(types::PaymentsSyncType::get_headers(self, req, connectors)?)
+            .headers(headers)
+            .build();
+        Ok(Some(request))
     }
 
     fn get_error_response(
```
jagan-jaya, line 380: Move the error_response handling logic to common place and use it wherever needed
https://github.com/juspay/hyperswitch/blob/b5720f1e3801b0b9965de8d330284f5d2daec9e7/crates/router/src/connector/cybersource.rs#L87

jagan-jaya (changes requested):

crates/router/src/connector/rapyd/transformers.rs
manoj-juspay: i couldn't find it in there doc, this is from the sample implementation where its defined as string. anyway in case we get the data field this status comes as `SUCCESS` which mean the api was a success

crates/router/src/connector/rapyd/transformers.rs
manoj-juspay: i couldn't find it in there doc, this is from the sample implementation where its defined as string. anyway in case we get the data field this status comes as `SUCCESS` which mean the api was a success

crates/router/src/connector/rapyd.rs
manoj-juspay: will take it up in differnet PR

crates/router/src/connector/utils.rs
manoj-juspay: `?` operator cannot convert from `Result<&str, _>` to `Result<std::string::String, _>`
`map(String::from)` is required

crates/router/src/connector/rapyd/transformers.rs
manoj-juspay: ok

crates/router/src/connector/rapyd.rs
manoj-juspay: okay

crates/router/src/connector/rapyd.rs
manoj-juspay: sure

crates/router/src/connector/rapyd/transformers.rs
manoj-juspay: untagged is needed

crates/router/src/connector/rapyd/transformers.rs
manoj-juspay: sure

crates/router/src/connector/rapyd.rs
manoj-juspay: sure

crates/router/src/connector/rapyd/transformers.rs
manoj-juspay: sure

crates/router/src/connector/rapyd.rs
manoj-juspay: its not exactly same as it is for cybersource, keeping it as it is

crates/router/src/connector/rapyd.rs
manoj-juspay: sure

crates/router/src/connector/utils.rs
jarnura: the return type need not be string it can be slice

jarnura (changes requested):

crates/router/src/connector/utils.rs
manoj-juspay: yes

SanchithHegde (approved):

crates/router/src/connector/rapyd/transformers.rs
```diff
@@ -435,51 +361,75 @@ impl TryFrom<&types::PaymentsCaptureRouterData> for CaptureRequest {
     }
 }
 
-impl TryFrom<types::PaymentsCaptureResponseRouterData<RapydPaymentsResponse>>
-    for types::PaymentsCaptureRouterData
+impl<F, T>
+    TryFrom<types::ResponseRouterData<F, RapydPaymentsResponse, T, types::PaymentsResponseData>>
```
jarnura, line 365: This is a generic response handler, and redirection handling also here. For all flows redirection handling won't be there in core. That won't be problem?

crates/router/src/connector/rapyd/transformers.rs
jarnura: What is this status field, What if the api is success and status is failure?

jarnura: If the status is mentioned as string, we need to use their response status only, we can't override with a hardcoded status.

jarnura (changes requested):

crates/router/src/connector/rapyd/transformers.rs
manoj-juspay: No for all other flow this data will be null, And i have tested it for all the flow that we currently integrated as of today

crates/router/src/connector/rapyd/transformers.rs
manoj-juspay: we first check the status and update the payment status , there won't be a case in webhook where this hardcoded value is used to update status

jarnura (approved):

SanchithHegde (approved):

jagan-jaya (approved):

PR #439, opened 2023-01-20
feat(core): added multiple payment_attempt support for payment_intent

Abhicodes-crypto:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates

## Description

A column 'attempt_id' was added to payment_intent table while ensuring that the current kv or postgres flows don't get disturbed. Having multiple payment_attempts associated with a single payment_intent helps us implement 'Retries'. 

### Additional Changes

## How did you test it?

Manual, verified if the payment flows were working as before.
KV 
image
image

POSTGRES
image
image

## Checklist


migrations/2023-01-20-113235_add_attempt_id_to_payment_intent/up.sql
```diff
@@ -0,0 +1,4 @@
+-- Your SQL goes here
+ALTER TABLE PAYMENT_INTENT ADD COLUMN attempt_id VARCHAR(255) DEFAULT NULL;
+CREATE UNIQUE INDEX payment_attempt_attempt_id_index ON payment_attempt (attempt_id);
+DROP index payment_attempt_payment_id_merchant_id_index;
```
Abhicodes-crypto, line 4: Instead of dropping this index completely , should I maintain indexing on (payment_id, merchant_id) but it will not be unique. I guess this would help us in all payment_attempts related to an intent. 

Or should I create unique index on (payment_id, merchant_id, attempt_id). I'm not really sure which one would be efficient.

crates/storage_models/src/payment_intent.rs
```diff
@@ -29,6 +29,7 @@ pub struct PaymentIntent {
     pub setup_future_usage: Option<storage_enums::FutureUsage>,
     pub off_session: Option<bool>,
     pub client_secret: Option<String>,
+    pub attempt_id: Option<String>,
```
Abhicodes-crypto, line 32: Should I remove payment attempt as optional and make it mandatory. Or is there any case where we would create a payment_intent but not have the attempt_id ?

Abhicodes-crypto: I made it mandatory.

migrations/2023-01-20-113235_add_attempt_id_to_payment_intent/up.sql
Abhicodes-crypto: We are querying based on (merchant_id, attempt_id) , (merchant_id, payment_id),  (merchant_id, connector_txn_id), 
so thinking of adding indexes on them. Let me know your thoughts.

jarnura: `(merchant_id, attempt_id)`-  unqiue index.
`(merchant_id, payment_id) (merchant_id, connector_txn_id)` - no unqiue index

crates/storage_models/src/payment_intent.rs
jarnura: we need to made this as non optional.

migrations/2023-01-20-113235_add_attempt_id_to_payment_intent/up.sql
jarnura: We should not keep the value as NULL

jarnura: The datatype of the `attempt_id` field should be same the` attempt_id` field in `payment_attempt`. VARCHAR(64).

SanchithHegde (dismissed):

crates/router/src/core/payments/operations/payment_session.rs
```diff
@@ -70,9 +70,9 @@ impl<F: Send + Clone> GetTracker<F, PaymentData<F>, api::PaymentsSessionRequest>
         )?;
 
         let mut payment_attempt = db
-            .find_payment_attempt_by_payment_id_merchant_id(
-                &payment_id,
+            .find_payment_attempt_by_merchant_id_attempt_id(
```
jarnura, line 73: we need payment_id also to fetch a payment_attempt. Without payment_id uniqueness of attempt can't be ensured. And for making the SQL and NoSql interfaces same we need payment_id also. And we might need indexes also for the same

jarnura (changes requested):

crates/storage_models/src/payment_intent.rs
```diff
@@ -122,6 +125,7 @@ pub struct PaymentIntentUpdateInternal {
     pub billing_address_id: Option<String>,
     pub shipping_address_id: Option<String>,
     pub modified_at: Option<PrimitiveDateTime>,
+    pub attempt_id: String,
```
SanchithHegde, line 128: Make this an `Option<String>` instead. When you use any of the enums variants other than `PaymentAttemptUpdate`, you'll end up setting the `attempt_id` to an empty `String` because of the `Default::default()` being called.

crates/router/src/core/payments/operations.rs
```diff
@@ -91,6 +91,22 @@ pub trait GetTracker<F, D, R>: Send {
...
+        _request: &R,
+        _mandate_type: Option<api::MandateTxnType>,
+        _merchant_account: &storage::MerchantAccount,
+    ) -> RouterResult<(BoxedOperation<'a, F, R>, D, Option<CustomerDetails>)>
+    where
+        Self: Sized,
+    {
+        Err(report!(errors::ApiErrorResponse::InternalServerError))
+            .attach_printable("post connector update tracker not found")
+    }
```
SanchithHegde, line 109: Is this block supposed to go in the current PR?

SanchithHegde (changes requested):

crates/router/src/core/payments/operations.rs
Abhicodes-crypto: No. These changes were pushed accidentally. Will revert these.

crates/storage_models/src/payment_intent.rs
Abhicodes-crypto: Yes. 

Apologies this change was pushed accidentally. I made the same code changes in `for-retry` branch and already used `Option` there.

Abhicodes-crypto:
Will Reopen the PR when I'm done with all the changes.

SanchithHegde:
> Will Reopen the PR when I'm done with all the changes.

Converting this to a draft PR instead.

crates/router/src/core/payments.rs
```diff
@@ -918,3 +909,14 @@ pub fn decide_connector(
...
+#[inline]
+pub fn get_connector_data(
+    connectors: &mut IntoIter<api::ConnectorData>,
+) -> RouterResult<api::ConnectorData> {
+    connectors
+        .next()
+        .ok_or(errors::ApiErrorResponse::InternalServerError)
+        .into_report()
+        .attach_printable("Failed to get connector from iterator of connectors")
+}
```
vspecky, line 922: This function is no longer required right?

vspecky (changes requested):

vspecky (approved):
Overall LGTM. Ensure a couple of things
- Whether in update trackers and post update trackers, you have a check that prevents a payment attempt from being updated if it's not the current payment attempt as per the payment intent. This is important to ensure that the combination of auto retries and webhooks doesn't cause invalid updates.
- Check with @dracarys18 once regarding the reverse lookup logic.

migrations/2023-01-20-113235_add_attempt_id_to_payment_intent/up.sql
```diff
@@ -0,0 +1,11 @@
+-- Your SQL goes here
+ALTER TABLE payment_intent ADD COLUMN attempt_id VARCHAR(64) NOT NULL DEFAULT 'xxx';
+
+UPDATE payment_intent SET attempt_id = payment_attempt.attempt_id from payment_attempt where payment_intent.payment_id = payment_attempt.payment_id;
+
+CREATE UNIQUE INDEX payment_attempt_attempt_id_merchant_id_index ON payment_attempt (attempt_id, merchant_id);
```
jarnura, line 6: payment_id needs to be added.

migrations/2023-03-15-082312_add_connector_txn_id_merchant_id_index_in_payment_attempt/up.sql
```diff
@@ -0,0 +1,2 @@
+-- Your SQL goes here
+CREATE UNIQUE INDEX payment_attempt_connector_transaction_id_merchant_id_index ON payment_attempt (connector_transaction_id, merchant_id);
```
jarnura, line 2: this is not a unique index

crates/storage_models/src/query/payment_attempt.rs
```diff
@@ -92,38 +79,6 @@ impl PaymentAttempt {
         .await
     }
 
-    pub async fn find_last_successful_attempt_by_payment_id_merchant_id(
```
jarnura, line 95: we need this back, we can't use active attempt id in payment_intent to make a refund for a particular attempt.

jarnura (changes requested):

crates/router/src/db/payment_attempt.rs
```diff
@@ -348,7 +275,7 @@ mod storage {
                 enums::MerchantStorageScheme::RedisKv => {
                     let key = format!(
                         "{}_{}",
-                        payment_attempt.merchant_id, payment_attempt.payment_id
+                        payment_attempt.merchant_id, payment_attempt.attempt_id
```
dracarys18, line 278: Shouldn't key be merchant_id_payment_id?

crates/router/src/db/payment_attempt.rs
```diff
@@ -465,33 +390,39 @@ mod storage {
...
-                        ReverseLookupNew {
-                            lookup_id: format!(
-                                "{}_{}",
-                                &updated_attempt.merchant_id, connector_transaction_id
-                            ),
-                            pk_id: key.clone(),
-                            sk_id: field.clone(),
-                            source: "payment_attempt".to_string(),
+                        (None, Some(connector_transaction_id)) => {
+                            add_connector_txn_id_to_reverse_lookup(
```
dracarys18, line 404: Here key is merchant_id_payment_id. But while creating key initially in insert they key is merchant_id_attempt_id.

dracarys18 (changes requested):

vspecky (dismissed):

vspecky (dismissed):

migrations/2023-01-20-113235_add_attempt_id_to_payment_intent/up.sql
```diff
@@ -0,0 +1,11 @@
+-- Your SQL goes here
+ALTER TABLE payment_intent ADD COLUMN attempt_id VARCHAR(64) NOT NULL DEFAULT 'xxx';
```
jarnura, line 2: attempt_id change to active_attempt_id

jarnura (changes requested):

dracarys18 (approved):

jarnura (approved):

SanchithHegde (approved):
Other than the optional change, looks good to me!

PR #442, opened 2023-01-20
fix(checkout): status mapping for payment

dracarys18:
## Type of Change

- [x] Bugfix

## Description

This fixes checkout getting `requires_capture` error even after an automatic capture payment.

## Motivation and Context

Checkout getting `requires_capture` status even in automatic capture because when you make a payment with automatic capture in checkout it will be authorized first and captured few seconds later. So in balances the `amount_to_capture` will not be 0 when we get the response.

## How did you test it?

image

## Checklist

### DIFF
--- a/crates/router/src/connector/checkout.rs
+++ b/crates/router/src/connector/checkout.rs
@@ -279,6 +279,7 @@ impl
         &self,
         res: types::Response,
     ) -> CustomResult<types::ErrorResponse, errors::ConnectorError> {
+        logger::debug!(raw_response=?res);
         let response: checkout::ErrorResponse = res
             .response
             .parse_struct("ErrorResponse")
@@ -368,11 +369,11 @@ impl
         data: &types::PaymentsAuthorizeRouterData,
         res: types::Response,
     ) -> CustomResult<types::PaymentsAuthorizeRouterData, errors::ConnectorError> {
+        logger::debug!(payments_create_response=?res);
         let response: checkout::PaymentsResponse = res
             .response
             .parse_struct("PaymentIntentResponse")
             .change_context(errors::ConnectorError::ResponseDeserializationFailed)?;
-        logger::debug!(payments_create_response=?response);
         types::RouterData::try_from(types::ResponseRouterData {
             response,
             data: data.clone(),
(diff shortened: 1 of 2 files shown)


crates/router/src/connector/checkout/transformers.rs
```diff
@@ -140,6 +140,34 @@ pub enum CheckoutPaymentStatus {
...
 
+impl From<transformers::Foreign<(CheckoutPaymentStatus, Option<enums::CaptureMethod>)>>
+    for transformers::Foreign<enums::AttemptStatus>
+{
+    fn from(
+        item: transformers::Foreign<(CheckoutPaymentStatus, Option<enums::CaptureMethod>)>,
+    ) -> Self {
+        let item = item.0;
+        let status = item.0;
+        let capture_method = item.1;
```
SanchithHegde, line 151:
```suggestion
        let (status, capture_method) = item.0;
```

SanchithHegde (commented):

crates/router/src/connector/checkout.rs
```diff
@@ -470,12 +471,13 @@ impl
         data: &types::PaymentsCancelRouterData,
         res: types::Response,
     ) -> CustomResult<types::PaymentsCancelRouterData, errors::ConnectorError> {
+        logger::debug!(payments_create_response=?res);
```
Abhicodes-crypto, line 474: payments_cancel_response

crates/router/src/connector/checkout/transformers.rs
```diff
@@ -140,6 +140,34 @@ pub enum CheckoutPaymentStatus {
...
+        match status {
+            CheckoutPaymentStatus::Authorized => {
+                if capture_method == Some(enums::CaptureMethod::Automatic)
+                    || capture_method.is_none()
```
Abhicodes-crypto, line 155: This capture method we are getting from request right ?

When doing a sync call, in request capture_method is `None` . And it would mapped to `Charged`, but it should be mapped to `Authorized`. 

Getting the capture method from DB , through router data would avoid that issue.

dracarys18: For sync call we are not using this From implementation, We would be using the one with balances.

Abhicodes-crypto: But if automatic capture payment call has happened and then immediately a sync call. 

Then in the response balances[available_to_capture] will be non zero and the status is returned as Authorized. 
Unless we have a guard saying if payment_intent status is `succeeded, don't call connector sync api`. I don't think we have this guard since some succeeded can become failure cc. @jarnura .

Can you check how it behaves for this case. If it is changing status from succeeded to requires_capture . Then I think we can store and get capture method info from DB(attempt table).

dracarys18: It shows succeeded @Abhicodes-crypto

Abhicodes-crypto (approved):
LGTM

crates/router/src/connector/checkout/transformers.rs
jarnura: After status reaches terminal status we won't sync.

jarnura (approved):
LGTM

PR #445, opened 2023-01-21
fix: move `connector_refund_id` to request

NishantJoshi00:
## Type of Change

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates

## Description

This change will move the `connector_refund_id` into request while also making the updation of router_data for `access_token` step in refund sync optional for connectors that doesn't support it.

This change resolves a bug where the refund sync was failing for every connector which doesn't have access token

### Additional Changes

## How did you test it?
This is an sample of refund with connector that needs access_token (i.e. braintree):
Screenshot 2023-01-21 at 9 19 25 PM

This the sample output of refund sync for connector that doesn't need access_token (i.e. stripe):
Screenshot 2023-01-21 at 9 21 44 PM

Here both the calls provide identical result as expected with refund sync

## Checklist

### DIFF
--- a/crates/router/src/connector/authorizedotnet/transformers.rs
+++ b/crates/router/src/connector/authorizedotnet/transformers.rs
@@ -466,13 +466,18 @@ impl<F> TryFrom<&types::RefundsRouterData<F>> for AuthorizedotnetCreateSyncReque
     type Error = error_stack::Report<errors::ConnectorError>;
 
     fn try_from(item: &types::RefundsRouterData<F>) -> Result<Self, Self::Error> {
-        let transaction_id = item.request.connector_refund_id.clone();
+        let transaction_id = item
+            .request
+            .connector_refund_id
+            .clone()
+            .get_required_value("connector_refund_id")
+            .change_context(errors::ConnectorError::MissingConnectorRefundID)?;
         let merchant_authentication = MerchantAuthentication::try_from(&item.connector_auth_type)?;
 
         let payload = Self {
             get_transaction_details_request: TransactionDetails {
                 merchant_authentication,
-                transaction_id,
+                transaction_id: Some(transaction_id),
             },
         };
         Ok(payload)
--- a/crates/router/src/connector/authorizedotnet/transformers.rs
+++ b/crates/router/src/connector/authorizedotnet/transformers.rs
@@ -466,11 +466,7 @@ impl<F> TryFrom<&types::RefundsRouterData<F>> for AuthorizedotnetCreateSyncReque
     type Error = error_stack::Report<errors::ConnectorError>;
 
     fn try_from(item: &types::RefundsRouterData<F>) -> Result<Self, Self::Error> {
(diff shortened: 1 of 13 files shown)


crates/router/src/core/refunds.rs
```diff
@@ -295,9 +295,11 @@ pub async fn sync_refund_with_gateway(
...
     logger::debug!(refund_retrieve_router_data=?router_data);
 
-    match add_access_token_result.access_token_result {
-        Ok(access_token) => router_data.access_token = access_token,
-        Err(connector_error) => router_data.response = Err(connector_error),
+    if add_access_token_result.connector_supports_access_token {
+        match add_access_token_result.access_token_result {
+            Ok(access_token) => router_data.access_token = access_token,
+            Err(connector_error) => router_data.response = Err(connector_error),
+        }
```
Narayanbhat166, line 302: Is this extra check necessary, now that you have made the changes to not read from response in routerdata.

Narayanbhat166 (dismissed):

crates/router/src/core/refunds.rs
NishantJoshi00: Currently, @jagan-jaya is making a change for `payu`, `cybersource` and `globalpay` which still uses the response for fetching the connector_refund_id, we discussed that once this PR get's merged we can raise another PR to remove those changes too. Till then, to keep the refund_retrieve_core from breaking. This change needs to be kept.

CC: @jagan-jaya

crates/router/src/connector/authorizedotnet/transformers.rs
```diff
@@ -466,11 +466,7 @@ impl<F> TryFrom<&types::RefundsRouterData<F>> for AuthorizedotnetCreateSyncReque
     type Error = error_stack::Report<errors::ConnectorError>;
 
     fn try_from(item: &types::RefundsRouterData<F>) -> Result<Self, Self::Error> {
-        let transaction_id = item
-            .response
-            .as_ref()
-            .map(|refund_response_data| refund_response_data.connector_refund_id.clone())
-            .ok();
+        let transaction_id = item.request.connector_refund_id.clone();
```
jagan-jaya, line 469: shouldn't we handle required field validation here?

jagan-jaya (changes requested):

crates/router/src/connector/authorizedotnet/transformers.rs
NishantJoshi00: According to the earlier code, we are handling the existence of the ID somewhere ahead in the code. 
I have done an inplace change for the connector refund id, to avoid any unwanted bugs to arise because of this change.

jarnura (approved):

Narayanbhat166 (approved):

jagan-jaya (approved):

jarnura (approved):
LGTM

PR #461, opened 2023-01-23
refactor(core): validate payment status using common function

Narayanbhat166:
## Type of Change

- [x] Refactor

## Description

Payments session did not have a check for creating the session token, which would allow users to create session token for any `succeeded` or `failed` payment.

## Motivation and Context

While adding this check to `payment_session` operation, I found out that the same logic is being used in other operations too. There is a function `helpers::validate_payment_status` which will validate the payment intent status based on the `statuses` which the intent is not allowed to have. It will throw an error if the condition fails.

## How did you test it?

- Update a payment which has status of `requires_capture`

image
- Confirm a payment which has already been `confirmed` and `succeeded`

Screenshot 2023-01-25 at 3 16 07 PM

## Checklist

### DIFF
--- a/crates/router/src/core/payments/helpers.rs
+++ b/crates/router/src/core/payments/helpers.rs
@@ -1,6 +1,6 @@
 use std::borrow::Cow;
 
-use common_utils::ext_traits::AsyncExt;
+use common_utils::{ext_traits::AsyncExt, fp_utils};
 // TODO : Evaluate all the helper functions ()
 use error_stack::{report, IntoReport, ResultExt};
 use masking::ExposeOptionInterface;
@@ -1150,18 +1150,18 @@ pub(crate) fn validate_payment_status(
     not_allowed_statuses: &[storage_enums::IntentStatus],
     action: &'static str,
 ) -> Result<(), errors::ApiErrorResponse> {
-    if not_allowed_statuses
-        .iter()
-        .any(|not_allowed| intent_status == not_allowed)
-    {
-        Err(errors::ApiErrorResponse::PreconditionFailed {
-            message: format!(
-                "You cannot {action} this Payment because it has already {intent_status}",
-            ),
-        })
-    } else {
-        Ok(())
-    }
+    fp_utils::when(
+        not_allowed_statuses
+            .iter()
+            .any(|not_allowed| intent_status == not_allowed),
+        || {
+            Err(errors::ApiErrorResponse::PreconditionFailed {
+                message: format!(
+                    "You cannot {action} this Payment because it has already {intent_status}",
+                ),
+            })
+        },
+    )
 }
 
 pub(crate) fn validate_pm_or_token_given(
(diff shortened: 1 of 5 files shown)


crates/router/src/core/payments/helpers.rs
```diff
@@ -1145,6 +1145,25 @@ pub(crate) fn authenticate_client_secret(
     }
 }
 
+pub(crate) fn validate_payment_status(
+    intent_status: &storage_enums::IntentStatus,
+    not_allowed_statuses: &[storage_enums::IntentStatus],
+    action: &'static str,
+) -> Result<(), errors::ApiErrorResponse> {
+    if not_allowed_statuses
```
jarnura, line 1153: use when instead

crates/router/src/core/payments/operations/payment_confirm.rs
```diff
@@ -51,6 +51,22 @@ impl<F: Send + Clone> GetTracker<F, PaymentData<F>, api::PaymentsRequest> for Pa
...
             .change_context(errors::ApiErrorResponse::PaymentNotFound)?;
 
+        payment_intent = db
+            .find_payment_intent_by_payment_id_merchant_id(&payment_id, merchant_id, storage_scheme)
+            .await
+            .map_err(|error| {
+                error.to_not_found_response(errors::ApiErrorResponse::PaymentNotFound)
+            })?;
+
+        helpers::validate_payment_status(
```
jarnura, line 61: the function name not express the logic inside the function. Prefer a better name

jarnura (changes requested):

crates/router/src/core/payments/helpers.rs
```diff
@@ -1150,18 +1150,18 @@ pub(crate) fn validate_payment_status(
...
-                "You cannot {action} this Payment because it has already {intent_status}",
-            ),
-        })
-    } else {
-        Ok(())
-    }
+    fp_utils::when(
+        not_allowed_statuses
+            .iter()
+            .any(|not_allowed| intent_status == not_allowed),
```
SanchithHegde, line 1156:
```suggestion
        not_allowed_statuses.contains(intent_status),
```

crates/router/src/core/payments/helpers.rs
```diff
@@ -1157,7 +1157,7 @@ pub(crate) fn validate_payment_status(
         || {
             Err(errors::ApiErrorResponse::PreconditionFailed {
                 message: format!(
-                    "You cannot {action} this Payment because it has already {intent_status}",
+                    "You cannot {action} this payment because it has already {intent_status}",
```
SanchithHegde, line 1160: It might be better to use something like:

```"You cannot {action} this payment because it has status `{intent_status}`"```

I feel `[...] it has already confirmed` or `[...] it has already voided` (just taking a few examples) would make the sentence grammatically incorrect.

I'll leave this up to you, go with whatever you see fit.

SanchithHegde (changes requested):

crates/router/src/core/payments/helpers.rs
Narayanbhat166: adding a been would make it gramatically correct, `[...] it has already been confirmed`

SanchithHegde (approved):

jarnura (approved):

PR #464, opened 2023-01-25
fix(connector): fix globalpay error handling

jagan-jaya:
## Type of Change

- [X] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates

## Description

Globalpay throw error message for failure transaction if it has the message.

### Additional Changes

## Motivation and Context

Previously for failure transactions only the `failed` status is shown, now adding the reason for failure also to the response

## How did you test it?

Ran unit tests in local

## Checklist

### DIFF
--- a/crates/router/src/connector/globalpay/response.rs
+++ b/crates/router/src/connector/globalpay/response.rs
@@ -344,7 +344,7 @@ pub enum FingerprintPresenceIndicator {
 }
 
 /// Indicates where a transaction is in its lifecycle.
-#[derive(Clone, Debug, Serialize, Deserialize)]
+#[derive(Clone, Copy, Debug, Serialize, Deserialize)]
 #[serde(rename_all = "SCREAMING_SNAKE_CASE")]
 pub enum GlobalpayPaymentStatus {
     /// A Transaction has been successfully authorized and captured. The funding
--- a/crates/router/src/connector/globalpay/response.rs
+++ b/crates/router/src/connector/globalpay/response.rs
@@ -344,7 +344,7 @@ pub enum FingerprintPresenceIndicator {
 }
 
 /// Indicates where a transaction is in its lifecycle.
-#[derive(Debug, Serialize, Deserialize)]
+#[derive(Clone, Debug, Serialize, Deserialize)]
 #[serde(rename_all = "SCREAMING_SNAKE_CASE")]
 pub enum GlobalpayPaymentStatus {
     /// A Transaction has been successfully authorized and captured. The funding
--- a/crates/router/src/connector/globalpay/transformers.rs
+++ b/crates/router/src/connector/globalpay/transformers.rs
@@ -185,7 +185,7 @@ impl<F, T>
             types::PaymentsResponseData,
         >,
     ) -> Result<Self, Self::Error> {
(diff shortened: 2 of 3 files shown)


crates/router/src/connector/globalpay/transformers.rs
```diff
@@ -149,6 +149,28 @@ impl From<GlobalpayPaymentStatus> for enums::RefundStatus {
...
+fn get_payment_response(
+    status: enums::AttemptStatus,
+    response: GlobalpayPaymentsResponse,
+) -> Result<types::PaymentsResponseData, ErrorResponse> {
+    match status {
+        enums::AttemptStatus::Failure => Err(ErrorResponse {
+            message: response
+                .payment_method
+                .and_then(|pm| pm.message)
+                .map_or("".to_string(), |f| f),
```
jarnura, line 161: There are constant for no message present, avoid using empty strings. Instead map_or, or will work. So that we don't want to use closure just to return the same input.

jarnura (changes requested):

crates/router/src/connector/globalpay/transformers.rs
```diff
@@ -149,6 +149,28 @@ impl From<GlobalpayPaymentStatus> for enums::RefundStatus {
...
+fn get_payment_response(
+    status: enums::AttemptStatus,
+    response: GlobalpayPaymentsResponse,
+) -> Result<types::PaymentsResponseData, ErrorResponse> {
+    match status {
+        enums::AttemptStatus::Failure => Err(ErrorResponse {
+            message: response
+                .payment_method
+                .and_then(|pm| pm.message)
+                .unwrap_or_default(),
```
jarnura, line 161: The default return empty sting?

Use this const `NO_ERROR_MESSAGE` as Default

jarnura (changes requested):

crates/router/src/connector/globalpay/response.rs
```diff
@@ -344,7 +344,7 @@ pub enum FingerprintPresenceIndicator {
 }
 
 /// Indicates where a transaction is in its lifecycle.
-#[derive(Debug, Serialize, Deserialize)]
+#[derive(Clone, Debug, Serialize, Deserialize)]
```
SanchithHegde, line 347: You can even derive `Copy` for enums with no fields (all variants are unit variants), as copying them is cheap. However, if your enum has variants with fields which cannot be copied (say `String` fields), you'll have to switch over to cloning them.

Since most "status" enums have only unit variants, you can derive `Copy` on them without much thought.

crates/router/src/connector/globalpay/transformers.rs
```diff
@@ -162,15 +185,10 @@ impl<F, T>
             types::PaymentsResponseData,
         >,
     ) -> Result<Self, Self::Error> {
+        let status = enums::AttemptStatus::from(item.response.status.clone());
```
SanchithHegde, line 188: Then you can get rid of this clone and elsewhere in the module, if you've done any.

SanchithHegde (commented):

jarnura (approved):

SanchithHegde (approved):

PR #472, opened 2023-01-27
refactor(errors): error enums to use 'static str

Narayanbhat166:
## Type of Change

- [x] Enhancement

## Description

Changes included in this PR  
- 'static str to be passed to error enums.
- `DuplicateValueFound` converted to `struct enum` -> provides better visibility as to what is to be passed.

## Motivation and Context

Few enum variants of error like `MissingRequiredField` used to take `owned` values, the problem with this is everytime you use a `static str` ( which is in all the cases ), there needs to be a conversion to owned `String`. This is not necessary since it will lead to extra heap allocations and conversion from `str` to `String` using the `tostring()` method everytime.

## How did you test it?

If it compiles, it works

## Checklist


crates/router/src/core/errors.rs
```diff
@@ -54,8 +54,8 @@ pub enum StorageError {
     DatabaseError(error_stack::Report<storage_errors::DatabaseError>),
     #[error("ValueNotFound: {0}")]
     ValueNotFound(String),
-    #[error("DuplicateValue: {0}")]
-    DuplicateValue(String),
+    #[error("DuplicateValue: {entity} already exists for key {key}")]
+    DuplicateValue { entity: &'static str, key: String },
```
SanchithHegde, line 58: Should we be exposing the `key` that we internally use to store the entity? I personally feel leaking an implementation detail is not a good idea.

(I'm specifically speaking with respect to ephemeral key.)

dracarys18: Yes exactly my thoughts since we store api-key as a key for a fetching it later in ephemeral key

Narayanbhat166: The database errors are not propogated to user, it will be shown only in the logs, the duplicate error in case of ephemeral key is raised only in case of creating the ephemeral key

jarnura (approved):

SanchithHegde (approved):

PR #481, opened 2023-01-30
feat(config): add fetch endpoint for config table

dracarys18:
## Type of Change

- [x] New feature

## Description

Added fetch endpoint for config table.

## Motivation and Context

This PR adds cached fetch to config table, for dynamically configuring application.

## How did you test it?

image

## Checklist

### DIFF
--- a/crates/router/src/core/configs.rs
+++ b/crates/router/src/core/configs.rs
@@ -2,17 +2,16 @@ use crate::{
     core::errors::{self, utils::StorageErrorExt, RouterResponse},
     db::StorageInterface,
     services::ApplicationResponse,
-    types::storage,
+    types::api,
 };
 
-pub async fn read_config(
-    store: &dyn StorageInterface,
-    key: &str,
-) -> RouterResponse<storage::Config> {
-    Ok(ApplicationResponse::Json(
-        store
-            .find_config_by_key_cached(key)
-            .await
-            .map_err(|err| err.to_not_found_response(errors::ApiErrorResponse::ConfigNotFound))?,
-    ))
+pub async fn read_config(store: &dyn StorageInterface, key: &str) -> RouterResponse<api::Config> {
+    let config = store
+        .find_config_by_key_cached(key)
+        .await
+        .map_err(|err| err.to_not_found_response(errors::ApiErrorResponse::ConfigNotFound))?;
+    Ok(ApplicationResponse::Json(api::Config {
+        key: config.key,
+        value: config.config,
+    }))
 }
--- a/crates/router/src/core/configs.rs
+++ b/crates/router/src/core/configs.rs
@@ -1,0 +1,18 @@
+use crate::{
+    core::errors::{self, utils::StorageErrorExt, RouterResponse},
+    db::StorageInterface,
+    services::ApplicationResponse,
+    types::storage,
+};
+
+pub async fn read_config(
+    store: &dyn StorageInterface,
+    key: &str,
+) -> RouterResponse<storage::Config> {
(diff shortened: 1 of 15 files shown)


crates/router/src/types/api/configs.rs
```diff
@@ -0,0 +1,5 @@
+#[derive(Clone, serde::Deserialize)]
+pub struct ConfigKeyCreate {
+    pub key: String,
+    pub value: String,
+}
```
SanchithHegde, line 5: Would be better if you could rename this as `Config` (or something similar), excluding the `Create`/`Retrieve` operation name. From the looks of it, both creating and retrieving a config involves only these 2 fields, so calling it `Config` would be preferable.

crates/router/src/core/configs.rs
```diff
@@ -0,0 +1,18 @@
...
+    core::errors::{self, utils::StorageErrorExt, RouterResponse},
+    db::StorageInterface,
+    services::ApplicationResponse,
+    types::storage,
+};
+
+pub async fn read_config(
+    store: &dyn StorageInterface,
+    key: &str,
+) -> RouterResponse<storage::Config> {
```
SanchithHegde, line 11: Please ensure that the type returned by the API is an API model and not a storage model.

crates/router/src/db/configs.rs
```diff
@@ -60,6 +65,30 @@ impl ConfigInterface for Store {
...
+                        .await
+                        .change_context(errors::StorageError::KVError)?;
+                    config
+                }
+                err => Err(report!(errors::StorageError::KVError)
+                    .attach_printable(format!("Error while fetching config {err}")))?,
+            },
+            Ok(val) => val,
+        })
+    }
```
SanchithHegde, line 91: Can we please have this "populate from database if not found in Redis" logic in a single place instead of implementing it for each entity independently?

SanchithHegde (changes requested):

crates/router/src/core/configs.rs
```diff
@@ -0,0 +1,17 @@
...
+
+pub async fn read_config(store: &dyn StorageInterface, key: &str) -> RouterResponse<api::Config> {
+    let config = store
+        .find_config_by_key_cached(key)
+        .await
+        .map_err(|err| err.to_not_found_response(errors::ApiErrorResponse::ConfigNotFound))?;
+    Ok(ApplicationResponse::Json(api::Config {
+        key: config.key,
+        value: config.config,
+    }))
```
SanchithHegde, line 16: Could you implement `From` or `ForeignFrom` instead?

crates/router/src/db/configs.rs
```diff
@@ -60,6 +65,30 @@ impl ConfigInterface for Store {
...
+                errors::RedisError::NotFound => {
+                    let config = self.find_config_by_key(key).await?;
+                    redis
+                        .serialize_and_set_key(&config.key, &config)
+                        .await
+                        .change_context(errors::StorageError::KVError)?;
+                    config
+                }
+                err => Err(report!(errors::StorageError::KVError)
+                    .attach_printable(format!("Error while fetching config {err}")))?,
```
SanchithHegde, line 87:
```suggestion
                error => err.change_context(errors::StorageError::KVError)
                    .attach_printable("Error while fetching config")?,
```

SanchithHegde (changes requested):

SanchithHegde (approved):

jarnura (approved):

SanchithHegde (approved):

PR #482, opened 2023-01-31
perf(router): remove redundant DB call during merchant account update

Abhicodes-crypto:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring
- [ ] Dependency updates

## Description

Merchant account is fetched before update in all cases. This is redundant and can be avoided by fetching only when required. This PR addresses that

### Additional Changes

## How did you test it?

Manual 
image

## Checklist

### DIFF
--- a/crates/router/src/db/merchant_account.rs
+++ b/crates/router/src/db/merchant_account.rs
@@ -26,7 +26,7 @@ pub trait MerchantAccountInterface {
         merchant_account: storage::MerchantAccountUpdate,
     ) -> CustomResult<storage::MerchantAccount, errors::StorageError>;
 
-    async fn normal_update_merchant(
+    async fn update_specific_fields_in_merchant(
         &self,
         merchant_id: &str,
         merchant_account: storage::MerchantAccountUpdate,
@@ -85,13 +85,13 @@ impl MerchantAccountInterface for Store {
             .into_report()
     }
 
-    async fn normal_update_merchant(
+    async fn update_specific_fields_in_merchant(
         &self,
         merchant_id: &str,
         merchant_account: storage::MerchantAccountUpdate,
     ) -> CustomResult<storage::MerchantAccount, errors::StorageError> {
         let conn = pg_connection(&self.master_pool).await;
-        storage::MerchantAccount::normal_update(&conn, merchant_id, merchant_account)
+        storage::MerchantAccount::update_with_specific_fields(&conn, merchant_id, merchant_account)
             .await
             .map_err(Into::into)
             .into_report()
@@ -193,7 +193,7 @@ impl MerchantAccountInterface for MockDb {
         Err(errors::StorageError::MockDbError)?
     }
 
-    async fn normal_update_merchant(
+    async fn update_specific_fields_in_merchant(
         &self,
         _merchant_id: &str,
         _merchant_account: storage::MerchantAccountUpdate,
--- a/crates/router/src/db/merchant_account.rs
+++ b/crates/router/src/db/merchant_account.rs
@@ -26,6 +26,12 @@ pub trait MerchantAccountInterface {
         merchant_account: storage::MerchantAccountUpdate,
     ) -> CustomResult<storage::MerchantAccount, errors::StorageError>;
 
+    async fn normal_update_merchant(
+        &self,
+        merchant_id: &str,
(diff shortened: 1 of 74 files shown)


crates/storage_models/src/merchant_account.rs
```diff
@@ -48,7 +48,7 @@ pub struct MerchantAccountNew {
 #[derive(Debug)]
 pub enum MerchantAccountUpdate {
     Update {
-        merchant_id: String,
+        merchant_id: Option<String>,
```
Abhicodes-crypto, line 51: Should we allow the update of api_key, merchant_id, publishable key ?

crates/router/src/db/merchant_account.rs
```diff
@@ -26,6 +26,12 @@ pub trait MerchantAccountInterface {
         merchant_account: storage::MerchantAccountUpdate,
     ) -> CustomResult<storage::MerchantAccount, errors::StorageError>;
 
+    async fn normal_update_merchant(
```
Abhicodes-crypto, line 29: Need function name suggestion.

crates/storage_models/src/query/generics.rs
```diff
@@ -111,6 +111,66 @@ where
         .attach_printable_lazy(|| format!("Error while updating {debug_values}"))
 }
 
+#[instrument(level = "DEBUG", skip_all)]
+pub async fn generic_update_with_unique_predicate_get_result<T, V, P, R>(
```
Abhicodes-crypto, line 115: merchant_id is a unique index, but if it was PK. Then this generic function can be avoided.

crates/router/src/db/merchant_account.rs
jarnura: update_specific_fields_in_merchant

crates/storage_models/src/merchant_account.rs
jarnura: merchant_id can't be updated.

api_key and publishable_key can be updated

crates/storage_models/src/query/generics.rs
jarnura: making unique identifier of a table to primary key we will take it up in a separate PR. After this we will deprecate this

crates/storage_models/src/query/merchant_account.rs
```diff
@@ -40,6 +40,24 @@ impl MerchantAccount {
         }
     }
 
+    pub async fn normal_update(
```
jarnura, line 43: change function name, update_with_specific_fields

jarnura (changes requested):

crates/storage_models/src/query/generics.rs
```diff
@@ -111,6 +111,66 @@ where
...
+        <<T as FilterDsl<P>>::Output as HasTable>::Table,
+        <<T as FilterDsl<P>>::Output as IntoUpdateTarget>::WhereClause,
+        <V as AsChangeset>::Changeset,
+    >: AsQuery + LoadQuery<'static, PgConnection, R> + QueryFragment<Pg> + Send,
+    R: Send + 'static,
+{
+    let debug_values = format!("{values:?}");
+
+    let query = diesel::update(<T as HasTable>::table().filter(predicate)).set(values);
+    logger::debug!(query = %debug_query::<Pg, _>(&query).to_string());
```
SanchithHegde, line 134: The panic would happen right here if there are no fields to update and not anywhere in the following code.

Can we raise a `NoFieldsToUpdate` error variant if we encounter a `ConnectionError::Query(DieselError::QueryBuilderError(_))`?

I'd prefer it if you used `get_result_async()` instead of `get_results_async()`.

Abhicodes-crypto: > The panic would happen right here if there are no fields to update and not anywhere in the following code.

Yes that would happen. I've mentioned that on line 154. Will handle it. I guess  It would also happen in `generic_update` and `generic_update_with_results` , will check and handle there also.

> I'd prefer it if you used get_result_async() instead of get_results_async().

I doubt if it would throw error or update multiple records and return the first record. Will verify and use accordingly.

SanchithHegde (approved):

PR #483, opened 2023-01-31
feat(cache): add generic cache functions to populate and invalidate cache

dracarys18:
## Type of Change

- [x] Enhancement

## Description

This adds generic helper functions which helps to populate cache and update cache.

## Motivation and Context

So this PR adds two generic functions to help you populate value from redis if it doesn't exist in redis it will get it from DB and insert the value in redis. Usage:

```rust
inhabit_cache(&store,&key, || async{ function_that_inserts_into_db().await}).await
```
And for redact it will update it directly in the DB and remove it from cache

```rust
redact_cache(&store,&key,|| async{function_that_updates_into_db().await}).await
```

## How did you test it?

Tested with #481 PR

## Checklist


crates/router/src/db/cache.rs
```diff
@@ -0,0 +1,56 @@
+#![allow(dead_code)]
```
Narayanbhat166, line 1: remove this

crates/router/src/db/cache.rs
```diff
@@ -0,0 +1,56 @@
...
+
+use error_stack::{report, ResultExt};
+
+use super::Store;
+use crate::core::errors::{self, CustomResult};
+
+pub async fn inhabit_cache<T, F, Fut>(
```
Narayanbhat166, line 8: Provide meaningful function name or add comments

dracarys18:
Closing this as this can't be generic because the logic would change for different use cases.

crates/router/src/db/cache.rs
```diff
@@ -0,0 +1,57 @@
...
+where
+    T: serde::Serialize + serde::de::DeserializeOwned + std::fmt::Debug,
+    F: FnOnce() -> Fut + Send,
+    Fut: futures::Future<Output = CustomResult<T, errors::StorageError>> + Send,
+{
+    let redis = &store.redis_conn;
+    let redis_val = redis
+        .get_and_deserialize_key::<T>(key, std::any::type_name::<T>())
+        .await;
+    Ok(match redis_val {
```
jarnura, line 22: in OK we can wrap with Ok and we don't want to use ? in line: 34

jarnura (changes requested):

SanchithHegde (approved):

PR #494, opened 2023-02-02
fix: don't panic in redis library on creating connection pool

dracarys18:
## Type of Change

- [x] Enhancement

## Description

Don't panic while creating connection pool in redis library.

## Motivation and Context

Removing panics in library code

## Checklist


crates/redis_interface/src/lib.rs
```diff
@@ -21,24 +21,20 @@ pub mod commands;
...
 
 impl RedisConnectionPool {
     /// Create a new Redis connection
-    ///
-    /// # Panics
-    ///
-    /// Panics if a connection to Redis is not successful.
-    #[allow(clippy::expect_used)]
-    pub async fn new(conf: &RedisSettings) -> Self {
+    pub async fn new(conf: &RedisSettings) -> Result<Self, RedisError> {
```
SanchithHegde, line 37: Do you think we should raise our error type instead, an `errors::RedisError::RedisConnectionError` variant maybe?

SanchithHegde (commented):

crates/redis_interface/src/lib.rs
dracarys18: i guess we could something like 
```rust
RedisConnectionError {err: fred::RedisError}
```

SanchithHegde: Either that, or return an `error_stack::Result` / `CustomResult`.

SanchithHegde (approved):

ashokkjag (approved):

PR #496, opened 2023-02-03
refactor: use generic function to push to drainer

### LINKED ISSUE #97: Make KV storage flow more generic
Model the KV storage flow as a series of common operations (For example, `set in redis -> get sql query -> push query to redis stream`) and define generic utilities that perform these operations for all KV-enabled storage interfaces (as opposed to rewriting code).

dracarys18:
## Type of Change

- [x] Refactoring

## Description

Use a generic function to push sql queries into drainer stream.

## Motivation and Context

Code reusability for pushing to drainer stream. This Closes #97.

## How did you test it?

Manual

## Checklist


crates/router/src/utils/db_utils.rs
```diff
@@ -8,3 +8,29 @@ pub fn generate_hscan_pattern_for_refund(sk: &str) -> String {
...
+
+#[cfg(feature = "kv_store")]
+pub(crate) async fn push_to_drainer_stream<T>(
+    store: &crate::services::Store,
+    redis_entry: storage_models::kv::TypedSql,
+    partition_key: crate::utils::storage_partitioning::PartitionKey<'_>,
+) -> crate::core::errors::CustomResult<(), crate::core::errors::StorageError>
+where
+    T: crate::utils::storage_partitioning::KvStorePartition,
+{
```
SanchithHegde, line 20: Could we have this as an implementation on `Store`?

SanchithHegde:
@dracarys18 Would merging this PR address #97, or are there more instances in the code where things could be more generic?

dracarys18:
> @dracarys18 Would merging this PR address #97, or are there more instances in the code where things could be more generic?

As far as I understand this PR should address that issue. Thoughts? @vspecky

crates/router/src/utils/db_utils.rs
dracarys18: Makes sense will do

SanchithHegde (approved):

crates/router/src/services.rs
```diff
@@ -45,4 +45,29 @@ impl Store {
...
+        &self,
+        redis_entry: storage_models::kv::TypedSql,
+        partition_key: crate::utils::storage_partitioning::PartitionKey<'_>,
+    ) -> crate::core::errors::CustomResult<(), crate::core::errors::StorageError>
+    where
+        T: crate::utils::storage_partitioning::KvStorePartition,
+    {
+        use error_stack::ResultExt;
+
+        let shard_key = T::shard_key(partition_key, self.config.drainer_num_partitions);
```
jarnura, line 60: the second argument `self.config.drainer_num_partitions` can be moved into `shard_key` function itself?.

jarnura (changes requested):

vspecky:
> > @dracarys18 Would merging this PR address #97, or are there more instances in the code where things could be more generic?
> 
> As far as I understand this PR should address that issue. Thoughts? @vspecky

Skimming through the code, the PR seems to address the example I gave in #97, however that is not the whole scope of the issue. For instance, reads can also be generalized.
```
fetch from redis -> if not found, read from postgres and set in redis
```

@dracarys18 This I think you can take up in a separate PR since it seems to be out of the scope of this one.

vspecky (approved):
LGTM, please address generalizing redis reads in a separate PR.

dracarys18:
> > > @dracarys18 Would merging this PR address #97, or are there more instances in the code where things could be more generic?
> > 
> > 
> > As far as I understand this PR should address that issue. Thoughts? @vspecky
> 
> Skimming through the code, the PR seems to address the example I gave in #97, however that is not the whole scope of the issue. For instance, reads can also be generalized.
> 
> ```
> fetch from redis -> if not found, read from postgres and set in redis
> ```
> 
> @dracarys18 This I think you can take up in a separate PR since it seems to be out of the scope of this one.

I think this has been done in #483. Is that what you are specifying?

vspecky:
> > > > @dracarys18 Would merging this PR address #97, or are there more instances in the code where things could be more generic?
> > > 
> > > 
> > > As far as I understand this PR should address that issue. Thoughts? @vspecky
> > 
> > 
> > Skimming through the code, the PR seems to address the example I gave in #97, however that is not the whole scope of the issue. For instance, reads can also be generalized.
> > ```
> > fetch from redis -> if not found, read from postgres and set in redis
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > @dracarys18 This I think you can take up in a separate PR since it seems to be out of the scope of this one.
> 
> I think this has been done in #483. Is that what you are specifying?

@dracarys18 Yes, that seems correct. Just needs implementation for all KV tables 👍

jarnura (approved):

PR #497, opened 2023-02-03
feat: add metrics to drainer

NishantJoshi00:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring
- [ ] Dependency updates

## Description

This change adds metrics to drainer.
The added metrics are listed in the `drainer/src/metrics.rs` file

### Additional Changes

## Checklist


jarnura (approved):

crates/drainer/src/lib.rs
```diff
@@ -55,9 +58,17 @@ async fn drainer(
 
     // parse_stream_entries returns error if no entries is found, handle it
     let (entries, last_entry_id) = utils::parse_stream_entries(&stream_read, stream_name)?;
-
     let read_count = entries.len();
 
+    metrics::JOBS_PICKED_PER_STREAM.add(
+        &metrics::CONTEXT,
+        u64::try_from(read_count).unwrap_or(u64::MAX),
```
lsampras, line 65: Preferably use `u64::MIN` here so that we can ignore errors

crates/common_utils/src/lib.rs
```diff
@@ -33,6 +33,15 @@ pub mod date_time {
...
         OffsetDateTime::now_utc().unix_timestamp()
     }
+
+    /// Calculate execution time for a async block in milliseconds
+    pub async fn time_it<T, Fut: futures::Future<Output = T>, F: FnOnce() -> Fut>(
+        block: F,
+    ) -> (T, f64) {
+        let start = Instant::now();
+        let result = block().await;
+        (result, start.elapsed().as_seconds_f64() * 1000f64)
```
SanchithHegde, line 43: If you need milliseconds as an integer, you could also use `as_millis()`.

`(result, start.elapsed().as_millis())`

crates/drainer/src/lib.rs
```diff
@@ -115,7 +165,11 @@ async fn drainer(
         utils::trim_from_stream(stream_name, last_entry_id.as_str(), &store.redis_conn).await?;
 
     if read_count != entries_trimmed {
-        // TODO: log
+        logger::error!(
+            read_entries = %read_count,
+            trimmed_entries = %entries_trimmed,
+            ?entries
+        );
```
SanchithHegde, line 172: It would be better if we could have an error message with some context here.

crates/drainer/src/metrics.rs
```diff
@@ -0,0 +1,66 @@
...
+
+pub(crate) static CONTEXT: Lazy<Context> = Lazy::new(Context::current);
+static DRAINER_METER: Lazy<Meter> = Lazy::new(|| global::meter("DRAINER"));
+
+// macro_rules! create_counter {
+//     ($name:ident, $meter:ident,) => {
+//         pub(crate) static $name: Lazy<Counter<u64>> =
+//             Lazy::new(|| $meter.u64_counter(stringify!($name)).init());
+//     };
+// }
```
SanchithHegde, line 17: Remove this snippet of commented code if this is not required.

crates/drainer/src/utils.rs
```diff
@@ -99,11 +122,15 @@ pub fn parse_stream_entries<'a>(
         .into_report()
 }
 
-pub fn increment_stream_index(index: u8, total_streams: u8) -> u8 {
+pub fn increment_stream_index(index: u8, jobs_picked: u8, total_streams: u8) -> (u8, u8) {
```
SanchithHegde, line 125: It would be better if you could either use a struct (since it is not clear what the 2 `u8`s represent), or document what each element in the tuple returned represents.

SanchithHegde (changes requested):

crates/common_utils/src/lib.rs
NishantJoshi00: Here we want the time in f64 milliseconds, as_millies returns u128 which will not preserve the precision. That's the reason I used seconds as f64

SanchithHegde (approved):

crates/drainer/src/lib.rs
```diff
@@ -55,9 +58,17 @@ async fn drainer(
...
     let (entries, last_entry_id) = utils::parse_stream_entries(&stream_read, stream_name)?;
-
     let read_count = entries.len();
 
+    metrics::JOBS_PICKED_PER_STREAM.add(
+        &metrics::CONTEXT,
+        u64::try_from(read_count).unwrap_or(u64::MIN),
+        &[metrics::KeyValue {
+            key: "stream".into(),
+            value: stream_name.to_string().into(),
```
jarnura, line 68: to_string and into both required here?

jarnura (changes requested):

crates/drainer/src/lib.rs
NishantJoshi00: Yes, as the `stream_name` is &str, we need to make it owned before converting into `Value`

jarnura (approved):

Issue #501, opened 2023-02-05
[FEATURE] Allow rotating and revoking API keys


SanchithHegde:
### Feature Description

As of now, the `merchant_account` table has an `api_key` field which restricts a merchant account to only have one API key. Furthermore, there are no mechanisms in place to revoke and regenerate API keys. To improve the convenience of rotating and revoking API keys, it'd be better to allow multiple API keys to be associated with a merchant account. Furthermore, API keys are stored in plaintext as of now. It'd be great if they could be hashed as well, when this request is addressed.

### Possible Implementation

A possible approach to address this is to add an `api_keys` table with `name`, `merchant_id`, `hashed_key` fields (including any additional fields as necessary). And for the API interface, this could include endpoints to create, retrieve, update and revoke an API key, and an endpoint to list all API keys associated with the current merchant account.

### Have you spent some time to check if this feature request has been raised before?

- [X] I checked and didn't find similar issue

### Have you read the Contributing Guidelines?

- [X] I have read the [Contributing Guidelines](https://github.com/juspay/hyperswitch/blob/main/docs/CONTRIBUTING.md)

### Are you willing to submit a PR?

Yes I am willing to submit a PR!

PR #503, opened 2023-02-05
fix(storage_models): fix inconsistent field order for merchant account types

SanchithHegde:
## Type of Change

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates

## Description

This PR fixes inconsistent field order for merchant account types as compared to the generated `schema.rs` file when the `diesel migration run` command is run on a fresh clone of the repository.

## Motivation and Context

Even though the previous code was not buggy, running database migrations would still cause incorrect data to be populated in the `routing_algorithm` and `metadata` fields for any user who freshly clones the repository and does not revert the changes to the `schema.rs` file. 

Me and @Abhicodes-crypto had a discussion about this a while ago, but we never took active steps to fix it. I noticed this again when working on #501.

## How did you test it?

N/A

## Checklist


ashokkjag (approved):
Should we leave some comment to inform users to keep the order based on the generated code?

jarnura (approved):

SanchithHegde:
> Should we leave some comment to inform users to keep the order based on the generated code?

@ashokkjag Where do you think would be appropriate to add that note? As of now, we don't have any document containing instructions to add new database migrations.

PR #509, opened 2023-02-06
feat: add timeout for `set` command on hashes and add function to allow retry in database

NishantJoshi00:
- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring
- [ ] Dependency updates

## Description

This PR includes 2 changes
- It adds timeout to function that does set data in hashes in Redis. marked with a configuration entry `default_hash_ttl`
- It adds functionality to lookup the data into database if not found in redis

### Additional Changes

```toml

[redis]
default_hash_ttl = 900
```

## Checklist


config/config.example.toml
```diff
@@ -43,6 +43,7 @@ pool_size = 5              # Number of connections to keep open
 reconnect_max_attempts = 5 # Maximum number of reconnection attempts to make before failing. Set to 0 to retry forever.
 reconnect_delay = 5        # Delay between reconnection attempts, in milliseconds
 default_ttl = 300          # Default TTL for entries, in seconds
+default_hash_ttl = 900     # Default TTL for hashes entries, in seconds
```
SanchithHegde, line 46: Should we also suggest them to keep this greater than the drainer's loop time (which doesn't seem to be present in config now) so that the value would be in database by the time it is cleared from cache?

crates/router/src/db/payment_intent.rs
```diff
@@ -199,20 +199,17 @@ mod storage {
...
+                            .get_hash_field_and_deserialize(&key, "pi", "PaymentIntent"),
+                        self.find_payment_intent_by_payment_id_merchant_id(
+                            payment_id,
+                            merchant_id,
+                            enums::MerchantStorageScheme::PostgresOnly,
+                        ),
+                    )
+                    .await
+
                     // Check for database presence as well Maybe use a read replica here ?
```
SanchithHegde, line 213: I believe this comment can be removed?

SanchithHegde (commented):

config/config.example.toml
NishantJoshi00: Currently there is no, interval tick in the drainer's loop and it will constantly keep syncing. But yes when we do add that interval tick we need to have this interval grater than the entire loop. Currently on the safer side I have kept it to 15 minutes. As, it's certain that the drainer will drain all the entries in the database with only db query and network delay to consider.

dracarys18 (dismissed):
LGTM

SanchithHegde (dismissed):

crates/router/src/db/payment_attempt.rs
```diff
@@ -534,15 +535,20 @@ mod storage {
...
+                        self.redis_conn.get_hash_field_and_deserialize(
                             &lookup.pk_id,
                             &lookup.sk_id,
                             "PaymentAttempt",
-                        )
-                        .await
-                        .map_err(|error| error.to_redis_failed_response(&key))
-                    // Check for database presence as well Maybe use a read replica here ?
+                        ),
+                        self.find_payment_attempt_by_payment_id_merchant_id(
```
jarnura, line 545: Remove recursion

crates/router/src/db/payment_attempt.rs
```diff
@@ -576,14 +582,17 @@ mod storage {
...
-                        .map_err(|error| error.to_redis_failed_response(key))
+
+                    db_utils::try_redis_get_else_try_database_get(
+                        self.redis_conn
+                            .get_hash_field_and_deserialize(
+                                key,
+                                &lookup.sk_id,
+                                "PaymentAttempt",
+                            ),
+                        self.find_payment_attempt_by_connector_transaction_id_payment_id_merchant_id(connector_transaction_id, payment_id, merchant_id, enums::MerchantStorageScheme::PostgresOnly),
```
jarnura, line 593: remove recursion everywhere.

jarnura (changes requested):

SanchithHegde (approved):

PR #514, opened 2023-02-07
refactor: use unique id for merchant_conenctor_id instead of int

dracarys18:
## Type of Change

- [x] Refactoring

## Description

Change the `merchant_connector_id` data type to use unique id.

### Additional Changes

## Motivation and Context

Since we want to implement KV for Merchant connector account table we need to have a unique id as key for the table.

## How did you test it?

image

## Checklist

### DIFF
--- a/crates/router/src/core/admin.rs
+++ b/crates/router/src/core/admin.rs
@@ -3,6 +3,7 @@ use error_stack::{report, FutureExt, ResultExt};
 use uuid::Uuid;
 
 use crate::{
+    consts,
     core::errors::{self, RouterResponse, RouterResult, StorageErrorExt},
     db::StorageInterface,
     env::{self, Env},
@@ -285,7 +286,7 @@ pub async fn create_payment_connector(
         merchant_id: Some(merchant_id.to_string()),
         connector_type: Some(req.connector_type.foreign_into()),
         connector_name: Some(req.connector_name),
-        merchant_connector_id: None,
+        merchant_connector_id: utils::generate_id(consts::ID_LENGTH, "mca"),
         connector_account_details: req.connector_account_details,
         payment_methods_enabled,
         test_mode: req.test_mode,
@@ -307,7 +308,7 @@ pub async fn create_payment_connector(
 pub async fn retrieve_payment_connector(
     store: &dyn StorageInterface,
     merchant_id: String,
-    merchant_connector_id: i32,
+    merchant_connector_id: String,
 ) -> RouterResponse<api::PaymentConnectorCreate> {
     let _merchant_account = store
         .find_merchant_account_by_merchant_id(&merchant_id)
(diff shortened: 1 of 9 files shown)


migrations/2023-02-07-070512_change_merchant_connector_id_data_type/down.sql
```diff
@@ -0,0 +1,10 @@
+CREATE SEQUENCE IF NOT EXISTS merchant_connector_id_seq OWNED BY merchant_connector_account.merchant_connector_id;
+
+UPDATE merchant_connector_account
+SET merchant_connector_id = id;
+
+ALTER TABLE merchant_connector_account
+ALTER COLUMN merchant_connector_id TYPE INTEGER USING (trim(merchant_connector_id)::integer);
+
+ALTER TABLE merchant_connector_account 
+ALTER COLUMN merchant_connector_id SET DEFAULT nextval('merchant_connector_id_seq');
```
dracarys18, line 10: So here since Serial is not a concrete type we can't use it in ALTER TABLE. So creating Sequence will emulate that behaviour.

migrations/2023-02-07-070512_change_merchant_connector_id_data_type/down.sql
```diff
@@ -0,0 +1,10 @@
+CREATE SEQUENCE IF NOT EXISTS merchant_connector_id_seq OWNED BY merchant_connector_account.merchant_connector_id;
+
+UPDATE merchant_connector_account
+SET merchant_connector_id = id;
```
dracarys18, line 4: So here for a safe conversion from varchar to int we set connector_id to id and then convert to Integer

migrations/2023-02-07-070512_change_merchant_connector_id_data_type/up.sql
```diff
@@ -0,0 +1,6 @@
+ALTER TABLE merchant_connector_account
+ALTER COLUMN merchant_connector_id TYPE VARCHAR(128) USING merchant_connector_id::varchar;
+
+
+ALTER TABLE merchant_connector_account
+ALTER COLUMN merchant_connector_id DROP DEFAULT;
```
dracarys18, line 6: So since in down.sql it creates default to sequence this is needed

jarnura (approved):

migrations/2023-02-07-070512_change_merchant_connector_id_data_type/up.sql
SanchithHegde: Would this `DEFAULT` constraint be present when `diesel migration run` is run on an empty database as well?

crates/router/src/core/admin.rs
```diff
@@ -307,7 +308,7 @@ pub async fn create_payment_connector(
 pub async fn retrieve_payment_connector(
     store: &dyn StorageInterface,
     merchant_id: String,
-    merchant_connector_id: i32,
+    merchant_connector_id: String,
```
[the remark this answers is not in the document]
dracarys18: Since we aren't cloning anywhere simply moving this from routes function. I kept it as String

migrations/2023-02-07-070512_change_merchant_connector_id_data_type/up.sql
dracarys18: Yes it will be there.

SanchithHegde (approved):

PR #515, opened 2023-02-07
feat: Impelement redis cache for MCA fetch and update

dracarys18:
## Type of Change

- [x] Enhancement

## Description

Implemented cache for MCA table.

## Motivation and Context

Impelemente redis cache for mca fetch and update for reduced latency.

## How did you test it?

Manual
image

## Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1018,8 +1018,11 @@ dependencies = [
  "serde",
  "serde_json",
  "serde_urlencoded",
+ "signal-hook",
+ "signal-hook-tokio",
  "thiserror",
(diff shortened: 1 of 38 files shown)


SanchithHegde (approved):

jarnura (approved):

dracarys18:
> Rather than have two feature based calls in every caller, would it make sense to call a single function, always get_or_populate_cache and either

But this feature based call is table based like `accounts_cache` is for merchant connector account and merchant account would it make sense to add it in the generic function.

PR #522, opened 2023-02-08
fix(core): use guard for access token result

Narayanbhat166:
## Type of Change

- [x] Bugfix

## Description

Klarna redirect payments were failing because of default error that was present in router data, which was populated by access token result.

## How did you test it?

- Create payment
Screenshot 2023-02-08 at 12 44 21 PM
- Successful redirection
Screenshot 2023-02-08 at 12 44 36 PM

## Checklist

### DIFF
--- a/crates/router/src/core/payments.rs
+++ b/crates/router/src/core/payments.rs
@@ -356,9 +356,11 @@ where
         .add_access_token(state, &connector, merchant_account)
         .await?;
 
-    match add_access_token_result.access_token_result {
-        Ok(access_token) => router_data.access_token = access_token,
-        Err(connector_error) => router_data.response = Err(connector_error),
+    if add_access_token_result.connector_supports_access_token {
+        match add_access_token_result.access_token_result {
+            Ok(access_token) => router_data.access_token = access_token,
+            Err(connector_error) => router_data.response = Err(connector_error),
+        }
     }
 
     let router_data_res = if !(add_access_token_result.connector_supports_access_token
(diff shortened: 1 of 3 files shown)


crates/router/src/core/payments.rs
```diff
@@ -356,9 +356,11 @@ where
         .add_access_token(state, &connector, merchant_account)
         .await?;
 
-    match add_access_token_result.access_token_result {
-        Ok(access_token) => router_data.access_token = access_token,
-        Err(connector_error) => router_data.response = Err(connector_error),
+    if add_access_token_result.connector_supports_access_token {
```
NishantJoshi00, line 359: can we make this into a function, we have the same pattern in 3 different places, will this guard be required in those other places?

NishantJoshi00 (changes requested):

NishantJoshi00 (dismissed):
LGTM

crates/router/src/core/payments/access_token.rs
```diff
@@ -48,6 +48,18 @@ pub fn router_data_type_conversion<F1, F2, Req1, Req2, Res1, Res2>(
     }
 }
 
+pub fn update_router_data_with_access_token_result<F, Req, Res>(
+    add_access_token_result: &types::AddAccessTokenResult,
+    router_data: &mut types::RouterData<F, Req, Res>,
+) {
+    if add_access_token_result.connector_supports_access_token {
```
jarnura, line 55: Add actual predicate for the issue

jarnura (changes requested):

jarnura (approved):

NishantJoshi00 (approved):

PR #525, opened 2023-02-09
fix: don't call connector if connector transaction id doesn't exist

NishantJoshi00:
## Type of Change

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates

## Description
In the case where we do not have the connector transaction id of a connector force_sync should not fail, rather we need it not to call the connector it this case.

### Additional Changes

## Checklist

### DIFF
--- a/crates/router/src/core/payments/helpers.rs
+++ b/crates/router/src/core/payments/helpers.rs
@@ -848,7 +848,7 @@ pub(crate) fn validate_payment_method_fields_present(
     Ok(())
 }
 
-pub fn can_call_connector(
+pub fn check_force_psync_precondition(
     status: &storage_enums::AttemptStatus,
     connector_transaction_id: &Option<String>,
 ) -> bool {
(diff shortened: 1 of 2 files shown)


crates/router/src/core/payments/helpers.rs
```diff
@@ -848,7 +848,10 @@ pub(crate) fn validate_payment_method_fields_present(
     Ok(())
 }
 
-pub fn can_call_connector(status: &storage_enums::AttemptStatus) -> bool {
+pub fn can_call_connector(
```
jarnura, line 851: change the function. This means this condition can be used everywhere

jarnura (changes requested):

jarnura (approved):

Narayanbhat166:
With the `force_sync` condition set to true, the merchant or the caller expects to always get the response from connector. How is setting `force_sync` to `true` and also adding our own validation help here? This fact is being hidden from the caller of the api. There is no way to know whether the status is synced or we have the stale status. If we are deciding whether to call the connector or not, then why are we asking the merchant or caller to set that field?

Narayanbhat166 (approved):

PR #527, opened 2023-02-09
refactor: Throw 500 error on database connection error instead of panic

dracarys18:
## Type of Change

- [x] Refactoring

## Description

Do not panic while getting db connection. Instead of that just throw 500 error. This PR also makes connection timeout in Database configurable.

### Additional Changes

## Motivation and Context

When database goes down we don't want our application panic and stop actix workers since restarting it has much overhead. So instead just 500 error.

## How did you test it?

Start the application and bring down the database and try to make a payment. Throws 500 error.
image

## Checklist

### DIFF
--- a/crates/router/src/connection.rs
+++ b/crates/router/src/connection.rs
@@ -1,8 +1,9 @@
 use async_bb8_diesel::{AsyncConnection, ConnectionError};
 use bb8::{CustomizeConnection, PooledConnection};
 use diesel::PgConnection;
+use error_stack::{IntoReport, ResultExt};
 
-use crate::configs::settings::Database;
+use crate::{configs::settings::Database, errors};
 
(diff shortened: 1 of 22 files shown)


jarnura (approved):

crates/router/src/connection.rs
```diff
@@ -56,8 +59,12 @@ pub async fn diesel_make_pg_pool(database: &Database, test_transaction: bool) ->
 #[allow(clippy::expect_used)]
 pub async fn pg_connection(
```
SanchithHegde, line 60: You can remove the `#[allow(...)]` annotation.

SanchithHegde (approved):

SanchithHegde (approved):

lsampras (approved):

jarnura (approved):

PR #531, opened 2023-02-10
fix: throw 500 error when redis goes down

dracarys18:
## Type of Change

- [x] Enhancement

## Description

Created a global flag to check if redis is available or not if it's not available send 500 error to all reamaining requests.

## Motivation and Context

Currently if redis goes down currently after `reconnect_attempts` reached fred will currently close the client and block the whole application. 

## How did you test it?

Manual
Screenshot 2023-02-10 at 6 10 18 PM
Screenshot 2023-02-10 at 6 10 31 PM

## Checklist

### DIFF
--- a/crates/router/src/services.rs
+++ b/crates/router/src/services.rs
@@ -3,10 +3,15 @@ pub mod authentication;
 pub mod encryption;
 pub mod logger;
 
-use std::sync::Arc;
+use std::sync::{atomic, Arc};
+
+use redis_interface::errors::RedisError;
 
 pub use self::{api::*, encryption::*};
-use crate::connection::{diesel_make_pg_pool, PgPool};
+use crate::{
+    connection::{diesel_make_pg_pool, PgPool},
+    core::errors,
+};
 
 #[derive(Clone)]
 pub struct Store {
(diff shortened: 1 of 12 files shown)


crates/router/src/services.rs
```diff
@@ -46,6 +58,20 @@ impl Store {
...
 
+    pub fn redis_conn(
+        &self,
+    ) -> errors::CustomResult<Arc<redis_interface::RedisConnectionPool>, RedisError> {
+        if self
+            .redis_conn
+            .is_redis_available
+            .load(atomic::Ordering::SeqCst)
+        {
+            Ok(self.redis_conn.clone())
```
jarnura, line 69: We can removing this clone() right?

jarnura (approved):

crates/router/src/services.rs
dracarys18: Since we take `&self` in this function and redis_conn doesn't implement copy. `redis_conn` will be moved out of self if we don't use clone here which will throw error

dracarys18: And since this is an `Arc` clone is very less expensive

SanchithHegde (approved):

SanchithHegde (approved):

lsampras (approved):

jarnura (approved):

PR #549, opened 2023-02-11
feat:H1116-L2 - Payment authorise and sync for Bambora

hey-ritikgupta:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates

### Additional Changes

## Checklist

### DIFF
--- a/crates/router/src/connector/bambora.rs
+++ b/crates/router/src/connector/bambora.rs
@@ -84,7 +84,80 @@ impl
         types::PaymentsCancelData,
         types::PaymentsResponseData,
     > for Bambora
-{}
+{
+    fn get_headers(
+        &self,
+        req: &types::PaymentsCancelRouterData,
+        connectors: &settings::Connectors,
+    ) -> CustomResult<Vec<(String, String)>, errors::ConnectorError> {
+        self.build_headers(req, connectors)
+    }
+
+    fn get_content_type(&self) -> &'static str {
+        self.common_get_content_type()
+    }
+
+    fn get_url(
+        &self,
+        req: &types::PaymentsCancelRouterData,
+        _connectors: &settings::Connectors,
+    ) -> CustomResult<String, errors::ConnectorError> {
+        let connector_payment_id = req.request.connector_transaction_id.clone();
+        Ok(format!(
+            "{},v1/payments/{}/void",
+            self.base_url(_connectors),
+            connector_payment_id
+        ))
+    }
+
+    fn get_request_body(
+        &self,
+        req: &types::PaymentsCancelRouterData
+    ) -> CustomResult<Option<String>,errors::ConnectorError> {
+        Ok(Some("{}".to_string()))
+    }
+
+    fn build_request(
+        &self,
+        req: &types::PaymentsCancelRouterData,
+        connectors: &settings::Connectors,
+    ) -> CustomResult<Option<services::Request>, errors::ConnectorError> {
+        Ok(Some(
+            services::RequestBuilder::new()
+                .method(services::Method::Post)
+                .url(&types::PaymentsVoidType::get_url(self, req, connectors)?)
+                .headers(types::PaymentsVoidType::get_headers(self, req, connectors)?)
+                .body(self.get_request_body(req)?)
+                .build(),
+        ))
+    }
+
+    fn handle_response(
+        &self,
+        data: &types::PaymentsCancelRouterData,
+        res: Response,
+    ) -> CustomResult<types::PaymentsCancelRouterData, errors::ConnectorError> {
+        let response: bambora:: BamboraPaymentsResponse = res
+            .response
+            .parse_struct("bambora PaymentsResponse")
+            .change_context(errors::ConnectorError::ResponseDeserializationFailed)?;
+        types::RouterData::try_from(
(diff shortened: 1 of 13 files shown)


crates/router/src/connector/bambora.rs
```diff
@@ -0,0 +1,554 @@
...
+
+    fn get_url(
+        &self,
+        _req: &types::PaymentsCaptureRouterData,
+        connectors: &settings::Connectors,
+    ) -> CustomResult<String, errors::ConnectorError> {
+        Ok(format!(
+            "{}{}",
+            self.base_url(connectors),
+            "v1/payments"
```
jagan-jaya, line 269: How is payments authorize different from payment capture for this connector?, i don't see that flow specifically handled. both of them are same. Ideally it should not be

crates/router/src/connector/bambora/transformers.rs
```diff
@@ -0,0 +1,194 @@
...
+    approved: BamboraPaymentStatus,
+    id: String,
+    response: Option<String>
+}
+
+impl<F,T> TryFrom<types::ResponseRouterData<F, BamboraPaymentsResponse, T, types::PaymentsResponseData>> for types::RouterData<F, T, types::PaymentsResponseData> {
+    type Error = error_stack::Report<errors::ParsingError>;
+    fn try_from(item: types::ResponseRouterData<F, BamboraPaymentsResponse, T, types::PaymentsResponseData>) -> Result<Self,Self::Error> {
+        Ok(Self {
+            status: enums::AttemptStatus::Charged,
```
jagan-jaya, line 110: how are the handling the void flow here, where are the status mappings for void responses
Refer: https://github.com/juspay/hyperswitch/blob/e4956820fdfea8f637600b7fd03ac666f3b19c21/crates/router/src/connector/adyen/transformers.rs#L654

jagan-jaya (changes requested):

crates/router/src/connector/bambora.rs
```diff
@@ -0,0 +1,589 @@
...
+    }
+}
+
+impl ConnectorCommon for Bambora {
+    fn id(&self) -> &'static str {
+        "bambora"
+    }
+
+    fn common_get_content_type(&self) -> &'static str {
+        "application/json"
```
Narayanbhat166, line 53: If you want to avoid spelling mistakes, `mime::APPLICATION_JSON` this can be used.

crates/router/src/connector/bambora/transformers.rs
```diff
@@ -0,0 +1,217 @@
...
+    }
+}
+
+impl TryFrom<&types::PaymentsAuthorizeRouterData> for BamboraPaymentsRequest {
+    type Error = error_stack::Report<errors::ConnectorError>;
+    fn try_from(item: &types::PaymentsAuthorizeRouterData) -> Result<Self, Self::Error> {
+        let payment_method_data = item.request.payment_method_data.clone();
+        let payment_method = match payment_method_data {
+            api::PaymentMethod::Card(ref _item) => String::from("card"),
+            _ => todo!(),
```
Narayanbhat166, line 38: Raise `ConnectorError::NotImplemented` instead of todo!()

crates/router/src/connector/bambora.rs
hey-ritikgupta: we are sending complete : false in the payload

it's named as Pre-Auth payments in Bambora
Ref: https://dev-apac.bambora.com/checkout/guides/handling-payments/payment-api#:~:text=Bambora%20offers%20the%20ability%20to,real%2Dtime%20via%20the%20API.

crates/router/src/connector/bambora/transformers.rs
```diff
@@ -0,0 +1,217 @@
...
+        match item {
+            BamboraPaymentStatus::ONE => Self::Charged,
+            BamboraPaymentStatus::ZERO => Self::Failure,
+        }
+    }
+}
+
+#[derive(Default, Debug, Clone, Serialize, Deserialize, PartialEq)]
+pub struct BamboraPaymentsResponse {
+    approved: BamboraPaymentStatus,
```
Narayanbhat166, line 97: Rename this field appropriately, it must signify that it is a status.

If Few fields are optional then keep them optional, This might fail if some fields are present only for specfic payment types. The deserialization might fail in that case

Also keep the fields that are only necessary, this will avoid unnecessary deserialization

crates/router/src/connector/bambora/transformers.rs
hey-ritikgupta: According to the doc of Bambora all void payments get cancelled 
we just need to pass id and amount
Screenshot 2023-02-12 at 1 41 17 AM

Narayanbhat166 (changes requested):

jarnura (commented):
Is this a duplicate PR?

ArjunKarthik:
Thanks for the contribution @hey-ritikgupta  , We are merging the changes from different participant as it has more flows implemented. You can still continue your contributions on the [open issues](https://github.com/juspay/hyperswitch/issues?q=is%3Aissue+is%3Aopen+label%3AA-connector-integration).

PR #569, opened 2023-02-12
feat(datetime): Changing date from utc to iso 

Vignesh-772:


jagan-jaya:
@Vignesh-772  Unfortunately we couldn't proceed with this PR as similar one is already merged. Thanks for your contribution. Closing in favour of https://github.com/juspay/hyperswitch/pull/955

PR #614, opened 2023-02-17
feat: Add support for a redis pubsub interface

dracarys18:
## Type of Change

- [x] New feature

## Description

This commit adds two seperate redis clients `subscriber` and `publisher` for pubsub interface.

## Motivation and Context

We need this pubsub interface to invalidate keys in redis. So a subscriber subscribes to a channel named `hyperswitch_invalidate` and there we publish the keys and invalidate it.

## How did you test it?

Manual.

## Checklist

### DIFF
--- a/crates/redis_interface/src/errors.rs
+++ b/crates/redis_interface/src/errors.rs
@@ -56,6 +56,6 @@ pub enum RedisError {
     RedisConnectionError,
     #[error("Failed to subscribe to a channel")]
     SubscribeError,
-    #[error("Failed to publish to a subscriber")]
+    #[error("Failed to publish to a channel")]
     PublishError,
 }
--- a/crates/redis_interface/src/errors.rs
+++ b/crates/redis_interface/src/errors.rs
@@ -54,4 +54,8 @@ pub enum RedisError {
     InvalidRedisEntryId,
     #[error("Failed to establish Redis connection")]
     RedisConnectionError,
+    #[error("Failed to subscribe to a channel")]
+    SubscribeError,
+    #[error("Failed to publish to a subscriber")]
+    PublishError,
 }
--- a/crates/redis_interface/src/lib.rs
+++ b/crates/redis_interface/src/lib.rs
@@ -191,10 +191,8 @@ impl PubSubInterface for RedisConnectionPool {
     }
     #[inline]
     async fn on_message(&self) {
-        let mut message = self.subscriber.on_message();
-        while let Some((_, key)) = message.next().await {
-            logger::error!("{:?}", key);
-        }
+        let message = self.subscriber.on_message();
+        message.for_each(|_key| futures::future::ready(())).await;
     }
 }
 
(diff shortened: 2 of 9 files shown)


crates/redis_interface/src/lib.rs
```diff
@@ -117,6 +172,32 @@ impl RedisConnectionPool {
...
+    #[inline]
+    async fn publish(&self, channel: &str, key: &str) -> CustomResult<usize, errors::RedisError> {
+        self.publisher.publish(channel, key).await
+    }
+    #[inline]
+    async fn on_message(&self) {
+        let mut message = self.subscriber.on_message();
+        while let Some((_, key)) = message.next().await {
+            logger::error!("{:?}", key);
+        }
```
SanchithHegde, line 197: 1. Can we use `for_each()` maybe?
2. Is the key being logged at the `ERROR` level intentional? Maybe include some context as to what's the error here?
   If you're updating this snippet, a minor change: `logger::error!("{key:?}");`

crates/redis_interface/src/lib.rs
```diff
@@ -35,9 +35,57 @@ pub struct RedisConnectionPool {
...
+        key: impl AsRef<str>,
+    ) -> CustomResult<usize, errors::RedisError> {
+        self.inner
+            .publish(channel.as_ref(), key.as_ref())
+            .await
+            .into_report()
+            .change_context(errors::RedisError::SubscribeError)
+    }
+    #[inline]
+    pub fn on_message(&self) -> fred::prelude::AsyncStream<(String, fred::prelude::RedisValue)> {
```
SanchithHegde, line 85: Can we prevent exposing `fred`'s types in return types? If it's not possible in this case, that's fine, not a necessity.

crates/redis_interface/src/lib.rs
dracarys18: that log is just a placeholder I will change it to appropriate behaviour

crates/redis_interface/src/lib.rs
```diff
@@ -35,9 +35,57 @@ pub struct RedisConnectionPool {
     pub pool: fred::pool::RedisPool,
     config: RedisConfig,
     join_handles: Vec<fred::types::ConnectHandle>,
+    subscriber: RedisClient,
+    publisher: RedisClient,
     pub is_redis_available: Arc<atomic::AtomicBool>,
 }
 
+pub struct RedisClient {
+    inner: fred::prelude::RedisClient,
```
Narayanbhat166, line 44: I think we can have a deref trait implemented for this to avoid accessng the client through inner all the time.

crates/redis_interface/src/errors.rs
```diff
@@ -54,4 +54,8 @@ pub enum RedisError {
     InvalidRedisEntryId,
     #[error("Failed to establish Redis connection")]
     RedisConnectionError,
+    #[error("Failed to subscribe to a channel")]
+    SubscribeError,
+    #[error("Failed to publish to a subscriber")]
```
Narayanbhat166, line 59: maybe rename this to failed to publish to a `channel`?

crates/redis_interface/src/lib.rs
SanchithHegde: Reopening this conversation since there's nothing being done with `key` as of now:

https://github.com/juspay/hyperswitch/blob/cd1e21b2de00536a6d20457143f7897d3feacdf1/crates/redis_interface/src/lib.rs#L182-L185

jarnura (approved):

SanchithHegde (approved):

NishantJoshi00 (approved):

ashokkjag (approved):
LGTM, Please check if delete_key is the only action required in on_message, or should we leave it generic for future use cases.

Issue #353, opened 2023-01-11
fix(connector): redirection status is `requires_payment_method` for failed payments created through stripe


Narayanbhat166:
### Bug Description

The payment status of a failed redirection payment using `stripe` says requires_payment_method instead of failed.

### Expected Behavior

The status should be failed, with appropriate error message if possible

### Actual Behavior

The status is `requires_payment_method`

### Steps To Reproduce

Provide an unambiguous set of steps to reproduce this bug. Include code or configuration to reproduce, if relevant.
1. Create a payment through stripe which requires a redirection flow
2. Fail the payment when prompted during redirection
3. The status should be visible in the url.

### Context For The Bug

Attaching the redirect url for reference which has this bug
http://google.com/success?status=requires_payment_method&payment_intent_client_secret=pay_QU75MkmagL59TuG2dAdn&amount=6545

### Environment

Are you using hyperswitch hosted version? No
If yes, please provide the value of the `x-request-id` response header for helping us debug your issue.

If not (or if building/running locally), please provide the following details:
1. Operating System or Linux distribution: Mac
2. Rust version (output of `rustc --version`): `1.66.0`
3. App version (output of `cargo r -- --version`): `0.2.0-48af156-2023-01-11T10:11:56Z`

### Have you spent some time to check if this bug has been raised before?

- [X] I checked and didn't find similar issue

### Have you read the Contributing Guidelines?

- [X] I have read the [Contributing Guidelines](https://github.com/juspay/hyperswitch/blob/main/contrib/CONTRIBUTING.md)

### Are you willing to submit a PR?

Yes I am willing to submit a PR!

dracarys18:
This shouldn't be failed status. Stripe sends the status as `requires_payment_method` with `last_payment_error`. So  we should cascade their behavior and keep the status as `requires_payment_method` with `error_message` same as last_payment_error.

Narayanbhat166:
@Abhicodes-crypto are we handling this in retries? Can you comment on the status that hyperswitch should have in this case?

Abhicodes-crypto:
Having the payment status flagged as "requires_payment_method" is important because it allows the SDK or merchant client app to make additional attempts at confirming the payment. 

On the other hand, if the status is categorized as "failed," this would indicate that the payment has definitively failed and cannot be retried without obtaining a new client secret. Such an approach would result in a poor user experience for the customer.

Narayanbhat166:
Since this is the expected behaviour, I will close this issue.

PR #615, opened 2023-02-20
feat(stripe): get error message for failed redirect payments

Narayanbhat166:
## Type of Change

- [x] Enhancement

## Description

During the consume flow i.e when the status is available in the url, if the status is failed then we are unable to get the error message as to why this payment failed because we will only be updating the status and not the error message and code. 
So call payments_sync if the status is failed to get the error message and code. The `PaymentSync` struct of stripe is created accordingly to get the `last_payment_error`.

## Motivation and Context

We will not be calling psync if the status is failed. So we never get to know why the payment failed during consume flow. So we call psync if the status is failed during the last step of redirection. 

## How did you test it?

Create a payment which uses consume flow ( klarna redirect ) and then fail it.
Screenshot 2023-02-20 at 1 45 18 PM

Screenshot 2023-02-20 at 1 31 07 PM

Error message and code updated in db.

## Checklist


crates/router/src/core/payments/operations/payment_response.rs
```diff
@@ -268,7 +268,7 @@ async fn payment_response_update_tracker<F: Clone, T>(
         Err(err) => (
             Some(storage::PaymentAttemptUpdate::ErrorUpdate {
                 connector: Some(router_data.connector.clone()),
-                status: storage::enums::AttemptStatus::Failure,
+                status: router_data.status.foreign_into(),
```
Narayanbhat166, line 271: This will use the same status as provided by the connector processing step for updating the status. referencing this issue for more information #353

crates/router/src/connector/stripe/transformers.rs
```diff
@@ -548,6 +564,74 @@ impl<F, T>
...
+            item.response
+                .payment_method_options
+                .to_owned()
+                .and_then(|payment_method_options| match payment_method_options {
+                    StripePaymentMethodOptions::Card {
+                        mandate_options, ..
+                    } => mandate_options.map(|mandate_options| mandate_options.reference),
+                    StripePaymentMethodOptions::Klarna {} => None,
+                    StripePaymentMethodOptions::Affirm {} => None,
+                    StripePaymentMethodOptions::AfterpayClearpay {} => None,
```
SanchithHegde, line 602:
```suggestion
                    StripePaymentMethodOptions::Klarna {}
                    | StripePaymentMethodOptions::Affirm {}
                    | StripePaymentMethodOptions::AfterpayClearpay {} => None,
```

jarnura (approved):

SanchithHegde (approved):

jarnura (approved):

PR #623, opened 2023-02-20
feat: Adding support for multiple account creation for a merchant<>connector based on country

manoj-juspay:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates

## Description
Adding support so that merchant can have multiple account for a connector based on country
The unique identity will be `connector_label` for mca record

### Additional Changes

## How did you test it?
Create MCA

## Checklist


crates/router/src/core/admin.rs
```diff
@@ -48,6 +48,15 @@ pub async fn create_merchant_account(
...
     );
 
+    let primary_business_details = Some(
+        utils::Encode::<api::PrimaryBusinessDetails>::encode_to_value(
+            &req.primary_business_details,
+        )
+        .change_context(errors::ApiErrorResponse::InvalidDataValue {
+            field_name: "default business details",
+        })?,
+    );
```
SanchithHegde, line 58: Wouldn't this also encode the `None` case as `Some(serde_json::Value::Null)` instead of directly being mapped as a `None`? Maybe use a `map()` instead?

migrations/2023-02-20-101809_update_merchant_connector_account/down.sql
```diff
@@ -0,0 +1,7 @@
+ALTER TABLE merchant_connector_account
```
SanchithHegde, line 1: It would have been better if the migration directory names provided a hint on the changes being made. Same for the other migration directory name as well.

SanchithHegde (commented):

crates/router/src/core/admin.rs
manoj-juspay: the entry in the database will be `null` in case its a `None` value

migrations/2023-02-20-101809_update_merchant_connector_account/down.sql
manoj-juspay: sure will try to name if possible as a hint of change in the future

vspecky (approved):
Other than the changes Sanchith mentioned, LGTM.

crates/storage_models/src/merchant_connector_account.rs
```diff
@@ -41,6 +44,9 @@ pub struct MerchantConnectorAccountNew {
     pub merchant_connector_id: String,
     pub payment_methods_enabled: Option<Vec<serde_json::Value>>,
     pub metadata: Option<serde_json::Value>,
+    pub connector_label: Option<String>,
```
Narayanbhat166, line 47: Since you have mentioned that this will be unique identity, why is this field optional?

manoj-juspay: this is taken care in api_models , there can be default values as well

SanchithHegde (approved):

crates/api_models/src/admin.rs
```diff
@@ -241,6 +255,15 @@ pub struct PaymentConnectorCreate {
...
     pub connector_name: String,
+    /// Connector label for specific country and Business
+    #[schema(example = "stripe_US_travel")]
+    pub connector_label: String,
+    /// Country through which payment should be processed
+    #[schema(example = "US")]
+    pub business_country: String,
+    ///Business Type of the merchant
+    #[schema(example = "travel")]
+    pub business_type: String,
```
jarnura, line 266: change this to `business_label`

jarnura (changes requested):

crates/api_models/src/admin.rs
manoj-juspay: done

crates/api_models/src/admin.rs
```diff
@@ -240,6 +251,13 @@ pub enum RoutingAlgorithm {
     Single(api_enums::RoutableConnectors),
 }
 
+#[derive(Clone, Debug, Deserialize, ToSchema, Serialize)]
+#[serde(deny_unknown_fields)]
+pub struct PrimaryBusinessDetails {
+    country: String,
```
jarnura, line 257: create a country enum. Can we taken as separate PR.

crates/router/src/core/admin.rs
```diff
@@ -34,19 +34,38 @@ pub async fn create_merchant_account(
...
-    let webhook_details = Some(
-        utils::Encode::<api::WebhookDetails>::encode_to_value(&req.webhook_details)
-            .change_context(errors::ApiErrorResponse::InvalidDataValue {
-                field_name: "webhook details",
-            })?,
-    );
+    let merchant_details = req
+        .merchant_details
+        .map(|md| {
+            utils::Encode::<api::PrimaryBusinessDetails>::encode_to_value(&md).change_context(
```
jarnura, line 40: Why merchant details are encoded into `PrimaryBusinessDetails`?

crates/api_models/src/admin.rs
```diff
@@ -66,6 +66,10 @@ pub struct MerchantAccountCreate {
     /// An identifier for the vault used to store payment method information.
     #[schema(example = "locker_abc123")]
     pub locker_id: Option<String>,
+
+    ///Default business details for connector routing
+    #[schema(value_type = Option<PrimaryBusinessDetails>)]
+    pub primary_business_details: Option<serde_json::Value>,
```
jarnura, line 72: We are going to support business_details for muliple connector, so we don't need this add for merchant account and other merchant account related structs also

jarnura (changes requested):

jarnura (commented):
change the base branch to release_candidate_v0.5.0

crates/api_models/src/admin.rs
manoj-juspay: we are using country as String, needs to be changed at all places ?

crates/api_models/src/admin.rs
manoj-juspay: This is the default business details for a merchant

crates/router/src/core/admin.rs
manoj-juspay: corrected!

crates/api_models/src/admin.rs
jarnura: yes

jarnura (approved):

SanchithHegde (approved):

PR #639, opened 2023-02-24
refactor(api_keys): use a KMS encrypted API key hashing key and remove key ID prefix from plaintext API keys

SanchithHegde:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [x] Refactoring
- [ ] Dependency updates

## Description

This PR modifies the API keys code to use a KMS encrypted API key hashing key instead of using a plaintext hashing key stored in the database.

This includes the following changes:

1. Extracts the KMS code to a separate sub-module within the `services` module.
2. Adds new application configuration for API keys.
3. Modifies API key hashing code to use a KMS encrypted API key hashing key.

### Additional Changes

The migrations can be found at: [`up.sql`](https://github.com/juspay/hyperswitch/blob/4f22018df818e23c6fd6664d6605697bf64e7afa/migrations/2023-02-21-094019_api_keys_remove_hash_key/up.sql) and [`down.sql`](https://github.com/juspay/hyperswitch/blob/4f22018df818e23c6fd6664d6605697bf64e7afa/migrations/2023-02-21-094019_api_keys_remove_hash_key/down.sql).

I'm also clearing API keys from the table since we can't use the old API keys once we add/remove the `hash_key` column to/from the `api_keys` table. Moreover, since the API key endpoints are not being actively used on sandbox, I don't expect this to be too much of a problem.

As for the application configuration:

1. If you're not using KMS (the `kms` feature is disabled), use the value generated by the following command for the `api_keys.hash_key` application configuration variable (we generate a 32 byte long random key and hex encode it):

   ```shell
   $ openssl rand -hex 32
   0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef
   ```

2. If you're using KMS:

   1. Set the `api_keys.aws_key_id` and `api_keys.aws_region` application configuration variables to the AWS key ID and region for the KMS key being used.
   2. Generate a 32 byte long random key, hex encode it, then base64 encode it using the command:

      ```shell
      $ openssl rand -hex 32 | tr -d \\n | base64
      MDEyMzQ1Njc4OWFiY2RlZjAxMjM0NTY3ODlhYmNkZWYwMTIzNDU2Nzg5YWJjZGVmMDEyMzQ1Njc4OWFiY2RlZg==
      ```

   3. Use the AWS KMS key to KMS encrypt the API key hashing key, and use the output ciphertext generated by the following command as the value for the `api_keys.kms_encrypted_hash_key` application configuration variable.

      ```shell
      aws kms encrypt --output json --key-id my-aws-key-id --plaintext MDEyMzQ1Njc4OWFiY2RlZjAxMjM0NTY3ODlhYmNkZWYwMTIzNDU2Nzg5YWJjZGVmMDEyMzQ1Njc4OWFiY2RlZg==
      ```

## Motivation and Context

Using a KMS encrypted hashing key should slightly increase the security, even if the database containing the API keys is breached.

## How did you test it?

Manual.

## Checklist


ashokkjag:
LGTM, but should the hashed keys etc be passed from toml config file?

SanchithHegde:
> LGTM, but should the hashed keys etc be passed from toml config file?

The encrypted keys can also go in the environment variables, due to the way our application configuration is set up.

crates/router/src/core/api_keys.rs
```diff
@@ -12,13 +13,45 @@ use crate::{
...
 };
 
+pub static HASH_KEY: tokio::sync::OnceCell<StrongSecret<[u8; PlaintextApiKey::HASH_KEY_LEN]>> =
+    tokio::sync::OnceCell::const_new();
+
+pub async fn get_hash_key(
+    api_key_config: &settings::ApiKeys,
+) -> errors::RouterResult<StrongSecret<[u8; PlaintextApiKey::HASH_KEY_LEN]>> {
+    #[cfg(feature = "kms")]
+    let hash_key = crate::services::kms::KeyHandler::get_kms_decrypted_key(
```
jarnura, line 23: remove absolute path here

migrations/2023-02-21-094019_api_keys_remove_hash_key/up.sql
```diff
@@ -0,0 +1,11 @@
+ALTER TABLE api_keys DROP COLUMN hash_key;
+
+/*
+ Once we've dropped the `hash_key` column, we cannot use the existing API keys
+ from the `api_keys` table anymore, as the `hash_key` is a random string that
+ we no longer have.
+ */
+DELETE FROM api_keys;
```
jarnura, line 8: Use Truncate instead of delete since there is no where clause.

jarnura: In down.sql too.

jarnura (changes requested):

crates/router/src/core/api_keys.rs
SanchithHegde: Then that import would end up behind the `kms` feature flag, since this is the only place where this function is used. Would you prefer that?

jarnura: Do the same in the import place itself so the we follow one approach of importing.

jarnura (dismissed):

jarnura:
Blocked because of database changes, will take this up on or after v0.4.0

ashokkjag (dismissed):
How did you test the KMS feature?

SanchithHegde:
> How did you test the KMS feature?

I set up KMS on an AWS account and verified that KMS worked as expected.

jarnura (dismissed):

jarnura (approved):

ashokkjag (approved):

PR #656, opened 2023-02-26
refactor: Pass country and currency as json format in MCA

manoj-juspay:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [x] Refactoring
- [ ] Dependency updates

## Description
Supported country and currency , the size exceed the allowed request size, refactoring how these field are passed in the MCA and in the List PM api

### Additional Changes

## Checklist

### DIFF
--- a/crates/api_models/src/admin.rs
+++ b/crates/api_models/src/admin.rs
@@ -270,14 +270,12 @@ pub struct PaymentConnectorCreate {
                 "Discover"
             ],
             "accepted_currencies": {
-                "enable_all":false,
-                "disable_only": ["INR", "CAD", "AED","JPY"],
-                "enable_only": ["EUR","USD"]
+                "type": "enable_only",
+                "list": ["USD", "EUR"]
             },
             "accepted_countries": {
-                "enable_all":false,
-                "disable_only": ["FR", "DE","IN"],
-                "enable_only": ["UK","AU"]
+                "type": "disable_only",
+                "list": ["FR", "DE","IN"]
             },
             "minimum_amount": 1,
             "maximum_amount": 68607706,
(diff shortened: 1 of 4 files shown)


crates/api_models/src/admin.rs
```diff
@@ -342,28 +338,26 @@ pub struct PaymentMethods {
     pub payment_experience: Option<Vec<api_enums::PaymentExperience>>,
 }
 
-/// List of enabled and disabled currencies
+/// List of enabled and disabled currencies, empty in case all currencies are enabled
```
Narayanbhat166, line 341: this could be an enum with serde tag as type with 3 variants, `EnableAll`, `EnableOnly(Vec<_>)`, `DisableOnly(Vec<_>)`.

manoj-juspay: can be with only `EnableOnly(Vec<_>)` and `DisableOnly(Vec<_>)`  as enum field
not passing means `enable_all`

manoj-juspay: @jarnura @bernard-eugine  ?

jarnura: `enable_all:true` is default, and they can pass `enable_all:false` means  disable all.

jarnura: @bernard-eugine Can you confirm once do we need disable all?

jarnura (changes requested):
Other than the disable all, LGTM.

crates/api_models/src/admin.rs
jarnura: DIsable all is not required, verified with @bernard-eugine

jarnura (approved):

crates/api_models/src/admin.rs
```diff
@@ -342,28 +338,26 @@ pub struct PaymentMethods {
...
 pub struct AcceptedCountries {
-    /// True in case all countries are supported
-    pub enable_all: bool,
-    /// List of disabled countries, provide in case only few of countries are not supported
-    pub disable_only: Option<Vec<String>>,
-    /// List of enable countries, provide in case only few of countries are supported
-    pub enable_only: Option<Vec<String>>,
+    /// Type of accepted countries (disable_only, enable_only)
+    #[serde(rename = "type")]
+    pub accept_type: String,
```
lsampras, line 358: this could be an enum

crates/api_models/src/admin.rs
```diff
@@ -342,28 +338,26 @@ pub struct PaymentMethods {
...
+    /// List of countries of the provided type
+    pub list: Option<Vec<String>>,
```
lsampras, line 360: are these 2 letter country codes?
these could probably be enum as well

crates/router/src/core/payment_methods/cards.rs
```diff
@@ -472,19 +472,25 @@ fn filter_pm_country_based(
 ) -> (Option<admin::AcceptedCountries>, Option<Vec<String>>, bool) {
     match (accepted_countries, req_country_list) {
         (None, None) => (None, None, true),
-        (None, Some(ref r)) => (None, Some(r.to_vec()), false),
+        (None, Some(ref r)) => (
```
lsampras, line 475: do we need `ref` seems you should be able to consume the value instead of `ref + to_owned`

lsampras (approved):
Minor nit,
Looks good to merge

PR #708, opened 2023-03-03
feat(connector): POC for using secret in headers

NishantJoshi00:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring
- [ ] Dependency updates

## Description

This PR does the following:
- It allows the `(key, value)` pairs in headers (i.e. while calling the connector) to have normal string values as well as `Secret<String>` which provides an abstraction for obfuscating it in some cases while still being able to log other values.
- This is still a POC and a minor refactoring will be required at the connector side, to have proper use of this mechanism.

### Additional Changes

## Motivation and Context
After a discussion with @jagan-jaya and @Narayanbhat166 we decided that we some some way to hide merchant related PII, from our logs, while not completely hiding it, for ease of debugging.

## How did you test it?

Screenshot 2023-03-03 at 4 22 16 PM

## Checklist


connector-template/mod.rs
```diff
@@ -43,7 +43,7 @@ where
         &self,
         _req: &types::RouterData<Flow, Request, Response>,
         _connectors: &settings::Connectors,
-    ) -> CustomResult<Vec<(String, String)>, errors::ConnectorError> {
+    ) -> CustomResult<Vec<\(String, Box<dyn services::request::HeaderValue>\)>, errors::ConnectorError> {
```
SanchithHegde, line 46: What's with these backslashes here in the connector template?

crates/masking/src/secret.rs
```diff
@@ -152,3 +152,9 @@ where
         SecretValue::default().into()
     }
 }
+
+impl ToString for Secret<String, crate::strategy::ApiKey> {
+    fn to_string(&self) -> String {
+        self.inner_secret.clone()
+    }
+}
```
SanchithHegde, line 160: Would this prevent `Secret<String>` types from being logged/formatted in plaintext when passed to either the logging macros or the `format!()` macro?

crates/router/src/connector/aci.rs
```diff
@@ -35,11 +36,15 @@ impl ConnectorCommon for Aci {
...
-    ) -> CustomResult<Vec<(String, String)>, errors::ConnectorError> {
+    ) -> CustomResult<Vec<(String, masking::Secret<String, masking::ApiKey>)>, errors::ConnectorError>
+    {
         let auth: aci::AciAuthType = auth_type
             .try_into()
             .change_context(errors::ConnectorError::FailedToObtainAuthType)?;
-        Ok(vec![(headers::AUTHORIZATION.to_string(), auth.api_key)])
+        Ok(vec![(
+            headers::AUTHORIZATION.to_string(),
+            auth.api_key.into(),
```
SanchithHegde, line 46: If possible, let's have the API key fields in connector-specific auth types as `Secret<_, ApiKey>` please.

crates/router/src/connector/aci.rs
```diff
@@ -102,16 +107,25 @@ impl
...
+        let api_key = self
+            .get_auth_header(&req.connector_auth_type)?
+            .into_iter()
+            .map(
+                |(key, value)| -> (String, Box<dyn services::request::HeaderValue>) {
+                    (key, Box::new(value))
+                },
+            );
+        // header.append(&mut api_key);
+        header.extend(api_key);
```
SanchithHegde, line 128: I believe you'd prefer using `concat_headers()` instead.

crates/router/src/connector/aci.rs
```diff
@@ -102,16 +107,25 @@ impl
         &self,
         req: &types::PaymentsSyncRouterData,
         _connectors: &settings::Connectors,
-    ) -> CustomResult<Vec<(String, String)>, errors::ConnectorError> {
-        let mut header = vec![
+    ) -> CustomResult<Vec<(String, Box<dyn services::request::HeaderValue>)>, errors::ConnectorError>
```
SanchithHegde, line 110: Consider using a type alias for `Box<dyn services::request::HeaderValue>` or `Vec<(String, Box<dyn services::request::HeaderValue>)>` maybe?

crates/router/src/connector/dlocal.rs
```diff
@@ -69,18 +70,18 @@ where
...
+            (headers::X_VERSION.to_string(), "2.1".to_string().into()),
+            (headers::X_DATE.to_string(), date.into()),
             (
                 headers::CONTENT_TYPE.to_string(),
-                Self.get_content_type().to_string(),
+                Self.get_content_type().to_string().into(),
             ),
         ];
-        Ok(headers)
+        Ok(concat_headers(headers, Vec::<(String, String)>::new()))
```
SanchithHegde, line 84: Directly return `headers` instead? Follow that for every other place you're concatinating with an empty `Vec`.

crates/router/src/connector/worldline.rs
```diff
@@ -87,9 +88,22 @@ where
             self.generate_authorization_token(auth, &http_method, content_type, &date, &endpoint)?;
 
         Ok(vec![
-            (headers::DATE.to_string(), date),
-            (headers::AUTHORIZATION.to_string(), signed_data),
-            (headers::CONTENT_TYPE.to_string(), content_type.to_string()),
+            (
+                headers::DATE.to_string(),
+                Box::new(Into::<masking::Secret<String, masking::ApiKey>>::into(date)),
```
SanchithHegde, line 93: Are the type specifications here necessary?

crates/router/src/types/api.rs
```diff
@@ -39,10 +39,12 @@ pub trait ConnectorCommon {
     fn id(&self) -> &'static str;
 
     /// HTTP header used for authorization.
+    #[allow(clippy::type_complexity)]
     fn get_auth_header(
         &self,
         _auth_type: &types::ConnectorAuthType,
-    ) -> CustomResult<Vec<(String, String)>, errors::ConnectorError> {
+    ) -> CustomResult<Vec<(String, masking::Secret<String, masking::ApiKey>)>, errors::ConnectorError>
```
SanchithHegde, line 46: I believe using type aliases would eliminate the need for this `allow` attribute.

crates/router/src/types/api.rs
```diff
@@ -77,11 +79,15 @@ pub trait ConnectorCommonExt<Flow, Req, Resp>:
...
+    #[allow(clippy::type_complexity)]
     fn build_headers(
         &self,
         _req: &types::RouterData<Flow, Req, Resp>,
         _connectors: &Connectors,
-    ) -> CustomResult<Vec<(String, String)>, errors::ConnectorError> {
+    ) -> CustomResult<
+        Vec<(String, Box<dyn crate::services::request::HeaderValue>)>,
+        errors::ConnectorError,
+    > {
```
SanchithHegde, line 90: Same here.

SanchithHegde (commented):

crates/masking/src/secret.rs
NishantJoshi00: Yes, for the purpose of logging and formatting, the `Display` and `Debug` implementation are used accordingly. In this case we are performing the same action as `peek`. But, having `ToString` on a secret isn't optimal in every case, thus this implementation is only limited to `ApiKey` masking strategy

crates/router/src/connector/aci.rs
NishantJoshi00: In this case, the `header` variable is already of type `Vec<(String, Box<dyn HeaderValue>)>` thus, directly extending it with an iterator will have the same effect as the function `concat_headers`. The function concat headers is for a case where both of the headers provided to the function are of strict types and we want to append they as well as convert them to the able mentioned type

crates/router/src/connector/dlocal.rs
NishantJoshi00: If we try to directly return the header, the transformation between, `impl HeaderValue -> Box<dyn HeaderValue>` isn't apparent, and the compiler cannot abstract it. For this to happen we need have a `.map` to perform `Box::new`. I'll create a new function for this. Thanks!

crates/masking/src/secret.rs
jarnura: Already we having `peek`. Why taking a different direction for `Apikey`?

jarnura (commented):
This Pull Request is not looking like a `POC`.

`POC` should have problem statement and possible solution discussion with pros and cons.

A mini-model of the problem and the suggested solution.

This feels like implementation and not a poc

connector-template/mod.rs
```diff
@@ -168,7 +168,7 @@ impl
         &self,
         req: &types::PaymentsSyncRouterData,
         connectors: &settings::Connectors,
-    ) -> CustomResult<Vec<(String, String)>, errors::ConnectorError> {
+    ) -> CustomResult<Vec<\(String, services::request::header::Value\)>, errors::ConnectorError> {
```
jarnura, line 171: why we need escape sequence here? If needed so, why at line:46 is not needed?

crates/router/src/types/api.rs
```diff
@@ -39,10 +39,12 @@ pub trait ConnectorCommon {
     fn id(&self) -> &'static str;
 
     /// HTTP header used for authorization.
+    #[allow(clippy::type_complexity)]
```
jarnura, line 42: Don't allow clippy warnings, if the warning can't be resolve or takes much Dev-effort then we can allow, this didn't look like one to allow.

crates/router/src/types/api.rs
```diff
@@ -81,7 +83,8 @@ pub trait ConnectorCommonExt<Flow, Req, Resp>:
         &self,
         _req: &types::RouterData<Flow, Req, Resp>,
         _connectors: &Connectors,
-    ) -> CustomResult<Vec<(String, String)>, errors::ConnectorError> {
+    ) -> CustomResult<Vec<(String, crate::services::request::header::Value)>, errors::ConnectorError>
```
jarnura, line 86: Remove the absolute path in imports.

crates/router/src/services/api/request.rs
```diff
@@ -110,12 +168,15 @@ impl RequestBuilder {
...
     pub fn header(mut self, header: &str, value: &str) -> Self {
-        self.headers.insert((header.into(), value.into()));
+        self.headers
+            .insert((header.into(), Box::new(value.to_string())));
         self
     }
 
-    pub fn headers(mut self, headers: Vec<(String, String)>) -> Self {
-        let mut h = headers.into_iter().map(|(h, v)| (h, v));
+    pub fn headers(mut self, headers: Vec<(String, Box<dyn HeaderValue>)>) -> Self {
```
jarnura, line 176: Why dynamic dispatch is required here, anyways this is a string or library's `HeaderValue`.  I don't see a `polymorphism` which needed to be solved in dynamic dispatch way.

crates/router/src/services/api/request.rs
```diff
@@ -10,7 +15,60 @@ use crate::{
     logger,
 };
 
-pub(crate) type Headers = collections::HashSet<(String, String)>;
+pub(crate) type Headers = collections::HashSet<(String, Box<dyn HeaderValue>)>;
+
+pub trait HeaderValue: ToString + Debug + erased_serde::Serialize + Send + Sync {}
```
jarnura, line 20: What is the use of this trait?

jarnura (changes requested):

connector-template/mod.rs
NishantJoshi00: I'll remove it

jarnura:
> 

@jagan-jaya  requirement is a valid one, here we need to make `Secrets` to the `headerValue`. And `HeaderValue` is a limited set of possible values only, we can go will simpler approach with `enum` and possible values as a variant's right?

NishantJoshi00:
Closed in favor of #1320

PR #704, opened 2023-03-02
refactor: Basilisk hs integration

manoj-juspay:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [x] Refactoring
- [ ] Dependency updates

## Description
Legacy locker will be depreciated as it doesn't have the key rotation feature,
We will be using basilisk-hs card vault to store card and other payment method

### Additional Changes

- [ ] This PR modifies the database schema
- [x] This PR modifies application configuration/environment variables

## How did you test it?
Tested it locally with basiliks-hs setup , we need to test basilisk-v3 as this is the services effected with the changes
Lets do it in custom pods post review

## Checklist

- [x] I formatted the code `cargo +nightly fmt --all`
- [x] I addressed lints thrown by `cargo clippy`
- [x] I reviewed submitted code
- [ ] I added unit tests for my changes where possible
- [ ] I added a [CHANGELOG](/CHANGELOG.md) entry if applicable

PR #718, opened 2023-03-07
feat: Integrate basilisk-hs for card storage

NishantJoshi00:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring
- [ ] Dependency updates

## Description

This PR is an extension for #704 . Earlier @manoj-juspay added the support for `basilisk_hs` permanent locker. But, in doing so removed the support of legacy locker. The current situation of the testing environment will still be using the legacy locker or some time, and for the mean while, we also need to support both lockers. Once, the testing environment also receives the required infra changes we can completely remove the deprecated locker code, while is being added in this PR.

### Additional Changes

This change adds a new field under the configuration `[locker]`
```toml
[locker]
locker_setup = "legacy_locker" // "basilisk_locker"
```

## How did you test it?

Testing is required on the sandbox environment to observer the full effect of this change

## Checklist


NishantJoshi00:
This force push was required, the history of the current branch and the base branch differed greatly, but as the base branch has been up-to-date with main, the latest commits were no longer required

crates/router/src/core/customers.rs
```diff
@@ -141,7 +141,7 @@ pub async fn delete_customer(
         Ok(customer_payment_methods) => {
             for pm in customer_payment_methods.into_iter() {
                 if pm.payment_method == enums::PaymentMethod::Card {
-                    cards::delete_card_from_hs_locker(
+                    cards::delete_card_wrapper(
```
jarnura, line 144: change the function name, this is not a wrapper function, `delete_card_from_locker` locker can be `legacy` or `basilisk`?

crates/router/src/core/payment_methods/transformers.rs
```diff
@@ -74,6 +74,58 @@ pub struct DeleteCardResp {
...
+}
+
+#[derive(Debug, Deserialize, Serialize)]
+#[serde(rename_all = "camelCase")]
+pub struct AddCardResponse {
+    pub card_id: String,
+    pub external_id: String,
+    pub card_fingerprint: Secret<String>,
+    pub card_global_fingerprint: Secret<String>,
+    #[serde(rename = "merchant_id")]
```
jarnura, line 97: why `merchant_id` is renamed with `merchant_id`?

crates/router/src/core/payment_methods/transformers.rs
```diff
@@ -280,6 +332,72 @@ pub fn mk_add_card_response_hs(
...
+        customer_id: req.customer_id,
+        payment_method_id: response.card_id,
+        payment_method: req.payment_method,
+        payment_method_type: req.payment_method_type,
+        card: Some(card),
+        metadata: req.metadata,
+        created: Some(common_utils::date_time::now()),
+        recurring_enabled: false,           // [#256]
+        installment_payment_enabled: false, // #[#256]
+        payment_experience: Some(vec![api_models::enums::PaymentExperience::RedirectToUrl]), // [#256]
```
jarnura, line 364: What is the reason for hard-coding the `PaymentExperience` to `RedirectToUrl`?

jarnura: Add a comment in the type definition and use defaults instead. Or add a comment here itself.

crates/router/src/core/payment_methods/transformers.rs
```diff
@@ -375,6 +531,37 @@ pub async fn mk_delete_card_request_hs(
...
+        card_id,
+    };
+    let body = utils::Encode::<GetCard<'_>>::encode(&delete_card_req)
+        .change_context(errors::VaultError::RequestEncodingFailed)?;
+    let mut url = locker.host.to_owned();
+    url.push_str("/card/deleteCard");
+    let mut request = services::Request::new(services::Method::Post, &url);
+    request.add_header(headers::X_ROUTER, "test");
+    request.add_header(headers::CONTENT_TYPE, "application/x-www-form-urlencoded");
+    //request.add_content_type(Content::FORMURLENCODED);
```
jarnura, line 550: remove commented code

jarnura (changes requested):

crates/router/src/core/payment_methods/transformers.rs
NishantJoshi00: I'll remove it, it was initially present, I didn't notice it while adding the code back

jarnura (approved):
@NishantJoshi00 clarify with @manoj-juspay and comments for that. Otherwise looks good to me

crates/router/src/core/payment_methods/transformers.rs
manoj-juspay: make it `None` , this field is basically the experience of different type of transaction for a particular PM which will be a `vec[]` with PMT with most successful txns rate as the first entry
In future if we are asked to choose through which PMT we should do the txns , we can refer this

manoj-juspay (changes requested):

config/Development.toml
```diff
@@ -33,6 +33,7 @@ connection_timeout = 10
 host = ""
 mock_locker = true
 basilisk_host = ""
+locker_setup = "legacy_locker"
```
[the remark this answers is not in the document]
NishantJoshi00: Yes, but I kept it as a configuration because we are exploring the possibility of using basilisk-hs in sandbox as well in the future, having it as a config gives us more control in this case.

NishantJoshi00:
> https://github.com/juspay/hyperswitch/pull/718#discussion_r1133546509
Yes, for the current scenario, I have set it as `None`

manoj-juspay (approved):
LGTM

crates/router/src/core/payment_methods/transformers.rs
NishantJoshi00: Sure, for right now though, I have set it to `None`

PR #722, opened 2023-03-07
ci(release): specify `fetch-depth` for code checkout and use official Docker GitHub actions

SanchithHegde:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [x] Refactoring
- [ ] Dependency updates

## Description

This PR fixes an issue with release builds not having the correct values for `VERGEN_GIT_*` environment variables. The fix is to clone the complete git history instead of just the latest commit. Additionally, this also updates the release workflows to use the official Docker GitHub actions for building and pushing Docker images.

## Motivation and Context

This should help us better identify the exact builds running in sandbox and production environments.

## How did you test it?

Created a manual build, found no `VERGEN_*` related warnings.

## Checklist

### DIFF
--- a/.github/workflows/auto-release-tag.yml
+++ b/.github/workflows/auto-release-tag.yml
@@ -11,6 +11,8 @@ jobs:
 
     steps:
       - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
 
       - name: Build router Docker image
         run: docker build . --file Dockerfile --tag juspaydotin/orca:${{ github.ref_name }} --build-arg BINARY=router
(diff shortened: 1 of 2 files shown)


lsampras (approved):
Do you know why the entire history is needed?

Do we know the underlying git commands for generating vergen...

SanchithHegde:
@lsampras 

> Do you know why the entire history is needed?

The `actions/checkout` action otherwise clones only the latest ref and when I tried the same commands (from the build logs) on my local system, I received the same `vergen` related warnings about it using default values for `VERGEN_GIT_*` environment variables. Having the entire history only seemed to solve the problem, and I couldn't find any documented option to fetch history of only one branch either.

> Do we know the underlying git commands for vergen...

It uses the `git2` crate under the hood to obtain this information, it does not require the `git` CLI to be installed on the machine where the app is built.

PR #724, opened 2023-03-09
enhance(storage_models): delete client secret when status is succeeded, failed, cancelled

Abhicodes-crypto:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring
- [ ] Dependency updates

## Description

Previously client secret was only deleted when payment was succeeded. 
Now the client secret gets deleted when status is succeeded/failed/cancelled

### Additional Changes

## How did you test it?

Manual
image
image

## Checklist


SanchithHegde (approved):

crates/storage_models/src/payment_intent.rs
```diff
@@ -231,20 +232,25 @@ impl From<PaymentIntentUpdate> for PaymentIntentUpdateInternal {
...
 ) -> Option<Option<String>> {
-    if status == Some(storage_enums::IntentStatus::Succeeded) {
-        Some(None)
-    } else {
-        None
-    }
+    status.and_then(|status| match status {
+        storage_enums::IntentStatus::Succeeded
+        | storage_enums::IntentStatus::Failed
+        | storage_enums::IntentStatus::Cancelled => Some(None),
```
jarnura, line 249: should be make this to bool instead of `Some(None)` and `None`?

jarnura (changes requested):

crates/storage_models/src/payment_intent.rs
Abhicodes-crypto: `Some(None)` deletes the value in the database and `None` leaves the value unedited. 

If we make return type bool we have to still write a wrapper function to convert it to Some(None) or None. 

Since `Some(None)` and `None` is specific to diesel we could have a generic wrapper function in generics file I guess.

Should I do it in this PR or take it up in a separate PR ?

crates/storage_models/src/payment_intent.rs
```diff
@@ -231,20 +232,25 @@ impl From<PaymentIntentUpdate> for PaymentIntentUpdateInternal {
...
+        storage_enums::IntentStatus::Processing
+        | storage_enums::IntentStatus::RequiresCustomerAction
+        | storage_enums::IntentStatus::RequiresPaymentMethod
+        | storage_enums::IntentStatus::RequiresConfirmation
+        | storage_enums::IntentStatus::RequiresCapture => None,
```
Narayanbhat166, line 254: could have used `_` pattern here

Abhicodes-crypto: Didn't use `_` pattern because if a new enum variant is added in future they won't be forced by compiler to handle the case here.

jarnura (approved):

Narayanbhat166 (approved):

PR #725, opened 2023-03-09
feat: Add HSTS headers to response

dracarys18:
## Type of Change

- [x] Enhancement

## Description

Added HSTS header for responses.

## Motivation and Context

This header informs browsers that the site should only be accessed using HTTPS.
For reference: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Strict-Transport-Security

## How did you test it?

image

## Checklist

### DIFF
--- a/crates/router/src/core/errors.rs
+++ b/crates/router/src/core/errors.rs
@@ -16,7 +16,7 @@ pub use self::{
     api_error_response::ApiErrorResponse,
     utils::{ConnectorErrorExt, StorageErrorExt},
 };
-use crate::services;
+use crate::{headers, services};
 pub type RouterResult<T> = CustomResult<T, ApiErrorResponse>;
 pub type RouterResponse<T> = CustomResult<services::ApplicationResponse<T>, ApiErrorResponse>;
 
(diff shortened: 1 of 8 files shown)


inventvenkat (dismissed):

crates/router/src/core/errors.rs
```diff
@@ -154,7 +154,8 @@ impl From<ConfigError> for ApplicationError {
 
 fn error_response<T: Display>(err: &T) -> actix_web::HttpResponse {
     actix_web::HttpResponse::BadRequest()
-        .append_header(("Via", "Juspay_Router"))
+        .append_header((headers::HSTS, 31536000))
```
SanchithHegde, line 157: The numerical value could have been a constant.

I'd prefer if it was explicit as to how the number was obtained:

```rust
const HSTS_MAX_AGE: u32 /* or u64 */ = 60 * 60 * 24 * 365; // 1 year
```

Also, shouldn't the header value be `max-age=31536000` and not just the number?

SanchithHegde (approved):

jarnura (approved):

PR #733, opened 2023-03-12
refactor(drainer, router): KMS decrypt database password when `kms` feature is enabled

SanchithHegde:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [x] Refactoring
- [ ] Dependency updates

## Description

This PR enables the use of KMS encrypted database passwords for the `router` and `drainer` crates when the `kms` feature is enabled.

### Additional Changes

I prefer being explicit that a variable expects a KMS-encryted value, so I use `kms_encrypted_password` and KMS decrypt it to obtain the password when the `kms` feature is enabled, and use the `password` config value directly whenever the `kms` feature is disabled. Let me know if it would be preferable to use the `password` config value in both scenarios.

When `kms` feature is enabled, the `ROUTER__MASTER_DATABASE__PASSWORD`, `ROUTER__REPLICA_DATABASE__PASSWORD` and `DRAINER__MASTER_DATABASE__PASSWORD` environment variables should now be replaced with `ROUTER__MASTER_DATABASE__KMS_ENCRYPTED_PASSWORD`, `ROUTER__REPLICA_DATABASE__KMS_ENCRYPTED_PASSWORD` and `DRAINER__MASTER_DATABASE__KMS_ENCRYPTED_PASSWORD` instead, while ensuring that their values are replaced with their KMS encrypted equivalents.

KMS encrypted database passwords can be obtained like so:

```shell
$ echo -n 'MyVerySecretDatabasePassword' | base64

# Use the obtained ciphertext as the value for the config variable
$ aws kms encrypt --output json --key-id my-aws-key-id --plaintext <BASE64_ENCODED_DATABASE_PASSWORD>
```

## Motivation and Context

This should add another layer of security for the database passwords.

## How did you test it?

Manually.

## Checklist

### DIFF
--- a/config/config.example.toml
+++ b/config/config.example.toml
@@ -19,23 +19,25 @@ request_body_limit = 16_384
 
 # Main SQL data store credentials
 [master_database]
-username = "db_user"      # DB Username
-password = "db_pass"      # DB Password
-host = "localhost"        # DB Host
-port = 5432               # DB Port
-dbname = "hyperswitch_db" # Name of Database
-pool_size = 5             # Number of connections to keep open
-connection_timeout = 10   # Timeout for database connection in seconds
+username = "db_user"        # DB Username
+password = "db_pass"        # DB Password. Only applicable when KMS is disabled.
+host = "localhost"          # DB Host
+port = 5432                 # DB Port
+dbname = "hyperswitch_db"   # Name of Database
+pool_size = 5               # Number of connections to keep open
+connection_timeout = 10     # Timeout for database connection in seconds
+kms_encrypted_password = "" # Base64-encoded (KMS encrypted) ciphertext of the database password. Only applicable when KMS is enabled.
 
 # Replica SQL data store credentials
 [replica_database]
-username = "replica_user" # DB Username
-password = "replica_pass" # DB Password
-host = "localhost"        # DB Host
-port = 5432               # DB Port
-dbname = "hyperswitch_db" # Name of Database
-pool_size = 5             # Number of connections to keep open
-connection_timeout = 10   # Timeout for database connection in seconds
+username = "replica_user"   # DB Username
+password = "replica_pass"   # DB Password. Only applicable when KMS is disabled.
+host = "localhost"          # DB Host
+port = 5432                 # DB Port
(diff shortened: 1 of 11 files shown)


config/config.example.toml
```diff
@@ -20,22 +20,40 @@ request_body_limit = 16_384
...
-password = "db_pass"      # DB Password
+password = "db_pass"      # DB Password. Only applicable when KMS is disabled.
 host = "localhost"        # DB Host
 port = 5432               # DB Port
 dbname = "hyperswitch_db" # Name of Database
 pool_size = 5             # Number of connections to keep open
 connection_timeout = 10   # Timeout for database connection in seconds
+# Key ID for the KMS managed key used to decrypt the database password.
+# Only applicable when KMS is enabled.
+aws_key_id = ""
```
jarnura, line 31: move all `aws_key_id` and `aws_region` to single place under kms itself.

jarnura (changes requested):

config/config.example.toml
SanchithHegde: I've opened #744 for this change. Marking this PR as blocked until it gets merged.

SanchithHegde:
Rebased and force pushed. The drainer database password is also KMS decrypted at startup now.

ashokkjag (approved):

crates/drainer/src/connection.rs
```diff
@@ -15,13 +17,29 @@ pub async fn redis_connection(
...
+    let password = kms::get_kms_client(kms_config)
+        .await
+        .decrypt(&database.kms_encrypted_password)
+        .await
+        .expect("Failed to KMS decrypt database password");
+
+    #[cfg(not(feature = "kms"))]
+    let password = &database.password;
+
     let database_url = format!(
```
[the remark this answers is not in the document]
SanchithHegde: Unfortunately, the `diesel::r2d2::ConnectionManager` stores the connection string.

https://github.com/diesel-rs/diesel/blob/3d68571069e0f28582ae20e2120482203243bd55/diesel/src/r2d2.rs#L121

PR #735, opened 2023-03-13
enhance(core): replace string with enum for country

Narayanbhat166:
## Type of Change

- [x] Enhancement

## Description

Use enums for country instead of accepting arbitrary string.

## Motivation and Context

Accepting string for country does not have any type restriction.

## How did you test it?

Create a payment with billing country as `US` via stripe.

## Checklist

### DIFF
--- a/crates/api_models/src/enums.rs
+++ b/crates/api_models/src/enums.rs
@@ -248,6 +248,31 @@ pub enum Currency {
     ZAR,
 }
 
+#[allow(clippy::upper_case_acronyms)]
+#[derive(
+    Clone,
+    Debug,
+    Eq,
+    Default,
+    Hash,
+    PartialEq,
+    serde::Deserialize,
+    serde::Serialize,
+    strum::Display,
+    strum::EnumString,
+    frunk::LabelledGeneric,
+)]
+#[rustfmt::skip]
+pub enum Country {
+    AL,DZ,AS,AO,AG,AR,AU,AT,AZ,BH,BY,BE,BR,BG,CA,CL,CO,HR,CZ,DK,DO,EG,
+    EE,FI,FR,DE,GR,HK,HU,IN,ID,IE,IL,IT,JP,JO,KZ,KE,KW,LV,LB,LT,LU,MY,
+    MX,NL,NZ,NO,OM,PK,PA,PE,PH,PL,PT,QA,RO,RU,SA,SG,SK,ZA,ES,LK,SE,CH,
+    TW,TH,TR,UA,AE,GB,UY,VN,CN,MO,AM,CY,FO,GE,GL,GG,IS,IM,JE,LI,MT,
+    MD,MC,ME,SM,RS,SI,CR,PS,UM,
+    #[default]
+    US
+}
+
 #[derive(
     Clone,
     Copy,
--- a/crates/router/src/core/payments/flows/session_flow.rs
+++ b/crates/router/src/core/payments/flows/session_flow.rs
@@ -83,7 +83,7 @@ fn create_gpay_session_token(
 
     let session_data = router_data.request.clone();
     let transaction_info = payment_types::GpayTransactionInfo {
-        country_code: session_data.country.unwrap_or_else(|| "US".to_string()),
+        country_code: session_data.country.unwrap_or_default(),
         currency_code: router_data.request.currency.to_string(),
         total_price_status: "Final".to_string(),
         total_price: router_data.request.amount,
--- a/crates/storage_models/src/enums.rs
+++ b/crates/storage_models/src/enums.rs
@@ -256,6 +256,33 @@ pub enum Currency {
     ZAR,
 }
 
+#[allow(clippy::upper_case_acronyms)]
+#[derive(
+    Clone,
+    Copy,
+    Debug,
+    Eq,
+    Default,
+    Hash,
+    PartialEq,
+    serde::Deserialize,
+    serde::Serialize,
+    strum::Display,
+    strum::EnumString,
+    frunk::LabelledGeneric,
+)]
+#[router_derive::diesel_enum(storage_type = "text")]
+#[rustfmt::skip]
+pub enum Country {
+    AL,DZ,AS,AO,AG,AR,AU,AT,AZ,BH,BY,BE,BR,BG,CA,CL,CO,HR,CZ,DK,DO,EG,
+    EE,FI,FR,DE,GR,HK,HU,IN,ID,IE,IL,IT,JP,JO,KZ,KE,KW,LV,LB,LT,LU,MY,
+    MX,NL,NZ,NO,OM,PK,PA,PE,PH,PL,PT,QA,RO,RU,SA,SG,SK,ZA,ES,LK,SE,CH,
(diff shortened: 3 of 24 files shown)


crates/api_models/src/enums.rs
```diff
@@ -248,6 +248,121 @@ pub enum Currency {
...
+    Hash,
+    PartialEq,
+    serde::Deserialize,
+    serde::Serialize,
+    strum::Display,
+    strum::EnumString,
+    frunk::LabelledGeneric,
+    ToSchema,
+)]
+pub enum Country {
```
SanchithHegde, line 265: You should be able to derive `Copy` as well, as long as all variants are unit variants. This would eliminate the need to clone this enum.

SanchithHegde:
Could you verify if all existing connector tests pass after this change?

I was thinking more along the lines of using descriptive country names in our code, while serializing/deserializing using one of Alpha-2, Alpha-3 or the numeric code, as per requirement. This would keep things flexible for us should a connector require us to send either the Alpha-3 or numeric country code at a later point in time, while ensuring readability and ease of debugging.

I'll include a snippet of how this could be done, @vspecky @NishantJoshi00 @dracarys18 @lsampras please let me know if you can think this can be improved somehow.

```rust
#[derive(Clone, Copy, Debug)]
pub enum Country {
    Afghanistan,
    Albania,
    // ...
}

#[derive(Clone, Copy, Debug)]
pub enum Alpha2CountryCode {
    AF,
    AL,
    // ...
}

impl Country {
    // Note the `const`. Similar implementation for `to_alpha2(&self)`.
    pub const fn from_alpha2(code: Alpha2CountryCode) -> Self {
        match code {
            Alpha2CountryCode::AF => Self::Afghanistan,
            Alpha2CountryCode::AL => Self::Albania,
            // ...
        }
    }
}

mod custom_serde {
    use super::*;

    pub mod alpha2_country_code {
        use super::*;

        pub fn serialize<S>(code: &Country, serializer: S) -> Result<S::Ok, S::Error>
        where
            S: serde::Serializer,
        {
            // Serialize country as Alpha-2 code
            Err(serde::ser::Error::custom("not implemented"))
        }

        pub fn deserialize<'a, D>(deserializer: D) -> Result<Country, D::Error>
        where
            D: serde::Deserializer<'a>,
        {   
            // Deserialize Alpha-2 code from string, then convert to country enum
            Err(serde::de::Error::custom("not implemented"))
        }
    }
}

#[derive(serde::Deserialize, serde::Serialize)]
struct Address {
    #[serde(with = "custom_serde::alpha2_country_code")]
    country: Country,
}
```

crates/router/src/core/payments/flows/session_flow.rs
```diff
@@ -92,7 +92,7 @@ fn create_gpay_session_token(
 
     let session_data = router_data.request.clone();
     let transaction_info = payment_types::GpayTransactionInfo {
-        country_code: session_data.country.unwrap_or_else(|| "US".to_string()),
+        country_code: session_data.country.unwrap_or_default(),
```
jagan-jaya, line 95: why did we remove "US" from default country?

jagan-jaya (changes requested):

crates/router/src/core/payments/flows/session_flow.rs
Narayanbhat166: `US` is still the default country for `Country` enum. So unwrap or default returns US if option is None

vspecky:
> Could you verify if all existing connector tests pass after this change?
> 
> I was thinking more along the lines of using descriptive country names in our code, while serializing/deserializing using one of Alpha-2, Alpha-3 or the numeric code, as per requirement. This would keep things flexible for us should a connector require us to send either the Alpha-3 or numeric country code at a later point in time, while ensuring readability and ease of debugging.
> 
> I'll include a snippet of how this could be done, @vspecky @NishantJoshi00 @dracarys18 @lsampras please let me know if you can think this can be improved somehow.
> 
> ```rust
> #[derive(Clone, Copy, Debug)]
> pub enum Country {
>     Afghanistan,
>     Albania,
>     // ...
> }
> 
> #[derive(Clone, Copy, Debug)]
> pub enum Alpha2CountryCode {
>     AF,
>     AL,
>     // ...
> }
> 
> impl Country {
>     // Note the `const`. Similar implementation for `to_alpha2(&self)`.
>     pub const fn from_alpha2(code: Alpha2CountryCode) -> Self {
>         match code {
>             Alpha2CountryCode::AF => Self::Afghanistan,
>             Alpha2CountryCode::AL => Self::Albania,
>             // ...
>         }
>     }
> }
> 
> mod custom_serde {
>     use super::*;
> 
>     pub mod alpha2_country_code {
>         use super::*;
> 
>         pub fn serialize<S>(code: &Country, serializer: S) -> Result<S::Ok, S::Error>
>         where
>             S: serde::Serializer,
>         {
>             // Serialize country as Alpha-2 code
>             Err(serde::ser::Error::custom("not implemented"))
>         }
> 
>         pub fn deserialize<'a, D>(deserializer: D) -> Result<Country, D::Error>
>         where
>             D: serde::Deserializer<'a>,
>         {   
>             // Deserialize Alpha-2 code from string, then convert to country enum
>             Err(serde::de::Error::custom("not implemented"))
>         }
>     }
> }
> 
> #[derive(serde::Deserialize, serde::Serialize)]
> struct Address {
>     #[serde(with = "custom_serde::alpha2_country_code")]
>     country: Country,
> }
> ```

Overall this makes sense to me. While accepting inputs from APIs and passing data around, we will always use the `Country`  enum (Even if our API takes Alpha-2 or Alpha-3 as input, the custom deserialization will deserialize to `Country`). If a specific connector requires it in a different format, then in the types of the connector we can use the `Alpha2CC` and `Alpha3CC` types and convert within the code using the conversion logic example you've given.

crates/storage_models/src/enums.rs
```diff
@@ -256,6 +256,32 @@ pub enum Currency {
...
+    Default,
+    Hash,
+    PartialEq,
+    serde::Deserialize,
+    serde::Serialize,
+    strum::Display,
+    strum::EnumString,
+    frunk::LabelledGeneric,
+)]
+#[router_derive::diesel_enum(storage_type = "text")]
```
jarnura, line 273: We can add enum in the database since this won't grow in future, add migrations

jarnura (changes requested):

Narayanbhat166:
Linking the issue for supporting additional country code variants #804

Narayanbhat166:
The `frunkLabelledGeneric` derive macro is not able to generate the implementation for the `CountryCode` enum which has 249 variants. I'm thinking of creating a macro for this which will manually do the mapping at compile time. This macro can replace the `frunkLabelledGeneric` macro.

Narayanbhat166:
Better to have a common place for enums which will be same in both `api_models` and `storage_models`. Will create a `common_enum` crate. @vspecky @jarnura.

Narayanbhat166:
The dependency is bumped to the recent version of [utoipa](https://github.com/juhaku/utoipa). The bug fix is included in this release. Unblocking this now.

jarnura (dismissed):

jarnura (approved):

jagan-jaya (approved):

PR #736, opened 2023-03-13
feat: populate fields from payment attempt in payment list

dracarys18:
## Type of Change

- [x] Enhancement

## Description

Populate some fields from payment attempt table in Paymentlist response.

## Motivation and Context

Populate necessary fields which we show in the dashboard from payment attempt.

## How did you test it?

Manual

## Checklist

### DIFF
--- a/crates/router/src/core/payments.rs
+++ b/crates/router/src/core/payments.rs
@@ -631,7 +631,8 @@ pub async fn list_payments(
                 .find_payment_attempt_by_payment_id_merchant_id(
                     &pi.payment_id,
                     merchant_id,
-                    merchant.storage_scheme,
+                    // since OLAP doesn't have KV. Force to get the data from PSQL.
+                    storage_enums::MerchantStorageScheme::PostgresOnly,
                 )
                 .await
                 .ok()?;
--- a/crates/router/src/core/payments.rs
+++ b/crates/router/src/core/payments.rs
@@ -609,9 +609,13 @@ pub async fn list_payments(
     merchant: storage::MerchantAccount,
     constraints: api::PaymentListConstraints,
 ) -> RouterResponse<api::PaymentListResponse> {
+    use futures::stream::StreamExt;
+
+    use crate::types::transformers::ForeignFrom;
+
     helpers::validate_payment_list_request(&constraints)?;
     let merchant_id = &merchant.merchant_id;
-    let payment_intent =
+    let payment_intents =
         helpers::filter_by_constraints(db, &constraints, merchant_id, merchant.storage_scheme)
             .await
             .map_err(|err| {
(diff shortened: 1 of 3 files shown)


crates/router/src/core/payments.rs
```diff
@@ -621,10 +625,23 @@ pub async fn list_payments(
...
-        .into_iter()
-        .map(types::transformers::ForeignInto::foreign_into)
-        .collect();
+    let pi = futures::stream::iter(payment_intents)
+        .filter_map(|pi| async {
+            let pa = db
+                .find_payment_attempt_by_payment_id_merchant_id(
+                    &pi.payment_id,
+                    merchant_id,
+                    merchant.storage_scheme,
```
jarnura, line 634: Should we add a comment and say this is `olap` function call and the storage.schema is always `psql` only?

crates/router/src/core/payments/transformers.rs
```diff
@@ -393,6 +393,30 @@ where
...
+            client_secret: pi.client_secret.map(|s| s.into()),
+            created: Some(pi.created_at),
+            currency: pi.currency.map(|c| c.to_string()).unwrap_or_default(),
+            description: pi.description,
+            metadata: pi.metadata,
+            customer_id: pi.customer_id,
+            connector: pa.connector,
+            payment_method: pa.payment_method.map(ForeignInto::foreign_into),
+            payment_method_type: pa.payment_method_type.map(ForeignInto::foreign_into),
+            ..Default::default()
```
jarnura, line 415: we need to populate other fields also right? like `refunds`. Can you confirm with the dashboard team about this.

jarnura (changes requested):

crates/router/src/core/payments/transformers.rs
dracarys18: Confirmed with dashboard team they only need `payment_method`,`connector`,`payment_method_type` fields which is being populated here.

SanchithHegde (approved):

ashokkjag:
@jarnura / @dracarys18 , 

Are there limits on how many payment intents are retrieved? If payment list goes into 100s, then for each call, we'll do 100s of payment attempts lookup.

Does payment attempt always have only 1 entry for payment intent + merchant id?

crates/router/src/core/payments.rs
```diff
@@ -609,9 +609,13 @@ pub async fn list_payments(
     merchant: storage::MerchantAccount,
     constraints: api::PaymentListConstraints,
 ) -> RouterResponse<api::PaymentListResponse> {
+    use futures::stream::StreamExt;
+
+    use crate::types::transformers::ForeignFrom;
```
jarnura, line 614: move this import to top of the file

jarnura (changes requested):
Other than import changes looks good to be

crates/router/src/core/payments.rs
jarnura: Because of feature enabled we can't move to top of the file

jarnura (approved):

dracarys18:
> @jarnura / @dracarys18 ,
> 
> Are there limits on how many payment intents are retrieved? If payment list goes into 100s, then for each call, we'll do 100s of payment attempts lookup.
> 
> Does payment attempt always have only 1 entry for payment intent + merchant id?

The limit is 100.  Since the call happens to OLAP database I think these calls is feasible and not heavy to the database.

PR #737, opened 2023-03-13
feat: add generic in-memory cache interface

dracarys18:
## Type of Change

- [x] New feature

## Description

Added a struct for Cache which takes any type as value and string as key.

## Motivation and Context

This PR adds support to generic cache struct. Which accepts any type as value and while getting it back from cache downcasts back to concrete type.

## How did you test it?

Manual

## Checklist

### DIFF
--- a/crates/router/src/cache.rs
+++ b/crates/router/src/cache.rs
@@ -1,0 +1,65 @@
+use dyn_clone::DynClone;
+use moka::future::Cache as MokaCache;
+use std::any::Any;
+
+/// Trait which defines the behaviour of types that's gonna be stored in Cache
+pub trait Cacheable: Any + Send + Sync + DynClone {
+    fn as_any(&self) -> &dyn Any;
+}
+
+impl<T> Cacheable for T
+where
+    T: Any + Clone + Send + Sync,
+{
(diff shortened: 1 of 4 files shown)


jarnura (approved):

crates/router/src/cache.rs
```diff
@@ -0,0 +1,66 @@
...
+                .time_to_live(std::time::Duration::from_secs(time_to_live))
+                .time_to_idle(std::time::Duration::from_secs(time_to_idle))
+                .build(),
+        }
+    }
+
+    pub fn get_val<T: Clone + Any>(&self, key: &str) -> Option<T> {
+        let val = self.get(key)?;
+        (*val).as_any().downcast_ref::<T>().cloned()
+    }
```
vspecky, line 51: See if we can have per-key time to live and design an API around that (not a blocker).

vspecky (approved):
LGTM

SanchithHegde (approved):

crates/router/src/cache.rs
dracarys18: Checked with their docs no such feature available yet.

PR #743, opened 2023-03-14
refactor: get connection pool based on olap/oltp features

dracarys18:
## Type of Change

- [x] Refactoring

## Description

Get connection pool from store based on the features that are enabled.

## Motivation and Context

the motivation behind this is whenever let's say a function from olap feature calls any DB functions, by default it goes to master pool which is not expected. So decide the DB pool inside the `pg_connection` function based on the feature that's enabled.

## How did you test it?

Manual

## Checklist

### DIFF
--- a/crates/router/src/db/configs.rs
+++ b/crates/router/src/db/configs.rs
@@ -2,7 +2,7 @@ use error_stack::IntoReport;
 
 use super::{cache, MockDb, Store};
 use crate::{
-    connection::pg_connection_read,
+    connection,
     core::errors::{self, CustomResult},
     types::storage,
 };
@@ -45,7 +45,7 @@ impl ConfigInterface for Store {
         &self,
         config: storage::ConfigNew,
     ) -> CustomResult<storage::Config, errors::StorageError> {
-        let conn = pg_connection_read(self).await?;
+        let conn = connection::pg_connection_write(self).await?;
         config.insert(&conn).await.map_err(Into::into).into_report()
     }
 
@@ -53,7 +53,7 @@ impl ConfigInterface for Store {
         &self,
         key: &str,
     ) -> CustomResult<storage::Config, errors::StorageError> {
-        let conn = pg_connection_read(self).await?;
+        let conn = connection::pg_connection_read(self).await?;
         storage::Config::find_by_key(&conn, key)
             .await
             .map_err(Into::into)
@@ -65,7 +65,7 @@ impl ConfigInterface for Store {
         key: &str,
         config_update: storage::ConfigUpdate,
     ) -> CustomResult<storage::Config, errors::StorageError> {
-        let conn = pg_connection_read(self).await?;
+        let conn = connection::pg_connection_write(self).await?;
         storage::Config::update_by_key(&conn, key, config_update)
             .await
             .map_err(Into::into)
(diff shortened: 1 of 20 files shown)


jarnura (dismissed):

crates/router/src/db/customers.rs
```diff
@@ -75,7 +75,7 @@ impl CustomerInterface for Store {
         merchant_id: String,
         customer: storage::CustomerUpdate,
     ) -> CustomResult<storage::Customer, errors::StorageError> {
-        let conn = pg_connection(self).await?;
+        let conn = pg_connection_read(self).await?;
```
SanchithHegde, line 78: Writable connection?

crates/router/src/db/customers.rs
```diff
@@ -108,7 +108,7 @@ impl CustomerInterface for Store {
         &self,
         customer_data: storage::CustomerNew,
     ) -> CustomResult<storage::Customer, errors::StorageError> {
-        let conn = pg_connection(self).await?;
+        let conn = pg_connection_read(self).await?;
```
SanchithHegde, line 111: Writable connection?

crates/router/src/db/customers.rs
```diff
@@ -121,7 +121,7 @@ impl CustomerInterface for Store {
         customer_id: &str,
         merchant_id: &str,
     ) -> CustomResult<bool, errors::StorageError> {
-        let conn = pg_connection(self).await?;
+        let conn = pg_connection_read(self).await?;
```
SanchithHegde, line 124: Writable connection?

crates/router/src/db/payment_intent.rs
```diff
@@ -135,7 +135,7 @@ mod storage {
         ) -> CustomResult<PaymentIntent, errors::StorageError> {
             match storage_scheme {
                 enums::MerchantStorageScheme::PostgresOnly => {
-                    let conn = pg_connection(self).await?;
+                    let conn = connection::pg_connection_read(self).await?;
```
SanchithHegde, line 138: Writable connection?

crates/router/src/db/refund.rs
```diff
@@ -182,7 +182,7 @@ mod storage {
         //     merchant_id: &str,
         //     refund_id: &str,
         // ) -> CustomResult<Refund, errors::StorageError> {
-        //     let conn = pg_connection(self).await;
+        //     let conn = pg_connection_read(self).await;
```
SanchithHegde, line 185: If you don't mind, please remove all commented out definitions of the `find_refund_by_payment_id_merchant_id_refund_id()` method.

crates/router/src/db/configs.rs
```diff
@@ -45,15 +45,15 @@ impl ConfigInterface for Store {
...
+        let conn = connection::pg_connection_write(self).await?;
         config.insert(&conn).await.map_err(Into::into).into_report()
     }
 
     async fn find_config_by_key(
         &self,
         key: &str,
     ) -> CustomResult<storage::Config, errors::StorageError> {
-        let conn = pg_connection_read(self).await?;
+        let conn = connection::pg_connection_read(self).await?;
```
jarnura, line 56: Config read also use write itself

jarnura (changes requested):

crates/router/src/db/configs.rs
dracarys18: Why?

jarnura: In dashboard, will updating a config value for feature flags we need to check it correctly updated or not, we need to read from master and not replica

jarnura (changes requested):

crates/router/src/db/configs.rs
dracarys18: Fixed

SanchithHegde (approved):

jarnura (approved):

Issue #379, opened 2023-01-15
[FEATURE] Determine card brand from card BIN number


ArjunKarthik:
### Feature Description

Some connectors like Worldline require a card brand name to be provided when providing payment method details. As of now, we have a limited set of regular expressions to identify card brands from card BIN number:

https://github.com/juspay/hyperswitch/blob/d107b44fd356f172ef7c9535567d266a6301e43d/crates/router/src/connector/worldline/transformers.rs#L20-L32

We'd need a more comprehensive way of identifying card brands. It would be great if the card identification logic is included in a new crate called `cards` which would contain this card brand identification logic and validation of card numbers (taken up on #606). It would also be great if the implementation minimizes CPU usage, so please include benchmarking code and times observed on your machine. Please also include unit tests in your PR.

_For external contributors, clarify any questions you may have, and let us know that you're interested in picking this up. We'll assign this issue to you to reduce duplication of effort on the same task._

### Possible Implementation

There are 2 ways to handle this: either using regular expressions, or maintaining a mapping of card BINs to brand names. You can decide to go with either of them, keeping the requirements in mind.

### Have you spent some time to check if this feature request has been raised before?

- [X] I checked and didn't find similar issue

### Have you read the Contributing Guidelines?

- [X] I have read the [Contributing Guidelines](https://github.com/juspay/hyperswitch/blob/main/contrib/CONTRIBUTING.md)

### Are you willing to submit a PR?

None

jeevaprakashdr:
@SanchithHegde is this feature needs to be added as part of  `cards info api` or it is part of new crate `cards` from #607 ?

SanchithHegde:
Closing as completed in #749.

PR #749, opened 2023-03-15
feat: cards info api

Narayanbhat166:
## Type of Change

- [x] New feature

## Description

This is an open endpoint which allows to fetch card information given the first 6 digits ( also known as card bin ). 

### Additional Changes

Add a new table called [cards_info](https://github.com/juspay/hyperswitch/blob/1f53c9d3cd122e86ad00d909e1791cbbec0d2f04/migrations/2023-03-14-123541_add_cards_info_table/up.sql)

## Motivation and Context

This will allow the sdk to show the card icon and pass 3ds information.

## How did you test it?

Screenshot 2023-03-21 at 4 30 35 PM

Screenshot 2023-03-16 at 3 11 02 PM

## Checklist


jarnura (dismissed):

SanchithHegde:
@Narayanbhat166 Would this close #379?

crates/router/src/db/cards_info.rs
```diff
@@ -0,0 +1,41 @@
...
+        &self,
+        card_iin: &str,
+    ) -> CustomResult<Option<CardInfo>, errors::StorageError> {
+        let conn = pg_connection(&self.master_pool).await?;
+        CardInfo::find_by_iin(&conn, card_iin)
+            .await
+            .map_err(Into::into)
+            .into_report()
+    }
+}
```
SanchithHegde, line 31: Please keep names consistent. If it is IIN that you're querying with, then call the parameter as `card_iin` instead of `card_bin` in the trait definition and all implementations.

crates/storage_models/src/cards_info.rs
```diff
@@ -0,0 +1,18 @@
+use diesel::{Identifiable, Queryable};
+
+use crate::schema::cards_info;
+
+#[derive(Clone, Debug, Queryable, Identifiable, serde::Deserialize, serde::Serialize)]
+#[diesel(table_name = cards_info)]
+#[diesel(primary_key(card_iin))]
```
SanchithHegde, line 7: You should be able to merge these `diesel` attributes (IIRC).

migrations/2023-03-14-123541_add_cards_info_table/up.sql
```diff
@@ -0,0 +1,15 @@
...
+    card_network TEXT,
+    card_type TEXT,
+    card_subtype TEXT,
+    card_issuing_country TEXT,
+    bank_code_id VARCHAR(32),
+    bank_code VARCHAR(32),
+    country_code VARCHAR(32),
+    date_created TIMESTAMP NOT NULL,
+    last_updated TIMESTAMP,
+    last_updated_provider TEXT
```
SanchithHegde, line 14: Are these 3 columns being used anywhere?

crates/router/src/core/cards_info.rs
```diff
@@ -0,0 +1,26 @@
...
+    db::StorageInterface,
+    services::ApplicationResponse,
+    types::transformers::ForeignFrom,
+};
+
+#[instrument(skip_all)]
+pub async fn retrieve_card_info(
+    store: &dyn StorageInterface,
+    card_iin: String,
+) -> RouterResponse<api_models::cards_info::CardInfoResponse> {
```
SanchithHegde, line 15: Shouldn't we validate that the IIN/BIN is a min. of 6 and a max. of 9 characters only?

SanchithHegde (changes requested):
Also, how would the inserts into database happen? I couldn't seem to find any code to update the cards information table.

migrations/2023-03-14-123541_add_cards_info_table/up.sql
Narayanbhat166: These are used internally, to keep track of when the card bin was created, updated and the provider which updated this card bin

Narayanbhat166:
> @Narayanbhat166 Would this close #379?

No. That is a different use case ( during payments ). This endpoint can be called by the sdk ( or anyone using the api ) to know card information prior to confirming the payment and then pass the same card information like card_network, issuer. The caller has to ensure that card network is always passed in the confirm payments by calling this cards info endpoint

crates/router/src/core/errors/api_error_response.rs
```diff
@@ -156,7 +156,7 @@ pub enum ApiErrorResponse {
     #[error(error_type = ErrorType::ObjectNotFound, code = "HE_04", message = "Address does not exist in our records")]
     AddressNotFound,
     #[error(error_type = ErrorType::ObjectNotFound, code = "HE_04", message = "Card with the provided bin does not exist")]
-    CardBinNotFound,
+    InvalidCardIIN,
```
SanchithHegde, line 159:
```suggestion
    InvalidCardIin,
```

Let's keep this consistent with the Rust API guidelines.

From https://rust-lang.github.io/api-guidelines/naming.html#casing-conforms-to-rfc-430-c-case:

> In `UpperCamelCase`, acronyms and contractions of compound words count as one word: use `Uuid` rather than `UUID`, `Usize` rather than `USize` or `Stdin` rather than `StdIn`.

crates/router/src/core/cards_info.rs
```diff
@@ -0,0 +1,31 @@
...
+    when(card_iin.len() != 6, || {
+        Err(errors::ApiErrorResponse::InvalidCardIinLength)
+    })?;
```
SanchithHegde, line 19: As far as I understand, length should be either 6 or 8 digits. Not sure if some cases need 9 chars as well.
So we can check if the length is in that range.

You'd have to update the error message as well, to convey the same.

manojradhakrishnan:
@Narayanbhat166 This API seems to have been built as unauthenticated API. Please enable authentication via client_secret and api_key

SanchithHegde (dismissed):

jarnura (approved):

SanchithHegde (dismissed):

crates/router/src/core/errors/api_error_response.rs
```diff
@@ -156,6 +156,10 @@ pub enum ApiErrorResponse {
     IncorrectConnectorNameGiven,
     #[error(error_type = ErrorType::ObjectNotFound, code = "HE_04", message = "Address does not exist in our records")]
     AddressNotFound,
+    #[error(error_type = ErrorType::InvalidRequestError, code = "HE_04", message = "Card with the provided iin does not exist")]
+    InvalidCardIin,
+    #[error(error_type = ErrorType::InvalidRequestError, code = "HE_04", message = "The provoded card iin length is invalid, please provide an iin with 6 or 8 digits")]
```
jagan-jaya, line 161:
```suggestion
    #[error(error_type = ErrorType::InvalidRequestError, code = "HE_04", message = "The provided card IIN length is invalid, please provide an iin with 6 or 8 digits")]
```

crates/router/src/core/errors/api_error_response.rs
```diff
@@ -429,9 +434,11 @@ impl common_utils::errors::ErrorSwitch<api_models::errors::types::ApiErrorRespon
             }
             Self::NotSupported { message } => {
                 AER::BadRequest(ApiError::new("HE", 3, "Payment method type not supported", Some(Extra {reason: Some(message.to_owned()), ..Default::default()})))
-            }
+            },
+            Self::InvalidCardIin => AER::BadRequest(ApiError::new("HE", 3, "The provided card IIN does not exist", None)),
+            Self::InvalidCardIinLength  => AER::BadRequest(ApiError::new("HE", 3, "The provoded card iin length is invalid, please provide an iin with 6 digits", None)),
```
jagan-jaya, line 439:
```suggestion
            Self::InvalidCardIinLength  => AER::BadRequest(ApiError::new("HE", 3, "The provoded card IIN length is invalid, please provide an IIN with 6 digits", None)),
```

jagan-jaya (changes requested):

SanchithHegde (dismissed):

crates/router/src/core/errors/api_error_response.rs
```diff
@@ -429,9 +434,11 @@ impl common_utils::errors::ErrorSwitch<api_models::errors::types::ApiErrorRespon
             }
             Self::NotSupported { message } => {
                 AER::BadRequest(ApiError::new("HE", 3, "Payment method type not supported", Some(Extra {reason: Some(message.to_owned()), ..Default::default()})))
-            }
+            },
+            Self::InvalidCardIin => AER::BadRequest(ApiError::new("HE", 3, "The provided card IIN does not exist", None)),
+            Self::InvalidCardIinLength  => AER::BadRequest(ApiError::new("HE", 3, "The provoded card IIN length is invalid, please provide an IIN with 6 digits", None)),
```
dracarys18, line 439:
```suggestion
            Self::InvalidCardIinLength  => AER::BadRequest(ApiError::new("HE", 3, "The provided card IIN length is invalid, please provide an IIN with 6 digits", None)),
```

SanchithHegde (approved):

jarnura (approved):

jagan-jaya (approved):

PR #750, opened 2023-03-15
feat: connector tokenization flow

Sangamesh26:
## Type of Change

- [x] New feature

## Description

Some connectors requires token for the authorize call, so tokenization flow is implemented.In general, we get token and we use this for the next flows.

**When we don't have the token:**
Based on the feature check, the connector is decided whether to call tokenization flow or not.If tokenization is required, we call the tokenization service and the token is stored in payment_data, and this token is used in the next payment flows.
In the payment request, if the store_connector_token is true, we store the token in our db as it is long lived, if the store_connectore_token is false, we don't store it as it is short lived.
The  token is stored in payment_method table in metadata, one thing that is considered is , if the same payment_method goes to another connector for tokenization , we get different token.So, in payment_method db table, in metadata, we store token like (example): [{"connector":"stripe","token":"sometoken"},{..},{..}].

**What happens in customer_payment_method api call:**
We get all the payment methods of a particular customer, then if there is a metadata with the token information,then the token information is temporarily stored in redis like **{connector_name}_token_hyperswitch_{hyperswitch_)token}**.

**Payment create call with token:**
When we have the token, we will check whether connector token is stored against this hyperswitch token, if there is a token, then there is no need of tokenization flow again, we directly use this token in next authorize flows.If the tokenization_check fails, we get the payment_method from the hyperswitch token and then call for tokenization flow to get the token and to proceed with that token for next flows.

**Test**
This tokenization flow is tested for stripe through applepay, the call for tokenization is : https://api.stripe.com/v1/tokens and after getting the token, we make the payment call with data : payment_method_data[card][token]=token_xxx and type as card for applepay.

### Additional Changes

## Motivation and Context

Some of the connectors require token for the authorize flow.So, the tokenization flow helps to get token from the connector and pass this token for the upcoming payment flows for the smooth payment experience.

## How did you test it?

Manual testing (tested taking applepay through stripe as it require tokenization and this token is short lived)

## Checklist


vspecky (changes requested):

crates/router/src/core/payments.rs
```diff
@@ -109,19 +102,70 @@ where
...
+                )
+                .await?;
+
+                payment_data = pdc::<_, F>(token_payment_data.to_owned());
+                payment_data.token = token_payment_data.payment_attempt.payment_token.to_owned();
+                payment_data.payment_attempt.connector = Some(connector_name);
+            }
+        };
+    };
+
```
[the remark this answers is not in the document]
vspecky: Also can we put all the tokenization code into a separate function? To keep the core clean.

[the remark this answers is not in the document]
Sangamesh26: done

crates/router/src/connector/aci.rs
```diff
@@ -51,6 +51,14 @@ impl api::PaymentVoid for Aci {}
 impl api::PaymentCapture for Aci {}
 impl api::PaymentSession for Aci {}
 impl api::ConnectorAccessToken for Aci {}
+impl api::PaymentToken for Aci {}
+
+impl
```
Narayanbhat166, line 56: Add this implementation in the default_impl macro in flow.rs file
https://github.com/juspay/hyperswitch/blob/main/crates/router/src/core/payments/flows.rs#L116

crates/router/src/connector/stripe.rs
```diff
@@ -84,6 +84,111 @@ impl
...
+        connectors: &settings::Connectors,
+    ) -> CustomResult<String, errors::ConnectorError> {
+        Ok(format!("{}{}", self.base_url(connectors), "v1/tokens"))
+    }
+
+    fn get_request_body(
+        &self,
+        req: &types::TokenizationRouterData,
+    ) -> CustomResult<Option<String>, errors::ConnectorError> {
+        let stripe_req = utils::Encode::<stripe::TokenRequest>::convert_and_url_encode(req)
```
Narayanbhat166, line 126: first do the try_from conversion and then encode to string. This will help to catch missing field errors

Narayanbhat166 (changes requested):

crates/router/src/configs/settings.rs
```diff
@@ -60,6 +60,32 @@ pub struct Settings {
...
+pub struct TokenizationConfig(pub HashMap<String, PaymentMethodTokenFilter>);
+
+#[derive(Debug, Deserialize, Clone, Default)]
+pub struct PaymentMethodTokenFilter {
+    #[serde(deserialize_with = "pm_deser")]
+    pub payment_method: HashSet<storage_models::enums::PaymentMethod>,
+    pub long_lived_token: bool,
+}
+
+fn pm_deser<'a, D>(
```
jarnura, line 77: - Add doc test
- Input and Output data for custom `deserializer`.

crates/router/src/core/payment_methods/cards.rs
```diff
@@ -45,7 +51,26 @@ pub async fn create_payment_method(
     customer_id: &str,
     payment_method_id: &str,
     merchant_id: &str,
+    connector: Option<&api::ConnectorData>,
```
jarnura, line 54: whenever token is there then connector also should be present right, create a tuple or struct and use the struct/tuple as `Option`

crates/router/src/core/payment_methods/cards.rs
```diff
@@ -45,7 +51,26 @@ pub async fn create_payment_method(
...
+    token: Option<String>,
 ) -> errors::CustomResult<storage::PaymentMethod, errors::StorageError> {
+    let pm_metadata = token
+        .map(|token| -> errors::CustomResult<_, errors::StorageError> {
+            let metadata = payment_methods::PaymentMethodMetadata {
+                connector: connector
```
jarnura, line 60: This is a plain structure and can be accommodate only one connector and token right?

jarnura (changes requested):

crates/router/src/connector/stripe/transformers.rs
```diff
@@ -619,6 +678,20 @@ impl TryFrom<&types::PaymentsAuthorizeRouterData> for PaymentIntentRequest {
             }
         };
 
+        payment_data = match item.request.payment_method_data.to_owned() {
```
dracarys18, line 681: No need of `to_owned` here.

crates/router/src/core/payment_methods/cards.rs
```diff
@@ -276,15 +335,87 @@ pub async fn add_card_hs(
...
+                let mut vec_value: Vec<Value> = Vec::new();
+                if let Some(pm_metadata) = pm.metadata.to_owned() {
+                    let mut pm_metadata_value = pm_metadata.expose();
+                    let vec_metadata_value = pm_metadata_value.as_array_mut().map(|array| {
+                        array.push(metadata_value.to_owned());
+                        array
+                    });
+
+                    vec_value = match vec_metadata_value {
+                        Some(val) => (*val.clone()).to_vec(),
```
dracarys18, line 368: This can be replaced with `val.clone()`

crates/router/src/core/payment_methods/cards.rs
```diff
@@ -340,26 +473,73 @@ pub async fn add_card(
...
```
dracarys18, line 503: This can be replaced with `val.clone()`

crates/router/src/core/payment_methods/cards.rs
```diff
@@ -340,26 +473,73 @@ pub async fn add_card(
...
+                        None => {
+                            vec_value.push(metadata_value);
+                            vec_value
+                        }
+                    };
```
dracarys18, line 508: This looks repeated can be extracted into a function?

crates/router/src/core/payments.rs
```diff
@@ -552,7 +558,144 @@ where
...
+                &state.conf.connectors,
+                &connector_name,
+                api::GetToken::Connector,
+            )
+            .change_context(errors::ApiErrorResponse::InternalServerError)
+            .attach_printable("Failed to get connector data")?;
+
+            if tokenization_pm_and_token_store_check.0 {
+                payment_data.store_connector_token = Some(tokenization_pm_and_token_store_check.1);
+                let token_pd = pdc::<_, api::Token>(payment_data);
```
dracarys18, line 621: What's `pdc`? Can we have a better name here

dracarys18 (changes requested):

crates/router/src/core/payment_methods/cards.rs
```diff
@@ -276,15 +335,87 @@ pub async fn add_card_hs(
...
+
+                let _updated_pm = db
+                    .update_payment_method(pm, pm_update)
+                    .await
+                    .change_context(errors::VaultError::UpdateInPMDTableFailed)?;
+            };
+        }
+        Err(err) => {
+            if err.current_context().is_db_not_found() {
+                create_payment_method(
```
manoj-juspay, line 390: why are we calling `create_payment_method` twice ?

manoj-juspay (changes requested):

crates/router/src/core/payment_methods/cards.rs
Sangamesh26: I have passed connector and token in the next functions, token is not passed  as it is not necessary at this place.

crates/router/src/core/payment_methods/cards.rs
Sangamesh26: While deserializing, i have deserialized to the vector of this struct, so that I will get the token for that particular connector easily.So plain structure would suffice here.

jarnura (changes requested):

crates/router/src/core/payment_methods/cards.rs
```diff
@@ -73,40 +79,31 @@ pub async fn add_payment_method(
...
                 installment_payment_enabled: false, //[#219]
                 payment_experience: Some(vec![api_models::enums::PaymentExperience::RedirectToUrl]), //[#219]
-            })
+            };
+            Ok((payment_method_response, false))
         }
-    }
-    .map(services::ApplicationResponse::Json)
+    };
+    // pm_responsec = (PaymentMethodResponse,is_duplicate)
```
jarnura, line 105: remove comments

crates/router/src/core/payments/flows/authorize_flow.rs
```diff
@@ -128,6 +152,164 @@ impl types::PaymentsAuthorizeRouterData {
...
+) -> Option<serde_json::Value> {
+    connector_token.map(|connector_and_token| {
+        let mut metadata_map = serde_json::Map::new();
+        metadata_map.insert(
+            connector_and_token.0.connector_name.to_string(),
+            serde_json::Value::String(connector_and_token.1),
+        );
+        serde_json::Value::Object(metadata_map)
+    })
+}
```
jarnura, line 305:
```suggestion
pub fn create_payment_method_metadata(
    metadata: Option<&pii::SecretSerdeValue>,
    connector_token: Option<(&api::ConnectorData, String)>,
) -> Option<serde_json::Value> {
	 let mut meta = match metadata {
	    None  =>  serde_json::Map::new(),
		Some(meta) => meta.clone().expose(),
	}
    connector_token.and_then(|connector_and_token| {
    	meta.insert(
    		connector_and_token.0.connector_name.to_string(),
    		serde_json::Value::String(connector_and_token.1),
        )
   })
}
```

jarnura: Similar to this, in future it will have more fields in metadata and create also takes metadata as a input. So single function itself handle this. create metadata if not passed.

crates/router/src/core/payments/flows/cancel_flow.rs
```diff
@@ -69,6 +69,15 @@ impl Feature<api::Void, types::PaymentsCancelData>
     ) -> RouterResult<types::AddAccessTokenResult> {
         access_token::add_access_token(state, connector, merchant_account, self).await
     }
+
+    async fn add_payment_method_token<'a>(
```
jarnura, line 73: Add this as trait's default implementation

crates/router/src/core/payments/helpers.rs
```diff
@@ -709,21 +677,41 @@ pub async fn create_customer_if_not_exist<'a, F: Clone, R>(
...
-#[allow(clippy::too_many_arguments)]
 pub async fn make_pm_data<'a, F: Clone, R>(
     operation: BoxedOperation<'a, F, R>,
     state: &'a AppState,
     payment_data: &mut PaymentData<F>,
 ) -> RouterResult<(BoxedOperation<'a, F, R>, Option<api::PaymentMethodData>)> {
     let request = &payment_data.payment_method_data;
     let token = payment_data.token.clone();
+    let hyperswitch_token = if let Some(token) = token {
+        let redis_conn = crate::connection::redis_connection(&state.conf).await;
```
jarnura, line 688:
```suggestion
        let redis_conn = connection::redis_connection(&state.conf).await;
```

crates/storage_models/src/query/payment_method.rs
```diff
@@ -97,4 +98,29 @@ impl PaymentMethod {
         )
         .await
     }
+
+    pub async fn update(
```
jarnura, line 102: change function name, update_with_payment_method_id

jarnura (changes requested):

SanchithHegde:
You'd also have to update the connector template to include corresponding changes from this PR and #826, by the way.

crates/router/src/core/payments/transformers.rs
```diff
@@ -649,3 +650,15 @@ impl<F: Clone> TryFrom<PaymentAdditionalData<F>> for types::CompleteAuthorizeDat
         })
     }
 }
+
+// impl<F: Clone> TryFrom<PaymentData<F>> for types::PaymentMethodTokenizationData {
```
jarnura, line 654: Remove comments code

jarnura (approved):
LGTM

manoj-juspay (changes requested):

crates/router/src/core/payments/flows/authorize_flow.rs
```diff
@@ -128,6 +152,158 @@ impl types::PaymentsAuthorizeRouterData {
...
+            payment_method_create_request.to_owned(),
+        )
+        .await?;
+        let is_duplicate = locker_response.1;
+
+        if is_duplicate {
+            let pm = db
+                .find_payment_method(&locker_response.0.payment_method_id)
+                .await
+                .change_context(errors::ApiErrorResponse::PaymentMethodNotFound)?;
```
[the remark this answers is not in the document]
Sangamesh26: Done

crates/router/src/core/payment_methods/cards.rs
```diff
@@ -1458,20 +1444,21 @@ pub async fn list_customer_payment_method(
         ));
     }
     let mut customer_pms = Vec::new();
+    let parent_payment_method_token = generate_id(consts::ID_LENGTH, "token");
```
manoj-juspay, line 1447: Why do we need to introduce this new token ? @jarnura @Sangamesh26 
PM will be passed in /payment api and `hyperswitch_token` is unique for every entry in PM table, so in case of tokenization flow we can store `Vec<PaymentMethodMetadata>` against `hyperswitch_token`

manoj-juspay:
@Sangamesh26 can you reply to all the comment in case they are resolved!

crates/router/src/core/payment_methods/cards.rs
jarnura: Here we are treating locker also as a token provider as other token provider, so there is a single reference token to connector all provider token.

jarnura (approved):

manoj-juspay (dismissed):

jarnura (dismissed):

manoj-juspay (dismissed):

vspecky (dismissed):
LGTM

dracarys18 (dismissed):

Narayanbhat166 (dismissed):

jarnura (dismissed):

Narayanbhat166 (dismissed):

crates/router/src/core/payments.rs
```diff
@@ -507,7 +520,160 @@ where
...
+            );
+
+            match redis_conn.get_key::<String>(&key).await.ok() {
+                None => {
+                    if is_connector_tokenization_enabled {
+                        TokenizationAction::TokenizeInConnector
+                    } else {
+                        TokenizationAction::SkipConnectorTokenization
+                    }
+                }
```
dracarys18, line 579: Why are we calling `get_key` twice here?

crates/router/src/core/errors.rs
```diff
@@ -309,6 +309,8 @@ pub enum VaultError {
     MissingRequiredField { field_name: &'static str },
     #[error("The card vault returned an unexpected response: {0:?}")]
     UnexpectedResponseError(bytes::Bytes),
+    #[error("Failed to update in PMD table")]
+    UpdateInPMDTableFailed,
```
SanchithHegde, line 313: Either follow the Rust naming convention:

```suggestion
    UpdateInPmdTableFailed,
```

or use the complete name:

```suggestion
    UpdateInPaymentMethodDataTableFailed,
```

And please be verbose in the error messages, do not use abbreviations there.

crates/router/src/core/payment_methods/cards.rs
```diff
@@ -73,40 +79,30 @@ pub async fn add_payment_method(
...
+            let payment_method_response = api::PaymentMethodResponse {
                 merchant_id: merchant_id.to_string(),
                 customer_id: Some(customer_id),
-                payment_method_id: payment_method_id.to_string(),
+                payment_method_id: pm_id,
                 payment_method: req.payment_method,
                 payment_method_type: req.payment_method_type,
                 card: None,
-                metadata: req.metadata,
+                metadata: None,
```
SanchithHegde, line 96: Why are we no longer passing metadata here?

crates/router/src/core/payments.rs
```diff
@@ -507,7 +520,160 @@ where
...
+}
+
+#[allow(clippy::too_many_arguments)]
+pub async fn get_connector_tokenization_action<F, Req>(
+    state: &AppState,
+    operation: &BoxedOperation<'_, F, Req>,
+    connector_details: Option<&api::ConnectorCallType>,
+    mut payment_data: PaymentData<F>,
+    validate_result: &operations::ValidateResult<'_>,
+    _req: &Req,
```
dracarys18, line 601: Why do you need `req` here if it's not used inside the function

crates/router/src/core/payments.rs
```diff
@@ -507,7 +520,160 @@ where
...
+) -> RouterResult<(PaymentData<F>, TokenizationAction)>
+where
+    F: Send + Clone,
+{
+    let payment_data_and_tokenization_action =
+        if get_connector_from_connector_type(connector_details).is_some()
```
dracarys18, line 607: You can use 
```rust
match get_connector_from_connector_type(connector_details) {
	Some(connector) if is_operation_confirm(&operation)
}
```
Instead of calling `get_connector_from_connector_type` again in line number 615

crates/router/src/core/payments/helpers.rs
```diff
@@ -710,21 +678,41 @@ pub async fn create_customer_if_not_exist<'a, F: Clone, R>(
...
+                .payment_attempt
+                .payment_method
+                .to_owned()
+                .get_required_value("payment_method")?,
+        );
+        let val = redis_conn.get_key::<String>(&key).await;
+        match val {
+            Ok(token) => Some(token),
+            Err(_) => None,
+        }
```
SanchithHegde, line 703:
```suggestion
        redis_conn.get_key::<String>(&key).await.ok()
```

dracarys18 (changes requested):

crates/router/src/connector/stripe/transformers.rs
```diff
@@ -191,10 +206,33 @@ pub struct StripeBankRedirectData {
...
+#[derive(Debug, Eq, PartialEq, Serialize)]
+#[serde(untagged)]
+pub enum StripeWallet {
+    ApplepayToken(StripeApplePay),
+    ApplepayPayment(ApplepayPayment),
+}
+
+#[derive(Debug, Eq, PartialEq, Serialize)]
+pub struct StripeApplePay {
+    pub pk_token: String,
```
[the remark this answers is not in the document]
Sangamesh26: I am not sure what it stands for, but applepay has the naming convention like pk_token and pk_payment etc and stripe even requires this naming way itself.

SanchithHegde (approved):

dracarys18 (approved):

jarnura (approved):

PR #751, opened 2023-03-15
feat: generic in memory cache interface

dracarys18:
## Type of Change

- [x] Enhancement

## Description

This PR adds in memory cache support to the config table, the config in memory cache is a static with time to live as 30 mins and time to idle as 30 mins. On a normal get operation it checks if the data is in in memory cache and returns if exists. Otherwise it goes to db and gets the cache and inserts it into both redis and in memory.

Also switched to Arc for the value type in cache interface since moka clones the inner value on every get cloning an Arc is much more cheaper than cloning a Box.

## Motivation and Context

This is lower the latency of fetch in config table.

## How did you test it?

image

## Checklist

### DIFF
--- a/crates/router/src/cache.rs
+++ b/crates/router/src/cache.rs
@@ -1,7 +1,11 @@
-use std::any::Any;
+use std::{any::Any, sync::Arc};
 
 use dyn_clone::DynClone;
 use moka::future::Cache as MokaCache;
+use once_cell::sync::Lazy;
+
+/// Config Cache with time_to_live as 30 mins and time_to_idle as 10 mins.
+pub static CONFIG_CACHE: Lazy<Cache> = Lazy::new(|| Cache::new(30 * 60, 10 * 60));
 
 /// Trait which defines the behaviour of types that's gonna be stored in Cache
 pub trait Cacheable: Any + Send + Sync + DynClone {
@@ -20,11 +24,11 @@ where
 dyn_clone::clone_trait_object!(Cacheable);
 
 pub struct Cache {
-    inner: MokaCache<String, Box<dyn Cacheable>>,
+    inner: MokaCache<String, Arc<dyn Cacheable>>,
 }
 
 impl std::ops::Deref for Cache {
-    type Target = MokaCache<String, Box<dyn Cacheable>>;
+    type Target = MokaCache<String, Arc<dyn Cacheable>>;
     fn deref(&self) -> &Self::Target {
         &self.inner
     }
@@ -45,7 +49,11 @@ impl Cache {
         }
     }
 
-    pub fn get_val<T: Clone + Any>(&self, key: &str) -> Option<T> {
+    pub async fn push<T: Cacheable>(&self, key: String, val: T) {
+        self.insert(key, Arc::new(val)).await;
+    }
+
+    pub fn get_val<T: Clone + Cacheable>(&self, key: &str) -> Option<T> {
         let val = self.get(key)?;
         (*val).as_any().downcast_ref::<T>().cloned()
     }
@@ -58,9 +66,7 @@ mod cache_tests {
     #[tokio::test]
     async fn construct_and_get_cache() {
         let cache = Cache::new(1800, 1800);
-        cache
-            .insert("key".to_string(), Box::new("val".to_string()))
-            .await;
+        cache.push("key".to_string(), "val".to_string()).await;
         assert_eq!(cache.get_val::<String>("key"), Some(String::from("val")));
     }
 }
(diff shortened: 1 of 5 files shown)


crates/router/src/cache.rs
```diff
@@ -1,7 +1,11 @@
-use std::any::Any;
+use std::{any::Any, sync::Arc};
 
 use dyn_clone::DynClone;
 use moka::future::Cache as MokaCache;
+use once_cell::sync::Lazy;
+
+/// Config Cache with time_to_live as 30 mins and time_to_idle as 10 mins.
+pub static CONFIG_CACHE: Lazy<Cache> = Lazy::new(|| Cache::new(30 * 60, 10 * 60));
```
SanchithHegde, line 8: Do you think the TTL and TTI should be configurable or static?

crates/router/src/db/configs.rs
```diff
@@ -86,8 +87,17 @@ impl ConfigInterface for Store {
...
+        if let Some(val) = val {
+            Ok(val)
+        } else {
+            let val = cache::get_or_populate_cache(self, key, || async {
+                self.find_config_by_key(key).await
+            })
+            .await?;
+            CONFIG_CACHE.push(key.to_string(), val.clone()).await;
+            Ok(val)
+        }
```
SanchithHegde, line 100: It would be preferable if `get_or_populate_cache()` either handled three levels of hierarchy, or if the function was named `get_or_populate_redis()`.

Personally, I'd prefer the former.

crates/router/src/db/cache.rs
```diff
@@ -49,7 +53,7 @@ where
     store
         .redis_conn()
         .map_err(Into::<errors::StorageError>::into)?
-        .delete_key(key)
+        .publish(consts::PUB_SUB_CHANNEL, key)
```
jarnura, line 56: in all redact_cache we should not call publish, invalidation of the key can be happen in other application(not only dashboard) also.

So we need to do delete separately and a function which do delete and publish.

Dashboard will call delete and publish function

jarnura (changes requested):

SanchithHegde (dismissed):

crates/router/src/db/cache.rs
dracarys18: Done

crates/router/src/db/configs.rs
```diff
@@ -86,8 +87,13 @@ impl ConfigInterface for Store {
...
+        cache::get_or_populate_in_memory(
+            self,
+            key,
+            || async { self.find_config_by_key(key).await },
+            &CONFIG_CACHE,
+        )
+        .await
     }
 
     async fn delete_config_by_key(&self, key: &str) -> CustomResult<bool, errors::StorageError> {
```
jarnura, line 99: for deleted also we need to remove from all cache

crates/router/src/db/cache.rs
```diff
@@ -54,3 +83,22 @@ where
...
+) -> CustomResult<T, errors::StorageError>
+where
+    F: FnOnce() -> Fut + Send,
+    Fut: futures::Future<Output = CustomResult<T, errors::StorageError>> + Send,
+{
+    let data = fun().await?;
+    store
+        .redis_conn()
+        .map_err(Into::<errors::StorageError>::into)?
+        .publish(consts::PUB_SUB_CHANNEL, key)
```
jarnura, line 100: where delete from cache is happening?, this is only publishing right?

jarnura (changes requested):

crates/router/src/db/cache.rs
dracarys18: Deleting happens in `on_message` function of pubsub interface

jarnura (dismissed):

SanchithHegde (approved):

crates/router/src/db/configs.rs
```diff
@@ -86,16 +88,29 @@ impl ConfigInterface for Store {
         &self,
         key: &str,
     ) -> CustomResult<storage::Config, errors::StorageError> {
-        cache::get_or_populate_cache(self, key, || async { self.find_config_by_key(key).await })
-            .await
+        cache::get_or_populate_in_memory(
+            self,
+            key,
+            || async { self.find_config_by_key(key).await },
```
NishantJoshi00, line 94: can't we just do:
```suggestion
            || self.find_config_by_key(key),
```

NishantJoshi00 (changes requested):

NishantJoshi00 (approved):

jarnura (approved):

PR #752, opened 2023-03-15
feat(router): add support for stateful straight through routing

vspecky:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates

## Description

This PR makes the straight through routing feature stateful (i.e. merchant can now give a straight through config in payments create which will be further used in payments confirm)

### Additional Changes

[payment_attempts.rs](https://github.com/juspay/hyperswitch/blob/fe4671f3079878a36a2375ebc6f11c1044d5baa0/crates/storage_models/src/schema.rs) - make the `connector` field of payment_attempt `JSONB`.

## Motivation and Context

Merchant needs to be able to send a straight through config in payments create and be able to use this in payments confirm without having to send it again.

## How did you test it?

Stripe payments locally, compiler-guided.

## Checklist

### DIFF
--- a/crates/api_models/src/payments.rs
+++ b/crates/api_models/src/payments.rs
@@ -6,7 +6,7 @@ use router_derive::Setter;
 use time::PrimitiveDateTime;
 use utoipa::ToSchema;
 
-use crate::{enums as api_enums, refunds};
+use crate::{admin, enums as api_enums, refunds};
 
 #[derive(Clone, Copy, Debug, Eq, PartialEq)]
 pub enum PaymentOp {
@@ -200,6 +200,9 @@ pub struct PaymentsRequest {
     /// Payment Method Type
     #[schema(value_type = Option<PaymentMethodType>, example = "google_pay")]
     pub payment_method_type: Option<api_enums::PaymentMethodType>,
+
+    /// Merchant connector details used to make payments.
+    pub merchant_connector_details: Option<admin::MerchantConnectorDetailsWrap>,
 }
 
 #[derive(Default, Debug, serde::Deserialize, serde::Serialize, Clone, Copy, PartialEq, Eq)]
@@ -251,6 +254,7 @@ pub struct VerifyRequest {
     pub setup_future_usage: Option<api_enums::FutureUsage>,
     pub off_session: Option<bool>,
     pub client_secret: Option<String>,
+    pub merchant_connector_details: Option<admin::MerchantConnectorDetailsWrap>,
 }
 
 impl From<PaymentsRequest> for VerifyRequest {
@@ -269,6 +273,7 @@ impl From<PaymentsRequest> for VerifyRequest {
             mandate_data: item.mandate_data,
             setup_future_usage: item.setup_future_usage,
             off_session: item.off_session,
+            merchant_connector_details: item.merchant_connector_details,
         }
     }
 }
@@ -632,19 +637,6 @@ impl Default for PaymentIdType {
     }
 }
 
-//#[derive(Debug, serde::Deserialize, serde::Serialize)]
-//#[serde(untagged)]
-//pub enum enums::CaptureMethod {
-//Automatic,
-//Manual,
-//}
-
-//impl Default for enums::CaptureMethod {
-//fn default() -> Self {
-//enums::CaptureMethod::Manual
-//}
-//}
-
 #[derive(
     Default,
     Clone,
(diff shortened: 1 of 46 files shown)


SanchithHegde (dismissed):

crates/router/src/core/payments.rs
```diff
@@ -760,47 +743,151 @@ pub async fn add_process_sync_task(
...
 
-pub async fn route_connector<F>(
+pub fn update_straight_through_routing<F>(
+    payment_data: &mut PaymentData<F>,
+    request_straight_through: serde_json::Value,
+) -> RouterResult<()>
+where
+    F: Send + Clone,
+{
+    let mut routing_data: storage::RoutingData = payment_data
```
jarnura, line 753: All places the error is lifted to `InternalServerError`.

We can implement like common lifting as follows
```rust
|| -> _ {
   \\ code block with common error with attach_printable if needed
}().change_context(errors::ApiErrorResponse::InternalServerError)

```

jarnura: we can use this approach in other places too if all errors inside the scope are common.

jarnura (changes requested):

crates/router/src/core/payments.rs
vspecky: In that case we cannot attach specific printables on the `InternalServerError` context. Altho, after taking a look, I think it's okay to attach all printables on the `ParsingError` context in this case. So i guess we can go ahead with this. I'll make the change.

crates/api_models/src/payments.rs
```diff
@@ -63,9 +63,16 @@ pub struct PaymentsRequest {
     #[serde(default, deserialize_with = "amount::deserialize_option")]
     pub amount: Option<Amount>,
 
+    #[schema(value_type = Option<RoutingAlgorithm>, example = json!({
+        "type": "single",
+        "data": "stripe"
+    }))]
+    pub routing: Option<serde_json::Value>,
```
Narayanbhat166, line 70: If possible, have this as a concrete type which will enforce type saftey so that we don't accept invalid data. While inserting it into the database it can be made into `serde_json::value`

crates/router/src/core/payments.rs
vspecky: Since all errors are the same here, just lifted the ApiErrorResponse change_context outside the function.

crates/router/src/core/payments.rs
```diff
@@ -760,47 +745,148 @@ pub async fn add_process_sync_task(
...
+pub fn update_straight_through_routing<F>(
+    payment_data: &mut PaymentData<F>,
+    request_straight_through: serde_json::Value,
+) -> CustomResult<(), errors::ParsingError>
+where
+    F: Send + Clone,
+{
+    let mut routing_data: storage::RoutingData = payment_data
+        .payment_attempt
+        .connector
```
Narayanbhat166, line 757: If the connector field is empty in payment_attempt, then this unwrap will return `{}` which will not be parsed in the next step. Is this intended behaviour?

Narayanbhat166 (changes requested):

crates/api_models/src/payments.rs
vspecky: The type will have different interpretations for OSS and VAS, that's why the decision was made to make it generic.

crates/router/src/core/payments.rs
vspecky: The type it's being parsed to has all fields as optional, so it will parse to the type with all values as null. This is intended behavior.

Narayanbhat166:
This change would break the payments list and payments retrieve for payments created prior to this change, are we handling that?

Narayanbhat166 (approved):

crates/router/src/core/payments.rs
```diff
@@ -99,27 +99,35 @@ where
...
-    let connector_details = operation
+    let connector_choice = operation
         .to_domain()?
-        .get_connector(
-            &merchant_account,
-            state,
-            &req,
-            payment_data.payment_attempt.connector.as_ref(),
-        )
+        .get_connector(&merchant_account, state, &req)
```
jarnura, line 104: Create a new function and add the below logic and this function in that.

jarnura (changes requested):

crates/router/src/core/payments.rs
vspecky: Are you sure we want to move `.get_connector()` in a separate function? It's using the `operation` and invoking a domain function.

jarnura: The agenda is core just call a connector function and all logic's abstracted inside that

jarnura (approved):

crates/api_models/src/payments.rs
vspecky: As per discussion with @jarnura, we will need to come up with some other strategy. For now we will go with this.

Narayanbhat166 (approved):

vspecky:
> would this change break the payments list and payments retrieve for payments created prior to this change, if so are we handling that?

Yes it is handled in this PR.

PR #759, opened 2023-03-16
feat(pm_list): handle client secret check

### LINKED ISSUE #756: [FEATURE] Addition of Client Secret Validation in Payment Methods API.
### Feature Description

With the current workflow the clientSecret essentially gets verified at the very end during confirm payment. What we basically want to do however is to make sure to not show the SDK or throw the appropriate errors when that is the case as soon as possible. So during Payment Methods API call that happens at the very beginning.

### Possible Implementation

So instead of sending the payment_methods  list , in the case of incorrect client secret, we want an error body which will tell us one of the 3 things. 
1. Expired ClientSecret. 
2. Completed/Failed ClientSecret. 
3. Incorrect ClientSecret. 
After knowing any 1 of these, we can throw an error in the UI for the same.

None

Narayanbhat166:
## Type of Change

- [x] New feature

## Description

If there is no client secret in payment_intent then it has expired, so if we receive a client secret in request and there is no client secret in payment intent then the client secret provided has expired.

## Motivation and Context

Closes #756 
For the client secret to be delete on failed payment #724 PR needs to be merged

## How did you test it?

- Create a successful payment
- Call payment_method list which fails saying client secret is invalid

## Checklist

### DIFF
--- a/crates/router/src/core/errors/api_error_response.rs
+++ b/crates/router/src/core/errors/api_error_response.rs
@@ -319,7 +319,7 @@ impl common_utils::errors::ErrorSwitch<api_models::errors::types::ApiErrorRespon
             Self::ClientSecretNotGiven => AER::BadRequest(ApiError::new(
                 "IR",
                 8,
-                "Client secret was not provided", None
+                "client_secret was not provided", None
             )),
             Self::ClientSecretInvalid => {
                 AER::BadRequest(ApiError::new("IR", 9, "The client_secret provided does not match the client_secret associated with the Payment", None))
--- a/crates/router/src/core/errors/api_error_response.rs
+++ b/crates/router/src/core/errors/api_error_response.rs
@@ -49,6 +49,8 @@ pub enum ApiErrorResponse {
     InvalidDataValue { field_name: &'static str },
     #[error(error_type = ErrorType::InvalidRequestError, code = "IR_08", message = "Client secret was not provided")]
     ClientSecretNotGiven,
+    #[error(error_type = ErrorType::InvalidRequestError, code = "IR_08", message = "Client secret has expired")]
+    ClientSecretExpired,
     #[error(error_type = ErrorType::InvalidRequestError, code = "IR_09", message = "The client_secret provided does not match the client_secret associated with the Payment")]
     ClientSecretInvalid,
     #[error(error_type = ErrorType::InvalidRequestError, code = "IR_10", message = "Customer has active mandate/subsciption")]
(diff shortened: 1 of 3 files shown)


crates/router/src/core/errors/api_error_response.rs
```diff
@@ -344,7 +347,12 @@ impl common_utils::errors::ErrorSwitch<api_models::errors::types::ApiErrorRespon
             Self::InvalidJwtToken => AER::Unauthorized(ApiError::new("IR", 17, "Access forbidden, invalid JWT token was used", None)),
             Self::GenericUnauthorized { message } => {
                 AER::Unauthorized(ApiError::new("IR", 18, message.to_string(), None))
-            }
+            },
+            Self::ClientSecretExpired => AER::BadRequest(ApiError::new(
+                "IR",
+                19,
+                "The given client secret has expired", None
```
SanchithHegde, line 354:
```suggestion
                "The provided client secret has expired", None
```

SanchithHegde (commented):
Other than a couple of optional changes, looks good to me!

jarnura (approved):

SanchithHegde (dismissed):

crates/router/src/core/errors/api_error_response.rs
```diff
@@ -49,6 +49,8 @@ pub enum ApiErrorResponse {
     InvalidDataValue { field_name: &'static str },
     #[error(error_type = ErrorType::InvalidRequestError, code = "IR_08", message = "Client secret was not provided")]
```
manojradhakrishnan, line 50: Please use consistent terminology in error messages. (i.e) "The client_secret was not provieded" instead of "Client secret..."

jarnura (dismissed):

crates/router/src/core/errors/api_error_response.rs
Narayanbhat166: This error message will not be sent in the api response, this is for internal use only. Have changed it anyways.

jarnura (approved):

crates/router/src/core/errors/api_error_response.rs
```diff
@@ -319,7 +319,7 @@ impl common_utils::errors::ErrorSwitch<api_models::errors::types::ApiErrorRespon
             Self::ClientSecretNotGiven => AER::BadRequest(ApiError::new(
                 "IR",
                 8,
-                "Client secret was not provided", None
+                "client_secret was not provided", None
```
SanchithHegde, line 322: If we're mentioning field names, can we enclose them in backticks? We can make this change across all error messages in a separate PR if necessary.

So the error message would be:
```
`client_secret` was not provided
```

SanchithHegde (approved):

PR #762, opened 2023-03-16
refactor(connector): update add_connector script

jagan-jaya:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [X] Refactoring
- [ ] Dependency updates

## Description

Update the add_connector script as per current code base

### Additional Changes

## Motivation and Context

### DIFF
--- a/scripts/add_connector.sh
+++ b/scripts/add_connector.sh
@@ -1,33 +1,40 @@
 #! /usr/bin/env bash
 
 pg=$1;
+base_url=$2;
 pgc="$(tr '[:lower:]' '[:upper:]' <<< ${pg:0:1})${pg:1}"
 src="crates/router/src"
 conn="$src/connector"
-tests="../../tests/connectors/"
+tests="../../tests/connectors"
 SCRIPT="$( cd -- "$(dirname "$0")" >/dev/null 2>&1 ; pwd -P )"
-if [[ -z "$pg" ]]; then 
-    echo 'Connector name not present: try "sh add_connector.sh <adyen>"'
+RED='\033[0;31m'
+GREEN='\033[0;32m'
+ORANGE='\033[0;33m'
+
+if [ -z "$pg" ] || [ -z "$base_url" ]; then 
+    echo "$RED Connector name or base_url not present: try $GREEN\"sh add_connector.sh adyen https://test.adyen.com\""
     exit
 fi
 cd $SCRIPT/..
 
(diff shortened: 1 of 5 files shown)


scripts/add_connector.sh
```diff
@@ -1,33 +1,60 @@
 #! /usr/bin/env bash
-
+function find_prev_connector() {
+    self=scripts/add_connector.sh
+    cp $self $self.tmp
+    connectors=(aci adyen airwallex applepay authorizedotnet bambora bluesnap braintree checkout cybersource dlocal fiserv forte forte forte forte forte forte forte forte forte forte forte forte forte forte globalpay klarna mollie multisafepay nuvei payu rapyd shift4 stripe trustpay utils worldline worldpay "$1")
```
SanchithHegde, line 5: 1. Is this going to be hard-coded, or do you plan to do something along the lines of: "list dirs in `router/src/connectors`, sort everything except `utils` in order"?.
2. `forte` is repeated multiple times.

scripts/add_connector.sh
```diff
@@ -1,33 +1,60 @@
...
+    if [ "${sorted[$i]}" = "$1" ] && [ $i != "0" ]; then
+        eval "$2='${sorted[i-1]}'"
+        mv $self.tmp $self
+        rm $self.tmp-e
+        return 0
+    fi
+    done
+    mv $self.tmp $self
+    rm $self.tmp-e
+    eval "$2='aci'"
```
SanchithHegde, line 19: What is exactly happening here? It'd be preferable if this had comments.

scripts/add_connector.sh
jagan-jaya: 1. it is going to be hardcoded, whenever some add a new connector,it will update this file also automatically
2. will fix it

scripts/add_connector.sh
jagan-jaya: we are trying to find the previous connector so that we can add our current connector next to it. this helps in keeping the connectors in sorted order

jagan-jaya: `Aci` is for corner case, if there is no connector with name `ab*` it will be added after Aci.

jarnura (approved):

jarnura (approved):

SanchithHegde (approved):

ArjunKarthik (approved):

PR #770, opened 2023-03-20
feat(connector): Add support to provide connector_payment_meta for capture and void request

jagan-jaya:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [X] Enhancement
- [ ] Refactoring
- [ ] Dependency updates

## Description

1. Adding support to provide meta_data for capture and void request also
2. Bug fixes in add_connector script

### Additional Changes

## Motivation and Context

Previously when a detail from payment authorize response is needed in any of the payment flow we only add it in the `connector_meta` of payment_attempt table, but it was only used in PaymentSync, now we are enabling the Capture and Void flows to make use of connector_meta from payment_attempt. 

## How did you test it?

Ran basic sanity cases locally

## Checklist

### DIFF
--- a/crates/router/src/core/payments/transformers.rs
+++ b/crates/router/src/core/payments/transformers.rs
@@ -508,6 +508,7 @@ impl<F: Clone> TryFrom<PaymentData<F>> for types::PaymentsCaptureData {
                 .connector_transaction_id
                 .ok_or(errors::ApiErrorResponse::MerchantConnectorAccountNotFound)?,
             amount: payment_data.amount.into(),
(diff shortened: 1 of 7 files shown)


vspecky (changes requested):
We'll also need to add this for `Void`.

vspecky (approved):
Ignore previous comment. Cancel = Void.

ArjunKarthik (approved):

jarnura (approved):
Make metadata as secret, other than that looks good to me.

PR #794, opened 2023-03-22
chore: prepare for building production Docker images

SanchithHegde:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring
- [ ] Dependency updates

## Description

This PR includes the initial changes required for building Docker images with the `production` feature of the `router` crate and using the `Production` run environment. In addition, the manual release workflow now accepts the environment to build the release image for.

## Motivation and Context

Moving to production! :tada::rocket:

## How did you test it?

Created a new build, verified that the image looks for a `Production.toml` config file.

## Checklist

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,7 @@
 FROM rust:1.65 as builder
+
+ARG RUN_ENV=Sandbox
+
 RUN apt-get update \
     && apt-get install -y libpq-dev libssl-dev
 
@@ -27,7 +30,9 @@ ENV RUSTUP_MAX_RETRIES=10
 ENV RUST_BACKTRACE="short"
 
 COPY . .
-RUN cargo build --release --features sandbox
+
+# Use bash variable substitution to convert environment name to lowercase
+RUN bash -c 'cargo build --release --features ${RUN_ENV@L}'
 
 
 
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -27,7 +30,9 @@ ENV RUSTUP_MAX_RETRIES=10
 ENV RUST_BACKTRACE="short"
 
 COPY . .
-RUN cargo build --release --features sandbox
+
+# Use bash variable substitution to convert environment name to lowercase
+RUN bash -c 'cargo build --release --features ${RUN_ENV@L}'
```
SanchithHegde, line 35: Honestly though, I'd prefer using lowercase names for `RUN_ENV` and config files (`development.toml` / `sandbox.toml` / `production.toml`) instead of doing such shell sorcery. For the config files, I'm even okay with using a single `config.toml` instead, if that's preferable.

I'm aware this would break a lot of things including sandbox, but I'd prefer doing it now rather than after we move to production. 

I've made the necessary changes in https://github.com/juspay/hyperswitch/commit/5863a988bc43e766831b0724b076ae752e7f1b8a, let me know what you think.

lsampras (approved):

jarnura (approved):

Dockerfile
SanchithHegde: Okay, I'll take it up on a separate PR.

PR #802, opened 2023-03-26
docker: build using latest stable Rust toolchain and use sparse index protocol

SanchithHegde:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring
- [ ] Dependency updates

## Description

This PR updates the `Dockerfile` to use the latest stable Rust toolchain instead of MSRV to build images, since newer versions of the compiler usually receive performance improvements as compared to the older versions. In addition, this PR updates the `Dockerfile` and CI workflow to use the new "sparse" index protocol to fetch crates from the crates.io registry.

## Motivation and Context

Faster build times.

## How did you test it?

N/A

## Checklist

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,9 +1,9 @@
-FROM rust:latest as builder
+FROM rust:slim as builder
 
 ARG RUN_ENV=Sandbox
 
 RUN apt-get update \
-    && apt-get install -y libpq-dev libssl-dev
+    && apt-get install -y libpq-dev libssl-dev pkg-config
 
 # Copying codebase from current dir to /router dir
 # and creating a fresh build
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM rust:1.65 as builder
+FROM rust:latest as builder
```
ashokkjag, line 1: Does slim make much difference?

SanchithHegde: As far as I understand, `slim` does not include some packages useful for building. I'll check if it makes too much of a difference.

jarnura (approved):

Dockerfile
SanchithHegde: `slim` required only `pkg-config` to be installed additionally. I'll switch to it.

ashokkjag (approved):

PR #813, opened 2023-03-29
fix(cads_info): add extra columns to cards_info struct

Narayanbhat166:
## Type of Change

- [x] Bugfix

## Description

There are some columns in `cards_info` table which will be used and updated by the crawlers. These columns are added in the migration. But when migrations are run, the columns are added to `schema.rs`. Since these fields are not in the `CardInfo` struct in `storage_models` the code will not compile. This PR adds those columns to the struct.

## Motivation and Context

Make the code compile.

## How did you test it?

Screenshot 2023-03-29 at 4 29 40 PM

## Checklist

### DIFF
--- a/crates/storage_models/src/cards_info.rs
+++ b/crates/storage_models/src/cards_info.rs
@@ -1,4 +1,5 @@
 use diesel::{Identifiable, Queryable};
+use time::PrimitiveDateTime;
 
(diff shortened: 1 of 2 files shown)


vspecky (approved):
LGTM

PR #816, opened 2023-03-30
feat: multiple connector account support for the same `country`

Narayanbhat166:
## Type of Change

- [x] New feature

## Description

This PR adds support for multiple connector account support for the same country. The changes are built on top of #623. 
In order to ensure backwards compatibility, default values are provided.
- Default value for `country` is `US`
- Default value for `business` is `default`.

The decision to pick a connector is based on `connector_label`. The format of this label is 
`{connector_name}_{business_country}_{business_label}_{business_sub_label: Optional}`. 
For a merchant with `grocery` business in `US` the label would look like `stripe_US_grocery`. 
Business sub label is used only in case of specific connectors where we would need additional data for choosing the connector.

It is mandatory to pass `business` data when creating the `merchant_account`. These values are considered as default values when a payment is created without providing the business data. 

Default values are provided in the database too by modifying few migrations. There was a need to modify the migration as those would not run because a value would be missing from a field which was marked as required.

### Additional Changes

## Motivation and Context

Allow merchants to create multiple connector accounts within the same country.

## How did you test it?

- Create a merchant account with Business details.
- Create a connector account with both `business_label` and `business_country`.
- Create a payment without business details -> autofilled by referencing the merchant account.
- Create a payment for all the above use cases.

## Checklist


jarnura (dismissed):

crates/api_models/src/admin.rs
```diff
@@ -347,6 +377,70 @@ pub struct MerchantConnector {
...
+    /// Type of the Connector for the financial use case. Could range from Payments to Accounting to Banking.
+    #[schema(value_type = ConnectorType, example = "payment_processor")]
+    pub connector_type: api_enums::ConnectorType,
+    /// Name of the Connector
+    #[schema(example = "stripe")]
+    pub connector_name: String,
+
+    /// Business Sub label of the merchant
+    #[schema(example = "chase")]
+    pub business_sub_label: Option<String>,
```
vspecky, line 393: We cannot let them change these fields when updating connector information since the connector label needs to be immutable. I'd suggest either removing them or skipping deserializing these.

crates/router/src/core/payments/operations/payment_create.rs
```diff
@@ -514,6 +521,8 @@ impl PaymentCreate {
             statement_descriptor_name: request.statement_descriptor_name.clone(),
             statement_descriptor_suffix: request.statement_descriptor_suffix.clone(),
             metadata: metadata.map(masking::Secret::new),
+            business_country,
+            business_label,
```
vspecky, line 525: Shouldn't we also store the sub label in the payment intent?

crates/router/src/core/payments/operations/payment_confirm.rs
```diff
@@ -369,6 +369,8 @@ impl<F: Clone> UpdateTracker<F, PaymentData<F>, api::PaymentsRequest> for Paymen
         let customer_id = customer.map(|c| c.customer_id);
         let return_url = payment_data.payment_intent.return_url.clone();
         let setup_future_usage = payment_data.payment_intent.setup_future_usage;
+        let business_label = Some(payment_data.payment_intent.business_label.clone());
+        let business_country = Some(payment_data.payment_intent.business_country);
 
         payment_data.payment_intent = db
             .update_payment_intent(
```
vspecky, line 376: We should also update the payment attempt with these details right?

crates/router/src/core/payments/operations/payment_update.rs
```diff
@@ -414,6 +422,8 @@ impl<F: Clone> UpdateTracker<F, PaymentData<F>, api::PaymentsRequest> for Paymen
 
         let return_url = payment_data.payment_intent.return_url.clone();
         let setup_future_usage = payment_data.payment_intent.setup_future_usage;
+        let business_label = Some(payment_data.payment_intent.business_label.clone());
+        let business_country = Some(payment_data.payment_intent.business_country);
 
         payment_data.payment_intent = db
             .update_payment_intent(
```
vspecky, line 429: We should also update payment attempt with these details right?

vspecky (changes requested):

crates/router/src/core/payments/operations/payment_create.rs
Narayanbhat166: Sublabel can change for a payment intent for retries, if the merchant wants to retry the payment with a different Sublabel. So this field will be stored in payment attempt

bernard-eugine: When retrieving a completed payment, which sub_label will be attached to the payment_intent? The successful payment_attempt's sub label?

Narayanbhat166: Yes, sub label will be of the payment attempt which was made recently, or in other words active payment attempt.

vspecky: Makes sense, cool.

crates/api_models/src/admin.rs
Narayanbhat166: This field is removed from the updatable fields.

crates/router/src/core/payments/operations/payment_confirm.rs
Narayanbhat166: Yes, payment attempt should be updated with business_sub_label

crates/router/src/core/payments/operations/payment_update.rs
Narayanbhat166: Yes, payment attempt should be updated with business_sub_label

crates/api_models/src/admin.rs
vspecky: Shouldn't we also remove the connector name?

Narayanbhat166: Oh right, that would also affect connector label. Yes connector name should also not be allowed to update.

crates/router/src/core/payments/transformers.rs
```diff
@@ -348,6 +364,7 @@ where
                                 .map(ForeignInto::foreign_into),
                         )
                         .set_metadata(payment_intent.metadata)
+                        .set_connector_label(connector_label)
```
vspecky, line 367: Here I think it's better if we also return the `business_country`, `business_label`, `business_sublabel` and `connector_name` separately alongside the `connector_label`. Will make accessing that data easier downstream as opposed to deconstructing the `connector_label` to get it.

vspecky (changes requested):

crates/router/src/core/payments/transformers.rs
Narayanbhat166: These will be 3 additional fields. We can send it, but I felt like the response will contain too many fields. Thinking of it, better to send them in the response.

crates/api_models/src/admin.rs
```diff
@@ -186,6 +193,9 @@ pub struct MerchantAccountResponse {
     /// An identifier for the vault used to store payment method information.
     #[schema(example = "locker_abc123")]
     pub locker_id: Option<String>,
+    ///Default business details for connector routing
+    #[schema(value_type = Option<PrimaryBusinessDetails>)]
+    pub primary_business_details: serde_json::Value,
```
jarnura, line 198:
```suggestion
    pub primary_business_details: pii::SecretSerdeValue,
```

jarnura (changes requested):

vspecky (approved):
Overall LGTM

crates/api_models/src/admin.rs
Narayanbhat166: This is in response, do we need to have secret value in all the other places?

jarnura: yes

jarnura (dismissed):

vspecky (dismissed):

vspecky (approved):
LGTM

vspecky (approved):

PR #511, opened 2023-02-06
feat(router): implement API endpoints for managing API keys

SanchithHegde:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates

## Description

This PR implements the endpoints to create, retrieve, update, revoke and list API keys. 

- As of now, the `last_used` field is included in the storage model, but I don't expect it to be actively used until API keys are cached in Redis or in-memory.
- The endpoints use admin API key authentication, but require an `X-Merchant-Id` header to be sent for creating an API key and to list API keys for a specific merchant.

This PR only implements the endpoints for managing API keys, I'll update merchant authentication to use these API keys in a separate PR.

### Additional Changes

- [x] This PR modifies the database schema
- [ ] This PR modifies application configuration/environment variables

The relevant migrations are [`up.sql`](https://github.com/juspay/hyperswitch/blob/api-keys-endpoints/migrations/2023-02-01-135102_create_api_keys_table/up.sql) and [`down.sql`](https://github.com/juspay/hyperswitch/blob/api-keys-endpoints/migrations/2023-02-01-135102_create_api_keys_table/down.sql).

## Motivation and Context

Closes #501.

## How did you test it?

I have added a few unit tests where database accesses are not involved.

I'll attach invocations of the APIs via `curl` to give an idea of how they look.

- Create API Key:
  
  ```shell
  curl --location --request POST 'http://localhost:8080/api_keys' \
  --header 'Content-Type: application/json' \
  --header 'Accept: application/json' \
  --header 'X-Merchant-Id: merchant123' \
  --header 'api-key: MyVerySecretAdminApiKey' \
  --data-raw '{
  "name": "API Key 1",
  "description": null,
  "expiration": "2023-09-23T01:02:03.000Z"
  }'
  {
    "key_id": "dev_PwX5u10PjAbs9XWac0B2",
    "merchant_id": "merchant123",
    "name": "API Key 1",
    "description": null,
    "api_key": "dev_PwX5u10PjAbs9XWac0B2-MyVerySecret64CharacterLongApiKey",
    "created": "2023-02-06T14:16:46.144Z",
    "expiration": "2023-09-23T01:02:03.000Z"
  }
  ```

- Retrieve API Key:

  ```shell
  curl --location --request GET 'http://localhost:8080/api_keys/dev_PwX5u10PjAbs9XWac0B2' \
  --header 'Content-Type: application/json' \
  --header 'Accept: application/json' \
[...]

PR #819, opened 2023-03-31
refactor: use newtype pattern for email addresses

### LINKED ISSUE #608: [FEATURE] Use newtype pattern for email addresses
### Feature Description

This issue covers one of the problems mentioned in #119, about using [the newtype pattern](https://rust-unofficial.github.io/patterns/patterns/behavioural/newtype.html) for email addresses. The requirement is that email addresses must be validated once during construction/deserialization and should never be validated again. This would also eliminate the validation being performed at the time of masking email addresses (specifically line 97 in the snippet):

https://github.com/juspay/hyperswitch/blob/d107b44fd356f172ef7c9535567d266a6301e43d/crates/common_utils/src/pii.rs#L87-L109

Please also include unit tests in your PR.

### Possible Implementation

You can use a newtype like so:

```rust
use masking::Secret;

struct Email(Secret<String>);

let email = Email::from_str("..."); // You can use the `FromStr` trait for this purpose
```

For validating emails, the existing regex validation should suffice. Also, implement `Serialize`, `Deserialize` and `masking::Strategy` for `Email`.

None

nachiketkanore:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [x] Enhancement
- [ ] Refactoring
- [ ] Dependency updates

## Description

- This fixes #608 
- Added testcases

### Additional Changes

## Motivation and Context

#608 

## How did you test it?

- ran `cargo test`
- Added 2 unit testcases for validation of implemented code
- img

## Checklist


jarnura (dismissed):

SanchithHegde (changes requested):
Thanks for your contribution, @nachiketkanore!

The motivation behind #608 was eliminating repeated validations of email addresses each time they were masked. However, with the existing set of changes, that hasn't been addressed.

I'll lay out a possible set of steps to address this:

1. Move the email validation logic from the `validate_email()` function to the `FromStr` implementation on `Email`.
2. Update the implementation of `Strategy` on `Email` to accept `&self` instead of a generic `T`, and remove the validation happening at the time of masking emails.
3. Implement `Serialize` and `Deserialize` for `Email` such that validation happens during deserialization.
4. Substitute email addresses in our request/response types to use `Email` instead of `String` or `Secret<String>`.

Let me know if you have any more questions.

nachiketkanore:
- Shouldn't std::fmt::Debug be implemented for struct Email(Secret<String>)?
Since, we will be calling format!(email) where email: Email(Secret<String>)
- So, Email(Secret<String>) will replace both Email(String) and Secret<String, Email>?
@SanchithHegde

SanchithHegde:
> * Shouldn't std::fmt::Debug be implemented for struct Email(Secret)?
>   Since, we will be calling format!(email) where email: Email(Secret)

You should be able to derive `Debug` on `Email(Secret<String>)` since `Secret` implements `Debug` already.

> * So, Email(Secret) will replace both Email(String) and Secret<String, Email>?

I'm not sure if there are any types of the form `Email(String)` in our code. But if there are any, please do replace that and `Secret<String, Email>` with `Email(Secret<String>)`.

nachiketkanore:
```rust
let secret: Secret<String, Email> = Secret::new("myemail@gmail.com".to_string());
assert_eq!("*******@gmail.com", format!("{secret:?}"));
```
What about these tests?

SanchithHegde:
You'd have to update/remove tests accordingly.

nachiketkanore:
Sorry, this got closed by mistake. 
I was trying to sync with the remote, but somehow the PR closed.
Still working on this. Can you reopen this?

nachiketkanore:
Sorry, this got closed by mistake.
I was trying to sync with the remote, but somehow the PR closed.

crates/common_utils/Cargo.toml
```diff
@@ -47,6 +36,7 @@ md5 = "0.7.0"
 # First party crates
 masking = { version = "0.1.0", path = "../masking" }
 router_env = { version = "0.1.0", path = "../router_env", features = ["log_extra_implicit_fields", "log_custom_entries_to_extra"], optional = true }
+diesel = "2.0.3"
```
SanchithHegde, line 39: It'd be great if we could have this behind a `diesel` feature flag.

@Narayanbhat166 @NishantJoshi00 Or would you prefer the type being moved to another crate?

crates/common_utils/src/pii.rs
```diff
@@ -88,20 +98,103 @@ where
...
+#[derive(serde::Serialize, serde::Deserialize, Debug, Clone, PartialEq, Eq, Default)]
+pub struct Email(
+    Secret<String>
+);
+
+impl From<String> for Email {
+    fn from(value: String) -> Self {
+        Email(Secret::new(value))
+    }
+}
```
SanchithHegde, line 110: We'd want to avoid this, as this doesn't validate if an email is correct. Construction of an email from a `String` or `&str` could fail, so the only way to perform that conversion is using the `FromStr` trait.

crates/common_utils/src/pii.rs
```diff
@@ -88,20 +98,103 @@ where
...
+    fn walk_ast(&self, mut out: AstPass<Pg>) -> QueryResult<()> {
+        // out.push_sql(" sql statement ");
+        Ok(())
+    }
+}
+
+
+impl Expression for Email {
+    type SqlType = sql_types::Nullable<Text>;
+}
```
SanchithHegde, line 126: Do you need any of these implementations? As for `Expression`, you should be able to implement `AsExpression` instead, similar to how it's been done for `Secret` in the `masking` crate.

crates/common_utils/src/pii.rs
```diff
@@ -88,20 +98,103 @@ where
...
+    ST: sql_types::SingleValue,
+    DB: Backend,
+    Email: FromSql<ST, DB>
+{
+    type Row = Self;
+
+    fn build(row: Self::Row) -> deserialize::Result<Self> {
+        Ok(row)
+    }
+}
```
SanchithHegde, line 139: I believe implementing `Queryable` manually isn't required either.

crates/common_utils/src/pii.rs
```diff
@@ -88,20 +98,103 @@ where
...
+
+impl<DB> FromSql<sql_types::Text, DB> for Email
+where
+    DB: Backend,
+    String: FromSql<sql_types::Text, DB>,
+{
+    fn from_sql(bytes: backend::RawValue<'_, DB>) -> deserialize::Result<Self> {
+        let val = String::from_sql(bytes)?;
+        return Ok(Email::from_str(val.as_str()).unwrap());
```
SanchithHegde, line 148: No `unwrap()`s please, we avoid panicking code once our application starts.

crates/common_utils/src/pii.rs
```diff
@@ -88,20 +98,103 @@ where
...
+impl<T> Strategy<T> for EmailStrategy
+where
+    T: fmt::Display
 {
     fn fmt(val: &T, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        let val_str: &str = val.as_ref();
-        let is_valid = validate_email(val_str);
+        write!(f, "{}", val.to_string().to_ascii_lowercase())
+    }
+}
```
SanchithHegde, line 171: Is this needed if you have `Strategy` implemented for `Email`?

crates/router/src/core/payments/operations/payment_session.rs
```diff
@@ -162,7 +162,7 @@ impl<F: Send + Clone> GetTracker<F, PaymentData<F>, api::PaymentsSessionRequest>
                 payment_attempt,
                 currency,
                 amount,
-                email: None::<Secret<String, pii::Email>>,
+                email: None::<pii::Email>,
```
SanchithHegde, line 165: You shouldn't need to specify the type here, just `None` should suffice.

crates/common_utils/src/pii.rs
```diff
@@ -88,20 +98,103 @@ where
 }
 
 /// Email address
-#[derive(Debug)]
-pub struct Email;
+#[derive(serde::Serialize, serde::Deserialize, Debug, Clone, PartialEq, Eq, Default)]
+pub struct Email(
+    Secret<String>
+);
```
SanchithHegde, line 104: How is the deserialization happening, if it's not using the `FromStr` implementation?

crates/router/src/types/api/customers.rs
```diff
@@ -20,19 +20,19 @@ pub(crate) trait CustomerRequestExt: Sized {
...
+//             .as_ref()
+//             .validate_opt(|email| utils::validate_email(email.peek()))
+//             .change_context(errors::ApiErrorResponse::InvalidDataFormat {
+//                 field_name: "email".to_string(),
+//                 expected_format: "valid email address".to_string(),
+//             })?;
+//
+//         Ok(self)
+//     }
+// }
```
SanchithHegde, line 35: You should be able to remove the commented code, and the `CustomerRequestExt` trait definition.

SanchithHegde (commented):
Other than that, looks good to me. 

As you mentioned on Discord, cleanup and addressing CI checks (formatting and clippy) is also pending.

crates/common_utils/src/pii.rs
nachiketkanore:
```
error[E0277]: the trait bound `Email: AppearsOnTable<schema::customers::table>` is not satisfied
  --> crates/storage_models/src/customers.rs:48:33
   |
48 | #[derive(Clone, Debug, Default, AsChangeset, router_derive::DebugAsDisplay)]
   |                                 ^^^^^^^^^^^ the trait `AppearsOnTable<schema::customers::table>` is not implemented for `Email`
```
AppearsOnTable is required for Email

SanchithHegde: You can derive `AsExpression` on `Email` like so:

```rust
#[derive(diesel::AsExpression)]
#[diesel(sql_type = diesel::sql_types::Text)]
pub struct Email(Secret<String>);
```

This should handle it, since `AsExpression` helps with obtaining an `Expression`, and `AppearsOnTable` is implemented automatically for types which implement `Expression`. I did something similar for `HashedApiKey` when working on #511. (You can check [this snippet](https://github.com/juspay/hyperswitch/blob/b12762e776205a1b6375a7c49447e8a3dc541b8a/crates/storage_models/src/api_keys.rs#L80-L82) and the [`diesel_impl` module](https://github.com/juspay/hyperswitch/blob/b12762e776205a1b6375a7c49447e8a3dc541b8a/crates/storage_models/src/api_keys.rs#L96).)

crates/router/src/core/customers.rs
```diff
@@ -202,7 +203,10 @@ pub async fn delete_customer(
 
     let updated_customer = storage::CustomerUpdate::Update {
         name: Some(REDACTED.to_string()),
-        email: Some(REDACTED.to_string().into()),
+        email: match Email::from_str(REDACTED) {
+            Ok(email) => Some(email),
+            Err(_) => None,
+        },
```
SanchithHegde, line 209: This would always fail and would cause unexpected behavior when we're trying to redact customer details, could you add a check in `Email::from_str()` that allows `REDACTED` as a valid email?

SanchithHegde (commented):
Apart from that, looks good to me!

crates/router/src/core/customers.rs
nachiketkanore: Changes done. Also added a test for checking the REDACTED value.

crates/common_utils/src/pii.rs
```diff
@@ -88,25 +100,73 @@ where
...
-            WithType::fmt(val, f)
+        match Self::from_str(val_str) {
+            Ok(_email) => match val_str.split_once('@') {
+                Some((a, b)) => write!(f, "{}@{}", "*".repeat(a.len()), b),
+                None => WithType::fmt(val, f),
+            },
+            Err(..) => WithType::fmt(val, f),
         }
     }
 }
```
vspecky, line 172: Hey, there seems to be a slight misunderstanding regarding the purpose of `Strategy<T>` and how it's supposed to work with a `Secret<D, S>`. Please refer to my comment [here](https://github.com/juspay/hyperswitch/pull/889#discussion_r1168201823).

To summarize, a `Secret<Data, Strategy>` by definition differentiates between data and a masking strategy, therefore a type (like `Email` here) that both, stores data and implements `Strategy<T>` is unintended.
Also even though `Email` implements `Strategy<T>` here, it's never used to mask the actual email in the definition (`struct Email(Secret<String>)`, no strategy provided to the secret). My suggestion would be to create a separate unit struct, implement `Strategy<T>` on said struct, and define this struct type as a strategy to the secret inside `Email`.

vspecky (changes requested):

crates/common_utils/src/pii.rs
nachiketkanore: Done

crates/common_utils/src/pii.rs
```diff
@@ -87,26 +99,77 @@ where
...
 
-/// Email address
+/// Strategy for masking Email
 #[derive(Debug)]
-pub struct Email;
+pub struct EmailStrategy;
```
SanchithHegde, line 104: You should be able to make this private, since no other module would need it.

```suggestion
struct EmailStrategy;
```

SanchithHegde (commented):

Also, please resolve conflicts with the `main` branch and ensure that CI checks pass.

crates/common_utils/src/pii.rs
nachiketkanore: `Queryable` requires EmailStrategy to be public, since it is inside public interface
> error[E0446]: private type `EmailStrategy` in public interface

SanchithHegde: Alright then, you can ignore my comment. Marking this conversation resolved.

nachiketkanore:
Done

crates/common_utils/src/pii.rs
```diff
@@ -87,26 +99,77 @@ where
...
 
-        if is_valid.is_err() {
-            return WithType::fmt(val, f);
+        match Email::from_str(val_str) {
+            Ok(_email) => match val_str.split_once('@') {
+                Some((a, b)) => write!(f, "{}@{}", "*".repeat(a.len()), b),
+                None => WithType::fmt(val, f),
+            },
+            Err(..) => WithType::fmt(val, f),
         }
```
SanchithHegde, line 119:
```suggestion
        match val_str.split_once('@') {
            Some((a, b)) => write!(f, "{}@{}", "*".repeat(a.len()), b),
            None => WithType::fmt(val, f),
        },
```

We won't need to perform the validation here when masking emails.

crates/router/src/core/customers.rs
```diff
@@ -200,7 +201,10 @@ pub async fn delete_customer(
...
```
SanchithHegde, line 207:
```suggestion
        email: Email::from_str(REDACTED).ok(),
```

crates/router/src/types/api/customers.rs
SanchithHegde: @nachiketkanore Could you remove the commented code?

SanchithHegde (commented):
Everything else looks good!

nachiketkanore:
Done
Made suggested changes

SanchithHegde (approved):
Looks good to me!

vspecky (dismissed):

jagan-jaya (dismissed):

jarnura (approved):

PR #822, opened 2023-03-31
feat(domain): add domain types and add support for Encryption for fields

NishantJoshi00:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [x] Refactoring
- [ ] Dependency updates

## Description

This PR is the drain for all the PRs that would be related to domain types. The PR will include the following changes
1. Add domain types & data conversations
2. Refactor `router` to incorporate domain types, for current scope the incorporation will be limited to storage model conversions. (This will be divided into 2 PRs)
	1. #823 
4. Adding types and behaviors for Encryption : #827 
5. Incorporate Encryption in the existing data, will a need of migrations. (This will be further divided into multiple sub PRs based size or tables) 

### Additional Changes

## Checklist


NishantJoshi00:
Hey folks, this PR is introducing a BREAKING CHANGE, and the merge ready stage of the PR touches a lot of areas. Please review areas that it touches and comment if any of the code that you added was removed during the merge process.
- disputes: @sai-harsha-vardhan 
- frm configurations: @SamraatBansal 
- flows: @Narayanbhat166 
- core

crates/router/src/core/admin.rs
```diff
@@ -212,6 +266,19 @@ pub async fn merchant_account_update(
...
+            .await
+            .transpose()
+    };
+
+    let encrypt_value = |inner: OptionalSecretValue| async {
+        inner
+            .async_map(|value| domain_types::encrypt(value, &key))
+            .await
+            .transpose()
+    };
```
Narayanbhat166, line 281: This seems to be repeated. can be moved to some utils

NishantJoshi00: This cannot be moved into a function of its own. the reason being
- The key is being moved from the current context into the closure
- The the only solution that is left is: we create a function that looks something like this:
```rust
fn encrypt_closure_maker<T: Clone, S: Strategy<T>>(key: Vec<u8>) -> 
	(impl Fn(Option<Secret<T, S>>)) -> 
		(impl futures::Future<Output = 
			CustomResult<Option<Encryptable<Secret<T, S>>>, errors::CryptoError>>
		)
	)
where
	Secret<T, S>: Send
```
These closure that this function will yield cannot be generic as they themselves had uncertain fields or returns. this only possible way is to do `impl ...`. But we cannot have and `impl` inside an `impl` statement. Thus, for now we can only work with the manual closures that we are creating here.

dracarys18:
Reopened in #1043

PR #828, opened 2023-04-03
feat(connector): [Shift4] add support for card 3DS payment

jagan-jaya:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [X] Enhancement
- [ ] Refactoring
- [ ] Dependency updates

## Description

Add support for shift4 Card 3DS

Explanation for Adding Core change: 
1. For a 3DS transaction , we need to make a /3d-authenticate call to shift4 , which will return the `enrollment status` of the card along with `redirect_url` and `card_token`.
2. If the card is enrolled for 3DS,
```{
    "enrolled": true,
    "version": "2.2.0",
    "redirectUrl": "https://api.securionpay.com/3d-secure/start/L5DVhJTGO8XawEhptqGp16KrTAV50gFVJBeg9s1a",
    "token": {
        "id": "tok_csSIl6EWrBy3LHIAJ7iDDov4"
       }
     }
```
We should redirect the customer to the given redirect_url and complete the payment in `/complete` call. on complete call only we will make an actual transaction. So to achieve this scenario, we will not call `/authorize` flow after `/init`and we return the redirection response from `init` flow directly as response. Now the transaction will be in `authentication_pending` state with no `connector_transaction_id`.  After the `/complete` call the status of the transaction and connector_transaction_id will be updated.

3. If the card is not enrolled for 3DS,
```{
    "enrolled": false,
    "token": {
        "id": "tok_olPomCZs72KsMvyMYJ7Ja1z7"
         }
      }
```
we can proceed with the authorize call and make a Non 3DS transaction , it will follow the normal flow

### Additional Changes

## How did you test it?

Manual testing via postman and verified that all previous tests are working fine
Screen Shot 2023-04-03 at 3 48 40 PM

## Checklist


crates/router/src/connector/shift4/transformers.rs
```diff
@@ -290,13 +408,73 @@ pub enum NextAction {
...
+                enums::AttemptStatus::AuthenticationPending
+            } else {
+                enums::AttemptStatus::Pending
+            },
+            request: types::PaymentsAuthorizeData {
+                enrolled_for_3ds: item.response.enrolled,
+                ..item.data.request
+            },
+            response: Ok(types::PaymentsResponseData::TransactionResponse {
+                resource_id: types::ResponseId::NoResponseId,
```
Narayanbhat166, line 450: How can we track this payment without any id?

Narayanbhat166 (approved):
LGTM.

crates/router/src/connector/shift4/transformers.rs
```diff
@@ -87,49 +125,94 @@ impl TryFrom<&types::PaymentsAuthorizeRouterData> for Shift4PaymentsRequest {
...
 }
 
-fn get_bank_redirect_request(
-    item: &types::PaymentsAuthorizeRouterData,
+fn get_bank_redirect_request<T>(
+    item: &types::RouterData<T, types::PaymentsAuthorizeData, types::PaymentsResponseData>,
     redirect_data: &payments::BankRedirectData,
 ) -> Result<Shift4PaymentsRequest, Error> {
-    let submit_for_settlement = submit_for_settlement(item);
+    let submit_for_settlement = item.request.is_auto_capture();
```
ArjunKarthik, line 197: We have 4 CaptureMethods `Automatic, Manual, ManualMultiple, Scheduled,`. 
Implementation for each capture type may differ based on connector or they may not support some capture type. Here because of `request.is_auto_capture()` method we are considering whether it is `Automatic` or `Manual`, we are ignoring the other 2 capture types (Even though right now we don't have support for these 2, future this will requires lot of refactoring). So instead of this why Can't we just have case condition for `Automatic`, `Manual` rest all others throw FlowNotSupported?

crates/router/src/services/api.rs
```diff
@@ -312,7 +313,13 @@ async fn send_request(
                 // Currently this is not used remove this if not required
                 // If using this then handle the serde_part
                 Some(ContentType::FormUrlEncoded) => {
-                    let url_encoded_payload = serde_urlencoded::to_string(&request.payload)
+                    let payload = match request.payload.clone() {
+                        Some(req) => serde_json::from_str(req.peek())
```
ArjunKarthik, line 317: is this conversion is really necessary? What happens if we directly try `serde_urlencoded::to_string(req)` incase of some and serde_urlencoded:: to_string(Secret::new("")) incase of None? 

If we can directly convert the Secret<String> to url_encoded string we can skip the `peek` and `serde_json::from_str` part

crates/router/src/connector/shift4/transformers.rs
jagan-jaya: After the `/complete` call the status of the transaction and connector_transaction_id will be updated.

crates/router/src/connector/shift4.rs
```diff
@@ -107,14 +119,127 @@ impl ConnectorIntegration<api::Verify, types::VerifyRequestData, types::Payments
...
+                .body(types::PaymentsAuthorizeType::get_request_body(self, req)?)
+                .build(),
+        ))
+    }
+    fn handle_response(
+        &self,
+        data: &types::PaymentsAuthorizeRouterData,
+        res: types::Response,
+    ) -> CustomResult<types::PaymentsAuthorizeRouterData, errors::ConnectorError> {
+        let response: shift4::Shift4NonThreeDsResponse = res
```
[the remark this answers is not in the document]
jagan-jaya: yes, we don't it will be covered during complete_authorize call

crates/router/src/services/api.rs
jagan-jaya:
```2023-04-05T10:53:49.131754Z ERROR router::services::api: error: {"error":{"type":"api","message":"Something went wrong","code":"HE_00"}}
├╴at crates/router/src/services/api.rs:517:42
│
├─▶ {"error":{"type":"server_not_available","code":"HE_00","message":"Something went wrong"}}
│   ╰╴at crates/router/src/core/payments/flows/authorize_flow.rs:102:44
│
├─▶ Failed to execute a processing step: None
│   ╰╴at crates/router/src/services/api.rs:266:30
│
├─▶ URL encoding of request payload failed
│   ├╴at crates/router/src/services/api.rs:317:26
│   ╰╴Unable to do url encoding on request: Some(*** alloc::string::String ***)
│
╰─▶ top-level serializer supports only maps and structs```
You will get above error saying maps and struct only can be urlencoded. That's why we are converting it to JSON first and then url_encode

File: crates/router/src/connector/shift4/transformers.rs (hunk 2 above)
Comment (jagan-jaya) [reply crates/router/src/connector/shift4/transformers.rs:197]: When you want to check other type add one more method to it like `is_manual_multiple` so instead of writing a switch case in each connector like how it is done here, we can use the same logic in all the places. i don't understand why you are mentioning `will requires lot of refactoring`

Comment (jagan-jaya) [reply crates/router/src/connector/shift4/transformers.rs:197]: When you want to check other type add one more method to it like `is_manual_multiple` so instead of writing a switch case in each connector like how it is done here, we can use the same logic in all the places. i don't understand why you are mentioning `will requires lot of refactoring`

Comment (ArjunKarthik) [reply crates/router/src/connector/shift4/transformers.rs:197]: First concern here is if we receive `CaptureMethods` other that `Automatic ` or `Manual ` we should propagate error. Instead you are allowing the payment .
Secondly I don't agree with keep on adding method for each CaptureMethods. As to decide the CaptureMethods we have to call all the methods like `is_auto_capture`, `is_manual_multiple`, `is_scheduled`. Which is nothing but the case matching like how we are doing for PaymentMethods)
Yes refactor would be required for the connectors which uses `is_auto_capture` method when we start to support new CaptureMethods. But we can skip that refactor if we handle `CaptureMethods ` in match statement within every connector code.

Comment (jagan-jaya) [reply crates/router/src/connector/shift4/transformers.rs:197]: Currently we have 
```
fn is_auto_capture(&self) -> bool {
        matches!(
            self.capture_method,
            Some(storage_models::enums::CaptureMethod::Automatic) | None
        )
    }
```
Tell me how this will change in future if we have one more new `Capture Method`
@ArjunKarthik

File: crates/router/src/connector/shift4.rs [hunk 5]
```diff
@@ -258,22 +381,113 @@ impl ConnectorIntegration<api::Capture, types::PaymentsCaptureData, types::Payme
...
+    fn build_request(
+        &self,
+        req: &types::PaymentsInitRouterData,
+        connectors: &settings::Connectors,
+    ) -> CustomResult<Option<services::Request>, errors::ConnectorError> {
+        Ok(Some(
+            services::RequestBuilder::new()
+                .method(services::Method::Post)
+                .url(&types::PaymentsInitType::get_url(self, req, connectors)?)
+                .content_type(ContentType::FormUrlEncoded)
```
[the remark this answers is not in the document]
Narayanbhat166: Okay. Yes if it returns an enum, we can create the request based on the enum type rather than manually creating the request string by calling `encode_to` methods.

crates/router/src/connector/shift4/transformers.rs
ArjunKarthik: Even now we have 4 values for CaptureMethod enum,

> pub enum CaptureMethod {
    #[default]
    Automatic,
    Manual,
    ManualMultiple,
    Scheduled,
}

jagan-jaya: Yes , even though it has 4 capture methods, it did not affect any other methods and served the auto_capture check too. so why should there be a change in future is required?
```
fn is_auto_capture(&self) -> bool {
        matches!(
            self.capture_method,
            Some(storage_models::enums::CaptureMethod::Automatic) | None
        )
    }
 ```

**Review by Narayanbhat166 — APPROVED** [review]

File: crates/router/src/connector/utils.rs [hunk 6]
```diff
@@ -166,14 +170,32 @@ impl PaymentsAuthorizeRequestData for types::PaymentsAuthorizeData {
     }
 }
 
+pub trait PaymentsCompleteAuthorizeRequestData {
+    fn is_auto_capture(&self) -> Result<bool, Error>;
+}
+
+impl PaymentsCompleteAuthorizeRequestData for types::CompleteAuthorizeData {
```
jarnura, line 177:
```suggestion
impl types::CompleteAuthorizeData {
```

we don't require `PaymentsCompleteAuthorizeRequestData` trait

crates/router/src/connector/shift4/transformers.rs
```diff
@@ -68,15 +95,26 @@ pub struct DeviceData;
...
+    pub exp_month: Secret<String>,
+    pub exp_year: Secret<String>,
+    pub cardholder_name: Secret<String>,
 }
 
-impl TryFrom<&types::PaymentsAuthorizeRouterData> for Shift4PaymentsRequest {
+#[derive(Debug, Serialize, Eq, PartialEq)]
+#[serde(untagged)]
+pub enum CardPayment {
+    RawCard(Card),
```
jarnura, line 107: Use Box for struct when it is used as a value in Enum's variant

jarnura (changes requested):

ArjunKarthik (dismissed):

ArjunKarthik (dismissed):

crates/router/src/connector/shift4/transformers.rs
```diff
@@ -268,6 +349,45 @@ pub struct Shift4PaymentsResponse {
...
+
+#[derive(Default, Debug, Deserialize)]
+pub struct Token {
+    pub id: String,
+    pub created: i64,
+    #[serde(rename = "objectType")]
+    pub object_type: String,
+    pub first6: String,
+    pub last4: String,
+    pub fingerprint: String,
```
jarnura, line 369: Make this field as secret

jarnura (changes requested):
Other Secret related comments, LGTM

Narayanbhat166 (dismissed):

Narayanbhat166 (dismissed):

jarnura (dismissed):

jarnura (approved):

ArjunKarthik (approved):

PR #836, opened 2023-04-04
feat: allow (de)serializing countries to/from alpha-2, alpha-3 and numeric country codes

### LINKED ISSUE #804: [FEATURE]: add support for three letter and numeric country codes
### Feature Description

We are currently using two_letter country codes internally since majority of connectors accept only two_letter codes, there may arise a situation where some connector will accept only three_letter country codes or numeric codes. There is a complete list of the mapping https://www.nationsonline.org/oneworld/country_code_list.htm. 

### Possible Implementation

We can have an enum `Country` which will be the actual country name and then provide mapping to two_letter, three_letter codes and even numeric_codes. All the variants map to same enum `Country`. This can be achieved by implementing a custom serializer and deserializer as shown above.

```rust
#[derive(Clone, Copy, Debug)]
pub enum Country {
    Afghanistan,
    Albania,
    // ...
}

#[derive(Clone, Copy, Debug)]
pub enum Alpha2CountryCode {
    AF,
    AL,
    // ...
}

impl Country {
    // Note the `const`. Similar implementation for `to_alpha2(&self)`.
    pub const fn from_alpha2(code: Alpha2CountryCode) -> Self {
        match code {
            Alpha2CountryCode::AF => Self::Afghanistan,
            Alpha2CountryCode::AL => Self::Albania,
            // ...
        }
    }
}

mod custom_serde {
    use super::*;

    pub mod alpha2_country_code {
        use super::*;

        pub fn serialize<S>(code: &Country, serializer: S) -> Result<S::Ok, S::Error>
        where
            S: serde::Serializer,
        {
            // Serialize country as Alpha-2 code
            Err(serde::ser::Error::custom("not implemented"))
        }

        pub fn deserialize<'a, D>(deserializer: D) -> Result<Country, D::Error>
        where
            D: serde::Deserializer<'a>,
        {   
            // Deserialize Alpha-2 code from string, then convert to country enum
            Err(serde::de::Error::custom("not implemented"))
        }
    }
}

#[derive(serde::Deserialize, serde::Serialize)]
struct Address {
    #[serde(with = "custom_serde::alpha2_country_code")]
    country: Country,
}
```

nachiketkanore:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [x] Enhancement
- [x] Refactoring
- [x] Dependency updates

## Description
fixes #804 

### Additional Changes

## How did you test it?

Added multiple test cases covering maximum scenarios
Tested with `cargo test` with all OK passed

## Checklist


Narayanbhat166 (approved):
Looks good to me. Awesome work!

nachiketkanore:
Any specific reason these tests are failing? 
My changes seem unrelated to them (except formatting)

crates/common_enums/src/enums.rs
```diff
@@ -37,3 +38,2097 @@ pub enum CountryCode {
...
+            92 => Self::VirginIslandsBritish,
+            850 => Self::VirginIslandsUS,
+            876 => Self::WallisandFutuna,
+            732 => Self::WesternSahara,
+            887 => Self::Yemen,
+            894 => Self::Zambia,
+            716 => Self::Zimbabwe,
+            _ => {
+                panic!()
+            }
```
SanchithHegde, line 1593: We can convert this function to return a `Result` instead and error out if an unknown country code is provided. But panicking and terminating a payments related system only because of an invalid input would not be a good experience for the users. They'd have no idea if their payments actually went through, for example.

crates/common_enums/src/enums.rs
```diff
@@ -37,3 +38,2097 @@ pub enum CountryCode {
...
+    Tunisia,
+    Turkey,
+    Turkmenistan,
+    TurksandCaicosIslands,
+    Tuvalu,
+    Uganda,
+    Ukraine,
+    UnitedArabEmirates,
+    UnitedKingdomofGreatBritainandNorthernIreland,
+    UnitedStatesofAmerica,
```
SanchithHegde, line 311:
```suggestion
    UnitedStatesOfAmerica,
```

We'd prefer if the names of countries followed `PascalCase` (or `UpperCamelCase`) for all words. Please follow this for words like "of" and "and" (and anything I may have left out) as well, so these would be `Of` and `And` respectively. Please update other country names to follow this naming convention as well.

crates/common_enums/src/enums.rs
```diff
@@ -37,3 +38,2097 @@ pub enum CountryCode {
...
+struct HyperswitchRequestAlpha3 {
+    #[serde(with = "custom_serde::alpha3_country_code")]
+    pub country: Country,
+}
+
+#[derive(serde::Serialize, serde::Deserialize)]
+struct HyperswitchRequestNumeric {
+    #[serde(with = "custom_serde::numeric_country_code")]
+    pub country: Country,
+}
```
SanchithHegde, line 1990: All these types could go in the `tests` module? I don't see them being used anywhere else in your PR.

crates/common_enums/Cargo.toml
```diff
@@ -13,3 +13,4 @@ diesel = { version = "2.0.3", features = ["postgres"] }
 
 # First party crates
 router_derive = { version = "0.1.0", path = "../router_derive" }
+serde_json = "1.0.95"
```
SanchithHegde, line 16: You should be able to move `serde_json` under the `[dev-dependencies]` table, since it's only used in tests as of now.

> Any specific reason these tests are failing?
> My changes seem unrelated to them (except formatting)

It looks like you ran a `cargo update` (or something else triggered it), which has bumped multiple dependencies and tests are breaking. You could try reverting the `Cargo.lock` to an earlier commit, move `serde_json` to dev-deps and run `cargo check`. That should update the `Cargo.lock` to a stable state.

SanchithHegde (commented):
Other than that, looks good to me! Thanks for your contribution by the way!

crates/common_enums/src/enums.rs
Narayanbhat166: Yes, please remove the `panic!`, as suggested return a `Result`

nachiketkanore:
Suggested changes done

crates/common_enums/src/enums.rs
```diff
@@ -1943,55 +1944,51 @@ mod custom_serde {
         where
             D: serde::Deserializer<'a>,
         {
-            u32::deserialize(deserializer).map(Country::from_numeric)
+            u32::deserialize(deserializer).map(|value| Country::from_numeric(value).unwrap())
```
SanchithHegde, line 1947:
```suggestion
            u32::deserialize(deserializer).map(Country::from_numeric).map_err(serde::de::Error::custom)
```

crates/common_enums/src/enums.rs
```diff
@@ -279,64 +279,67 @@ pub enum Country {
...
-    WallisandFutuna,
+    WallisAndFutuna,
     WesternSahara,
     Yemen,
     Zambia,
     Zimbabwe,
 }
 
+#[derive(Debug)]
+pub struct NumericCountryCodeParseError;
```
SanchithHegde, line 328: Please implement `Display` for `NumericCountryCodeParseError` so that a useful error message is displayed whenever this error is logged.

SanchithHegde (commented):

crates/common_enums/src/enums.rs
nachiketkanore: This won't work because `Country::from_numeric` returns a Result<Country, _>

expected enum `Result<enums::Country, <D as Deserializer<'a>>::Error>`
found enum `Result<Result<enums::Country, NumericCountryCodeParseError>, _>`

Narayanbhat166: use `and_then` instead of map
```suggestion
u32::deserialize(deserializer)
                .and_then(|value| Country::from_numeric(value).map_err(serde::de::Error::custom));
```

nachiketkanore: Actually, we are getting 2 types of errors (NumericParseError and deserialization Error). 
So, they cannot be directly mapped to custom error

Narayanbhat166 (dismissed):
Looks Good to me!

SanchithHegde (approved):

PR #847, opened 2023-04-09
refactor(router_env): improve logging setup

SanchithHegde:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [x] Refactoring
- [ ] Dependency updates

## Description

This PR improves the module responsible for setting up the logging system. It includes the following changes:

- Makes the implicitly accessed environment variables explicit. They all now reside in the `router_env::env::vars` module.
- Extracts the OpenTelemetry tracing pipeline setup to a different function.
- Makes the OpenTelemetry metrics pipeline setup conditional, it is now run only if `log.telemetry.enabled` is `true`.
- Simplifies the file path construction for file logging.
- Build script now sets the `CARGO_WORKSPACE_MEMBERS` environment variable with a comma-separated list of all crates in the current cargo workspace. This is used to enable verbose logging for all workspace members by default.
- Uses `Targets` only for filtering crates/targets to log and removes usages of `EnvFilter` for filtering.
- Uses `RUST_FILE_LOG` and `RUST_CONSOLE_LOG` environment variables to override `Targets` directives for file and console logging respectively, instead of `RUST_LOG`.

### Additional Changes

The `RUST_LOG` environment variable is no longer accessed by the application, the `ROUTER__LOG__CONSOLE__LEVEL` or `ROUTER__LOG__FILE__LEVEL` environment variable should be used to set the logging level instead, and if advanced target filtering is required, the `RUST_CONSOLE_LOG` or `RUST_FILE_LOG` environment variables must be used instead.

## Motivation and Context

My initial motivation was to allow overriding the `Targets` filtering directive at runtime, so that better debugging is possible at runtime, if required. Since I had anyway ventured into the rabbit hole, I ended up refactoring parts of it.

## How did you test it?

Manually.

## Checklist

### DIFF
--- a/crates/router_env/build.rs
+++ b/crates/router_env/build.rs
@@ -1,0 +1,24 @@
+include!("src/vergen.rs");
+
+fn main() {
+    generate_cargo_instructions();
+
+    #[allow(clippy::expect_used)] // Safety: panicking in build scripts is okay for us
+    let metadata = cargo_metadata::MetadataCommand::new()
+        .exec()
+        .expect("Failed to obtain cargo metadata");
+    let workspace_members = metadata
+        .workspace_members
+        .iter()
+        .map(|package_id| {
+            #[allow(clippy::expect_used)] // Safety: panicking in build scripts is okay for us
+            package_id
+                .repr
+                .split_once(' ')
+                .expect("Unknown cargo metadata package ID format")
+                .0
+        })
+        .collect::<Vec<_>>()
+        .join(",");
+    println!("cargo:rustc-env=CARGO_WORKSPACE_MEMBERS={workspace_members}",);
+}
--- a/crates/router_env/src/env.rs
+++ b/crates/router_env/src/env.rs
@@ -1,21 +1,16 @@
-//!
-//! Current environment related stuff.
-//!
+//! Information about the current environment.
 
 use std::path::PathBuf;
 
 use serde::{Deserialize, Serialize};
 use strum::{Display, EnumString};
 
-/// Parent dir where Cargo.toml is stored
+/// Parent directory where `Cargo.toml` is stored.
 pub const CARGO_MANIFEST_DIR: &str = "CARGO_MANIFEST_DIR";
-/// Env variable that sets Development/Production env
+/// Environment variable that sets Development/Sandbox/Production environment.
 pub const RUN_ENV: &str = "RUN_ENV";
 
-///
 /// Current environment.
-///
-
 #[derive(Debug, Default, Deserialize, Serialize, Clone, Copy, Display, EnumString)]
 pub enum Env {
     /// Development environment.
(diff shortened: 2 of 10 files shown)


crates/router_env/src/env.rs
```diff
@@ -1,21 +1,36 @@
...
+    pub const RUST_CONSOLE_LOG: &str = "RUST_CONSOLE_LOG";
+
+    /// Environment variable that sets the file logging level.
+    pub const RUST_FILE_LOG: &str = "RUST_FILE_LOG";
+
+    /// Target to which the OpenTelemetry exporter is going to send signals.
+    pub const OTEL_EXPORTER_OTLP_ENDPOINT: &str = opentelemetry_otlp::OTEL_EXPORTER_OTLP_ENDPOINT;
+
+    /// Max waiting time for the OpenTelemetry backend to process each signal batch.
+    pub const OTEL_EXPORTER_OTLP_TIMEOUT: &str = opentelemetry_otlp::OTEL_EXPORTER_OTLP_TIMEOUT;
```
SanchithHegde, line 30: For now, I've kept them as environment variables independent of our application configuration. I can update the code to pick these up from application configuration instead (then the `ROUTER__LOG__*` environment variables would have to be set instead), if required.

crates/router_env/build.rs
```diff
@@ -0,0 +1,24 @@
...
+            #[allow(clippy::expect_used)] // Safety: panicking in build scripts is okay for us
+            package_id
+                .repr
+                .split_once(' ')
+                .expect("Unknown cargo metadata package ID format")
+                .0
+        })
+        .collect::<Vec<_>>()
+        .join(",");
+    println!("cargo:rustc-env=CARGO_WORKSPACE_MEMBERS={workspace_members}",);
```
jarnura, line 23: Just add assert to test the format.

jarnura (commented):
Use Configs to fetch env variables and don't use absolute path in function calls

ashokkjag:
LGTM

crates/router_env/src/logger/setup.rs
```diff
@@ -12,150 +10,209 @@ use opentelemetry::{
...
+    default_log_level: config::Level,
+    filter_log_level: config::Level,
+    crates_to_filter: impl AsRef<[&'static str]>,
+) -> Targets {
+    filtering_directive
+        .and_then(|filter| {
+            // Try to create target filter from specified filtering directive, if set
+            Targets::from_str(filter)
+                .map_err(|error| {
+                    eprintln!("Invalid target filtering directive `{filter}`: {error}")
```
SanchithHegde, line 199: Should we panic here instead?

crates/router_env/src/env.rs
SanchithHegde: Done, moved to application configuration.

crates/router_env/src/logger/setup.rs
SanchithHegde: No longer applicable, since I switched to using `EnvFilter` instead.

jarnura (approved):

PR #831, opened 2023-04-03
added field frm_configs to MerchantConnectorAccount

rishavkar:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates

## Description

We need to accept frm_configs from the dashboard. So needed to do development for accepting new fields for endpoint: `/account/:account_id/connectors`

### Additional Changes

- [x] This PR modifies the database schema : 
                    `ALTER TABLE "merchant_connector_account" ADD COLUMN frm_configs jsonb;`
- [ ] This PR modifies application configuration/environment variables

## Motivation and Context

It a new feature...we are going to integrate FRM services into hyperswitch, for which we required additional fields for create connector request. 
This PR will accept frm_configs field for create connector request, and accordingly persist in the db.

## How did you test it?

Following are the steps:
- Adding an extra field to the request body, to existing curl.
- Getting 200 and Expecting this same body in response.
  ```
"frm_configs" : {
      "frm_enabled_pms" : ["card"],
      "frm_enabled_pm_types" : ["credit"],
      "frm_enabled_gateways" : ["stripe"],
      "frm_action": "cancel_txn",
      "frm_preferred_flow_type" : "pre"
  }
```

## Checklist

- [ ] I formatted the code `cargo +nightly fmt --all`
- [ ] I addressed lints thrown by `cargo clippy`
- [ ] I reviewed submitted code
- [ ] I added unit tests for my changes where possible
- [ ] I added a [CHANGELOG](/CHANGELOG.md) entry if applicable

PR #848, opened 2023-04-10
Revert "added field frm_configs to MerchantConnectorAccount"

jarnura:
Reverts juspay/hyperswitch#831

The frm_configs handled as a mandatory field in flows.


Narayanbhat166 (approved):

crates/router/src/types/transformers.rs
```diff
@@ -341,16 +341,6 @@ impl ForeignTryFrom<storage::MerchantConnectorAccount> for api_models::admin::Me
                 .change_context(errors::ApiErrorResponse::InternalServerError)?,
             None => None,
         };
-        let configs_for_frm_value = merchant_ca
```
jarnura, line 344: This place we are throwing error if the data is not found.

SanchithHegde (approved):

PR #856, opened 2023-04-11
feat(access_token): scheduler workflow implementation for access token

Narayanbhat166:
## Type of Change

- [x] New feature

## Description

This PR adds support for generating the access token by the scheduler. This will offload the task of generating the access token by a new payment and save us some time for a new payment.

## Motivation and Context

Reduce access token creation time during payments.

## How did you test it?

- For testing purposes, the next schedule time is set to 60 seconds.
- Create a payment using a connector which requires `access_token` for payment.
- The process tracker entry is created.
Screenshot 2023-04-11 at 4 12 09 PM

- After one minute, new access token is created and stored in redis. The task is requeued and new schedule time is created.
Screenshot 2023-04-11 at 4 14 58 PM

## Checklist


crates/router/src/core/payments/access_token.rs
```diff
@@ -160,10 +186,59 @@ pub async fn add_access_token<
     }
 }
 
+fn get_requeue_schedule_time(access_token_ttl: i64) -> time::PrimitiveDateTime {
+    // Scheduler to get a new access token 60 seconds before it expires
+    // If for some reason the ttl is negative, get the time until refresh to be 0 seconds
+    // The problem arises if the access_token_ttl is greater than 30 minutes in the opposite time duration
+    // as this will not be picked up the scheduler.
+
+    let access_token_ttl = access_token_ttl - 60;
```
jarnura, line 195: make this 60secs as a config can be served from env.

crates/router/src/core/payments/access_token.rs
```diff
@@ -196,3 +271,150 @@ pub async fn refresh_connector_auth(
...
+    let custom_merchant_mapping = db.find_config_by_key(pt_mapping_key).await?;
+
+    let connector_mapping = custom_merchant_mapping
+        .config
+        .parse_struct::<ConnectorPTMapping>("CustomPTMapping");
+
+    let mapping = match connector_mapping {
+        Ok(x) => x,
+        Err(err) => {
+            crate::logger::error!("Error: while getting connector mapping: {}", err);
```
jarnura, line 336: remove absolute path in function calls
```suggestion
            logger::error!("Error: while getting connector mapping: {}", err);
```

crates/router/src/core/payments/access_token.rs
```diff
@@ -196,3 +271,150 @@ pub async fn refresh_connector_auth(
...
+                    .change_context(errors::ApiErrorResponse::InternalServerError)
+                    .attach_printable("Redis error when setting the access token");
+
+                // Requeue the task for next scheduled time
+                let next_schedule_time = get_requeue_schedule_time(access_token.expires);
+                process.requeue(store, next_schedule_time).await?;
+            }
+            Err(error_response) => {
+                // Retry the task
+                crate::logger::error!(access_token_refresh_task_error=?error_response);
```
jarnura, line 399:
```suggestion
                logger::error!(access_token_refresh_task_error=?error_response);
```

jarnura (changes requested):

crates/router/src/core/payments/access_token.rs
```diff
@@ -196,3 +271,150 @@ pub async fn refresh_connector_auth(
...
+                // Update the access token in redis
+                let store = &*state.store;
+                let _ = state
+                    .store
+                    .set_access_token(
+                        &tracking_data.merchant_id,
+                        connector.connector.id(),
+                        access_token.clone(),
+                    )
+                    .await
```
Abhicodes-crypto, line 389: Can you add comments over here on why the error should not be propagated.

crates/router/src/core/payments/access_token.rs
```diff
@@ -121,30 +141,36 @@ pub async fn add_access_token<
...
-                            access_token.clone(),
+
+                refresh_connector_auth(state, connector, &refresh_token_router_data)
+                    .await?
+                    .async_map(|access_token| async {
+                        //Store the access token in db
+                        let store = &*state.store;
+                        // This error should not be propagated, we don't want payments to fail once we have
+                        // the access token, the next request will create new access token
+                        let _ = store
```
Abhicodes-crypto, line 152:
```suggestion
                        _ = store
```

Abhicodes-crypto: Similarly in other places also it can be replaced.

crates/router/src/core/payments/access_token.rs
Narayanbhat166: This flow will be on the critical path of `payments`. Once we have the access token, if saving it in the database (redis) fails, we should not fail the payments. The next payment will again create new access token.

Narayanbhat166: I have drafted this PR as of now, as there are some changes that have been merged into the main which will have conflicts with this change. Will pick this up later. I will add comments.

Abhicodes-crypto: Sure. Can you add the same comments in the codebase as well. Also if errors should not be thrown then why is there a fallible function on line 395.

Narayanbhat166:
Closing this PR since changes will be taken up in a separate PR. This has a lot of conflicts and better to copy over changes to a separate PR because a lot has changed since this change was made.

PR #857, opened 2023-04-11
Frm integration with hyperswitch

rishavkar:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates

## Description

We need to accept frm_configs from the dashboard. So needed to do development for accepting new fields for endpoint: /account/:account_id/connectors

### Additional Changes

- [x] This PR modifies the database schema : `ALTER TABLE "merchant_connector_account" ADD COLUMN frm_configs jsonb;`

## Motivation and Context

It a new feature...we are going to integrate FRM services into hyperswitch, for which we required additional fields for create connector request.
This PR will accept frm_configs field for create connector request, and accordingly persist in the db.

## How did you test it?

Following are the steps:

- Adding an extra field to the request body, to existing curl of the create connector request body.
- Getting 200 and Expecting this same body in response.
```
"frm_configs" : {
"frm_enabled_pms" : ["card"],
"frm_enabled_pm_types" : ["credit"],
"frm_enabled_gateways" : ["stripe"],
"frm_action": "cancel_txn",
"frm_preferred_flow_type" : "pre"
}
```
- Also this is an optional field...without this field also (previous request format), we will get a success 

## Checklist

## Assignee
- @shakthi-22 
- @AnandKGanesh


jarnura (changes requested):

crates/api_models/src/admin.rs
```diff
@@ -372,6 +372,8 @@ pub struct MerchantConnectorCreate {
     /// You can specify up to 50 keys, with key names up to 40 characters long and values up to 500 characters long. Metadata is useful for storing additional, structured information on an object.
     #[schema(value_type = Option<Object>,max_length = 255,example = json!({ "city": "NY", "unit": "245" }))]
     pub metadata: Option<pii::SecretSerdeValue>,
+    /// contains the frm configs for the merchant... it should be of this format:: "\"frm_configs\" : { \"frm_enabled_pms\" : [\"card\"], \"frm_enabled_pm_types\" : [\"credit\"], \"frm_enabled_gateways\" : [\"stripe\"], \"frm_action\": \"cancel_txn\", \"frm_preferred_flow_type\" : \"pre\" }"
+    pub frm_configs: Option<FrmConfigs>,
```
jarnura, line 376: Add schema with example, refer the payment_methods_enabled key in the same struct. Remove format from comments and move to schema

jarnura (changes requested):

jarnura (approved):

SamraatBansal (approved):

crates/storage_models/src/merchant_connector_account.rs
```diff
@@ -28,6 +28,7 @@ pub struct MerchantConnectorAccount {
     pub payment_methods_enabled: Option<Vec<serde_json::Value>>,
     pub connector_type: storage_enums::ConnectorType,
     pub metadata: Option<pii::SecretSerdeValue>,
+    pub frm_configs: Option<Secret<serde_json::Value>>, //Option<FrmConfigs>
```
SanchithHegde, line 31: Hey @rishavkar, I understand this is a PR that's already merged, but wanted to bring this to your attention.

Whenever you're adding fields to a struct corresponding to a database table, please add it at the last, so that it is in correspondence with how the database has it. When we do a `diesel migration run`, Postgres adds the column to the end of the table, and `diesel` updates the `schema.rs` file to match the database table state, thus adding `frm_configs` to the end. Because of how `diesel` handles queries, it expects the column order in `schema.rs` and the field order in the struct here to be the same. If that is not the case, we'd end up having unexpected behavior (such as data for one column being populated in another) when running our application.

PR #871, opened 2023-04-13
fix: Update events table after notifying merchant

SamraatBansal:
## Type of Change

- [X] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates

## Description

Update the `is_webhook_notified` column in `events` table after an outgoing Webhook notification is successfully triggered.

### Additional Changes

## Motivation and Context

After triggering Webhook notification, db was not getting updated to record that a successful notification has been delivered to the merchant.

## How did you test it?

Triggered Webhooks and accordingly based on outgoing Webhook notification, observed the changes in `events` table.

Screenshot 2023-04-13 at 12 24 40 PM

## Checklist

### DIFF
--- a/crates/router/src/core/webhooks.rs
+++ b/crates/router/src/core/webhooks.rs
@@ -439,7 +439,9 @@ async fn trigger_webhook_to_merchant<W: api::OutgoingWebhookType>(
         .change_context(errors::WebhooksFlowError::MerchantWebhookURLNotConfigured)
         .map(ExposeInterface::expose)?;
 
-    let transformed_outgoing_webhook = W::from(webhook.clone());
+    let outgoing_webhook_event_id = webhook.event_id.clone();
+
+    let transformed_outgoing_webhook = W::from(webhook);
 
     let response = reqwest::Client::new()
         .post(&webhook_url)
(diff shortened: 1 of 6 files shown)


crates/storage_models/src/events.rs
```diff
@@ -18,6 +18,27 @@ pub struct EventNew {
...
 
+#[derive(Debug)]
+pub struct EventUpdate {
+    pub event_type: Option<storage_enums::EventType>,
+    pub event_class: Option<storage_enums::EventClass>,
+    pub is_webhook_notified: Option<bool>,
+    pub intent_reference_id: Option<String>,
+    pub primary_object_id: Option<String>,
+    pub primary_object_type: Option<storage_enums::EventObjectType>,
+}
```
vspecky, line 29: This should be an enum where the variants are the different types of updates that need to be done and the corresponding fields that need to be updated. refer to other tables.
```suggestion
pub enum EventUpdate {
    NotifyUpdate {
        is_webhook_notified: bool,
    }
}
```

crates/router/src/core/webhooks.rs
```diff
@@ -459,9 +459,21 @@ async fn trigger_webhook_to_merchant<W: api::OutgoingWebhookType>(
...
         }
         Ok(res) => {
+            let update_event = storage::EventUpdate {
+                event_type: None,
+                event_class: None,
+                is_webhook_notified: Some(true),
+                intent_reference_id: None,
+                primary_object_id: None,
+                primary_object_type: None,
+            };
```
vspecky, line 469: Move this inside the `else` branch. Also it would be more readable if we invert the if-else condition.

vspecky (changes requested):

crates/router/src/core/webhooks.rs
```diff
@@ -439,7 +439,7 @@ async fn trigger_webhook_to_merchant<W: api::OutgoingWebhookType>(
         .change_context(errors::WebhooksFlowError::MerchantWebhookURLNotConfigured)
         .map(ExposeInterface::expose)?;
 
-    let transformed_outgoing_webhook = W::from(webhook);
+    let transformed_outgoing_webhook = W::from(webhook.clone());
```
Narayanbhat166, line 442: rather than cloning the `webhook` object, clone the `event_id`

crates/router/src/core/webhooks.rs
```diff
@@ -459,9 +459,21 @@ async fn trigger_webhook_to_merchant<W: api::OutgoingWebhookType>(
                 .change_context(errors::WebhooksFlowError::CallToMerchantFailed)?;
         }
         Ok(res) => {
+            let update_event = storage::EventUpdate {
+                event_type: None,
+                event_class: None,
+                is_webhook_notified: Some(true),
+                intent_reference_id: None,
+                primary_object_id: None,
+                primary_object_type: None,
```
Narayanbhat166, line 468: If the fields which are passed as `None` are never to be updated, would be better to remove them from the update object.

crates/storage_models/src/events.rs
Narayanbhat166: Other fields are never to be updated, we can remove them

Narayanbhat166 (changes requested):

Narayanbhat166 (approved):

jarnura (approved):

vspecky (approved):
LGTM

PR #872, opened 2023-04-13
ci(manual_release): add `multiple_mca` feature in ci

Narayanbhat166:
## Type of Change

- [x] New feature

## Description

This will add support for choosing whether the `backwards_compatibility` feature when creating the manual release.

## Motivation and Context

To support `backwards_compatibility` feature in CI.

## Checklist

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,7 @@
 FROM rust:slim as builder
 
 ARG RUN_ENV=Sandbox
+ARG EXTRA_FEATURES=""
 
 RUN apt-get update \
     && apt-get install -y libpq-dev libssl-dev pkg-config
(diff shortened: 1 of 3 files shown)


jarnura (dismissed):

Dockerfile
```diff
@@ -34,7 +34,7 @@ ENV CARGO_REGISTRIES_CRATES_IO_PROTOCOL="sparse"
 COPY . .
 
 # Use bash variable substitution to convert environment name to lowercase
-RUN bash -c 'cargo build --release --features ${RUN_ENV@L}'
+RUN bash -c 'cargo build --release --features ${RUN_ENV@L} ${EXTRA_FEATURES@L}'
```
SanchithHegde, line 37: You don't need the lowercase transformation here, that was only required in case of `RUN_ENV` due to the way it was handled by the application to load appropriate configuration files.

You'd also need to add an `ARG EXTRA_FEATURES=""` line for the `builder` stage.

SanchithHegde (approved):

PR #874, opened 2023-04-13
feat(cards): validate card security code and expiration

### LINKED ISSUE #607: [FEATURE] Validate card security codes and expiration month/year
### Feature Description

As of now, we don't perform any validations on card security codes (CVC/CVV/CID code) and card expiration month and year. It'd be great if the following validations can be done:

1. Card security code must be 3 or 4 digits only
2. Expiration month is valid (1-12)
3. Expiration year is a 4 digit number only
4. Card has not expired (expiration month and year correspond to a date in the future)

This logic can reside in a new crate called `cards`. Please also include unit tests in your PR.

### Possible Implementation

For card security codes, you could do:

```rust
use masking::StrongSecret;

struct CardSecurityCode(StrongSecret<u16>);

let card_security_code = CardSecurityCode::try_from(123); // You can use the `TryFrom` trait for this purpose

struct CardExpirationMonth(StrongSecret<u8>);

let card_expiration_month = CardExpirationMonth::try_from(1); // You can use the `TryFrom` trait for this purpose

impl CardExpirationMonth {
    fn inner(&self) -> u8;
    fn two_digits(&self) -> String;
}

struct CardExpirationYear(StrongSecret<u16>);

let card_expiration_year = CardExpirationYear::try_from(2038); // You can use the `TryFrom` trait for this purpose

impl CardExpirationYear {
    fn inner(&self) -> u16;
    fn four_digits(&self) -> String;
    fn two_digits(&self) -> String;
}

struct CardExpiration {
    pub month: CardExpirationMonth,
    pub year: CardExpirationYear,
}

impl CardExpiration {
    fn is_expired(&self) -> bool;
}
```

You can implement `Serialize` and `Deserialize` traits on `CardSecurityCode`. Use the 2-digit and 4-digit representations for the month and year, respectively, for the `Serialize` and `Deserialize` implementations of the card expiry month and year. You can implement any more methods as necessary.

None

divinenaman:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates

## Description
Add basic validations for card security codes (CVC/CVV/CID code) and card expiration month and year. This logic resides in a new crate called cards. Validations included (as of now):
 - [x] Card security code must be 3 or 4 digits only
 - [x] Expiration month is valid (1-12)
 - [x] Expiration year is a 4 digit number only
 - [x] Card has not expired (expiration month and year correspond to a date in the future)

### Additional Changes

## Motivation and Context

This fixes #607 

## How did you test it?

run `cargo test --cards` 

## Checklist


crates/cards/src/lib.rs
```diff
@@ -0,0 +1,157 @@
...
+    }
+
+    pub fn two_digits(&self) -> String {
+        let month = self.0.peek();
+        let month_str: String = month.to_string();
+        if month_str.len() == 1 {
+            format!("0{}", month_str)
+        } else {
+            month_str
+        }
```
SanchithHegde, line 44:
```suggestion
        format!("{:02}", self.0.peek())
```

crates/cards/src/lib.rs
```diff
@@ -0,0 +1,157 @@
...
+    }
+}
+
+pub struct CardExpirationYear(StrongSecret<u16>);
+
+impl CardExpirationYear {
+    pub fn new(secret: StrongSecret<u16>) -> errors::CustomResult<Self, errors::ValidationError> {
+        let year = secret.peek();
+
+        if *year >= 1997 {
```
SanchithHegde, line 54: Is there a specific reason why this check is done against 1997? Won't it be better to check if it's `>= current_year`?

crates/cards/src/lib.rs
```diff
@@ -0,0 +1,157 @@
...
+            }))
+        }
+    }
+
+    pub fn four_digits(&self) -> String {
+        self.0.peek().to_string()
+    }
+
+    pub fn two_digits(&self) -> String {
+        let year = &self.0.peek().to_string()[2..4];
```
SanchithHegde, line 68:
```suggestion
        let year = self.0.peek() % 100;
```

crates/cards/src/lib.rs
```diff
@@ -0,0 +1,157 @@
...
+                match card_year {
+                    Ok(cy) => Ok(Self {
+                        month: cm,
+                        year: cy,
+                    }),
+                    Err(e) => Err(e),
+                }
+            }
+            Err(e) => Err(e),
+        }
```
SanchithHegde, line 98:
```suggestion
        let month = CardExpirationMonth::new(secret_month)?;
        let year = CardExpirationYear::new(secret_year)?;

        Ok(Self {
            month,
            year,
        })
```

You might have to run the formatter though.

crates/cards/src/lib.rs
```diff
@@ -0,0 +1,157 @@
...
+            i32::from(*self.year.0.peek()),
+            (*self.month.0.peek()).try_into().unwrap(),
+        );
+
+        let expiration_date = Date::from_calendar_date(
+            i32::from(*self.year.0.peek()),
+            (*self.month.0.peek()).try_into().unwrap(),
+            expiration_day,
+        )
+        .unwrap();
```
SanchithHegde, line 115: We preferably don't want to panic for invalid values.  

@NishantJoshi00 @vspecky Should we have this function return `Result<bool, std::convert::Infallible>`?

crates/cards/Cargo.toml
```diff
@@ -0,0 +1,14 @@
...
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+time = { version = "0.3.20", features = ["serde", "serde-well-known", "std"] }
+error-stack = "0.3.1"
+
+# First Party crates
+masking = { version = "0.1.0", path = "../masking" }
+common_utils = { version = "0.1.0", path = "../common_utils", features = ["signals", "async_ext"] }
```
SanchithHegde, line 14:
```suggestion
common_utils = { version = "0.1.0", path = "../common_utils" }
```

I don't think we need either of the features here.

SanchithHegde (commented):
Could you please move the logic in the constructors to the `TryFrom` trait implementations for `CardSecurityCode`, `CardExpirationMonth` and `CardExpirationYear`?

So:
1. `CardSecurityCode::new()` would become `CardSecurityCode::try_from(u16)`
2. `CardExpirationMonth::new()` would become `CardExpirationMonth::try_from(u8)`
3. `CardExpirationYear::new()` would become `CardExpirationYear::try_from(u16)`

And regarding the `Serialize` and `Deserialize` implementations, would you want to take them up in this PR or a separate one? Let us know if you have any questions.

crates/cards/src/lib.rs
NishantJoshi00: I see 2 approaches in which we can handle this:
1. We can how it return `Option<boo>` and then using `.ok()` to convert those `Result`s into `Option`s.
2. We can do the same thing but instead of return type being an option we can keep it as bool and do `unwrap_or(false)`

crates/cards/src/lib.rs
```diff
@@ -0,0 +1,157 @@
...
+    fn peek(&self) -> &StrongSecret<u8> {
+        &self.0
+    }
+}
+
+impl PeekInterface<StrongSecret<u16>> for CardExpirationYear {
+    fn peek(&self) -> &StrongSecret<u16> {
+        &self.0
+    }
+}
```
NishantJoshi00, line 157: Hey @divinenaman, implementing `PeekInterface<StrongSecret<u16>>` provides use access to all the methods that were present on `&StrongSecret<u16>`, from which the most prominent is `.peek()`, so instead of writing `.peek().peek()`, consider using [deref coercion](https://doc.rust-lang.org/std/ops/trait.Deref.html#more-on-deref-coercion)

example:
```rust
impl Deref for CardExpirationYear {
    type Target = StrongSecret<u16>;
    fn deref(&self) -> &Self::Target {
        &self.0
    }
}
```

As you can see it's not that different, but it definitely contributes to ease of use in future

for instance, in the tests that you have written below instead of doing `.get_year().peek().peek()` you can simply `.get_year().peek()` and it would still work!

crates/cards/src/lib.rs
vspecky: I'd prefer returning a `Result<bool, SomeError>` here. We cannot just unwrap the result since in certain cases, the date may be invalid (eg: `29th Feb 2025` which is invalid cuz 2025 is not a leap year)

That, or we perform this validation during construction (makes more sense) in the `new()` function. Then, given that all members are private, it should be fine to `.expect()` here.

cc: @NishantJoshi00

crates/cards/src/lib.rs
```diff
@@ -0,0 +1,157 @@
...
+
+        // card expiry day is last day of the expiration month
+        let expiration_day = days_in_year_month(
+            i32::from(*self.year.0.peek()),
+            (*self.month.0.peek()).try_into().unwrap(),
+        );
+
+        let expiration_date = Date::from_calendar_date(
+            i32::from(*self.year.0.peek()),
+            (*self.month.0.peek()).try_into().unwrap(),
```
vspecky, line 112: Prefer `.expect("panic message")` over `.unwrap()`

crates/cards/tests/basic.rs
```diff
@@ -0,0 +1,27 @@
...
+
+    let card_security_code = CardSecurityCode::new(csc).unwrap();
+    let card_expiration = CardExpiration::new(expiry_month, expiry_year).unwrap();
+
+    assert_eq!(*card_security_code.peek().peek(), 1234);
+    assert_eq!(*card_expiration.get_month().peek().peek(), 12);
+    assert_eq!(*card_expiration.get_year().peek().peek(), 2023);
+
+    Ok(())
+}
```
vspecky, line 27: This is an integration test. I'd suggest using unit tests in the crate using `#[cfg(test)]`. Also, add test cases for validations for assurance that the code will be panic free at runtime, eg:
- 31st April 20xx (error, April has only 30 days)
- 31st Feb 20xx (error, Feb has max 28/29 days)
- 29th Feb 2020 (Valid leap year)
- 29th Feb 2021 (error, 2021 is not a leap year)

Add more as you see fit.

vspecky (changes requested):
Also look into adding `serde::{Serialize, Deserialize}` implementations with appropriate error reporting. Thanks for taking this up!

crates/cards/tests/basic.rs
vspecky: Ignore this, I just saw that the `time` crate handles all of these cases for you! :)

crates/cards/src/lib.rs
vspecky: Okay so apparently the `time` crate handles cases like the one I mentioned above. In that case, I'm fine with an `.expect()` here. We can add a `time` sanity test that ensures any crate version bump does not break panic safety.

Also, for every `.expect()`, include a comment
```
// Panic Safety: reason why this will not panic
```

crates/cards/src/lib.rs
divinenaman: sure, I'll change that

crates/cards/src/lib.rs
divinenaman: sure, I'll look into that

divinenaman:
> Could you please move the logic in the constructors to the `TryFrom` trait implementations for `CardSecurityCode`, `CardExpirationMonth` and `CardExpirationYear`?
> 
> So:
> 
>     1. `CardSecurityCode::new()` would become `CardSecurityCode::try_from(u16)`
> 
>     2. `CardExpirationMonth::new()` would become `CardExpirationMonth::try_from(u8)`
> 
>     3. `CardExpirationYear::new()` would become `CardExpirationYear::try_from(u16)`

@SanchithHegde should I remove the new implementation and just implement a TryFrom instance or keep both ?

SanchithHegde:
> @SanchithHegde should I remove the new implementation and just implement a TryFrom instance or keep both ?

Yes, remove the `new()` method and have only a `TryFrom` implementation. `new()` would be suitable when the constructor is not expected to fail (doesn't return a `Result`), which clearly isn't the case here.

crates/cards/src/lib.rs
divinenaman: @vspecky @NishantJoshi00 can you check me if below implementation is as suggested 

```rust
pub fn is_expired(&self) -> bool {
        let current_datetime_utc = date_time::now();
        
        // Panic Safety: struct CardExpirationMonth validates month  
        let expiration_month = (*self.month.0.peek()).try_into().expect("valid month in range 1 to 12");

        let expiration_year = *self.year.0.peek();

        // card expiry day is last day of the expiration month
        let expiration_day = days_in_year_month(i32::from(expiration_year), expiration_month);

        // Panic Safety: year, month, day validated
        let expiration_date =
            Date::from_calendar_date(i32::from(expiration_year), expiration_month, expiration_day)
                .expect("valid date");

        let expiration_time = Time::MIDNIGHT;

        // actual expiry date specified on card w.r.t. local timezone
        // max diff b/w utc and other timezones is 14 hours
        let mut expiration_datetime_utc = PrimitiveDateTime::new(expiration_date, expiration_time);

        // compensating time difference b/w local and utc timezone by adding a day
        expiration_datetime_utc = expiration_datetime_utc.saturating_add(Duration::days(1));

        if current_datetime_utc > expiration_datetime_utc {
            true
        } else {
            false
        }
    }

```

crates/cards/Cargo.toml
```diff
@@ -8,7 +8,9 @@ edition = "2021"
 [dependencies]
 time = { version = "0.3.20", features = ["serde", "serde-well-known", "std"] }
 error-stack = "0.3.1"
+serde = { version = "1.0.155", features = ["derive"] }
+serde_json = "1.0.94"
```
SanchithHegde, line 12: I don't see any usages of these crates. Do you plan to use them in this PR?

crates/cards/src/lib.rs
```diff
@@ -0,0 +1,147 @@
...
+    pub fn two_digits(&self) -> String {
+        format!("{:02}", self.peek())
+    }
+}
+
+impl TryFrom<u8> for CardExpirationMonth {
+    type Error = error_stack::Report<errors::ValidationError>;
+    fn try_from(month: u8) -> Result<Self, Self::Error> {
+        if month >= 1 && month <= 12 {
+            Ok(Self(StrongSecret::<u8>::new(month)))
```
SanchithHegde, line 35:
```suggestion
            Ok(Self(StrongSecret::new(month)))
```

crates/cards/Cargo.toml
divinenaman: @SanchithHegde  Yes, how do I add serialization and deserialization support for the new structs ?

SanchithHegde: For serialization of tuple structs with a single field, you should be able to use [`#[serde(transparent)]`](https://serde.rs/container-attrs.html#transparent). As for deserialization, you'd have to [implement a custom deserializer](https://serde.rs/impl-deserialize.html) which internally uses the `TryFrom` implementations.

divinenaman: @SanchithHegde can you help me with these : 

- To derive serializable implementation of `StrongSecret<u16>` or  `StrongSecret<u8>`, `u8` and `u16` primitive types should implement `SerializableSecret` trait defined in masking/serde.rs. I tried implementing it but   It seems I cannot implement it within cards and I need to add the impl in the crate where `SerializableSecret` trait it defined.

- does this deserialize look as expected:  
```rust
impl<'de> Deserialize<'de> for CardSecurityCode {
    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
    where
        D: serde::Deserializer<'de>,
    {
        let csc = u16::deserialize(deserializer)?;
        csc.try_into().map_err(de::Error::custom)
    }
}
```

crates/cards/src/lib.rs
divinenaman: @SanchithHegde @vspecky @NishantJoshi00 thoughts on this ?

crates/cards/Cargo.toml
SanchithHegde: > To derive serializable implementation of `StrongSecret<u16>` or  `StrongSecret<u8>`, `u8` and `u16` primitive types should implement `SerializableSecret` trait defined in masking/serde.rs. I tried implementing it but   It seems I cannot implement it within cards and I need to add the impl in the crate where `SerializableSecret` trait it defined.

Yeah, you can implement the trait for `u8` and `u16` in the `masking` crate itself.

> does this deserialize look as expected:

Looks good, could you add tests to verify that the deserialization works as expected, and that invalid card security codes / expiration months / expiration years raise an error?

crates/cards/Cargo.toml
```diff
@@ -5,10 +5,13 @@ edition = "2021"
...
+[features]
+default = ["serde"]
+
 [dependencies]
-time = { version = "0.3.20", features = ["serde", "serde-well-known", "std"] }
+time = { version = "0.3.20" }
 error-stack = "0.3.1"
-serde = { version = "1.0.155", features = ["derive"] }
+serde = { version = "1", features = ["derive"], optional = true }
 serde_json = "1.0.94"
```
SanchithHegde, line 15: You can move `serde_json` to the `[dev-dependencies]` table since it's only used in tests.

crates/cards/tests/basic.rs
```diff
@@ -0,0 +1,88 @@
+#![allow(dead_code, clippy::unwrap_used, clippy::panic_in_result_fn)]
```
SanchithHegde, line 1:
```suggestion
#![allow(clippy::unwrap_used)]
```

I don't think we violate the other 2 lints here.

crates/cards/tests/basic.rs
```diff
@@ -0,0 +1,88 @@
...
+    let invalid_card_exp = CardExpiration::try_from((13, curr_year + 1));
+
+    assert_eq!(*card_exp.get_month().peek(), 3);
+    assert_eq!(*card_exp.get_year().peek(), curr_year + 1);
+    assert_eq!(card_exp.is_expired().unwrap(), false);
+
+    assert!(invalid_card_exp.is_err());
+
+    let serialized = serde_json::to_string(&card_exp).unwrap();
+    let expected_string = format!("{{\"month\":{},\"year\":{}}}", 3, curr_year + 1);
```
SanchithHegde, line 82:
```suggestion
    let expected_string = format!(r#"{{"month":{},"year":{}}}"#, 3, curr_year + 1);
```

You can make use of Rust's raw string literals to avoid escaping quotes.

SanchithHegde (commented):
Other than that, looks good to me! Thanks again for the PR!

crates/cards/tests/basic.rs
```diff
@@ -0,0 +1,88 @@
...
+use common_utils::date_time;
+use masking::PeekInterface;
+
+#[test]
+fn test_card_security_code() {
+    // no panic
+    let valid_card_security_code = CardSecurityCode::try_from(1234).unwrap();
+
+    // will panic on unwrap
+    let invalid_card_security_code = CardSecurityCode::try_from(12);
```
SanchithHegde, line 13:
```suggestion
    let invalid_card_security_code = CardSecurityCode::try_from(12);
    let invalid_card_security_code_deserialization = serde_json::from_str::<CardSecurityCode>("12"); // Check if this is an error
```

Could you also add tests like this (for each of the newtypes) to ensure deserialization does not allow invalid values?

crates/cards/tests/basic.rs
```diff
@@ -79,10 +88,15 @@ fn test_card_expiration() {
...
+    let expected_string = format!(r#"{{"month":{},"year":{}}}"#, 3, curr_year + 1);
     assert_eq!(serialized, expected_string);
 
     let derialized = serde_json::from_str::<CardExpiration>(&serialized).unwrap();
     assert_eq!(*derialized.get_month().peek(), 3);
     assert_eq!(*derialized.get_year().peek(), curr_year + 1);
+
+    let invalid_serialized_string = format!(r#"{{"month":{},"year":{}}}"#, 13, 123);
+    let invalid_deserialization =
+        serde_json::from_str::<CardExpiration>(&invalid_serialized_string);
```
SanchithHegde, line 100:
```suggestion
    let invalid_serialized_string = r#"{"month":13,"year":123}"#;
    let invalid_deserialization =
        serde_json::from_str::<CardExpiration>(invalid_serialized_string);
```

SanchithHegde (commented):

crates/cards/Cargo.toml
SanchithHegde: @divinenaman let me know if you'd need any help with this.

crates/cards/tests/basic.rs
```diff
@@ -0,0 +1,102 @@
...
+
+    let serialized = serde_json::to_string(&card_exp).unwrap();
+    let expected_string = format!(r#"{{"month":{},"year":{}}}"#, 3, curr_year + 1);
+    assert_eq!(serialized, expected_string);
+
+    let derialized = serde_json::from_str::<CardExpiration>(&serialized).unwrap();
+    assert_eq!(*derialized.get_month().peek(), 3);
+    assert_eq!(*derialized.get_year().peek(), curr_year + 1);
+
+    let invalid_serialized_string = r#"{{"month":13,"year":123}}"#;
```
SanchithHegde, line 98: You can remove one pair of curly braces:

```suggestion
    let invalid_serialized_string = r#"{"month":13,"year":123}"#;
```

SanchithHegde (approved):

crates/cards/tests/basic.rs
```diff
@@ -0,0 +1,102 @@
...
+    // will panic on unwrap
+    let invalid_card_security_code = CardSecurityCode::try_from(12);
+
+    assert_eq!(*valid_card_security_code.peek(), 1234);
+    assert!(invalid_card_security_code.is_err());
+
+    let serialized = serde_json::to_string(&valid_card_security_code).unwrap();
+    assert_eq!(serialized, "1234");
+
+    let derialized = serde_json::from_str::<CardSecurityCode>(&serialized).unwrap();
```
vspecky, line 21: Not a blocker but prefer `.expect("what was expected")` over `.unwrap()` since it makes finding the cause of panic easier.

vspecky (dismissed):
LGTM

SanchithHegde:
@divinenaman Please address CI checks.

SanchithHegde (approved):

vspecky (approved):

jarnura (approved):

PR #876, opened 2023-04-13
fix(stripe): remove cancel reason validation for stripe

ArjunKarthik:
## Type of Change

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates

## Description
For stripe in case of invalid cancellation reason, we should not throw RequestEncodingFailed as it will be returned as InternalServer in our response. Instead validation should happen at connector and connector error should be shown in our response.

### Additional Changes

## How did you test it?
Screen Shot 2023-04-13 at 6 32 45 PM

## Checklist

### DIFF
--- a/crates/router/src/connector/stripe/transformers.rs
+++ b/crates/router/src/connector/stripe/transformers.rs
@@ -1,5 +1,3 @@
-use std::str::FromStr;
-
 use api_models::{self, enums as api_enums, payments};
 use base64::Engine;
 use common_utils::{fp_utils, pii::Email};
@@ -1207,25 +1205,16 @@ pub struct StripeRedirectResponse {
     pub source_type: Option<Secret<String>>,
 }
 
-#[derive(Debug, Serialize, Clone, Copy)]
+#[derive(Debug, Serialize)]
 pub struct CancelRequest {
-    cancellation_reason: Option<CancellationReason>,
+    cancellation_reason: Option<String>,
 }
 
 impl TryFrom<&types::PaymentsCancelRouterData> for CancelRequest {
     type Error = error_stack::Report<errors::ConnectorError>;
     fn try_from(item: &types::PaymentsCancelRouterData) -> Result<Self, Self::Error> {
-        let cancellation_reason = match &item.request.cancellation_reason {
-            Some(c) => Some(
-                CancellationReason::from_str(c)
-                    .into_report()
-                    .change_context(errors::ConnectorError::RequestEncodingFailed)?,
-            ),
-            None => None,
-        };
-
         Ok(Self {
-            cancellation_reason,
+            cancellation_reason: item.request.cancellation_reason.clone(),
         })
     }
 }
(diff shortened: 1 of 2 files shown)


crates/router/src/connector/stripe/transformers.rs
```diff
@@ -1207,25 +1205,16 @@ pub struct StripeRedirectResponse {
     pub source_type: Option<Secret<String>>,
 }
 
-#[derive(Debug, Serialize, Clone, Copy)]
+#[derive(Debug, Serialize)]
 pub struct CancelRequest {
-    cancellation_reason: Option<CancellationReason>,
+    cancellation_reason: Option<String>,
```
Abhicodes-crypto, line 1210: The enum `CancellationReason` is not deleted.

Abhicodes-crypto:
According to stripe docs it accepts only these values `duplicate, fraudulent, requested_by_customer, or abandoned` for cancellation reason. If any other value is sent it will respond with an error. 

By having an enum we ensure that only these values are passed. By making it a string, type safety is compromised and call to the connector is made unnecessarily. 

I think a proper mapping should be done for `RequestEncodingFailedWithReason` at https://github.com/juspay/hyperswitch/blob/cf902f19e5e08c6ee538ab4be7411f50a4079b55/crates/router/src/core/errors/utils.rs#L116 and that would help throw a proper error to user rather than internal server error.

crates/router/src/connector/stripe/transformers.rs
ArjunKarthik: @Abhicodes-crypto 

RequestEncodingFailedWithReason should be handled only as InternalServerError as this is a failure happened while building connector request at our side.

However, we can solve this problems in two ways,
1. Having validation at our side (using enums).   Incase validation fails we can have a new error type which will propagate proper message instead of throwing `InternalServerError`. But till now we don't have any unified logic to handle connector level field validation. And we have to maintain a proper documentation for the possible validation errors.Incase validation fails we can have a new error type which will propagate proper message instead of throwing `InternalServerError`. But till now we don't have any unified logic to handle connector level field validation. And we have to maintain a proper documentation for the possible validation errors.

2. Relying on connector's validation. Exact error message from connector will be shown as ConnectorError in the response instead of us throwing error on behalf of connector.

I discussed both the solution with @jarnura, Considering the complexity involves in first he suggested to go with 2 solution.

Abhicodes-crypto: Cool. 

I guess you can still delete `enum CancellationReason` if it is not being used anywhere in `stripe/transformers.rs`.

Also I guess an appropriate error message could be given here. 
https://github.com/juspay/hyperswitch/blob/cf902f19e5e08c6ee538ab4be7411f50a4079b55/crates/router/src/connector/stripe.rs#L558

ArjunKarthik: Yeah you are right. In stripe.rs it should be RequestEncodingFailed error. I will change there too

Abhicodes-crypto (approved):

jagan-jaya (approved):

jarnura (approved):

PR #878, opened 2023-04-13
fix: redis issue in tokenization call

Sangamesh26:
## Type of Change

- [x] Bugfix

## Description

* Fix redis issue in tokenization call
* Fix requires_payment_method_data error

## Motivation and Context

fix for requires_payment_method_data

## How did you test it?

Manual

## Checklist

### DIFF
--- a/crates/router/src/core/payments.rs
+++ b/crates/router/src/core/payments.rs
@@ -319,7 +319,7 @@ impl PaymentRedirectFlow for PaymentRedirectCompleteAuthorize {
             metadata: Some(Metadata {
                 order_details: None,
                 data: masking::Secret::new("{}".into()),
-                payload: Some(req.json_payload.unwrap_or("{}".into()).into()),
+                payload: Some(req.json_payload.unwrap_or_else(|| "{}".into()).into()),
             }),
             ..Default::default()
         };
@@ -597,13 +597,13 @@ async fn decide_payment_method_tokenize_action(
     payment_method: &storage::enums::PaymentMethod,
     pm_parent_token: Option<&String>,
     is_connector_tokenization_enabled: bool,
-) -> TokenizationAction {
+) -> RouterResult<TokenizationAction> {
     match pm_parent_token {
         None => {
             if is_connector_tokenization_enabled {
-                TokenizationAction::TokenizeInConnectorAndRouter
+                Ok(TokenizationAction::TokenizeInConnectorAndRouter)
             } else {
-                TokenizationAction::TokenizeInRouter
+                Ok(TokenizationAction::TokenizeInRouter)
             }
         }
         Some(token) => {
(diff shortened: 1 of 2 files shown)


vspecky (approved):

SanchithHegde (approved):

crates/router/src/core/payments.rs
```diff
@@ -319,7 +319,7 @@ impl PaymentRedirectFlow for PaymentRedirectCompleteAuthorize {
             metadata: Some(Metadata {
                 order_details: None,
                 data: masking::Secret::new("{}".into()),
-                payload: Some(req.json_payload.unwrap_or("{}".into()).into()),
+                payload: Some(req.json_payload.unwrap_or_else(|| "{}".into()).into()),
```
NishantJoshi00, line 322: Correct me If I am wrong, here:
- json_payload: `Option<Value>`
- payload: `Option<SecretSerdeValue>`,

By doing this operation that is performed here, the value that is being yielded when json_payload is `None` will be `serde_json::Value::String("{}")`, But shouldn't this be instead `serde_json::Value::Object {}`

crates/router/src/core/payments/helpers.rs
```diff
@@ -691,7 +691,17 @@ pub async fn make_pm_data<'a, F: Clone, R>(
...
+        let hyperswitch_token_option = redis_conn
+            .get_key::<Option<String>>(&key)
+            .await
+            .change_context(errors::ApiErrorResponse::InternalServerError)
+            .attach_printable("Failed to fetch the token from redis")?;
+
+        match hyperswitch_token_option {
+            Some(hyperswitch_token) => Some(hyperswitch_token),
+            None => Some(token),
+        }
```
NishantJoshi00, line 704: can we do something like, `hyperswitch_token_option.or_else(|| Some(token))`

NishantJoshi00 (changes requested):
Looks good otherwise!

NishantJoshi00 (changes requested):
Looks good otherwise!

crates/router/src/core/payments.rs
NishantJoshi00:
```suggestion
                payload: Some(req.json_payload.unwrap_or(serde_json::json!({}))).into()),
```

crates/router/src/core/payments.rs
```diff
@@ -615,14 +615,19 @@ async fn decide_payment_method_tokenize_action(
...
+                None => {
                     if is_connector_tokenization_enabled {
-                        TokenizationAction::TokenizeInConnector
+                        Ok(TokenizationAction::TokenizeInConnector)
                     } else {
-                        TokenizationAction::SkipConnectorTokenization
+                        Ok(TokenizationAction::TokenizeInRouter)
                     }
                 }
             }
```
NishantJoshi00, line 633: Hey, minor nit pick, now that you are using a variable you can do something like this:
```rust
match (connector_token_option, is_connector_tokenization_enabled) {
    (Some(connector_token), _) => Ok(TokenizationAction::ConnectorToken(connector_token)),
    (None, true) => Ok(TokenizationAction::TokenizeInConnector),
    (None, false) => Ok(TokenizationAction::TokenizeInRouter)
}
```

NishantJoshi00 (commented):
Everything looks good!

jarnura (approved):

crates/router/src/core/payments.rs
NishantJoshi00: can change this, moving forward

NishantJoshi00 (approved):

PR #883, opened 2023-04-14
feat(router): add support for filtering PaymentMethodTypes at a session

kashif-m:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [x] Enhancement
- [ ] Refactoring
- [ ] Dependency updates

## Description
A new field allowed_payment_method_types is introduced in /payments (POST) for limiting a payment intent to specified payment method types. If passed, it will be stored as metadata in payment_intent table and will be used to filter the PaymentMethodTypes during /payment_methods (GET) call.

### Additional Changes

## Motivation and Context
This feature enables merchant to filter certain payment methods only for a particular payment intent. This helps in cases where the merchants wants to limit the payment methods shown for payment intents based on the requirements.

Eg: limiting PaymentMethods for a particular period / amount

## How did you test it?
Tested manually.

## Checklist

### DIFF
--- a/crates/api_models/src/payments.rs
+++ b/crates/api_models/src/payments.rs
@@ -212,6 +212,10 @@ pub struct PaymentsRequest {
     /// Merchant connector details used to make payments.
     pub merchant_connector_details: Option<admin::MerchantConnectorDetailsWrap>,
 
+    /// Allowed Payment Method Types for a given PaymentIntent
+    #[schema(value_type = Option<Vec<PaymentMethodType>>)]
+    pub allowed_payment_method_types: Option<Vec<api_enums::PaymentMethodType>>,
+
     /// Business sub label for the payment
     pub business_sub_label: Option<String>,
 }
@@ -978,6 +982,10 @@ pub struct PaymentsResponse {
 
     /// The business_sub_label for this payment
     pub business_sub_label: Option<String>,
+
+    /// Allowed Payment Method Types for a given PaymentIntent
+    #[schema(value_type = Option<Vec<PaymentMethodType>>)]
+    pub allowed_payment_method_types: Option<Vec<api_enums::PaymentMethodType>>,
 }
 
 #[derive(Clone, Debug, serde::Deserialize, ToSchema)]
@@ -1238,6 +1246,10 @@ pub struct Metadata {
     pub data: pii::SecretSerdeValue,
     /// Payload coming in request as a metadata field
     pub payload: Option<pii::SecretSerdeValue>,
+
+    /// Allowed payment method types for a payment intent
+    #[schema(value_type = Option<Vec<PaymentMethodType>>)]
+    pub allowed_payment_method_types: Option<Vec<api_enums::PaymentMethodType>>,
 }
 
 #[derive(Debug, serde::Deserialize, serde::Serialize, Clone, ToSchema)]
--- a/crates/router/src/core/payment_methods/cards.rs
+++ b/crates/router/src/core/payment_methods/cards.rs
@@ -1056,6 +1056,26 @@ async fn filter_payment_methods(
         let parse_result = serde_json::from_value::<PaymentMethodsEnabled>(payment_method);
         if let Ok(payment_methods_enabled) = parse_result {
             let payment_method = payment_methods_enabled.payment_method;
+            let allowed_payment_method_types = payment_intent
+                .map(|payment_intent|
+                    payment_intent
+                        .metadata
+                        .as_ref()
+                        .and_then(|masked_metadata| {
+                            let metadata = masked_metadata.peek().clone();
+                            let parsed_metadata: Option<api_models::payments::Metadata> =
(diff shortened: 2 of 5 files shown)


jarnura (dismissed):

crates/api_models/src/payments.rs
```diff
@@ -212,6 +212,9 @@ pub struct PaymentsRequest {
     /// Merchant connector details used to make payments.
     pub merchant_connector_details: Option<admin::MerchantConnectorDetailsWrap>,
 
+    /// Allowed Payment Method Types for a given PaymentIntent
+    pub allowed_payment_method_types: Option<Vec<api_enums::PaymentMethodType>>,
```
SanchithHegde, line 216:
```suggestion
    #[schema(value_type = Option<Vec<PaymentMethodType>>)]
    pub allowed_payment_method_types: Option<Vec<api_enums::PaymentMethodType>>,
```

This so that the OpenAPI spec generated is proper.

crates/api_models/src/payments.rs
```diff
@@ -1224,6 +1227,9 @@ pub struct Metadata {
     pub data: pii::SecretSerdeValue,
     /// Payload coming in request as a metadata field
     pub payload: Option<pii::SecretSerdeValue>,
+
+    /// Allowed payment method types for a payment intent
+    pub allowed_payment_method_types: Option<Vec<api_enums::PaymentMethodType>>,
```
SanchithHegde, line 1232:
```suggestion
    #[schema(value_type = Option<Vec<PaymentMethodType>>)]
    pub allowed_payment_method_types: Option<Vec<api_enums::PaymentMethodType>>,
```

crates/router/src/core/payment_methods/cards.rs
```diff
@@ -1060,6 +1060,28 @@ async fn filter_payment_methods(
         let parse_result = serde_json::from_value::<PaymentMethodsEnabled>(payment_method);
         if let Ok(payment_methods_enabled) = parse_result {
             let payment_method = payment_methods_enabled.payment_method;
+            let allowed_payment_method_types = if let Some(payment_intent) = payment_intent {
```
SanchithHegde, line 1063: You can use `payment_intent.map()` instead of using `if let Some(...) = payment_intent { ... } else { None }`.

crates/router/src/core/payment_methods/cards.rs
```diff
@@ -1060,6 +1060,28 @@ async fn filter_payment_methods(
...
+                payment_intent
+                    .metadata
+                    .as_ref()
+                    .and_then(|masked_metadata| {
+                        let metadata = masked_metadata.peek().clone();
+                        let parsed_metadata: Option<api_models::payments::Metadata> =
+                            match serde_json::from_value(metadata) {
+                                Ok(metadata) => Some(metadata),
+                                _ => None,
+                            };
```
SanchithHegde, line 1073:
```suggestion
                            serde_json::from_value(metadata)
                                .map_err(|error| logger::error!(%error, "Failed to deserialize payment intent metadata"))
                                .ok()
```

@shakthi-22 Don't we need to raise errors in case parsing metadata fails?

crates/router/src/core/payment_methods/cards.rs
```diff
@@ -1060,6 +1060,28 @@ async fn filter_payment_methods(
...
+                        parsed_metadata.and_then(|pm| {
+                            logger::info!(
+                                "Only given PaymentMethodTypes will be allowed {:?}",
+                                pm.allowed_payment_method_types
```
Narayanbhat166, line 1077: Was this meant for debugging purposes? if so I think this can be removed.

crates/api_models/src/payments.rs
Narayanbhat166, line 1232: We should be populating the response too, the metadata struct should be parsed and passed here. You can check the [payments_to_payments_response](https://github.com/juspay/hyperswitch/blob/43aaf8f735eb2b74e12ab0fe49026dd6767e546e/crates/router/src/core/payments/transformers.rs#L233) function, and add this field there.

Narayanbhat166 (changes requested):

kashif-m:
**PaymentIntent Creation with specified allowed PaymentMethod types**
Expectation - PaymentIntent is created and only the allowed PaymentMethod types will be shown when using the PaymentMethod list API



**Fetching PaymentMethods using for that session (using client_secret)**
Expectation - PaymentMethods are filtered on the basis of the payment_method_types allowed for that session



**Fetching PaymentMethods for the merchant (session-less)**
Expectation - PaymentMethods are unfiltered and contains all the methods enabled for that merchant



**Fetching the PaymentIntent**
Expectation - PaymentsResponse contains the new field which was added in the request - `allowed_payment_method_types`

SanchithHegde (approved):

Narayanbhat166 (approved):

jarnura (approved):

PR #886, opened 2023-04-14
refactor(storage_models, errors): impl StorageErrorExt for error_stack::Result<T, errors::StorageError>

Abhicodes-crypto:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [x] Refactoring
- [ ] Dependency updates

## Description

changing the below pattern code 
```
.map_err(|error| {
            error.to_duplicate_response(errors::ApiErrorResponse::DuplicateMerchantAccount)
        })?;
```
to 
```
.to_duplicate_response(errors::ApiErrorResponse::DuplicateMerchantAccount)?;
```

### Additional Changes

## How did you test it?

Manually, It compiles.

## Checklist

### DIFF
--- a/crates/router/src/core/admin.rs
+++ b/crates/router/src/core/admin.rs
@@ -155,9 +155,7 @@ pub async fn create_merchant_account(
     let merchant_account = db
         .insert_merchant(merchant_account)
         .await
-        .map_err(|error| {
-            error.to_duplicate_response(errors::ApiErrorResponse::DuplicateMerchantAccount)
-        })?;
+        .to_duplicate_response(errors::ApiErrorResponse::DuplicateMerchantAccount)?;
 
     Ok(service_api::ApplicationResponse::Json(
         ForeignTryFrom::foreign_try_from(merchant_account).change_context(
@@ -175,9 +173,7 @@ pub async fn get_merchant_account(
     let merchant_account = db
         .find_merchant_account_by_merchant_id(&req.merchant_id)
         .await
-        .map_err(|error| {
-            error.to_not_found_response(errors::ApiErrorResponse::MerchantAccountNotFound)
-        })?;
+        .to_not_found_response(errors::ApiErrorResponse::MerchantAccountNotFound)?;
 
     Ok(service_api::ApplicationResponse::Json(
         ForeignTryFrom::foreign_try_from(merchant_account).change_context(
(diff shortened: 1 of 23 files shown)


jarnura (approved):

crates/router/src/core/admin.rs
```diff
@@ -421,9 +409,7 @@ pub async fn create_payment_connector(
     let mca = store
         .insert_merchant_connector_account(merchant_connector_account)
         .await
-        .map_err(|error| {
-            error.to_duplicate_response(errors::ApiErrorResponse::DuplicateMerchantConnectorAccount)
-        })?;
+        .to_duplicate_response(errors::ApiErrorResponse::DuplicateMerchantConnectorAccount)?;
```
Narayanbhat166, line 412: Can you add a field called `connector_label` which will provide an information about what was the connector label that was created, because we construct this field.

crates/router/src/core/admin.rs
```diff
@@ -438,19 +424,15 @@ pub async fn retrieve_payment_connector(
...
     let mca = store
         .find_by_merchant_connector_account_merchant_id_merchant_connector_id(
             &merchant_id,
             &merchant_connector_id,
         )
         .await
-        .map_err(|error| {
-            error.to_not_found_response(errors::ApiErrorResponse::MerchantConnectorAccountNotFound)
-        })?;
+        .to_not_found_response(errors::ApiErrorResponse::MerchantConnectorAccountNotFound)?;
```
Narayanbhat166, line 435: Same here, would be better if we add the `connector_label` field

Narayanbhat166 (changes requested):
Minor changes, Overall seems fine

crates/router/src/core/admin.rs
Abhicodes-crypto: Sure. If this not urgent we could add it as good first issue.

crates/router/src/core/admin.rs
Narayanbhat166: Not urgent, but good to have. Yes better to create the issue. Please create one with the good first issue label and mention it here

Narayanbhat166 (approved):

crates/router/src/core/admin.rs
Abhicodes-crypto: created an issue here #899

PR #889, opened 2023-04-14
feat(cards): add credit card number validation

### LINKED ISSUE #606: [FEATURE] Validate card numbers when accepting payment method information
### Feature Description

As of now, we just pass the card information to the payment processor without performing any form of validation on the provided card number. It'd be great if the card number could have been validated on our end whenever a payment is being created or payment method is being created, instead of invalid card numbers being caught by the payment processor.

It'd be preferable if the card number validation logic is included in a new crate called `cards` which would contain this validation of card numbers and card brand identification logic (taken up on #379). This must check that the card number provided is of correct length, and also handle whitespaces being included in the card number input. Please also include unit tests in your PR.

### Possible Implementation

For validating card numbers, you can use the [Luhn Algorithm](https://en.wikipedia.org/wiki/Luhn_algorithm). I'm thinking of an API like so (refer to [the newtype pattern](https://rust-unofficial.github.io/patterns/patterns/behavioural/newtype.html)):

```rust
use masking::StrongSecret;

struct CardNumber(StrongSecret<String>);

let card_number = CardNumber::from_str("..."); // You can use the `FromStr` trait for this purpose
```

This would ensure that a `CardNumber` once constructed is always valid and no further validation is necessary. You can also implement `Serialize` and `Deserialize` traits on `CardNumber`.

None

phillyphil91:
Fixes #606

## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates

## Description

The PR adds a new crate called `cards`. Which currently just has a single module named validate which exposes the `CardNumber` struct as well as a `CCValError`.

The validation of a correct CC number is done via the Luhn Algorithm implemented in the `luhn` crate.

### Additional Changes

## Motivation and Context

This PR address #606.

## How did you test it?

Tests for successful and failing CC number validation have been added.

## Checklist


phillyphil91:
I wasn't sure what to do regarding the Error that `StrongSecret::from_str(s)` could produce (`Infallible`).

I guess I could have also made the associative type `Err` in `FromStr` to be `type Err = Box<dyn std::error::Error>;` but I feel like the current approach makes more sense. Otherwise I would have had to box `CCValError`

crates/cards/src/lib.rs
SanchithHegde, line 1: Could you re-export `CardNumber` and `CCValError` at the root of the crate?

SanchithHegde (commented):
Thanks for the PR, @phillyphil91!

Could you add test cases to verify handling of whitespaces in the input, and to ensure that invalid card numbers are caught during deserialization?

Could you also port over the implementation of `Strategy` to `cards::CardNumber` from the `common_utils` crate ([reference](https://github.com/juspay/hyperswitch/blob/b12762e776205a1b6375a7c49447e8a3dc541b8a/crates/common_utils/src/pii.rs#L16-L29))? I forgot to include this in the issue description.

> Also please advice if the test should go into their own module or if it's okay to keep them in the file as is.

It's fine to keep them in the same file.

phillyphil91:
Hi @SanchithHegde 

I refactored a bigger chunk of the code when I was moving `Strategy` trait over from masking, because I realized that there would otherwise be two places for `CardNumber`, the new one in `crates/cards` and the one that was in `common_utils:pii`. 

Since quite a few places were using `common_utils:pii::CardNumber` i replaced that everywhere with the new `cards::CardNumber`.

I also moved the test regarding masking of `CardNumber` to `cards::validate`. Creating a secret from a valid and a non valid CC number works when specifying `Strategy` of `CardNumber`: 
```rust
let secret: Secret<String, CardNumber>
```

When creating a `CardNumber` the inner value is a `StrongSecret` which uses the `WithType` `Strategy` per default which seems to work as well. 

Still not really sure how to use the error properly,  i.e. what to do in the masking case when the CC number is not valid. Right now it's masked just as a valid CC number. 

> and to ensure that invalid card numbers are caught during deserialization?

Also I wasn't sure what you meant by this.

Let me know if this works for you and what to change. Or if this is too much I can always go back to the previous commit and just add the stuff that you mentioned in your earlier comment.

crates/cards/src/validate.rs
```diff
@@ -0,0 +1,88 @@
...
+    fn fmt(val: &T, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        let val_str: &str = val.as_ref();
+
+        if val_str.len() < 15 || val_str.len() > 19 {
+            return WithType::fmt(val, f);
+        }
+
+        write!(f, "{}{}", &val_str[..6], "*".repeat(val_str.len() - 6))
+    }
+}
```
vspecky, line 45: Hey, you might have had a slight misunderstanding regarding how a masking `Strategy` is supposed to function. I'll briefly explain the relationship between the `Data` and the `Strategy` in `masking::Secret<Data, Strategy>` in case there's some confusion.

The `Data` type parameter is a type that represents the sensitive data, and the `Strategy` type parameter represents the strategy using which the `Data` is to be masked. All the data is stored in `Data`, the `Strategy` is just intended to be a simple (usually a unit struct) type that provides an impl of the `Strategy<T>::fmt` method (which doesn't take any `&self`).

Here, your `CardNumber` struct is doing both, storing a card number `Data` and providing the `Strategy<T>` impl. This means that when you specify `masking::Secret<String, CardNumber>`, the `String` is storing your card number, and `CardNumber` just provides the strategy even tho it's capable of doing both, along with providing validations.

Since you're trying to do validations, my suggestion would be to implement a separate unit struct `CardNumberStrategy` which provides the `Strategy<T>` impl, updating the `CardNumber` struct to `struct CardNumber(StrongSecret<String, CardNumberStrategy>)` and update usages of `masking::Strategy<String, CardNumber>` in the code with just `cards::CardNumber`.
With this, you have a wrapper type (`CardNumber`) around a `StrongSecret` storing your card number which provides validations, and internally, the `StrongSecret` holds the masking strategy.

If you're going down this route, also consider implementing `Deref` on `CardNumber`.

vspecky (changes requested):

SanchithHegde:
> > and to ensure that invalid card numbers are caught during deserialization?
> 
> Also I wasn't sure what you meant by this.

Could you add tests to ensure deserialization catches invalid card numbers? So a call like `serde_json::from_str::<CardNumber>(r#""1234 5678""#)` should fail deserialization because of the invalid card number.

Regarding the refactor, it's fine, I did expect it to happen.

phillyphil91:
Hi @SanchithHegde 
thank you for the detailed explanation of how the masking Strategy works. It confirmed how I understood it before, but I thought maybe it's okay that `CardNumber` does both: hold the data as well as provide a masking Strategy.
But I totally agree that it makes more sense to split the two things into dedicated types. I created `CardNumberStrategy` as you suggested and replaced everything that uses `CardNumber` with the new Strategy.

One thing that I wasn't sure about is that in a lot of places use something similar to:
```rust 
pub card_number: Secret<String, CardNumberStrategy>
```
Would it make sense to replace all those with the new `CardNumber` struct that actually holds the data?

Also regarding `Deref` : I thought this is strictly for `Smart Pointers` and hence i created `get_secret()` instead. Let me know what you think.

I think the cc number check does not happen before deserialization unfortunately. Not sure what to do about this. Implement `Deserialize` manually?

vspecky:
Hey @phillyphil91,

It would be fine to replace `masking::Secret<String, Strategy>` with `CardNumber` throughout the code, since it's just a wrapper around the actual `StrongSecret`. Typical case of the [newtype pattern](https://rust-unofficial.github.io/patterns/patterns/behavioural/newtype.html), and due to Rust's zero cost abstractions, the `CardNumber` wrapper won't have any size.

We'll also most probably go ahead with a custom implementation for `Deserialize` that internally performs validation and returns an appropriate error message in case of failure (unless @SanchithHegde @NishantJoshi00, you guys see an issue). The idea is that card number validation gets done at the API layer itself by actix when it deserializes the HTTP request body into the required type, and the developer does not have to manually call any function to ensure validity.
Serde has documentation on how to do the same. If you'd like a reference, check out the `mod`s at the bottom of [crates/api_models/src/payments.rs](https://github.com/juspay/hyperswitch/blob/573a4d384ee6a9d72648ab537804799a3993e1e8/crates/api_models/src/payments.rs)

We don't have to worry too much about serialization. We can just `#[derive(Serialize)]` and annotate with `#[serde(transparent)]`, and the serialization logic will skip the `CardNumber` struct and just serialize the inner secret.

> Also regarding Deref : I thought this is strictly for Smart Pointers and hence i created get_secret() instead. Let me know what you think.

While `Deref` works great for smart pointers, it's not restricted to that domain and can be used wherever needed. Usage of `Deref` is pretty popular with the newtype pattern. It allows us to seamlessly go from having to write `card_number.0.some_method_on_strong_secret()` to being able to write `card_number.some_method_on_strong_secret()` (no `.0`).

phillyphil91:
Hi @vspecky 
thanks for the comment. Just to be clear, should I try to implement `Deserialize` manually for `CardNumber` according to the example that you provided?

crates/cards/src/validate.rs
```diff
@@ -30,12 +32,25 @@ impl FromStr for CardNumber {
...
+fn from_str<'de, D>(deserializer: D) -> Result<StrongSecret<String, CardNumberStrategy>, D::Error>
+where
+    D: Deserializer<'de>,
+{
+    let s: String = Deserialize::deserialize(deserializer)?;
+    match CardNumber::from_str(&s) {
+        Ok(secret) => Ok(secret.clone()),
+        Err(e) => Err(serde::de::Error::custom(e.to_string())),
+    }
+}
```
SanchithHegde, line 52: Can you manually implement `Deserialize` for `CardNumber` instead? Something along the lines of:

```rust
str::deserialize(deserializer)?.try_into().map_err(de::Error::custom)
```

phillyphil91: I currently did it via:

```rust
deserialize<D: Deserializer<'de>>(d: D) -> Result<Self, D::Error> {
        let s = String::deserialize(d)?;
        match Self::from_str(&s) {
            Ok(card_number) => Ok(card_number),
            Err(e) => Err(serde::de::Error::custom(e.to_string())),
        }
    }
```

If I want to do it via `try_into()` I would also have to implement `TryFrom<&str>` or `TryFrom<String>` for `CardNumber` right? And in there also use the card number validation.

Is the `try_into()` variant preferred?

crates/cards/src/validate.rs
```diff
@@ -0,0 +1,128 @@
...
+
+impl<'de> Deserialize<'de> for CardNumber {
+    fn deserialize<D: Deserializer<'de>>(d: D) -> Result<Self, D::Error> {
+        let s = String::deserialize(d)?;
+        match Self::from_str(&s) {
+            Ok(card_number) => Ok(card_number),
+            Err(_) => Err(serde::de::Error::custom(CCValError)),
+        }
```
SanchithHegde, line 50:
```suggestion
        Self::from_str(&s).map_err(serde::de::Error::custom)
```

crates/cards/src/validate.rs
```diff
@@ -0,0 +1,128 @@
...
+        let card_number = CardNumber::from_str("3714 4963 5398 431").unwrap();
+        let secret = &(*card_number);
+        assert_eq!("3714 4************", format!("{secret:?}"));
+    }
+
+    #[test]
+    fn test_valid_card_number_deserialization() {
+        let card_number = serde_json::from_str::<CardNumber>(r#""3714 4963 5398 431""#).unwrap();
+        let secret = card_number.to_string();
+        assert_eq!("\"3714 4************\"", format!("{secret:?}"));
```
SanchithHegde, line 119: You could also use raw string literals here, not a necessary change though.

crates/api_models/src/payment_methods.rs
```diff
@@ -72,7 +73,7 @@ pub struct PaymentMethodUpdate {
 pub struct CardDetail {
     /// Card Number
     #[schema(value_type = String,example = "4111111145551142")]
-    pub card_number: masking::Secret<String, pii::CardNumber>,
+    pub card_number: masking::Secret<String, CardNumberStrategy>,
```
SanchithHegde, line 76: You could just use `CardNumber` as the type instead. Please make this change elsewhere too.

```suggestion
    pub card_number: CardNumber,
```

SanchithHegde (commented):

crates/cards/src/validate.rs
SanchithHegde: It's fine, the `try_into()` is not required.

phillyphil91:
Hi @SanchithHegde 

I changed all occurrences of Secret wherever it was used to hold the card number to CardNumber (there where actually quite a lot of places :smile:) 
A few issues arose due to this. Mainly:
1. Beforehand creating a Secret from a CC Number could  not fail, i.e. `into()` always worked. Now I have to use `try_into()` and `unwrap()` to get the `CardNumber`. Same with `from_str()` which can also return the `CCValError`. I know this isn't really good but I was not sure how to handle or where to propagate the error to. E.g.:
`card_number: Some(response.card_number.try_into().unwrap()),`
2. There are some tests that create empty CC numbers to check if API calls fail with a missing number. Now that there is a validation check on CC numbers, one with an empty string can no longer be created and would result in an Error. Again the issue is how to handle this. E.g.:
`card_number: cards::CardNumber::from_str("").unwrap(),` -> will panic, because the `Result` will be an `Err` variant.

I haven't really looked into the call stack of some places that just had a "always" working Secret before, if is easy to propagate the error. Maybe you have an idea for this?

crates/cards/src/validate.rs
```diff
@@ -0,0 +1,146 @@
...
+
+impl CardNumber {
+    pub fn remove_whitespace(self) -> Self {
+        let cc_no_whitespace: String = self.peek().split_whitespace().collect();
+        let secret = StrongSecret::new(cc_no_whitespace);
+        Self(secret)
+    }
+}
+
+pub struct CardNumberStrategy;
```
SanchithHegde, line 63: You should be able to mark this private, since no types outside this module would need it.

```suggestion
struct CardNumberStrategy;
```

crates/cards/src/validate.rs
```diff
@@ -0,0 +1,146 @@
...
+    }
+}
+
+impl CardNumber {
+    pub fn remove_whitespace(self) -> Self {
+        let cc_no_whitespace: String = self.peek().split_whitespace().collect();
+        let secret = StrongSecret::new(cc_no_whitespace);
+        Self(secret)
+    }
+}
```
SanchithHegde, line 61: @Narayanbhat166 Can we remove all whitespaces when constructing a `CardNumber` itself, or are there any connectors which require us to send card numbers with spaces?

crates/router/src/connector/authorizedotnet/transformers.rs
```diff
@@ -636,7 +638,7 @@ pub struct AuthorizedotnetErrorResponse {
 
 fn construct_refund_payment_details(masked_number: String) -> PaymentDetails {
     PaymentDetails::CreditCard(CreditCardDetails {
-        card_number: masked_number.into(),
+        card_number: cards::CardNumber::from_str(&masked_number).unwrap(),
         expiration_date: "XXXX".to_string().into(),
         card_code: None,
     })
```
SanchithHegde, line 644: @Narayanbhat166 Would this fail if the card number passed to the function is a masked one? How do you suggest we handle this?

SanchithHegde (commented):

To answer your queries:

> 1. Beforehand creating a Secret from a CC Number could  not fail, i.e. `into()` always worked. Now I have to use `try_into()` and `unwrap()` to get the `CardNumber`. Same with `from_str()` which can also return the `CCValError`. I know this isn't really good but I was not sure how to handle or where to propagate the error to. E.g.:
>        `card_number: Some(response.card_number.try_into().unwrap()),`

If it's about `unwrap()`s in tests, that's fine. If it's in production code however, you'd have to avoid it. As far as I could see, there was only one `unwrap()` within code, and everything else were in tests. If there are more instances in code, remove them, or let us know if you need help.

> 2. There are some tests that create empty CC numbers to check if API calls fail with a missing number. Now that there is a validation check on CC numbers, one with an empty string can no longer be created and would result in an Error. Again the issue is how to handle this. E.g.:
>         `card_number: cards::CardNumber::from_str("").unwrap(),` -> will panic, because the `Result` will be an `Err` variant.

We should be able to remove that test case from all connector tests, since we're handling it ourselves. @jagan-jaya @ArjunKarthik Please confirm.

crates/cards/src/validate.rs
phillyphil91: The problem is that I use it in the `Deref` trait which is a public interface and hence I have to make `CardNumberStrategy` public as well.
If I understood everything correctly.

phillyphil91:
hi @SanchithHegde 


> If it's about `unwrap()`s in tests, that's fine. If it's in production code however, you'd have to avoid it. As far as I could see, there was only one `unwrap()` within code, and everything else were in tests. If there are more instances in code, remove them, or let us know if you need help.

Unfortunately in `crates/router/src/core/payment_methods/cards.rs` there are numerous places as well as in `crates/router/src/core/payment_methods/vault.rs` for example.
I wonder if it makes sense to replace `CardNumber` with `Result<CardNumber, CCValError>` in some places, such as in `AddCardResponse` in `crates/router/src/core/payment_methods/transformers.rs` or find some other way to include the fact that a card number can now no longer always successfully be created but might fail sometimes. 

> We should be able to remove that test case from all connector tests, since we're handling it ourselves.

Does this mean I leave them in for now and you take care of this?

crates/cards/src/validate.rs
SanchithHegde: Alright then you can ignore my comment. Marking this conversation resolved.

crates/cards/src/validate.rs
Narayanbhat166: It's not usually the case that connectors accept card numbers with spaces only, without spaces also it should work as expected. But if there arises any case where the connector requires the card number to be sent with spaces only then we can have the conversion to add whitespaces for that connector only. To be consistent, I would say better to remove all the whitespaces. Currently with the connectors that I have come across, they accept card numbers without spaces. @jagan-jaya can share his opinion on this.

SanchithHegde: I did discuss this with @jagan-jaya, we can go ahead and remove spaces when constructing the `CardNumber` type. @phillyphil91 Please make that change, and remove the `remove_whitespace()` method you've added here.

phillyphil91: So basically remove all whitespaces in the from_str implementation of CardNumber so that at the end we will always have a CC number without whitespaces? (If the number is valid in general of course)

crates/router/src/core/payment_methods/cards.rs
```diff
@@ -627,7 +627,7 @@ pub async fn mock_add_card(
         card_fingerprint: response.card_fingerprint.into(),
         card_global_fingerprint: response.card_global_fingerprint.into(),
         merchant_id: Some(response.merchant_id),
-        card_number: Some(response.card_number.into()),
+        card_number: Some(response.card_number.try_into().unwrap()),
```
Narayanbhat166, line 630: You can throw this value error 
https://github.com/juspay/hyperswitch/blob/f624eb52d61561c365cce21e58b08281d096d904/crates/router/src/core/errors.rs#L307

crates/router/src/core/payment_methods/cards.rs
```diff
@@ -654,7 +654,7 @@ pub async fn mock_get_card<'a>(
         card_fingerprint: locker_mock_up.card_fingerprint.into(),
         card_global_fingerprint: locker_mock_up.card_global_fingerprint.into(),
         merchant_id: Some(locker_mock_up.merchant_id),
-        card_number: Some(locker_mock_up.card_number.into()),
+        card_number: Some(locker_mock_up.card_number.try_into().unwrap()),
```
Narayanbhat166, line 657: Same here, You can raise the same error as mentioned above.

SanchithHegde:
> Does this mean I leave them in for now and you take care of this?

You can either leave it in and one of us will take care of removing the test case for each of the connectors (and the connector template), or you can remove it in this PR. I'm fine with either.

crates/router/src/core/payment_methods/cards.rs
Narayanbhat166: Add an `attach_printable` to provide clear context as to what failed here. In this case the printable message would be `Invalid card number format from the mock locker`.

crates/router/src/connector/authorizedotnet/transformers.rs
Narayanbhat166: In case of authorizedotnet, the card number that we receive in the response will be masked already, it will be in the format `1234XX`. Validating this will always fail. You should change the type of card number ( only for `authorizedotnet` ) to not have any validation. 

This case is special for this connector because it expects the first four digits of the card number to be sent in the `refunds` call. In `payments` response it sends a masked card number which we need to send in the refunds call. So we save this data in the `payment_attempt`.

crates/router/src/core/payment_methods/cards.rs
phillyphil91: Like this?
```rust
card_number: { match response.card_number.try_into() {                                                                                                                
               Ok(ccnum) => Some(ccnum),                                                                                       
               Err(_) => return Err(errors::VaultError::ResponseDeserializationFailed).into_report().attach_printable("Invalid card number format from the mock locker")
            }},
```
I have not yet dealt with the `error_stack` crate which I think is where the `attach_printable` method comes from, correct? 
I can also implement `From<CCValError>` for `VaultError` but I felt like this would be kind of displaced. As there are no other `From` implementations for `VaultError`

phillyphil91:
> You can either leave it in and one of us will take care of removing the test case for each of the connectors (and the connector template), or you can remove it in this PR. I'm fine with either.

I can remove them in this PR

crates/router/src/core/payment_methods/cards.rs
SanchithHegde: You can do something like this. This would create a `Report` which contains both the initial error (`CCValError` in this case) and the error we'd like to raise (`VaultError::ResponseDeserializationFailed`) from this function.

```rust
card_number: match response.card_number.try_into() {                                                                                                                
                Ok(ccnum) => Some(ccnum),                                                                                       
                error => return error.into_report()
                                .change_context(errors::VaultError::ResponseDeserializationFailed)
                                .attach_printable("Invalid card number format from the mock locker")
            },
```

crates/cards/src/validate.rs
```diff
@@ -73,7 +68,7 @@ where
             return WithType::fmt(val, f);
         }
 
-        write!(f, "{}{}", &val_str[..6], "*".repeat(val_str.len() - 6))
+        write!(f, "{}{}", &val_str[..5], "*".repeat(val_str.len() - 5))
```
SanchithHegde, line 71: Is there any specific reason for this number to be decremented to 5?

crates/router/src/connector/utils.rs
```diff
@@ -282,7 +282,7 @@ impl CardData for api::Card {
         Secret::new(year[year.len() - 2..].to_string())
     }
     fn get_card_issuer(&self) -> Result<CardIssuer, Error> {
-        let card: cards::CardNumber = self.card_number.clone().remove_whitespace();
+        let card: cards::CardNumber = self.card_number.clone();
         get_card_issuer(card.peek().clone().as_str())
```
SanchithHegde, line 286: You should be able to do this and eliminate the clones?

```rust
get_card_issuer(self.card_number.peek())
```

crates/router/src/core/payment_methods/cards.rs
```diff
@@ -627,7 +627,16 @@ pub async fn mock_add_card(
...
+        card_number: match response
+            .card_number
+            .try_into()
+            .into_report()
+            .change_context(errors::VaultError::ResponseDeserializationFailed)
+            .attach_printable("Invalid card number format from the mock locker")
+        {
+            Ok(ccnum) => Some(ccnum),
+            Err(e) => return Err(e),
+        },
```
SanchithHegde, line 639:
```suggestion
        card_number: response
            .card_number
            .try_into()
            .into_report()
            .change_context(errors::VaultError::ResponseDeserializationFailed)
            .attach_printable("Invalid card number format from the mock locker")
            .map(Some)?,
```

crates/router/src/core/payment_methods/cards.rs
```diff
@@ -654,7 +663,16 @@ pub async fn mock_get_card<'a>(
...
+        card_number: match locker_mock_up
+            .card_number
+            .try_into()
+            .into_report()
+            .change_context(errors::VaultError::ResponseDeserializationFailed)
+            .attach_printable("Invalid card number format from the mock locker")
+        {
+            Ok(ccnum) => Some(ccnum),
+            Err(e) => return Err(e),
+        },
```
SanchithHegde, line 675: You can perform a similar conversion here.

```suggestion
        card_number: locker_mock_up
            .card_number
            .try_into()
            .into_report()
            .change_context(errors::VaultError::ResponseDeserializationFailed)
            .attach_printable("Invalid card number format from the mock locker")
            .map(Some)?,
```

crates/router/tests/connectors/adyen.rs
```diff
@@ -358,27 +358,6 @@ async fn should_fail_payment_for_incorrect_card_number() {
     );
 }
 
-// Creates a payment with empty card number.
```
SanchithHegde, line 359: Please remove this test from the connector template as well:

https://github.com/juspay/hyperswitch/blob/17e8305a5de49a37aca9205f4dcdc8765c7cd0ec/connector-template/test.rs#L305-L326

SanchithHegde:
Once you're done with the suggested changes, please resolve merge conflicts with the `main` branch. CI checks seem to be blocked on it.

crates/router/src/core/payment_methods/cards.rs
phillyphil91: good call! :+1:

crates/router/src/connector/paypal/transformers.rs
```diff
@@ -9,7 +9,10 @@ use crate::{
         PaymentsAuthorizeRequestData,
     },
     core::errors,
+<<<<<<< Updated upstream
     pii, services,
+=======
+>>>>>>> Stashed changes
```
SanchithHegde, line 15: There's been a minor derp during conflict resolution.

SanchithHegde (commented):
Also, please address the CI checks.

crates/router/src/connector/checkout/transformers.rs
```diff
@@ -95,7 +95,7 @@ impl<F, T>
 pub struct CardSource {
     #[serde(rename = "type")]
     pub source_type: CheckoutSourceTypes,
-    pub number: pii::Secret<String, pii::CardNumber>,
+    pub number: Option<cards::CardNumber>,
```
SanchithHegde, line 98: Why was this updated to be an `Option`? Could you revert this to be a required value instead?

phillyphil91: not sure to be honest. there were so many files that I had to change that I lost overview a bit and this might have slipped in. Thanks for catching that.

phillyphil91:
Hi @SanchithHegde 
I hope everything is fixed now. It was a bit tough keeping up with all the changes on `main`. Let me know if there is something missing.

crates/cards/Cargo.toml
```diff
@@ -12,10 +12,13 @@ default = ["serde"]
 time = { version = "0.3.20" }
 error-stack = "0.3.1"
 serde = { version = "1", features = ["derive"], optional = true }
+serde_json = "1.0.96"
```
SanchithHegde, line 15: Do we need `serde_json` to be among our dependencies, or would it suffice to have it in `dev-dependencies`?

SanchithHegde (commented):
> I hope everything is fixed now. It was a bit tough keeping up with all the changes on main.

Everything else looks good. Thanks for keeping up with the changes on the `main` branch. I can take it up from here, if you'd prefer that.

crates/cards/Cargo.toml
phillyphil91: > Everything else looks good. Thanks for keeping up with the changes on the `main` branch. I can take it up from here, if you'd prefer that.

there are still some CI errors that I am not too sure about as I don't get these locally.

SanchithHegde: I'll handle it then, no worries!

phillyphil91: thank you. and I believe `serde_json` is sufficient in the `dev-dependencies`

vspecky (dismissed):

phillyphil91:
Hi @SanchithHegde 
Just making sure: There is nothing needed from my side for this to be merged right?

SanchithHegde:
> Hi @SanchithHegde Just making sure: There is nothing needed from my side for this to be merged right?

@phillyphil91 Yup, I'll handle merge conflicts and getting it merged in.

jarnura (approved):

vspecky (approved):

vspecky (approved):

jarnura (approved):

phillyphil91:
It was a lot of fun @SanchithHegde

PR #904, opened 2023-04-18
refactor: use `CountryAlpha2` instead of `CountryCode` for country codes

pixincreate:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [x] Refactoring
- [ ] Dependency updates

## Description

Refactor countryCode for `CountryAlpha2` and remove the existing enum `CountryAlpha2` to reduce redundant code.

It is recommended to clean up the local database once merged by running `cargo clean` to avoid errors.

### Additional Changes

## Motivation and Context

Removes the redundant code to tidy the `enums.rs`

## How did you test it?

crates/common_enums/src/transformers.rs   
image

image

## Checklist


jarnura (commented):
Add labels and all required reviewers

crates/api_models/src/payment_methods.rs
```diff
@@ -284,7 +284,7 @@ pub struct PaymentMethodListRequest {
 
     /// The two-letter ISO currency code
     #[schema(value_type = Option<Vec<Country>>, example = json!(["US", "UK", "IN"]))]
```
Narayanbhat166, line 286: Change the type here to `CountryAlpha2`

crates/api_models/src/payments.rs
```diff
@@ -203,7 +203,7 @@ pub struct PaymentsRequest {
 
     /// Business country of the merchant for this payment
     #[schema(example = "US")]
```
Narayanbhat166, line 205: add value_type = CountryAlpha2 here. The idea is to add this wherever the enum/struct is imported from a module, value_type has to be added.

crates/api_models/src/payments.rs
```diff
@@ -709,8 +709,8 @@ pub struct AddressDetails {
     pub city: Option<String>,
 
     /// The two-letter ISO country code for the address
-    #[schema(value_type = Option<Country>, max_length = 2, min_length = 2, example = "US")]
-    pub country: Option<api_enums::CountryCode>,
+    #[schema(value_type = Option<CountryAlpha2>, max_length = 2, min_length = 2, example = "US")]
```
Narayanbhat166, line 712: the min and max length can be removed from this, as this has been converted to enum type

Narayanbhat166 (changes requested):
In `api_models` wherever `CountryAlpha2` is used, add 
```rust
#[schema(value_type = ..example = "US")]
```
The value type can be either Option<CountryAlpha2> or Vec<CountryAlpha2> based on the usage.

ArjunKarthik:
@pixincreate Please run unit tests and attach screenshot in the PR

ArjunKarthik:
@pixincreate I couldn't find any commits which actually resolved @Narayanbhat166 comments

ArjunKarthik:
> In `api_models` wherever `CountryAlpha2` is used, add
> 
> ```rust
> #[schema(value_type = ..example = "US")]
> ```
> 
> The value type can be either Option or Vec based on the usage.

@pixincreate as per @Narayanbhat166 comment you have to add #[schema(value_type = ..example = "US")] to all the fields which uses CountryAlpha2 inside api_models. Please make sure you are following the same in all the 3 files [admin.rs](https://github.com/juspay/hyperswitch/pull/904/files#diff-c50d31f1c8ecb900891725f09240e0b31819cc74ddc380905bd8aafad949d30f), [payment_methods.rs](https://github.com/juspay/hyperswitch/pull/904/files#diff-34595e668b14ba308e1a500bc2091b66c87df3c0ad9d224bf799f58be6a1ddd2), [payments.rs](https://github.com/juspay/hyperswitch/pull/904/files#diff-460b9de06d0bc282f72b2b826d1516d58aa4a4e67c0004bea409664a9e2fa375) in api_models

Narayanbhat166 (dismissed):
Looks good to me. Generate the openapi spec using this command 
`cargo r --features openapi -- generate-openapi-spec` and ensure that there are no errors in the `generated.json` file

crates/router/src/connector/applepay/transformers.rs
jarnura, line 1: Why this file is required for this changes?

If this accidental please remove.

jarnura (changes requested):

jarnura (approved):

crates/storage_models/src/enums.rs
```diff
@@ -3,7 +3,7 @@ pub mod diesel_exports {
     pub use super::{
         DbAttemptStatus as AttemptStatus, DbAuthenticationType as AuthenticationType,
         DbCaptureMethod as CaptureMethod, DbConnectorType as ConnectorType,
-        DbCountryCode as CountryCode, DbCurrency as Currency, DbDisputeStage as DisputeStage,
+        DbCountryAlpha2 as CountryAlpha2, DbCurrency as Currency, DbDisputeStage as DisputeStage,
```
SanchithHegde, line 6: As far as I understand, this will cause the `schema.rs` file to go to an inconsistent state (and possibly fail to build) when you run `diesel migration run` on a fresh clone of the repository, since the database type would be named `CountryCode`, while our code expects it to be `CountryAlpha2`.

There are a few ways to handle this, I feel the simplest way is to just rename the database type to be also called `CountryAlpha2`.

SanchithHegde (commented):

crates/storage_models/src/enums.rs
pixincreate: So `DbCountryCode as CountryAlpha2` would have been better if we're not renaming in the database?  
Also for some reason, on my local, it shows 60+ problems with this branch while it passes all the test cases here. `cargo clean` seems to fix the issue locally btw.

Narayanbhat166: Yes you can add a migration for this. It's always better to have proper naming and be explicit rather than having the developer make an effort to look up the database type to know its details.

migrations/2023-04-25-061159_rename_country_code_to_country_alpha2/down.sql
```diff
@@ -0,0 +1,2 @@
+-- This file should undo anything in `up.sql`
+ALTER TYPE "CountryAlpha2" RENAME TO "CountryCode"
```
SanchithHegde, line 2:
```suggestion
ALTER TYPE "CountryAlpha2" RENAME TO "CountryCode";
```

migrations/2023-04-25-061159_rename_country_code_to_country_alpha2/up.sql
```diff
@@ -0,0 +1,2 @@
+-- Your SQL goes here
+ALTER TYPE "CountryCode" RENAME TO "CountryAlpha2"
```
SanchithHegde, line 2:
```suggestion
ALTER TYPE "CountryCode" RENAME TO "CountryAlpha2";
```

SanchithHegde (dismissed):

ArjunKarthik (dismissed):

jarnura (approved):

SanchithHegde (approved):

jarnura (approved):

PR #908, opened 2023-04-18
feat: add relevant ids for payment calls & make json logs 

NishantJoshi00:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates

## Description
The PR introduces 2 changes
- Adds `payment_id` and `merchant_id` for any payment_core call
- putting start/end span json log events under feature flag, (currently set to disabled)

### Additional Changes

## Checklist


crates/router/src/core/payments.rs
```diff
@@ -67,7 +68,9 @@ where
         .to_validate_request()?
         .validate_request(&req, &merchant_account)?;
 
-    tracing::Span::current().record("payment_id", &format!("{:?}", validate_result.payment_id));
+    tracing::Span::current()
+        .record("payment_id", &format!("{}", validate_result.payment_id))
+        .record("merchant_id", merchant_account.merchant_id.as_str());
```
vspecky, line 73: I think better to record the `merchant_id` before the `validate_request()` call, at the top of the function since its immediately available. So even if `validate_request()` throws an error, we can still check which account the request was made from.

Actually, the same can be done for `payment_id` as well right?

vspecky (changes requested):

crates/router/src/core/payments.rs
NishantJoshi00: For mandate_id, I was also thinking about adding it in the `payments_core` function itself, will do that. But, in case of `payment_id` that won't be possible as in cases like `PaymentCreate` the `payment_id` in only available after validate request, as it is being generated there

jarnura (approved):

vspecky (approved):

SanchithHegde (approved):

lsampras (approved):

PR #911, opened 2023-04-18
refactor(errors): add parsing error types for context info

lsampras:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring
- [ ] Dependency updates

## Description

- Update parsing error to an enum to provide more context about the type of parsing failure...

### Additional Changes

## Checklist


crates/common_utils/src/errors.rs
```diff
@@ -8,36 +8,23 @@
...
-    };
+/// Parsing Errors
+#[derive(Debug, thiserror::Error)]
+pub enum ParsingError {
+    ///Failed to parse enum
+    #[error("Failed to parse enum: {0}")]
+    EnumParseFailure(&'static str),
+    ///Failed to parse struct
+    #[error("Failed to parse struct: {0}")]
+    StructParseFailure(&'static str),
```
SanchithHegde, line 19: Do you think we should call these two together as deserialization failure?

So, if a parsing error occurs, it's either during serialization (encoding), deserialization, or an unknown error?

lsampras: The reason for 2 different types is different error messages.

if we club them into a single deserialization variant
say something like

```
#[error("Failed to deserialize {0}")]
DeserializeFailure(&'static str)
```

We lose some info about whether it was enum or struct,
Also I think we should be able to extend the parse errors to parsing beyond serde...

SanchithHegde: Fair enough.

SanchithHegde (dismissed):

jarnura (dismissed):

Sangamesh26 (approved):

PR #927, opened 2023-04-20
feat(pm_list): add support for bank debits

Narayanbhat166:
## Type of Change

- [x] New feature

## Description

This PR adds support for `bank_redirects` in payment methods list. The response will be of the form 
```json
{
  "payment_method": "bank_debit",
  "payment_method_types": [
    {
      "payment_method_type": "sepa",
      "payment_experience": null,
      "card_networks": null,
      "bank_names": null,
      "bank_debits": {
        "eligible_connectors": [
          "stripe",
          "adyen"
        ]
      }
    },
    {
      "payment_method_type": "ach",
      "payment_experience": null,
      "card_networks": null,
      "bank_names": null,
      "bank_debits": {
        "eligible_connectors": [
          "adyen"
        ]
      }
    }
  ]
}
```

## How did you test it?

- Create a payment connector with `sepa` and `ach` enabled.
- Create a payment with confirm set to false
- Call the pm list.

## Checklist

### DIFF
--- a/crates/api_models/src/payment_methods.rs
+++ b/crates/api_models/src/payment_methods.rs
@@ -183,6 +183,11 @@ pub struct CardNetworkTypes {
 }
 
 #[derive(Debug, Clone, serde::Serialize, serde::Deserialize, ToSchema, PartialEq, Eq)]
+pub struct BankDebitTypes {
+    pub eligible_connectors: Vec<String>,
+}
+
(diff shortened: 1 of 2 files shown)


jarnura (commented):
Avoid using if let some pattern for case match scenarios, use  if let some pattern if only the any one variant needs to be handled

Narayanbhat166:
In this PR, if let Some(_) is used only on `Option<_>` where only one case needs to be handled. The consolidation logic will be removed in the upcoming days.

jarnura (approved):

PR #951, opened 2023-04-21
refactor: accept any field in mca update for backwards compatibality

Narayanbhat166:
## Type of Change

- [x] New feature

## Description

Some fields which are indexed fields are not allowed to be updated. But this change will break the changes for the ones using these apis before.

## How did you test it?

Manual, Compiler guided.

## Checklist


Abhicodes-crypto (changes requested):
The previous fields can be added in this struct and  `deny_unknown_fields` can be left as it is.

Narayanbhat166:
Closing this branch as the tag is recent, should have created the branch from 0.5.6 tag

Issue #920, opened 2023-04-19
[FEATURE] Add Refund Sync route in compatibiltiy


Abhicodes-crypto:
### Feature Description

There is a refund sync route https://github.com/juspay/hyperswitch/blob/573a4d384ee6a9d72648ab537804799a3993e1e8/crates/router/src/routes/app.rs#L184 

A similar route needs to be added in the compatibility layer in this file - https://github.com/juspay/hyperswitch/blob/main/crates/router/src/compatibility/stripe/refunds.rs 

### Possible Implementation

refer similar route of payments as an example. - https://github.com/juspay/hyperswitch/blob/573a4d384ee6a9d72648ab537804799a3993e1e8/crates/router/src/compatibility/stripe/app.rs#L12 

### Have you spent some time to check if this feature request has been raised before?

- [X] I checked and didn't find similar issue

### Have you read the Contributing Guidelines?

- [X] I have read the [Contributing Guidelines](https://github.com/juspay/hyperswitch/blob/main/docs/CONTRIBUTING.md)

### Are you willing to submit a PR?

None

shubhanshu-mani:
Hi I would like to pick this up , though I need some additional details on this. 

1. How is refund sync different from refund-retrieve https://github.com/juspay/hyperswitch/blob/573a4d384ee6a9d72648ab537804799a3993e1e8/crates/router/src/compatibility/stripe/app.rs#L41
2. Do you want something like refund_retrieve_with_gateway_creds

Abhicodes-crypto:
> How is refund sync different from refund-retrieve

It calls the same function of refund sync. Only difference is that in the request body the gateway creds are passed.

> Do you want something like refund_retrieve_with_gateway_creds

Yes.

shubhanshu-mani:
Hey do I need to fork this repo and create a PR from that? Or are people allowed to push branches directly

SanchithHegde:
> Hey do I need to fork this repo and create a PR from that? Or are people allowed to push branches directly

Yes @shubhanshu-mani, you can fork the repository and open a PR from your fork. Users who aren't members of the `juspay` GitHub organization aren't allowed to create branches on our repository.

shubhanshu-mani:
@SanchithHegde sure I have created the PR from fork . BTW my account is part of juspay organisation don't know why it was allowing me to push branch directly

SanchithHegde:
> BTW my account is part of juspay organisation don't know why it was not allowing me to push branch directly

[...]

PR #958, opened 2023-04-22
feat(compatibility): add refund retrieve endpoint which accepts gateway creds

shubhanshu-mani:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

This PR is based on the issue #920 to add refund sync with gateway creds in stripe compatibility.  
Adds a new service  refund_retrieve_with_gateway_creds which parses the JSON to type RefundsRetrieveRequest which already has an optional field merchant_connector_details to accept gateway credentials and then calls the normal refund retrieve API

### Additional Changes

## Checklist


crates/router/src/compatibility/stripe/refunds.rs
```diff
@@ -49,6 +49,47 @@ pub async fn refund_create(
     .await
 }
 
+#[instrument(skip_all)]
+#[get("/{refund_id}")]
```
SanchithHegde, line 53: Would this have to be `/sync` @Abhicodes-crypto? As far as I remember, trying to register two handlers for the same route with the same HTTP method would cause the Actix server to panic at startup.

SanchithHegde (commented):

crates/router/src/compatibility/stripe/refunds.rs
Abhicodes-crypto, line 53: The route should be `post("/sync")`.

Abhicodes-crypto (changes requested):
Route change requested.

crates/router/src/compatibility/stripe/refunds.rs
shubhanshu-mani: Done

SanchithHegde (approved):

jarnura (approved):

PR #963, opened 2023-04-24
feat(validate): implementing validators for mca and payment list

NishantJoshi00:
…st-meta-validate## Type of Change

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

This PR adds validation for payment method specific behaviour. This validation is applied while creating MCA and also while filtering payment method in payment method list API

### Additional Changes

## How did you test it?
Compiler guided

## Checklist


crates/router/src/connector/adyen.rs
```diff
@@ -801,3 +802,103 @@ impl api::IncomingWebhook for Adyen {
...
+            payment_method
+        {
+            let _inner: api_models::payments::GpaySessionTokenData = metadata
+                .parse_value("GpayTokenParameters")
+                .change_context(errors::ValidationError::InvalidValue {
+                    message: "Failed while getting google pay metadata".to_string(),
+                })?;
+        }
+
+        Ok(())
```
Abhicodes-crypto, line 822: Seems like this implementation is same in most the places so can it be the default impl in the trait ?

jarnura (commented):
We should not call DB calls just for validation.

Remove all DB calls which made for validations

crates/router/src/connector/adyen.rs
NishantJoshi00: True, but we cannot have this as default as the validation that is done here is specific to a connector. This was decided based on the features that are implemented or possible for a certain connector

NishantJoshi00:
@jarnura , I agree that having DB calls in validation isn't performant, but in case of `adyen` and `cybersource` we need to validate a field in the address. That's the reason why, the DB calls were present in the mentioned connector. In order to remove those DB calls we will remove the validation itself.
@bernard-eugine any thoughts on this, I agree with @jarnura about not having redundant DB calls, but then that will not allow us to do validation for `Authorize` in payment method list

NishantJoshi00:
Resolved by adding hint based validations by @Chethan-rao. This change no longer required

PR #964, opened 2023-04-24
feat(payment_request): add field `amount` to `OrderDetails` and make `order_details` a `Vec` in `payments_create` request

rishavkar:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

Made changes to order_details, to contain an amount(per quantity of the product) field.
Also order_details is now a vector
Why the change?
the order can have multiple products, with unique prices and quantity...the old structure didnt accomodate for it 

### Additional Changes

Payments create request, will have the order_details field in metadata inside.
```
"metadata": {
    "order_details" : [{
        "product_name" : "gillete creme",
        "quantity" : 2,
        "amount" : 600
    },
    {
        "product_name" : "gillete razor",
        "quantity" : 1,
        "amount" : 300
    }],
    "udf1": "value1",
    "new_customer": "true",
    "login_date": "2019-09-10T10:11:12Z"
  }
```
Metadata Type in DB will also change, as it will contain order_details as a vector : 
```
pub struct Metadata {
    /// Information about the product and quantity for specific connectors. (e.g. Klarna)
    pub order_details: Option<Vec<OrderDetails>>,
    /// Any other metadata that is to be provided
    #[schema(value_type = Object, example = r#"{ "city": "NY", "unit": "245" }"#)]
    #[serde(flatten)]
    pub data: pii::SecretSerdeValue,
    /// Payload coming in request as a metadata field
    pub payload: Option<pii::SecretSerdeValue>,
    /// Allowed payment method types for a payment intent
    #[schema(value_type = Option<Vec<PaymentMethodType>>)]
    pub allowed_payment_method_types: Option<Vec<api_enums::PaymentMethodType>>,
}
```

## Checklist


Narayanbhat166:
@rishavkar Can you please follow the PR title standards?

crates/router/src/connector/klarna/transformers.rs
```diff
@@ -49,9 +49,12 @@ impl TryFrom<&types::PaymentsSessionRouterData> for KlarnaSessionRequest {
...
                 order_lines: vec![OrderLines {
-                    name: order_details.product_name,
-                    quantity: order_details.quantity,
-                    unit_price: request.amount,
+                    name: order_details
+                        .iter()
+                        .map(|v| v.product_name.clone())
+                        .collect(),
+                    quantity: order_details.iter().map(|v| v.quantity).collect(),
+                    unit_price: order_details.iter().map(|v| v.amount).collect(),
```
Abhicodes-crypto, line 57: Should the vector be populated with each order_detail ? Right now I guess it would populate it as `[{"name":["name1", "name2"]}]` .

cc: @Narayanbhat166

Narayanbhat166: Yes, you are right @Abhicodes-crypto. According to their [docs](https://docs.klarna.com/api/payments/#operation/createOrder!path=order_lines&t=request) this should be the vector of order lines.

crates/router/src/connector/klarna/transformers.rs
```diff
@@ -94,9 +97,12 @@ impl TryFrom<&types::PaymentsAuthorizeRouterData> for KlarnaPaymentsRequest {
...
```
Narayanbhat166, line 105: The same applies here too. It should be `Vec<OrderDetails>`

crates/router/src/connector/adyen/transformers.rs
```diff
@@ -788,12 +788,13 @@ fn get_line_items(item: &types::PaymentsAuthorizeRouterData) -> Vec<LineItem> {
     let line_item = LineItem {
         amount_including_tax: Some(item.request.amount),
         amount_excluding_tax: Some(item.request.amount),
-        description: order_details.map(|details| details.product_name.clone()),
+        description: order_details
+            .map(|details| details.iter().map(|v| v.product_name.clone()).collect()),
```
Narayanbhat166, line 792: I guess this also should be sent as a `Vec<LineItem>`. Can @ArjunKarthik @jagan-jaya confirm?

Narayanbhat166 (changes requested):
Once the changes are made, please test the connectors which are impacted by this change.

crates/router/src/connector/adyen/transformers.rs
```diff
@@ -836,18 +836,31 @@ fn get_address_info(address: Option<&api_models::payments::Address>) -> Option<A
...
+            })
+            .collect(),
+        None => {
+            let line_item = LineItem {
+                amount_including_tax: Some(item.request.amount),
+                amount_excluding_tax: Some(item.request.amount),
+                description: None,
+                id: Some(String::from("Items #1")),
+                tax_amount: None,
+                quantity: None,
```
Narayanbhat166, line 859: This should be 1 by default i guess by looking at the previous code.

crates/router/src/connector/zen/transformers.rs
```diff
@@ -203,12 +203,21 @@ impl TryFrom<&types::PaymentsAuthorizeRouterData> for ZenPaymentsRequest {
...
+                    price: data.amount.to_string(),
+                    line_amount_total: order_amount.clone(),
+                })
+                .collect(),
+            // items: vec![ZenItemObject {
+            //     name: order_details.product_name,
+            //     price: order_amount.clone(),
+            //     quantity: 1,
+            //     line_amount_total: order_amount,
+            // }],
```
Narayanbhat166, line 220: If this is not required, then please remove it

Narayanbhat166:
The amount field in `OrderDetails` struct should contain the unit amount or total amount? If it is unit amount, then can the field name be changed so that it is clear whether to pass the total amount ( quantity * unit price ) or just unit_price ?

crates/router/src/connector/adyen/transformers.rs
rishavkar: changed to :
`quantity: Some(1),`

crates/router/src/connector/zen/transformers.rs
rishavkar: done

Narayanbhat166 (approved):

crates/api_models/src/payments.rs
```diff
@@ -1410,12 +1410,14 @@ pub struct OrderDetails {
...
     pub quantity: u16,
+    /// the amount per quantity of product
+    pub amount: i64,
 }
 
 #[derive(Default, Debug, Eq, PartialEq, serde::Deserialize, serde::Serialize, Clone, ToSchema)]
 pub struct Metadata {
     /// Information about the product and quantity for specific connectors. (e.g. Klarna)
-    pub order_details: Option<OrderDetails>,
+    pub order_details: Option<Vec<OrderDetails>>,
```
jagan-jaya, line 1420: is this tested for both klarna and zen?

jagan-jaya:
If the testing is done for Klarna and zen, then it is good to me

crates/api_models/src/payments.rs
Narayanbhat166: This is tested for klarna and adyen, @rishavkar can confirm if this is tested for zen connector

jagan-jaya: What happens to existing orders in sandbox? it will still try to fetch as single OrderDetails right?
don't have to write migration to fix that?

rishavkar: - i have tested for zen and adyen
- sandbox has entries with order_details as objects....so migration would be needed ideally.... would i have to add the migration script in this PR? 
@jagan-jaya

jarnura: Yes, Add the migration and steps needs to done before and after the deployment.

Is the changes are back-compatible?

migrations/2023-05-12-103127_payment_intent.sql/up.sql
```diff
@@ -0,0 +1,7 @@
+ALTER TABLE payment_intent ADD COLUMN meta_data jsonb;
```
Narayanbhat166, line 1: There already is a column called metadata for this table, can you use the same column and not create another instead.

jagan-jaya: @Narayanbhat166 this new column is added for backward_compatibility. once things are stable, we will remove old one and update the existing one

jagan-jaya (dismissed):

jagan-jaya (approved):

Narayanbhat166 (approved):

jarnura (approved):

PR #965, opened 2023-04-24
feat(pm_list): Add support for bank transfers in pm list

Sangamesh26:
## Type of Change

- [x] New feature

## Description

This PR adds support for bank_transfer in payment methods list. The response will be of the form

```
        {
            "payment_method": "bank_transfer",
            "payment_method_types": [
                {
                    "payment_method_type": "ach",
                    "payment_experience": null,
                    "card_networks": null,
                    "bank_names": null,
                    "bank_transfers": {
                        "eligible_connectors": [
                            "stripe"
                        ]
                    }
                }
            ]
        }
```

## How did you test it?

* Create a payment connector with ach enabled.
* Create a payment with confirm set to false
* Call the pm list.

## Checklist


crates/router/src/core/payment_methods/cards.rs
```diff
@@ -954,6 +960,28 @@ pub async fn list_payment_methods(
...
+        if element.payment_method == api_enums::PaymentMethod::BankRedirect {
+            let connector = element.connector.clone();
+            if let Some(vector_of_connectors) =
+                banks_consolidated_hm.get_mut(&element.payment_method_type)
+            {
+                vector_of_connectors.push(connector);
+            } else {
+                banks_consolidated_hm.insert(element.payment_method_type, vec![connector]);
+            }
+        }
```
Abhicodes-crypto, line 973: This code seems to be the same as code from line 953 to 962. Was this accidentally committed.

Abhicodes-crypto: Also can this be broken down into smaller functions as there is too much nesting. `#[inline]` can be done on those smaller functions.

Narayanbhat166 (approved):

PR #966, opened 2023-04-24
refactor(core): in PaymentData and RouterData make flow as concrete type

Abhicodes-crypto:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [x] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

There is a generic type `F` which has been used across the codebase. It represents Flow. In this PR the following changes are made :
- Marker trait `Flow` is created and all the entities using `F` are constrained by `F: Flow`.
- Trait Flow implements Debug additionally. 
- A derive macro for Flow has been implemented and used wherever appropriate.
- If we want Flow as a string we use `format!("{:?}", F::default())),`, see `crates/router/src/services/api.rs:193` . Edit now it is changed to `F::default().to_string()`

While making the above changes I observed that `crates/router/src/connector/utils.rs:53` There was a trait RouterData which had getter methods on RouterData. So I removed the trait and instead made it `impl RouterData`. Let me know if this change needs to be reverted.

Also derive Debug impl seems to be unstable across rustc versions https://doc.rust-lang.org/std/fmt/trait.Debug.html#stability . So I will impl Display on Flow in the next commit. Edit: Done.

Having marker trait flow makes it easy to add constraints on `Flow` as we just have to add the constraint in one place and impl  the trait wherever necessary. 

Have added the ZDisplay macro which implements Display on unit structs using their name. 

Enhanced `Flow` macro to also do the same functionality as ZDisplay while also adding the marker trait flow.

`Debug` is impl on Operation as well and we are using it in payments core to do a match on the struct. Will make it `to_string()` as well using ZDisplay. Edit : Done

### Additional Changes

## How did you test it?

It compiled. But let me know if I have to test any specific flow.

## Checklist


crates/router/src/connector/utils.rs
```diff
@@ -47,32 +50,32 @@ impl AccessTokenRequestInfo for types::RefreshTokenRouterData {
...
-    fn to_connector_meta<T>(&self) -> Result<T, Error>
-    where
-        T: serde::de::DeserializeOwned;
-    fn is_three_ds(&self) -> bool;
-    fn get_payment_method_token(&self) -> Result<String, Error>;
-}
-
-impl<Flow, Request, Response> RouterData for types::RouterData<Flow, Request, Response> {
-    fn get_billing(&self) -> Result<&api::Address, Error> {
+// pub trait RouterData {
```
jarnura, line 53: Remove comment code

jarnura (changes requested):
Looks good to me :)

Narayanbhat166 (approved):

Narayanbhat166 (dismissed):

ashokkjag:
@Abhicodes-crypto , can you check and rebase. can't have 406 files changed.

Narayanbhat166:
Closing this PR since this change is no longer required

PR #974, opened 2023-04-25
fix(refund_list): updated refund list response status code when no refunds found.

hrithikesh026:
## Type of Change

- [x] Bugfix

## Description

* Changed `refund_list()` function in `crates/router/src/core/refunds.rs`  such that it returns only Ok response with Vec of refunds.
* Added  `size: usize` field to `RefundListResponse` in `crates/api_models/src/refunds.rs` just like in `PaymentListResponse`.

## Motivation and Context
"Get Refund List" would send an error response with status code 400 when no refunds are found.
Fixed it so it returns an empty list with status code 200 like "Get Payment List".

## How did you test it?

Manual

## Checklist

### DIFF
--- a/crates/api_models/src/refunds.rs
+++ b/crates/api_models/src/refunds.rs
@@ -144,8 +144,9 @@ pub struct RefundListRequest {
 
 #[derive(Debug, Clone, Eq, PartialEq, Deserialize, Serialize, ToSchema)]
 pub struct RefundListResponse {
-    /// The list of refund response
+    /// The number of refunds included in the list
     pub size: usize,
+    /// The List of refund reponse object
     pub data: Vec<RefundResponse>,
 }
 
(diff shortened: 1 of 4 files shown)


crates/api_models/src/refunds.rs
```diff
@@ -145,6 +145,7 @@ pub struct RefundListRequest {
 #[derive(Debug, Clone, Eq, PartialEq, Deserialize, Serialize, ToSchema)]
 pub struct RefundListResponse {
     /// The list of refund response
+    pub size: usize,
```
Abhicodes-crypto, line 148: Fix schema

Abhicodes-crypto (changes requested):
Schema change requested. Other than that LGTM.

Abhicodes-crypto (approved):

SanchithHegde (approved):

jarnura (approved):

PR #976, opened 2023-04-25
fix(storage_models): fix incorrect field order in `MerchantConnectorAccount`

SanchithHegde:
## Type of Change

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

This PR fixes the `schema.rs` file being inconsistent (incorrect field order) when `diesel migration run` was run on a fresh clone of the repository and with a fresh database.

## Motivation and Context

Without this fix, if migrations are run on a fresh clone of the repository, the code fails to build because of the inconsistent field order, and the error messages we receive are quite cryptic. The only fix was to revert changes to the `schema.rs` file, but that is not a good experience for external contributors.

## How did you test it?

Ensured that the fields in `MerchantConnectorAccount` have the expected values.

## Checklist


Abhicodes-crypto (dismissed):

Narayanbhat166 (dismissed):
I believe, for someone running only few migrations ( incremental ones ) this would produce the same results, right?

SanchithHegde:
> I believe, for someone running only few migrations ( incremental ones ) this would produce the same results, right?

It should, provided they run migrations each time they pull the `main` branch.

jarnura (dismissed):

dracarys18 (approved):

jarnura (approved):

PR #999, opened 2023-04-26
feat(pm_list): add available capture methods filter

prajjwalkumar17:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [x] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description
We are adding a filter for the payment methods list that whenever the capture method on payment_intent is set to Manual it will evaluate whether that connector issues for the manual capture of payment or not. If not it won't be visible in payments method list

### Additional Changes

## Motivation and Context\
As of now even if manual capture is not present for a connector, our route was showing the option from now it won't happen and only the methods allowed for manual capture will be visible to choose from

## How did you test it?
When we will make the payment create in manual mode we  will won't get the cards available as it don't supports manual capture

Screenshot 2023-04-26 at 18 26 15

## Checklist


.vscode/settings.json
SanchithHegde, line 1: Please untrack this file.

config/development.toml
```diff
@@ -197,3 +221,4 @@ region = ""
 [tokenization]
 stripe = { long_lived_token = false, payment_method = "wallet"}
 checkout = { long_lived_token = false, payment_method = "wallet"}
+        #ACI,BrainTree,Klarna,Mollie,MultisafePay,Trustpay,Zen
```
SanchithHegde, line 224: I believe you should be able to remove this comment?

crates/router/src/core/payment_methods/cards.rs
```diff
@@ -1184,6 +1192,19 @@ fn filter_pm_based_on_config<'a>(
         .unwrap_or(true)
 }
 
+fn filter_for_capture_method(
```
SanchithHegde, line 1195: This function could either use a better name or have a brief documentation as to what is does.

crates/common_enums/src/enums.rs
```diff
@@ -41,6 +41,11 @@ pub enum CountryCode {
     US
 }
 
+#[derive(Clone, Copy, Debug, Serialize, Deserialize)]
+pub enum AbsentImpl {
+    ManualCapture,
+}
```
NishantJoshi00, line 47: can we have it in snake case? it stands out quite a bit in the configuration file

crates/router/src/core/payment_methods/cards.rs
```diff
@@ -17,7 +17,7 @@ use common_utils::{
 };
 use error_stack::{report, IntoReport, ResultExt};
 use router_env::{instrument, tracing};
-use storage_models::payment_method;
+use storage_models::{enums::CaptureMethod, payment_method};
```
NishantJoshi00, line 20: avoid specific imports, instead import `enums` and in the later part of the code use `enums::CaptureMethod` instead

crates/router/src/core/payment_methods/cards.rs
```diff
@@ -1184,6 +1193,20 @@ fn filter_pm_based_on_config<'a>(
...
+    payment_method_filters: &settings::PaymentMethodFilters,
+    payment_method_type: &api_enums::PaymentMethodType,
+) -> bool {
+    payment_method_filters
+        .0
+        .get(&settings::PaymentMethodFilterKey::PaymentMethodType(
+            *payment_method_type,
+        ))
+        .and_then(|v| v.absent_impl)
+        .is_none()
```
NishantJoshi00, line 1207: It seems you are not using the value that, was specified for `absent_impl`, even though currently we don't require it, please add a match case, so in the future if any new filters were to be added, the developer would understand the logic that you have written and also the modifications would be much easier

NishantJoshi00 (changes requested):
Other than that, looks good!

Narayanbhat166 (approved):

SanchithHegde (dismissed):

NishantJoshi00 (dismissed):

crates/router/src/configs/settings.rs
```diff
@@ -128,7 +129,12 @@ pub struct CurrencyCountryFilter {
     pub currency: Option<HashSet<api_models::enums::Currency>>,
     #[serde(deserialize_with = "string_set_deser")]
     pub country: Option<HashSet<api_models::enums::CountryCode>>,
-    pub absent_impl: Option<api_models::enums::AbsentImpl>,
+    pub not_available_flows: Option<NotAvailableFlows>,
```
Narayanbhat166, line 132: The parent struct will have to be renamed, it's not only currency and country filter now.

NishantJoshi00 (dismissed):

Narayanbhat166 (approved):

jarnura (approved):

jarnura (dismissed):

Narayanbhat166 (dismissed):

jarnura (approved):

dracarys18 (approved):
LGTM

PR #1020, opened 2023-04-28
refactor(config): add independent toggles for enabling traces and metrics

SanchithHegde:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [x] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

This PR adds two configuration variables for toggling traces and metrics independently instead of using a single toggle to enable both. In addition, this PR introduces one more configuration variable to ignore errors during traces and metrics setup, if required.

### Additional Changes

The `log.telemetry.enabled` configuration variable is no longer being used, the `log.telemetry.traces_enabled` and `log.telemetry.metrics_enabled` configuration variables must be used instead for toggling traces and metrics, respectively. In addition, errors during traces or metrics setup can be ignored using the `log.telemetry.ignore_errors` configuration variable.

## Motivation and Context

This would allow us to have metrics and traces independently enabled, as required.

## How did you test it?

Manually.

## Checklist

### DIFF
--- a/crates/router_env/src/logger/setup.rs
+++ b/crates/router_env/src/logger/setup.rs
@@ -10,7 +10,6 @@ use opentelemetry::{
         propagation::TraceContextPropagator,
         trace, Resource,
     },
-    trace::TraceError,
     KeyValue,
 };
 use opentelemetry_otlp::{TonicExporterBuilder, WithExportConfig};
@@ -33,22 +32,19 @@ pub fn setup(
     config: &config::Log,
     service_name: &'static str,
     crates_to_filter: impl AsRef<[&'static str]>,
-) -> Result<TelemetryGuard, opentelemetry::metrics::MetricsError> {
+) -> TelemetryGuard {
     let mut guards = Vec::new();
 
     // Setup OpenTelemetry traces and metrics
-    let (telemetry_tracer, _metrics_controller) = if config.telemetry.enabled {
-        global::set_text_map_propagator(TraceContextPropagator::new());
-        (
-            setup_tracing_pipeline(&config.telemetry, service_name),
-            setup_metrics_pipeline(&config.telemetry),
-        )
+    let traces_layer = if config.telemetry.traces_enabled {
+        setup_tracing_pipeline(&config.telemetry, service_name)
     } else {
-        (None, None)
+        None
     };
-    let telemetry_layer = match telemetry_tracer {
-        Some(Ok(ref tracer)) => Some(tracing_opentelemetry::layer().with_tracer(tracer.clone())),
-        _ => None,
+    let _metrics_controller = if config.telemetry.metrics_enabled {
+        setup_metrics_pipeline(&config.telemetry)
+    } else {
+        None
     };
 
     // Setup file logging
@@ -75,7 +71,7 @@ pub fn setup(
     };
 
     let subscriber = tracing_subscriber::registry()
-        .with(telemetry_layer)
+        .with(traces_layer)
         .with(StorageSubscription)
         .with(file_writer);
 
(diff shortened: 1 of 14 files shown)


crates/router_env/src/logger/setup.rs
```diff
@@ -33,22 +32,19 @@ pub fn setup(
...
-        (None, None)
+        None
     };
-    let telemetry_layer = match telemetry_tracer {
-        Some(Ok(ref tracer)) => Some(tracing_opentelemetry::layer().with_tracer(tracer.clone())),
-        _ => None,
+    let _metrics_controller = if config.telemetry.metrics_enabled {
+        setup_metrics_pipeline(&config.telemetry)
+    } else {
+        None
```
Abhicodes-crypto, line 47:
```suggestion
    let _metrics_controller = config
        .telemetry
        .metrics_enabled
        .then(|| setup_metrics_pipeline(&config.telemetry))
        .flatten();
```
optional change.

Abhicodes-crypto (dismissed):

crates/router_env/src/logger/setup.rs
SanchithHegde: I'd personally prefer keeping it the way it is since I find it to be more readable, and since the suggested change requires understanding what `bool::then()` does to understand what's happening here. I'm open to any more opinions on this, I'll leave this conversation open for a while.

crates/router_env/src/logger/setup.rs
```diff
@@ -148,14 +142,24 @@ fn setup_tracing_pipeline(
...
+    if config.ignore_errors {
+        traces_layer_result
+            .map_err(|error| {
+                eprintln!("Failed to create an `opentelemetry_otlp` tracer: {error:?}")
+            })
+            .ok()
+    } else {
+        // Safety: This is conditional, there is an option to avoid this behavior at runtime.
+        #[allow(clippy::expect_used)]
+        Some(traces_layer_result.expect("Failed to create an `opentelemetry_otlp` tracer"))
```
lsampras, line 161:
```suggestion
        traces_layer_result.expect("Failed to create an `opentelemetry_otlp` tracer")
```

crates/router_env/src/logger/setup.rs
```diff
@@ -180,9 +184,17 @@ fn setup_metrics_pipeline(config: &config::LogTelemetry) -> Option<BasicControll
...
+        .build();
+
+    if config.ignore_errors {
+        metrics_controller_result
+            .map_err(|error| eprintln!("Failed to setup metrics pipeline: {error:?}"))
+            .ok()
+    } else {
+        // Safety: This is conditional, there is an option to avoid this behavior at runtime.
+        #[allow(clippy::expect_used)]
+        Some(metrics_controller_result.expect("Failed to setup metrics pipeline"))
```
lsampras, line 196:
```suggestion
        traces_layer_result.expect("Failed to create an `opentelemetry_otlp` tracer")
```

crates/router_env/src/logger/setup.rs
Abhicodes-crypto: Sure. It's an optional change. I don't prefer it either but suggested because there were already similar functions(ex: `fp_utils::when`) used in the codebase. You don't have to make the change. Resolving conv.

lsampras (approved):

SanchithHegde:
> I wonder if we should update the config structure here as well...
> 
> currently we have telmetry
> 
> I think the below could be a better representation... tel-new

I'd prefer to take this up in a separate PR.

crates/router_env/src/logger/setup.rs
SanchithHegde: I'll need the `Some` since this allows us to either provide a trace layer when traces are enabled or `None` when traces are disabled. This works out since `Option<_>` also implements the required traits.

crates/router_env/src/logger/setup.rs
SanchithHegde: Same here

crates/router_env/src/logger/setup.rs
SanchithHegde: Marking this resolved.

Abhicodes-crypto (approved):

jarnura (approved):

PR #1026, opened 2023-04-30
fix: fix internal server errors on merchant connector account creation

hrithikesh026:
## Type of Change

- [x] Bugfix

## Description
Added created_at and modified_at timestamps for "MerchantConnectorAccountNew" struct to match with "MerchantConnectorAccount". 

This was causing the "Create Merchant Connector Account" operation to fail with the below error message. 

> Failed to issue a query: Unexpected null for non-null column

Screenshot 2023-04-30 at 7 24 01 PM

## Motivation and Context

This change will fix the status code 500 error while creating a new merchant connector account. This is due to the lack of newly added created_at and modified_at fields in MerchantConnectorAccountNew struct in `crates/storage_models/src/merchant_connector_account.rs`

## How did you test it?

Manual

## Checklist


.vscode/launch.json
Narayanbhat166, line 1: You can remove this file since this is developer specific

hrithikesh026: Copy that

crates/storage_models/src/merchant_connector_account.rs
```diff
@@ -54,6 +54,8 @@ pub struct MerchantConnectorAccountNew {
     pub business_country: storage_enums::CountryCode,
     pub business_label: String,
     pub business_sub_label: Option<String>,
+    pub created_at: Option<time::PrimitiveDateTime>,
+    pub modified_at: Option<time::PrimitiveDateTime>,
```
SanchithHegde, line 58: You can have these as required values for insertion, since they must always be provided, or the database would throw errors. The existing implementation would still allow the possibility of not passing any timestamps for the `created_at` and `modified_at` fields.

SanchithHegde (changes requested):

Also, please follow the [commit message guidelines](https://github.com/juspay/hyperswitch/blob/main/docs/CONTRIBUTING.md#commits) for PR titles as well, since it ultimately ends up in the commit message when the PR gets merged.

crates/storage_models/src/merchant_connector_account.rs
hrithikesh026: Since `MerchantConnectorAccountNew` implements `Default` trait through `#[derive()] ` macro, every data member of the `MerchantConnectorAccountNew` struct needs to implement `Default` Macro. But `time::PrimitiveDateTime `doesn't implement it. 

But `Option<time::PrimitiveDateTime>` would defaults to None.

We would need to manually implement Default trait and for MerchantConnectorAccountNew and provide implementation for the `Default::default()` function.
[https://stackoverflow.com/questions/68346169/is-there-a-short-way-to-implement-default-for-a-struct-that-has-a-field-that-doe](url)

hrithikesh026: PaymentIntentNew also has `Option<time::PrimitiveDateTime>` for `created_at` and `modified_at` .

SanchithHegde: From what I can find, we do not use the `Default` implementation for `MerchantConnectorAccountNew` anywhere in our codebase, and in the only place where a `MerchantConnectorAccountNew` is constructed (in `core/admin.rs`), all fields are specified explicitly. It should be safe to remove the `#[derive(Default)]` annotation on `MerchantConnectorAccountNew`.

SanchithHegde (dismissed):

Narayanbhat166 (approved):

crates/router/src/core/admin.rs
```diff
@@ -414,6 +414,8 @@ pub async fn create_payment_connector(
         business_country,
         business_label,
         business_sub_label: req.business_sub_label,
+        created_at: common_utils::date_time::now(),
+        modified_at: common_utils::date_time::now(),
```
Narayanbhat166, line 418: When the connector is being modified, the modified time also should be updated. Is this being done already?

hrithikesh026: Yes.
This is how `MerchantConnectorAccountUpdateInternal` is defined. 
modified_at is still `Option<time::PrimitiveDateTime>` here. Should I change it to `time::PrimitiveDateTime`?
```
pub struct MerchantConnectorAccountUpdateInternal {
    merchant_id: Option<String>,
    connector_type: Option<storage_enums::ConnectorType>,
    connector_account_details: Option<Secret<serde_json::Value>>,
    test_mode: Option<bool>,
    disabled: Option<bool>,
    merchant_connector_id: Option<String>,
    payment_methods_enabled: Option<Vec<serde_json::Value>>,
    frm_configs: Option<Secret<serde_json::Value>>,
    modified_at: Option<time::PrimitiveDateTime>,
}
impl From<MerchantConnectorAccountUpdate> for MerchantConnectorAccountUpdateInternal {
    fn from(merchant_connector_account_update: MerchantConnectorAccountUpdate) -> Self {
        match merchant_connector_account_update {
            MerchantConnectorAccountUpdate::Update {
                merchant_id,
                connector_type,
                connector_account_details,
                test_mode,
                disabled,
                merchant_connector_id,
                payment_methods_enabled,
                metadata,
                frm_configs,
            } => Self {
                merchant_id,
                connector_type,
                connector_account_details,
                test_mode,
                disabled,
                merchant_connector_id,
                payment_methods_enabled,
                metadata,
                frm_configs,
                modified_at: Some(common_utils::date_time::now()),
            },
        }
    }
}
```

Narayanbhat166: Yes, the modified field should be updated when an update is made. So this field should not be optional. But if somewhere the `default()` function is being used, then we need to remove the default if possible or implement default on `datetime` to provide the current time.

jarnura (dismissed):

SanchithHegde (approved):

jarnura (approved):

PR #1029, opened 2023-05-02
feat:Payments cancellation can now be authenticated using publishable key

prajjwalkumar17:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description
We weren't able to use publishable key as authentication for the payments cancel. So the publishable key auth is required for the authentication type of cancel payment. We will require the publishable key auth and moreover we will require to add the client_secret in the body of the cancel payment.

### Additional Changes

## Motivation and Context
We are implementing the same authentication as was present in the confirm payments

## How did you test it?
tested it manually using PM.
Screenshot 2023-05-02 at 17 33 50

## Checklist


Abhicodes-crypto:
If publishable key auth is enabled for cancel API. Then customer can cancel the payment after `Authorizing`. I don't think this is a desirable feature and only the merchant should be able to cancel the payment. As far as I know payment gateways like stripe also don't support cancelling payments using publishable keys.

prajjwalkumar17:
@Abhicodes-crypto I have added you to the thread in slack where this feature is being discussed

Abhicodes-crypto:
> Moreover we aren't sending the request to connector

@prajjwalkumar17 , afaik if the payment is in `requires_capture` state then request is sent to connector to cancel the payment. But for other status connector is not called.

prajjwalkumar17:
@Abhicodes-crypto Yeah so for that status we already have the cancel payments implemented and for the other statuses we have mentioned we are creating our own check methods that what I have done.

Narayanbhat166:
> If publishable key auth is enabled for cancel API, then customer can cancel the payment after `Authorizing`. I don't think this is a desirable feature and only the merchant should be able to cancel the payment. As far as I know payment gateways like stripe also don't support cancelling payments using publishable keys.

cc @bernard-eugine. @Abhicodes-crypto is right, should we allow this?

jarnura (approved):

jarnura (commented):
Currently Payments cancellation only available for auth and capture payments.

For non auth and capture payments, via SDK we don't want to allow the cancellation.

There are auto expiry logic's to handle the same.

For payment links and user cancellation features we might need this, but while implementing the feature we need to revamp the solution.

PR #1051, opened 2023-05-04
fix: impl `Drop` for `RedisConnectionPool`

### LINKED ISSUE #1035: [BUG] panic on RedisPoolConnection close.
### Bug Description

Currently the RedisPoolConnection close panics on [this line](https://github.com/juspay/hyperswitch/blob/ed99655ebc11d53f4b2ffcb8c0eb9ef6b56f32c4/crates/router/src/bin/router.rs#L47). The root cause of this issue is `Arc::get_mut` throws None when `Arc` has 2 or more strong immutable reference. Since we are creating async tasks with reference to `Arc<RedisConnectionPool>` [here](https://github.com/juspay/hyperswitch/blob/ed99655ebc11d53f4b2ffcb8c0eb9ef6b56f32c4/crates/router/src/services.rs#L110).

### Expected Behavior

Runs successfully and closes the connection to redis and db connection.

### Actual Behavior

```text
thread 'main' panicked at 'Redis connection pool cannot be closed', crates/router/src/db.rs:74:14
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace.
```

### Steps To Reproduce

1. Run `cargo r`
2.  Stop it with `CTRL+C` 
3. You will get:

   ```text
   thread 'main' panicked at 'Redis connection pool cannot be closed', crates/router/src/db.rs:74:14
   note: run with RUST_BACKTRACE=1 environment variable to display a backtrace
   ```

### Possible Resolution

Possible resolution would be removing `&mut self` from close function. Since `redis.quit_pool` doesn't really need mutable reference. Other one would be something like.
```rust
impl Drop for RedisConnectionPool {
	fn drop(self) {
		tokio_rt.block(async {
			// call to the function that closes the connections
		})
	}
}
```

### Environment

1. Operating System or Linux distribution: MacOS
2. Rust version (output of `rustc --version`): ``1.69
3. App version (output of `cargo r -- --version`): ``0.5.9

None

dracarys18:
## Type of Change

- [x] Bugfix

## Description

This fixes #1035 

## Motivation and Context

#1035 

## How did you test it?

Manual

## Checklist

### DIFF
--- a/crates/router/src/bin/router.rs
+++ b/crates/router/src/bin/router.rs
@@ -39,7 +39,7 @@ async fn main() -> ApplicationResult<()> {
     logger::info!("Application started [{:?}] [{:?}]", conf.server, conf.log);
 
     #[allow(clippy::expect_used)]
-    let (server, _) = router::start_server(conf)
+    let server = router::start_server(conf)
         .await
         .expect("Failed to create the server");
     let _ = server.await;
--- a/crates/router/src/bin/router.rs
+++ b/crates/router/src/bin/router.rs
@@ -39,12 +39,10 @@ async fn main() -> ApplicationResult<()> {
     logger::info!("Application started [{:?}] [{:?}]", conf.server, conf.log);
 
     #[allow(clippy::expect_used)]
-    let (server, mut state) = router::start_server(conf)
+    let (server, _) = router::start_server(conf)
         .await
         .expect("Failed to create the server");
     let _ = server.await;
-
-    state.store.close().await;
 
     Err(ApplicationError::from(std::io::Error::new(
         std::io::ErrorKind::Other,
--- a/crates/drainer/src/main.rs
+++ b/crates/drainer/src/main.rs
@@ -33,6 +33,5 @@ async fn main() -> DrainerResult<()> {
     )
     .await?;
 
-    store.close().await;
     Ok(())
 }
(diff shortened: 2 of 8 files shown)


jarnura (dismissed):

crates/router/src/bin/router.rs
```diff
@@ -39,13 +39,11 @@ async fn main() -> ApplicationResult<()> {
     logger::info!("Application started [{:?}] [{:?}]", conf.server, conf.log);
 
     #[allow(clippy::expect_used)]
-    let (server, mut state) = router::start_server(conf)
+    let (server, _) = router::start_server(conf)
```
NishantJoshi00, line 42: If we don't need the state here, then can't we change the function signature of `start_server`

NishantJoshi00 (approved):
Left an optional comment, not an immediate priority. Everything looks good!

crates/router/src/bin/router.rs
Abhicodes-crypto, line 42: Sorry for interrupting but I have a small question, 

In `start_server` function the returned value is a clone of the `state`. 

So here if a `state` clone is dropped, and that triggers the redis_connection pool drop function. 

Does that mean anywhere in the code base if a `state` clone is dropped then redis connection is dropped(closed) ?

dracarys18: Since state has `Arc<RedisConnectionPool>` dropping would decrement the reference counter

Abhicodes-crypto: Got it. Thanks.

Abhicodes-crypto (approved):

SanchithHegde (approved):

jarnura (approved):

PR #1052, opened 2023-05-04
refactor: use `strum::EnumString` implementation for connector name conversions

### LINKED ISSUE #1042: [FEATURE] replace manual implementation using `from_str` function of strum
### Feature Description

In the `convert_connector` function the manual implementation of converting the connector from string can be replaced using the `from_str` function provided by `strum`

### Possible Implementation

Use `from_str` function of strum

nachiketkanore:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [x] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

- fixes #1042 
- added test cases for type conversions

### Additional Changes

## How did you test it?

- added 3 unit tests supporting the change

## Checklist


SanchithHegde:
@nachiketkanore Could you comment on #1042 that you've taken it up? Please ensure that you do so for any further issues that you take up.

jarnura (approved):

crates/router/src/types/api.rs
```diff
@@ -162,6 +165,42 @@ pub enum ConnectorCallType {
     Single(ConnectorData),
 }
 
+#[derive(EnumString, Serialize, Deserialize, PartialEq, Debug)]
+#[strum(serialize_all = "lowercase")]
+enum ConnectorName {
```
SanchithHegde, line 170: @vspecky @Narayanbhat166 Correct me if I'm wrong, I believe we'd prefer to use the [`RoutableConnectors`](https://github.com/juspay/hyperswitch/blob/3d05e50abcb92fe7e6c4472faafc03fb70920048/crates/api_models/src/enums.rs#L655) enum instead of adding a new one with the same enum variants.

NishantJoshi00 (approved):

crates/router/src/types/api.rs
vspecky: We would prefer to use the `[Connector](https://github.com/juspay/hyperswitch/blob/3d05e50abcb92fe7e6c4472faafc03fb70920048/crates/api_models/src/enums.rs#LL582C21-L582C21)` enum. Just to re-iterate, `Connector` is more general whereas `RoutableConnectors` is intended for routing purposes and is a direct subset of `Connector`.

SanchithHegde: @nachiketkanore Could you please make the change to use the existing `Connector` enum instead?

nachiketkanore: The ApplePay variant is not implemented inside [connector](https://github.com/juspay/hyperswitch/blob/e5cc0d9d45d41c391720ceb3f6c18151ac5a00f2/crates/router/src/connector.rs)
How do you want to handle [ApplePay](https://github.com/juspay/hyperswitch/blob/3d05e50abcb92fe7e6c4472faafc03fb70920048/crates/api_models/src/enums.rs#L586) and [Dummy](https://github.com/juspay/hyperswitch/blob/3d05e50abcb92fe7e6c4472faafc03fb70920048/crates/api_models/src/enums.rs#L594) for this?

vspecky: @nachiketkanore You can just remove the Applepay variant. It is residual (@Sangamesh26). Other than that, I'm not completely sure if we need both `Dummy` and `DummyConnector` variants. Is there any reason we should have both? @Narayanbhat166 @SanchithHegde

SanchithHegde: We don't need both. `Dummy` was added initially as a default value. Now that the dummy connector is being implemented under the `DummyConnector` variant, it should be fine to remove the `Dummy` variant, and make the `DummyConnector` variant as the default.

nachiketkanore: DummyConnector cannot be made default since it is behind the feature #[cfg(feature = "dummy_connector")]

Narayanbhat166: Can you remove the default trait? I don't see it being used anywhere

crates/router/src/types/api.rs
```diff
@@ -172,41 +169,7 @@ pub enum ConnectorCallType {
...
-    Stripe,
-    Worldline,
-    Worldpay,
-    Multisafepay,
-    Nexinets,
-    Paypal,
-    Trustpay,
-    Zen,
-}
+use api_models::enums::Connector as ConnectorName;
```
SanchithHegde, line 172: Can we avoid using this rename?

nachiketkanore: There is a trait [link](https://github.com/juspay/hyperswitch/blob/c158e043fb4be0ad389ddc6baa3d7a82bb3065d9/crates/router/src/types/api.rs#LL102C21-L102C21) with the same name

nachiketkanore: used `enums::Connector` because there's a trait named `Connector` already present in that file

SanchithHegde (dismissed):

SanchithHegde (approved):

jarnura (dismissed):

SanchithHegde (dismissed):

@jagan-jaya The `add_connector` script will have to be updated to accommodate this change. I feel it's best we have it as a separate PR.

SanchithHegde (approved):

jarnura (approved):

PR #1056, opened 2023-05-05
ci(migrations): add diesel migrations validations

lsampras:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [x] CI/CD

## Description

- Automatically check for migrations on database changes

### Additional Changes

## Checklist


.github/workflows/migration-check.yaml
```diff
@@ -0,0 +1,67 @@
+name: Migration Consistency Tests
+
+on:
+  pull_request:
+    types:
+      - "labeled"
+      - "opened"
+      - "ready_for_review"
+      - "synchronize"
```
SanchithHegde, line 9: We should be able to remove the filtering on event types.

.github/workflows/migration-check.yaml
```diff
@@ -0,0 +1,67 @@
...
+      - name: Install Rust
+        uses: dtolnay/rust-toolchain@master
+        with:
+          toolchain: stable
+
+      - uses: Swatinem/rust-cache@v2.2.0
+
+      - name: Install diesel_cli
+        shell: bash
+        run: cargo install diesel_cli
```
SanchithHegde, line 55:
```suggestion
        run: cargo install diesel_cli --no-default-features --features "postgres"
```

SanchithHegde (commented):

SanchithHegde (approved):

SanchithHegde:
@jarnura Please include this in the required checks for pull requests made against the `main` branch and the merge queue checks for the `main` branch.

.github/workflows/migration-check.yaml
```diff
@@ -0,0 +1,66 @@
...
+          --health-timeout 5s
+          --health-retries 5
+        ports:
+          - 5432:5432
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v3.1.0
+
+      - name: Install Rust
```
ashokkjag, line 45: Do we need this. I believe the images already come with rust.
https://github.com/actions/runner-images/blob/main/images/linux/Ubuntu2204-Readme.md#rust-tools

.github/workflows/migration-check.yaml
```diff
@@ -0,0 +1,66 @@
...
+        shell: bash
+        env:
+          DATABASE_URL: postgres://postgres:postgres@localhost:5432/postgres
+        run: diesel migration run --locked-schema
+
+      - name: Verify `diesel migration redo`
+        shell: bash
+        env:
+          DATABASE_URL: postgres://postgres:postgres@localhost:5432/postgres
+        run: diesel migration redo -a --locked-schema
```
ashokkjag, line 66: Should we verify undo also? I see sometimes people leave an empty file which prevents undo.

SanchithHegde: I've fixed some of the revert migrations with a `SELECT 1` query so that the revert runs successfully.

.github/workflows/migration-check.yaml
SanchithHegde: I'd still prefer to have the latest stable toolchain of Rust pulled from their distribution, at least for CI checks. It might not matter much for migration checks though. I'm fine with either keeping it or removing it, let me know what you think.

jarnura (approved):

PR #1086, opened 2023-05-09
feat: add accounts in-memory cache

dracarys18:
## Type of Change

- [x] New feature

## Description

Added accounts in memory cache static.  This also seperates publisher key through `CacheKind` enum so the publisher publishes the key into the channel in the form of `{cachekind}_rediskey`, to differentiate between accounts and config cache.

## Motivation and Context

Added in memory for merchant account table and merchant connector table for better latency

## How did you test it?

Manual

## Checklist

### DIFF
--- a/crates/router/src/cache.rs
+++ b/crates/router/src/cache.rs
@@ -8,6 +8,12 @@ use redis_interface::RedisValue;
 
 use crate::core::errors;
 
+/// Prefix for config cache key
+const CONFIG_CACHE_PREFIX: &str = "config";
+
+/// Prefix for accounts cache key
+const ACCOUNTS_CACHE_PREFIX: &str = "accounts";
+
 /// Time to live 30 mins
 const CACHE_TTL: u64 = 30 * 60;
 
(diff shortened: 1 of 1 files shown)


crates/router/src/cache.rs
```diff
@@ -1,23 +1,69 @@
...
+        };
+        let kind = kind.as_string().ok_or(validation_err.clone())?;
+        let mut split = kind.split(',');
+        match split.next().ok_or(validation_err.clone())? {
+            "accounts" => Ok(Self::Accounts(Cow::Owned(
+                split.next().ok_or(validation_err)?.to_string(),
+            ))),
+            "config" => Ok(Self::Config(Cow::Owned(
+                split.next().ok_or(validation_err)?.to_string(),
+            ))),
```
SanchithHegde, line 61: Would be preferable if these keys were `const`s instead of static string literals repeated multiple times.

crates/redis_interface/src/types.rs
```diff
@@ -5,9 +5,36 @@
...
+    }
+}
+
+impl RedisValue {
+    pub fn new(value: FredRedisValue) -> Self {
+        Self { inner: value }
+    }
+    pub fn into_inner(self) -> FredRedisValue {
+        self.inner
+    }
```
NishantJoshi00, line 30: Why are we exposing values of `fred type`, don't we have conversions and transformations written for this. Shouldn't those be exposed from here?

NishantJoshi00 (commented):

crates/redis_interface/src/types.rs
dracarys18: It's needed here https://github.com/juspay/hyperswitch/pull/1086/files#diff-0867d0a0cd2736dd09355b0280de09e00aa7114048b1b45a860318463f8ee48aR59

NishantJoshi00 (approved):
LGTM

PR #1097, opened 2023-05-09
fix(connector_customer): create connector_customer on requirement basis

Narayanbhat166:
## Type of Change

- [x] Bugfix

## Description

There was a function called `should_call_connector_create_customer` which was called after constructing the router data. This PR includes the change to move it outside and break if creating the customer is not required.

### Additional Changes

Remove the space and change this value to
```toml
[connector_customer]
connector_list = "bluesnap,stripe"
```

## Motivation and Context

BUGFIX

## How did you test it?

Manual, create a payment and customer is also created. Subsequent calls are not made to create customer again
Screenshot 2023-05-09 at 8 47 51 PM

Tested redirect payments too as it was failing previously.
Screenshot 2023-05-09 at 9 22 37 PM

Redirect success url https://www.google.com/?status=requires_capture&payment_intent_client_secret=pay_Rs7BYdDa0XODut0zQwpx&amount=10000

## Checklist

### DIFF
--- a/config/development.toml
+++ b/config/development.toml
@@ -229,7 +229,7 @@ stripe = { long_lived_token = false, payment_method = "wallet", payment_method_t
 checkout = { long_lived_token = false, payment_method = "wallet"}
 
 [connector_customer]
-connector_list = "bluesnap, stripe"
+connector_list = "bluesnap,stripe"
 
 [dummy_connector]
 payment_ttl = 172800
--- a/crates/router/src/core/payments/customers.rs
+++ b/crates/router/src/core/payments/customers.rs
@@ -15,68 +15,62 @@ use crate::{
 pub async fn create_connector_customer<F: Clone, T: Clone>(
     state: &AppState,
     connector: &api::ConnectorData,
-    customer: &Option<storage::Customer>,
     router_data: &types::RouterData<F, T, types::PaymentsResponseData>,
     customer_request_data: types::ConnectorCustomerData,
+    connector_customer_map: Option<serde_json::Map<String, serde_json::Value>>,
 ) -> RouterResult<(Option<String>, Option<storage::CustomerUpdate>)> {
-    let (is_eligible, connector_customer_id, connector_customer_map) =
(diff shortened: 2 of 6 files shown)


crates/router/src/core/payments/customers.rs
```diff
@@ -15,68 +15,62 @@ use crate::{
...
+    let connector_customer_id = match resp.response {
+        Ok(response) => match response {
+            types::PaymentsResponseData::ConnectorCustomerResponse {
+                connector_customer_id,
+            } => Some(connector_customer_id),
+            _ => None,
+        },
+        Err(err) => {
+            logger::debug!(payment_method_tokenization_error=?err);
+            None
```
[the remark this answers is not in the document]
jarnura: Yes, we should fail the payment if the customer create is failed. There are some where customer id is mandatory, in those flow connector flow will throw the required error.

jarnura (approved):

config/development.toml
```diff
@@ -229,7 +229,7 @@ stripe = { long_lived_token = false, payment_method = "wallet", payment_method_t
 checkout = { long_lived_token = false, payment_method = "wallet"}
 
 [connector_customer]
-connector_list = "bluesnap, stripe"
+connector_list = "bluesnap,stripe"
```
SanchithHegde, line 232: It's quite easy to make the same mistake of including spaces again. Can we have this configuration variable as a `Vec<String>` instead?

If you do make that change, ensure you also include it in the list of keys to be parsed as a list:

https://github.com/juspay/hyperswitch/blob/bc4ac529aa981150de6882d425bd274bc6272e30/crates/router/src/configs/settings.rs#L512-L513

Narayanbhat166: Yes, right. There are other values too which we are reading as string and then `deserializing` it to a `Vec<String>`.  We have to find a better way to do this. Linking #1109 as this refactor can be taken up in this issue

jagan-jaya (approved):

vspecky (approved):

SanchithHegde (approved):

PR #1100, opened 2023-05-09
feat(router): add route to invalidate cache entry

### LINKED ISSUE #995: [FEATURE] add a route that will invalidate cache
### Feature Description

Currently we don't have any way to manually invalidate cache that's stored in redis and in-memory, Add a route that given a key will invalidate both of redis and inmemory cache.

### Possible Implementation

So the in memory cache currently has `invalidate` function that will delete the entry in the cache, and in our redis interface we have `delete_key` function that will delete the entry from redis. You need to create a route in `app.rs` that will call this on `AdminApiKeyAuth`. 

None

jeevaprakashdr:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description
A new route to invalidate the in-memory cache and redis

### Additional Changes

## Motivation and Context
Currently we don't have any way to manually invalidate cache that's stored in redis and in-memory, 

Closes #995.

## How did you test it?
A unit test is added to test the in-memory cache and postman script used for redis cache.

## Checklist


crates/router/src/core/errors/api_error_response.rs
```diff
@@ -182,6 +182,8 @@ pub enum ApiErrorResponse {
     MissingFilePurpose,
     #[error(error_type = ErrorType::InvalidRequestError, code = "HE_04", message = "File content type not found / valid")]
     MissingFileContentType,
+    #[error(error_type = ErrorType::RouterError, code = "RE_01", message = "Failed to remove redis cache key")]
+    RemoveRedisKeyFailure,
```
SanchithHegde, line 186: @dracarys18 @Narayanbhat166 Please confirm if we would need a new error to be introduced for this scenario, or if we would prefer an `InternalServerError` being raised instead.

.gitignore
```diff
@@ -258,3 +258,5 @@ loadtest/*.tmp/
 result*
 
 .idea/
+
+.vscode/settings.json
```
SanchithHegde, line 262: Could you please revert this change? We have ignored most files under the `.vscode` directory. This specific file could be used for sharing project specific settings across developers working on the project, on a later date.

In fact, if you think such a thing would be helpful for you, I'll include project specific settings that can be followed across by developers working on the project, in a separate PR.

crates/router/src/core/cache.rs
```diff
@@ -0,0 +1,20 @@
+use common_utils::errors::CustomResult;
+
+use super::errors;
+use crate::{
+    cache::CONFIG_CACHE,
+    db::StorageInterface,
+    services::{self},
```
SanchithHegde, line 7:
```suggestion
    services,
```

crates/router/src/core/cache.rs
```diff
@@ -0,0 +1,20 @@
...
+
+pub async fn invalidate(
+    store: &dyn StorageInterface,
+    key: &str,
+) -> CustomResult<services::api::ApplicationResponse<serde_json::Value>, errors::ApiErrorResponse> {
+    CONFIG_CACHE.remove(key).await;
+
+    match store.get_redis_conn().delete_key(key).await {
+        Ok(_) => Ok(services::api::ApplicationResponse::StatusOk),
+        _ => Err(errors::ApiErrorResponse::RemoveRedisKeyFailure.into()),
```
SanchithHegde, line 18: Preferably, we should be returning a response with 404 status code in case a key was not found, and a response with 500 status code if a Redis related issue occurred.

crates/router/src/routes/cache.rs
```diff
@@ -0,0 +1,30 @@
...
+#[instrument(skip_all)]
+pub async fn invalidate(
+    state: web::Data<AppState>,
+    req: HttpRequest,
+    key: web::Path<String>,
+) -> impl Responder {
+    let flow = Flow::CacheInvalidate;
+
+    let key = key.into_inner().to_owned();
+    println!("cache key is {}", &key);
```
SanchithHegde, line 19: Please remove this `println!()` statement.

crates/router/tests/cache.rs
SanchithHegde, line 1: Once you've updated the code to return a response with a 400 status code for an invalid key, please add a test case for it.

SanchithHegde (commented):

crates/router/src/core/errors/api_error_response.rs
Narayanbhat166: we can use InternalServerError in its place. This function/route does only one thing, that is to remove the key from redis, there are not a lot of moving parts. This error tells that `RedisKeyRemovalFailed`, in order to know why it failed we anyway have to look into the logs and debug. So This error will not add much value over using `InternalServerError`.

crates/router/src/core/cache.rs
jeevaprakashdr: good spot!

crates/router/src/core/cache.rs
jeevaprakashdr: I am happy to return the InternalServerError(HTTP 500) as part of the error. 

In case of the key does not exist what needs to be done may be up for discussion. As of now RedisConnectionPool's `del(key)` encapsulation states  

> Removes the specified keys. A key is ignored if it does not exist. --> Based on this currently we are returning HTTP 200 status code. 

If we need to return the HTTP 404 status code,  need to extend `RedisConnectionPool::delete_key`.  This has currently been referenced in a couple of places.

crates/router/tests/cache.rs
jeevaprakashdr: can be done based on the outcome of 
[previous sugestion ](https://github.com/juspay/hyperswitch/pull/1100#discussion_r1190380458)

.gitignore
jeevaprakashdr: of course, I can. If i do come across this often, will remove a new issue.

crates/router/src/core/cache.rs
SanchithHegde: @dracarys18 @Narayanbhat166 What should be the behavior here?

Narayanbhat166: We can change this function https://github.com/juspay/hyperswitch/blob/cef8914372fa051f074e89fc76b76c6aee0d7bca/crates/redis_interface/src/commands.rs#L151 
to return a https://github.com/juspay/hyperswitch/blob/cef8914372fa051f074e89fc76b76c6aee0d7bca/crates/redis_interface/src/errors.rs#L52 error, and then the caller can match based on the error type?

dracarys18: @SanchithHegde @jeevaprakashdr As I can see in the fred documentation del will ignore if key does not exist in the database so it won't throw an error, So if del failed possibility is that something went wrong with redis so makes sense to throw InternalServerError.

Ref: https://docs.rs/fred/latest/fred/interfaces/trait.KeysInterface.html#method.del

SanchithHegde: @dracarys18 That seems to be how Redis handles it as well (official docs for the [`DEL` command](https://redis.io/commands/del/)). However, since we are only deleting one key, the output of the `DEL` command as an integer must be either 0 or 1. If we absolutely need to return a 404, we can match over the response (0 or 1), and return an appropriate result (`Ok(())` or `Err(NotFound)`). If we get any other response than 0 or 1, we'd have to error out stating an unexpected response from Redis.

Narayanbhat166: > @dracarys18 That seems to be how Redis handles it as well (official docs for the [`DEL` command](https://redis.io/commands/del/)). However, since we are only deleting one key, the output of the `DEL` command as an integer must be either 0 or 1. If we absolutely need to return a 404, we can match over the response (0 or 1), and return an appropriate result (`Ok(())` or `Err(NotFound)`). If we get any other response than 0 or 1, we'd have to error out stating an unexpected response from Redis.

This seems to be a good option.

Narayanbhat166: > @SanchithHegde @jeevaprakashdr As I can see in the fred documentation del will ignore if key does not exist in the database so it won't throw an error, So if del failed possibility is that something went wrong with redis so makes sense to throw InternalServerError.
> 
> Ref: https://docs.rs/fred/latest/fred/interfaces/trait.KeysInterface.html#method.del

Returning 200 will be ambiguous in this case, you are not sure whether the key was deleted or the key was given in the wrong format ( key not found ). It would be better if the end user is known that key was not found and hence the deletion did not happen

crates/redis_interface/src/commands.rs
```diff
@@ -147,13 +147,17 @@ impl super::RedisConnectionPool {
...
-        self.pool
+        match self.pool
             .del(key)
-            .await
-            .into_report()
-            .change_context(errors::RedisError::DeleteFailed)
+            .await{
+                Ok(0) | Ok(i32::MIN..=-1_i32) => Err(errors::RedisError::NotFound).into_report(),
+                Ok(1_i32..=i32::MAX) =>  Ok(()),
+                Err(fred::error::RedisError {..}) => return Err(errors::RedisError::DeleteFailed).into_report(),
```
SanchithHegde, line 159: We'd prefer matching against 0 and 1 only, while keeping the other values as unexpected responses. In the previous conversation, I forgot that we were doing similar response handling elsewhere in the `redis_interface` crate, in a slightly neater way. Could you define an enum `DelReply` like so, which implements the `fred::types::FromRedis` trait:

```rust
enum DelReply {
    KeyDeleted,
    KeyNotDeleted, // Key not found
}
```

You can refer to this snippet for implementing the `FromRedis` trait:

https://github.com/juspay/hyperswitch/blob/d7cfb4a179083580a7e195fa07077af23a262ceb/crates/redis_interface/src/types.rs#L135-L152

Then, you can match on the `DelReply` enum and return either an `Ok(())`, `Err(RedisError::NotFound)`, or `Err(RedisError::DeleteFailed)`.

crates/router/src/core/cache.rs
```diff
@@ -0,0 +1,16 @@
...
+pub async fn invalidate(
+    store: &dyn StorageInterface,
+    key: &str,
+) -> CustomResult<services::api::ApplicationResponse<serde_json::Value>, errors::ApiErrorResponse> {
+    CONFIG_CACHE.remove(key).await;
+
+    match store.get_redis_conn().delete_key(key).await {
+        Ok(_) => Ok(services::api::ApplicationResponse::StatusOk),
+        _ => Err(errors::ApiErrorResponse::InternalServerError.into()),
+    }
```
SanchithHegde, line 15: This would have to return a response with a 404 status code when the key is not deleted.

crates/router/src/cache.rs
```diff
@@ -63,6 +63,10 @@ impl Cache {
         let val = self.get(key)?;
         (*val).as_any().downcast_ref::<T>().cloned()
     }
+
+    pub async fn remove(&self, key: &str) {
```
dracarys18, line 67: Hey @jeevaprakashdr can you also replace it `invalidate` with `remove` in the other places where `invalidate` is being used

jeevaprakashdr: @dracarys18  ya can do. then in that case following will be the changes. is this okay?
*Url Endpoint will be renamed 
               **From**  http://localhost:8080/cache/invalidate/{key}
               **To** http://localhost:8080/cache/remove/{key}
*core/cache.rs method is renamed 
            **from** invalidate(key)
            **to** remove(key)

crates/redis_interface/src/commands.rs
```diff
@@ -21,7 +21,7 @@ use fred::{
         RedisKey, RedisMap, RedisValue, Scanner, SetOptions, XCap, XReadResponse,
     },
 };
-use futures::StreamExt;
+use futures::{StreamExt};
```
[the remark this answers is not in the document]
jeevaprakashdr: done

crates/redis_interface/src/commands.rs
jeevaprakashdr: done

crates/router/src/cache.rs
dracarys18: It's fine to keep it as invalidate @jeevaprakashdr

dracarys18:
Can you add invalidate for `ACCOUNTS_CACHE` as well since [this](https://github.com/juspay/hyperswitch/commit/da4d721424d329af618a63034aabe2d9248eb041) change.

jeevaprakashdr:
@Narayanbhat166 @jarnura @ashokkjag  The changes are ready for review. Please can you have a look at the PR, when you get a chance.

crates/router/src/core/cache.rs
```diff
@@ -0,0 +1,16 @@
+use common_utils::errors::CustomResult;
+
+use super::errors;
+use crate::{cache::CONFIG_CACHE, db::StorageInterface, services};
+
+pub async fn invalidate(
+    store: &dyn StorageInterface,
+    key: &str,
+) -> CustomResult<services::api::ApplicationResponse<serde_json::Value>, errors::ApiErrorResponse> {
+    CONFIG_CACHE.remove(key).await;
```
dracarys18, line 10: @jeevaprakashdr can you add `ACCOUNT_CACHE.remove(key).await`

jeevaprakashdr: ahh okie, now it is clear a new in-memory cache has been introduced as part of  [this change](https://github.com/juspay/hyperswitch/commit/da4d721424d329af618a63034aabe2d9248eb041) and it should be invalidated through the new endpoint. 

The changes are done and pushed.

crates/drainer/src/utils.rs
```diff
@@ -100,6 +100,7 @@ pub async fn make_stream_available(
         .delete_key(stream_name_flag)
         .await
         .map_err(DrainerError::from)
+        .map(|_r| ())
```
SanchithHegde, line 103: Please revert this change.

crates/redis_interface/src/commands.rs
```diff
@@ -148,12 +148,17 @@ impl super::RedisConnectionPool {
...
+        let result: Result<DelReply, fred::error::RedisError> = self.pool.del(key).await;
+        match result {
+            Ok(del_reply) => Ok(del_reply),
+            Err(e) => match e.kind() {
+                fred::error::RedisErrorKind::NotFound => {
+                    Err(errors::RedisError::NotFound).into_report()
+                }
+                _ => Err(errors::RedisError::DeleteFailed).into_report(),
+            },
+        }
```
SanchithHegde, line 161:
```suggestion
        self.pool.del(key).await.into_report().change_context(errors::RedisError::DeleteFailed)
```

You should be able to do something similar, since the `NotFound` error is never raised in the current implementation.

crates/redis_interface/src/commands.rs
```diff
@@ -665,29 +670,81 @@ impl super::RedisConnectionPool {
 #[cfg(test)]
 mod tests {
     #![allow(clippy::unwrap_used)]
+    #![allow(clippy::expect_used)]
```
SanchithHegde, line 673:
```suggestion
    #![allow(clippy::expect_used, clippy::unwrap_used)]
```

crates/router/src/core/cache.rs
```diff
@@ -0,0 +1,21 @@
...
+    ACCOUNTS_CACHE.remove(key).await;
+
+    match store.get_redis_conn().delete_key(key).await {
+        Ok(_del_reply) => Ok(services::api::ApplicationResponse::StatusOk),
+        _ => Err(errors::ApiErrorResponse::InvalidRequestUrl.into()),
```
SanchithHegde, line 19: Here, we can match on `del_reply` and return a 200, 404 or 500 status code, something like this:

```rust
Ok(DelReply::KeyDeleted) => Ok(services::api::ApplicationResponse::StatusOk),
Ok(DelReply::KeyNotDeleted) => Err(errors::ApiErrorResponse::InvalidRequestUrl.into()),
error => error.into_report().change_context(errors::ApiErrorResponse::InternalServerError).attach_printable("Failed to invalidate cache")
```

You'd have to run the formatter by the way.

crates/router/tests/cache.rs
```diff
@@ -0,0 +1,76 @@
...
+        ))
+        .insert_header(api_key)
+        .send()
+        .await
+        .unwrap();
+
+    // Assert
+    let response_body = response.body().await;
+    println!("invalidate Cache: {response:?} : {response_body:?}");
+    assert_eq!(response.status(), awc::http::StatusCode::OK);
```
SanchithHegde, line 75: Once you update the logic for the invalidation endpoint, this would then become `StatusCode::NotFound`.

SanchithHegde (commented):

crates/router/src/core/cache.rs
jeevaprakashdr: done

crates/router/tests/cache.rs
jeevaprakashdr: good spot

crates/drainer/src/utils.rs
SanchithHegde: @jeevaprakashdr This is still pending.

SanchithHegde (commented):

crates/drainer/src/utils.rs
jeevaprakashdr: I thought I have dropped a message,  on this.  WE would need the error part  else we get the following error 

image

SanchithHegde (dismissed):

crates/router/src/cache.rs
```diff
@@ -131,17 +135,25 @@ mod cache_tests {
...
     }
 
     #[tokio::test]
-    async fn eviction_on_size_test() {
-        let cache = Cache::new(2, 2, Some(0));
+    async fn invalidate_cache_for_key() {
+        let cache = Cache::new(1800, 1800, None);
         cache.push("key".to_string(), "val".to_string()).await;
+
+        cache.remove("key").await;
```
dracarys18, line 149: Can you add the assert it with for value present since this is a test.

dracarys18 (commented):
Other than LGTM

crates/router/src/cache.rs
jeevaprakashdr: the changes are pushed now.

dracarys18 (approved):

SanchithHegde (dismissed):

crates/router/src/core/cache.rs
```diff
@@ -0,0 +1,25 @@
...
+    CONFIG_CACHE.remove(key).await;
+    ACCOUNTS_CACHE.remove(key).await;
+
+    match store.get_redis_conn().delete_key(key).await {
+        Ok(DelReply::KeyDeleted) => Ok(services::api::ApplicationResponse::StatusOk),
+        Ok(DelReply::KeyNotDeleted) => Err(errors::ApiErrorResponse::InvalidRequestUrl.into()),
+        Err(error) => Err(error
+            .change_context(errors::ApiErrorResponse::InternalServerError)
+            .attach_printable("Failed to invalidate cache")),
+    }
```
dracarys18, line 24: My bad i missed it but can you use `publish_and_redact_cache` function to clear the cache.

dracarys18: Will address in a seperate PR

dracarys18 (approved):

SanchithHegde (approved):

PR #1104, opened 2023-05-10
feat(error): add feature-gated stacktrace to error received from API

NishantJoshi00:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description
This change provides a feature-gated functionality for exporting the `stacktrace` in the API response, for easier debugging and faster RCAs

### Additional Changes

This change only affects the outbound error response, adding a optional field `stacktrace` which is only provided when the specific feature is enabled.

## How did you test it?

via. postman

Screenshot 2023-05-10 at 1 33 25 PM

## Checklist

### DIFF
--- a/crates/router/Cargo.toml
+++ b/crates/router/Cargo.toml
@@ -26,6 +26,7 @@ vergen = ["router_env/vergen"]
 multiple_mca = ["api_models/multiple_mca"]
 dummy_connector = ["api_models/dummy_connector"]
 external_access_dc = ["dummy_connector"]
+detailed_errors = ["api_models/detailed_errors"]
 
 
 [dependencies]
@@ -45,7 +46,7 @@ crc32fast = "1.3.2"
 diesel = { version = "2.0.3", features = ["postgres"] }
 dyn-clone = "1.0.11"
 encoding_rs = "0.8.32"
-error-stack = "0.3.1"
+error-stack = { version = "0.3.1", features = [ "serde" ] }
 frunk = "0.4.1"
 frunk_core = "0.4.1"
 futures = "0.3.28"
--- a/crates/router/Cargo.toml
+++ b/crates/router/Cargo.toml
@@ -26,7 +26,7 @@ vergen = ["router_env/vergen"]
 multiple_mca = ["api_models/multiple_mca"]
 dummy_connector = ["api_models/dummy_connector"]
 external_access_dc = ["dummy_connector"]
-detailed_errors = ["api_models/detailed_errors"]
+detailed_errors = ["api_models/detailed_errors", "error-stack/serde"]
 
 
 [dependencies]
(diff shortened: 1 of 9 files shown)


crates/router/src/core/errors.rs
```diff
@@ -461,3 +461,45 @@ pub enum WebhooksFlowError {
     #[error("Dispute webhook status validation failed")]
     DisputeWebhookValidationFailed,
 }
+
+#[cfg(feature = "detailed_errors")]
```
SanchithHegde, line 465: Optional change: You could move these new types and impls to a module to simplify the feature gates.

SanchithHegde (commented):
One more request: Can we manage to un-escape the strings in the list of errors returned? It would be great if it can be done, it would improve readability by a great extent. Let me know if it involves too much effort and/or performance hits, I'll approve this PR.

NishantJoshi00:
> One more request: Can we manage to un-escape the strings in the list of errors returned? It would be great if it can be done, it would improve readability by a great extent. Let me know if it involves too much effort and/or performance hits, I'll approve this PR.

It is definitely a performance hit, but regardless there is one more concern, the context as shown in the example won't always be a `json`, in some cases it would just be a simple string. I should have mentioned a example for this scenario. Please refer the following:

Screenshot 2023-05-10 at 5 24 21 PM

SanchithHegde:
> It is definitely a performance hit, but regardless there is one more concern, the context as shown in the example won't always be a `json`, in some cases it would just be a simple string.

Even when it's a string, don't you think it would improve readability if it's un-escaped? As for the performance hit, we're expecting to enable this for debug and/or testing builds, so I believe it is acceptable?

crates/router/Cargo.toml
```diff
@@ -26,6 +26,7 @@ vergen = ["router_env/vergen"]
 multiple_mca = ["api_models/multiple_mca"]
 dummy_connector = ["api_models/dummy_connector"]
 external_access_dc = ["dummy_connector"]
+detailed_errors = ["api_models/detailed_errors", "error-stack/serde"]
```
Narayanbhat166, line 29: Can you add this feature in `integ`, also we would need this in sandbox as well right?

NishantJoshi00: As you can see there is not integ feature flag here, and it's not preferred to add it here

NishantJoshi00:
> > It is definitely a performance hit, but regardless there is one more concern, the context as shown in the example won't always be a `json`, in some cases it would just be a simple string.
> 
> Even when it's a string, don't you think it would improve readability if it's un-escaped? As for the performance hit, we're expecting to enable this for debug and/or testing builds, so I believe it is acceptable?

Hey, this side effect of having escaped characters isn't because of my implementation is handling this behaviour, it's how it is being stored. That's what causing this issue

jarnura (approved):

PR #1112, opened 2023-05-10
fix(Google Pay): allow custom fields in `GpayTokenParameters` for Google Pay via Stripe

jagan-jaya:
## Type of Change

- [X] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

Stripe requires 
``` 
"gateway": "stripe"
"stripe:version": "2018-10-31"
"stripe:publishableKey": "YOUR_PUBLIC_STRIPE_KEY"
  ```
dashboard started sending these fields from front-end but backend is not handling the same, this PR will fix the issue.
Previously it was only
```
  "gateway": "assist"
  "gatewayMerchantId": "YOUR_GATEWAY_MERCHANT_ID"
```
### Additional Changes

## Checklist

### DIFF
--- a/crates/api_models/src/payments.rs
+++ b/crates/api_models/src/payments.rs
@@ -1442,11 +1442,12 @@ pub struct GpayTokenParameters {
     /// The merchant ID registered in the connector associated
     #[serde(skip_serializing_if = "Option::is_none")]
     pub gateway_merchant_id: Option<String>,
-    #[serde(skip_serializing_if = "Option::is_none")]
-    #[serde(rename = "stripe:version")]
+    #[serde(skip_serializing_if = "Option::is_none", rename = "stripe:version")]
     pub stripe_version: Option<String>,
-    #[serde(skip_serializing_if = "Option::is_none")]
-    #[serde(rename = "stripe:publishableKey")]
+    #[serde(
+        skip_serializing_if = "Option::is_none",
+        rename = "stripe:publishableKey"
+    )]
     pub stripe_publishable_key: Option<String>,
 }
 
--- a/crates/api_models/src/payments.rs
+++ b/crates/api_models/src/payments.rs
@@ -1440,7 +1440,14 @@ pub struct GpayTokenParameters {
     /// The name of the connector
     pub gateway: String,
     /// The merchant ID registered in the connector associated
-    pub gateway_merchant_id: String,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub gateway_merchant_id: Option<String>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    #[serde(rename = "stripe:version")]
+    pub stripe_version: Option<String>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    #[serde(rename = "stripe:publishableKey")]
+    pub stripe_publishable_key: Option<String>,
 }
 
 #[derive(Debug, Clone, serde::Serialize, serde::Deserialize, ToSchema)]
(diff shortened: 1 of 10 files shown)

### CODE REVIEW

File: crates/router/src/connector/utils.rs [hunk 1]
```diff
@@ -319,6 +319,47 @@ impl RefundsRequestData for types::RefundsData {
...
+
+#[derive(Clone, Debug, serde::Serialize)]
+pub struct GpayTokenizationData {
+    #[serde(rename = "type")]
+    pub token_type: String,
+    pub token: String,
+}
+
+impl From<api_models::payments::GooglePayWalletData> for GooglePayWalletData {
+    fn from(data: api_models::payments::GooglePayWalletData) -> Self {
```
Narayanbhat166, line 347: Both structs have the same name. Some developer integrating Googlepay might be confused as to which type to use and why this conversion is necessary, can you please explain as to why we have to do it like this?

Narayanbhat166 (approved):
LGTM

crates/api_models/src/payments.rs
```diff
@@ -1440,7 +1440,14 @@ pub struct GpayTokenParameters {
...
     /// The merchant ID registered in the connector associated
-    pub gateway_merchant_id: String,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub gateway_merchant_id: Option<String>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    #[serde(rename = "stripe:version")]
+    pub stripe_version: Option<String>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    #[serde(rename = "stripe:publishableKey")]
+    pub stripe_publishable_key: Option<String>,
```
SanchithHegde, line 1450:
```suggestion
    #[serde(skip_serializing_if = "Option::is_none", rename = "stripe:version")]
    pub stripe_version: Option<String>,
    #[serde(skip_serializing_if = "Option::is_none", rename = "stripe:publishableKey")]
    pub stripe_publishable_key: Option<String>,
```

crates/router/src/connector/utils.rs
jagan-jaya: All connectors requires the googlepay_data in camelcase but we serialize and deserialize it in snake_case, so we initially made serialization in camelcase and deserialization in snake_case previously. Now an issue happened when they stored the gpay token in camelcase in vault and tried retrieving in snake_case, so to avoid this issue we revert back to old mechanism of all snake_case concept. now whatever connector that requires the google token strictly in camelcase we add a custom logic for those to convert them to camelcase. That's why both have same name but are in different crates

crates/api_models/src/payments.rs
```diff
@@ -704,7 +704,7 @@ pub enum WalletData {
 }
 
 #[derive(Eq, PartialEq, Clone, Debug, serde::Deserialize, serde::Serialize, ToSchema)]
-#[serde(rename_all(serialize = "camelCase", deserialize = "snake_case"))]
```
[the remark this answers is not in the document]
jagan-jaya: no, it's not , we accept all the input in snake_case, so there won't be any impact in api-contract

crates/router/src/connector/utils.rs
Narayanbhat166: Thanks for the explanation.

jagan-jaya:
Closing in favour of #1125

Issue #1091, opened 2023-05-09
[FEATURE] add support for filtering in `payments_session`


Narayanbhat166:
### Feature Description

After supporting multiple connector accounts through `business_label` , `business_country` and `business_sub_label`, PaymentsSession flow ( creating session tokens - wallets ) should filter based on this.

### Possible Implementation

Add a similar filter which is in `list_payment_methods`

### Have you spent some time to check if this feature request has been raised before?

- [X] I checked and didn't find similar issue

### Have you read the Contributing Guidelines?

- [X] I have read the [Contributing Guidelines](https://github.com/juspay/hyperswitch/blob/main/docs/CONTRIBUTING.md)

### Are you willing to submit a PR?

Yes I am willing to submit a PR!

PR #1128, opened 2023-05-11
refactor(session_token): add support for business filtering in payments session

Narayanbhat166:
## Type of Change

- [x] Refactoring

## Description

- This PR refactors the get connector function of `PaymentsSession` operation to use idiomatic rust. 
- Add business country and business label filtering.

## Motivation and Context

 #1091 

## How did you test it?

Tested manually. 
- Create payments
- Create session token
  - With `wallets` fields as empty list
  - With `wallets` field as `applepay` only

## Checklist


crates/router/src/core/payments/operations/payment_session.rs
```diff
@@ -299,84 +309,88 @@ where
...
+                    .unwrap_or_default()
+                    .into_iter()
+                    .map(|payment_methods_enabled| {
+                        payment_methods_enabled
+                            .parse_value::<PaymentMethodsEnabled>("payment_methods_enabled")
+                    })
+                    .filter_map(|parsed_payment_method_result| {
+                        let error = parsed_payment_method_result.as_ref().err();
+                        logger::error!(session_token_parsing_error=?error);
+                        parsed_payment_method_result.ok()
```
vspecky, line 334: Shouldn't we throw an error here?

crates/router/src/core/payments/operations/payment_session.rs
```diff
@@ -299,84 +309,88 @@ where
...
+                api::GetToken::from(payment_method_type),
+            ) {
+                Ok(connector_data) => session_connector_data.push(api::SessionConnectorData {
+                    payment_method_type,
+                    connector: connector_data,
+                    business_sub_label,
+                }),
+                Err(error) => {
+                    logger::error!(session_token_error=?error)
                 }
```
vspecky, line 387: Here too, shouldn't we throw an error?

vspecky (changes requested):

crates/router/src/core/payments/operations/payment_session.rs
Narayanbhat166: No, if any one connector is misconfigured ( with invalid data ), we shouldn't break the flow and throw the error. Other connectors where we can create the session token, we should create.

crates/router/src/core/payments/operations/payment_session.rs
Narayanbhat166: No we shouldn't, for the same reason as specified before. I'm thinking of collecting all the errors and then return them in a separate field in the response.

crates/router/src/core/payments/operations/payment_session.rs
```diff
@@ -299,84 +309,88 @@ where
...
+                    .payment_methods_enabled
+                    .unwrap_or_default()
+                    .into_iter()
+                    .map(|payment_methods_enabled| {
+                        payment_methods_enabled
+                            .parse_value::<PaymentMethodsEnabled>("payment_methods_enabled")
+                    })
+                    .filter_map(|parsed_payment_method_result| {
+                        let error = parsed_payment_method_result.as_ref().err();
+                        logger::error!(session_token_parsing_error=?error);
```
Abhicodes-crypto, line 333: So every time this code is run , this error log is printed right ? If the error is none it logs `None` and when the error is `Some` it logs the error. Can we have the log only when the error occurs.

jarnura (approved):

PR #1136, opened 2023-05-11
fix(session_token): log error only when it occurs

Abhicodes-crypto:
## Type of Change

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

```
let error = parsed_payment_method_result.as_ref().err();
logger::error!(session_token_parsing_error=?error);
parsed_payment_method_result.ok()
```

This logic logs the error always. Changed it to log error only when error occurs. 

### Additional Changes

## How did you test it?

Compiles. But please verify the logic while review.

## Checklist

### DIFF
--- a/crates/common_utils/src/ext_traits.rs
+++ b/crates/common_utils/src/ext_traits.rs
@@ -3,7 +3,6 @@
 //! & inbuilt datatypes.
 //!
 
-use std::fmt::Display;
 
 use error_stack::{IntoReport, ResultExt};
 use masking::{ExposeInterface, Secret, Strategy};
@@ -448,7 +447,7 @@ pub trait ResultExtLog<T> {
 
 impl<T, E> ResultExtLog<T> for Result<T, E>
 where
-    E: Display + std::fmt::Debug,
+    E: std::fmt::Display,
 {
     fn log_err_and_ok(self, key: &str) -> Option<T> {
         self.map_err(|err| {
--- a/crates/common_utils/src/ext_traits.rs
+++ b/crates/common_utils/src/ext_traits.rs
@@ -3,9 +3,12 @@
 //! & inbuilt datatypes.
 //!
 
+use std::fmt::Display;
+
 use error_stack::{IntoReport, ResultExt};
 use masking::{ExposeInterface, Secret, Strategy};
 use quick_xml::de;
+use router_env::logger;
 use serde::{Deserialize, Serialize};
 
 use crate::errors::{self, CustomResult};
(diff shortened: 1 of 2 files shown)


crates/router/src/core/payments/operations/payment_session.rs
```diff
@@ -329,9 +328,7 @@ where
                             .parse_value::<PaymentMethodsEnabled>("payment_methods_enabled")
                     })
                     .filter_map(|parsed_payment_method_result| {
-                        let error = parsed_payment_method_result.as_ref().err();
```
Narayanbhat166, line 332: A `map_error` would be sufficient in this case.

crates/common_utils/src/ext_traits.rs
```diff
@@ -434,3 +435,24 @@ impl XmlExt for &str {
...
+pub trait ResultExtLog<T> {
+    ///
+    /// Log error if it is of `Err` type else return the `Ok` value wrapped in `Some`
+    ///
+    fn log_err_and_ok(self, key: &str) -> Option<T>;
+}
+
+impl<T, E> ResultExtLog<T> for Result<T, E>
+where
+    E: std::fmt::Display,
```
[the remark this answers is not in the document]
Abhicodes-crypto: We could use , but since most of the error types have display impl and usually display impl is logged , so added this. For errors with no Display impl, direct `map_err` can be used.

Narayanbhat166 (approved):

crates/common_utils/src/ext_traits.rs
```diff
@@ -434,3 +435,24 @@ impl XmlExt for &str {
         de::from_str(self)
     }
 }
+
+/// Extension trait for logging the error of a Result and then discarding the error
+pub trait ResultExtLog<T> {
+    ///
+    /// Log error if it is of `Err` type else return the `Ok` value wrapped in `Some`
+    ///
+    fn log_err_and_ok(self, key: &str) -> Option<T>;
```
SanchithHegde, line 444:
```suggestion
    #[track_caller]
    fn log_err_and_ok(self, key: &str) -> Option<T>;
```

Please check if the log message contains the location of where this function has been called, when this annotation is included. If yes, please do include the annotation.

SanchithHegde (commented):

crates/common_utils/src/ext_traits.rs
Abhicodes-crypto: I don't think it works that way. But will verify and update here.

Abhicodes-crypto: Yep, there is no difference with and without it. 
with -
image
without -
image

Abhicodes-crypto: It was decided to remove this trait based on offline discussions with @vspecky

SanchithHegde: Alright, marking this resolved.

SanchithHegde (approved):

SanchithHegde (approved):

Narayanbhat166 (approved):

PR #1141, opened 2023-05-11
fix(merchant_connector_account): add validation for the `disabled` flag

NishantJoshi00:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

The `disabled` field in the merchant connector account, was being populated, but it's presence wasn't being validated during a payment. This PR resolves this issue.

### Additional Changes

## How did you test it?

Screenshot 2023-05-12 at 3 04 35 AM

## Checklist

### DIFF
--- a/crates/router/src/core/payments/helpers.rs
+++ b/crates/router/src/core/payments/helpers.rs
@@ -1570,6 +1570,13 @@ impl MerchantConnectorAccountType {
             Self::CacheVal(val) => val.connector_account_details.peek().to_owned(),
         }
     }
+
+    pub fn is_disabled(&self) -> bool {
+        match self {
+            Self::DbVal(ref inner) => inner.disabled.unwrap_or(false),
+            Self::CacheVal(_) => false,
+        }
+    }
 }
 
 pub async fn get_merchant_connector_account(
--- a/crates/router/src/core/payments/helpers.rs
+++ b/crates/router/src/core/payments/helpers.rs
@@ -1574,6 +1574,8 @@ impl MerchantConnectorAccountType {
     pub fn is_disabled(&self) -> bool {
         match self {
             Self::DbVal(ref inner) => inner.disabled.unwrap_or(false),
+            // Cached merchant connector account, only contains the account details,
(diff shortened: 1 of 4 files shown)


crates/router/src/core/payments/helpers.rs
```diff
@@ -1570,6 +1570,13 @@ impl MerchantConnectorAccountType {
             Self::CacheVal(val) => val.connector_account_details.peek().to_owned(),
         }
     }
+
+    pub fn is_disabled(&self) -> bool {
+        match self {
+            Self::DbVal(ref inner) => inner.disabled.unwrap_or(false),
+            Self::CacheVal(_) => false,
```
SanchithHegde, line 1577: Would be great if you could include comments as to why we are always returning `false` in case of `CacheVal`.

SanchithHegde (commented):

SanchithHegde (approved):

crates/router/src/core/payments/transformers.rs
```diff
@@ -52,6 +53,10 @@ where
     )
     .await?;
 
+    fp_utils::when(merchant_connector_account.is_disabled(), || {
+        Err(errors::ApiErrorResponse::MerchantConnectorAccountDisabled)
```
Narayanbhat166, line 57: Can you add the connector label here? It will be more clear as to which mca was disabled.

NishantJoshi00: I agree, but the type is a abstraction over db value and cache value. And getting to look inside, the cache value doesn't provide this label.

Narayanbhat166: You already have the connector label that you are querying with, here https://github.com/juspay/hyperswitch/pull/1141/files#diff-6f19e8956b6645147dedca51bdadb551b2cfbe7ce0a7092e514a4dfa41deb3bcR51 Use that

PR #1143, opened 2023-05-11
feat(router): add mandates list api

sai-harsha-vardhan:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description
Mandates - List => Lists mandates for a merchant based on some filters like mandate_status, connector etc.

### Additional Changes

## How did you test it?
Tested Manually
image

## Checklist


crates/router/src/types/storage/mandate.rs
```diff
@@ -1,3 +1,69 @@
...
+            filter = filter.limit(limit);
+        }
+
+        logger::debug!(query = %diesel::debug_query::<diesel::pg::Pg, _>(&filter).to_string());
+
+        filter
+            .get_results_async(conn)
+            .await
+            .into_report()
+            .change_context(errors::DatabaseError::NotFound)
```
SanchithHegde, line 66: This would end up raising an incorrect error for the scenario. The query built here returns an empty `Vec` when no records are found, and if any error does occur, it would be an internal database error, due to which we must be raising a `DatabaseError::Unknown` error.

Please update this to raise a `DatabaseError::Unknown` error, and feel free to include comments as to why this is being done, if you feel this needs to be documented.

crates/router/src/types/storage/mandate.rs
```diff
@@ -1,3 +1,69 @@
...
+            .attach_printable_lazy(|| "Error filtering records by predicate")
```
SanchithHegde, line 67: Since this does not involve any `format!()` calls, this can just be `attach_printable()`.

```suggestion
            .attach_printable("Error filtering mandates by specified constraints")
```

crates/router/src/core/mandate.rs
```diff
@@ -259,6 +260,25 @@ where
...
+pub async fn retrieve_mandates_list(
+    state: &AppState,
+    merchant_account: storage::MerchantAccount,
+    constraints: api_models::mandates::MandateListConstraints,
+) -> RouterResponse<Vec<api_models::mandates::MandateResponse>> {
+    let mandates = state
+        .store
+        .find_mandates_by_merchant_id(&merchant_account.merchant_id, constraints)
+        .await
+        .to_not_found_response(errors::ApiErrorResponse::InternalServerError)
```
SanchithHegde, line 273:
```suggestion
        .change_context(errors::ApiErrorResponse::InternalServerError)
```

crates/router/src/core/mandate.rs
```diff
@@ -259,6 +260,25 @@ where
...
+        .attach_printable("Unable to retrieve mandates")?;
+    let mandates_list = future::try_join_all(mandates.iter().map(|mandate| {
+        mandates::MandateResponse::from_db_mandate(state, mandate.clone(), &merchant_account)
+    }))
+    .await?;
```
SanchithHegde, line 278: You should be able to use `into_iter()` instead of `iter()` and can avoid the clone.

crates/router/src/core/mandate.rs
Narayanbhat166: Since we are searching unsing the merchant_id, shouldn't this error be `MerchantAccountNotFound`

SanchithHegde: > Since we are searching unsing the merchant_id, shouldn't this error be `MerchantAccountNotFound`

Nope, we are returning a list of mandates, which would be empty when there are no results. See https://github.com/juspay/hyperswitch/pull/1143#discussion_r1192261008.

SanchithHegde (approved):

PR #1147, opened 2023-05-12
feat(connector): add multiple dummy connectors and enable them

### LINKED ISSUE #726: [FEATURE] Create a dummy connector for internal and external testing of payments and refunds flows
### Feature Description

A dummy test connector needs to be created with the ability to process all the payment methods available on Hyperswitch. This dummy connector will be useful for internal and external testing and it will be used extensively by those trying to test or integrate Hyperswitch without setting up any active accounts with any of the processors. 

The dummy test connector will be able to simulate most basic scenarios (success/failure/pending) for all the available payment methods by having separate test payment method data for each of these flows. Payment failure can include further breakdown in scenarios according to the payment methods. 

For example, there should be separate test card numbers for simulating success, failure, pending statuses of 3DS and Non 3DS flows. Under failure, different test card numbers should enable simulating errors like invalid card info, insufficient funds, declined by gateway, declined by issuing bank, declines due to bad fraud score, etc. 

### Possible Implementation

We would need to create a dummy connector to test all the payment flows, refund flows, webhooks, etc.

The two major tasks are:
1. Server implementation for dummy connector for the following endpoints
a. payments #980
b. refunds #1071 
c. webhooks 

2. Integrating Hyperswitch with the dummy connector for the following endpoints
a. payments #1084 
b. refunds #1084 
c. webhooks 

3. Enable Dummy Connector and make it default

**Cards:**
Non- 3DS payments flow https://github.com/juspay/hyperswitch/issues/727

ThisIsMani:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

Add multiple dummy connectors and enable dummy connector by default

### Additional Changes

## Motivation and Context

Closes #726.

## Checklist

### DIFF
--- a/crates/router/src/connector/dummyconnector.rs
+++ b/crates/router/src/connector/dummyconnector.rs
@@ -21,22 +21,22 @@ use crate::{
 };
 
 #[derive(Debug, Clone)]
-pub struct DummyConnector<const T: i32>;
+pub struct DummyConnector<const T: u8>;
 
-impl<const T: i32> api::Payment for DummyConnector<T> {}
-impl<const T: i32> api::PaymentSession for DummyConnector<T> {}
-impl<const T: i32> api::ConnectorAccessToken for DummyConnector<T> {}
-impl<const T: i32> api::PreVerify for DummyConnector<T> {}
-impl<const T: i32> api::PaymentAuthorize for DummyConnector<T> {}
-impl<const T: i32> api::PaymentSync for DummyConnector<T> {}
-impl<const T: i32> api::PaymentCapture for DummyConnector<T> {}
-impl<const T: i32> api::PaymentVoid for DummyConnector<T> {}
-impl<const T: i32> api::Refund for DummyConnector<T> {}
-impl<const T: i32> api::RefundExecute for DummyConnector<T> {}
-impl<const T: i32> api::RefundSync for DummyConnector<T> {}
-impl<const T: i32> api::PaymentToken for DummyConnector<T> {}
+impl<const T: u8> api::Payment for DummyConnector<T> {}
+impl<const T: u8> api::PaymentSession for DummyConnector<T> {}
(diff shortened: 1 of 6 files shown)


crates/router/src/connector/dummyconnector.rs
```diff
@@ -21,32 +21,33 @@ use crate::{
...
-impl api::PaymentSync for DummyConnector {}
-impl api::PaymentCapture for DummyConnector {}
-impl api::PaymentVoid for DummyConnector {}
-impl api::Refund for DummyConnector {}
-impl api::RefundExecute for DummyConnector {}
-impl api::RefundSync for DummyConnector {}
-impl api::PaymentToken for DummyConnector {}
-
-impl
+pub struct DummyConnector<const T: i32>;
```
SanchithHegde, line 24: You could also have this as a `u8` instead, since I don't expect the number of dummy connectors to exceed 255.

SanchithHegde (commented):
Could you also update the PR description with the motivation behind this change?

NishantJoshi00 (approved):
LGTM

SanchithHegde (approved):
> Could you also update the PR description with the motivation behind this change?

I see that you've updated the PR description. To clarify, my question was rather about why we needed multiple dummy connectors. What problem does this solve?

jarnura (approved):

NishantJoshi00:
> > Could you also update the PR description with the motivation behind this change?
> 
> I see that you've updated the PR description. To clarify, my question was rather about why we needed multiple dummy connectors. What problem does this solve?

It's actually a requirement that was proposed by @koushikg14 and some other folks, making multiple connectors which behave as dummy connectors to test out orchestrations

PR #1179, opened 2023-05-17
revert(connector): fix stripe status to attempt status map

Narayanbhat166:
reverts #1132

This status is required in case of `sofort` bank_redirect payment method. Incorrect comments caused this to be removed.


Abhicodes-crypto (approved):
LGTM

PR #1183, opened 2023-05-17
feat(documentation): add polymorphic `generate_schema` macro

Narayanbhat166:
## Type of Change

- [x] New feature

## Description

This PR will allow us to specify which fields are mandatory in `PaymentsRequest` for a particular flow. A new macro is added called `PolymorphicSchema` ( Name is open for discussion ), which will generate different schemas as specified by 
`#[generate_schemas(PaymentsCreateRequest)]` , where we can specify which fields will be mandatory in PaymentsCreateRequest by adding the `mandatory_in` attribute as follows

```rust
#[mandatory_in(PaymentsCreateRequest)]
amount: Option<u64>
```
, this will add the `#[schema( required = true )` for that field when creating the schema.

For more information, you can refer to the macro documentation here.

## Motivation and Context

Make meaningful documentation which is easy to use.

## How did you test it?

Manually. Generate the openapi file and check if fields are marked as mandatory.

## Checklist


jarnura (dismissed):
LGTM

crates/router_derive/src/macros/generate_schema.rs
```diff
@@ -0,0 +1,199 @@
...
+    let fields = get_struct_fields(input.data)
+        .map_err(|error| syn::Error::new(proc_macro2::Span::call_site(), error))?;
+
+    // Go through all the fields and create a mapping of required fields for a schema
+    // PaymentsCreate -> ["amount","currency"]
+    // This will be stored in a hashset
+    // mandatory_hashset -> ((PaymentsCreate, amount), (PaymentsCreate,currency))
+
+    let mut mandatory_hashset = HashSet::<(syn::Ident, syn::Ident)>::new();
+    let mut other_fields_hm = HashMap::<syn::Field, Vec<syn::Attribute>>::new();
```
SanchithHegde, line 95: Do you think it would help to use a `HashMap<NewTypeName, Vec<RequiredField>>` instead?

Narayanbhat166: The lookup for this will be `Is this field mandatory in this NewType`, this can be done in a single lookup using `Hashset<(NewType, field)>`, but if we convert this to hashmap, then first we would have to lookup the NewType and then in the vector we should find if the field is present, so Hashset will be better in this case.

SanchithHegde (commented):

SanchithHegde (approved):

dracarys18 (approved):
LGTM

jarnura (approved):

PR #1198, opened 2023-05-17
refactor: extend PayoutMethodData with additional bank fields, add Mu…

kashif-m:
…ltiAuthKey AuthType support for Adyen, utilize review_key in Adyen's payout confirm and cancel API's, restrict operations on terminal state payouts, update payout_method_data in locker for payout update flow

## Type of Change

- [x] Bugfix
- [ ] New feature
- [x] Enhancement
- [x] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

### Additional Changes

## Checklist

### DIFF
--- a/crates/router/src/core/payouts/helpers.rs
+++ b/crates/router/src/core/payouts/helpers.rs
@@ -47,7 +47,7 @@ pub async fn make_payout_method_data<'a>(
             )?;
             Ok(pm)
         }
-        (Some(payout_method), None) => {
+        (Some(payout_method), _) => {
             let payout_token = vault::Vault::store_payout_method_data_in_locker(
                 state,
                 None,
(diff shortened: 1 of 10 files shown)


crates/router/src/core/payouts/helpers.rs
```diff
@@ -47,10 +47,10 @@ pub async fn make_payout_method_data<'a>(
             )?;
             Ok(pm)
         }
-        (Some(payout_method), None) => {
+        (Some(payout_method), payout_token) => {
```
manoj-juspay, line 50: it should be None `store_payout_method_data_in_locker ` will give you the token and we need to update that in `payout_create`

kashif-m: For the cases where payout_method_data is present in the request and payout_token is also present in DB, it will return None.

This case was failing during Update flow.

We need to either disallow changing the payout_type and method_data once a payout is created, or else we need to handle it using the changes above

PR #1208, opened 2023-05-18
refactor: payout flow corrections

kashif-m:
- add status checks before payout fulfillment and cancellation
- fix default PayoutStatus across flows to RequiresCreation

## Type of Change

- [x] Bugfix
- [ ] New feature
- [x] Enhancement
- [x] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

### Additional Changes

## Checklist

### DIFF
--- a/crates/router/src/core/payouts.rs
+++ b/crates/router/src/core/payouts.rs
@@ -96,9 +96,8 @@ pub async fn payouts_update_core(
     if helpers::is_payout_terminal_state(status) {
         return Err(report!(errors::ApiErrorResponse::InvalidRequestData {
             message: format!(
-                "{} is in terminal state [status: {}]",
-                payout_create.payout_id.to_owned(),
-                status
+                "Payout {} cannot be updated for status {}",
+                payout_data.payout_create.payout_id, status
             ),
         }));
     }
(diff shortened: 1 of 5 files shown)


crates/router/src/core/payouts/helpers.rs
```diff
@@ -20,12 +19,12 @@ use crate::{
 
 pub async fn make_payout_method_data<'a>(
     state: &'a AppState,
-    request: &api::PayoutCreateRequest,
+    payout_method_data: &Option<api::PayoutMethodData>,
```
manoj-juspay, line 22: are you getting this from request when you are passing it to this function ?

crates/router/src/core/payouts.rs
```diff
@@ -228,45 +227,69 @@ pub async fn payouts_cancel_core(
...
-        &connector_data,
-        &mut payout_data,
-        &pmd,
-    )
-    .await
-    .attach_printable("Payout fulfillment failed for given Payout request")?;
+        // Trigger fulfillment
+        let payout_method_data = helpers::make_payout_method_data(state, &None, &payout_create)
+            .await?
+            .get_required_value("payout_method_data")?;
```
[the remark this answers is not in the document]
kashif-m: We are passing this in both payout_fulfill and payout_cancel because we have generic functions across all the flows, and this is a mandatory field in PayoutsData which is passed to the connector.

We can remove this dependency by creating separate types for separate actions or making this field Option type

crates/router/src/core/payouts/helpers.rs
kashif-m: Yes, multiple flows needs to use this function and pmd is not present in every request, so we're directly passing the payout_method_data

PR #1230, opened 2023-05-22
fix: remove multiple call to locker

Sangamesh26:
## Type of Change

- [x] Bugfix

## Description

If we have hyperswitch_token as well as pmd, we get pmd against that hyperswitch token from basilisk.
We used to update the pmd again against that token w.r.t wallets and bank transfers, as both the data will be same, we get error.

## Motivation and Context

Remove duplicate storage in locker

## How did you test it?

Manual

## Checklist

### DIFF
--- a/crates/router/src/core/payments/helpers.rs
+++ b/crates/router/src/core/payments/helpers.rs
@@ -782,40 +782,16 @@ pub async fn make_pm_data<'a, F: Clone, R>(
                     }
                 }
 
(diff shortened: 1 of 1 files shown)


manoj-juspay:
@Sangamesh26 if we have hyperswitch_token and also PMD , we can choose to update PMD again ?

jarnura (approved):

Issue #1245, opened 2023-05-23
[BUG] The response body in the OpenAPI spec should derive the `ToResponse` attribute and not the `ToSchema` attribute; "required" tag should not appear in the response bodies


kritikmodi:
### Bug Description

The OpenAPI spec for every endpoint derives the `ToSchema` attribute for the request as well as the response body. This makes the response body contain fields with the `required` tag on them.

### Expected Behavior

The response body should derive the `ToResponse` attribute which can be found [here](https://docs.rs/utoipa/latest/utoipa/derive.ToResponse.html); or the Response bodies must be changed accordingly to remove the `required` tag.

### Actual Behavior

The response body derives the `ToSchema` attribute and some fields are tagged as `required`.

### Steps To Reproduce

The request and response bodies for each endpoint can be seen within `generated.json`.

### Context For The Bug

request-response bodies

### Environment

Local

### Have you spent some time to check if this bug has been raised before?

- [X] I checked and didn't find similar issue

### Have you read the Contributing Guidelines?

- [X] I have read the [Contributing Guidelines](https://github.com/juspay/hyperswitch/blob/main/docs/CONTRIBUTING.md)

SanchithHegde:
Closing this as "wont fix", having fields marked as required in response schemas is indeed beneficial.

My comment on the linked PR (https://github.com/juspay/hyperswitch/pull/1255#issuecomment-1561788426):

> Fields marked as "required" in the response schema indicate that they would always be populated with a non-null value by us. This is indeed a necessary feature and would be beneficial to downstream users of our APIs, and in automated client code generation.

PR #1255, opened 2023-05-24
fix(openapi): Make all fields optional in the response body visible in the OpenAPI spec

kritikmodi:
## Type of Change

- [x] Bugfix

## Description

Made the `required` fields optional in the response body of all the endpoints by wrapping their types with `Option<>` in the derived `schema` trait. This would not affect their actual type, only their types visible on the docs.

## Motivation and Context

This PR solves [this](https://github.com/juspay/hyperswitch/issues/1245) issue.

## How did you test it?

Manually.

## Checklist


SanchithHegde:
Fields marked as "required" in the response schema indicate that they would always be populated with a non-null value by us. This is indeed a necessary feature and would be beneficial to downstream users of our APIs, and in automated client code generation.

Closing this PR.

PR #1264, opened 2023-05-25
feat(kms): reduce redundant kms calls

NishantJoshi00:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

This PR introduces changes to changes to KMS decryption:
- It decrypts and stores the KMS keys for vault and locker in a `AppState`.
- It performs concurrent decryption while startup for those KMS keys
- It also supports, retry mechanism for decryption for failures

### Additional Changes

## Motivation and Context

Improve the performance

## How did you test it?
compiler guided

## Checklist

### DIFF
--- a/crates/router/src/configs/settings.rs
+++ b/crates/router/src/configs/settings.rs
@@ -39,6 +39,13 @@ pub enum Subcommand {
     GenerateOpenapiSpec,
 }
 
+#[cfg(feature = "kms")]
+/// Store the decrypted kms secret values for active use in the application
+#[derive(Clone)]
+pub struct ActiveKmsSecrets {
+    pub jwekey: masking::Secret<Jwekey>,
+}
+
 #[derive(Debug, Deserialize, Clone, Default)]
 #[serde(default)]
 pub struct Settings {
--- a/crates/router/src/configs/settings.rs
+++ b/crates/router/src/configs/settings.rs
@@ -41,6 +41,9 @@ pub enum Subcommand {
 
 #[cfg(feature = "kms")]
 /// Store the decrypted kms secret values for active use in the application
+/// Currently using `StrongSecret` won't have any effect as this struct have smart pointers to heap
+/// allocations.
+/// note: we can consider adding such behaviour in the future with custom implementation
 #[derive(Clone)]
 pub struct ActiveKmsSecrets {
     pub jwekey: masking::Secret<Jwekey>,
--- a/crates/router/src/routes/app.rs
+++ b/crates/router/src/routes/app.rs
@@ -12,6 +12,8 @@ use super::{admin::*, api_keys::*, disputes::*, files::*};
 use super::{configs::*, customers::*, mandates::*, payments::*, payouts::*, refunds::*};
 #[cfg(feature = "oltp")]
 use super::{ephemeral_key::*, payment_methods::*, webhooks::*};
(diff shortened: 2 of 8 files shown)


crates/router/src/routes/app.rs
```diff
@@ -66,6 +68,20 @@ impl AppState {
             StorageImpl::Mock => Box::new(MockDb::new(&conf).await),
         };
 
+        #[cfg(feature = "kms")]
+        let kms_secrets = retry(
+            || {
+                crate::configs::kms::KmsDecrypt::decrypt_inner(
```
jarnura, line 74:
```suggestion
                configs::kms::KmsDecrypt::decrypt_inner(
```

crates/router/src/configs/settings.rs
```diff
@@ -39,6 +39,13 @@ pub enum Subcommand {
     GenerateOpenapiSpec,
 }
 
+#[cfg(feature = "kms")]
+/// Store the decrypted kms secret values for active use in the application
+#[derive(Clone)]
+pub struct ActiveKmsSecrets {
+    pub jwekey: masking::Secret<Jwekey>,
```
ashokkjag, line 46: StrongSecret so that it will be zero'ed out on drop.

jarnura (changes requested):

crates/router/src/routes/app.rs
```diff
@@ -75,14 +91,41 @@ impl AppState {
...
         Self::with_storage(conf, StorageImpl::Postgresql, shut_down_signal).await
     }
 }
 
+#[allow(clippy::panic)]
+///
+/// # Panics
+/// panics of the function fails 2 times
+///
+pub async fn retry<F, T, E, Fut>(func: F, retry_count: u8) -> T
```
ashokkjag, line 109: What error are you trying to recover from? If network issues, an immediate retry won't be of any use.

NishantJoshi00: as @SanchithHegde suggested, I would have added a delay, but it's uncertain is how much time will the network recover. It would be better to just do it once as the server will not boot up, and we can simply redeploy.

crates/router/src/configs/settings.rs
NishantJoshi00: So, my thought process was on the performance side
> The `AppState` is cloned on every request and then dropped at the end. 
> Having it as `StrongSecret` will incur higher cost for every request.

So, what do you think, should we still go with `StrongSecret`?

ashokkjag: If you think the zero'ing out would be costly due to per request cloning, Go ahead with secret but put a note of the same in code.

crates/router/src/routes/app.rs
ashokkjag: Yes Nishant, either put some delay like 1 second. Or give up after first try itself, since this is startup scenario. Anything while the server is up, we may have to think on handling failure/retries.

crates/router/src/configs/settings.rs
NishantJoshi00: Done

jarnura (approved):

ashokkjag (approved):

PR #1269, opened 2023-05-25
feat(core): add custom headers capability for application response and make `JWTAuth` private

### LINKED ISSUE #1291: [ENHANCEMENT] - Add custom headers support while sending response
### Feature Description

Currently, there is no way to control the headers that are being sent in the response. Adding support for that would be a nice feature that many other features could benefit from.

### Possible Implementation

Create a new `ApplicationResponse` which can take custom headers and attach them to the `HttpResponse`.

Yes I am willing to submit a PR!

ThisIsMani:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

- Add custom headers capability to application response. 
- As we are not using `JWTAuth` in other crates., this one is made private.
- `jwt_auth_or` function is removed, which can be replaced by `auth_type`.
- Create a new `ApiErrorResponse` for 403 (`AccessForbidden`).

### Additional Changes

## Motivation and Context

As we can only add headers to `HttpResponse`, which can be only used in `server_wrap`, it's better if we can control it in the inner function itself, which only has access to `ApplicationResponse`. So I have added this new response type.
Closes #1291 

## How did you test it?

Postman

## Checklist

### DIFF
--- a/crates/router/src/compatibility/wrap.rs
+++ b/crates/router/src/compatibility/wrap.rs
@@ -40,29 +40,33 @@ where
         api_authentication,
     )
     .await;
-    handle_application_response::<Q, S, E>(request, resp)
-}
-
-pub fn handle_application_response<Q, S, E>(
-    request: &HttpRequest,
-    server_resp: Result<api::ApplicationResponse<Q>, error_stack::Report<E>>,
-) -> HttpResponse
-where
-    Q: Serialize + std::fmt::Debug,
-    E: Serialize + error_stack::Context + actix_web::ResponseError + Clone,
-    S: TryFrom<Q> + Serialize,
-    error_stack::Report<E>: services::EmbedError,
-{
-    match server_resp {
+    match resp {
         Ok(api::ApplicationResponse::ResponseWithCustomHeader(response, headers)) => {
-            let mut final_response = handle_application_response::<Q, S, E>(request, Ok(*response));
+            let mut final_response = handle_application_response::<Q, S>(request, *response);
             let inner_headers = final_response.headers_mut();
             headers
                 .iter()
                 .for_each(|(name, value)| inner_headers.append(name.to_owned(), value.to_owned()));
             final_response
         }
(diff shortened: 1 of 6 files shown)


crates/router/src/compatibility/wrap.rs
```diff
@@ -40,7 +40,28 @@ where
...
+) -> HttpResponse
+where
+    Q: Serialize + std::fmt::Debug,
+    E: Serialize + error_stack::Context + actix_web::ResponseError + Clone,
+    S: TryFrom<Q> + Serialize,
+    error_stack::Report<E>: services::EmbedError,
+{
+    match server_resp {
+        Ok(api::ApplicationResponse::ResponseWithCustomHeader(response, headers)) => {
+            let mut final_response = handle_application_response::<Q, S, E>(request, Ok(*response));
```
jarnura, line 58: Why the function calling itself?

jarnura (changes requested):
Add sub-issue, milestone, projects.

crates/router/src/compatibility/wrap.rs
ThisIsMani: `ResponseWithCustomHeader` has two fields. The first field is `response` of type `ApplicationResponse`, and the second field is `headers` which is a `HeaderMap`. The function `handle_application_response` takes an `ApplicationResponse` as input and returns an `HttpResponse`. To obtain the `HttpResponse` for the inner `response`, the function is recursively called again. Similar behavior can be observed in other functions as well.

lsampras:
@ThisIsMani 
As discussed offline,
can we try to do this via an actix middleware,

you can close this PR if that approach works out...

cc: @SanchithHegde @jarnura

ThisIsMani:
This is closed. Actix middleware and server_wrap is being used to achieve this.

PR #1286, opened 2023-05-28
chore(common_utils): Apply the new type pattern for phone numbers

vlad-onis:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

* Used the phonenumber crate to validate the phonenumber
* Implemented usual traits: FromStr, TryFrom, Deref, DerefMut, Queryable, FromSql, ToSql
* Used the testcase crate to write a few phone number validation unit tests

### Additional Changes

## Checklist


crates/common_utils/src/pii.rs
```diff
@@ -25,6 +26,103 @@ pub const REDACTED: &str = "Redacted";
 /// Type alias for serde_json value which has Secret Information
 pub type SecretSerdeValue = Secret<serde_json::Value>;
 
+/// Phone Number
+#[derive(Debug, serde::Deserialize, serde::Serialize)]
```
lsampras, line 29: use serde [try_from](https://serde.rs/container-attrs.html#try_from) attribute for deserialization here....
Currently Deserialize won't validate for phone numbers,

You'll need to 
```
impl TryFrom<String> for PhoneNumber
or
impl TryFrom<Secret<String>> for PhoneNumber
```

we can instead use `try_from = String/Secret<String>` that way serde deserializes it to the corresponding struct and then runs the try_from method to get the correct PhoneNumber or throw a ParsingError

crates/common_utils/src/pii.rs
```diff
@@ -25,6 +26,103 @@ pub const REDACTED: &str = "Redacted";
...
+        let secret = Secret::<String>::new(phone_number.to_string());
+        Ok(Self(secret))
+    }
+}
+
+impl TryFrom<String> for PhoneNumber {
+    type Error = error_stack::Report<errors::ParsingError>;
+
+    fn try_from(value: String) -> Result<Self, Self::Error> {
+        Self::from_str(&value).change_context(errors::ParsingError::EmailParsingError)
```
lsampras, line 66: Can you update this to return a PhoneNumberParsingError to better reflect the error?

You can add the corresponding error variant in the ParsingError enum

crates/common_utils/src/errors.rs
```diff
@@ -29,11 +29,14 @@ pub enum ParsingError {
     /// Failed to parse email
     #[error("Failed to parse email")]
     EmailParsingError,
+    /// Failed to parse email
```
lsampras, line 32:
```suggestion
    /// Failed to parse phone number
```

crates/common_utils/src/pii.rs
```diff
@@ -24,6 +25,103 @@ pub const REDACTED: &str = "Redacted";
...
+}
+
+impl ops::DerefMut for PhoneNumber {
+    fn deref_mut(&mut self) -> &mut Self::Target {
+        &mut self.0
+    }
+}
+
+impl<DB> Queryable<diesel::sql_types::Text, DB> for PhoneNumber
+where
```
lsampras, line 93: Wondering whether we can reduce this repetition associated with NewTypes...
maybe we can create a generic trait or a macro_rules....

We could take this up in the future maybe
@SanchithHegde should we create an issue for this^

crates/common_utils/Cargo.toml
```diff
@@ -33,6 +33,8 @@ tokio = { version = "1.27.0", features = ["macros", "rt-multi-thread"], optional
 thiserror = "1.0.40"
 time = { version = "0.3.20", features = ["serde", "serde-well-known", "std"] }
 md5 = "0.7.0"
+phonenumber = "0.3.2+8.13.9"
+test-case = "3.1.0"
```
lsampras, line 37: Not sure if we wanna add this crate,
we could do some analysis to check how much value it adds compared to the additional compilation time/size etc

Anyways can you move this to dev-dependencies...

lsampras (commented):
Thanks for the PR,

I have a couple of suggestions/comments...
Let me know what you think of them.

crates/common_utils/src/pii.rs
SanchithHegde: Yeah we can create an issue for this.

crates/common_utils/src/pii.rs
vlad-onis: Got some inspiration from the Email new type, will clean all leftovers

vlad-onis:
I have a question sorry to tag you both again @lsampras @SanchithHegde 

When applying the changes needed to use the new type throughout the codebase, I managed so far but I'm uncertain about a few things:

1. Should Request types only contain Secret<String> (I think yes) or should they also use the newType instead?
2. Address types (e.g. storage_models/address.rs Address) contain the phone number as secret, if I change that to PhoneNumber it creates a huge ripple effect. If I don't change it to PhoneNumber I need a way to create a PhoneNumber from a Secret<String> but this will break the validation as there is no way to ensure the Secret is right. Or is there a way to retrieve the inner value of the Secret?

SanchithHegde:
> This PR still has some way to go until it's done. I will mark it as draft for now. Trying to finish it this week otherwise I will only get the chance to go back to it next Friday (9th of June).

I've attached the `S-in-progress` label.



> When applying the changes needed to use the new type throughout the codebase, [...]

@vlad-onis Could you please restrict this PR to only the introduction of the newtype? Preferably, we'd like to take up the refactor of replacing phone number fields with the newtype in a separate PR, since it would cause conflicts to you and other folks. I can create a new issue for the refactor, if you'd prefer.

SanchithHegde:
You should be able to use the newtype in both the request/response types, and in the storage types.

As for the implementations required for the storage types, you can use the `Email` type as a reference.

crates/common_utils/src/pii.rs
```diff
@@ -24,6 +25,103 @@ pub const REDACTED: &str = "Redacted";
...
+
+        write!(
+            f,
+            "{}{}{}",
+            &val_str[..2],
+            "*".repeat(val_str.len() - 5),
+            &val_str[(val_str.len() - 3)..]
+        )
+    }
+}
```
SanchithHegde, line 51: Please use a different unit struct for `PhoneNumberStrategy`, implement `Strategy` for `PhoneNumberStrategy` and update `PhoneNumber` to use it instead, like so:

```rust
struct PhoneNumberStrategy;

impl Strategy for PhoneNumber {
    // ...
}

struct PhoneNumber(Secret<String, PhoneNumberStrategy>);
```

For more information, please refer to this comment: https://github.com/juspay/hyperswitch/pull/889#discussion_r1168201823

lsampras:
> Should Request types only contain Secret (I think yes) or should they also use the newType instead?

We prefer to use them in newtypes because then we can re-use the error-handling scaffold for de-serialize errors, where serde neatly formats out the error message as ApiResponseError along with appropriate status code


> .... I need a way to create a PhoneNumber from a Secret but this will break the validation as there is no way to ensure the Secret is right. Or is there a way to retrieve the inner value of the Secret?

Yup, you are right we shouldn't support `impl From<Secret<...>> for PhoneNumber`,
I see that our implementation with `impl From<Secret<...>> for Email` is also built with incorrect assumptions...

We should instead implement `impl TryFrom<Secret<...>> for Email/PhoneNumber`...

@SanchithHegde Should we create another issue for fixing `impl From<Secret<...>> for Email`....

SanchithHegde:
> Yup, you are right we shouldn't support `impl From<Secret<...>> for PhoneNumber`, I see that our implementation with `impl From<Secret<...>> for Email` is also built with incorrect assumptions...
> 
> We should instead implement `impl TryFrom<Secret<...>> for Email/PhoneNumber`...
> 
> @SanchithHegde Should we create another issue for fixing `impl From<Secret<...>> for Email`....

I believe the `From` implementation is not used anywhere in our codebase, PR #1074 can be reverted. @vspecky Please confirm.

vspecky:
> I believe the `From` implementation is not used anywhere in our codebase, PR #1074 can be reverted. @vspecky Please confirm.
Yes, the code in the mentioned change can be removed.

vlad-onis:
The above comments should now be addressed, together with the clippy warnings

crates/common_utils/src/pii.rs
```diff
@@ -25,6 +25,125 @@ pub const REDACTED: &str = "Redacted";
...
 pub type SecretSerdeValue = Secret<serde_json::Value>;
 
+/// Strategy for masking a PhoneNumber
+#[derive(Debug)]
+pub struct PhoneNumberStrategy;
+
+/// Phone Number
+#[derive(Debug, serde::Deserialize, serde::Serialize)]
+#[serde(try_from = "String")]
+pub struct PhoneNumber(Secret<String, PhoneNumberStrategy>);
```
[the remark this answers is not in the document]
lsampras: Do you see a use case for custom formatting?

Currently I'd prefer this format due to its less complexity

e.g
```
struct Student {
    id: u8,
    phone_number: PhoneNumber,
    email_address: Email,
}
```

vs

```
struct Student {
    id: u8,
    phone_number: Secret<PhoneNumber, PhoneNumberStrategy>,
    email_address: Secret<Email, EmailStrategy>,
}
```

Are there any other rules/guidelines that we would be breaking in a contextual sense?

For e.g in this case we are equating that a phone number is always gonna be a secret, should the type phonenumber worry about the formatting of its value or should it only care about its data validity...

For now since we haven't reached any problems that need to separate this concerns, let's keep it this (Since it's easy to extend as well)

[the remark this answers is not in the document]
SanchithHegde: Fair enough, resolving this conversation.

vlad-onis:
Resolved the last few comments above.

crates/common_utils/src/errors.rs
SanchithHegde: @vlad-onis This minor change is pending.

SanchithHegde (commented):

SanchithHegde (approved):
Thanks for the PR, @vlad-onis!

PR #1302, opened 2023-05-29
refactor(core): use HMAC-SHA512 to calculate payments response hash

ShankarSinghC:
## Type of Change

- [x] Refactoring

## Description

Use the SHA-512 to hash the payment response as SHA-512 is generally considered to be more secure than SHA-256 due to its longer hash output. Also the length of the hash key being generate is modified (64 bytes), if not SHA-512 does padding to make the input key length into 64 bytes. And having 32 bytes padded in a key is not secure.

## How did you test it?

Hash key generated screenshots
Screenshot 2023-05-29 at 3 11 20 PM

Screenshot 2023-05-29 at 3 20 57 PM

## Checklist


Narayanbhat166 (approved):

SanchithHegde (approved):

ashokkjag (approved):
Just verify if some SDK side changes are needed.

PR #1311, opened 2023-05-29
fix(Adyen): Address Internal Server Error when calling PSync without redirection

pixincreate:
## Type of Change

- [x] Bugfix

## Description

This PR address Internal Server Error that arised when calling PSync in non-redirection flow.

## Motivation and Context

Doing a PSync call used to `500, Internal Server Error` as the `encoded_data` passed `None` value during non-redirection flow.

## How did you test it?

Couldn't test by running `adyen.rs` as API keys contain symbols such as `)` and `]` that result in internal `toml` error. Tested manually as of now.
image

## Checklist


Narayanbhat166 (changes requested):
This is not how we are supposed to handle it. The call should not reach connector module if it cannot handle it. Handle this case at the core level, in this place
https://github.com/juspay/hyperswitch/blob/c3e4556096d772b7b1b5f84d761de7a680bac31a/crates/router/src/core/payments/operations/payment_status.rs#L280

crates/router/src/connector/adyen.rs
```diff
@@ -314,7 +314,10 @@ impl
             .encoded_data
             .clone()
             .get_required_value("encoded_data")
-            .change_context(errors::ConnectorError::RequestEncodingFailed)?;
+            .change_context(errors::ConnectorError::FlowNotSupported {
+                flow: String::from("PSync"),
+                connector: String::from("Adyen"),
```
vspecky, line 319: Not a blocker but better to use `ConnectorCommon::id()` here (`self.id()`)

vspecky (approved):
LGTM

vspecky (approved):

Sangamesh26 (dismissed):
LGTM

vspecky (approved):

lsampras (approved):

PR #1318, opened 2023-05-30
feat: filter out payment_methods which does not support mandates during list api call

Chethan-rao:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

When a mandate payment is created, sdk makes a payment_method_list api call which lists all the payment_methods. But not all of them support mandate payments.
Added a filter for payment_methods which doesn't support mandates and list only those which supports, during payment_method_list api call.
Created an hashmap containing info about which payment_methods through which connector, supports mandates. And while listing payment methods, validate between the merchant configured payment_methods and the supported payment_methods.

### Additional Changes

crates/router/src/configs/settings.rs

## Motivation and Context

List the payment_methods which support mandates filtering out others.

## How did you test it?

Postman
Configured payment methods - 
Screenshot 2023-05-30 at 4 08 41 PM
Screenshot 2023-05-30 at 4 09 01 PM

Filtered payment methods - 
Screenshot 2023-05-30 at 4 10 07 PM

## Checklist


crates/router/src/core/payment_methods/cards.rs
```diff
@@ -1218,6 +1223,21 @@ pub async fn filter_payment_methods(
                         &payment_method_object.payment_method_type,
                     );
 
+                    let connector_variant = api_enums::Connector::from_str(connector.as_str())
+                        .into_report()
+                        .change_context(errors::ApiErrorResponse::InternalServerError)?;
```
lsampras, line 1228: 400 bad request invalid connector?

lsampras (approved):

crates/router/src/core/payment_methods/cards.rs
Chethan-rao: right, shall i change_context to `ConnectorError::InvalidConnectorName` first and then to `ApiErrorResponse::InternalServerError` ?

crates/router/src/core/payment_methods/cards.rs
```diff
@@ -1225,7 +1225,11 @@ pub async fn filter_payment_methods(
 
                     let connector_variant = api_enums::Connector::from_str(connector.as_str())
                         .into_report()
-                        .change_context(errors::ApiErrorResponse::InternalServerError)?;
+                        .change_context(errors::ConnectorError::InvalidConnectorName)
+                        .change_context(errors::ApiErrorResponse::InternalServerError)
```
lsampras, line 1229: invalid connector bad request?

lsampras (changes requested):

crates/router/src/core/payment_methods/cards.rs
Chethan-rao: U mean this cannot be internal server error? could u please provide more info

crates/router/src/configs/settings.rs
```diff
@@ -83,6 +83,7 @@ pub struct Settings {
     pub dummy_connector: DummyConnector,
     #[cfg(feature = "email")]
     pub email: EmailSettings,
+    pub supported_payment_methods_for_mandate: SupportedPaymentMethodsForMandate,
```
jarnura, line 86: In settings, new things where added. Add the below this too.
- Add info in config.example.toml.
- Basic configs in development.toml

jarnura (changes requested):

Narayanbhat166 (dismissed):

SanchithHegde (approved):

Narayanbhat166 (approved):

SanchithHegde (approved):

lsampras (approved):

crates/router/src/configs/settings.rs
```diff
@@ -127,6 +128,27 @@ pub struct DummyConnector {
...
 }
 
+#[derive(Debug, Deserialize, Clone)]
+pub struct Mandates {
+    pub supported_payment_methods: SupportedPaymentMethodsForMandate,
+}
+
+#[derive(Debug, Deserialize, Clone)]
+pub struct SupportedPaymentMethodsForMandate(
+    pub HashMap<enums::PaymentMethod, SupportedPaymentMethodTypesForMandate>,
```
ashokkjag, line 138: Do we need this hashmap be built through config or can there be a global static? I understand that having a config based approach will let us turn things on/off during startup, but the config structure keeps getting bloated, which gets cloned, passed to each actix thread.

ashokkjag (approved):

jarnura (approved):

crates/router/src/configs/settings.rs
Chethan-rao: I agree that config structure is getting messy, but making it a global static will require re-deploying the application if any changes in the config right?

PR #1333, opened 2023-06-01
fix(router/webhooks): correct webhook error mapping and make source verification optional for all connectors

vspecky:
## Type of Change

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

This PR is intended to perform two bugfixes :-
- Correction in the error mapping from webhook connector errors to API error response. This is done specifically for webhook event type mapping failure cases.
- Making source verification merchant secret retrieval optional for all connectors. This can be optional because some webhook consumption flows can go for a fallback when source verification has failed.

### Additional Changes
None

## Motivation and Context

This set of bug fixes is aimed at eliminating redundant 5xx API errors, instead making the API errors more sensitive to the actual context in which the error occurred.

## How did you test it?

Manual, compiler-guided.

## Checklist

### DIFF
--- a/crates/router/src/connector/adyen.rs
+++ b/crates/router/src/connector/adyen.rs
@@ -11,6 +11,7 @@ use storage_models::enums as storage_enums;
 use self::transformers as adyen;
 use crate::{
     configs::settings,
+    connector::utils as conn_utils,
     consts,
     core::errors::{self, CustomResult},
     db::StorageInterface,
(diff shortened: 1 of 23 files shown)


crates/router/src/connector/adyen.rs
```diff
@@ -797,13 +798,17 @@ impl api::IncomingWebhook for Adyen {
...
         merchant_id: &str,
     ) -> CustomResult<Vec<u8>, errors::ConnectorError> {
-        let key = format!("whsec_verification_{}_{}", self.id(), merchant_id);
-        let secret = db
-            .get_key(&key)
-            .await
-            .change_context(errors::ConnectorError::WebhookVerificationSecretNotFound)?;
-
-        Ok(secret)
+        let key = conn_utils::get_webhook_merchant_secret_key(self.id(), merchant_id);
```
sai-harsha-vardhan, line 801: Since this logic is redundant in all the connectors, can we have it as the default implementation?

sai-harsha-vardhan (changes requested):

crates/router/src/connector/adyen.rs
sai-harsha-vardhan: I think it's done here -> https://github.com/juspay/hyperswitch/pull/1331/files#diff-8f740d4eb7ef77de424329a12be7f7a12b09fb805a6551d25e7d8edbda606ddf @vspecky

vspecky: @sai-harsha-vardhan Yes, so I think we can avoid blocking this PR on the default implementation as opposed to doing the implementation again here.

sai-harsha-vardhan (approved):

SanchithHegde (approved):

lsampras (approved):

PR #1336, opened 2023-06-02
fix: set `locker_id` to a default value in db

ThisIsMani:
## Type of Change

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

This will set a default value for the `locker_id` field in `merchant_account` table.

### Additional Changes

## Motivation and Context

This is to fix the saved cards not being accessed because of `locker_id` is null

## How did you test it?

Postman

## Checklist

### DIFF
--- a/crates/masking/tests/basic.rs
+++ b/crates/masking/tests/basic.rs
@@ -40,11 +40,6 @@ fn basic() -> Result<(), Box<dyn std::error::Error + Send + Sync>> {
         not_secret,
     };
 
-    // clone
-
-    let composite2 = composite.clone();
-    assert_eq!(composite, composite2);
-
     // format
 
     let got = format!("{composite:?}");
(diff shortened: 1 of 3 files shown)


dracarys18 (approved):

lsampras (approved):

SanchithHegde:
Although this may be a solution to the problem of not being able to use saved cards because of `locker_id` being `NULL` in the database, I'd not prefer having it, since the locker ID `m0010` does not provide any extra information; the default locker ID could have been something generic as well, such as `default_locker_id` or just `default`. Moreover, this does not seem like the right solution to me, period.

Will be closing this PR for the same reason.

PR #1341, opened 2023-06-02
fix(payment_method): do not save card in locker in case of error from connector

dracarys18:
## Type of Change

- [x] Bugfix

## Description

Don't save card in card locker in case of error from connector.

## Motivation and Context

Currently we are saving card in card locker even in case of error in connector right now.

## How did you test it?

Manual

## Checklist


Narayanbhat166 (dismissed):

crates/router/src/core/payments/tokenization.rs
```diff
@@ -31,102 +31,115 @@ pub async fn save_payment_method<F: Clone, FData>(
...
+                                        .change_context(
+                                            errors::ApiErrorResponse::InternalServerError,
+                                        )
+                                        .attach_printable("Failed to add payment method in db")
+                                    }
+                                    _ => {
+                                        Err(report!(errors::ApiErrorResponse::InternalServerError))
+                                    }
+                                },
+                                _ => Err(report!(errors::ApiErrorResponse::InternalServerError)),
```
SanchithHegde, line 117: Consider adding reasons for raising `InternalServerError`.

crates/router/src/core/payments/tokenization.rs
```diff
@@ -31,102 +31,115 @@ pub async fn save_payment_method<F: Clone, FData>(
...
-            .attach_printable("Failed to add payment method in db")?;
-        };
-        Some(locker_response.0.payment_method_id)
-    } else {
-        None
-    };
-    Ok(pm_id)
+            Ok(pm_id)
+        }
+        Err(_) => Ok(None),
```
lsampras, line 141: What are the standards followed on this,
do we not save the card because it might be invalid?

What if we receive errors that point to other errors such as insufficient balance etc...

SanchithHegde (approved):

crates/router/src/core/payments/tokenization.rs
dracarys18: Connectors have different errors for card specific errors it's not easy enough to map them into one

Narayanbhat166 (approved):

lsampras (approved):

crates/router/src/core/payments/tokenization.rs
lsampras: #1599 Added an issue to pick this up later

PR #1352, opened 2023-06-05
fix(encryption): do not log encrypted binary data

dracarys18:
## Type of Change

- [x] Enhancement

## Description

Add secret to encrypted fields so it wont be logged. It will be logged as `*** Encrypted ***`.

## Motivation and Context

Dont log huge encrypted binary data.

## How did you test it?

image

## Checklist

### DIFF
--- a/crates/router/src/types/domain/types.rs
+++ b/crates/router/src/types/domain/types.rs
@@ -55,7 +55,7 @@ impl<
         crypt_algo: V,
     ) -> CustomResult<Self, errors::CryptoError> {
         let encrypted = encrypted_data.into_inner();
-        let data = crypt_algo.decode_message(key, encrypted.clone().expose())?;
+        let data = crypt_algo.decode_message(key, encrypted.clone())?;
 
         let value: String = std::str::from_utf8(&data)
             .into_report()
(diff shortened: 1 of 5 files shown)


SanchithHegde (approved):

SanchithHegde (dismissed):

crates/router/src/types/domain/types.rs
```diff
@@ -94,7 +94,7 @@ impl<
         crypt_algo: V,
     ) -> CustomResult<Self, errors::CryptoError> {
         let encrypted = encrypted_data.into_inner();
-        let data = crypt_algo.decode_message(key, encrypted.clone())?;
+        let data = crypt_algo.decode_message(key, encrypted.clone().expose())?;
```
NishantJoshi00, line 97: don't expose if possible pass the secret directly. If need be you can take it up in a separate PR

crates/router/src/types/domain/types.rs
```diff
@@ -128,7 +128,7 @@ impl<
...
```
NishantJoshi00, line 131: same here

NishantJoshi00 (changes requested):

NishantJoshi00 (approved):

SanchithHegde (approved):

Issue #1357, opened 2023-06-06
[FEATURE] add domain type for client secret 


Abhicodes-crypto:
### Feature Description

Create domain type for client secret

### Possible Implementation

Make type `ClietSecret(String)` and use it wherever applicable

### Have you spent some time to check if this feature request has been raised before?

- [X] I checked and didn't find similar issue

### Have you read the Contributing Guidelines?

- [X] I have read the [Contributing Guidelines](https://github.com/juspay/hyperswitch/blob/main/docs/CONTRIBUTING.md)

### Are you willing to submit a PR?

No, but I'm happy to collaborate on a PR with someone else

slelyukh:
Hello, can you assign this issue to me?

tusharv01:
Any suggestions where to make changes as Client Secret files are not there in the codebase.

amankrx:
Can you assign me this issue?

Abhicodes-crypto:
You can add the type in this file https://github.com/juspay/hyperswitch/blob/main/crates/api_models/src/payments.rs 
And you need to implement serialize and dezerialize for the type. You can refer this as an example https://github.com/juspay/hyperswitch/blob/379d1d1375783f2c35edbf4dda6bbb0eb9351a3c/crates/api_models/src/api_keys.rs#L191 

Also add test cases to make sure the serialize and deserialize work. For an idea of the format of client secret refer PR #1358

VedantKhairnar:
Hello @tusharv01  , please let us know if you're still working on this issue or still interested in working on it. There hasn't been any recent activity on this thread, or any open PRs.
.
Meanwhile we would be un-assigning you from the issue and close the PR but if you still want to work on it, please let us know.

cb-dhivyadandapani:
Can i pick this up?

deepanshu-iiitu:
@cb-dhivyadandapani I have assigned 2 other issues to you. We can assign you more once you complete those 2.

Sushank34:
Hello team, could you please assign this issue to me?

deepanshu-iiitu:
Hey @cb-dhivyadandapani @Sushank34 
This issue is not a part of Hacktoberfest.
You can use the labels Hacktoberfest and S-awaiting-triage to look for unassigned issues that are part of Hacktoberfest 2023

wowinter13:
@Abhicodes-crypto @VedantKhairnar hello there guys, could you assign this issue to me? 
[...]

PR #1358, opened 2023-06-06
fix(payments): fix client secret parsing

Abhicodes-crypto:
## Type of Change

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

Client secret parsing logic could create bugs changed it to avoid them. 

Ideally , it will be good to pass payment_id everywhere and ask for payment_id in the request wherever applicable and never extract payment_id from client_secret. To address it in this way many changes need to be made So I have addressed the issue in this PR differently.

It would also be good to have a domain type for client secret #1357 issue is created for it.

Do let me know if payment id is extracted from client secret else where in the codebase. I have done a basic search and changed it wherever I found.

Also I have used this error `ClientSecretInvalid` for now. Once we have domain type and appropriate error would be raised during Deserialising. But do let me know if new error should be created for this.

### Additional Changes

## How did you test it?

Added and Ran tests succesfully

## Checklist


jarnura (dismissed):

sai-harsha-vardhan (dismissed):

Abhicodes-crypto:
> what if we provide a function called find_payment_intent_by_client_secret instead of doing the client secret split and extract payment id out of it

Then we have to add index on it. We already have index on payment_id. Too many indexes on DB is not desirable. 

Plus in which ever flow we have client secret it is better to accept payment_id also. I have created an issue for the same. See PR description.

Narayanbhat166 (approved):

sai-harsha-vardhan (approved):

dracarys18 (approved):

PR #1360, opened 2023-06-06
refactor(disputes): resolve incorrect 5xx error mappings for disputes

Abhicodes-crypto:
## Type of Change

- [x] Bugfix
- [x] Refactoring

## Description

Mappings have been created from Connectors to Disputes to handle 5xx errors.

## Motivation and Context

This change is required since right now every error is being marked as 500 which is not correct.

## How did you test it?

Manual testing/compiles.

## Checklist


crates/router/src/core/errors/api_error_response.rs
```diff
@@ -262,6 +264,7 @@ impl actix_web::ResponseError for ApiErrorResponse {
             | Self::VerificationFailed { .. }
             | Self::PaymentUnexpectedState { .. }
             | Self::MandateValidationFailed { .. }
+            | Self::DisputeFailed { .. }
```
sai-harsha-vardhan, line 267: why are we marking `DisputeFailed` as `BAD_REQUEST`?

crates/router/src/core/errors/api_error_response.rs
```diff
@@ -428,6 +431,9 @@ impl common_utils::errors::ErrorSwitch<api_models::errors::types::ApiErrorRespon
             Self::PaymentCaptureFailed { data } => {
                 AER::BadRequest(ApiError::new("CE", 3, "Capture attempt failed while processing with connector", Some(Extra { data: data.clone(), ..Default::default()})))
             }
+            Self::DisputeFailed { data } => {
+                AER::BadRequest(ApiError::new("CE", 1, "Dispute failed during authorization with connector. Retry operation", Some(Extra { data: data.clone(), ..Default::default()})))
```
sai-harsha-vardhan, line 435:
```suggestion
                AER::BadRequest(ApiError::new("CE", 1, "Dispute operation failed while processing with connector. Retry operation", Some(Extra { data: data.clone(), ..Default::default()})))
```

sai-harsha-vardhan (changes requested):

crates/router/src/core/errors/api_error_response.rs
kritikmodi: Not sure :/

kritikmodi (approved):

PR #1363, opened 2023-06-06
feat: add error type for empty connector list

Narayanbhat166:
## Type of Change

- [x] New feature

## Description

This PR adds a new error type called `PaymentMethodUnactivated` which can be used when we get no possible connector after eligibility analysis.

https://stripe.com/docs/error-codes#payment-method-unactivated

## Motivation and Context

Change the 5xx to 4xx error since this is a misconfiguration error.

## How did you test it?

The error type will be used in a different crate.

## Checklist


Abhicodes-crypto (dismissed):

crates/router/src/compatibility/stripe/errors.rs
```diff
@@ -198,6 +198,8 @@ pub enum StripeErrorCode {
     FileNotAvailable,
     #[error(error_type = StripeErrorType::HyperswitchError, code = "", message = "There was an issue with processing webhooks")]
     WebhookProcessingError,
+    #[error(error_type = StripeErrorType::InvalidRequestError, code = "payment_method_unactivated", message = "The operation cannot be performed as the payment method used has not been activated. Activate the payment method in the Dashboard, then try again.")]
+    PaymentMethodUnactivated,
```
vspecky, line 202: An inactive payment method is just one of the factors involved in an eligibility analysis failure. There can be multiple other factors like
- amount exceeding maximum amount
- amount less than minimum amount
- payment method not supported by the connector
- capture method not supported by the connector
- 3DS not supported by the connector
etc.

Pinpointing the exact issue in eligibility analysis is a hard problem since it can be one factor or a combination of n factors for every connector of the merchant. Solving this problem is currently in the future scope, so we cannot pinpoint a particular factor for the failure. For now, it will be a general `EligibilityAnalysisFailure` only

vspecky (changes requested):

crates/router/src/compatibility/stripe/errors.rs
Narayanbhat166: But this should not be a 5xx error, ( which it is currently ). The exact error should be propogated to the user / caller. In the current case the only way to check if something happened is to check logs and mca configuration

Narayanbhat166: The error type variant is changed to `IncorrectpaymentMethodConfigured`

crates/router/src/core/errors/api_error_response.rs
```diff
@@ -194,6 +194,8 @@ pub enum ApiErrorResponse {
     WebhookBadRequest,
     #[error(error_type = ErrorType::RouterError, code = "WE_03", message = "There was some issue processing the webhook")]
     WebhookProcessingFailure,
+    #[error(error_type = ErrorType::InvalidRequestError, code = "HE_04", message = "Required payment method is not configured / configured incorrectly for the connector")]
```
vspecky, line 197:
```suggestion
    #[error(error_type = ErrorType::InvalidRequestError, code = "HE_04", message = "required payment method is not configured or configured incorrectly for all configured connectors")]
```

crates/router/src/compatibility/stripe/errors.rs
vspecky: Can we make this change in Stripe Compatibility as well?

vspecky (changes requested):

crates/router/src/compatibility/stripe/errors.rs
Narayanbhat166: Stripe does not have this error type, so we map it to the available error in stripe

bernard-eugine (approved):

NishantJoshi00 (approved):

PR #1371, opened 2023-06-06
feat(Connector): [Noon] Add reference id in Order Struct

SamraatBansal:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [X] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

Add additional field in the Noon Payments Request, so that payment can be identified in the flows like Webhooks

### Additional Changes

## Motivation and Context

The tunnelling via Euler needs this to recognise Webhooks

## Checklist

### DIFF
--- a/crates/router/src/connector/noon/transformers.rs
+++ b/crates/router/src/connector/noon/transformers.rs
@@ -37,6 +37,7 @@ pub struct NoonOrder {
     currency: Option<storage_models::enums::Currency>,
     channel: NoonChannels,
     category: Option<String>,
+    reference: String,
     //Short description of the order.
     name: String,
 }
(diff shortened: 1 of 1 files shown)


crates/router/src/connector/noon/transformers.rs
```diff
@@ -142,6 +143,7 @@ impl TryFrom<&types::PaymentsAuthorizeRouterData> for NoonPaymentsRequest {
             currency,
             channel: NoonChannels::Web,
             category,
+            reference: item.payment_id.clone(),
```
jagan-jaya, line 146: why are we sending the payment_id here,  use `item.attempt_id

jagan-jaya (changes requested):

crates/router/src/connector/noon/transformers.rs
jagan-jaya: As discussed with Shakthi, we are passing this payment_id from Euler directly, so that it can be used as a reference during webhooks.

jagan-jaya (approved):

jarnura (approved):

PR #1376, opened 2023-06-07
refactor(payments): attempt to address unintended 5xx and 4xx in payments

Abhicodes-crypto:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [x] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

attempt to address unintended 5xx and 4xx in payments

### Additional Changes

## Checklist

### DIFF
--- a/crates/router/src/core/payments/helpers.rs
+++ b/crates/router/src/core/payments/helpers.rs
@@ -843,7 +843,7 @@ pub async fn create_customer_if_not_exist<'a, F: Clone, R>(
             None => None,
             Some(customer_id) => db
                 .find_customer_optional_by_customer_id_merchant_id(customer_id, merchant_id)
-                .await?
+                .await? // if customer_id is present in payment_intent but not found in customer table then shouldn't an error be thrown ?
                 .map(Ok),
         },
     };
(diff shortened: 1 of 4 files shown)


jarnura (approved):

crates/router/src/core/payments/helpers.rs
```diff
@@ -843,7 +843,7 @@ pub async fn create_customer_if_not_exist<'a, F: Clone, R>(
             None => None,
             Some(customer_id) => db
                 .find_customer_optional_by_customer_id_merchant_id(customer_id, merchant_id)
-                .await?
+                .await? // if customer_id is present in payment_intent but not found in customer table then shouldn't an error be thrown ?
```
Abhicodes-crypto, line 846: if customer_id is present in payment_intent but not found in customer table then shouldn't an error be thrown ? @jarnura @Narayanbhat166

Narayanbhat166: This cannot be the case

Abhicodes-crypto: If the customer_id was found in payment_intent and then if customer is not found then the function would return None. But then it should be returning Internal Server Error. Here there is no distinction between customer_id found in request or payment_intent . 

I feel this needs to be looked into . 

Resolving this for now.

Narayanbhat166: If the customer id is not found, then it will create the customer with that customer id. 
Please ask such doubts when the PR is raised by pointing to appropriate line in the diff and not in the code.

PR #1382, opened 2023-06-07
fix(connector): implement ConnectorErrorExt for error_stack::Result<T, ConnectorError>

Abhicodes-crypto:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [x] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

 impl ConnectoExt for Result<T, ConnectorError> to reduce some code.

### Additional Changes

## How did you test it?

compiles

## Checklist

### DIFF
--- a/crates/router/src/core/payments/operations/payment_response.rs
+++ b/crates/router/src/core/payments/operations/payment_response.rs
@@ -300,8 +300,8 @@ async fn payment_response_update_tracker<F: Clone, T: types::Capturable>(
             Some(storage::PaymentAttemptUpdate::ErrorUpdate {
                 connector: None,
                 status: match err.status_code {
-                    400..=499 => storage::enums::AttemptStatus::Failure,
-                    _ => storage::enums::AttemptStatus::Pending,
+                    500..=511 => storage::enums::AttemptStatus::Pending,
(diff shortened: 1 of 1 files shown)


Narayanbhat166 (dismissed):

crates/router/src/core/errors/utils.rs
```diff
@@ -43,111 +43,119 @@ impl<T> StorageErrorExt<T, errors::ApiErrorResponse>
     }
 }
 
-pub trait ConnectorErrorExt {
+pub trait ConnectorErrorExt<T> {
```
SanchithHegde, line 46: Do you think naming this `ConnectorResultExt` makes more sense, since we're implementing the trait on `Result`?

SanchithHegde (commented):

crates/router/src/core/errors/utils.rs
Abhicodes-crypto: Since we don't have a type `ConnectorResult`  and the trait doesn't do anything with the Ok part . I think It makes sense to keep it as it is.

Narayanbhat166 (approved):

jarnura (approved):

PR #1387, opened 2023-06-08
feat(db): implement `MandateInterface` for `MockDb`

### LINKED ISSUE #1256: Implement `MandateInterface` for `MockDb`
Spin out from https://github.com/juspay/hyperswitch/issues/172. Please refer to that issue for more information.

nachiketkanore:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [x] Enhancement
- [x] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

- fixes #1256 

### Additional Changes

## Checklist


SanchithHegde (commented):

crates/router/src/db/mandate.rs
```diff
@@ -108,46 +112,150 @@ impl MandateInterface for Store {
...
+            mandate_type: mandate_new.mandate_type,
+            customer_accepted_at: mandate_new.customer_accepted_at,
+            customer_ip_address: mandate_new.customer_ip_address,
+            customer_user_agent: mandate_new.customer_user_agent,
+            network_transaction_id: mandate_new.network_transaction_id,
+            previous_attempt_id: mandate_new.previous_attempt_id,
+            created_at: mandate_new.created_at.unwrap_or_else(|| {
+                let now = OffsetDateTime::now_utc();
+                PrimitiveDateTime::new(now.date(), now.time())
+            }),
```
SanchithHegde, line 247: You can use the `common_utils::date_time::now()` funtion instead.

```suggestion
            created_at: mandate_new.created_at.unwrap_or_else(common_utils::date_time::now),
```

You might have to run the formatter though.

crates/router/src/db/mandate.rs
```diff
@@ -108,46 +111,149 @@ impl MandateInterface for Store {
...
+            .lock()
+            .await
+            .iter()
+            .find(|mandate| mandate.merchant_id == merchant_id && mandate.mandate_id == mandate_id)
+        {
+            Some(mandate) => Ok(mandate.clone()),
+            None => {
+                Err(errors::StorageError::ValueNotFound("mandate not found".to_string()).into())
+            }
+        }
```
NishantJoshi00, line 128: instead of match, can we go with a more functional approach, not a hard requirement, but as a improvement, we can consider using `.ok_or_else` after `.find`

crates/router/src/db/mandate.rs
```diff
@@ -108,46 +111,149 @@ impl MandateInterface for Store {
...
+                let storage_mandate_status: storage_models::enums::MandateStatus =
+                    mandate_status.foreign_into();
+                checker &= mandate.mandate_status == storage_mandate_status;
+            }
+            checker
+        });
+
+        let mandates: Vec<storage::Mandate> = if let Some(limit) = mandate_constraints.limit {
+            #[allow(clippy::as_conversions)]
+            mandates_iter.take(limit as usize).cloned().collect()
```
NishantJoshi00, line 217: can we have a bit of runtime safety here, consider this
limit is 
`Option<i64>` -> `i64`
to convert `i64` to `usize` we need to handle negative numbers
```
x if x < 0 => 0,
x => x as usize,
```

NishantJoshi00 (changes requested):
Looks good,

crates/router/src/db/mandate.rs
SanchithHegde: Should we consider using `usize::MAX` if `limit` is negative?

NishantJoshi00: if the limit is not passed, then we are displaying the same behaviour as `usize::MAX`, so I think in case of 0 we should return no entires

SanchithHegde (approved):

NishantJoshi00 (approved):

SanchithHegde:

PR #1378, opened 2023-06-07
save payment_method_type when doing a transaction

pixincreate:
## Type of Change

- [x] Bugfix

## Description

When saving the payment method record to the payment_method table after a payment, only the PaymentMethod is saved but not the PaymentMethodType even if it is present in the request.
This PR fixes that.

## Motivation and Context

I fixed a bug! 

## How did you test it?

Manually
Screenshot 2023-06-07 at 15 41 13
Screenshot 2023-06-07 at 15 41 32
Screenshot 2023-06-07 at 15 41 53
Screenshot 2023-06-07 at 15 43 15

## Checklist

- [x] I formatted the code `cargo +nightly fmt --all`
- [x] I addressed lints thrown by `cargo clippy`
- [x] I reviewed submitted code

PR #1415, opened 2023-06-12
fix(core): `payment_method_type` not set in the payment attempt when making a recurring mandate payment

pixincreate:
## Type of Change

- [x] Bugfix

## Description

When doing a recurring mandate payment, the `payment_method` is set in the newly created payment attempt but the `payment_method_type` is not set. This PR aims to fix that.

**[#1378](https://github.com/juspay/hyperswitch/pull/1378) HAS TO BE MERGED BEFORE THIS PR IS MERGED AS THE CURRENT PR IS DEPENDENT ON THAT!**
**WHEN CREATING A MANDATE, MAKE SURE YOU THAT YOU GIVE `payment_method_type` IN THE REQUEST.**

## Motivation and Context

I fixed another bug! 

## How did you test it?

Manually.

### IMAGES
image
image
image

## Checklist


crates/router/src/core/payments/operations/payment_complete_authorize.rs
```diff
@@ -110,10 +110,8 @@ impl<F: Send + Clone> GetTracker<F, PaymentData<F>, api::PaymentsRequest> for Co
 
         payment_attempt.payment_method = payment_method.or(payment_attempt.payment_method);
         payment_attempt.browser_info = browser_info;
-        payment_attempt.payment_method_type = request
```
jagan-jaya, line 113: why are we ignoring the payment_method_type from request?

jagan-jaya (dismissed):

crates/router/src/core/payments/operations/payment_confirm.rs
```diff
@@ -129,10 +129,8 @@ impl<F: Send + Clone> GetTracker<F, PaymentData<F>, api::PaymentsRequest> for Pa
 
         payment_attempt.payment_method = payment_method.or(payment_attempt.payment_method);
         payment_attempt.browser_info = browser_info;
-        payment_attempt.payment_method_type = request
```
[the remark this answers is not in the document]
pixincreate: When we try to get the `payment_method_type` from the request, it always return a `None` for recurring mandate as we'll not be using the `payment_method_type` that we get from the `get_token_pm_type_mandate_details()`.

crates/router/src/core/payments/operations/payment_complete_authorize.rs
pixincreate: When we try to get the `payment_method_type` from the request, it always return a `None` for recurring mandate as we'll not be using the `payment_method_type` that we get from the `get_token_pm_type_mandate_details()`.

vspecky (approved):

Narayanbhat166 (approved):

vspecky (approved):

vspecky (changes requested):
Actually, @jagan-jaya was correct in his previous two comments. We should also be taking the payment method type from the request. The precedence for taking the payment method type would be `mandate pmt OR request pmt OR original pmt from attempt`

vspecky (approved):
Ignore previous comment. Said comment is taken care of in the `helpers.rs` function

Narayanbhat166 (approved):

vspecky (approved):

PR #1447, opened 2023-06-15
refactor(core): accept customer data in customer object

### LINKED ISSUE #1446: [REFACTOR] accept customer details in customer object
Currently, for a payment, the `CustomerDetails` such as email, phone, name are accepted in the `PaymentsRequest` struct 
https://github.com/juspay/hyperswitch/blob/795500797d1061630b5ca493187a4e19d98d26c0/crates/api_models/src/payments.rs#L59

It is not clear that these fields are related to the customer. One may think that the fields are for a payment and one time use only. But these are stored in the `Customers` table.

A better api design would be to accept the customer details in a `customers` field in `PaymentsRequest`.

It would look something like this
```json
{
  "customer" : {
    "id": "cus_1235",
    "email" : "example@example.com"
  }
}
```

Narayanbhat166:
## Type of Change

- [x] Bugfix

## Description

This PR will support sending the customer details in the `customers` object in `PaymentsRequest` in all three endpoints 
- payments-create
- payments-update
- payments-confirm

The changes are added in a backwards compatible manner, the older fields will be deprecated soon.

### Additional Changes

Fixes #1446  

## How did you test it?

- Create a payment with customer details outside.
Screenshot 2023-06-15 at 3 28 27 PM

- Create a payment with customer details in customers field.
Screenshot 2023-06-15 at 3 27 19 PM

- Create ambiguity by passing the same field in both the places
Screenshot 2023-06-16 at 4 13 48 PM

## Checklist


crates/router/src/core/payments/helpers.rs
```diff
@@ -810,6 +810,30 @@ pub async fn get_customer_from_details<F: Clone>(
     }
 }
 
+/// Get the customer details from customer field if present
+/// or from the individual fields in `PaymentsRequest`
+pub fn get_customer_details_from_request(
```
Abhicodes-crypto, line 815: There was a similar case for order_details, backward compatibility was added. But there an error was thrown if data was present in both the fields. Should it be the same case here ?

Abhicodes-crypto (approved):
Minor comment. Other than LGTM

crates/router/src/core/payments/helpers.rs
Narayanbhat166: This error has been handled in the validate request.

jarnura (approved):

Abhicodes-crypto (approved):

PR #1449, opened 2023-06-15
ci(postman-collection-runner): use matrix to run tests

pixincreate:
## Type of Change

- [x] CI/CD

## Description

Uses `newman` to run postman collection tests on GitHub Actions for validating the integrity and working of the connectors. This helps better maintain code integrity and keep every developer in check.  

### Usage
- Add a new connector for testing in [this](https://github.com/juspay/hyperswitch/blob/main/.github/workflows/postman-collection-runner.yml) workflow
- It is recommended that you develop the postman collection by following the conventions that is already developed by @Gnanasundari24 for `Stripe` and `Adyen` instead of writing unit tests in `crates/router/tests` as the postman collection that you develop will be used for validation of the integrity
- For every postman collection developed, follow the below mentioned convention in the `postman` directory:
```
<connector_name>.postman_collection.json
```

## Motivation and Context

Current PR reduces the human interaction and helps speed up development as well as reduce the availability of bad code on the repo by improving the code quality.

## How did you test it?

It is running and will run in the actions every time a PR is pushed. As of now, it failing due to lack of collection in the repo.  
`hyperswitch.post_colleciton.json` has been failing and the reason is unknown. Have to debug.

## Checklist


scripts/postman_test_automation.sh
```diff
@@ -0,0 +1,51 @@
...
+API_KEY=""
+API_SECRET=""
+KEY1=""
+
+# Unused as of now, will be useful once we start using this properly
+CERTIFICATE=$4
+CERTIFICATE_KEY=$5
+
+path_generation() {
+    local NAME=$1
```
SanchithHegde, line 15: 1. Local variables can be `snake_case`. I personally feel it makes it easier to identify variables which have a global scope against local ones.
2. Always quote the usages of variables/arguments.

```suggestion
    local name="$1"
```

Optionally, you can also prefer to wrap them within `{}` like so:

```suggestion
    local name="${1}"
```

So that a variable substitution like `"$12"` do not unintentionally search for the 12th argument when you instead wanted to suffix a `2` to the value of the 1st argument: `"${1}2"`. You can read more about this.

scripts/postman_test_automation.sh
```diff
@@ -0,0 +1,51 @@
...
+get_api_keys "$CONNECTOR_NAME" > /dev/null
+COLLECTION_PATH=$(path_generation $CONNECTOR_NAME)
+
+if [[ "$KEY_TYPE" == "HeaderKey" ]]; then
+    newman run $COLLECTION_PATH --env-var admin_api_key=$ADMIN_API_KEY --env-var baseUrl=$BASE_URL --env-var connector_api_key=$API_KEY
+elif [[ "$KEY_TYPE" == "BodyKey" ]]; then
+    newman run $COLLECTION_PATH --env-var admin_api_key=$ADMIN_API_KEY --env-var baseUrl=$BASE_URL --env-var connector_api_key=$API_KEY --env-var connector_key1=$KEY1
+elif [[ "$KEY_TYPE" == "SignatureKey" ]]; then
+    newman run $COLLECTION_PATH --env-var admin_api_key=$ADMIN_API_KEY --env-var baseUrl=$BASE_URL --env-var connector_api_key=$API_KEY --env-var connector_api_secret=$API_SECRET --env-var connector_key1=$KEY1
+fi
```
SanchithHegde, line 51: Maybe use the `case` statement instead?

scripts/postman_test_automation.sh
SanchithHegde, line 1: 1. Quote all usages of variables which are strings (which I assume most of them are). Please make this change across the script.
2. Optionally, run it through `shellcheck`. It'll point out some common improvements to the script.
3. Preferably, add comments as to what a certain piece of the script does. Looking at the script alone does not provide enough information as to what file types are being parsed and what the input looks like, etc.

.github/workflows/postman-collection-runner.yml
```diff
@@ -1,6 +1,8 @@
 name: Postman Collection API test
 
 on:
+  pull_request:
```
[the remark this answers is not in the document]
pixincreate: I don't remember adding this. However, I would recommend keeping this with some extra modifications done to it. Like, it should only work for newly integrated connectors when the developer of that connector raises a PR for that. Thoughts?

I'll remove this if you think this will be useless.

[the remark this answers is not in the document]
SanchithHegde: I don't think it's easy to filter and run a workflow on newly added files in a pull request. Moreover, we'll end up spamming the connectors each time a pull request is updated with new commits, which wouldn't be preferable. Let's only keep the schedule trigger for now. If absolutely required, consider adding a merge queue trigger.

SanchithHegde (dismissed):

SanchithHegde (approved):

jarnura (approved):

PR #1466, opened 2023-06-17
refactor(payments): add udf field and remove refactor metadata

Abhicodes-crypto:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [x] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

In this PR , 
`routing_parameters` and `data`  are removed from metadata.
`udf` field was added in `PaymentsRequest`

Backwards compatibility for order_details in metadata is maintained but will be deprecated in the future.

### Additional Changes

## How did you test it?

I tested everything and it worked fine. 
image

## Checklist

### DIFF
--- a/crates/api_models/src/payments.rs
+++ b/crates/api_models/src/payments.rs
@@ -283,7 +283,7 @@ pub struct PaymentsRequest {
     pub manual_retry: bool,
 
     /// Any user defined fields can be passed here.
-    #[schema(value_type = Object, example = r#"{ "city": "NY", "unit": "245" }"#)]
+    #[schema(value_type = Option<Object>, example = r#"{ "udf1": "some-value", "udf2": "some-value" }"#)]
     pub udf: Option<pii::SecretSerdeValue>,
 }
 
@@ -1381,9 +1381,9 @@ pub struct PaymentsResponse {
 
     /// ephemeral_key for the customer_id mentioned
     pub ephemeral_key: Option<EphemeralKeyCreateResponse>,
-
+    
     /// Any user defined fields can be passed here.
-    #[schema(value_type = Object, example = r#"{ "city": "NY", "unit": "245" }"#)]
+    #[schema(value_type = Option<Object>, example = r#"{ "udf1": "some-value", "udf2": "some-value" }"#)]
     pub udf: Option<pii::SecretSerdeValue>,
 }
 
--- a/crates/api_models/src/payments.rs
+++ b/crates/api_models/src/payments.rs
@@ -1381,7 +1381,7 @@ pub struct PaymentsResponse {
 
     /// ephemeral_key for the customer_id mentioned
     pub ephemeral_key: Option<EphemeralKeyCreateResponse>,
-    
+
     /// Any user defined fields can be passed here.
     #[schema(value_type = Option<Object>, example = r#"{ "udf1": "some-value", "udf2": "some-value" }"#)]
     pub udf: Option<pii::SecretSerdeValue>,
--- a/crates/api_models/src/payments.rs
+++ b/crates/api_models/src/payments.rs
@@ -1,4 +1,4 @@
-use std::{collections::HashMap, num::NonZeroI64};
+use std::num::NonZeroI64;
 
 use cards::CardNumber;
 use common_utils::{
@@ -281,6 +281,10 @@ pub struct PaymentsRequest {
     /// If enabled payment can be retried from the client side until the payment is successful or payment expires or the attempts(configured by the merchant) for payment are exhausted.
     #[serde(default)]
     pub manual_retry: bool,
+
+    /// Any user defined fields can be passed here.
+    #[schema(value_type = Object, example = r#"{ "city": "NY", "unit": "245" }"#)]
+    pub udf: Option<pii::SecretSerdeValue>,
 }
 
 #[derive(Default, Debug, serde::Deserialize, serde::Serialize, Clone, Copy, PartialEq, Eq)]
@@ -1377,6 +1381,10 @@ pub struct PaymentsResponse {
 
     /// ephemeral_key for the customer_id mentioned
     pub ephemeral_key: Option<EphemeralKeyCreateResponse>,
+
+    /// Any user defined fields can be passed here.
+    #[schema(value_type = Object, example = r#"{ "city": "NY", "unit": "245" }"#)]
+    pub udf: Option<pii::SecretSerdeValue>,
 }
 
 #[derive(Clone, Debug, serde::Deserialize, ToSchema)]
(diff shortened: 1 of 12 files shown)


crates/api_models/src/payments.rs
```diff
@@ -1641,6 +1641,22 @@ pub struct RedirectResponse {
...
 }
 
+#[derive(
+    Default,
+    Debug,
+    serde::Deserialize,
+    serde::Serialize,
+    Clone,
+    ToSchema,
+    router_derive::PolymorphicSchema,
```
lsampras, line 1651: Do we need `PolymorphicSchema` here?

crates/api_models/src/payments.rs
```diff
@@ -239,16 +244,15 @@ pub struct PaymentsRequest {
...
-
     /// Business sub label for the payment
     pub business_sub_label: Option<String>,
 
     /// If enabled payment can be retried from the client side until the payment is successful or payment expires or the attempts(configured by the merchant) for payment are exhausted.
     #[serde(default)]
     pub manual_retry: bool,
+
+    /// metadata that will be deprecated. For now only order details can be passed here.
+    pub metadata: Option<MetadataExternal>,
```
lsampras, line 255: will need to add schema here?

probably can specify in the schema that this field is gonna be deprecated

crates/router/src/core/payments/operations/payment_confirm.rs
```diff
@@ -103,6 +103,8 @@ impl<F: Send + Clone> GetTracker<F, PaymentData<F>, api::PaymentsRequest> for Pa
             .map(ForeignInto::foreign_into)
             .or(payment_intent.setup_future_usage);
 
+        payment_intent.udf = payment_intent.udf.or(request.udf.clone());
```
lsampras, line 106: If the merchant hasn't specified udf when he created intent,
we shouldn't be picking one up from the confirm request as well right?

crates/api_models/src/payments.rs
```diff
@@ -1355,12 +1360,14 @@ pub struct PaymentsResponse {
...
-    pub allowed_payment_method_types: Option<Vec<api_enums::PaymentMethodType>>,
-
     /// ephemeral_key for the customer_id mentioned
     pub ephemeral_key: Option<EphemeralKeyCreateResponse>,
+
+    /// the metadata that will be deprecated
+    pub metadata: Option<MetadataExternal>,
+
+    /// user defined values
+    pub udf: Option<pii::SecretSerdeValue>,
```
lsampras, line 1370: Why are these fields not included in API schema

crates/storage_models/src/payment_intent.rs
```diff
@@ -122,6 +124,7 @@ pub enum PaymentIntentUpdate {
         business_label: Option<String>,
         order_details: Option<Vec<pii::SecretSerdeValue>>,
         metadata: Option<pii::SecretSerdeValue>,
+        udf: Option<pii::SecretSerdeValue>,
```
lsampras, line 127: we could remove this if udf is not intented to be updated...

lsampras (commented):
I don't understand the motivation behind the metada related change...

> From a users perspective there is no field metadata. He can populate fields inside metadata by adding the same fields in the payment request directly.

1. since we don't add metadata in the schema he doesn't know about it...
2. But why would he even add fields like `order_details` if they are also not present in our API schema...
3. If we do decide to make such fields public why not restrict them to using metadata...
4. In general we should keep the struct format & API request json similar in structure so that we can easily correlate the request when looking at logs...

Unless there's a compelling need to flatten the metadata I'd prefer it to remain the same...

crates/api_models/src/payments.rs
Abhicodes-crypto: `metadata` would be deprecated so we don't want it in API schema. Will add udf.

crates/router/src/core/payments/operations/payment_confirm.rs
Abhicodes-crypto: I feel the same But the task was to add udf to PaymentsRequest and whichever API's using it.

crates/api_models/src/payments.rs
```diff
@@ -184,7 +184,12 @@ pub struct PaymentsRequest {
     pub statement_descriptor_suffix: Option<String>,
 
     /// You can specify up to 50 keys, with key names up to 40 characters long and values up to 500 characters long. Metadata is useful for storing additional, structured information on an object.
-    pub metadata: Option<Metadata>,
+    #[serde(flatten)]
+    pub metadata_internal: Option<Metadata>,
```
jarnura, line 188: Follow the specs to name the fields and structs.

This field is metadata only.

crates/api_models/src/payments.rs
```diff
@@ -239,16 +244,15 @@ pub struct PaymentsRequest {
...
-    pub allowed_payment_method_types: Option<Vec<api_enums::PaymentMethodType>>,
-
     /// Business sub label for the payment
     pub business_sub_label: Option<String>,
 
     /// If enabled payment can be retried from the client side until the payment is successful or payment expires or the attempts(configured by the merchant) for payment are exhausted.
     #[serde(default)]
     pub manual_retry: bool,
+
+    /// metadata that will be deprecated. For now only order details can be passed here.
```
jarnura, line 254: Remove this comment and add the valid comment, this field is not going to deprecated.

For naming convention refer the spec sheet

jarnura (changes requested):

crates/storage_models/src/payment_intent.rs
Abhicodes-crypto: It is being updated in payment update and confirm call

crates/api_models/src/payments.rs
```diff
@@ -281,6 +281,10 @@ pub struct PaymentsRequest {
     /// If enabled payment can be retried from the client side until the payment is successful or payment expires or the attempts(configured by the merchant) for payment are exhausted.
     #[serde(default)]
     pub manual_retry: bool,
+
+    /// Any user defined fields can be passed here.
+    #[schema(value_type = Object, example = r#"{ "city": "NY", "unit": "245" }"#)]
+    pub udf: Option<pii::SecretSerdeValue>,
```
Narayanbhat166, line 287:
```suggestion
    #[schema(value_type = Option<Object>, example = r#"{ "city": "NY", "unit": "245" }"#)]
    pub udf: Option<pii::SecretSerdeValue>,
```

crates/api_models/src/payments.rs
```diff
@@ -1377,6 +1381,10 @@ pub struct PaymentsResponse {
 
     /// ephemeral_key for the customer_id mentioned
     pub ephemeral_key: Option<EphemeralKeyCreateResponse>,
+
+    /// Any user defined fields can be passed here.
+    #[schema(value_type = Object, example = r#"{ "city": "NY", "unit": "245" }"#)]
+    pub udf: Option<pii::SecretSerdeValue>,
```
Narayanbhat166, line 1387:
```suggestion
    #[schema(value_type = Option<Object>, example = r#"{ "city": "NY", "unit": "245" }"#)]
    pub udf: Option<pii::SecretSerdeValue>,
```

crates/api_models/src/payments.rs
```diff
@@ -1636,12 +1644,7 @@ pub struct OrderDetails {
 pub struct Metadata {
     /// Information about the product and quantity for specific connectors. (e.g. Klarna)
     pub order_details: Option<OrderDetails>,
-    /// Information used for routing
-    pub routing_parameters: Option<HashMap<String, String>>,
```
Narayanbhat166, line 1640: Why was this field removed, is it no longer required? If it is required, then how are we accepting it?

Abhicodes-crypto: Yeah I have confirmed with @bernard-eugine and @vspecky . It can be removed.

Narayanbhat166 (approved):

jarnura (approved):

Issue #888, opened 2023-04-14
[FEATURE] separate payments_session from payments core


Narayanbhat166:
### Feature Description

In the current implementation, the [payments_session](https://github.com/juspay/hyperswitch/blob/main/crates/router/src/core/payments/operations/payment_session.rs) is a part of [payments core](https://github.com/juspay/hyperswitch/blob/main/crates/router/src/core/payments.rs) logic. This was done to reuse certain parts of code which would avoid redundancy, this would also be maintained along with the payments core. The major anamoly from the core logic would be that multiple connectors would be called in parallel.

### Possible Implementation

As we have come a long way in our journey of integrating the connectors / gateways, we have not come across any connector / gateway, where the session token generation ( required to initiate the sdk ) logic, is tied to a payment flow. It would be better to have this logic separated from the core logic. This would help us in
- Deploying it separately since this will have many reads 
- Flexibility to add changes without any breaking changes to the payments logic.
- No routing support is required.

This can be implemented as a separate function as which would perform the required functionalities. This can be build in a similar way to [payment_method_list](https://github.com/juspay/hyperswitch/blob/main/crates/router/src/core/payment_methods/cards.rs#L784) api. Additional step would be to call the connector whenever necessary.

### Have you spent some time to check if this feature request has been raised before?

- [X] I checked and didn't find similar issue

### Have you read the Contributing Guidelines?

- [X] I have read the [Contributing Guidelines](https://github.com/juspay/hyperswitch/blob/main/docs/CONTRIBUTING.md)

### Are you willing to submit a PR?

No, but I'm happy to collaborate on a PR with someone else

shubhanshu-mani:
Hi I would like to pick this up , need some clarifications though

1. Is payment_session an API ? 
2. Do you just want to move the existing implementation to another module ? 
3. Is payment_session's implementation is based on payment flow and you want the impl to be independent of it ??

Narayanbhat166:
Hi @shubhanshu-mani. Thanks for showing interest. To clarify your doubts
- Yes `payments_session` is an api endpoint currently. It can be followed from request till response from [here](https://github.com/juspay/hyperswitch/blob/8eac8cb3709ccab8819aa25c34b64b01a6d43750/crates/router/src/routes/app.rs#L90)
[...]

PR #1472, opened 2023-06-18
refactor(core): move update trackers after build request

### LINKED ISSUE #1471: [BUG] payment status is processing if there is a validation error
### Bug Description

The payment flow is as follows
- Merchant server creates the Payments with amount and currency and other merchant specific fields.
- The SDK confirms the payment with `PaymentMethodData`.

During confirm, the connector / gateway is called to create the actual transaction, we update the payment status before this, i.e make the payment status as `processing` since after we get the response from connector it will be updated to a terminal status ( happy case ). If for some reason, the connector request cannot be created, then the payment status will still be processing which is not the intended behaviour.

### Expected Behavior

If there is a bad request error, then the trackers are updated and payment status is changed to processing.

### Actual Behavior

The payment status should not be changed.

### Steps To Reproduce

Provide an unambiguous set of steps to reproduce this bug. Include code or configuration to reproduce, if relevant.
1. Create a `zen` payment, without order details.
2. Confirm with payment method data - fails saying that `order_details` is a required field.
3. Retrieve the payment, the status will be `processing`

### Context For The Bug

_No response_

### Environment

Local environment

Yes I am willing to submit a PR!

Narayanbhat166:
## Type of Change

- [x] Refactoring

## Description

Update trackers has been moved before callling the connector. This is to make sure that if there is any validation error, then the payment status will not be updated. 
For other flows where connector does not need to be called ( except `session_token` flow, there is an issue to move it outside payment core #888), it will be called as normal.

## Motivation and Context

Fixes #1471 

## How did you test it?

- Create a payment with zen, without email - Payment status is `requires_payment_method`.
- Confirm the payment - error is Missing required param: email
- Retrieve the payment - status will be `requires_payment_method`.

## Checklist


crates/router/src/core/payments.rs
```diff
@@ -189,11 +178,24 @@ where
...
             }
         };
+
         if should_delete_pm_from_locker(payment_data.payment_intent.status) {
             vault::Vault::delete_locker_payment_method_by_lookup_key(state, &payment_data.token)
                 .await
         }
+    } else {
+        operation
+            .to_update_tracker()?
```
Abhicodes-crypto, line 188: In PR description it is mentioned that update trackers would be called before calling the connector . But here it is called after calling the connector .

Narayanbhat166: This is a normal flow where connector does not need to be called, this is the else branch ( when we don't have connector). These include flows like `payments_update`,  so we just go with the normal flow.

Abhicodes-crypto (dismissed):

crates/router/src/core/payments.rs
```diff
@@ -155,11 +143,12 @@ where
                     state,
                     &merchant_account,
                     connector,
-                    &operation,
+                    operation.to_update_tracker()?,
```
jarnura, line 146: send whole operation itself. call_connector_service is not just specific to just one operation.

crates/router/src/core/payments/flows/authorize_flow.rs
```diff
@@ -155,36 +167,24 @@ impl types::PaymentsAuthorizeRouterData {
                 );
 
                 logger::debug!(completed_pre_tasks=?true);
+
                 if self.should_proceed_with_authorize() {
                     self.decide_authentication_type();
                     logger::debug!(auth_type=?self.auth_type);
-                    let resp = services::execute_connector_processing_step(
```
jarnura, line 161: this block is a conditional block for `self.should_proceed_with_authorize()` condition, in decide flow the same is established?

jarnura (changes requested):

jagan-jaya (changes requested):

crates/router/src/core/payments/flows/authorize_flow.rs
```diff
@@ -51,24 +51,46 @@ impl Feature<api::Authorize, types::PaymentsAuthorizeData> for types::PaymentsAu
...
-        let resp = self
-            .decide_flow(
+        let connector_integration: services::BoxedConnectorIntegration<
+            '_,
+            api::Authorize,
+            types::PaymentsAuthorizeData,
+            types::PaymentsResponseData,
+        > = connector.connector.get_connector_integration();
+
+        if self.should_proceed_with_authorize() {
```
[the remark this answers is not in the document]
Narayanbhat166: The first check is to construct the connector request. Here the actual call to connector happens, so it is added again here

jagan-jaya (approved):

crates/router/src/core/payments/flows/authorize_flow.rs
Narayanbhat166: Yes. I will refactor the method in another PR to decide to call connector or not based only on whether the connector request is passed. I did not take that up in this PR as it would be a big refactor and add risk

jarnura (approved):

pixincreate:
I think this PR broke `payment_methods`. #1378 got conflicts when this PR got merged. I fixed the conflicts and while testing I observed list payment_methods for customer is broken.  
I passed customer_id to the URL ({{baseUrl}}/customers/:customer_id/payment_methods) and the response is always empty.  
```
{
    "customer_payment_methods": []
}
```

cURL:
curl --location 'http://localhost:8080/customers/StripeCustomer/payment_methods' --header 'Accept: application/json' --header 'api-key: api_key_comes_here'

PR #1496, opened 2023-06-21
fix(update_trackers): handle preprocessing steps status update

### LINKED ISSUE #1495: [BUG] : Bank Transfer status mapping is wrong for stripe
### Bug Description

While making the payment for bank transfer its going to "processing" state instead of "requires_customer_action".
Even after payment sync also its in processing state.

### Expected Behavior

It should be in "requires_customer_action" state

### Actual Behavior

Its going to "processing" state

### Steps To Reproduce

1.Create a payment intent
2.confirm the payment with bank transfer payment method details
3.The status is going to processing instead of requires_customer_action.

### Context For The Bug

_No response_

### Environment

Integ

None

Narayanbhat166:
## Type of Change

- [x] Bugfix

## Description

The status of `router_data` was being updated after update trackers to `processing`. But in case of Bank transfers, the status of payment is updated in pre processing steps, so we need to persist the routerdata instead of overriding it.

## Motivation and Context

Fixes #1495.

## How did you test it?

- Test stripe card payment
- Test ach bank transfers.
- Test shift4 3ds
- Test applepay trustpay

## Checklist


crates/router/src/connector/trustpay/transformers.rs
```diff
@@ -950,6 +950,7 @@ impl<F, T>
                     },
                 ))),
             }),
+            status: storage_models::enums::AttemptStatus::Pending,
```
jagan-jaya, line 953: by default also it is Pending right?

jagan-jaya (changes requested):

crates/router/src/connector/trustpay/transformers.rs
Narayanbhat166: It's better to be explicit than rely on the caller. This flow defies the normal conventions that we follow.

jarnura (approved):

PR #1499, opened 2023-06-21
fix(ci/cd): refactor newman command to fix the runner

pixincreate:
## Type of Change

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [x] CI/CD

## Description

This PR aims at resolving the bug that forced the runner to fail always especially at the scheduled times (6 AM and 6 PM).  
The Action was failing due to the params being considered as external arguments (`newman run "${COLLECTION_PATH}" "${args[@]}"`) instead of considering the params as interal / direct arguments (newman run postman/collection.json --env-var api=api_key --env-var ...). 

## Motivation and Context

It is not working, trying to fix.

## How did you test it?

Manually tested the runner by creating a personal private repo and tried running the tests there fixing fixing the bugs until the failure rate drastically reduced from 121 to merely 34. This must be due to the locker being down.  

## Checklist


jarnura (approved):

.github/workflows/postman-collection-runner.yml
```diff
@@ -39,12 +46,13 @@ jobs:
...
-
-          # Stripe specific env variables, will be replaced by connector_auth.toml
           GATEWAY_MERCHANT_ID: ${{ secrets.STRIPE_GATEWAY_MERCHANT_ID }}
           GPAY_CERTIFICATE: ${{ secrets.STRIPE_GPAY_CERTIFICATE }}
           GPAY_CERTIFICATE_KEYS: ${{ secrets.STRIPE_GPAY_CERTIFICATE_KEYS }}
-        shell: bash
-        run: ./scripts/postman_test_automation.sh ${{ matrix.connector }}
+        uses: nick-fields/retry@v2
+        with:
+          timeout_minutes: 6
```
SanchithHegde, line 55: How is the timeout decided?

scripts/postman_test_automation.sh
```diff
@@ -19,16 +19,23 @@ get_api_keys() {
     # [<connector_name>]
     # api_key = "HeadKey of <connector_name>"
 
-    API_KEY=$(echo "${result}" | awk -F ' = ' '$1 == "api_key" { print $2 }')
-    KEY1=$(echo "${result}" | awk -F ' = ' '$1 == "key1" { print $2 }')
-    API_SECRET=$(echo "${result}" | awk -F ' = ' '$1 == "api_secret" { print $2 }')
-
+    # Keys are exported as environment variables since some API Keys for connectors such as ACI
+    # base Base64 based and requires "Bearer" to be prefexed such as "Bearer Skst45645gey5r#&$==".
```
SanchithHegde, line 23: Is there an error around this part?

> [...] such as ACI base Base64 based and requires "Bearer" to be prefexed [...]

Also, there's a typo:

```suggestion
    # Keys are exported as environment variables since some API Keys for connectors such as ACI
    # base Base64 based and requires "Bearer" to be prefixed such as "Bearer Skst45645gey5r#&$==".
```

scripts/postman_test_automation.sh
```diff
@@ -38,50 +45,19 @@ get_api_keys() {
...
-        [[ -n "$GPAY_CERTIFICATE" ]] && args+=("--env-var" "certificate=${GPAY_CERTIFICATE}")
-        [[ -n "$GPAY_CERTIFICATE_KEYS" ]] && args+=("--env-var" "certificate_keys=${GPAY_CERTIFICATE_KEYS}")
-        newman run "${COLLECTION_PATH}" "${args[@]}"
-        ;;
-esac
+# Newman runner
+newman run "${COLLECTION_PATH}" \
+    --env-var "admin_api_key=${ADMIN_API_KEY}" \
+    --env-var "baseUrl=${BASE_URL}" \
+    --env-var "connector_api_key=${API_KEY}" \
```
SanchithHegde, line 56: Instead of `export`ing these as environment variables, can you use single quotes around these values? Would that help prevent the shell from reading values with spaces as separate arguments?

```suggestion
    --env-var "admin_api_key='${ADMIN_API_KEY}'" \
    --env-var "baseUrl='${BASE_URL}'" \
    --env-var "connector_api_key='${API_KEY}'" \
```

Same change for the following lines?

Also, just being sure, would these API keys be printed in the logs, either when the script is run or by `newman`?

.github/workflows/postman-collection-runner.yml
pixincreate: I had updated that in the description today afternoon. Pasting the same here:
> Retries have been added since collections sometimes fail for no reason and this believed to be inconsistencies with matrix itself. Retrying usually fixes unless the error is specific to the collection. It takes 1 - 2 minutes for a collection to run on average. Since, they all run in parallel, timeout of 3 minutes max would be enough. However, maximum number of retries for any collection to run is set to3. So, timeout of 6 minutes have been set as of now. It can be scaled depending on the necessity.

SanchithHegde: Alright. Resolving this conversation.

scripts/postman_test_automation.sh
pixincreate: This cannot be done. Using single quotes directly considers the API_KEYS along with single quotes like `'sdkjfdskdgd'` instead of `sdkjfdskdgd`. 
The same behaviour can be observed when trying to print the contents of a variable as well. If you do `echo '${something}'`, it will print `${something}` instead of the contents within it. However, when using it as a command, it prints the contents along with single quotes.  
Unfortunately, when you run Payment Connector - Create, yes, API keys do get printed. Thanks for reminding that, I feel like it is better to remove logs all over for safety and use it whenever needed locally.

scripts/postman_test_automation.sh
pixincreate: Auto fill issue.

scripts/postman_test_automation.sh
SanchithHegde: > This cannot be done. Using single quotes directly considers the API_KEYS along with single quotes like `'sdkjfdskdgd'` instead of `sdkjfdskdgd`.

> The same behaviour can be observed when trying to print the contents of a variable as well. If you do `echo '${something}'`, it will print `${something}` instead of the contents within it. However, when using it as a command, it prints the contents along with single quotes.

These two are independent things: the first one is about how `newman` parses environment variables, the second is common shell behavior.

Can you do without `export`ing them? As far as I understand, since you aren't using `local`, the variables should be accessible anywhere within the script, but not outside it. Something like this: `MY_VAR=value`?

scripts/postman_test_automation.sh
```diff
@@ -19,16 +19,23 @@ get_api_keys() {
     # [<connector_name>]
     # api_key = "HeadKey of <connector_name>"
 
-    API_KEY=$(echo "${result}" | awk -F ' = ' '$1 == "api_key" { print $2 }')
-    KEY1=$(echo "${result}" | awk -F ' = ' '$1 == "key1" { print $2 }')
-    API_SECRET=$(echo "${result}" | awk -F ' = ' '$1 == "api_secret" { print $2 }')
-
+    # Keys are exported as environment variables since some API Keys for connectors such as ACI
+    # which is Base64 based and requires "Bearer" to be prefixed such as "Bearer Skst45645gey5r#&$==".
```
SanchithHegde, line 23:
```suggestion
    # Keys are set as variables since some API Keys for connectors such as ACI
    # are Base64 encoded and require "Bearer" to be prefixed such as "Bearer Skst45645gey5r#&$==".
```

Make changes as you see fit.

SanchithHegde (commented):

SanchithHegde (approved):

SanchithHegde (approved):

jarnura (approved):

Issue #1512, opened 2023-06-22
[BUG] Validate and update 4xx in Files - Delete endpoint


kritikmodi:
### Bug Description

The following error is being thrown with a 400 error code when we're hitting the Files - Delete endpoint : 

`"error": {
        "type": "invalid_request",
        "message": "Payment method type not supported",
        "code": "HE_03",
        "reason": "Not Supported if provider is not Router"
    }`

### Expected Behavior

The error should make more sense. The message should be updated accordingly.

### Actual Behavior

It throws a Bad Request Error (400) with an irrelevant message.

### Steps To Reproduce

Hit the Files - Delete endpoint.

### Environment

Local

### Have you spent some time to check if this bug has been raised before?

- [X] I checked and didn't find similar issue

### Have you read the Contributing Guidelines?

- [X] I have read the [Contributing Guidelines](https://github.com/juspay/hyperswitch/blob/main/docs/CONTRIBUTING.md)

### Are you willing to submit a PR?

Yes I am willing to submit a PR!

PR #1531, opened 2023-06-23
fix(disputes): update 4xx error for Files - Delete endpoint

kritikmodi:
## Type of Change

- [x] Bugfix

## Description

Introduced new error `FileProviderNotSupported` and fixed the error mapping for the `Files - Delete` endpoint thrown when the file storage is not `router`.

## Motivation and Context

This PR solves [this](https://github.com/juspay/hyperswitch/issues/1512) issue.

## How did you test it?

Tested manually using Postman.

## Checklist


crates/router/src/compatibility/stripe/errors.rs
```diff
@@ -199,6 +199,8 @@ pub enum StripeErrorCode {
     FileNotFound,
     #[error(error_type = StripeErrorType::HyperswitchError, code = "", message = "File not available")]
     FileNotAvailable,
+    #[error(error_type = StripeErrorType::InvalidRequestError, code = "IR_23", message = "Not Supported because provider is not Router")]
```
SanchithHegde, line 202: I don't think we should be sending our error codes via the Stripe compatibility layer.

@Abhicodes-crypto Please confirm.

lsampras: I believe the ideal solution for such cases should be non-centralized API errors which we had taken up earlier...

from a business perspective we should have valid stripe codes for all errors thrown by payments, refunds & customers API

But since we use the same enum for multiple other API endpoints... we can't guarantee that every payment/refund/customer API has a corresponding stripe error code.
At the same time non-stripe compatible API's also need to be mapped to a stripe compatible error (although it's never gonna be used)

kritikmodi: @SanchithHegde I've removed the error code from the Stripe error stack.

SanchithHegde (approved):

SanchithHegde (approved):

inventvenkat (approved):

Issue #1523, opened 2023-06-23
[BUG] "reason" field not being updated while hitting the Refunds - Update endpoint


kritikmodi:
### Bug Description

Upon hitting the `Refunds - Update` endpoint, we receive a 200 OK response but the `reason` field is not being updated.

### Expected Behavior

The `reason` field should be updated as per the specified message.

### Steps To Reproduce

Hit the `Refunds - Update` endpoint and pass a different reason message than the previous one.

### Environment

Local

### Have you spent some time to check if this bug has been raised before?

- [X] I checked and didn't find similar issue

### Have you read the Contributing Guidelines?

- [X] I have read the [Contributing Guidelines](https://github.com/juspay/hyperswitch/blob/main/docs/CONTRIBUTING.md)

### Are you willing to submit a PR?

Yes I am willing to submit a PR!

PR #1544, opened 2023-06-26
fix(refunds): modify refund fields to process updating of refund_reason

kritikmodi:
## Type of Change

- [x] Bugfix

## Description

The `reason` field is incorrectly mapped in the core logic so this PR modifies that. Also, the `refund_reason` field is missing in the db update function so this PR adds that.

## Motivation and Context

This PR fixes [this](https://github.com/juspay/hyperswitch/issues/1523) issue.

## How did you test it?

Tested manually using Postman.

## Checklist


crates/api_models/src/refunds.rs
```diff
@@ -76,7 +76,7 @@ pub struct RefundsRetrieveRequest {
 #[serde(deny_unknown_fields)]
 pub struct RefundUpdateRequest {
     /// An arbitrary string attached to the object. Often useful for displaying to users and your customer support executive
-    #[schema(max_length = 255, example = "Customer returned the product")]
+    #[schema(value_type = Option<String>, max_length = 255, example = "Customer returned the product")]
```
dracarys18, line 79: This is not necessary.

dracarys18 (changes requested):

crates/api_models/src/refunds.rs
kritikmodi: Okay

NishantJoshi00 (approved):

inventvenkat (approved):

dracarys18 (approved):

SanchithHegde (approved):

Issue #1439, opened 2023-06-14
[BUG] Payments list doesn't work with pagination parameters


lsampras:
### Context
- Currently the API accepts payment_id as pagination parameters for e.g starting_after = `pay_Ajabdhakjhah`
- However the underlying code relies on the auto-increment characteristic of databases
- While we do some sort of conversion here our pagination filters are based on `id` whereas the sort is `modified_at`
- This causes the response list to be not actually be paginated but return inconsistent & wrong results

#### Possible Solutions

- Either we can update the pagination filters to use `modified_at` (which can cause inconsistencies with data lists since `modified_at` can be changed)
- We can also try to sort out the list by incremental id

Abhicodes-crypto:
It was sorted by `id` before it is changed in this PR - https://github.com/juspay/hyperswitch/pull/317 
The same thing has been repeated for disputes and refunds as well.

NavinShrinivas:
Hey @lsampras I'd like to work on this, can I be assigned to it?

lsampras:
Ideally we would want it sorted by the `modified_at` column in payment_intent.

Also we use `futures::stream::iter` when getting the associated [payment_attempts](https://github.com/juspay/hyperswitch/blob/2b71d4d8c40c3697e902398fc76bc1256d5b25ee/crates/router/src/core/payments.rs#L1154),
If possible you can try to make this concurrent as well for extra challenge/brownie points

NavinShrinivas:
Just as a clarification, we want the payments_list to be sorted by create_at in DESC ordered as possibly the front-end pagination depends on it being sorted that way?

NavinShrinivas:
Update : I have some code waiting for this issue, it implements async db fetches and uses a map, this has some considerable changes and would like to test before I open a PR.

NavinShrinivas:
![image](https://github.com/juspay/hyperswitch/assets/42774281/c343f09f-2679-483f-aa72-7e4b69e768fa)
This is despite havbing created an API key and it beign stored to variables, any idea?

NavinShrinivas:
@lsampras if the above commit seems fine, I'll open a p.  I've tested in local and it returns payment lists from most recent to oldest payments.

SanchithHegde:
Closing as completed in #1556.

PR #1556, opened 2023-06-28
fix(router): desc payment list for pagination

NavinShrinivas:
## Type of Change

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description
Sorted the payment list order in API response for descending based on `created_at` parameter. It allows the database fetches to be asycn and is collected in an hashmap and collected later in a descending order.

### Additional Changes

## Motivation and Context
fixes issue #1439 

## How did you test it?

There were no new tests written (Although if needed we can write 1 test to check descending order in postman). But the endpoints behaviour was tested using postman.

## Checklist

- [ ] I added a [CHANGELOG](/CHANGELOG.md) entry if applicable (Im not sure if this is needed)


crates/router/src/core/payments.rs
```diff
@@ -1177,20 +1184,24 @@ pub async fn list_payments(
...
+    let mut payment_intents =
         helpers::filter_by_constraints(db, &constraints, merchant_id, merchant.storage_scheme)
             .await
             .to_not_found_response(errors::ApiErrorResponse::PaymentNotFound)?;
-
-    let pi = futures::stream::iter(payment_intents)
-        .filter_map(|pi| async {
-            let pa = db
+    payment_intents.sort();
+    payment_intents.reverse();
```
lsampras, line 1196: if possible we'd prefer to sort this via order clause in the diesel query...

That way the pagination would also be more user friendly

crates/router/src/core/payments.rs
```diff
@@ -1177,20 +1184,24 @@ pub async fn list_payments(
...
+
+    let payment_intent_attempt_map: Arc<RwLock<HashMap<&str, Vec<storage::PaymentAttempt>>>> =
+        Arc::new(RwLock::new(HashMap::new()));
```
lsampras, line 1199: I don't think interior mutability is required here...

you could do something like
```rust
// psuedo code
let pi_pa_mapping = payment_intents.iter().map(|pi| async {
     let payment_attempt = get_attempt_from_intent(pi);
     (pi.payment_id, payment_attempt)
}).collect::<HashMap<String, PaymentAttempt>>;
let sorted_payment_attempts = payment_intents.iter().map(|pi| pi_pa_mapping.pop(pi.payment_id).collect::<Vec<PaymentAttempt>>;
```

While this isn't a performance sensitive part, locks generally have a considerable overhead,
& I think doing it without locks might lead to prettier code as well

lsampras (commented):
Hey,
thanks for the PR,

I think we can do away with the locks lmk what you think

crates/router/src/core/payments.rs
NavinShrinivas: Ah yes! That looks much better. I did not know HashMaps could be collected (TIL). Ill look into sorting payment intents in diesel queries itself.

NavinShrinivas: A single payment intent can have many payment attempts right? If I were to collect a hashmap of <&str, PaymentAttempt> Id be overwriting and only storing last payment attempt.

dracarys18: since you'll be fetching via `active_attempt_id` which is like a primary key on attempt id, you'll be fetching only 1 attempt per payment intent (which is the current active one)

NavinShrinivas:
```rust
    let collected_futures = payment_intents.iter().map(|pi| async {
        match db
            .find_payment_attempt_by_payment_id_merchant_id_attempt_id(
                &pi.payment_id,
                merchant_id,
                &pi.active_attempt_id,
                // since OLAP doesn't have KV. Force to get the data from PSQL.
                storage_enums::MerchantStorageScheme::PostgresOnly,
            )
            .await
        {
            Ok(pa) => return (pi.payment_id, pa),
            Err(e) => {
                logger::warn!("Error getting payment_attempts using payment_id");
                return ("".to_string(), storage::PaymentAttempt::default());
            }
        };
    });
```

This is somewhat I've been writing the  async function inside the map, I really don't like what I've in the Err arm of the match function. What would your recommendation be? (In the above case I'd have to ignore all keys that are empty and moreover there is not default impl for PaymentAttempts)

lsampras:
It seems that as per our current implementation we used to throw an error if we did not have an `active_attempt_id` for an intent...
@Narayanbhat166 Is this a valid assumption that an intent will always have a corresponding attempt (is it an internal server error if we have inconsistency/missing attempts here)...

I think the desired flow here would be to log out any errors & still list out the available payment_attempts...

In this case you could do something like
```
    let collected_futures = payment_intents.iter().filter_map(|pi| async {
        match db
            .find_payment_attempt_by_payment_id_merchant_id_attempt_id(
                &pi.payment_id,
                merchant_id,
                &pi.active_attempt_id,
                // since OLAP doesn't have KV. Force to get the data from PSQL.
                storage_enums::MerchantStorageScheme::PostgresOnly,
            )
            .await {
                Ok(pa) => Some(Ok(pi.payment_id, pa)),
                Err(er) if er.current_context().is_db_not_found() => {
                    logger::error!(attempt_id=pi.active_attempt_id, "Error getting payment_attempts using payment_id");
                    None
                },
               er@Err(_) => Some(er)
            }
    }).collect::<StorageResult<HashMap<String, PaymentAttempt>>>?;
```

such that we throw errors for actual DB errors (connection timeout etc), log & ignore not found errors

filter_map will automatically drop the none values & since we are collecting a result of HashMap, the iteration should stop at the first encountered error

crates/router/src/core/payments.rs
```diff
@@ -1211,55 +1198,35 @@ pub async fn list_payments(
...
+        .flatten()
+        .collect::<HashMap<&str, storage::PaymentAttempt>>();
+
+    //Iterating over sorted Intents and sorting Attempts
+    let pi_pa_tuple_vec: Vec<(storage::PaymentIntent, storage::PaymentAttempt)> = payment_intents
+        .iter()
+        .filter_map(|v| {
+            //usinf for_each to get ownership of values
+            match pi_pa_map.remove(v.payment_id.as_str()) {
+                Some(pa) => Some((v.clone(), pa)),
```
NavinShrinivas, line 1223: I still this clone should be avoided, that was the entire reason behind using iter and maps for this operation, what do you think? @lsampras

crates/router/src/types/storage/payment_intent.rs
```diff
@@ -40,7 +40,7 @@ impl PaymentIntentDbExt for PaymentIntent {
         // when https://github.com/rust-lang/rust/issues/52662 becomes stable
         let mut filter = <Self as HasTable>::table()
             .filter(dsl::merchant_id.eq(merchant_id.to_owned()))
-            .order(dsl::modified_at.desc())
+            .order(dsl::created_at.desc())
```
NavinShrinivas, line 43: The intent were sorting based on modified_at before, I'm not sure if changing this will cause issues elsewhere (I have tested in postman in local and payments list is working as intended)

NavinShrinivas:
As for the failing clippy tests, I think the current way the code is better, clippys suggestions feel cryptic to me

crates/router/src/types/storage/payment_intent.rs
lsampras: 👍 Yeah this should be fine...

crates/router/src/core/payments.rs
lsampras: Cool,
can convert this `.iter()` to `.into_iter()` as well
Don't see usages of payment_intents after this

crates/router/src/core/payments.rs
```diff
@@ -1199,13 +1197,40 @@ pub async fn list_payments(
...
+        .await
+        .into_iter()
+        .flatten()
+        .collect::<HashMap<&str, storage::PaymentAttempt>>();
+
+    //Iterating over sorted Intents and sorting Attempts
+    let pi_pa_tuple_vec: Vec<(storage::PaymentIntent, storage::PaymentAttempt)> = payment_intents
+        .iter()
+        .filter_map(|v| {
+            //usinf for_each to get ownership of values
```
lsampras, line 1221: typo?

crates/router/src/core/payments.rs
```diff
@@ -1199,13 +1197,40 @@ pub async fn list_payments(
...
                 )
                 .await
-                .ok()?;
-            Some((pi, pa))
+            {
+                Ok(pa) => return Some((pi.payment_id.as_str(), pa)),
+                Err(e) => {
+                    logger::warn!("Error getting payment_attempts using payment_id : {}", e);
+                    None
+                }
```
lsampras, line 1205: This will also end up supressing actual errors related to database

crates/router/src/core/payments.rs
lsampras: Also just realized that since we are calling collect on the futures iterator It automatically is converted to [FuturesUnordered](https://rust-lang-nursery.github.io/futures-api-docs/0.1.27/futures/stream/struct.FuturesUnordered.html)...

We might not need to worry about ordering since collecting the futures should lead to ordered results (as compared to calling poll on FuturesUnordered which can return a random result)
So in the initial case instead of returning `(payment_id, attempt_id)` & then sorting them out we can just return a vec of `(PaymentIntent, PaymentAttempt)` and call it a day.... (might need to verify this tho)

crates/router/src/core/payments.rs
NavinShrinivas: yeh :sweat_smile:

crates/router/src/core/payments.rs
```diff
@@ -1199,13 +1197,40 @@ pub async fn list_payments(
...
+            .await
+            .into_iter()
+            .flatten() //Will ignore `None`, will only flatten 1 level
+            .collect::<Result<Vec<(storage::PaymentIntent, storage::PaymentAttempt)>, _>>();
+    //Will collect responses in same order async, leading to sorted responses
+
+    //Converting Intent-Attempt array to Response if no error
+    let data: Vec<api::PaymentsResponse> = match pi_pa_tuple_vec {
+        Ok(vector) => vector.into_iter().map(ForeignFrom::foreign_from).collect(),
+        Err(e) => {
```
NavinShrinivas, line 1229: We are pulling out the error here now, I think putting it as internal server is right here.

lsampras: Hey,
we could use change_context here...

e.g 
```
let data: Vec<api::PaymentsResponse>  = pi_pa_tuple_vec
    .change_context(errors::ApiErrorResponse::InternalServerError)?
    .into_iter()
    .map(ForeignFrom::foreign_from)
    .collect()
```

Narayanbhat166:
>@Narayanbhat166 Is this a valid assumption that an intent will always have a corresponding attempt (is it an internal server error if we have inconsistency/missing attempts here)

Yes an intent will always have an active attempt id. It is an InternalServerError if attempt id is None

NavinShrinivas:
> > @Narayanbhat166 Is this a valid assumption that an intent will always have a corresponding attempt (is it an internal server error if we have inconsistency/missing attempts here)
> 
> Yes @lsampras, an intent will always have an active attempt id. It is an InternalServerError if attempt id is None

I think this is the behaviour we have now in the latest commit

lsampras (approved):

SanchithHegde (approved):

lsampras (approved):

SanchithHegde:
Thanks for the PR, @NavinShrinivas!

PR #1582, opened 2023-07-01
ci(validate-openapi-spec): fail check if OpenAPI spec is not up-to-date

SanchithHegde:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [x] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [x] CI/CD

## Description

This PR updates the OpenAPI spec validation check workflow to fail the check if the OpenAPI spec is not-up-to-date. I tried to have it auto-commit the updated spec, but it stalls checks from running again, due to GitHub limitations. A solution to that is to use a custom PAT for the workflow, which I don't prefer.

In addition, this updates the `generate_schemas` macro to preserve field order for the newly generated schemas.

## Motivation and Context

This helps us ensure that the OpenAPI spec is up-to-date.

## How did you test it?

The check fails for commit `9c7a090` in this PR because of the spec not being up-to-date, which is fixed by the subsequent commit.
## Checklist


crates/router_derive/src/macros/generate_schema.rs
```diff
@@ -44,10 +43,10 @@ pub fn polymorphic_macro_derive_inner(
 
     // Go through all the fields and create a mapping of required fields for a schema
     // PaymentsCreate -> ["amount","currency"]
-    // This will be stored in a hashset
-    // mandatory_hashset -> ((PaymentsCreate, amount), (PaymentsCreate,currency))
-    let mut mandatory_hashset = HashSet::<(syn::Ident, syn::Ident)>::new();
-    let mut other_fields_hm = HashMap::<syn::Field, Vec<syn::Attribute>>::new();
+    // This will be stored in an `IndexSet`
+    // required_fields -> ((PaymentsCreate, amount), (PaymentsCreate,currency))
+    let mut required_fields = IndexSet::<(syn::Ident, syn::Ident)>::new();
```
Narayanbhat166, line 48: this need not be changed, as it is used only for the lookup

dracarys18 (approved):

Narayanbhat166 (approved):

jarnura (approved):

PR #1603, opened 2023-07-04
refactor(router): add psync task to process tracker after building connector request in payments flow

sai-harsha-vardhan:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [x] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description
add psync task to process tracker after building connector request in payments flow to prevent duplicate entries error in case of 4xx threw from connector flows

### Additional Changes

## Motivation and Context
we were adding psync task to process tracker for every payment attempt even though we threw 4xx from our side, in these cases if we do confirm again we were getting duplicate entry error from PT because it's the same attempt. To prevent this have moved psync task addition to process tracker after building connector request

## How did you test it?
Tested locally
task getting added only after building connector request
image

## Checklist

### DIFF
--- a/crates/router/src/core/payments.rs
+++ b/crates/router/src/core/payments.rs
@@ -36,7 +36,7 @@ use crate::{
     db::StorageInterface,
     logger,
     routes::{metrics, AppState},
-    scheduler::utils as pt_utils,
+    scheduler::{utils as pt_utils, workflows::payment_sync},
     services::{self, api::Authenticate},
     types::{
         self, api, domain,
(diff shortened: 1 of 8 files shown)


crates/router/src/core/payments.rs
```diff
@@ -596,6 +602,20 @@ where
         (None, false)
     };
 
+    if should_add_task_to_process_tracker {
```
Narayanbhat166, line 605: Why is the schedule time calculated before and inserting task is done later? Can't these be done at this place

sai-harsha-vardhan: This is to ensure that we do psync tasks based on schedule config set by merchant within fulfillment time

Narayanbhat166 (dismissed):

jarnura (approved):

Narayanbhat166 (approved):

PR #1627, opened 2023-07-06
feat(router): add connector_request_reference_id in router_data based on merchant config

sai-harsha-vardhan:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description
Currently connectors are sending payment_intent_id or payment_attempt_id to connectors as merchant reference and being used in webhook flows.
But for retries we need to standardize reference_id based on config set by merchant and send a standard value to connectors which connector flows can pass it to connector without any modification.

### Additional Changes

## How did you test it?
Tested manually
1. without config set
image
image

2. config set but merchant not enabled
image
image
 
3. config set and merchant enabled
image
image

## Checklist

### DIFF
--- a/crates/router/src/core/payments/transformers.rs
+++ b/crates/router/src/core/payments/transformers.rs
@@ -13,6 +13,7 @@ use crate::{
     core::{
         errors::{self, RouterResponse, RouterResult},
         payments::{self, helpers},
+        utils as core_utils,
     },
     routes::{metrics, AppState},
     services::{self, RedirectForm},
@@ -122,7 +123,13 @@ where
         reference_id: None,
         payment_method_token: payment_data.pm_token,
         connector_customer: payment_data.connector_customer_id,
-        preprocessing_id: payment_data.payment_attempt.preprocessing_step_id,
+        preprocessing_id: payment_data.payment_attempt.preprocessing_step_id.clone(),
+        connector_request_reference_id: core_utils::get_connector_request_reference_id(
+            &*state.store,
+            &merchant_account.merchant_id,
+            &payment_data.payment_attempt,
+        )
+        .await,
     };
 
     Ok(router_data)
--- a/crates/router/src/core/payments/transformers.rs
+++ b/crates/router/src/core/payments/transformers.rs
@@ -123,13 +123,13 @@ where
         reference_id: None,
         payment_method_token: payment_data.pm_token,
         connector_customer: payment_data.connector_customer_id,
-        preprocessing_id: payment_data.payment_attempt.preprocessing_step_id.clone(),
         connector_request_reference_id: core_utils::get_connector_request_reference_id(
             &*state.store,
             &merchant_account.merchant_id,
             &payment_data.payment_attempt,
         )
         .await,
+        preprocessing_id: payment_data.payment_attempt.preprocessing_step_id,
     };
 
     Ok(router_data)
--- a/crates/router/src/core/payments/transformers.rs
+++ b/crates/router/src/core/payments/transformers.rs
@@ -124,11 +124,10 @@ where
         payment_method_token: payment_data.pm_token,
         connector_customer: payment_data.connector_customer_id,
         connector_request_reference_id: core_utils::get_connector_request_reference_id(
-            &*state.store,
+            &state.conf,
             &merchant_account.merchant_id,
             &payment_data.payment_attempt,
-        )
-        .await,
+        ),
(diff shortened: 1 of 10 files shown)


crates/router/src/core/utils.rs
```diff
@@ -543,6 +574,43 @@ pub async fn construct_retrieve_file_router_data<'a>(
...
     Ok(router_data)
 }
+
+pub async fn get_connector_request_reference_id(
+    db: &dyn db::StorageInterface,
+    merchant_id: &str,
+    payment_attempt: &storage_models::payment_attempt::PaymentAttempt,
+) -> String {
+    let lookup_key = "send_payment_id_as_connector_request_reference_id";
+    let option_config_map = match db.find_config_by_key(lookup_key).await {
```
Narayanbhat166, line 589: Should this be stored in redis instead, as this db call will happen in all the flows where connector is called.

Narayanbhat166: Also, how can the merchant update this field? or is it us who have to update it as requested by the merchant.

sai-harsha-vardhan: We have to update this field while onboarding a merchant

crates/router/src/core/utils.rs
```diff
@@ -543,6 +577,43 @@ pub async fn construct_retrieve_file_router_data<'a>(
         reference_id: None,
         payment_method_token: None,
         preprocessing_id: None,
+        connector_request_reference_id: "irrelevant_connector_request_reference_id_in_dispute_flow"
```
Abhicodes-crypto, line 580: Can this be made a const.

crates/router/src/core/utils.rs
```diff
@@ -543,6 +577,43 @@ pub async fn construct_retrieve_file_router_data<'a>(
...
     Ok(router_data)
 }
+
+pub async fn get_connector_request_reference_id(
+    db: &dyn db::StorageInterface,
+    merchant_id: &str,
+    payment_attempt: &storage_models::payment_attempt::PaymentAttempt,
+) -> String {
+    let lookup_key = SEND_PAYMENT_ID_AS_CONNECTOR_REQUEST_REFERENCE_ID;
+    let option_config_map = match db.find_config_by_key(lookup_key).await {
```
Narayanbhat166, line 592: You can use this 
https://github.com/juspay/hyperswitch/blob/e913bfc4958da613cd352eca9bc38b23ab7ac38e/crates/router/src/db/configs.rs#L88

crates/router/src/core/payments/transformers.rs
```diff
@@ -122,7 +123,13 @@ where
...
         payment_method_token: payment_data.pm_token,
         connector_customer: payment_data.connector_customer_id,
-        preprocessing_id: payment_data.payment_attempt.preprocessing_step_id,
+        preprocessing_id: payment_data.payment_attempt.preprocessing_step_id.clone(),
+        connector_request_reference_id: core_utils::get_connector_request_reference_id(
+            &*state.store,
+            &merchant_account.merchant_id,
+            &payment_data.payment_attempt,
+        )
+        .await,
```
Abhicodes-crypto, line 132:
```suggestion
        connector_request_reference_id: core_utils::get_connector_request_reference_id(
            &*state.store,
            &merchant_account.merchant_id,
            &payment_data.payment_attempt,
        )
        .await,
        preprocessing_id: payment_data.payment_attempt.preprocessing_step_id,
```

Abhicodes-crypto (changes requested):

Abhicodes-crypto:
Just a ques, in case of `retries`/ `requeue` , if payment_id is passed as reference id how would webhook flow identify which payment_attempt is the webhook for ?

jarnura (changes requested):
We can move from config to environment variable.

sai-harsha-vardhan:
> Just a ques, in case of `retries`/ `requeue` , if payment_id is passed as reference id how would webhook flow identify which payment_attempt is the webhook for ?

`payment_id` is passed as reference id only if a merchant is enabled through the config, in that case both payment_id and attempt_id is same as retries isn't enabled for them. So, in webhooks flow it would always be connector_transaction_id or attempt_id

jarnura (approved):

Narayanbhat166 (approved):

Abhicodes-crypto (approved):

PR #1633, opened 2023-07-06
fix(core): Fix wallet payments throwing `Invalid 'payment_method_type' provided` and UI test issues

pixincreate:
## Type of Change

- [x] Bugfix

## Description

- This PR fixes `Invalid 'payment_method_type' provided` error thrown by the connector.
- Fix UI Test issues
- Address Klarna failures
- Update the credentials

## Motivation and Context

We fixed many bugs.

## How did you test it?

Manual through Github CI and through [hs-payments-test](hs-payments-test.netlify.app/)

## Checklist


.github/scripts/run_ui_tests.sh
```diff
@@ -28,4 +28,10 @@ while netstat -lnt | awk '$4 ~ /:8080$/ {exit 1}'; do
     else 
         sleep 10
     fi
-done
\ No newline at end of file
+done
+
+IN="$INPUT"
+for i in $(echo $IN | tr ";" "\n"); do
```
SanchithHegde, line 34: I'd suggest using a different separator like comma instead of semicolons, since semicolons are a common terminator used in shell scripts:

```shell
cd dir1; ls -la
```

crates/router/tests/connectors/selenium.rs
```diff
@@ -338,12 +338,24 @@ pub trait SeleniumTest {
         &self,
         c: WebDriver,
```
SanchithHegde, line 339: Please avoid single character variable names.

.github/workflows/connector-ui-sanity-tests.yml
```diff
@@ -101,8 +101,8 @@ jobs:
       - name: Install latest compiler
         uses: actions-rs/toolchain@v1
         with:
-            toolchain: stable
-            override: true
+          toolchain: stable
+          override: true
```
SanchithHegde, line 105: I don't think we need this to be specified, we use the same action in other workflows and haven't specified this input.

pixincreate: I checked with  `connector-sanity-tests.yml` and yes, we don't need `override` to be set to `true`. But, `toolchain` version is recommended to be kept as it is.

SanchithHegde (dismissed):

SanchithHegde (approved):

PR #1636, opened 2023-07-06
fix(locker): remove delete_locker_payment_method_by_lookup_key from payments_operation_core

ShankarSinghC:
## Type of Change

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

The function call responsible for deleting card details from payments_operations_core() has been removed in this PR. This modification ensures that the redundant deletion of already deleted card details no longer occurs, preventing the unnecessary error from being triggered.

### Additional Changes

## Motivation and Context

Previously, there was a function call within payments_operations_core() that deleted card details stored in the Basilisk Locker V3. However, it was discovered that the process tracker was also attempting to delete the same card details, resulting in an unnecessary error when trying to delete an already deleted card.

## How did you test it?

Manual, compiler-guided

## Checklist

### DIFF
--- a/crates/router/src/core/payments.rs
+++ b/crates/router/src/core/payments.rs
@@ -28,10 +28,7 @@ use self::{
 use super::errors::StorageErrorExt;
 use crate::{
     configs::settings::PaymentMethodTypeTokenFilter,
-    core::{
-        errors::{self, CustomResult, RouterResponse, RouterResult},
-        payment_methods::vault,
-    },
+    core::errors::{self, CustomResult, RouterResponse, RouterResult},
     db::StorageInterface,
     logger,
     routes::{metrics, AppState},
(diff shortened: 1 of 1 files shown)


vspecky (approved):

crates/router/src/core/payments.rs
```diff
@@ -198,11 +195,6 @@ where
                 .await?
             }
         };
-
-        if should_delete_pm_from_locker(payment_data.payment_intent.status) {
-            vault::Vault::delete_locker_payment_method_by_lookup_key(state, &payment_data.token)
-                .await
-        }
```
Narayanbhat166, line 205: Do not delete this function call. based on the operation decide whether to delete the payment method or not

Narayanbhat166 (changes requested):

crates/router/src/core/payments.rs
ShankarSinghC: There is an entry being made in the process tracker to delete card from vault after a specified amount of time which will delete the card details. Keeping the above function would return error when the process tracker tries to perform the delete operation as the card details would already be deleted by the above function.

Narayanbhat166 (approved):

SanchithHegde (approved):

PR #1652, opened 2023-07-07
fix(core): db error when doing payments through saved card flow

pixincreate:
## Type of Change

- [x] Bugfix

## Description

This PR aims to fix the database error when doing payments with payment token i.e., the saved card flow.  

When a payment is done succesfully, we get a `payment_token` which is `K1` i.e., the `hyperswitch_token`. On the other hand, `K2` i.e., `connector_token` is generated form the connector's end.  
Once the customer ID is used to save the payment method, we receive yet another `payment_token` and this `payment_token` is `parent_token` that has both `K1` and `K2`.  
As of now, when a successful payment is done, it is trying to delete `K1` which is intended, however, it is trying to delete the `parent_token` which does not exist in the database. This PR, extracts the `K1` from the `parent_token` and allows the core to delete the token after the payment is successful.

## How did you test it?

Manually through postman.
image

## Checklist

### DIFF
--- a/crates/router/src/core/payment_methods/vault.rs
+++ b/crates/router/src/core/payment_methods/vault.rs
@@ -17,7 +17,7 @@ use crate::{
         api,
         storage::{self, enums},
     },
-    utils::{self, StringExt},
+    utils::{self, OptionExt, StringExt},
 };
 #[cfg(feature = "basilisk")]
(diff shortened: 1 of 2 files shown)


pixincreate:
Converting to draft for better diagnosis of the issue.

PR #1659, opened 2023-07-08
feat(router): restricted customer update in payments-confirm and payments-update call via clientAuth

sahkal:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

Restrict customer_id update via clientAuth in payment-confirm and payment-update call to improve security measures.

### Additional Changes

## Motivation and Context

currently if we dont pass customer_id in payment_create call with confirm as false, and later pass customer_id in payment-confirm or payment-update call we are able to update the customer details via clientAuth. But client-secret and public-key both can be tracked through web browser and anyone can use customer card Token to make a payment, so to restrict this fraud case, we wont let customer_id updation using clientAuth it can be done only via api-key based auth.

## How did you test it?

tested through postman.
Screen Shot 2023-07-08 at 10 49 44 PM

## Checklist


crates/router/src/core/payments/helpers.rs
```diff
@@ -2540,3 +2540,16 @@ pub async fn get_additional_payment_data(
         }
     }
 }
+
+pub fn validate_customer_access(
+    auth_flow: services::AuthFlow,
+    request: &api::PaymentsRequest,
+) -> Result<(), errors::ApiErrorResponse> {
+    if auth_flow == services::AuthFlow::Client && request.customer_id.is_some() {
```
jarnura, line 2548: Here we are directly resisting they can't pass customer_id itself. Please verify any dependent services will break because of this. Ideal case they can pass id but it should be same.

jarnura (changes requested):

jarnura (approved):

crates/router/src/core/payments/helpers.rs
sahkal: done

crates/router/src/core/payments/operations/payment_confirm.rs
```diff
@@ -59,6 +60,25 @@ impl<F: Send + Clone> GetTracker<F, PaymentData<F>, api::PaymentsRequest> for Pa
...
+            let response = request
+                .clone()
+                .customer_id
+                .and_then(|customer| {
+                    payment_intent
+                        .clone()
+                        .customer_id
+                        .map(|payment_customer| payment_customer != customer)
+                })
+                .unwrap_or_default();
```
Narayanbhat166, line 73:
```suggestion
                .unwrap_or(false);
```

crates/router/src/core/payments/operations/payment_confirm.rs
```diff
@@ -59,6 +60,25 @@ impl<F: Send + Clone> GetTracker<F, PaymentData<F>, api::PaymentsRequest> for Pa
             .await
             .to_not_found_response(errors::ApiErrorResponse::PaymentNotFound)?;
 
+        if auth_flow == services::AuthFlow::Client && request.customer_id.is_some() {
+            let response = request
```
Narayanbhat166, line 64: An appropriate name for this variable would be `is_same_customer`

crates/router/src/core/payments/operations/payment_confirm.rs
```diff
@@ -59,6 +60,25 @@ impl<F: Send + Clone> GetTracker<F, PaymentData<F>, api::PaymentsRequest> for Pa
...
+                        .clone()
+                        .customer_id
+                        .map(|payment_customer| payment_customer != customer)
+                })
+                .unwrap_or_default();
+            if response {
+                return Err(errors::ApiErrorResponse::GenericUnauthorized {
+                    message: "Unauthorised access to update customer".to_string(),
+                }
+                .into());
```
Narayanbhat166, line 78:
```suggestion
                Err(errors::ApiErrorResponse::GenericUnauthorized {
                    message: "Unauthorised access to update customer".to_string(),
                }
                .into())?;
```

crates/router/src/core/payments/operations/payment_update.rs
```diff
@@ -64,6 +66,25 @@ impl<F: Send + Clone> GetTracker<F, PaymentData<F>, api::PaymentsRequest> for Pa
...
+                        .map(|payment_customer| payment_customer != customer)
+                })
+                .unwrap_or_default();
+            if response {
+                return Err(errors::ApiErrorResponse::GenericUnauthorized {
+                    message: "Unauthorised access to update customer".to_string(),
+                }
+                .into());
+            }
+        }
```
Narayanbhat166, line 86: this logic could have been moved to a helper function in order to avoid duplication, or else apply the same set of changes as suggested above.

Narayanbhat166 (changes requested):

crates/router/src/core/payments/operations/payment_update.rs
sahkal: done!

crates/router/src/core/payments/operations/payment_confirm.rs
sahkal: done!

crates/router/src/core/payments/operations/payment_confirm.rs
sahkal: done!

crates/router/src/core/payments/operations/payment_confirm.rs
sahkal: done

crates/router/src/core/payments/operations/payment_confirm.rs
Narayanbhat166: My bad, since this is checking whether the customer is not same, it should be `is_not_same_customer`. Can you please correct it?

sahkal: done!

Narayanbhat166 (approved):

lsampras (approved):

PR #1674, opened 2023-07-11
feat:  Convert QrData into Qr data image source url

Sangamesh26:
## Type of Change

- [x] New feature

## Description

Qr image data is converted into image data source url.
I have used `qrcode` and `image` external crates to convert data into qr code and then to image data source url.
Some of the connectors send QrCode data and ask us to convert it into Qr code image data url, so this util functions add support for the issue.

## Motivation and Context

Add support for QrData conversion into image data source url.

## How did you test it?

Screen Shot 2023-07-11 at 2 17 44 PM

Screen Shot 2023-07-11 at 2 20 29 PM

## Checklist


crates/router/src/utils.rs
```diff
@@ -155,3 +158,51 @@ pub fn to_currency_base_unit_asf64(
...
+            .change_context(common_utils::errors::QrCodeError::FailedToCreateQrCode)?;
+
+        // Renders the QR code into an image.
+        let qrcode_image_buffer = qr_code.render::<Luma<u8>>().build();
+        let qrcode_dynamic_image = image::DynamicImage::ImageLuma8(qrcode_image_buffer);
+
+        let mut image_bytes: Vec<u8> = Vec::new();
+
+        // Encodes qrcode_dynamic_image and write it to image_bytes
+        let _ = qrcode_dynamic_image.write_to(&mut image_bytes, image::ImageOutputFormat::Png);
```
dracarys18, line 188: You can use `BufWriter` here if in the future if you do two or more writes this is suggested.

dracarys18:
Also you can use https://github.com/etemesi254/zune-image instead of `image`. Optional

crates/router/src/utils.rs
```diff
@@ -155,3 +158,51 @@ pub fn to_currency_base_unit_asf64(
...
+
+#[derive(Debug)]
+pub struct QrImage;
+
+pub trait CreateQrImageUrlFromQrData {
+    fn create_image_data_source_url(
+        &self,
+        _qr_code_data: String,
+    ) -> Result<String, error_stack::Report<common_utils::errors::QrCodeError>>;
+}
```
dracarys18, line 170: Why is this trait needed? You are not gonna use this for any other type right? You can just use

```rust
impl QrImage {
	pub fn create_image_data_source_url(&self,_qr_code_data:String){}
}
```

dracarys18 (changes requested):

crates/router/src/utils.rs
Sangamesh26: There is no `BufWriter` function implemented for `DynamicImage`

dracarys18: Not needed

dracarys18 (approved):

jagan-jaya (approved):

AkshayaFoiger (approved):

ArjunKarthik (approved):

PR #1675, opened 2023-07-11
feat(core): create capture entity for each manual capture request

hrithikesh026:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

For each manual capture request, a new entry is created in captures table (irrespective of partial or full). 
This will be useful for implementing multiple partial capture.

This PR only creates an entry in capture table. For status update of capture, we need to do the changes at connector's end in future.

### Additional Changes

## Motivation and Context

With current DB models, it is not possible to make and track multiple partial capture. To solve this, a new captures table must be created.

## How did you test it?

Manual
Screenshot 2023-07-11 at 5 10 04 PM

## Checklist


migrations/2023-07-07-091223_create_captures_table/up.sql
```diff
@@ -0,0 +1,25 @@
+
+CREATE TYPE "CaptureStatus" AS ENUM (
+    'started',
+    'charged',
+    'pending',
+    'failure'
+);
+CREATE TABLE captures(
+    id SERIAL PRIMARY KEY,
```
ashokkjag, line 9: I don't think we use auto generated id's. it will be difficult to move the data in future.

hrithikesh026:
Closed in favour of #1721 .

migrations/2023-07-07-091223_create_captures_table/up.sql
hrithikesh026: All the other tables in DB has `id` field which is auto increment. 
I just did the same here.

PR #1687, opened 2023-07-12
feat(connector): [PowerTranz] Add cards support for PowerTranz connector

ArjunKarthik:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description
Cards Non 3ds support for powertranz connector

### Additional Changes

## Motivation and Context
Cards Non 3ds support for powertranz connector

## How did you test it?
Unit Test

Screen Shot 2023-07-11 at 9 12 30 PM

Postman:
Automatic Payment
auto_create
Manual Payment
manual_create
Cancel Payment
Cancel
Capture Payment
capture
Refund
Refund

## Checklist


crates/router/src/connector/powertranz/transformers.rs
```diff
@@ -1,210 +1,446 @@
...
+        }?;
+        let billing_address = get_address_details(&item.address.billing, &item.request.email);
+        let shipping_address = get_address_details(&item.address.shipping, &item.request.email);
+        Ok(Self {
+            transaction_identifier: Uuid::new_v4().to_string(),
+            total_amount: utils::to_currency_base_unit_asf64(
+                item.request.amount,
+                item.request.currency,
+            )?,
+            currency_code: String::foreign_from(item.request.currency),
```
SamraatBansal, line 75: Move this to `connector_utils`

SamraatBansal (changes requested):

crates/router/src/connector/powertranz.rs
```diff
@@ -103,20 +111,22 @@ impl ConnectorCommon for Powertranz {
...
+        let message = first_error.map(|error| error.message.clone());
 
         Ok(ErrorResponse {
             status_code: res.status_code,
-            code: response.code,
-            message: response.message,
-            reason: response.reason,
+            code: code.unwrap_or_else(|| consts::NO_ERROR_CODE.to_string()),
+            message: message.unwrap_or_else(|| consts::NO_ERROR_MESSAGE.to_string()),
+            reason: None,
```
prasunna09, line 122: Why are we sending reason "None" when the message is errors.first(). Why can't we concatenate the other errors send them in reason?

crates/router/src/connector/powertranz/transformers.rs
```diff
@@ -1,210 +1,332 @@
...
-#[derive(Default, Debug, Serialize, Eq, PartialEq)]
+#[derive(Debug, Serialize)]
+#[serde(rename_all = "PascalCase")]
 pub struct PowertranzPaymentsRequest {
-    amount: i64,
-    card: PowertranzCard,
+    transaction_identifier: String,
+    total_amount: f64,
+    currency_code: String,
+    three_d_s_ecure: bool,
```
prasunna09, line 21: Is it supposed to be "ThreeDSEcure" or "ThreeDSecure"? If "ThreeDSecure" then it must be "three_d_secure".

crates/router/src/connector/powertranz/transformers.rs
```diff
@@ -1,210 +1,332 @@
...
-                    cvc: req_card.card_cvc,
-                    complete: item.request.is_auto_capture()?,
-                };
-                Ok(Self {
-                    amount: item.request.amount,
-                    card,
+        let source = match item.request.payment_method_data.clone() {
+            api::PaymentMethodData::Card(card) => Ok(Source::from(&card)),
+            _ => Err(errors::ConnectorError::NotImplemented(
+                "Payment methods".to_string(),
```
prasunna09, line 64:
```suggestion
                "Payment method".to_string(),
```

config/development.toml
```diff
@@ -143,7 +143,7 @@ payeezy.base_url = "https://api-cert.payeezy.com/"
 payme.base_url = "https://sandbox.payme.io/"
 paypal.base_url = "https://www.sandbox.paypal.com/"
 payu.base_url = "https://secure.snd.payu.com/"
-powertranz.base_url = "https://staging.ptranz.com/api/spi/"
+powertranz.base_url = "https://staging.ptranz.com/api/"
```
prasunna09, line 146: Observed that you have modified base url here, can you please modify in other places as well, as the base url there is still "https://staging.ptranz.com/api/spi/"

SamraatBansal (approved):

prasunna09 (dismissed):

crates/router/src/connector/utils.rs
```diff
@@ -1031,3 +1041,117 @@ impl ForeignTryFrom<String> for CanadaStatesAbbreviation {
         }
     }
 }
+
+impl ForeignFrom<diesel_models::enums::Currency> for String {
```
SanchithHegde, line 1045: Would be preferable to have this as a method on `Currency` itself.

```rust
impl Currency {
    pub fn iso_code(&self) -> &'static str {
        // ...
    }
}

File: crates/router/src/connector/powertranz/transformers.rs [hunk 7]
```diff
@@ -1,210 +1,336 @@
...
+        let source = match item.request.payment_method_data.clone() {
+            api::PaymentMethodData::Card(card) => Ok(Source::from(&card)),
+            _ => Err(errors::ConnectorError::NotImplemented(
+                "Payment method".to_string(),
+            )),
+        }?;
+        let billing_address = get_address_details(&item.address.billing, &item.request.email);
+        let shipping_address = get_address_details(&item.address.shipping, &item.request.email);
+        Ok(Self {
+            transaction_identifier: Uuid::new_v4().to_string(),
```
SanchithHegde, line 70: Shouldn't we be sending an identifier that we store as the transaction identifier? Wouldn't that help us with recon later, if required?

crates/router/src/connector/powertranz/transformers.rs
```diff
@@ -1,210 +1,336 @@
...
+                true => Self::Voided,
+                false => Self::VoidFailed,
+            },
+            // Refund
+            5 => match approved {
+                true => Self::AutoRefunded,
+                false => Self::Failure,
+            },
+            // Risk Management
+            _ => match approved {
```
SanchithHegde, line 183: Will this be the case for any transaction type > 5?
Would be good to have a known upper cap for transaction type (say 10, this would depend on connector docs of course) beyond which we consider it an invalid transaction type.

ArjunKarthik: 1,2,3,4,5 are the possible values

SanchithHegde: Should we then consider raising an error if anything more than 5 is sent?

crates/router/src/connector/powertranz/transformers.rs
ArjunKarthik: transaction_identifier should be in GUID format, we are passing our payment id in order_identifier field

SanchithHegde: Okay I figured out that we're storing it as connector transaction ID. Should be fine, marking this resolved.

prasunna09 (dismissed):

crates/router/src/connector/powertranz/transformers.rs
SanchithHegde: I understand. What would happen if the connector does end up sending transaction type other than these values?

Won't it be better to error out as invalid transaction type instead of mapping status to pending/failure?

SanchithHegde (approved):

PR #1693, opened 2023-07-12
feat(payments): add client secret in redirect response 

Abhicodes-crypto:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

add client secret in redirect response

### Additional Changes

## How did you test it?

Manual 
image

## Checklist

### DIFF
--- a/crates/router/src/core/payments/helpers.rs
+++ b/crates/router/src/core/payments/helpers.rs
@@ -1435,6 +1435,7 @@ pub fn get_handle_response_url(
         merchant_account,
         redirection_response,
         payments_return_url,
+        response.client_secret.as_ref(),
     )
     .attach_printable("Failed to make merchant url with response")?;
 
@@ -1445,6 +1446,7 @@ pub fn make_merchant_url_with_response(
     merchant_account: &domain::MerchantAccount,
     redirection_response: api::PgRedirectResponse,
     request_return_url: Option<&String>,
+    client_secret: Option<&masking::Secret<String>>,
 ) -> RouterResult<String> {
     // take return url if provided in the request else use merchant return url
     let url = request_return_url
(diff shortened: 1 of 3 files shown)


crates/router/src/core/payments/helpers.rs
```diff
@@ -1453,14 +1455,20 @@ pub fn make_merchant_url_with_response(
 
     let status_check = redirection_response.status;
 
-    let payment_intent_id = redirection_response.payment_id;
+    let payment_client_secret = client_secret
+        .ok_or(errors::ApiErrorResponse::InternalServerError)
+        .into_report()
+        .attach_printable("Expected client secret to be `Some`")?;
```
Abhicodes-crypto, line 1461: @jarnura @Narayanbhat166 Should error be thrown over here as InternalServerError or client_secret_destroyed.  Currently client secret is deleted only when status is canceled or when it is expired.

Abhicodes-crypto: Based on offline discussion , client secret will be made mandatory so it will be removed as optional in a separate PR.

Narayanbhat166: The client secret should not be deleted, if we do not find the client secret in `PaymentAttempt` then it is an invalid state. We should throw internal server error in this case.

Narayanbhat166 (approved):

crates/router/src/core/payments/helpers.rs
Abhicodes-crypto: * `PaymentIntent`

inventvenkat (approved):

PR #1651, opened 2023-07-07
refactor(payment_methods): Fix db insert for payment method create

Sarthak1799:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

Currently, payment method create API does not insert payment method record in the Db. 
have added the db call to insert the record if not already present.

### Additional Changes

- [ ] This PR modifies the API contract
- [ ] This PR modifies the database schema
- [ ] This PR modifies application configuration/environment variables

## How did you test it?

image

image

image

## Checklist

- [x] I formatted the code `cargo +nightly fmt --all`
- [x] I addressed lints thrown by `cargo clippy`
- [x] I reviewed submitted code
- [ ] I added unit tests for my changes where possible
- [ ] I added a [CHANGELOG](/CHANGELOG.md) entry if applicable

PR #1604, opened 2023-07-04
ci(runner): rewrite `collection_runner.sh` in rust

pixincreate:
## Type of Change

- [x] Enhancement
- [x] CI/CD

## Description

This PR aims at re-writing the collection_runner shell script to rust. With this, it drastically reduces the developer efforts to handle shell code along side the rust code by completely removing the shell code requirement.   

**Note:** #1651 needs to be merged to get Stripe and ACI collection working.

#### Usage
- Make sure you that you've the postman collection for the connector is ready in the `postman` dir with the name `<connector_name>.postman_collection.json`
- Add the connector credentials to the `connector_auth.toml` / `auth.toml`
- In terminal, execute:
  ```zsh
  export CONNECTOR_AUTH_FILE_PATH=/path/to/auth.toml
  cargo run --package test_utils --bin test_utils -- --connector_name=<connector_name> --base_url=<base_url> --admin_api_key=<admin_api_key>    
  ```

## Motivation and Context

`collection_runner.sh` required additional maintainence and rustification of the same eliminates the necessity of maintaining shell code separately. With that, everytime a new connector is created, the developer don't need to make connector specific changes (if any) to the shell code. 

## How did you test it?

- Via GitHub Actions:
https://github.com/juspay/hyperswitch/actions/runs/5506050847

- Via Termina:
  - NMI
image
  - Bluesnap
image
  - Checkout
image
  - Worldline
image

## Checklist

- [x] I formatted the code `cargo +nightly fmt --all`
- [x] I addressed lints thrown by `cargo clippy`
- [x] I reviewed submitted code

PR #1695, opened 2023-07-12
fix(core): Address 500 when deleting payment method and add logs to postman collections

pixincreate:
## Type of Change

- [x] Bugfix

## Description

- When a payment is done through saved card flow, the details is stored in Vault. To delete this, we need to pass `payment_method_id`. Since the `payment_method_id` is **not** exposed, it is impossible to delete the payment data
- When a `PaymentMethod` is created without doing a payment, it is stored in the database (`payment_methods` table) and not in the `configs` table
- But when delete `PaymentMethod` request method is made from the postman, `payment_method_delete_api` is called in `api.rs`. This `payment_method_delete_api` calls `delete_payment_method`
- `500` is raised when `get_payment_method_data_from_locker` is called where `find_config_by_key` is called. It requires `key` which is the temporary `payment_token` (Obtained when `PaymentCreate` is done), but `payment_method_id` is being supplied which is not intended

Flow:
- 2 calls to database is made and 1 call to locker
- 1st call to database is made to get the key (`payment_method_id`), for safety
- Now, with `pm_id`, the Card details from the locker is deleted. If the deletion is successful, it is logged, if not, we've the key stored previosuly to make delete call again
- The `payment_method_id` is then deleted from the database

**Note:** It is required that the collections need to update the `PaymentMethods` folder by taking the updated Stripe collection into consideration since for deleting a Payment Method, we should be using `payment_method_id` and not `payment_token`.

## Motivation and Context

This bug came into light while running the postman collections through #1604. When RCA was done, I learnt that we're not storing the `payment_method_id` in the first place. This was fixed by @Sarthak1799 with #1651. However, deletion of `payment_method_id` still remained an issue that this PR addresses.

## How did you test it?

Manual through postman [13/07/2023]
image

Stripe collection for used as reference for testing Payment Methods (Update and Delete)
image

## Checklist


crates/router/src/core/payment_methods/cards.rs
```diff
@@ -1993,37 +1993,45 @@ pub async fn retrieve_payment_method(
...
+            state,
+            &key.customer_id,
+            &key.merchant_id,
+            pm_id.payment_method_id.as_str(),
+        )
+        .await?;
+
         if response.status == "SUCCESS" {
-            print!("Card From locker deleted Successfully")
+            db.delete_payment_method_by_merchant_id_payment_method_id(
```
NishantJoshi00, line 2014: add `logger::info!` indicating deletion

crates/router/src/core/payment_methods/cards.rs
```diff
@@ -1993,37 +1993,45 @@ pub async fn retrieve_payment_method(
...
-
-    Ok(services::ApplicationResponse::Json(
-        api::PaymentMethodDeleteResponse {
-            payment_method_id: pm.payment_method_id,
-            deleted: true,
-        },
-    ))
+    } else {
+        // This part of code is like a place holder and the control will never come here since everything is handled above
+        Err(errors::ApiErrorResponse::PaymentMethodNotFound)?
```
NishantJoshi00, line 2035: In cases where the payment method isn't `Card`. we should have certain handling in place to redact or remove the payment method. 
What do you think @jarnura, @lsampras?

NishantJoshi00: No need to handle this in the PR, but would be good to create a issue if necessary

NishantJoshi00 (changes requested):

NishantJoshi00 (approved):

crates/router/src/core/payment_methods/cards.rs
```diff
@@ -2010,36 +2010,43 @@ pub async fn retrieve_payment_method(
...
+            pm_id.payment_method_id.as_str(),
+        )
+        .await?;
+
         if response.status == "SUCCESS" {
-            print!("Card From locker deleted Successfully")
+            logger::info!("Card From locker deleted Successfully!");
         } else {
-            print!("Error: Deleting Card From Locker")
+            logger::info!("Error: Deleting Card From Locker!");
```
ashokkjag, line 2033: logger.error? Is some context available in the logs for debugging the error?

crates/router/src/core/payment_methods/cards.rs
```diff
@@ -2010,36 +2010,43 @@ pub async fn retrieve_payment_method(
...
-            delete_card_from_locker(state, &pm.customer_id, &pm.merchant_id, &payment_method_id)
-                .await?;
+    if key.payment_method == enums::PaymentMethod::Card {
+        let response = delete_card_from_locker(
+            state,
+            &key.customer_id,
+            &key.merchant_id,
+            pm_id.payment_method_id.as_str(),
+        )
+        .await?;
```
ashokkjag, line 2028: for the `?`s , should the failure always abort or should the deletion from DB still happen? what action can the user (customer/merchant?) take based on the response to get the payment_method deleted?

pixincreate: We cannot delete from the database if this fails ( due to some locker error ) as we will lose the token that we need to delete from locker. If there is any error, then the user can retry after sometime ( after we resolve the error ).

crates/router/src/core/payment_methods/cards.rs
```diff
@@ -1885,36 +1885,43 @@ pub async fn retrieve_payment_method(
 pub async fn delete_payment_method(
     state: &routes::AppState,
     merchant_account: domain::MerchantAccount,
-    pm: api::PaymentMethodId,
+    pm_id: api::PaymentMethodId,
 ) -> errors::RouterResponse<api::PaymentMethodDeleteResponse> {
-    let (_, supplementary_data) =
-        vault::Vault::get_payment_method_data_from_locker(state, &pm.payment_method_id).await?;
```
Narayanbhat166, line 1891: Why is this removed? Please have a look if this is necessary and add it back

pixincreate: Please read the description.   

To delete from Vault (locker), we need `payment_token` which can be obtained when doing `save card flow` or `mandates`. Other than that, we're not storing the `payment_method_id` or its data. When we create a `payment_method`, we create its related `id` and never store that in the locker but rather database. So it is useless checking there and it just throws 500.

crates/router/src/core/payment_methods/cards.rs
```diff
@@ -1885,36 +1885,43 @@ pub async fn retrieve_payment_method(
...
+        )
+        .await?;
+
         if response.status == "SUCCESS" {
-            print!("Card From locker deleted Successfully")
+            logger::info!("Card From locker deleted Successfully!");
         } else {
-            print!("Error: Deleting Card From Locker")
+            logger::info!("Error: Deleting Card From Locker! {:#?}", response);
+            return Err(errors::ApiErrorResponse::UnprocessableEntity {
```
vspecky, line 1909: Why is the error `UnprocessableEntity`? Shouldn't it be `InternalServerError`? (since it's an internal system issue)

vspecky (changes requested):

vspecky (dismissed):

Narayanbhat166 (dismissed):

crates/router/src/core/payment_methods/cards.rs
```diff
@@ -2059,36 +2059,41 @@ pub async fn retrieve_payment_method(
...
+            pm_id.payment_method_id.as_str(),
+        )
+        .await?;
+
         if response.status == "SUCCESS" {
-            print!("Card From locker deleted Successfully")
+            logger::info!("Card From locker deleted Successfully!");
         } else {
-            print!("Error: Deleting Card From Locker")
+            logger::info!("Error: Deleting Card From Locker! {:#?}", response);
```
ashokkjag, line 2082: logger::error!?

ashokkjag (approved):

crates/router/src/core/payment_methods/cards.rs
pixincreate: I misunderstood. I've updated the line.

vspecky (approved):

ashokkjag (approved):

NishantJoshi00 (approved):

Issue #1532, opened 2023-06-23
[BUG] "mandate_data" returned as null in the PaymentsCreate response body while creating a new mandate payment


kritikmodi:
### Bug Description

While creating a new mandate payment, the `mandate_data` field is being returned as **null** in the response body. This occurs even after passing the right `mandate_data` content in the request body.

### Expected Behavior

`mandate_data` should be set and returned in accordance to what is being passed in the request body.

### Steps To Reproduce

Create a mandate payment by passing a valid set of mandate data and check the response body.

### Environment

Local

### Have you spent some time to check if this bug has been raised before?

- [X] I checked and didn't find similar issue

### Have you read the Contributing Guidelines?

- [X] I have read the [Contributing Guidelines](https://github.com/juspay/hyperswitch/blob/main/docs/CONTRIBUTING.md)

### Are you willing to submit a PR?

Yes I am willing to submit a PR!

PR #1715, opened 2023-07-14
fix(payments): populate mandate_data in the response body of the PaymentsCreate endpoint

kritikmodi:
## Type of Change

- [x] Bugfix

## Description

Earlier `mandate_data` was being returned as `null`. This PR fixes the same and populates `mandate_data` with the same content as sent in the request body.

## Motivation and Context

This PR closes [this](https://github.com/juspay/hyperswitch/issues/1532) issue.

## How did you test it?

Locally using Postman

## Checklist


kritikmodi:
Screen Shot 2023-07-14 at 4 55 24 PM

crates/router/src/core/payments/transformers.rs
```diff
@@ -269,6 +270,7 @@ pub fn payments_to_payments_response<R, Op>(
     operation: &Op,
     ephemeral_key_option: Option<ephemeral_key::EphemeralKey>,
     session_tokens: Vec<api::SessionToken>,
+    mandate_data: Option<api_models::payments::MandateData>,
```
Narayanbhat166, line 273: add `[auth_based]` attribute to this. This field should only be sent in the response if `api_key` is used. Check how it is done for `payment_method_data` in `PaymentsResponse`.

Narayanbhat166 (changes requested):

crates/router/src/core/payments/transformers.rs
kritikmodi: Okay

Narayanbhat166 (approved):

vspecky (approved):

PR #1725, opened 2023-07-17
feat(router): add attempt_count field in attempt update record of payment_intent

sai-harsha-vardhan:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description
While updating the payment_intent's active attempt id in automatic retries we need to update attempt_count too. To support that have added attempt_count field in attempt update record of payment_intent

### Additional Changes

## How did you test it?
Compiler guided

## Checklist

### DIFF
--- a/crates/diesel_models/src/payment_intent.rs
+++ b/crates/diesel_models/src/payment_intent.rs
@@ -134,8 +134,9 @@ pub enum PaymentIntentUpdate {
         order_details: Option<Vec<pii::SecretSerdeValue>>,
         metadata: Option<pii::SecretSerdeValue>,
     },
-    PaymentAttemptUpdate {
+    PaymentAttemptAndAttemptCountUpdate {
         active_attempt_id: String,
+        attempt_count: i16,
     },
     StatusAndAttemptUpdate {
         status: storage_enums::IntentStatus,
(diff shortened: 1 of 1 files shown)


Narayanbhat166 (approved):

crates/diesel_models/src/payment_intent.rs
```diff
@@ -134,8 +134,9 @@ pub enum PaymentIntentUpdate {
         order_details: Option<Vec<pii::SecretSerdeValue>>,
         metadata: Option<pii::SecretSerdeValue>,
     },
-    PaymentAttemptUpdate {
+    PaymentAttemptAndAttemptCountUpdate {
```
Abhicodes-crypto, line 137: Should we rename this as `AutoRetryAttemptUpdate` and below variant as `ManualRetryAttemptUpdate` ? I guess there are no other use cases for these structs

Abhicodes-crypto (approved):

crates/diesel_models/src/payment_intent.rs
sai-harsha-vardhan: But these names imply what are we updating, so this would be better even if it's used anywhere else in future right!

SanchithHegde (approved):

PR #1730, opened 2023-07-17
fix(connector): [Tsys] Update endpoint and unit tests

SamraatBansal:
## Type of Change

- [X] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

Update the case of the endpoint in payments and refunds flow to make it lowercase consistent.
Update Failing Unit Tests 

### Additional Changes

## Motivation and Context

Update the case of the endpoint in payments and refunds flow to make it lowercase consistent.
Update Failing Unit Tests 

## How did you test it?

**Tests**
Screen Shot 2023-07-17 at 6 09 21 PM

## Checklist

- [X] I added unit tests for my changes where possible


crates/router/src/connector/tsys/transformers.rs
```diff
@@ -464,6 +464,7 @@ impl From<TsysTransactionDetails> for enums::RefundStatus {
     fn from(item: TsysTransactionDetails) -> Self {
         match item.transaction_status {
             TsysTransactionStatus::Approved => Self::Pending,
+            //Connector calls refunds as Void
             TsysTransactionStatus::Void => Self::Success,
```
pixincreate, line 468: If refund is called as void, what do they call for voiding a payment? Or do they even support?

SamraatBansal: Both Void and Refunds are addressed as `Void`
We handle them accordingly in response on the designated call either payments or refunds

crates/router/tests/connectors/tsys.rs
```diff
@@ -286,9 +296,14 @@ async fn should_refund_succeeded_payment_multiple_times() {
 #[actix_web::test]
 async fn should_sync_refund() {
     let refund_response = CONNECTOR
-        .make_payment_and_refund(payment_method_details(), None, get_default_payment_info())
+        .make_payment_and_refund(
+            payment_method_details(100),
+            None,
```
ArjunKarthik, line 301: pass full amount for refund

crates/router/tests/connectors/tsys.rs
```diff
@@ -286,9 +296,14 @@ async fn should_refund_succeeded_payment_multiple_times() {
...
+            get_default_payment_info(),
+        )
         .await
         .unwrap();
+    tokio::time::sleep(Duration::from_secs(10)).await;
```
ArjunKarthik, line 306: Do we still need this sleep time?

ArjunKarthik (changes requested):

crates/router/tests/connectors/tsys.rs
SamraatBansal: Yes in this transaction is registered after a lag, that is why sleep is added

crates/router/tests/connectors/tsys.rs
SamraatBansal: It will by default pick 100, Rest at every place it is handled

ArjunKarthik (approved):

NishantJoshi00 (approved):

PR #1731, opened 2023-07-18
fix(router): add parsing for `connector_request_reference_id` env

sai-harsha-vardhan:
## Type of Change

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description
add parsing for `connector_request_reference_id` env

### Additional Changes

## How did you test it?
Env
image

Config while running application
image

## Checklist

### DIFF
--- a/crates/router/src/configs/settings.rs
+++ b/crates/router/src/configs/settings.rs
@@ -553,7 +553,7 @@ pub struct DelayedSessionConfig {
 
 #[derive(Debug, Deserialize, Clone, Default)]
 pub struct ConnectorRequestReferenceIdConfig {
-    pub merchant_ids_send_payment_id_as_connector_request_id: HashSet<String>,
+    pub merchant_ids_send_payment_id_as_connector_request_id: Vec<String>,
 }
 
 fn delayed_session_deser<'a, D>(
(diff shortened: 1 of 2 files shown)


crates/router/src/core/utils.rs
```diff
@@ -599,7 +599,7 @@ pub fn is_merchant_enabled_for_payment_id_as_connector_request_id(
     let config_map = &conf
         .connector_request_reference_id_config
         .merchant_ids_send_payment_id_as_connector_request_id;
-    config_map.contains(merchant_id)
+    config_map.iter().any(|id| id == merchant_id)
```
lsampras, line 602: what's the difference?

crates/router/src/configs/settings.rs
```diff
@@ -553,7 +553,7 @@ pub struct DelayedSessionConfig {
 
 #[derive(Debug, Deserialize, Clone, Default)]
 pub struct ConnectorRequestReferenceIdConfig {
-    pub merchant_ids_send_payment_id_as_connector_request_id: HashSet<String>,
+    pub merchant_ids_send_payment_id_as_connector_request_id: Vec<String>,
```
lsampras, line 556: why this?
Does config crate not support reading hashset from comma separated envs?

lsampras:
> How did you test it?
>
> Compiler guided

you should be able to test this via running it locally with env vars since compiler won't point much here

crates/router/src/core/utils.rs
sai-harsha-vardhan: contains expects &String, so had to convert &str to &String. To avoid that, changed it to iter any logic

lsampras (approved):

SanchithHegde (approved):

PR #1739, opened 2023-07-18
ci(postman): Run Postman collections against PRs 

pixincreate:
## Type of Change

- [x] Refactoring
- [x] CI/CD

## Description

This PR aims at running the postman collections against the PRs raised instead of running on schedule on Integ so that any changes that break the main will be caught at PR level.  

Keeping costs in mind, I would suggest running the collections at the time of merge queue rather than running at every commit. 

## Motivation and Context

It becomes a headache to track down the PR that broke the main. So, running tests globally for every PR would fix the issue. 

## Checklist


.github/workflows/postman-collection-runner.yml
```diff
@@ -1,68 +1,124 @@
-name: Postman Collection Runner
+name: Run collections against your PRs
 
 on:
   workflow_dispatch:
-  schedule:
-    - cron: "30 0,12 * * *" # Run workflow at 6 AM and 6 PM IST
+  merge_group:
```
ashokkjag, line 5:
```suggestion
  merge_group:
      types: [checks_requested]
```

.github/workflows/postman-collection-runner.yml
```diff
@@ -1,68 +1,124 @@
...
-          - bluesnap
-          # - braintree
-          - checkout
-          # - iatapay
-          # - multisafepay
-          - nmi
-          - stripe
-          # - trustpay
-          - worldline
+          - aci, checkout, stripe
```
ashokkjag, line 27: verify that this is what you want? syntax may be wrong here.

.github/workflows/postman-collection-runner.yml
```diff
@@ -1,68 +1,124 @@
...
+        id: config_path
+        shell: bash
+        run: |
+          echo "CONNECTOR_AUTH_FILE_PATH=$HOME/target/test/connector_auth.toml" >> $GITHUB_ENV
+
+      - name: Fetch keys
+        env:
+          TOML_PATH: "./config/development.toml"
+        run: |
+          KEY_LINE=$(grep -E '^\[secrets\]$|^\s*admin_api_key\s*=' "$TOML_PATH")
```
ashokkjag, line 74: I see `yq` is already available in the ubuntu runner.

you can use 
```bash
> yq '.secrets.admin_api_key' config/development.toml
test_admin
```

crates/test_utils/src/main.rs
```diff
@@ -97,11 +96,11 @@ fn main() {
             }
             // Handle other ConnectorAuthType variants
             _ => {
-                eprintln!("Invalid authentication type.");
+                eprintln!("\x1b[31mInvalid authentication type.\x1b[0m");
```
ashokkjag, line 99: Any reason for adding the ansi codes. these are OK for terminal but not useful for github actions.

ashokkjag (changes requested):

crates/test_utils/src/main.rs
pixincreate: This did work in GitHub actions when I tested it yesterday. I'll remove this for `--color` command in Newman

.github/workflows/postman-collection-runner.yml
ashokkjag: After discussion, intention is to pass multiple `,` separated values to each job to reduce the startup/runtime.

ashokkjag (approved):

.github/workflows/postman-collection-runner.yml
pixincreate: Yea, this reduces startup time and hence allows us to reuse already set-up server for upcoming connectors.   
With this, we could effectively reduce the run time from a mere 16+ minutes to ~7 minutes

ashokkjag (dismissed):

SanchithHegde (commented):

SanchithHegde (approved):

SanchithHegde (dismissed):

ashokkjag (approved):

PR #1765, opened 2023-07-21
feat(router): add grouping and priority logic in connector utils to handle multiple errors in connector flows

sai-harsha-vardhan:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description
In few connectors, we may get list of error codes and messages in a response.
In this case, we need a grouping and priority logic of errors to identify and give most prior error in error code and error message.

Add a `ConnectorErrorType` enum which has following groups with respective priorities
UnknownError = 1
UserError = 2
BusinessError = 3
TechnicalError = 4

Add a `get_error_code_error_message_based_on_priority` utils fn which take list of error codes and messages and gives most prior error based on connector error type

### Additional Changes

## How did you test it?
Add a Unit test
image

## Checklist


crates/router/src/types/api.rs
```diff
@@ -131,7 +142,8 @@ impl<
             + Dispute
             + FileUpload
             + ConnectorTransactionId
-            + Payouts,
+            + Payouts
+            + ConnectorErrorTypeMapping,
```
Narayanbhat166, line 146: Do we need to enforce this for every connector? This is not necessary for all the connectors right?

Narayanbhat166 (changes requested):

crates/router/src/types/api.rs
sai-harsha-vardhan: Yeah this is optional and need to be implemented only when the utils function is called where connector sends list of errors

Narayanbhat166 (approved):

SanchithHegde (approved):

PR #1786, opened 2023-07-25
refactor(redis_interface): remove the `Drop` implementation on `RedisConnectionPool`

### LINKED ISSUE #1782: [REFACTOR] The `Drop` implementation on `RedisConnectionPool` seems unnecessary
### Description

I'll keep this issue as a source of documentation and decisions taken regarding the graceful shutdown of Redis connections.

When I was working on #1743, I noticed that the `Drop` implementation on `RedisConnectionPool` caused panics when tests were being run.

https://github.com/juspay/hyperswitch/blob/1ab4226c780e9205785f012fd1c48c7a4bafb48f/crates/redis_interface/src/lib.rs#L197-L203

When I went down the rabbit hole of identifying possible solutions to it, I found that none of the log messages within [`close_connections()`](https://github.com/juspay/hyperswitch/blob/1ab4226c780e9205785f012fd1c48c7a4bafb48f/crates/redis_interface/src/lib.rs#L157-L179) were ever printed. To address this, I temporarily added a `std::mem::forget(_guard)` line in `router/src/bin/router.rs` so that the logs would be available. 

This brought out the fact that all attempts by us to close Redis connections (the connection pool, publisher client and subscriber client) were failing with the following warning:

```text
Fatal error sending QUIT command to router. Client may be stopped or not yet initialized.
```

When I attached the debugger, set a debugger breakpoint at the first line of the `Drop` implementation and listed Redis clients using the `redis-cli` tool, there were no Redis clients arising from the application in the list, and the warning was indeed correct. 

It seems that Redis connections are being closed before even the `Drop` implementation has been entered, and I'm unable to explain why. I couldn't find anything relevant in the `fred` codebase either. I was able to reproduce this with a Redis cluster set up using Docker Compose.

SanchithHegde:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [x] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

This PR drops the `Drop` implementation (pun intended) on `RedisConnectionPool`.

## Motivation and Context

Fixes #1782.

## How did you test it?

Locally, with both a standalone Redis instance and a Redis cluster setup with Docker Compose.

## Checklist


crates/redis_interface/src/lib.rs
```diff
@@ -33,7 +33,7 @@ pub use self::{commands::*, types::*};
 pub struct RedisConnectionPool {
     pub pool: fred::pool::RedisPool,
     config: RedisConfig,
-    join_handles: Vec<fred::types::ConnectHandle>,
+    _join_handles: Vec<fred::types::ConnectHandle>,
```
NishantJoshi00, line 36: do we need to hold on to `ConnectHandle`, if we aren't planning to use it?

NishantJoshi00 (changes requested):

ashokkjag (approved):

PR #1799, opened 2023-07-27
fix: storage of generic payment methods in permanent locker

kashif-m:
## Type of Change

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description
Permanent locker requires either card details or encrypted card data to be stored in DB. In case both the fields are present during addition in locker, card details takes preference over the latter. This leads to the locker retrieval API returning null for the required enc_card_data field where generic payment method data is stored.

This is solved by segregating the data types for card addition API request in locker to send only card details or encrypted card data based on the usecase.

### Additional Changes

## How did you test it?
Tested in custom integ pod for customer `locker_bank_test`. Bank details for SEPA and BACS are stored for this customer post fulfillment of the payout transaction.

```
curl --location 'https://integ-api.hyperswitch.io/customers/locker_bank_test/payment_methods' \
--header 'x-feature: integ-custom'
```
NOTE: DM for api_key.

## Checklist

### DIFF
--- a/crates/router/src/core/payment_methods/cards.rs
+++ b/crates/router/src/core/payment_methods/cards.rs
@@ -264,7 +264,7 @@ pub async fn add_card_hs(
             card_exp_year: card.card_exp_year.to_owned(),
             card_brand: None,
             card_isin: None,
-            nick_name: card.nick_name.to_owned().map(masking::Secret::expose),
+            nick_name: card.nick_name.as_ref().map(masking::Secret::peek).cloned(),
         },
     });
     let store_card_payload = call_to_locker_hs(state, &payload, &customer_id).await?;
--- a/crates/router/src/core/payment_methods/cards.rs
+++ b/crates/router/src/core/payment_methods/cards.rs
@@ -13,8 +13,6 @@ use api_models::{
     },
     payments::BankCodeResponse,
 };
-#[cfg(feature = "payouts")]
-use common_utils::ext_traits::ByteSliceExt;
 use common_utils::{
     consts,
     ext_traits::{AsyncExt, StringExt, ValueExt},
@@ -256,8 +254,20 @@ pub async fn add_card_hs(
     customer_id: String,
     merchant_account: &domain::MerchantAccount,
 ) -> errors::CustomResult<(api::PaymentMethodResponse, bool), errors::VaultError> {
-    let store_card_payload =
-        call_to_card_hs(state, &card, None, &customer_id, merchant_account).await?;
+    let payload = payment_methods::StoreLockerReq::LockerCard(payment_methods::StoreCardReq {
+        merchant_id: &merchant_account.merchant_id,
+        merchant_customer_id: customer_id.to_owned(),
+        card: payment_methods::Card {
+            card_number: card.card_number.to_owned(),
+            name_on_card: card.card_holder_name.to_owned(),
+            card_exp_month: card.card_exp_month.to_owned(),
+            card_exp_year: card.card_exp_year.to_owned(),
+            card_brand: None,
+            card_isin: None,
+            nick_name: card.nick_name.to_owned().map(masking::Secret::expose),
+        },
+    });
+    let store_card_payload = call_to_locker_hs(state, &payload, &customer_id).await?;
 
     let payment_method_resp = payment_methods::mk_add_card_response_hs(
         card,
@@ -269,6 +279,28 @@ pub async fn add_card_hs(
         payment_method_resp,
         store_card_payload.duplicate.unwrap_or(false),
     ))
+}
+
+#[instrument(skip_all)]
+pub async fn decode_and_decrypt_locker_data(
+    key_store: &domain::MerchantKeyStore,
+    enc_card_data: String,
+) -> errors::CustomResult<Secret<String>, errors::VaultError> {
+    // Fetch key
+    let key = key_store.key.get_inner().peek();
+    // Decode
+    let decoded_bytes = hex::decode(&enc_card_data)
+        .into_report()
+        .change_context(errors::VaultError::ResponseDeserializationFailed)
+        .attach_printable("Failed to decode hex string into bytes")?;
+    // Decrypt
+    async { decrypt(Some(Encryption::new(decoded_bytes.into())), key).await }
+        .await
+        .change_context(errors::VaultError::FetchPaymentMethodFailed)?
+        .map_or(
+            Err(report!(errors::VaultError::FetchPaymentMethodFailed)),
+            |d| Ok(d.into_inner()),
+        )
 }
 
 #[instrument(skip_all)]
@@ -285,7 +317,7 @@ pub async fn get_payment_method_from_hs_locker<'a>(
     #[cfg(feature = "kms")]
     let jwekey = &state.kms_secrets;
 
-    if !locker.mock_locker {
+    let payment_method_data = if !locker.mock_locker {
         let request = payment_methods::mk_get_card_request_hs(
             jwekey,
             locker,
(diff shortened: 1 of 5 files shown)


crates/router/src/core/payment_methods/cards.rs
```diff
@@ -314,44 +346,34 @@ pub async fn get_payment_method_from_hs_locker<'a>(
             .payload
             .get_required_value("RetrieveCardRespPayload")
             .change_context(errors::VaultError::FetchPaymentMethodFailed)?;
-        retrieve_card_resp
+        let enc_card_data = retrieve_card_resp
             .enc_card_data
             .get_required_value("enc_card_data")
-            .change_context(errors::VaultError::FetchPaymentMethodFailed)
+            .change_context(errors::VaultError::FetchPaymentMethodFailed)?;
+        decode_and_decrypt_locker_data(key_store, enc_card_data.peek().to_string()).await?
```
NishantJoshi00, line 353: Why are we using the key store to decrypt this data?
The key store based encryption is used for database-domain layer encryption and not for cards. Can you provide me with a bit more insight on this?

crates/router/src/core/payment_methods/cards.rs
```diff
@@ -494,31 +516,47 @@ pub async fn delete_card_from_hs_locker<'a>(
...
         card_fingerprint: uuid::Uuid::new_v4().to_string(),
         card_global_fingerprint: uuid::Uuid::new_v4().to_string(),
-        merchant_id: "mm01".to_string(),
-        card_number: card.card_number.peek().to_string(),
-        card_exp_year: card.card_exp_year.peek().to_string(),
-        card_exp_month: card.card_exp_month.peek().to_string(),
+        merchant_id: "".to_string(),
+        card_number: "4111111111111111".to_string(),
+        card_exp_year: "2099".to_string(),
+        card_exp_month: "12".to_string(),
```
NishantJoshi00, line 535: can we have this back to the old values, this could cause some issues while performing card specific tests

crates/router/src/core/payment_methods/cards.rs
```diff
@@ -587,21 +625,7 @@ pub async fn mock_get_payment_method<'a>(
...
-            .attach_printable("Failed to decode hex string into bytes")?;
-        // Decrypt
-        async { decrypt(Some(Encryption::new(decoded_bytes.into())), key).await }
-            .await
-            .change_context(errors::VaultError::FetchPaymentMethodFailed)?
-            .map_or(
-                Err(report!(errors::VaultError::FetchPaymentMethodFailed)),
-                |d| Ok(d.into_inner()),
-            )
+        decode_and_decrypt_locker_data(key_store, e).await
```
NishantJoshi00, line 628: From what I know, there is no field in the locker mock up that undergoes domain level encryption

crates/router/src/core/payment_methods/cards.rs
```diff
@@ -1821,8 +1845,9 @@ pub async fn get_lookup_key_for_payout_method(
     .attach_printable("Error getting payment method from locker")?;
     let pm_parsed: api::PayoutMethodData = payment_method
         .peek()
-        .as_bytes()
-        .to_vec()
+        .to_string()
+        // .as_bytes()
+        // .to_vec()
```
NishantJoshi00, line 1850: remove comments

crates/router/src/core/payment_methods/transformers.rs
```diff
@@ -15,12 +15,25 @@ use crate::{
...
     pub merchant_customer_id: String,
     pub card: Card,
-    pub enc_card_data: Option<String>,
+}
+
+#[derive(Debug, Deserialize, Serialize)]
+pub struct StoreGenericReq<'a> {
+    pub merchant_id: &'a str,
+    pub merchant_customer_id: String,
+    pub enc_card_data: String,
```
NishantJoshi00, line 36: Will this be `enc_card_data`, we are planning to use this to store any type of payment method right? can we have a bit of a generic name

NishantJoshi00 (changes requested):

crates/router/src/core/payment_methods/cards.rs
kashif-m: It will be overriden here - https://github.com/juspay/hyperswitch/pull/1799/files#diff-0cad9ec56cd7135aea31f7f8fb51ee0197f2125ebc87b716e9d023318361a827R546 in case of cards

crates/router/src/core/payment_methods/cards.rs
kashif-m: This approach was decided earlier; out of scope for this change

cc: @jagan-jaya

NishantJoshi00 (approved):
Looks good, though requires sign-off from testing team! as locker needs to be tested in integ

crates/router/src/core/payment_methods/cards.rs
```diff
@@ -256,8 +254,20 @@ pub async fn add_card_hs(
...
+        merchant_id: &merchant_account.merchant_id,
+        merchant_customer_id: customer_id.to_owned(),
+        card: payment_methods::Card {
+            card_number: card.card_number.to_owned(),
+            name_on_card: card.card_holder_name.to_owned(),
+            card_exp_month: card.card_exp_month.to_owned(),
+            card_exp_year: card.card_exp_year.to_owned(),
+            card_brand: None,
+            card_isin: None,
+            nick_name: card.nick_name.to_owned().map(masking::Secret::expose),
```
pixincreate, line 267: How does this differ from `name_on_card`?  
I suggest you use `peek` from `peekinterface` instead of using `expose`

kashif-m: `name_on_card` refers to the `card_holder_name` which is different from `nick_name`

kashif-m: Why should we use `peek` instead of `expose`? I have a value which I'm mapping to expose, which returns the inner value.

If we were to use `peek`, the reference will be passed to the `peek` function, which would return a reference to the inner value, which would need to be dereferenced again

With use of `expose` we don't have to reference and de-reference it.

pixincreate: - By using `peek()` you're just passing a reference and not the complete value and this increases the security by not giving complete ownership like what `expose()` does
- By using `expose()`, there exist a possibility to get the values changed since it is different value and not the reference to its parent

pixincreate (dismissed):

pixincreate:
Please update the information regarding how did you test the changes. Presently provided information is not enough.  

LGTM!

crates/router/src/core/payment_methods/cards.rs
```diff
@@ -271,6 +281,28 @@ pub async fn add_card_hs(
...
+) -> errors::CustomResult<Secret<String>, errors::VaultError> {
+    // Fetch key
+    let key = key_store.key.get_inner().peek();
+    // Decode
+    let decoded_bytes = hex::decode(&enc_card_data)
+        .into_report()
+        .change_context(errors::VaultError::ResponseDeserializationFailed)
+        .attach_printable("Failed to decode hex string into bytes")?;
+    // Decrypt
+    async { decrypt(Some(Encryption::new(decoded_bytes.into())), key).await }
```
NishantJoshi00, line 297: wrapping this with `async { ... }` not required, it's already async and can be directly awaited on.

NishantJoshi00 (changes requested):
Other than that looks good!

jagan-jaya (approved):

NishantJoshi00 (approved):
LGTM

pixincreate (approved):

PR #1815, opened 2023-07-28
fix(webhook): provide acknowledgment for webhooks with unsupported event types

NishantJoshi00:
## Type of Change

- [x] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

The change allows acknowledgment of event types which aren't configured and allows the webhooks to pass through instead of sending an error to the connector.

### Additional Changes

## How did you test it?
compiler guided

## Checklist


SanchithHegde (approved):

sai-harsha-vardhan (changes requested):
@NishantJoshi00 @SanchithHegde 
Currently `ConnectorError::WebhookEventTypeNotFound` is being thrown by connectors when there's deserialization failure of complete webhook response in `get_webhook_event_type`, ignoring this and sending acknowledgement is risky! 
I would suggest the following
1. Instead of deserializing the entire response, deserialize only event type enum
2. We can rely on `ConnectorError::WebhookEventTypeNotFound` and`EventNotSupported` webhook event given by connector, in both these cases we can safely ignore the event and give acknowledgement to the connector

NishantJoshi00:
Hey @sai-harsha-vardhan, though it seems risky it is the only way we can solve certain issues that we are facing with webhooks that we receive on the hosted environments. But following are some of my insights about the issues that you have raised.
> 1. Instead of deserializing the entire response, deserialize only event type enum

I agree that performing a lightweight deserialization is the best bet to avoid other issue in the webhook body to affect me, but its only possible if the webhook integration wasn't done properly and tested thoroughly, and I trust that most of the structs that are defined for response body parsing are accurate to the documentation and tested thoroughly. If otherwise we need to test this flow and other auxiliary flows which depend on webhooks, the problem of not parsing the webhook will be least of our worries in that case.

> 2. We can rely on `WebhookEventTypeNotFound` and `EventNotSupported` webhook event given by connector, in both these cases we can safely ignore the event and give acknowledgement to the connector

Thanks for the insight, I'll also add `EventNotSupported` in the error handling logic that is implemented in the PR.

NishantJoshi00:
It seems that the use case of `EventNotSupported` is already handled in the `lookup_webhook_event`, so this won't be a issue it seems

sai-harsha-vardhan:
@NishantJoshi00
Should we add any metrics to recognize genuine deserialization errors in case of ConnectorError::WebhookEventTypeNotFound, or how are we going to identify if all the webhooks start failing due to deserialization, and we continue to send acknowledgments? This makes it seem like we are consuming webhooks and updating status, but in reality, we are just ignoring them!

NishantJoshi00:
Sure, that makes sense, I'll add a log line and a metric for this!

sai-harsha-vardhan (approved):

SanchithHegde (approved):

PR #1850, opened 2023-08-02
fix(webhooks): do not send duplicate webhooks 

Narayanbhat166:
## Type of Change

- [x] New feature

## Description

Hyperswitch triggers a webhook only when it receives a webhook. We will need to trigger a webhook for payments too, but in that case, if we receive a webhook from upstream, that will also trigger a webhook to the merchant. To avoid this, event id is generated using `primary_resource_id` and `event_type`. This will prevent the same event type webhook being repeated.

### Additional Changes

migrations/2023-08-01-165717_make_event_id_unique_for_events_table/down.sql
migrations/2023-08-01-165717_make_event_id_unique_for_events_table/up.sql

## How did you test it?

- Create a payment and test if webhooks are received in the webhook site.
Screenshot 2023-08-02 at 7 13 52 PM

## Checklist

### DIFF
--- a/crates/api_models/src/payments.rs
+++ b/crates/api_models/src/payments.rs
@@ -2415,7 +2415,7 @@ pub struct PaymentsCancelRequest {
     pub merchant_connector_details: Option<admin::MerchantConnectorDetailsWrap>,
 }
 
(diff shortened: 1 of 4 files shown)


NishantJoshi00 (approved):
Looks good!

jarnura (approved):

PR #1859, opened 2023-08-03
ci(postman): Use dynamic amount for Authorizedotnet, remove retries, add zen collection and disable globalpay

pixincreate:
## Type of Change

- [x] Bugfix
- [x] CI/CD

## Description

- Modified Authorize.net collection to use dynamic amount instead of adding backoff since this backoff seems to do more harm than good. Also, removed retries and disabled Globalpay from running given its inconsistencies.
- Added Zen Collection. It works under Static IP, hence added, it will fail here since the IP is not static, and hence the Zen collection is disabled.

## Motivation and Context

Fix

## How did you test it?

Ran the collection.

## Checklist


ashokkjag (approved):

SanchithHegde (dismissed):

jarnura (approved):

jarnura (changes requested):
Ask respective to dev to check the connector doc's, this will happen only in sandbox of their system?

SanchithHegde (approved):

jarnura (approved):

jarnura (approved):

SanchithHegde (approved):

ashokkjag (approved):

PR #1864, opened 2023-08-04
feat(docs): add multiple examples support and webhook schema

Narayanbhat166:
## Type of Change

- [x] New feature

## Description

Add webhooks schema and multiple examples support for openapi.

## Motivation and Context

Multiple examples would be needed to convey the right set of fields required to be sent for a particular kind of payment. This Pr adds support to add multiple examples by specifying the examples as a json string in the `openapi_consts` file in `common_utils` crate.

## How did you test it?

Generate the openapi file and check in any swagger ui availble online.

## Checklist


crates/api_models/src/webhooks.rs
```diff
@@ -81,20 +82,33 @@ pub struct IncomingWebhookDetails {
...
+    /// The type of event this webhook corresponds to.
+    #[schema(value_type = EventType)]
     pub event_type: api_enums::EventType,
+
+    /// This is specific to the flow, for ex: it will be `PaymentsResponse` for payments flow
     pub content: OutgoingWebhookContent,
     #[serde(default, with = "custom_serde::iso8601")]
+
+    /// The time at which webhook was sent
     pub timestamp: PrimitiveDateTime,
```
SanchithHegde, line 102: Don't we need a `serde(with = "...")`?

crates/common_utils/src/openapi_consts.rs
SanchithHegde, line 1: Can we move these to the `router` crate if it's the only crate accessing it? Maybe a module in the `openapi` module?

crates/api_models/src/webhooks.rs
Narayanbhat166: This would be a breaking change for users who are consuming this field, as we are changing the format in which timestamp is sent.

jarnura (approved):

SanchithHegde (approved):

SanchithHegde (approved):

jarnura (approved):

PR #1866, opened 2023-08-04
refactor(connector): use utility function to raise payment method not implemented errors

swangi-kumari:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description
Issue: defaults case in match shouldn't be handled using "_"

Solution: "_" will never let the developer know impact of his changes incase of new addition of enum variants. By having the all cases in match compiler itself will throw errors wherever new enums needs to be handled

(In this PR it is done for Braintree, Bambora, GlobalPay, Paypal, Payu, Mollie, Rapyd, Fiserv and Authorizedotnet Connectors)

### Additional Changes

## How did you test it?
Screen Shot 2023-08-04 at 1 38 50 PM 2
Screen Shot 2023-08-04 at 1 44 00 PM 2

## Checklist


ArjunKarthik (approved):

srujanchikke (approved):

crates/router/src/connector/authorizedotnet/transformers.rs
```diff
@@ -133,12 +133,21 @@ fn get_pm_and_subsequent_auth_detail(
...
+                | api::PaymentMethodData::BankDebit(_)
+                | api::PaymentMethodData::BankTransfer(_)
+                | api::PaymentMethodData::Crypto(_)
+                | api::PaymentMethodData::MandatePayment
+                | api::PaymentMethodData::Reward(_)
+                | api::PaymentMethodData::Upi(_)
+                | api::PaymentMethodData::Voucher(_)
+                | api::PaymentMethodData::GiftCard(_) => {
+                    Err(errors::ConnectorError::NotImplemented(
+                        utils::get_unimplemented_payment_method_error_message("authorizedotnet"),
```
SanchithHegde, line 148: Should we consider storing the connector name in a module level `const`, say `CONNECTOR_NAME`, so that all error messages or functions that need the connector name can use it instead?

Something like:

```rust
// This would go in the same file (`transformers.rs`), or in the connector-specific file (`authorizedotnet.rs`)
const CONNECTOR_NAME: &str = "authorizedotnet";

// Then the error propagation code would look like:
utils::get_unimplemented_payment_method_error_message(CONNECTOR_NAME)
```

I'm also fine with obtaining the string off of the `Display` implementation of an enum if that's available.

CC: @ArjunKarthik

SanchithHegde (approved):

crates/router/src/connector/authorizedotnet/transformers.rs
ArjunKarthik: @SanchithHegde Agree, That would be a neat solution

SanchithHegde (approved):

SanchithHegde (approved):

PR #1870, opened 2023-08-04
fix(connector): [ACI] Response Handling in case of `ErrorResponse`

Sakilmostak:
## Type of Change

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

In case of Error Response in ACI, it doesn't contain the `id` field, resulting in deserialization error.

## Checklist


crates/router/src/connector/aci.rs
```diff
@@ -425,10 +425,10 @@ impl
...
         res: types::Response,
     ) -> CustomResult<types::ErrorResponse, errors::ConnectorError> {
-        let response: aci::AciPaymentsResponse =
-            res.response
-                .parse_struct("AciPaymentsResponse")
-                .change_context(errors::ConnectorError::ResponseDeserializationFailed)?;
+        let response: aci::AciErrorResponse = res
+            .response
+            .parse_struct("AciErrorResponse")
+            .change_context(errors::ConnectorError::ResponseDeserializationFailed)?;
```
ArjunKarthik, line 431: We can move this to common method

ArjunKarthik (changes requested):
Please do the suggested changes

crates/router/src/connector/aci.rs
Sakilmostak: Done👍

ArjunKarthik (approved):

crates/router/src/connector/aci.rs
```diff
@@ -50,6 +50,29 @@ impl ConnectorCommon for Aci {
...
+        Ok(types::ErrorResponse {
+            status_code: res.status_code,
+            code: response.result.code,
+            message: response.result.description,
+            reason: response.result.parameter_errors.and_then(|errors| {
+                errors.first().map(|error_description| {
```
prasunna09, line 67: here, connector always have only one element or more than one?? if more than one element, can we concatenate all the errors and pass it in reason?

prasunna09 (changes requested):

crates/router/src/connector/aci.rs
Sakilmostak: Done👍

crates/router/src/connector/aci.rs
```diff
@@ -50,6 +50,33 @@ impl ConnectorCommon for Aci {
...
+                errors
+                    .into_iter()
+                    .map(|error_description| {
+                        format!(
+                            "Field is {} and the message is {}",
+                            error_description.name, error_description.message
+                        )
+                    })
+                    .collect::<Vec<String>>()
+                    .join("\n")
```
ArjunKarthik, line 76: lets join this with ;

ArjunKarthik (changes requested):

crates/router/src/connector/aci.rs
Sakilmostak: Done👍

ArjunKarthik (approved):

prasunna09 (approved):

PR #1898, opened 2023-08-08
feat(payments): add api locking for payments core

Abhicodes-crypto:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

This PR introduces an API locking framework and uses it with the core payments flow. For now both are under the `locking` feature flag. After thorough testing they can be moved under release feature. 
Problem statement: 
- When webhooks, sync, redirection happen almost at the same time there could be race conditions and inconsistency in the status of the payment. 
- What if capture and cancel happens at the same time.

API locking prevents that by locking the flow. 

- Another use case of API locking, while attempting to not send duplicate webhooks to merchant, the logic was to fetch a table from DB and then check a value if it is false send the webhook or else don't send it . But if some other request changes the value to true after the previous request fetches it. Here race condition could happen and multiple webhooks could still be done. So API locking can solve this.

TODO : 
- Add status validations.
- Speculate what action should be taken in case of validation fails. 
- Add logs.
- Add metrics
- Remove hard coded values and take config from Redis.
- What action to take when the retries are exhausted. 
- Refactor such that the get_key_and_lock_resource function returns only the status `Acquired` or gives error and only over acquired status `release_lock` is implemented.
- Code clean up

### Additional Changes

## How did you test it?

Monitored redis if the locks are getting created or not. 
image

Webhooks testing still in progress

## Checklist


crates/router/Cargo.toml
```diff
@@ -27,6 +27,7 @@ dummy_connector = ["api_models/dummy_connector"]
 external_access_dc = ["dummy_connector"]
 detailed_errors = ["api_models/detailed_errors", "error-stack/serde"]
 payouts = []
+locking = []
```
NishantJoshi00, line 30: can we have this explicitly `api_locking`? locking seems a bit vague

crates/router/src/core/locking.rs
```diff
@@ -0,0 +1,191 @@
...
+        true,
+        expiry_in_seconds,
+        delay_between_retries_in_seconds,
+        retries,
+    )
+    .await
+}
+
+#[inline]
+fn get_redis_key_for_locks(a: String) -> String {
```
NishantJoshi00, line 99: This seems like a unnecessary allocation for input, could either have `&str` or `impl AsRef<str>`

crates/router/src/core/locking.rs
```diff
@@ -0,0 +1,191 @@
...
+pub async fn release_lock(
+    state: &routes::AppState,
+    mut retries: i32,
+    lock: LockStatus,
+) -> RouterResult<LockStatus> {
+    let redis_conn = state.store.clone().get_redis_conn();
+
+    match lock {
+        LockStatus::Acquired(key) | LockStatus::AlreadyLocked(key) => {
+            while retries != 0 {
```
NishantJoshi00, line 159: This doesn't prevent negative cases, can we force `retries` variable to be `u32`?

crates/router/src/core/payments.rs
```diff
@@ -76,6 +78,16 @@ where
         .validate_request(&req, &merchant_account)?;
 
     tracing::Span::current().record("payment_id", &format!("{}", validate_result.payment_id));
+
+    #[cfg(feature = "locking")]
+    let acquired_lock = locking::get_key_and_lock_resource(
+        &state.clone(),
```
NishantJoshi00, line 84: Is this `.clone` required?

crates/api_models/src/payments.rs
```diff
@@ -1189,6 +1189,17 @@ impl std::fmt::Display for PaymentIdType {
     }
 }
 
+impl PaymentIdType {
+    pub fn get_id(&self) -> String {
+        match &self {
+            Self::PaymentIntentId(a)
+            | Self::ConnectorTransactionId(a)
+            | Self::PaymentAttemptId(a)
+            | Self::PreprocessingId(a) => a.to_owned(),
```
inventvenkat, line 1198: Why only PreprocessingId is owned?

crates/router/src/core/payments/operations.rs
```diff
@@ -75,6 +77,18 @@ pub struct ValidateResult<'a> {
     pub requeue: bool,
 }
 
+#[cfg(feature = "locking")]
+impl locking::GetLockingInput for ValidateResult<'_> {
+    fn get_locking_input(&self) -> RouterResult<locking::LockingInput> {
+        Ok(locking::LockingInput {
+            unique_locking_key: self.payment_id.get_id(),
+            api_identifier: "PAYMENTS".to_string(),
```
NishantJoshi00, line 85: will prefer if this is a constant, defined somewhere

NishantJoshi00 (changes requested):

crates/router/src/core/payments.rs
```diff
@@ -210,6 +222,17 @@ where
             .await?;
     }
 
+    #[cfg(feature = "locking")]
+    let _ = locking::release_lock(state, 3, acquired_lock.to_owned())
```
inventvenkat, line 226: Will it always reach here?, I mean, even in case of any errors inside the flow?. We should never fail to release the lock

inventvenkat: Let's add metrics in case of failure. And monitor.

crates/router/src/core/locking.rs
```diff
@@ -0,0 +1,191 @@
...
+            Ok(redis::SetnxReply::KeyNotSet) => {
+                logger::error!("Lock not acquired, retrying");
+                actix_time::sleep(tokio::time::Duration::from_millis(expiry_in_seconds * 1000))
+                    .await;
+            }
+            Err(error) => {
+                logger::error!(error=%error.current_context(), "Error while locking");
+                actix_time::sleep(tokio::time::Duration::from_millis(expiry_in_seconds * 1000))
+                    .await;
+            }
```
dracarys18, line 142: Can we have like a seperate interface for retry like
```rust
struct Retry;
impl Retry {
	fn perform(fun: F,retries: i32) 
	where
		F: FnOnce-> Result<T,E>
	{
	}
}
```
Or a macro since this could be used in multiple places

crates/router/src/core/locking.rs
```diff
@@ -0,0 +1,191 @@
...
+    key: &str,
+    value: bool,
+    expiry_in_seconds: u64,
+    _delay_between_retries_in_seconds: i64,
+    mut retries: i32,
+) -> RouterResult<LockStatus> {
+    let redis_conn = state.store.clone().get_redis_conn();
+
+    while retries != 0 {
+        retries -= 1;
```
dracarys18, line 115: Can't we use for loop?

crates/router/src/core/locking.rs
```diff
@@ -0,0 +1,191 @@
...
+pub enum ActionOnWaitTimeout {
+    Default,
+}
+
+#[derive(Debug)]
+pub struct LockingInput {
+    pub unique_locking_key: String,
+    pub api_identifier: String,
+    pub action_on_wait_timeout: ActionOnWaitTimeout,
+    pub merchant_id: String,
```
inventvenkat, line 37: Why do we need merchant_id here?

crates/router/src/core/locking.rs
```diff
@@ -0,0 +1,191 @@
...
+
+#[instrument(skip(state))]
+pub async fn get_key_and_lock_resource(
+    state: &routes::AppState,
+    locking_input: LockingInput,
+    _should_retry: bool,
+) -> RouterResult<LockStatus> {
+    let api_identifier = locking_input.api_identifier.as_str();
+
+    let is_locking_enabled_for_merchant = true;
```
inventvenkat, line 48: I don' think we will need this feature, merchant wise.

crates/router/src/core/locking.rs
Abhicodes-crypto: Yep. Good catch

crates/router/src/core/locking.rs
```diff
@@ -0,0 +1,191 @@
...
+            )
+            .await;
+
+        match is_lock_acquired {
+            Ok(redis::SetnxReply::KeySet) => {
+                return Ok(LockStatus::Acquired(key.to_owned()));
+            }
+            Ok(redis::SetnxReply::KeyNotSet) => {
+                logger::error!("Lock not acquired, retrying");
+                actix_time::sleep(tokio::time::Duration::from_millis(expiry_in_seconds * 1000))
```
inventvenkat, line 135: shouldn't you have used `delay_between_retries_in_seconds` instead of `expiry_in_seconds`?

crates/router/src/core/locking.rs
```diff
@@ -0,0 +1,191 @@
...
+                    .await;
+            }
+            Err(error) => {
+                logger::error!(error=%error.current_context(), "Error while locking");
+                actix_time::sleep(tokio::time::Duration::from_millis(expiry_in_seconds * 1000))
```
inventvenkat, line 140: Same here

dracarys18 (changes requested):

crates/router/src/core/locking.rs
```diff
@@ -0,0 +1,191 @@
...
+#[instrument(skip(state))]
+pub async fn release_lock(
+    state: &routes::AppState,
+    mut retries: i32,
+    lock: LockStatus,
+) -> RouterResult<LockStatus> {
+    let redis_conn = state.store.clone().get_redis_conn();
+
+    match lock {
+        LockStatus::Acquired(key) | LockStatus::AlreadyLocked(key) => {
```
inventvenkat, line 158: When do we get AlreadyLocked status during release_lock?

crates/router/src/core/payments/operations.rs
Abhicodes-crypto: Yep will remove all hardcodings. Had already added in TODO.

crates/router/src/core/locking.rs
inventvenkat: I also thinking, we should not depend on the status while releasing. We should release irrespective of the status

crates/router/src/core/locking.rs
inventvenkat, line 159: In case of failure of release after retries, we should insert into scheduler

crates/router/src/core/payments.rs
```diff
@@ -76,6 +78,16 @@ where
         .validate_request(&req, &merchant_account)?;
 
     tracing::Span::current().record("payment_id", &format!("{}", validate_result.payment_id));
+
+    #[cfg(feature = "locking")]
```
inventvenkat, line 82: Are we sure, all APIs go through this flow?

Abhicodes-crypto: Afaik , most of the payments api go through this flow. The payments webhooks , and redirection definitely go through this flow.

crates/router/src/core/locking.rs
Abhicodes-crypto: Yes thanks.

crates/router/src/core/locking.rs
Abhicodes-crypto: This will be used only here. No multiple place implementation.

crates/router/src/core/locking.rs
Abhicodes-crypto: Sure will remove it.

crates/router/src/core/locking.rs
Abhicodes-crypto: When we insert we insert with a time_to_expiry, so it is going to be released eventually.

crates/api_models/src/payments.rs
Abhicodes-crypto: No it is owned for all the above values as well

crates/router/src/core/locking.rs
dracarys18: Retry is a generic mechanism you can create an interface and in future if you wanna use it somewhere else you can use it

crates/router/src/core/payments.rs
Abhicodes-crypto: Even in case of error the payments flow completes the post_update_trackers so that the payment status should be in a conclusive state. But if we are to use in any other flows we have to make sure the flow completes . If not handle release in the parent function based on result being `Ok` or `Err`. 

> Let's add metrics in case of failure. And monitor.

Will be adding metrics in the next PR. added it in TODO.

crates/router/src/core/payments.rs
```diff
@@ -76,6 +78,13 @@ where
         .validate_request(&req, &merchant_account)?;
 
     tracing::Span::current().record("payment_id", &format!("{}", validate_result.payment_id));
+
+    #[cfg(feature = "api_locking")]
+    let acquired_lock =
```
jarnura, line 83: api locking should be in server wrap or kind of a middle-ware solution, the solution should not be in `payments_operation_core`.

crates/router/src/core/payments/operations.rs
```diff
@@ -75,6 +77,18 @@ pub struct ValidateResult<'a> {
     pub requeue: bool,
 }
 
+#[cfg(feature = "api_locking")]
+impl api_locking::GetLockingInput for ValidateResult<'_> {
+    fn get_locking_input(&self) -> RouterResult<api_locking::LockingInput> {
+        Ok(api_locking::LockingInput {
+            unique_locking_key: self.payment_id.get_id(),
```
jarnura, line 84: Only payment_id is the unique_locking_key and other's not.

crates/router/src/core/api_locking.rs
```diff
@@ -0,0 +1,190 @@
...
+use error_stack::{IntoReport, ResultExt};
+use redis_interface as redis;
+use router_env::{instrument, logger, tracing};
+
+use super::errors::{self, RouterResult};
+use crate::routes;
+
+#[derive(Debug)]
+pub enum Lock {
+    Acquired(String),
```
jarnura, line 11: Add comments about what are those strings.

crates/router/src/core/api_locking.rs
```diff
@@ -0,0 +1,190 @@
...
+    Release(String),
+}
+
+pub struct LockingError {}
+
+#[derive(Clone, Debug)]
+pub enum LockStatus {
```
jarnura, line 18: Add comments about what are those strings.

jarnura (changes requested):

crates/router/src/core/api_locking.rs
```diff
@@ -0,0 +1,190 @@
...
+                .await;
+            }
+        }
+    }
+
+    Ok(LockStatus::AcquireFailedRetriesExhausted(key.to_owned()))
+}
+
+#[instrument(skip(state))]
+pub async fn release_lock(
```
inventvenkat, line 149: change the status to just have acquired and released. And check requestId as authentication to release the lock

crates/router/src/core/api_locking.rs
```diff
@@ -0,0 +1,155 @@
...
+                    .change_context(errors::ApiErrorResponse::InternalServerError)?;
+
+                let redis_locking_key = input.get_redis_locking_key(merchant_id);
+                let delay_between_retries_in_milliseconds = state
+                    .conf()
+                    .lock_settings
+                    .delay_between_retries_in_milliseconds;
+                let redis_lock_expiry_seconds =
+                    state.conf().lock_settings.redis_lock_expiry_seconds;
+                let lock_retries: u32 = ((redis_lock_expiry_seconds * 1000)
```
Abhicodes-crypto, line 70: Will move this logic to the place where Settings is initialised and use check math functions here instead of direct operators.

jarnura (approved):

crates/router/src/core/api_locking.rs
```diff
@@ -0,0 +1,167 @@
...
+                    .change_context(errors::ApiErrorResponse::InternalServerError)?;
+
+                let redis_locking_key = input.get_redis_locking_key(merchant_id);
+                let delay_between_retries_in_milliseconds = state
+                    .conf()
+                    .lock_settings
+                    .delay_between_retries_in_milliseconds;
+                let redis_lock_expiry_seconds =
+                    state.conf().lock_settings.redis_lock_expiry_seconds;
+                let lock_retries = state.conf().lock_settings.lock_retries;
```
inventvenkat, line 70: There are 2 cases, 
1. Wait till the locked gets unlocked and execute (Add 1 to the no.of retries in current logic)
2. Don't wait for the entire time and throw error after a certain time period (Add env to get the time period).

inventvenkat (changes requested):

crates/router/src/core/api_locking.rs
```diff
@@ -0,0 +1,167 @@
...
+}
+
+#[derive(Clone, Debug)]
+pub enum LockAction {
+    // Sleep until the lock is acquired
+    Hold { input: LockingInput },
+    // Queue it but return response as 2xx, could be used for webhooks
+    QueueWithOk,
+    // Ignore it and return response as 2xx, could be used for webhooks
+    IgnoreWithOk,
```
inventvenkat, line 28: We should not have this IgnoreWithOk

inventvenkat (changes requested):

inventvenkat (approved):

inventvenkat (approved):

jarnura (approved):

dracarys18 (approved):

PR #1908, opened 2023-08-09
fix(router): replace the occurrences of `gen_range` with a safer alternative

### LINKED ISSUE #1754: [BUGFIX] Replacing the occurrences of `gen_range` with a safe alternative
## Description

Following up with the issue https://github.com/rust-random/rand/issues/1326 created by @lsampras. The function `gen_range` in itself can cause panics and error handling isn't done for some cases. This is a unpredictable behaviour and needs to be replaced with a safe alternative which allows for error handling and recoverability.

## Objective

- Deciding open a safe alternative for `gen_range` while keeping the behaviour relatively same
- Replacing the occurrences of `gen_range` with the decided safe alternative

## Guidelines

- Please mention the approach for solving the above mentioned issue and reasoning of it being recoverable.

anupj:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [x] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

This is a bugfix for issue [1754](https://github.com/juspay/hyperswitch/issues/1754).
I have replaced the instances of `gen_range()` with `Uniform.sample`.
This should reduce the risk of `gen_range()` throwing a `panic!`.

### Additional Changes

## Motivation and Context

This is a bugfix for issue [1754](https://github.com/juspay/hyperswitch/issues/1754)
Closes #1754.

## How did you test it?
I did not write any new tests but I ran `cargo test -p router --lib` and all tests passed.

NOTE: I was not able to run all integration tests because my laptop is not setup for running integration tests. 

## Checklist

### DIFF
--- a/crates/diesel_models/src/business_profile.rs
+++ b/crates/diesel_models/src/business_profile.rs
@@ -70,3 +70,51 @@ pub struct BusinessProfileUpdateInternal {
     pub payout_routing_algorithm: Option<serde_json::Value>,
     pub is_recon_enabled: Option<bool>,
 }
+
+impl From<BusinessProfileNew> for BusinessProfile {
+    fn from(new: BusinessProfileNew) -> Self {
+        Self {
+            profile_id: new.profile_id,
+            merchant_id: new.merchant_id,
+            profile_name: new.profile_name,
+            created_at: new.created_at,
+            modified_at: new.modified_at,
+            return_url: new.return_url,
+            enable_payment_response_hash: new.enable_payment_response_hash,
+            payment_response_hash_key: new.payment_response_hash_key,
+            redirect_to_merchant_with_http_post: new.redirect_to_merchant_with_http_post,
+            webhook_details: new.webhook_details,
+            metadata: new.metadata,
+            routing_algorithm: new.routing_algorithm,
+            intent_fulfillment_time: new.intent_fulfillment_time,
+            frm_routing_algorithm: new.frm_routing_algorithm,
+            payout_routing_algorithm: new.payout_routing_algorithm,
+            is_recon_enabled: new.is_recon_enabled,
+        }
+    }
+}
+
+impl BusinessProfileUpdateInternal {
+    pub fn apply_changeset(self, source: BusinessProfile) -> BusinessProfile {
+        BusinessProfile {
+            profile_name: self.profile_name.unwrap_or(source.profile_name),
+            modified_at: self.modified_at.unwrap_or(source.modified_at),
+            return_url: self.return_url,
+            enable_payment_response_hash: self
+                .enable_payment_response_hash
+                .unwrap_or(source.enable_payment_response_hash),
+            payment_response_hash_key: self.payment_response_hash_key,
+            redirect_to_merchant_with_http_post: self
+                .redirect_to_merchant_with_http_post
+                .unwrap_or(source.redirect_to_merchant_with_http_post),
+            webhook_details: self.webhook_details,
+            metadata: self.metadata,
+            routing_algorithm: self.routing_algorithm,
+            intent_fulfillment_time: self.intent_fulfillment_time,
+            frm_routing_algorithm: self.frm_routing_algorithm,
+            payout_routing_algorithm: self.payout_routing_algorithm,
+            is_recon_enabled: self.is_recon_enabled.unwrap_or(source.is_recon_enabled),
+            ..source
+        }
+    }
+}
--- a/crates/router/src/db/business_profile.rs
+++ b/crates/router/src/db/business_profile.rs
@@ -114,38 +114,90 @@ impl BusinessProfileInterface for Store {
 impl BusinessProfileInterface for MockDb {
     async fn insert_business_profile(
         &self,
-        _business_profile: business_profile::BusinessProfileNew,
+        business_profile: business_profile::BusinessProfileNew,
     ) -> CustomResult<business_profile::BusinessProfile, errors::StorageError> {
-        Err(errors::StorageError::MockDbError)?
+        let business_profile_insert = business_profile::BusinessProfile::from(business_profile);
+        self.business_profiles
+            .lock()
+            .await
+            .push(business_profile_insert.clone());
+        Ok(business_profile_insert)
     }
 
     async fn find_business_profile_by_profile_id(
         &self,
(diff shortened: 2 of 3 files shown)


crates/router/src/routes/dummy_connector/utils.rs
```diff
@@ -15,8 +15,11 @@ use crate::{configs::settings, routes::AppState};
 
 pub async fn tokio_mock_sleep(delay: u64, tolerance: u64) {
     let mut rng = rand::thread_rng();
-    let effective_delay = rng.gen_range((delay - tolerance)..(delay + tolerance));
-    tokio::sleep(tokio::Duration::from_millis(effective_delay)).await
+    let effective_delay = Uniform::from((delay - tolerance)..(delay + tolerance));
```
lsampras, line 18: can we use `Uniform::try_from` ?.

(although that also does not seem to help)
https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=4a99d8a5b9a9604292cf7065e83e2866

Maybe it'll be fixed in future, we can use the safer API on our side in the meanwhile

crates/router/src/scheduler/consumer.rs
```diff
@@ -42,10 +42,11 @@ pub async fn start_consumer(
 ) -> CustomResult<(), errors::ProcessTrackerError> {
     use std::time::Duration;
 
-    use rand::Rng;
+    use rand::distributions::{Distribution, Uniform};
 
-    let timeout = rand::thread_rng().gen_range(0..=settings.loop_interval);
-    tokio::time::sleep(Duration::from_millis(timeout)).await;
+    let mut rng = rand::thread_rng();
+    let timeout = Uniform::from(0..=settings.loop_interval);
```
lsampras, line 48: `Uniform::try_from`

lsampras (changes requested):

crates/router/src/routes/dummy_connector/utils.rs
anupj: @lsampras  I started with `try_from`:
`    let timeout = Uniform::try_from(0..=settings.loop_interval)?;`
But the compiler complains that:

```bash
the trait `std::convert::From<Infallible>` is not implemented for `error_stack::Report<ProcessTrackerError>`
```

We can't just implement `From<Infallible>` for `error_stack::Report<ProcessTrackerError>` because of Rusts orphan rule. If we want to get around this then we may have to add significant amount of workaround in the `errors.rs` file potentially changing the API contract. 
Plus as you can see above it says to implement something that is Infallible that is unlikely to fail so we are not gaining much here. And another reason is that the `from` method just does type conversion here which is infallible; the actual random number work is done by the `sample` method which again doesn't return a `Result`.

Does that reasoning make sense?

anupj: @lsampras @SanchithHegde any thoughts on the above comment?

lsampras:
```
let timeout = Uniform::try_from(0..=settings.loop_interval).into_report().change_context(ProcessTrackerError::InvalidInput)?;
```
I believe this should fix it^

yeah the result is currently infallible but in the future it's not gonna be,
we won't have much benefit from adding it rn but we'll be benefitting from this when a new version of rand is released

anupj: @lsampras cool, thanks. That is good to know. I'll resubmit the PR.

anupj:
@SanchithHegde I've updated the PR, let me know if there are any other changes required here. Thanks.

crates/router/src/routes/dummy_connector/utils.rs
SanchithHegde: I believe this still needs to be changed to `Uniform::try_from()`?

SanchithHegde (commented):

Thanks for the PR, @anupj!

crates/router/src/scheduler/producer.rs
```diff
@@ -22,9 +22,13 @@ pub async fn start_producer(
...
 ) -> CustomResult<(), errors::ProcessTrackerError> {
-    use rand::Rng;
-    let timeout = rand::thread_rng().gen_range(0..=scheduler_settings.loop_interval);
-    tokio::time::sleep(std::time::Duration::from_millis(timeout)).await;
+    use rand::distributions::{Distribution, Uniform};
+
+    let mut rng = rand::thread_rng();
+    let timeout = Uniform::try_from(0..=scheduler_settings.loop_interval)
+        .into_report()
+        .change_context(errors::ProcessTrackerError::InvalidInput)?;
```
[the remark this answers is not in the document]
anupj: Is this PR blocked because of this?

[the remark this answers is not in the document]
SanchithHegde: @anupj Not exactly blocked by the error propagation, there's still one remaining usage of `Uniform::from()` that would have to be replaced.

https://github.com/juspay/hyperswitch/pull/1908#discussion_r1296186863

Let us know if you need any help.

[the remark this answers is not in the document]
anupj: Ok, I'll fix this today/tomorrow and update PR for review. Thanks.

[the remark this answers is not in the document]
anupj: @SanchithHegde I remember now why I didn't change this to try_from() because it will mean that we'll have to change the function signature for tokio_mock_sleep() to return a Result which could turn out to be a bigger change than is required for this PR as we haave to then change the fn signature where tokioo_mock_sleep() is called. Is that desirable? ?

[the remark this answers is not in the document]
lsampras: Can you then add a TODO here & if possible create a new issue for this that can be picked up later?

SanchithHegde:
@anupj Could you please address the merge conflicts?

anupj:
> @anupj Could you please address the merge conflicts?

Oh dear, looks like I might have accidentally closed the PR. Is it better to open a new PR to address the merge issues?

SanchithHegde:
> Oh dear, looks like I might have accidentally closed the PR. Is it better to open a new PR to address the merge issues?

Either is fine.

PR #1919, opened 2023-08-11
feat(api_client): add api client trait

harsh-sharma-juspay:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

Currently we use a shared client via once cell to prevent the CPU overhead associated with creating a new `reqwest::Client`
Instead I'd propose to use a trait that can be used for defining this behavior associated with making API calls while re-using the reqwest::Client

There are 2 distinct components for this design
1. ApiClient (to be shared throughout the app & holds the underlying tcp sockets/connections etc)
2. RequestBuilder (Generated each time an request is to be made holds info specific to that request)

Both of the above components are represented as a trait ^

```rs
pub trait RequestBuilder: Send + Sync {
    /// I'm constrained to not using generics anywhere in this trait as to make it still possible to build a vtable for dyn dispatch
    /// fn json<T: Serialize>(&mut self, body: T); won't work since I won't be able to build a dyn dispatch trait out of this

    /// fn json<T: Serialize>(&mut self, body: &T); Although using reference of the generic might work
    /// ^^^will get to this once I verify that we don't need ownership of this struct
    fn json(&mut self, body: serde_json::Value);
    fn url_encoded_form(&mut self, body: serde_json::Value);
    fn timeout(&mut self, timeout: Duration);
    fn multipart(&mut self, form: Form);
    fn header(&mut self, key: String, value: Maskable<String>) -> CustomResult<(), ApiClientError>;
    /// Have avoided the usage of async_trait crate here for personal preference
    fn send(
        self,
    ) -> CustomResult<Box<(dyn core::future::Future<Output = Result<reqwest::Response, reqwest::Error>> + 'static)>, ApiClientError>;
}

pub trait ApiClient
where
    Self: Sized + Send + Sync,
{
    fn new(
        proxy_config: Proxy,
        whitelisted_urls: Vec<String>,
    ) -> CustomResult<Self, ApiClientError>;
    /// Returning a Box<dyn> here to avoid any limitations introduced by generics (mostly ergonomics)
    /// can be re-considered once the integration is done
    fn request<U: IntoUrl>(&self, method: Method, url: U) -> Box<dyn RequestBuilder>;
}
```

This can be used to implement a mock `ApiClient` as well to test connectors.

### Additional Changes

## Motivation and Context

There are a couple of existing limitations that can be solved via this:
1. limited control over initialization of the client since the initialization info needs to be passed at all points of usage
2. reduced ergonomics & abstraction since the proxy config, whitelisted URL's, certificate info needs to be passed everywhere
3. Limitations in terms of the logging & metrics we can offer since we don't have a control/abstraction over how requests are built
4. Currently unit tests can't be run on more than 1 thread due to lazy static initialization failure on multiple threads simultaneously, now we can have a separate reqwest::Client for each thread (Or also build a mock for it)

## How did you test it?

- This only covers the initial spec for defining the trait, matching the spec for existing ApiClient & requesting comments on the overall trait definitions & functionality
- Integration for this would be taken up separately

## Checklist


crates/router/src/services/api/client.rs
```diff
@@ -110,3 +116,145 @@ pub(super) fn proxy_bypass_urls(locker: &Locker) -> Vec<String> {
...
+        Ok(Self {
+            proxy_client,
+            non_proxy_client,
+            whitelisted_urls,
+        })
+    }
+
+    fn request<U: IntoUrl>(&self, method: Method, url: U) -> Box<dyn RequestBuilder> {
+        Box::new(RouterRequestBuilder {
+            inner: Some(self.proxy_client.request(method, url)),
```
lsampras, line 256:
```suggestion
            inner: Some(self.get_reqwest_client(...)),
```

Also add the below  methods to `trait ApiClient` & use them in this function with the default values as None

```
pub fn request_with_certificate(&self, certificate: Option<String>, certificate_key: Option<String>);
```

lsampras (changes requested):

crates/router/src/services/api/client.rs
```diff
@@ -152,20 +158,31 @@ impl ProxyClient {
...
+                        "Failed to construct client with certificate and certificate key",
+                    )?)
+            }
+            (_, _) => {
                 if self.whitelisted_urls.contains(&base_url) {
-                    self.non_proxy_client.clone()
+                    Ok(self.non_proxy_client.clone())
                 } else {
-                    self.proxy_client.clone()
+                    Ok(self.proxy_client.clone())
```
lsampras, line 183: if a  client certificate is present we should build a proxied client (unless whitelisted URL's)
you can store the proxy config in the `ProxyClient` struct for this case

lsampras (changes requested):

lsampras (dismissed):

lsampras (approved):

SanchithHegde (approved):

PR #1925, opened 2023-08-11
fix(frm): add new column frm_config instead of alterning the existing…

jagan-jaya:
## Type of Change

- [X] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

We added a change for frm_config to support array in MCA with the below changes 
```UPDATE merchant_connector_account set frm_configs = null ;

ALTER TABLE merchant_connector_account 
ALTER COLUMN frm_configs TYPE jsonb[]
USING ARRAY[frm_configs]::jsonb[];

UPDATE merchant_connector_account set frm_configs = null ;
```
This will result in production downtime as we are touching the mca. 

So this PR will fix the downtime by adding a new column
### Additional Changes

## Motivation and Context

NA

## How did you test it?

Ran sanity tests manually in local

## Checklist


inventvenkat (dismissed):

migrations/2023-08-11-073905_add_frm_config_in_mca/up.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE "merchant_connector_account" ADD COLUMN frm_config jsonb[];
+ALTER TABLE "merchant_connector_account" DROP COLUMN IF EXISTS frm_configs;
```
jarnura, line 2: The running application asks for the column and the column is removed means that will break the current application which is running in environments.

So we can't drop a existing column which a application schema uses it.

jarnura (changes requested):

jarnura (approved):

inventvenkat (approved):

jagan-jaya:
This will require the sanity testing for a normal payment. As it is a breaking change this will affect all the payment flows and frm flows. so if the normal payment works frm will also work.

PR #1935, opened 2023-08-15
feat(router): add total count for refunds list

apoorvdixit88:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [x] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description
Added support for total count of available refunds in refunds list. This will support better pagination and also work well with filters applied on refunds list.

### Additional Changes

## Motivation and Context
The current refund list api doesn't have support for proper pagination.

## How did you test it?
Screenshot 2023-08-16 at 1 58 53 AM

## Checklist


crates/router/src/core/refunds.rs
```diff
@@ -665,9 +665,19 @@ pub async fn refund_list(
...
         .collect();
 
+    let total_count = db
+        .get_total_count_of_refunds(
+            &merchant_account.merchant_id,
+            &req,
+            merchant_account.storage_scheme,
+        )
+        .await
+        .to_not_found_response(errors::ApiErrorResponse::InternalServerError)?;
```
inventvenkat, line 675: Shouldn't this be 0 instead of error response?

crates/api_models/src/refunds.rs
```diff
@@ -160,7 +162,9 @@ pub struct TimeRange {
 #[derive(Debug, Clone, Eq, PartialEq, Deserialize, Serialize, ToSchema)]
 pub struct RefundListResponse {
     /// The number of refunds included in the list
-    pub size: usize,
+    pub count: usize,
```
inventvenkat, line 165: This is not backward compatible.

inventvenkat (changes requested):

crates/router/src/core/refunds.rs
apoorvdixit88: count 0 case is handled, error can occur in cases, such as DB is down or some unexpected behaviour.

crates/api_models/src/refunds.rs
apoorvdixit88: For now it is not being used

inventvenkat (approved):

PR #1936, opened 2023-08-16
fix(core): added is_terminated field in fraud_check table to support 3ds flow

rishavkar:
## Type of Change

- [X] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

Currently, frm module dosent work for 3ds flow.
It is because, payments operation core is called multiple times in the same request, and so is pre and post frm.
This creates duplicate entries, and also creates discrepencies.
To fix this, we are adding a field is_terminated in the fraud check table. This is to indicate that the frm flow has ended, and no furthur frm calls are to be made for the payment(be that 2nd 3ds call, or psync etc)

### Additional Changes

- [X] This PR modifies the database schema

- is_terminated is added to table fraud_check.

## How did you test it?

- do a 3ds flow, with frm.
- for both pre and post flow, at the end of the flow, the is_terminated field should be set to true in the fraud_check table.
- for pre flow, signifyd should be called twice(once before and once after transaction), not 4 times.
- for post flow, it should be called only hitting the redirection link, and completing the authentication, i.e., only after the transaction is complete,i.e., frm should be called just once, not twice.

## Checklist

### DIFF
--- a/crates/diesel_models/src/fraud_check.rs
+++ b/crates/diesel_models/src/fraud_check.rs
@@ -24,6 +24,7 @@ pub struct FraudCheck {
     pub payment_details: Option<serde_json::Value>,
     pub metadata: Option<serde_json::Value>,
     pub modified_at: PrimitiveDateTime,
+    pub is_terminated: bool,
 }
 
 #[derive(router_derive::Setter, Clone, Debug, Insertable, router_derive::DebugAsDisplay)]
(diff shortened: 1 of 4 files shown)


jagan-jaya:
Closing in favour of #1944

PR #1946, opened 2023-08-16
feat(storage_impl): split payment intent interface implementation

lsampras:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [x] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

- split payment intent interface declaration from it's implementation
- Currently the data_models crate defines the `PaymentIntentInterface` which is devoid of any diesel related functionality
- For now the struct representations are identical, however would be changed in the future where data_model structs start to model the business use cases & invariants (getting more nested) whereas the ones in storage_impl would represent the table structure & be rather flat
- Have added a simple trait called `DataModelExt` to set the mapping between a data model & its storage model, for now this is public since router uses both of there representations but it would be private to the crate once all storage specific implementations are removed from router
- As part of this also added a `StorageError` in data_models which defines the `PaymentIntentInterface`, maybe instead of having a data_models crate we should club common_utils, common_enums to create some sort of router_core or data_models can be treated as the replacement for storage_models... (but more representative of the domain use cases rather than api models)
- Add a common struct for filtering payments

- **Payments List API now uses read replica always (unless configured with a non replicated database i.e otlp only)**

### Additional Changes

## Motivation and Context

- eventually splitting the implementation constraints from domain structs
- currently router does not depend on `diesel_models::payment_intent::PaymentIntent` and this would be extented to other structs where the implementation is abstracted from the domain model

## How did you test it?

- followed the quick start api's in postman for payment creation/retrieve/update/list

## Checklist


crates/data_models/Cargo.toml
```diff
@@ -0,0 +1,26 @@
+[package]
+name = "data_models"
+version = "0.1.0"
+edition = "2021"
```
SanchithHegde, line 4: Please include more of the manifest fields, can refer to any of the other crates.

crates/router/src/core/errors.rs
```diff
@@ -80,6 +81,67 @@ pub enum StorageError {
...
+                storage_errors::DatabaseError::NotFound => {
+                    DataStorageError::ValueNotFound(String::from("db value not found"))
+                }
+                // TODO: Update this error type to encompass & propagate the duplicate type (instead of generic `db value not found`)
+                storage_errors::DatabaseError::UniqueViolation => {
+                    DataStorageError::DuplicateValue {
+                        entity: "db entity",
+                        key: None,
+                    }
+                }
```
SanchithHegde, line 108: Would this be done now, or as part of a later change?

crates/router/src/core/errors.rs
```diff
@@ -80,6 +81,67 @@ pub enum StorageError {
...
+            StorageError::MockDbError => DataStorageError::MockDbError,
+            StorageError::CustomerRedacted => DataStorageError::CustomerRedacted,
+            StorageError::DeserializationFailed => DataStorageError::DeserializationFailed,
+            StorageError::EncryptionError => DataStorageError::EncryptionError,
+            StorageError::DecryptionError => DataStorageError::DecryptionError,
+            StorageError::RedisError(i) => match i.current_context() {
+                // TODO: Update this error type to encompass & propagate the missing type (instead of generic `redis value not found`)
+                RedisError::NotFound => {
+                    DataStorageError::ValueNotFound("redis value not found".to_string())
+                }
```
SanchithHegde, line 136: Same here

crates/router/src/core/errors/utils.rs
```diff
@@ -10,6 +10,43 @@ pub trait StorageErrorExt<T, E> {
...
+    }
+
+    #[track_caller]
+    fn to_duplicate_response(
+        self,
+        duplicate_response: errors::ApiErrorResponse,
+    ) -> error_stack::Result<T, errors::ApiErrorResponse> {
+        self.map_err(|err| {
+            let new_err = match err.current_context() {
+                data_models::errors::StorageError::DuplicateValue { entity: _, key: _ } => {
```
SanchithHegde, line 40:
```suggestion
                data_models::errors::StorageError::DuplicateValue { .. } => {
```

crates/router/src/core/payments/operations/payment_status.rs
```diff
@@ -397,7 +397,13 @@ pub async fn get_payment_intent_payment_attempt(
     merchant_id: &str,
     storage_scheme: enums::MerchantStorageScheme,
 ) -> RouterResult<(storage::PaymentIntent, storage::PaymentAttempt)> {
-    (|| async {
+    let le: error_stack::Result<
```
SanchithHegde, line 400: Could we use a better variable name here (unless I'm missing out something obvious about what `le` stands for)?

SanchithHegde (commented):

crates/router/src/core/errors.rs
lsampras: later since this involves changes to the more commonly used Database Error affecting all storage trait implementations

crates/router/src/core/payments/operations/payment_status.rs
lsampras: removed this as no longer needed

SanchithHegde (approved):

SanchithHegde (approved):

PR #1947, opened 2023-08-16
fix(payment_methods): return parent_payment_method_token for other payment methods (BankTransfer, Wallet, BankRedirect)

ShankarSinghC:
## Type of Change

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

During the `make_pm_data` we construct a key (`pm_token_{token}_{payment_method}_hyperswitch`) using the token that is received in the request and try to get the value store in redis for that key. If there is no value present for that particular key then we were just returning the token we got thinking it is a locker (V3) token not the parent token. There was a pr (#1929) in which parent token was being generated even while doing a payment using the new card, so now in the `make_pm_data` the code to return the received token if its not parent token was removed thinking we will always receive parent token. The similar change was not done for the payment methods like  BankTransfer, Wallet, BankRedirect because of which it was still returning the locker (V3) token and it was failing during the fetching the value for it from redis as it was not the parent token. 

This pr includes the change to return the parent token for the above mentioned payment methods as well.
### Additional Changes

## How did you test it?

Cards
image_720
Wallets
image_720-1
Bank_transfer
image_720-2
Bank_redirect
image_720-3
image_720-4

## Checklist

### DIFF
--- a/crates/router/src/core/payments/helpers.rs
+++ b/crates/router/src/core/payments/helpers.rs
@@ -27,7 +27,7 @@ use crate::{
         payments,
     },
     db::StorageInterface,
-    routes::{metrics, payment_methods::ParentPaymentMethodToken, AppState},
+    routes::{metrics, payment_methods, AppState},
     scheduler::metrics as scheduler_metrics,
     services,
     types::{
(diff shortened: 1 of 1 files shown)


jarnura (approved):

crates/router/src/core/payments/helpers.rs
```diff
@@ -1360,6 +1342,36 @@ pub async fn make_pm_data<'a, F: Clone, R>(
...
+    let router_token = vault::Vault::store_payment_method_data_in_locker(
+        state,
+        None,
+        payment_method_data,
+        payment_intent.customer_id.to_owned(),
+        payment_method,
+    )
+    .await?;
+    let parent_payment_method_token = generate_id(consts::ID_LENGTH, "token");
+    let key_for_hyperswitch_token = payment_attempt.payment_method.map(|payment_method| {
```
Sangamesh26, line 1361: Please check why `key_for_hyperswitch_token` is optional?
We are about to store in vault, so we should always get the key.

So, we have access to pmd, try getting pm from it and then generate the key

Sangamesh26 (changes requested):

Sangamesh26 (approved):
LGTM :)

PR #1970, opened 2023-08-21
fix(core): Update Webhooks Event Mapping and Forced Psync preconditions

SamraatBansal:
## Type of Change

- [X] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

Update Webhooks Event Mapping to avoid getting 5XX errors and handle all the variant cases
Update Preconditions for Forced Sync call to connector

### Additional Changes

## Motivation and Context

Webhooks are throwing 5XX if an unsupported Event is encountered, Also its not triggering a PSync if source is not verified due to preconditions on the status

## Checklist


crates/router/src/compatibility/stripe/webhooks.rs
```diff
@@ -142,6 +142,7 @@ fn get_stripe_event_type(event_type: api_models::enums::EventType) -> &'static s
         api_models::enums::EventType::DisputeChallenged => "dispute.challenged",
         api_models::enums::EventType::DisputeWon => "dispute.won",
         api_models::enums::EventType::DisputeLost => "dispute.lost",
+        api_models::enums::EventType::UnsupportedEvent => "unsupported.event",
```
vspecky, line 145: Minor nitpick: should be `event.unsupported`

crates/router/src/core/webhooks.rs
```diff
@@ -385,24 +383,27 @@ pub async fn disputes_incoming_webhook_flow<W: types::OutgoingWebhookType>(
...
+                merchant_account,
+                event_type,
+                enums::EventClass::Disputes,
+                None,
+                dispute_object.dispute_id,
+                enums::EventObjectType::DisputeDetails,
+                api::OutgoingWebhookContent::DisputeDetails(disputes_response),
+            )
+            .await?;
+            metrics::INCOMING_DISPUTE_WEBHOOK_MERCHANT_NOTIFIED_METRIC.add(
```
vspecky, line 401: Add metrics to also catch hyperswitch-unsupported events. We can take this up later. For now, please create an issue.

vspecky (approved):
Other than that, LGTM

crates/common_enums/src/enums.rs
```diff
@@ -757,6 +757,7 @@ pub enum EventType {
     DisputeChallenged,
     DisputeWon,
     DisputeLost,
+    UnsupportedEvent,
```
sai-harsha-vardhan, line 760: Instead of doing this, can we throw some other error like `UnprocessableEntity` - 422 instead of `InternalServerError`

sai-harsha-vardhan (changes requested):

crates/router/src/core/payments/helpers.rs
```diff
@@ -1648,7 +1648,7 @@ pub fn check_force_psync_precondition(
             | storage_enums::AttemptStatus::Authorized
             | storage_enums::AttemptStatus::Started
             | storage_enums::AttemptStatus::Failure
-    ) && connector_transaction_id.is_some()
+    ) || connector_transaction_id.is_some()
```
sai-harsha-vardhan, line 1651: This would lead to unnecessary connector calls, in case of terminal statuses too

NishantJoshi00: This is not the location that we need to change the condition. This needs to be left unchanged

The or logic needs to be present here:
`crates/router/src/core/payments/operations/payment_status.rs:340:33`

crates/router/src/core/webhooks.rs
```diff
@@ -457,24 +458,22 @@ async fn bank_transfer_webhook_flow<W: types::OutgoingWebhookType>(
...
-                payment_id,
-                enums::EventObjectType::PaymentDetails,
-                api::OutgoingWebhookContent::PaymentDetails(payments_response),
-            )
-            .await?;
+            let event_type: enums::EventType = payments_response.status.foreign_into();
+
+            // If event is NOT an UnsupportedEvent, trigger Outgoing Webhook
+            if !matches!(event_type, enums::EventType::UnsupportedEvent) {
+                create_event_and_trigger_outgoing_webhook::<W>(
```
NishantJoshi00, line 465: also, can't we move this logic inside the function
`create_event_and_trigger_outgoing_webhook`.

That would reduce replication and be a little fault tolerant.
what do you think?

NishantJoshi00 (changes requested):

sai-harsha-vardhan (approved):

crates/router/src/core/webhooks.rs
SamraatBansal: For all the Webhooks flows this common function is being called, it might affect the flexibility of the function, Currently not updating it, if required can be taken up in refactoring

vspecky (approved):

NishantJoshi00 (dismissed):

sai-harsha-vardhan (approved):

Narayanbhat166 (approved):

sai-harsha-vardhan (approved):

SanchithHegde (approved):

PR #1978, opened 2023-08-22
fix(locker): fix locker to throw error only when no card details are retrieved

Aprabhat19:
## Type of Change

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

We fix the locker such that when and if we are unable to fetch any cards , then only we would throw a 500 else we would show the details of the cards that we were able to retrieve from the locker.

### Additional Changes

## Motivation and Context
Previously, the locker was throwing a 500 , if it wasn't able to fetch even a single card from the locker .

## Checklist

### DIFF
--- a/crates/router/src/core/payment_methods/cards.rs
+++ b/crates/router/src/core/payment_methods/cards.rs
@@ -37,14 +37,14 @@ use crate::{
         },
         payments::helpers,
     },
-    db, logger,
+    db,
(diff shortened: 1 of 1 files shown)


Aprabhat19:
Not required

Issue #217, opened 2022-12-22
[FEATURE] Schedule webhook for retry


dracarys18:
## Description

Refactor webhooks core to add retry logic in webhooks core.

Currently, we store information about whether we have sent a webhook or not, improving on this we can implement retry logic to make this more reliable

trkohler:
hello! I would try to cover this feature, if possible.

PR #2000, opened 2023-08-23
[WIP] Schedule webhook for retry

trkohler:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description
Webhook request should be fired again if retry count is left. Time difference between requests is set to 5 seconds.

### Additional Changes

- ❌ This PR modifies the API contract 
- ❌ This PR modifies the database schema
- ❌ This PR modifies application configuration/environment variables

`crates/router/src/core/webhooks.rs`

## Motivation and Context
resolve https://github.com/juspay/hyperswitch/issues/217

## How did you test it?
[current status] I have difficulties with testing it locally. 

## Checklist

### DIFF
--- a/crates/router/src/core/webhooks.rs
+++ b/crates/router/src/core/webhooks.rs
@@ -28,6 +28,8 @@ use crate::{
 
 const OUTGOING_WEBHOOK_TIMEOUT_SECS: u64 = 5;
 const MERCHANT_ID: &str = "merchant_id";
+const WEBHOOK_RETRIES: u8 = 3;
+const WEBHOOK_RETRY_INTERVAL_SECS: u64 = 5;
 
 #[instrument(skip_all)]
 pub async fn payments_incoming_webhook_flow<W: types::OutgoingWebhookType>(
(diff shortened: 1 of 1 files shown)


crates/router/src/core/webhooks.rs
```diff
@@ -634,33 +636,31 @@ pub async fn trigger_webhook_to_merchant<W: types::OutgoingWebhookType>(
...
 
-    let request = services::RequestBuilder::new()
-        .method(services::Method::Post)
-        .url(&webhook_url)
-        .attach_default_headers()
-        .headers(header)
-        .body(Some(transformed_outgoing_webhook_string))
-        .build();
+    let mut retries = WEBHOOK_RETRIES;
+    loop {
```
NishantJoshi00, line 640: Hey @trkohler, Thanks for drafting a PR. 

Though the solution for a webhook is a bit more nuanced and we could follow a bit different approach towards solving it. Let's break it down.
- [x] Write a solution to perform retries for failed webhook.
- [x] Have a defined limit on how many retires are allowed.

🎉 you have already solved for the above mentioned problems.
Now, let's understand and address some API performance and future reliability issues.
- [ ] Having a loop clause could result in a infinite loop not now (Of course!), but in the future when someone else contributes and makes some change here
- [ ] Let's consider the scenario that the merchant API might be failing a lot, this means that we are holding the connection with the *connector* (stripe, adyen, etc) for a longer duration of time, this could result in a timeout from their side.
- [ ] Finally, we are adding extra processing load on the API which should in theory be very lightweight.

**Let's look at some ways to solve these**
1. We should remove this from the critical path if the first outbound webhook fails. (How?)
2. To tackle this we can use the scheduler service that's already present. (check `crates/router/src/scheduler/workflows.rs`)
3. While having retry count and retry delay as constants, it would provide more control if we can have these as some kind of configs. (check `crates/router/src/db/configs.rs`)

> Overall the goal is to move intensive and time consuming stuff out of the critical path, and execute it somewhere else, (in case of initial failures)

trkohler:
I'm sorry to take so much time with my PR. I can't currently put much focus and attention to is because of other life priorities.

PR #2016, opened 2023-08-25
fix(request): add `idle_pool_connection_timeout` as a config

NishantJoshi00:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

Adding a configuration to change the idle pool connection timeout

### Additional Changes

## Checklist


crates/router/src/services/api/client.rs
```diff
@@ -22,7 +22,13 @@ fn get_client_builder(
...
     should_bypass_proxy: bool,
 ) -> CustomResult<reqwest::ClientBuilder, ApiClientError> {
-    let mut client_builder = reqwest::Client::builder().redirect(reqwest::redirect::Policy::none());
+    let mut client_builder = reqwest::Client::builder()
+        .redirect(reqwest::redirect::Policy::none())
+        .pool_idle_timeout(
+            proxy_config
+                .idle_pool_connection_timeout
+                .map(std::time::Duration::from_secs),
+        );
```
lsampras, line 31: only apply this function if the value is provided,

if you call it with None, it will set no timeout

lsampras (approved):

crates/router/src/services/api/client.rs
```diff
@@ -22,7 +22,11 @@ fn get_client_builder(
     proxy_config: &Proxy,
     should_bypass_proxy: bool,
 ) -> CustomResult<reqwest::ClientBuilder, ApiClientError> {
-    let mut client_builder = reqwest::Client::builder().redirect(reqwest::redirect::Policy::none());
+    let mut client_builder = reqwest::Client::builder()
+        .redirect(reqwest::redirect::Policy::none())
+        .pool_idle_timeout(std::time::Duration::from_secs(
+            proxy_config.idle_pool_connection_timeout.unwrap_or(90),
```
SanchithHegde, line 28: Can we instead use the `Default` trait implementation for this purpose? Using a number with `unwrap_or()` can be harder to track defaults like this.

NishantJoshi00: sure

SanchithHegde (approved):

inventvenkat (approved):

PR #2020, opened 2023-08-25
refactor(core): use profile id to find connector

Narayanbhat166:
## Type of Change

- [x] Refactoring

## Description

This PR will make the necessary changes to use `profile_id` to query the connector in payments, refunds, disputes and payouts flow. 
For a merchant with single business_profile there would be no difference, for merchants with more than one business profile, `profile_id` will have to be sent in the payments request. 

## Backwards Compatibility

In order to keep supporting the existing merchants to migrate to new change, passing business country and business label is still supported. This will be deprecated soon and only profile_id will be accepted.

When a merchant account is created by passing `primary_business_details` for ex
```json
"primary_business_details": [
      {
          "country": "US",
          "business": "food"
      }
  ]
```
- A business profile is created with the `profile_name` as `US_food`. If there is only one primary_business_detail, then the `default_profile` field will be populated. 
- If there are more than one `business_labels`, then that many business profiles are created, but no `default_profile` will be populated.

### Support for existing merchants
In order to keep supporting existing merchants who have already created more than one business profile, a backfill will be done and business profiles will be created with the `profile_name` as the `business_label`. Now in case `business_label` is passed, `profile_id` is retrieved and then it will be used everywhere.

### Additional Changes

- [x] This PR modifies the API contract.
- [x] This PR modifies database schema. 

## Motivation and Context

To move towards using `profile_id`.

## How did you test it?

- Create a payment, default profile id should be used to query the connector.
Screenshot 2023-08-25 at 4 16 07 PM

## Checklist


jarnura (approved):

migrations/2023-08-28-131238_make_business_details_optional/up.sql
```diff
@@ -0,0 +1,21 @@
...
+ALTER COLUMN business_label DROP NOT NULL;
+
+ALTER TABLE merchant_connector_account
+ALTER COLUMN connector_label DROP NOT NULL;
+
+DROP INDEX IF EXISTS merchant_connector_account_merchant_id_connector_label_index;
+
+CREATE UNIQUE INDEX IF NOT EXISTS merchant_connector_account_profile_id_connector_id_index ON merchant_connector_account(profile_id, connector_name);
+
+CREATE UNIQUE INDEX IF NOT EXISTS business_profile_merchant_id_profile_name_index ON business_profile(merchant_id, profile_name);
```
jarnura, line 21: we need to have profile_id alone as unique.

jarnura (changes requested):

jarnura (dismissed):

ashokkjag (approved):

jarnura (dismissed):

jarnura (approved):

bernard-eugine (approved):

PR #2043, opened 2023-08-29
feat(connector): add support for multiple captures PSync

hrithikesh026:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

Currently Checkout.com doesn't support Psync for multiple captures.

There are two ways to sync captures:
1. `Bulk` (sync all captures in a single api request) (using parent `connector_transaction_id`)
2. `Individual` (sync individual captures one by one) (using `connector_capture_id`)

Checkout supports `Bulk`

In this PR, when force_sync if true and payment is not in terminal state, we will retrieve the list of actions for the parent payment from checkout and update the status of captures in DB. Which in turn will update payment attempt and intent status in DB.

### Additional Changes

## How did you test it?

Manual

Before PSync:

payment_intent:
checkout intent before

payment_attempt
checkout attempt before

captures
checkout captures before

After PSync:

payment_intent
checkout intent after

payment_attempt
checkout attempt after

captures
checkout captures after

## Checklist


jarnura (dismissed):

jarnura (dismissed):

ArjunKarthik (changes requested):

crates/router/src/connector/checkout/transformers.rs
```diff
@@ -882,7 +996,11 @@ pub struct CheckoutWebhookData {
     pub id: String,
     pub payment_id: Option<String>,
     pub action_id: Option<String>,
+    pub reference: Option<String>,
     pub amount: i32,
+    pub balances: Option<Balances>,
```
ArjunKarthik, line 1001: Was this added because of try_from for PaymentsResponse?

ArjunKarthik (changes requested):

crates/router/src/connector/checkout/transformers.rs
hrithikesh026: yes

crates/router/src/connector/checkout/transformers.rs
```diff
@@ -340,6 +340,32 @@ pub enum CheckoutPaymentStatus {
     Captured,
 }
 
+impl From<CheckoutWebhookEventType> for CheckoutPaymentStatus {
+    fn from(value: CheckoutWebhookEventType) -> Self {
```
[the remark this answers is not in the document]
hrithikesh026: which connector error should I return in case non payment events?

crates/router/src/connector/checkout.rs
```diff
@@ -1185,12 +1213,22 @@ impl api::IncomingWebhook for Checkout {
...
-        let details: checkout::CheckoutWebhookObjectResource = request
+        let event_type_data: checkout::CheckoutWebhookEventTypeBody = request
             .body
-            .parse_struct("CheckoutWebhookObjectResource")
-            .change_context(errors::ConnectorError::WebhookResourceObjectNotFound)?;
-
-        Ok(details.data)
+            .parse_struct("CheckoutWebhookBody")
+            .change_context(errors::ConnectorError::WebhookEventTypeNotFound)?;
+        let resource_object = if checkout::is_chargeback_event(&event_type_data.transaction_type) {
```
ArjunKarthik, line 1220: Only in case of payments event it should be payment response. In other cases we should return the object as it is

ArjunKarthik (changes requested):

jarnura (approved):

Issue #2061, opened 2023-08-31
[FEATURE] Function for triggering an auto-refund


kritikmodi:
### Feature Description

We cannot use the `refund_create_core` function for issuing an auto-refund since it uses the latest payment attempt. Instead, we need a separate function for processing auto-refunds.

### Possible Implementation

Add a new function in `core/refunds.rs`.

### Have you spent some time checking if this feature request has been raised before?

- [X] I checked and didn't find a similar issue

### Have you read the Contributing Guidelines?

- [X] I have read the [Contributing Guidelines](https://github.com/juspay/hyperswitch/blob/main/docs/CONTRIBUTING.md)

### Are you willing to submit a PR?

Yes, I am willing to submit a PR!

PR #2063, opened 2023-08-31
feat(refunds): add function for processing auto-refunds

kritikmodi:
## Type of Change

- [x] New feature

## Description

This PR adds a function for processing auto-refunds.

## Motivation and Context

This PR closes https://github.com/juspay/hyperswitch/issues/2061.

## How did you test it?

Not tested.

## Checklist

### DIFF
--- a/crates/router/src/core/refunds.rs
+++ b/crates/router/src/core/refunds.rs
@@ -115,6 +115,79 @@ pub async fn refund_create_core(
 }
 
 #[instrument(skip_all)]
+pub async fn auto_refund_core(
+    state: &AppState,
+    merchant_account: domain::MerchantAccount,
+    payment_attempt: storage::PaymentAttempt,
+    key_store: domain::MerchantKeyStore,
+    req: refunds::RefundRequest,
+) -> RouterResponse<refunds::RefundResponse> {
+    
+    let db = &*state.store;
+
+    let (merchant_id, payment_intent, amount);
+
+    merchant_id = &merchant_account.merchant_id;
+
+    payment_intent = db
+        .find_payment_intent_by_payment_id_merchant_id(
+            &req.payment_id,
+            merchant_id,
+            merchant_account.storage_scheme,
+        )
+        .await
+        .to_not_found_response(errors::ApiErrorResponse::PaymentNotFound)?;
+
+    utils::when(
+        payment_intent.status != enums::IntentStatus::Succeeded,
+        || {
+            Err(report!(errors::ApiErrorResponse::PaymentNotSucceeded)
+                .attach_printable("unable to refund for a unsuccessful payment intent"))
+        },
+    )?;
+
+    amount = req.amount.unwrap_or(
+        payment_intent
+            .amount_captured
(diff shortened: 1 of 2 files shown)


crates/router/src/core/refunds.rs
```diff
@@ -114,6 +114,77 @@ pub async fn refund_create_core(
...
+    key_store: domain::MerchantKeyStore,
+    req: refunds::RefundRequest,
+) -> RouterResponse<refunds::RefundResponse> {
+    let db = &*state.store;
+
+    let (merchant_id, payment_intent, amount);
+
+    merchant_id = &merchant_account.merchant_id;
+
+    payment_intent = db
```
Abhicodes-crypto, line 131: Why do we need payment_intent , can auto refund be performed with attempt alone, for that particular attempt.

dracarys18:
Most of the things look the same as `refund_core` why do we need seperate function for this.
You can either take payment _attempt as 
`Option<PaymentAttempt>` or create an enum like
```rust
 enum RefundType {
 Manual,
 Auto(PaymentAttempt)
 }
```
in refund_core

crates/router/src/core/refunds.rs
kritikmodi: No, we need `payment_intent` as well; otherwise we'll need to write a separate flow for this case.

crates/router/src/core/refunds.rs
```diff
@@ -114,6 +114,69 @@ pub async fn refund_create_core(
...
+        .async_map(|mcd| async {
+            payments::helpers::insert_merchant_connector_creds_to_config(
+                db,
+                merchant_id.as_str(),
+                mcd,
+            )
+            .await
+        })
+        .await
+        .transpose()?;
```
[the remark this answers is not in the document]
kritikmodi: Okay.

kritikmodi:
Postman test for this function  : 
Screenshot 2023-09-01 at 5 09 33 PM

crates/router/src/core/refunds.rs
```diff
@@ -114,6 +114,59 @@ pub async fn refund_create_core(
...
+        &payment_attempt,
+        &payment_intent,
+        amount,
+        req,
+        None,
+    )
+    .await
+    .map(services::ApplicationResponse::Json)
+}
+
```
dracarys18, line 169: I suggest you to re-use `refund_core`. If you want to change the flow for both Auto and Manual refunds you can use enum_dispatch. Something like,

```rust 
trait Operation {
      fn do_stff(self)
}

enum RefundOperation {
   Auto(PaymentAttempt),
   Manual
}

impl Operation for RefundOperation {
   fn do_stuff(self) {
     match self {
        match Self::Auto(payment_attempt)=> {},
        match Self::Manual=> {}
     }
  }
}

fn refund_core(operation: impl Operation) {
  operation.do_stuff();
}
```

dracarys18 (changes requested):

crates/router/src/core/refunds.rs
kritikmodi: Okay, I'll refactor this.

PR #2098, opened 2023-09-07
feat(payments): make database calls parallel for `payments_confirm` operation

NishantJoshi00:
…irm operations

## Type of Change

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

The PR introduces concurrency to the database calls limited only to the payments confirm operation.

### Additional Changes

## How did you test it?
compiler guided

## Checklist

### DIFF
--- a/crates/router/src/core/payments/operations/payment_confirm.rs
+++ b/crates/router/src/core/payments/operations/payment_confirm.rs
@@ -4,6 +4,7 @@ use api_models::enums::FrmSuggestion;
 use async_trait::async_trait;
 use common_utils::ext_traits::{AsyncExt, Encode};
 use error_stack::ResultExt;
+use futures::FutureExt;
 use router_derive::PaymentOperation;
 use router_env::{instrument, tracing};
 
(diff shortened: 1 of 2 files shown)


crates/router/src/core/payments/operations/payment_confirm.rs
```diff
@@ -50,16 +51,27 @@ impl<F: Send + Clone> GetTracker<F, PaymentData<F>, api::PaymentsRequest> for Pa
...
+
+        let mandate_details_fut = helpers::get_token_pm_type_mandate_details(
+            state,
+            request,
+            mandate_type.clone(),
+            merchant_account,
+        );
+
+        let (mut payment_intent, mandate_details) =
+            futures::try_join!(payment_intent_fut, mandate_details_fut)?;
```
Narayanbhat166, line 74: `helpers::get_token_pm_type_mandate_details`. This function internally does two database calls, and when joined with payment_intent db call, which has only one db call, to get the result the max time would be two db call.

Can you try doing a tokio::spawn for `get_token_pm_type_mandate_details`?

Narayanbhat166 (changes requested):

crates/router/src/core/payments/operations/payment_confirm.rs
NishantJoshi00: Instead internally, we could make the calls concurrent. This we create a 3 node tree for resolution.

NishantJoshi00: also, doing tokio spawn will make it so that, if any failure occurs in it, it will not affect the flow, which shouldn't be the case

NishantJoshi00: seems like its not possible to internally parallelize this, as the calls are hierarchically

Narayanbhat166: You can catch the error by awaiting on the join handle. These are just read queries, so if there is any error we can catch it there.

NishantJoshi00: If we await the join handle, then that's exactly equivalent to what's happening here

Narayanbhat166: What I'm telling is to await only for that join handle and not club it with payment_intent fetch

Narayanbhat166 (dismissed):

jarnura (approved):

Narayanbhat166 (approved):

PR #2140, opened 2023-09-12
ci(postman): Modify server start logic to address timeouts in actions

pixincreate:
## Type of Change

- [x] Refactoring
- [x] CI/CD

## Description

This PR aims at refactoring the server start logic of `postman-runner.yml` file to address timeout issue that killed the server if the server did not start within 15 minutes.
For context: 
image

The above failure is because of the timeout that occurred. 

## Motivation and Context

With this PR going in, we'll be building first and then starting the server next in the background in 2 different steps. This way, if the build fails, the check should anyway fail. If the server does not start within the next 2 minutes, fail the check again.

## How did you test it?

Now that it consumes less than 10 minutes to run, we've reduced the time consumption by 7+ minutes. 
image

## Checklist

### DIFF
--- a/.github/workflows/postman-collection-runner.yml
+++ b/.github/workflows/postman-collection-runner.yml
@@ -118,7 +118,7 @@ jobs:
           # Wait for the server to start in port 8080
           COUNT=0
           while ! nc -z localhost 8080; do
-            if [ $COUNT -gt 12 ]; then # Wait for up to 2 minutes (12 * 10 seconds)
+            if [ $COUNT -gt 90 ]; then # Wait for up to 2 minutes (90 * 10 seconds)
               echo "Server did not start within a reasonable time. Exiting."
               kill $SERVER_PID
               exit 1
--- a/.github/workflows/postman-collection-runner.yml
+++ b/.github/workflows/postman-collection-runner.yml
@@ -25,7 +25,7 @@ jobs:
 
     services:
       redis:
-        image: 'redis'
+        image: "redis"
         options: >-
           --health-cmd "redis-cli ping"
           --health-interval 10s
@@ -34,7 +34,7 @@ jobs:
         ports:
           - 6379:6379
       postgres:
-        image: 'postgres:14.5'
+        image: "postgres:14.5"
         env:
           POSTGRES_USER: db_user
           POSTGRES_PASSWORD: db_pass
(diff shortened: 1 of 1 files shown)


.github/workflows/postman-collection-runner.yml
```diff
@@ -97,23 +97,37 @@ jobs:
         env:
           DATABASE_URL: postgres://db_user:db_pass@localhost:5432/hyperswitch_db
         run: diesel migration run
-
+      
+      - name: Build project
+        if: ${{ ((github.event_name == 'pull_request') && (github.event.pull_request.head.repo.full_name == github.event.pull_request.base.repo.full_name)) || (github.event_name == 'merge_group')}}
+        run: cargo build
```
SanchithHegde, line 103: We can be specific here about the package and binary required to be built.

```suggestion
        run: cargo build --package router --bin router
```

.github/workflows/postman-collection-runner.yml
```diff
@@ -97,23 +97,37 @@ jobs:
...
+  
       - name: Setup Local Server
         if: ${{ ((github.event_name == 'pull_request') && (github.event.pull_request.head.repo.full_name == github.event.pull_request.base.repo.full_name)) || (github.event_name == 'merge_group')}}
         run: |
+          # Check if the build succeeded
+          if [ $? -ne 0 ]; then
+            echo "Cargo build failed. Exiting."
+            exit 1
+          fi
```
SanchithHegde, line 112: This is not necessary, since the workflow would stop if any previous step fails.

.github/workflows/postman-collection-runner.yml
```diff
@@ -97,23 +97,37 @@ jobs:
...
+
+          # Start the server in the background
           cargo run &
```
SanchithHegde, line 115:
```suggestion
          target/debug/router &
```

.github/workflows/postman-collection-runner.yml
```diff
@@ -97,23 +97,37 @@ jobs:
...
-              if [ $COUNT -gt 90 ];
-              then
-                  exit 1
-              else 
-                  COUNT=$((COUNT+1))
-                  sleep 10
-              fi
+          COUNT=0
+          while ! nc -z localhost 8080; do
+            if [ $COUNT -gt 90 ]; then # Wait for up to 2 minutes (90 * 10 seconds)
```
SanchithHegde, line 121:
```suggestion
            if [ $COUNT -gt 12 ]; then # Wait for up to 2 minutes (12 * 10 seconds)
```

SanchithHegde (commented):

SanchithHegde (approved):

PR #2146, opened 2023-09-13
fix: payment status fix in trustpay for 3ds and wallets

Sangamesh26:
## Type of Change

- [x] Bugfix

## Description

  We don't get payment_status only in case, when the user doesn't complete the authentication step. So, the status should be mapped to Authentication pending
  If we receive status, then return the proper status based on the connector response

## Motivation and Context

Payment status fix

## Checklist

### DIFF
--- a/crates/router/src/connector/trustpay/transformers.rs
+++ b/crates/router/src/connector/trustpay/transformers.rs
@@ -435,24 +435,30 @@ fn get_pending_status_based_on_redirect_url(redirect_url: Option<Url>) -> enums:
 }
 
 fn get_transaction_status(
-    payment_status: &str,
+    payment_status: Option<String>,
     redirect_url: Option<Url>,
 ) -> CustomResult<(enums::AttemptStatus, Option<String>), errors::ConnectorError> {
-    let (is_failed, failure_message) = is_payment_failed(payment_status);
-    let pending_status = get_pending_status_based_on_redirect_url(redirect_url);
-    if payment_status == "000.200.000" {
-        return Ok((pending_status, None));
+    // We don't get payment_status only in case, when the user doesn't complete the authentication step.
+    // If we recieve status, then return the proper status based on the connector response
+    if let Some(payment_status) = payment_status {
+        let (is_failed, failure_message) = is_payment_failed(&payment_status);
+        if is_failed {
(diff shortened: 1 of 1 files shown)


crates/router/src/connector/trustpay/transformers.rs
```diff
@@ -435,24 +435,30 @@ fn get_pending_status_based_on_redirect_url(redirect_url: Option<Url>) -> enums:
...
+                enums::AttemptStatus::AuthorizationFailed,
+                Some(failure_message.to_string()),
+            ));
+        }
+
+        if is_payment_successful(&payment_status)? {
+            return Ok((enums::AttemptStatus::Charged, None));
+        }
+
+        let pending_status = get_pending_status_based_on_redirect_url(redirect_url);
```
ArjunKarthik, line 456: @sai-harsha-vardhan If redirect_url is None we are setting payment status as Authorizing rather than Pending ?

crates/router/src/connector/trustpay/transformers.rs
```diff
@@ -435,24 +435,30 @@ fn get_pending_status_based_on_redirect_url(redirect_url: Option<Url>) -> enums:
...
-    }
-    if is_payment_successful(payment_status)? {
-        return Ok((enums::AttemptStatus::Charged, None));
+    // We don't get payment_status only in case, when the user doesn't complete the authentication step.
+    // If we receive status, then return the proper status based on the connector response
+    if let Some(payment_status) = payment_status {
+        let (is_failed, failure_message) = is_payment_failed(&payment_status);
+        if is_failed {
+            return Ok((
+                enums::AttemptStatus::AuthorizationFailed,
```
ArjunKarthik, line 447: Change status to Failed

crates/router/src/connector/trustpay/transformers.rs
ArjunKarthik: @Sangamesh26 get_pending_status_based_on_redirect_url should return pending incase redirect_url is None

ArjunKarthik (changes requested):

ArjunKarthik (approved):

sai-harsha-vardhan (approved):

PR #2148, opened 2023-09-13
fix: handle 5xx during multiple capture call

hrithikesh026:
## Type of Change

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

In case connector returns 5xx during capture call, we wouldn't have connector_capture_id updated in captures table. But capture may be created at connector. Thus capture sync wouldn't be possible without connector_capture_id. 

In this PR, if such things happen, we will try to map the extra unmapped capture list returned by the connector to captures at our end based on connector_capture_reference_id(our capture_id) or amount captured.

### Additional Changes

## How did you test it?

Manual. 
manually set connector_capture_id and connector_capture_reference_id to null and changed capture status to pending in DB and did force_sync.

Before PSync
Screenshot 2023-09-13 at 3 29 41 PM

After PSync
Screenshot 2023-09-13 at 3 29 48 PM

## Checklist


crates/router/src/core/payments/operations/payment_response.rs
```diff
@@ -621,16 +621,54 @@ fn response_to_capture_update(
...
+        .into_iter()
+        .cloned()
+        .collect();
+    for capture_sync_response in unmapped_capture_sync_response_list {
+        if let Some(capture) = captures_without_connector_capture_id
+            .iter()
+            .find(|capture| {
+                capture_sync_response.get_connector_response_reference_id()
+                    == Some(capture.capture_id.clone())
+                    || capture_sync_response.get_amount_captured() == Some(capture.amount)
```
Narayanbhat166, line 660: what if two captures have the same amount?

Narayanbhat166 (approved):

jarnura (approved):
LGTM.

crates/router/src/connector/globalpay/transformers.rs
```diff
@@ -506,6 +506,12 @@ impl utils::MultipleCaptureSyncResponse for GlobalpayPaymentsResponse {
         true
     }
 
+    fn get_amount_captured(&self) -> Option<i64> {
+        match self.amount.clone() {
+            Some(amount) => amount.parse().ok(),
```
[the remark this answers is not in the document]
hrithikesh026: Yes. amount field in globalpay response body is a string. We only try to convert it into int64.

crates/router/src/core/payments/operations/payment_response.rs
hrithikesh026: Then we map such captures in any order, wouldn't be an issue since, amount will be same.

Issue #1573, opened 2023-06-30
[FEATURE] Use OS specific profile path in selenium tests


jagan-jaya:
### Feature Description

Currently the default profile path for browser in selenium.rs file is hardcoded with mac specific file path. it should be made OS  specific instead of hardcoding.

### Possible Implementation

Add a feature for OS and handle the paths in selenium.rs

### Have you spent some time to check if this feature request has been raised before?

- [X] I checked and didn't find similar issue

### Have you read the Contributing Guidelines?

- [X] I have read the [Contributing Guidelines](https://github.com/juspay/hyperswitch/blob/main/docs/CONTRIBUTING.md)

### Are you willing to submit a PR?

None

PR #2178, opened 2023-09-17
ci(ui_test): Refactor core of UI tests to use injected browser data

pixincreate:
## Type of Change

- [x] Enhancement
- [x] CI/CD

## Description

This PR aims at automating Googlepay tests in UI Tests for connectors: Adyen_UK, Bluesnap and Stripe.  
Main problem with Google Pay UI automated testing is Captchas.  
Google detects automated bots / scripts when you try to login and throws in a captcha the moment `Next` button is pressed after entering in the `email_id` by elongated the content with Captcha (It is in fact hard for even humans to understand the text written on it). 
This hides the `Passwd` element and hence the UI test fails as it cannot login to the Google account anymore by timing out.  

Now, we're injecting the browser data (that contains the login data) into the browser and hence it does not ask for login now. So, no captcha now (Yes, I cannot guarantee, if tomorrow Google throws in captcha in other places). 

## Motivation and Context

Reduces manual testing burden.

## How did you test it?

See Connector UI tests below in CI checks

## Checklist


.github/workflows/connector-ui-sanity-tests.yml
```diff
@@ -129,23 +130,46 @@ jobs:
...
-      - name: View test results
+      - name: Start server
         if: ${{ (github.event_name == 'pull_request') && (github.event.pull_request.head.repo.full_name == github.event.pull_request.base.repo.full_name) }}
+        env:
+          BROWSER_DATA_PASSPHRASE: ${{ secrets.CONNECTOR_AUTH_PASSPHRASE }}
+          INPUT: ${{ matrix.connector }}
+          UI_TESTCASES_PATH: ${{ secrets.UI_TESTCASES_PATH }}
         shell: bash
-        run: cat tests/test_results.log
+        run: .github/scripts/run_ui_tests.sh
```
SanchithHegde, line 144: If the script only starts the server for running UI tests but doesn't actually run the tests, we'll have to name the script suitably.

.gitignore
```diff
@@ -261,3 +261,5 @@ result*
 
 # node_modules
 node_modules/
+
+connector_auth.toml
```
SanchithHegde, line 265: Do you think prefixing a `**` would help? It would prevent a file with name `connector_auth.toml` in any directory (including the root) from being committed.

```suggestion
**/connector_auth.toml
```

crates/test_utils/tests/connectors/selenium.rs
```diff
@@ -847,7 +877,12 @@ fn get_firefox_profile_path() -> Result<String, WebDriverError> {
             fp.join(&MAIN_SEPARATOR.to_string())
         })
         .unwrap();
-    base_path.push_str(r#"/Library/Application Support/Firefox/Profiles/hs-test"#); //Issue: 1573
+    if env::consts::OS == "macos" {
+        base_path.push_str(r#"/Library/Application Support/Firefox/Profiles/hs-test"#);
+    //Issue: 1573
+    } else if env::consts::OS == "linux" {
+        return Ok(r#"/home/runner/.mozilla/firefox/hs-test"#.to_string());
+    }
```
SanchithHegde, line 885: 1. You can mark that this PR closes #1573.
2. The path and Firefox profile name is very specific to GitHub Actions runner. Can we substitute `/home/runner` by the value of `${HOME}` environment variable?
3. For setting the name of the profile for GitHub Actions, we can make use of the `CI` or `GITHUB_ACTIONS` environment variables (both will be `true`). Reference: https://docs.github.com/en/actions/learn-github-actions/variables#default-environment-variables

.github/workflows/connector-ui-sanity-tests.yml
pixincreate: I actually wanted to get rid of this instead of just renaming the file. What do you say?

.github/scripts/run_ui_tests.sh
```diff
@@ -1,10 +1,11 @@
...
+#! /usr/bin/env bash
+
+sudo apt-get update
+sudo apt-get install net-tools
 mkdir tests
 
 COUNT=0
-#download connector ui tests
+# Download connector ui tests
 while [ ! -f $HOME/target/test/connector_tests.json ]
```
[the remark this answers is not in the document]
pixincreate: I won't need to install net-tools and update in this case, right?

.gitignore
```diff
@@ -261,3 +261,5 @@ result*
 
 # node_modules
 node_modules/
+
+**/connector_auth.toml
```
jagan-jaya, line 265: add sample_auth also

crates/test_utils/tests/connectors/selenium.rs
```diff
@@ -398,9 +423,9 @@ pub trait SeleniumTest {
             Event::Trigger(Trigger::Goto(url)),
             Event::Trigger(Trigger::Click(By::Css(".gpay-button"))),
             Event::Trigger(Trigger::SwitchTab(Position::Next)),
-            Event::Trigger(Trigger::Sleep(5)),
+            Event::Trigger(Trigger::Sleep(10)),
```
jagan-jaya, line 426: revert this change, keep it 5

jagan-jaya (changes requested):

crates/test_utils/tests/connectors/selenium.rs
pixincreate: Do you think it is worth randomizing?

jagan-jaya (approved):

SanchithHegde (dismissed):

jagan-jaya (dismissed):

SanchithHegde (approved):

jagan-jaya (approved):

PR #2190, opened 2023-09-20
refactor(core): eliminate business profile database queries in payments confirm flow

BallaNitesh:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [x] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description
Removing db calls which are unnecessary

### Additional Changes

## Checklist

### DIFF
--- a/crates/router/src/core/utils.rs
+++ b/crates/router/src/core/utils.rs
@@ -211,6 +211,7 @@ pub async fn construct_refund_router_data<'a, F>(
         merchant_account,
         payment_intent.profile_id.as_ref(),
         &*state.store,
+        true,
     )
     .await
     .change_context(errors::ApiErrorResponse::InternalServerError)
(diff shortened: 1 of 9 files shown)


crates/router/src/core/utils.rs
```diff
@@ -992,16 +997,19 @@ pub async fn get_profile_id_from_business_details(
     merchant_account: &domain::MerchantAccount,
     request_profile_id: Option<&String>,
     db: &dyn StorageInterface,
+    should_validate: bool,
```
Narayanbhat166, line 1000: validate only in case of payments and mca create, in other cases validation is not needed

Narayanbhat166 (changes requested):

crates/router/src/core/utils.rs
BallaNitesh: Is it required in list payment methods also?

Narayanbhat166: No its not required in payment methods list. The entry point is payments create and once a profile id is given for a payment, it cannot be changed in any other flow

inventvenkat (changes requested):
What is this should validate is been passed everywhere?

inventvenkat (approved):

PR #2252, opened 2023-09-21
fix(router): fix attempt status for techincal failures in psync flow

sai-harsha-vardhan:
## Type of Change

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description
fix attempt status for techincal failures in psync flow
1. In Psync flow for 4xx / 5xx errors mark the previous attempt status instead of marking it as processing / failure

### Additional Changes

## How did you test it?
1. 4xx in payments confirm flow, payment status would be marked as failure
image
image

2. 4xx in payments retrieve flow with force_sync = true for processing payment, status is unchanged
image
image

## Checklist

### DIFF
--- a/crates/router/src/core/payments/operations/payment_response.rs
+++ b/crates/router/src/core/payments/operations/payment_response.rs
@@ -12,6 +12,7 @@ use crate::{
         errors::{self, RouterResult, StorageErrorExt},
         mandate,
         payments::{types::MultipleCaptureData, PaymentData},
+        utils as core_utils,
     },
     db::StorageInterface,
     routes::metrics,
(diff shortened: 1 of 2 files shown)


crates/router/src/core/payments/operations/payment_response.rs
```diff
@@ -311,10 +312,17 @@ async fn payment_response_update_tracker<F: Clone, T: types::Capturable>(
...
                 None => {
-                    let status = match err.status_code {
-                        500..=511 => storage::enums::AttemptStatus::Pending,
-                        _ => storage::enums::AttemptStatus::Failure,
-                    };
+                    let flow_name = core_utils::get_flow_name::<F>()?;
+                    let status =
+                        // mark previous attempt status for technical failures in PSync flow
+                        if flow_name == "PSync" {
+                            payment_data.payment_attempt.status
```
Narayanbhat166, line 319: here we are not allowing the status to be changed in case of genuine error too, this can be handled by checking the status code and then updating status

Narayanbhat166 (changes requested):

Narayanbhat166 (dismissed):

crates/router/src/core/payments/operations/payment_response.rs
```diff
@@ -311,10 +312,20 @@ async fn payment_response_update_tracker<F: Clone, T: types::Capturable>(
...
-                    let status = match err.status_code {
-                        500..=511 => storage::enums::AttemptStatus::Pending,
-                        _ => storage::enums::AttemptStatus::Failure,
-                    };
+                    let flow_name = core_utils::get_flow_name::<F>()?;
+                    let status =
+                        // mark previous attempt status for technical failures in PSync flow
+                        if flow_name == "PSync" {
+                            match err.status_code {
+                                200..=299 => storage::enums::AttemptStatus::Failure,
```
jarnura, line 320: Why 2xx are marked as failures?

jarnura (changes requested):

crates/router/src/core/payments/operations/payment_response.rs
sai-harsha-vardhan: If we get 2xx response from connector, and connector flows sends us the error response then it's a payment failure right? If we don't do this we would never be marking payment to terminal state in psync flows

jarnura (dismissed):

Narayanbhat166 (approved):

jarnura (approved):

PR #2448, opened 2023-10-04
feat(router): add support for payment_type field in payment intent

vspecky:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

This PR adds a new `payment_type` column in the `payment_intent` table, the value of which is inferred during payment creation, and is also included in the response of the list payment methods call.

### Additional Changes

## Motivation and Context

The `payment_type` field will be detrimental in deciding the right mandate flow. Certain merchants may want to just set up the mandate without debiting any amount, whereas some merchants may want to debit some amount as part of the setup. This distinction is done based on the `payment_type`.

## Checklist

### DIFF
--- a/crates/router/src/core/payments/helpers.rs
+++ b/crates/router/src/core/payments/helpers.rs
@@ -2396,7 +2396,7 @@ mod tests {
             profile_id: None,
             merchant_decision: None,
             payment_confirm_source: None,
-            payment_type: api_enums::PaymentType::Normal,
+            payment_type: Some(api_enums::PaymentType::Normal),
         };
         let req_cs = Some("1".to_string());
         let merchant_fulfillment_time = Some(900);
@@ -2444,7 +2444,7 @@ mod tests {
             profile_id: None,
             merchant_decision: None,
             payment_confirm_source: None,
-            payment_type: api_enums::PaymentType::Normal,
+            payment_type: Some(api_enums::PaymentType::Normal),
         };
         let req_cs = Some("1".to_string());
         let merchant_fulfillment_time = Some(10);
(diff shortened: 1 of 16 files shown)


migrations/2023-10-04-120026_add_payment_type_column_in_payment_intent/up.sql
```diff
@@ -0,0 +1,11 @@
...
+
+CREATE TYPE "PaymentType" AS ENUM (
+    'normal',
+    'new_mandate',
+    'setup_mandate',
+    'recurring_mandate'
+);
+
+ALTER TABLE payment_intent
+ADD COLUMN payment_type "PaymentType" NOT NULL DEFAULT 'normal';
```
Narayanbhat166, line 11: This will mark all the previous mandate / setup mandate transactions as 'normal' payments. Is that okay?

vspecky: Ah, good catch. We probably don't want to do that. I'll update it.

crates/router/src/core/payments/helpers.rs
```diff
@@ -632,6 +632,25 @@ pub fn validate_card_data(
     Ok(())
 }
 
+pub fn infer_payment_type(
+    req: &api::PaymentsRequest,
+    mandate_type: Option<&api::MandateTransactionType>,
+) -> api_enums::PaymentType {
+    match mandate_type {
+        Some(api::MandateTransactionType::NewMandateTransaction) => {
+            if let Some(api::Amount::Value(_)) = req.amount {
```
jarnura, line 641: if the amount is zero then only its a setup mandate.

jarnura (changes requested):

Narayanbhat166 (approved):

jarnura (approved):

PR #2475, opened 2023-10-06
fix: fetch data directly from DB in OLAP functions

dracarys18:
## Type of Change

- [x] Bugfix

## Description

Don't  check for storage scheme in OLAP functions.

## Motivation and Context

We don't need KV in OLAP functions, so just fetch things from DB

## How did you test it?

NA

## Checklist

### DIFF
--- a/crates/router/src/db/refund.rs
+++ b/crates/router/src/db/refund.rs
@@ -638,21 +638,21 @@ mod storage {
             &self,
             merchant_id: &str,
             refund_details: &api_models::refunds::RefundListRequest,
-            storage_scheme: enums::MerchantStorageScheme,
+            _storage_scheme: enums::MerchantStorageScheme,
             limit: i64,
             offset: i64,
         ) -> CustomResult<Vec<diesel_models::refund::Refund>, errors::StorageError> {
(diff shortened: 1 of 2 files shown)


Sangamesh26 (approved):

jarnura (approved):

crates/router/src/db/refund.rs
```diff
@@ -638,61 +638,53 @@ mod storage {
             &self,
             merchant_id: &str,
             refund_details: &api_models::refunds::RefundListRequest,
-            storage_scheme: enums::MerchantStorageScheme,
+            _storage_scheme: enums::MerchantStorageScheme,
             limit: i64,
             offset: i64,
         ) -> CustomResult<Vec<diesel_models::refund::Refund>, errors::StorageError> {
-            match storage_scheme {
```
inventvenkat, line 645: Can we write a utility function which says what to use `Postgres` or `KV`, and in that function, we can decide based on feature. And use that function everywhere. So that we will not miss any OLTP

inventvenkat (changes requested):

PR #2478, opened 2023-10-06
feat: add support for ddc flows and kount frm connector

Sangamesh26:
## Type of Change

- [x] New feature

## Description

Add support for DDC for Kount frm connector.

### Additional Changes

## Motivation and Context

Add support for DDC

## Checklist

### DIFF
--- a/crates/router/src/core/payments/operations/payments_device_data_collection.rs
+++ b/crates/router/src/core/payments/operations/payments_device_data_collection.rs
@@ -1,0 +1,303 @@
+use std::marker::PhantomData;
+
+use api_models::enums::FrmSuggestion;
+use async_trait::async_trait;
+use error_stack::ResultExt;
+use router_derive::PaymentOperation;
+use router_env::{instrument, tracing};
+
+use super::{BoxedOperation, Domain, GetTracker, Operation, UpdateTracker, ValidateRequest};
+use crate::{
+    core::{
+        errors::{self, CustomResult, RouterResult, StorageErrorExt},
+        payments::{helpers, operations, CustomerDetails, PaymentAddress, PaymentData},
+    },
+    db::StorageInterface,
+    routes::AppState,
+    services,
+    types::{
+        api::{self, PaymentIdTypeExt},
+        domain,
+        storage::{self, enums as storage_enums},
+    },
+    utils::OptionExt,
+};
+
+#[derive(Debug, Clone, Copy, PaymentOperation)]
+#[operation(ops = "all", flow = "ddc")]
+pub struct PaymentDeviceDataCollection;
+
+#[async_trait]
+impl<F: Send + Clone>
+    GetTracker<F, PaymentData<F>, api_models::payments::PaymentsDeviceDataCollectionRequest>
+    for PaymentDeviceDataCollection
+{
+    #[instrument(skip_all)]
+    async fn get_trackers<'a>(
+        &'a self,
+        state: &'a AppState,
+        payment_id: &api::PaymentIdType,
+        _request: &api_models::payments::PaymentsDeviceDataCollectionRequest,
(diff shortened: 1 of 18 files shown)


crates/router/src/core/payments/transformers.rs
```diff
@@ -474,12 +474,24 @@ where
                     amount,
                     currency: currency.to_string(),
                 }))
+            } else if payments::is_ddc_operation(&operation) {
+                services::ApplicationResponse::Form(Box::new(services::RedirectionFormData {
+                    redirect_form: RedirectForm::Kount {
```
jagan-jaya, line 479: please create a ticket and mark it here to make this logic generic

crates/router/src/services/api.rs
```diff
@@ -1258,6 +1263,71 @@ pub fn build_redirection_form(
...
+                                isSinglePageApp: true,
+                                isDebugEnabled: true
+                            }};
+                            const sdk = new kountSDK(kountConfig, sessionID);
+                        }});
+                        var f = document.createElement('form');
+                        f.action = window.location.pathname.replace(new RegExp("payments/ddc/(\\w+)/(\\w+)"), "payments/complete_ddc/$1/$2");
+                        f.method='POST';
+                        document.body.appendChild(f);
+                        f.submit();
```
jagan-jaya, line 1322: This block will get executed even before the Kount SDK call. Add this within `DOMContentLoaded` block and execute with SetTimeout so that we can be sure that kountSDK is called before redirection
```suggestion
                        var f = document.createElement('form');
                        f.action = window.location.pathname.replace(new RegExp("payments/ddc/(\\w+)/(\\w+)"), "payments/complete_ddc/$1/$2");
                        f.method='POST';
                        document.body.appendChild(f);
                        f.submit();
```

crates/router/src/core/payments/operations/payments_device_data_collection.rs
```diff
@@ -0,0 +1,310 @@
...
+impl<
+        F: Clone + Send,
+        Ctx: PaymentMethodRetrieve,
+        Op: Send + Sync + Operation<F, api_models::payments::PaymentsDeviceDataCollectionRequest, Ctx>,
+    > Domain<F, api_models::payments::PaymentsDeviceDataCollectionRequest, Ctx> for Op
+where
+    for<'a> &'a Op: Operation<F, api_models::payments::PaymentsDeviceDataCollectionRequest, Ctx>,
+{
+    #[instrument(skip_all)]
+    async fn get_or_create_customer_details<'a>(
```
jagan-jaya, line 226: do we need to call `create_customer` flow here as it is only responsible for building the response HTML ?

crates/router/src/core/payments/operations/payments_device_data_collection.rs
```diff
@@ -0,0 +1,310 @@
...
+        _auth_flow: services::AuthFlow,
+    ) -> RouterResult<(
+        BoxedOperation<'a, F, api_models::payments::PaymentsDeviceDataCollectionRequest, Ctx>,
+        PaymentData<F>,
+        Option<CustomerDetails>,
+    )> {
+        let (mut payment_intent, payment_attempt, currency, amount);
+        let db = &*state.store;
+
+        let merchant_id = &merchant_account.merchant_id;
```
jagan-jaya, line 56: can you verify if you need to build all the entities here as we are going to build only the HTML response for this flow

jagan-jaya (changes requested):

jagan-jaya:
@Sangamesh26 @SMohata  This is a high risk PR as it will affect the existing bluesnap <> Kount integration and by the time it gets merged SDK team should also remove the Kount call from SDK side.

PR #2488, opened 2023-10-07
refactor(connector): [Worldpay] Remove Default Case Handling

### LINKED ISSUE #2290: [REFACTOR]: [Worldpay] Remove Default Case Handling

:bookmark: Note: All the changes needed should be contained within `hyperswitch/crates/router/src/connector/`

sudharsangs:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [x] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description
Instead of relying on a default match case `_` I have added conditions for each type in match statements. 

Changes are made at https://github.com/juspay/hyperswitch/blob/main/crates/router/src/connector/worldpay/transformers.rs

## Motivation and Context
Closes #2290

## Checklist


crates/router/src/connector/worldpay/transformers.rs
```diff
@@ -144,7 +144,24 @@ impl TryFrom<&types::ConnectorAuthType> for WorldpayAuthType {
                     api_key: Secret::new(auth_header),
                 })
             }
-            _ => Err(errors::ConnectorError::FailedToObtainAuthType)?,
+            types::ConnectorAuthType::CurrencyAuthKey { auth_key_map: _ } => {
```
prasunna09, line 147: Hi @sudharsangs, 
Can you please revert back this change as wildcard match arms (_) is not in scope for AuthType?

prasunna09 (changes requested):

crates/router/src/connector/worldpay/transformers.rs
sudharsangs: I've made the change @prasunna09

prasunna09 (dismissed):

SanchithHegde (dismissed):
Looks good to me!

Thanks for the PR, @sudharsangs!

prasunna09:
@sudharsangs  Can you please revert back the Cargo.lock file?

sudharsangs:
> Hey @sudharsangs , Please revert back the cargo.lock file. Run `git checkout main -- Cargo.lock` command and push your code.

I did a ` git checkout upstream/main -- Cargo.lock`   but it's not changing I am not sure why

Sangamesh26 (approved):

SanchithHegde (approved):

VedantKhairnar:
Hey @sudharsangs  ,
Thanks a bunch for all your contributions! We've got a little something for you to show our appreciation.
Just take a moment to fill out this [form](https://forms.gle/yqbtjFuP6oxLaBKN8), and get ready for some awesome swag coming your way.
Thanks!

PR #2501, opened 2023-10-09
fix: consume profile_id throughout payouts flow

kashif-m:
## Type of Change

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [x] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description
This PR modifies logic used for fetching connector and storing customer's details using a profile_id. This had already been done for the core functionalities, including payments, refunds, etc.

Core PR - https://github.com/juspay/hyperswitch/pull/2020

### Additional Changes

## Motivation and Context
Move towards using `profile_id` as opposed to `business_label`.

## How did you test it?
Tested locally

#1 Current State [payouts flow is broken]
image

#2 Updated State [using profile_id]
image

## Checklist

### DIFF
--- a/crates/api_models/src/payouts.rs
+++ b/crates/api_models/src/payouts.rs
@@ -146,6 +146,10 @@ pub struct PayoutCreateRequest {
     /// Provide a reference to a stored payment method
     #[schema(example = "187282ab-40ef-47a9-9206-5099ba31e432")]
     pub payout_token: Option<String>,
+
+    /// The business profile to use for this payment, if not passed the default business profile
+    /// associated with the merchant account will be used.
+    pub profile_id: Option<String>,
 }
 
 #[derive(Debug, Clone, Deserialize, Serialize, ToSchema)]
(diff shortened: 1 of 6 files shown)


crates/api_models/src/payouts.rs
```diff
@@ -146,6 +146,10 @@ pub struct PayoutCreateRequest {
     /// Provide a reference to a stored payment method
     #[schema(example = "187282ab-40ef-47a9-9206-5099ba31e432")]
     pub payout_token: Option<String>,
+
+    /// The business profile to use for this payment, if not passed the default business profile
+    /// associated with the merchant account will be used.
+    pub profile_id: Option<String>,
```
jagan-jaya, line 152: have we tested if we don't provide profile_id, it is taking the default profile?

jagan-jaya (approved):

crates/api_models/src/payouts.rs
Narayanbhat166: After a certain point of time `profile_id` will be mandatory and has to be passed in every create request

Narayanbhat166 (approved):

crates/api_models/src/payouts.rs
kashif-m: Sure, will make this change once we need to mandate this

PR #2583, opened 2023-10-13
feat(router): Better UI payment link and order details product image and merchant config support

sahkal:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

Added better UI with animation in payment link along with orderDetails to support product_img_link param. 
It also adds, new params in merchant Account to configure payment link styles

### Additional Changes

`"order_details":[{
        "product_name":"Tea",
        "quantity":1,
        "amount":10,
        "product_img_link":"https://static-00.iconduck.com/assets.00/clock-small-icon-2048x2048-srddipq8.png"
    }`
    
    `pub struct OrderDetailsWithAmount {
    /// Name of the product that is being purchased
    #[schema(max_length = 255, example = "shirt")]
    pub product_name: String,
    /// The quantity of the product to be purchased
    #[schema(example = 1)]
    pub quantity: u16,
    /// the amount per quantity of product
    pub amount: i64,
    /// The image URL of the product
    pub product_img_link: Option<String>,
    }`

`"payment_link_metadata": {
      "merchant_logo": "https://i.pinimg.com/736x/4d/83/5c/4d835ca8aafbbb15f84d07d926fda473.jpg",
      "color_scheme":{
          "primary_color": "#34ebeb",
          "primary_accent_color": null,
          "secondary_color": null
      }
  }`
  
  
 `pub struct PaymentLinkMetadata {
    pub merchant_logo: Option<String>,
    pub color_scheme: Option<PaymentLinkColorSchema>,
}
`

`pub struct PaymentLinkColorSchema {
    pub primary_color: Option<String>,
    pub primary_accent_color: Option<String>,
    pub secondary_color: Option<String>,
}`

## How did you test it?

Screenshot 2023-10-13 at 10 10 21 PM

## Checklist


crates/router/src/core/payment_link.rs
```diff
@@ -56,32 +70,47 @@ pub async fn intiate_payment_link_flow(
...
+        merchant_name: merchant_account.merchant_name,
+        order_details,
+        return_url: payment_intent.return_url.unwrap_or_default(),
+        expiry: fulfillment_time,
+        pub_key: merchant_account.publishable_key.unwrap_or_default(),
+        client_secret: payment_intent.client_secret.unwrap_or_default(),
+        merchant_logo: payment_link_metadata
+            .clone()
+            .map(|pl_metadata| pl_metadata.merchant_logo.unwrap_or_default())
+            .unwrap_or_default(),
```
Narayanbhat166, line 104: why so many unwrap_or_default? can you handle the cases where these will be None? or if there is no such case then these can be made as mandatory, another case to handle this is to raise the error if value is missing

crates/router/src/core/payment_link.rs
```diff
@@ -93,80 +122,39 @@ The get_js_script function is used to inject dynamic value to payment_link sdk,
...
-            }},
-          ]
-    }});
+    let (primary_color, primary_accent_color, secondary_color) = payment_link_metadata
+        .and_then(|pl_metadata| {
+            pl_metadata.color_scheme.map(|color| {
+                (
+                    color.primary_color.unwrap_or("#C6C7C8".to_string()),
+                    color.primary_accent_color.unwrap_or("#6A8EF5".to_string()),
+                    color.secondary_color.unwrap_or("#0C48F6".to_string()),
```
Narayanbhat166, line 143: store these as const values

crates/router/src/core/payment_link.rs
```diff
@@ -93,80 +129,80 @@ The get_js_script function is used to inject dynamic value to payment_link sdk,
...
+      --secondary-color: {secondary_color};
+    }}"
     )
 }
+
+fn validate_sdk_requirements(
+    pub_key: Option<String>,
+    currency: Option<api_models::enums::Currency>,
+    client_secret: Option<String>,
+    return_url: (Option<String>, Option<String>),
```
Narayanbhat166, line 183: take these are two different fields, one as `merchant_account_return_url` and the other as `payment_return_url`, this way it is clear which url to pass in which field when calling the function. The problem in this case is that the caller has to look into the function definition as to know which url to pass where as both are option of string

Narayanbhat166: You can also calculate the return url before calling this function

crates/router/src/core/payment_link.rs
```diff
@@ -93,80 +129,80 @@ The get_js_script function is used to inject dynamic value to payment_link sdk,
...
+
+    let return_url = if let Some(merchant_account_return_url) = return_url.0 {
+        merchant_account_return_url
+    } else {
+        return_url
+            .1
+            .ok_or(errors::ApiErrorResponse::MissingRequiredField {
+                field_name: "return_url",
+            })?
+    };
```
Narayanbhat166, line 205: payment return url should take priority over merchant return url

jarnura (approved):

Narayanbhat166 (approved):

PR #2653, opened 2023-10-20
chore(env): add ttl as env variable

dracarys18:
## Type of Change

- [x] Enhancement

## Description

Add TTL for KV as environment variable.

### Additional Changes

## Motivation and Context

Add TTL for KV as environment variable.

## How did you test it?

- Run the application 
- Check the startup config debug log
Screenshot 2023-10-20 at 4 44 53 PM

## Checklist


SanchithHegde (dismissed):

jarnura (approved):

jarnura (approved):

sai-harsha-vardhan (approved):

config/config.example.toml
```diff
@@ -435,4 +435,9 @@ apple_pay_merchant_cert_key = "APPLE_PAY_MERCHNAT_CERTIFICATE_KEY"      #Private
...
 
 [payment_link]
-sdk_url = "http://localhost:9090/dist/HyperLoader.js"
\ No newline at end of file
+sdk_url = "http://localhost:9090/dist/HyperLoader.js"
+
+# Config for KV setup
+[kv_config]
+# TTL for KV in seconds
+ttl = 900
```
Abhicodes-crypto, line 443: Authentication of client secret is done based on merchant_intent_fullfillment time, so shouldn't data live in KV for merchant_intent_fullfillment time interval. 

Will failing to fetch from KV fallback to DB ?
Will the data be found in DB always as soon as it is expired in Redis ?

Abhicodes-crypto (approved):
Few questions apart from that LGTM

config/config.example.toml
dracarys18: We will have a bigger TTL in prod env so this problem won't occur

PR #2681, opened 2023-10-25
feat(events): add masked json serializer for logging PII values

lsampras:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

- Add a serializer that replicates the existing JSON serializer
- Add a type check in the secret serialize methods where it can optionally mask values when the custom serializer is used
- add a util method to provide masked serialization

### Additional Changes

## Motivation and Context

- while we use the secret type to mask values
- the behaviour for debug representations is pretty common whereas the serialization behaviour needs to satisfy both the use cases of masking or non-masking serializations
- Currently there is no OOTB approach where we can pass metadata to serde to control the serialization behaviour
- In this case we are relying on the std::any::type_name to identify whether a custom serializer was passed...

In this case I'm sticking with no-masking by default & specific functions to mask values

## How did you test it?

image

## Checklist


crates/masking/src/serde.rs
```diff
@@ -46,9 +49,19 @@ where
...
+            format!("{self:?}").serialize(serializer)
+        } else if std::any::type_name::<S>()
+            == std::any::type_name::<
+                serde::__private::ser::FlatMapSerializer<'_, SerializeMap<PIISerializer>>,
+            >()
+        {
+            HashMap::<String, String>::from([]).serialize(serializer)
+        } else {
+            self.peek().serialize(serializer)
+        }
```
NishantJoshi00, line 64: can we make this into a function to promote reusibility

NishantJoshi00 (changes requested):
LGTM

NishantJoshi00 (approved):

dracarys18 (approved):

PR #2720, opened 2023-10-29
refactor(connector): [Cybersource] Remove Default Case Handling

### LINKED ISSUE #2268: [REFACTOR]: [Cybersource] Remove Default Case Handling

:bookmark: Note: All the changes needed should be contained within `hyperswitch/crates/router/src/connector/`

Rutam21:
## Type of Change

- [x] Refactoring

## Description

This PR removes the default case handling for CyberSource and adds all the error cases explicitly.

Fixes #2268 

## How did you test it?

 I ran `cargo clippy --all-features` and it didn't throw any errors.

test

## Checklist

### DIFF
--- a/crates/router/src/connector/cybersource/transformers.rs
+++ b/crates/router/src/connector/cybersource/transformers.rs
@@ -199,7 +199,20 @@ impl TryFrom<&CybersourceRouterData<&types::PaymentsAuthorizeRouterData>>
                     client_reference_information,
                 })
             }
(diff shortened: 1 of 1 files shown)


Rutam21:
Please review and suggest changes, if any. Thanks.

cc: @SamraatBansal @Sangamesh26 @ArjunKarthik

Rutam21:
As the other PR still needs changes and the contributor has not yet committed to the PR with the latest updates, I believe this PR can be reviewed instead and merged before Hacktoberfest ends. Thanks.

cc: @VedantKhairnar @swangi-kumari

PR #2758, opened 2023-11-02
ci(postman): Fix auto-update collection files

pixincreate:
## Type of Change

- [x] Bugfix
- [x] CI/CD

## Description

Fixes https://github.com/juspay/hyperswitch/actions/runs/6694332240/job/18187377846

## Motivation and Context

This PR aims at fixing auto-collection update that was failing when only new collection was added to `colleciton-dir` and not to `collection-json`

## How did you test it?

This can only be tested when you add a new connector's collection to dir and not to json

## Checklist


.github/workflows/release-new-version.yml
```diff
@@ -74,7 +74,17 @@ jobs:
...
             newman dir-import ${POSTMAN_DIR}/${connector} -o ${POSTMAN_JSON_DIR}/${connector}.postman_collection.json
           done
-          (git diff --quiet && git diff --staged --quiet) || (git commit -am 'test(postman): update postman collection files' && echo "Committed changes") || (echo "Unable to commit the following changes:" && git diff)
+
+          if git diff --quiet && git diff --staged --quiet; then
+              echo "No changes detected."
+          elif git add postman; then
+              echo "Changes have been staged."
+          elif git commit -am 'test(postman): update postman collection files'; then
+              echo "Committed changes."
```
SanchithHegde, line 83: Won't this only stage the files and not commit them if there are any unstaged files? Moreover, this is quite confusing and seems unnecessarily complicated.

.github/workflows/release-new-version.yml
```diff
@@ -74,7 +74,17 @@ jobs:
...
+          else
+              echo "Unable to commit the following changes:"
+              git diff
+          fi
```
SanchithHegde, line 87: Can't we just do something like:

```suggestion
          git add postman
          if ! git diff --staged --quiet postman; then
            git commit --message 'test(postman): update postman collection files'
          fi
```

Performing the `git add` before checking for changed files also handles untracked files in the `postman` directory.
And any errors in the workflow should stop the workflow, which I think is fine.

.github/workflows/release-new-version.yml
pixincreate: Oh! Yes! That got a bit confusing which resulted in creation of mess. Will push a fix for that in a few minutes.

.github/workflows/release-new-version.yml
pixincreate: Had a similar thought.

.github/workflows/release-new-version.yml
```diff
@@ -74,7 +74,11 @@ jobs:
             connector=$(basename ${connector_dir})
             newman dir-import ${POSTMAN_DIR}/${connector} -o ${POSTMAN_JSON_DIR}/${connector}.postman_collection.json
           done
-          (git diff --quiet && git diff --staged --quiet) || (git commit -am 'test(postman): update postman collection files' && echo "Committed changes") || (echo "Unable to commit the following changes:" && git diff)
+
+          if git add postman && ! git diff --staged --quiet postman; then
+            git commit -am 'test(postman): update postman collection files'
```
SanchithHegde, line 79:
```suggestion
            git commit --message 'test(postman): update postman collection files'
```

Sufficient to commit changes, since we've already staged changes in the `postman` directory.

SanchithHegde (approved):

ShankarSinghC (approved):

Chethan-rao (approved):

PR #2780, opened 2023-11-03
feat: spawn webhooks and async scheduling in background

jarnura:
## Type of Change

- [x] Enhancement

## Description

- Sending webhooks to the merchants is a push event, and that does not need to be a blocking await in payments since the webhooks events mechanism has retry and async scheduling with it.

- Async scheduling, like polling status jobs, can spawn separately and need not be blocking await in payments flows.

## Solution

Spawn a new thread using tokio::spawn, and runtime will run these in the background and not in the current thread.

Runtime won't handle the lifetime of these background threads when the server crashes or shuts down. Since we have a graceful shutdown with an offset, it gives time for these threads to complete. In future, we will move to a channel-based approach.

## Motivation and Context

These jobs need not be a blocking await for payments and add much latency to Confirm APIs.

## How did you test it?

If tracing is enabled, we can check the traces waterfall to find the calls are running concurrently 
with others. 

If tracing is not enabled, we can check the timestamp in logs with subsequent calls to find these run concurrently with others.

Screenshot 2023-11-03 at 6 51 52 PM

Screenshot 2023-11-03 at 6 54 16 PM

## Checklist


crates/router/src/core/payments/operations/payment_confirm.rs
```diff
@@ -555,7 +555,27 @@ impl<F: Clone + Send, Ctx: PaymentMethodRetrieve> Domain<F, api::PaymentsRequest
         requeue: bool,
         schedule_time: Option<time::PrimitiveDateTime>,
     ) -> CustomResult<(), errors::ApiErrorResponse> {
-        helpers::add_domain_task_to_pt(self, state, payment_attempt, requeue, schedule_time).await
+        // This spawns this futures in a background thread, the execption inside this future won't affect
+        // the current thread and the lifecycle of spawn thread is not handled by runtime.
+        // So when server shutdown won't wait for this thread's completion.
+        let m_payment_attempt = payment_attempt.clone();
+        let m_state = state.clone();
```
dracarys18, line 562: This is an expensive call ideally we should have wrapper struct like
```rust
struct AppState {
	inner: Arc<ActualAppState>
}
```

jarnura: Yes, but touching `AppState` introduces lots of changes in the whole, and that changes can taking separately.

dracarys18: Cool created Issue for the same https://github.com/juspay/hyperswitch/issues/2781

dracarys18 (dismissed):

crates/router/src/core/payments/operations/payment_confirm.rs
```diff
@@ -555,7 +555,27 @@ impl<F: Clone + Send, Ctx: PaymentMethodRetrieve> Domain<F, api::PaymentsRequest
...
+                helpers::add_domain_task_to_pt(
+                    &m_self,
+                    m_state.as_ref(),
+                    &m_payment_attempt,
+                    requeue,
+                    schedule_time,
+                )
+                .await
+            }
+            .in_current_span(),
```
NishantJoshi00, line 575: Does this also propagate the fields captured in the current span? how the log structure will be done on the log aggregator?

NishantJoshi00 (approved):
Looks good!

crates/router/src/core/payments/operations/payment_confirm.rs
jarnura: the log structure will be same and it can capture the fields in current span.

```
{
    "message": "[GENERIC_INSERT - EVENT] diesel_models::query::generics",
    "pid": 8,
    "level": "DEBUG",
    "target": "diesel_models::query::generics",
    "service": "router",
    "line": 88,
    "file": "hyperswitch_oss/crates/diesel_models/src/query/generics.rs",
    "fn": "generic_insert",
    "full_name": "diesel_models::query::generics::generic_insert",
    "flow": "PaymentsConfirm",
    "request_method": "POST",
    "request_url_path": "/payments/***/confirm",
    "request_id": "***",
    "merchant_id": "***",
    "extra": {
        ...
    }
}
```

The *** are captured values

akshay-97 (approved):

Narayanbhat166 (approved):

PR #2794, opened 2023-11-06
fix(connector): Add attempt_status in field in error_response

srujanchikke:
## Type of Change

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

Payment attempt status would remain in non terminal state if psync gets 4xx or 5xx .This needs to handled in connector error response .This PR introduces optional field called attempt_status field in ErrorResponse struct by which we can update payment status in case 4xx and 5xx in connector `get_error_response` .

### Additional Changes

## Motivation and Context

Payment attempt status would remain in non terminal state if psync gets 4xx or 5xx .This needs to handled in connector error response .This PR introduces optional field called attempt_status field in ErrorResponse struct by which we can update payment status in case 4xx and 5xx in connector `get_error_response` .

## How did you test it?

tested using postman .
This can't be tested with exact scenario .Although we can duplicate this scenarion with other case 
1) make a failure payment using failed test scenarios
2)update the status in db to pending
3) psync the payment setting force_synce =true

## Checklist

### DIFF
--- a/crates/router/src/services/api.rs
+++ b/crates/router/src/services/api.rs
@@ -10,7 +10,7 @@ use std::{
 };
 
 use actix_web::{body, web, FromRequest, HttpRequest, HttpResponse, Responder, ResponseError};
-use api_models::enums::{AttemptStatus, CaptureMethod};
+use api_models::enums::CaptureMethod;
 pub use client::{proxy_bypass_urls, ApiClient, MockApiClient, ProxyClient};
 pub use common_utils::request::{ContentType, Method, Request, RequestBuilder};
 use common_utils::{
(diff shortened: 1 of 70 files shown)


crates/router/src/services/api.rs
```diff
@@ -406,16 +408,9 @@ where
...
-                                                && error_res.status_code == 403
-                                                && error_res.reason
-                                                    == Some(format!(
-                                                        "{} in bluesnap dashboard",
-                                                        consts::REQUEST_TIMEOUT_PAYMENT_NOT_FOUND
-                                                    ))
-                                            {
-                                                router_data.status = AttemptStatus::Failure;
-                                            };
+                                            router_data.status = error_res
```
sai-harsha-vardhan, line 411: update this value only if attempt_status is some, no need to update otherwise

sai-harsha-vardhan (changes requested):

sai-harsha-vardhan (approved):

ArjunKarthik (approved):

SanchithHegde (approved):

likhinbopanna (approved):

PR #2915, opened 2023-11-17
fix: status goes from pending to partially captured in psync

hrithikesh026:
## Type of Change

- [x] Bugfix

## Description

Status would always go from `Processing` to `PartiallyCaptured` if force sync is done. Fixed it.

Other fixes and changes: 

1. In payment_create, with `confirm = true` and `capture_method = manual`, if some `amount_to_capture` is sent in request body, it is updated in DB but the amount sent to connector will always be `amount`. So it is possible that amount_to_capture can be less than original amount but the original amount is sent to connector. Fixed it by requiring `amount_to_capture` to be equal to `amount` if `confirm = true` and `automatic` capture

2. surcharge amount wouldn't get updated in DB if connector returns error response. Fixed it.
3. amount_to_capture was not getting updated in DB if sent in payment_capture call. Fixed it.

### Additional Changes

## How did you test it?

Manual

Before PSync
Screenshot 2023-11-20 at 7 34 48 PM

After PSync
Screenshot 2023-11-20 at 7 34 59 PM

## Checklist


jarnura (approved):

prasunna09 (approved):

crates/router/src/core/payments/operations/payment_capture.rs
```diff
@@ -251,20 +251,29 @@ impl<F: Clone, Ctx: PaymentMethodRetrieve>
     where
         F: 'b + Send,
     {
-        payment_data.payment_attempt = match &payment_data.multiple_capture_data {
-            Some(multiple_capture_data) => db
-                .store
+        payment_data.payment_attempt = if payment_data.multiple_capture_data.is_some()
+            || payment_data.payment_attempt.amount_to_capture.is_some()
+        {
```
Narayanbhat166, line 256: This can be achieved using match statement
```suggestion
        match &payment_data.multiple_capture_data {
            Some(multiple_capture_data) if payment_data.payment_attempt.amount_to_capture.is_some()
        { }
```

I think this is what you want to achieve right?

Narayanbhat166 (approved):

crates/router/src/core/payments/operations/payment_capture.rs
hrithikesh026: I wanted to eliminate unnecessary DB Update call. Update DB only if Some data is present in either in `payment_data.multiple_capture_data`, or `payment_data.payment_attempt.amount_to_capture`

jarnura (approved):

PR #2916, opened 2023-11-17
fix: api lock on PaymentsCreate

dracarys18:
## Type of Change

- [x] Bugfix

## Description

Added API locking for `PaymentsCreate` and removed unnecessary `get_or_generate_id` calls.

## Motivation and Context

This fixes race condition when you do `PaymentsCreate` with `confirm=true`. A scenario when this can happen is when 
you do `PaymentsCreate` with `confirm = true` and while it's happening Connector sends a webhook and it updates PaymentAttempt but it hasn't send the response yet so the process is running with old `PaymentAttempt` state which causes inconsistency.

## How did you test it?

- 	PaymentsCreate with `confirm = true` to `adyen` connector which has the above behaviour.
- And sanity test all `Payments` API.

## Checklist

### DIFF
--- a/crates/router/src/core/payments/operations/payment_confirm.rs
+++ b/crates/router/src/core/payments/operations/payment_confirm.rs
@@ -6,7 +6,7 @@ use api_models::{
 };
 use async_trait::async_trait;
 use common_utils::ext_traits::{AsyncExt, Encode};
-use error_stack::ResultExt;
+use error_stack::{report, IntoReport, ResultExt};
 use futures::FutureExt;
 use redis_interface::errors::RedisError;
 use router_derive::PaymentOperation;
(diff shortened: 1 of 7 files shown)


crates/router/src/core/payments/operations/payment_confirm.rs
```diff
@@ -794,14 +786,19 @@ impl<F: Send + Clone, Ctx: PaymentMethodRetrieve> ValidateRequest<F, api::Paymen
...
+            .clone()
+            .ok_or(report!(errors::ApiErrorResponse::PaymentNotFound))?;
 
         Ok((
             Box::new(self),
             operations::ValidateResult {
                 merchant_id: &merchant_account.merchant_id,
-                payment_id: api::PaymentIdType::PaymentIntentId(payment_id),
+                payment_id: payment_id
+                    .and_then(|id| crate::core::utils::validate_id(id, "payment_id"))
```
jarnura, line 800:
```suggestion
                    .and_then(|id| core::utils::validate_id(id, "payment_id"))
```

jarnura (dismissed):

Abhicodes-crypto (dismissed):

jarnura (approved):

Abhicodes-crypto (approved):

PR #2928, opened 2023-11-21
fix: kv logs when KeyNotSet is returned

dracarys18:
## Type of Change

- [x] Bugfix

## Description

Return error on KeyNotSet.

## Motivation and Context

This fixes log throwing `success` logs in case of `KeyNotSet` in Redis.

## How did you test it?

- `PaymentsCreate` with same `payment_id` twice.
- On the second call log throws this error

Screenshot 2023-11-21 at 12 59 03 PM

## Checklist


jarnura (approved):

Narayanbhat166 (approved):

preetamrevankar:
Bro @dracarys18 what do we need to verify here whether do we need to check that log present or it should not come.

PR #2956, opened 2023-11-22
feat(events): add type info to outgoing requests & maintain structural & PII type info

lsampras:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

- Currently we rely on the `send_request` function and the `Request` object to send outgoing api calls
- This internally relies on assuming that the caller has passed a correct content type & correctly serialized the json payload
- Instead we now accept a generic request that implements serialize + maskedSerialize s.t we can use it to send requests + log the masked bodies for such requests

### Additional Changes

## Motivation and Context

- This allows us to be more robust about our request body types and avoid mismatch's
- The masking capability also allows us to now log the raw request as needed

### Future enhancements
- we can also require the bodies to implement debug fmt for quick logging
- we can also extract metadata from the request bodies such as payment_id/refund_id/flow type via common traits

## How did you test it?

- make payment calls with confirm=true
- the underlying connector call would be logged with the following info
image

## Checklist


crates/router/src/core/payment_methods/transformers.rs
```diff
@@ -555,31 +543,25 @@ pub async fn mk_delete_card_request_hs(
...
     Ok(request)
 }
 
-pub fn mk_delete_card_request<'a>(
+pub fn mk_delete_card_request(
     locker: &settings::Locker,
-    merchant_id: &'a str,
-    card_id: &'a str,
+    merchant_id: &'static str,
+    card_id: &'static str,
```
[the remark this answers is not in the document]
lsampras: this isn't auto-inferred it seems...

crates/router/src/connector/powertranz.rs
```diff
@@ -283,19 +278,14 @@ impl
...
             .parse_value("PowerTranz RedirectResponsePayload")
             .change_context(errors::ConnectorError::ResponseDeserializationFailed)?;
         let spi_token = format!(r#""{}""#, redirect_payload.spi_token);
-        let powertranz_req =
-            types::RequestBody::log_and_get_request_body(&spi_token, |spi_token| {
-                Ok(spi_token.to_string())
-            })
-            .change_context(errors::ConnectorError::RequestEncodingFailed)?;
-        Ok(Some(powertranz_req))
+        Ok(RequestContent::Json(Box::new(spi_token)))
```
ArjunKarthik, line 288: request body here is `text/plain`. Let's introduce one more type in `RequestContent` which handle text/plain body

crates/pm_auth/src/types/api.rs
```diff
@@ -38,8 +38,8 @@ pub trait ConnectorIntegration<T, Req, Resp>: ConnectorIntegrationAny<T, Req, Re
     fn get_request_body(
         &self,
         _req: &super::PaymentAuthRouterData<T, Req, Resp>,
-    ) -> CustomResult<Option<RequestBody>, ConnectorError> {
-        Ok(None)
+    ) -> CustomResult<RequestContent, ConnectorError> {
+        Ok(RequestContent::Json(Box::new(serde_json::json!(r#""#))))
```
ArjunKarthik, line 42:
```suggestion
        Ok(RequestContent::Json(Box::new(serde_json::json!(r#"{}"#))))
```
Since this method will return http request body, let's default it to empty json object instead of empty string

crates/router/src/services/api.rs
```diff
@@ -133,8 +137,8 @@ pub trait ConnectorIntegration<T, Req, Resp>: ConnectorIntegrationAny<T, Req, Re
         &self,
         _req: &types::RouterData<T, Req, Resp>,
         _connectors: &Connectors,
-    ) -> CustomResult<Option<types::RequestBody>, errors::ConnectorError> {
-        Ok(None)
+    ) -> CustomResult<RequestContent, errors::ConnectorError> {
+        Ok(RequestContent::Json(Box::new(json!(r#""#))))
```
ArjunKarthik, line 141: Since this method will return http request body, let's default it to empty json object instead of empty string
```suggestion
Ok(RequestContent::Json(Box::new(json!(r#"{}"#))))
```

crates/router/src/services/api.rs
```diff
@@ -278,6 +282,15 @@ where
 {
     // If needed add an error stack as follows
     // connector_integration.build_request(req).attach_printable("Failed to build request");
+    logger::debug!(connector_request=?connector_request);
+    let masked_conn_req = connector_request.as_ref().map(|req| match &req.body {
+        Some(RequestContent::Json(payload))
+        | Some(RequestContent::FormUrlEncoded(payload))
+        | Some(RequestContent::Xml(payload)) => payload.masked_serialize().unwrap_or_default(),
+        _ => serde_json::Value::Null,
```
ArjunKarthik, line 290: If any new enum variant added we will never know changes required here. So let's not use default, instead let's handle case by case.

crates/router/src/services/api.rs
```diff
@@ -351,10 +364,48 @@ where
...
+                    let masked_request_body = match &request.body {
+                        Some(request) => match request {
+                            RequestContent::Json(i)
+                            | RequestContent::FormUrlEncoded(i)
+                            | RequestContent::Xml(i) => i
+                                .masked_serialize()
+                                .unwrap_or(json!({ "error": "failed to mask serialize"})),
+                            RequestContent::FormData(_) => json!({"request_type": "FORM_DATA"}),
+                        },
+                        None => json!({"error": "EMPTY_REQUEST_BODY"}),
```
ArjunKarthik, line 377: in case of Get calls to the connector, body will be always empty. So instead of passing random json let's pass Null

ArjunKarthik (changes requested):

ArjunKarthik (dismissed):

NishantJoshi00 (approved):

ThisIsMani (approved):
Reviewed dashboard specific changes.

crates/router/src/connector/adyen.rs
```diff
@@ -202,7 +198,7 @@ impl
                 .url(&types::SetupMandateType::get_url(self, req, connectors)?)
                 .attach_default_headers()
                 .headers(types::SetupMandateType::get_headers(self, req, connectors)?)
-                .body(types::SetupMandateType::get_request_body(
+                .set_body(types::SetupMandateType::get_request_body(
```
SanchithHegde, line 201: We'd have to call `set_body()` instead of `body()` in the connector template?

SanchithHegde (approved):

ThisIsMani (approved):

PR #2958, opened 2023-11-23
fix(drainer): increase jobs picked only when stream is not empty

dracarys18:
## Type of Change

- [x] Bugfix

## Description

Increment jobs picked only when stream is not empty

## Motivation and Context

Currently we are incrementing `jobs_picked` whenever we try to read a stream which gives the wrong notion that drainer is picking the job in every stream read. So only increment `jobs_picked` when stream is not empty.

## How did you test it?

- Do a payment with KV Merchant.
- Run on data source VM-Single
```
router_CYCLES_COMPLETED_SUCCESSFULLY_total
router_CYCLES_COMPLETED_UNSUCCESSFULLY_total
```
- Check if `router_CYCLES_COMPLETED_SUCCESSFULLY_total` and `router_CYCLES_COMPLETED_UNSUCCESSFULLY_total` both are increasing 

 Screenshot 2023-11-23 at 1 23 29 PM

## Checklist

### DIFF
--- a/config/docker_compose.toml
+++ b/config/docker_compose.toml
@@ -15,7 +15,7 @@ level = "DEBUG" # What you see in your terminal.
 
 [log.telemetry]
 traces_enabled = false                                      # Whether traces are enabled.
-metrics_enabled = false                                     # Whether metrics are enabled.
+metrics_enabled = true                                     # Whether metrics are enabled.
 ignore_errors = false                                       # Whether to ignore errors during traces or metrics pipeline setup.
 otel_exporter_otlp_endpoint = "https://otel-collector:4317" # Endpoint to send metrics and traces to.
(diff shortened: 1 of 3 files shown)


akshay-97 (approved):

jarnura (approved):

config/docker_compose.toml
```diff
@@ -15,7 +15,7 @@ level = "DEBUG" # What you see in your terminal.
 
 [log.telemetry]
 traces_enabled = false                                      # Whether traces are enabled.
-metrics_enabled = false                                     # Whether metrics are enabled.
+metrics_enabled = true                                     # Whether metrics are enabled.
```
Abhicodes-crypto, line 18: Is this pushed accidentally ?

PR #3045, opened 2023-12-04
feat: include version number in response headers and on application startup

SanchithHegde:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

This PR includes changes to:

- Include version number in `x-hyperswitch-version` response header.
- Print out binary name and version number at startup, after the application configuration has been deserialized, but before the logging system has been set up.

## Motivation and Context

This should help us identify the version of code running slightly easier.

## How did you test it?

Running all binaries (router, scheduler, drainer) locally.

## Checklist


Narayanbhat166 (dismissed):

crates/drainer/src/main.rs
```diff
@@ -20,6 +20,9 @@ async fn main() -> DrainerResult<()> {
     let shutdown_intervals = conf.drainer.shutdown_interval;
     let loop_interval = conf.drainer.loop_interval;
 
+    #[cfg(feature = "vergen")]
+    println!("Starting drainer (Version: {})", router_env::git_tag!());
```
lsampras, line 24:
```suggestion
    eprintln!("Starting drainer (Version: {})", router_env::git_tag!());
```
should we print to stderr?
or maybe print this in json compatible format maybe...

SanchithHegde: I wouldn't prefer printing to stderr, as this isn't exactly some information about something going wrong.

As for the JSON format, I don't think it's absolutely necessary, since we're also logging out the console and file logging filters directly to stdout. If there's a reason to use JSON other than this being parsed by Loki/Promtail, please do let me know.

lsampras (dismissed):

Narayanbhat166 (approved):

lsampras (approved):

Issue #3112, opened 2023-12-12
[FEATURE] onboard customer's bank details using /payment_methods CREATE


kashif-m:
### Feature Description

Add capability to onboard customers and their bank details prior to making a txn.

**Current State**
Only card payment method can be persisted using /payment_methods endpoint. A txn has to be completed for persisting bank details.

**Use case**
Use payment_tokens for initiating a bank txn.

### Possible Implementation

Add capability in /payment_methods CREATE API to persist bank details.

### Have you spent some time checking if this feature request has been raised before?

- [X] I checked and didn't find a similar issue

### Have you read the Contributing Guidelines?

- [X] I have read the [Contributing Guidelines](https://github.com/juspay/hyperswitch/blob/main/docs/CONTRIBUTING.md)

### Are you willing to submit a PR?

Yes, I am willing to submit a PR!

kashif-m:
This is merged and deployed in sandbox.

PR #3113, opened 2023-12-12
feat(payment_method): add capability to store bank details using /payment_methods endpoint

kashif-m:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [x] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description
Ref - Issue #3112 

### Additional Changes

## How did you test it?

Tested locally

Postman collection - https://galactic-capsule-229427.postman.co/workspace/My-Workspace~2b563e0d-bad3-420f-8c0b-0fd5b278a4fe/collection/9906252-142ea0ea-f61f-48e7-8317-5f819dd535e9?action=share&creator=9906252

**Test Cases**
There are two major flows to be tested out which are fixed in this PR

1. Check payouts operation works end to end (payout creation + payout fulfillment for bank / card)
   - Create a payout with `auto_fulfill` false
   - Fulfill this payout using /payout/fulfill API call
   - This ensures the intermediary data is being persisted in temp locker
2. Payment methods saved for a customer can be used for any operation (payments / payouts) using payment_token
   - Create and fulfill a payout with `recurring` true against a cusomter_id
   - List SPMs for this customer
   - Using payment_token to carry out a payment or a payout transaction

## Checklist


crates/router/src/connector/adyen/transformers.rs
```diff
@@ -4087,19 +4071,41 @@ pub struct AdyenPayoutCancelRequest {
     merchant_account: Secret<String>,
 }
 
+#[cfg(feature = "payouts")]
+impl TryFrom<&PayoutMethodData> for PayoutCardDetails {
+    type Error = Error;
+    fn try_from(item: &PayoutMethodData) -> Result<Self, Self::Error> {
+        match item {
+            PayoutMethodData::Card(card) => Ok(Self {
+                _type: "scheme".to_string(), // FIXME: Remove hardcoding
```
vspecky, line 4080: We should avoid having underscores in identifier names since Rust interprets these differently (unused). If this is for JSON (de)serialization support, you can name the field `card_type` and annotate it with `#[serde(rename = "_type")]` to use `_type` for (de)serialization

crates/router/src/connector/adyen/transformers.rs
```diff
@@ -4087,19 +4071,41 @@ pub struct AdyenPayoutCancelRequest {
...
+                expiry_year: card.expiry_year.peek().to_string(),
+                holder_name: card
+                    .card_holder_name
+                    .clone()
+                    .get_required_value("card_holder_name")
+                    .change_context(errors::ConnectorError::MissingRequiredField {
+                        field_name: "payout_method_data.card.holder_name",
+                    })?
+                    .peek()
+                    .to_string(),
```
vspecky, line 4092: Is it okay for this to be mandatory? I don't have much context about payouts but for normal payments, the card holder name is not mandatory for a lot of connectors.

crates/router/src/connector/klarna.rs
```diff
@@ -306,6 +306,7 @@ impl
                     | storage_enums::PaymentExperience::OneClick
                     | storage_enums::PaymentExperience::RedirectToUrl,
                     api_models::enums::PaymentMethodType::Ach
+                    | api_models::enums::PaymentMethodType::AchBankTransfer
```
vspecky, line 309: We already have `Ach`, `Bacs` and `Sepa` in the `PaymentMethodType` enum. We do not need separate variants specifically for bank transfers. Wherever the context of the payment method is needed alongside Ach, Bacs and Sepa, please pass an instance of the `PaymentMethod` enum (here `PaymentMethod::BankTransfer`)

crates/router/src/core/payment_methods/cards.rs
```diff
@@ -111,30 +134,44 @@ pub async fn add_payment_method(
...
-                recurring_enabled: false,           //[#219]
-                installment_payment_enabled: false, //[#219]
-                payment_experience: Some(vec![api_models::enums::PaymentExperience::RedirectToUrl]), //[#219]
-            };
-            Ok((payment_method_response, false))
-        }
+
+    let response = match req.payment_method {
+        api_enums::PaymentMethod::BankTransfer => match req.bank_transfer.clone() {
+            Some(bank) => add_bank_to_locker(
```
vspecky, line 140: @NishantJoshi00 @Chethan-rao @ShankarSinghC It's better if we have your review here for locker-related stuff.

crates/router/src/core/payment_methods/cards.rs
```diff
@@ -208,6 +246,60 @@ pub async fn update_customer_payment_method(
...
+    customer_id: &String,
+) -> errors::CustomResult<(api::PaymentMethodResponse, bool), errors::VaultError> {
+    let key = key_store.key.get_inner().peek();
+    let payout_method_data = api::PayoutMethodData::Bank(bank.clone());
+    let enc_data = async {
+        serde_json::to_value(payout_method_data.to_owned())
+            .into_report()
+            .change_context(errors::VaultError::SavePaymentMethodFailed)
+            .attach_printable("Unable to encode payout method data")
+            .ok()
```
vspecky, line 264: Are we sure we want to ignore the error here? Please ensure error propagation is not a requirement for this flow. If it's not a requirement, please log the error `crate::logger::error!()`.

vspecky (changes requested):

crates/common_enums/src/enums.rs
```diff
@@ -1101,6 +1101,7 @@ pub enum PaymentExperience {
 #[strum(serialize_all = "snake_case")]
 pub enum PaymentMethodType {
     Ach,
+    AchBankTransfer,
```
Sarthak1799, line 1104: Why is this separate variant required for bank transfers?

crates/router/src/connector/adyen/transformers.rs
kashif-m: For this API, this is a required parameter.

crates/common_enums/src/enums.rs
kashif-m: This is required since a given `payment_method_type` can map to a single `payment_method`. For payout use-cases, the `payment_method` for bank payouts is `BankTransfer`.

Ach, Bacs and Sepa are mapped to `BankDebit`. So we need new variants

crates/router/src/connector/klarna.rs
kashif-m: https://github.com/juspay/hyperswitch/pull/3113#discussion_r1429191182

crates/router/src/core/payment_methods/cards.rs
kashif-m: I have updated the code, and now logging the error before changing context.

crates/common_enums/src/enums.rs
vspecky: @kashif-m The `PaymentMethodType` enum is shared by a lot of modules including Payments, Smart Routing, and Payouts now. So we shouldn't make concessions for any particular module if it's not going to be useful for the other dependents.

It's not necessary for a `PaymentMethodType` to map onto a single payment method. Certain payment method types like `Ach` like you mentioned can be mapped onto multiple payment methods. Hence wherever we are dependent on payment method type information in context of payment method information, we need to have both a `PaymentMethod` as well as a `PaymentMethodType`.

If you have come across a case where we're breaking this convention, then that's a discrepancy. Please create an issue for it and inform me.

kashif-m: @vspecky you can take a look at this function `validate` which uses a `foreign_from`. This makes sure a given variant in `PaymentMethodType` can only map to a single variant `PaymentMethod`.

ref - https://github.com/juspay/hyperswitch/blob/main/crates/router/src/types/api/payment_methods.rs#L24

crates/router/src/core/payment_methods/cards.rs
kashif-m: @NishantJoshi00 @Chethan-rao @ShankarSinghC - can you guys add your review for the locker bit?

NishantJoshi00 (dismissed):
The request construction and and fields looks to be aligned with the locker schema

vspecky (approved):

crates/router/src/connector/adyen/transformers.rs
```diff
@@ -4028,7 +4012,7 @@ pub struct AdyenPayoutEligibilityRequest {
 #[serde(rename_all = "camelCase")]
 pub struct PayoutCardDetails {
     #[serde(rename = "type")]
-    _type: String,
+    payment_method_type: String,
     number: String,
     expiry_month: String,
     expiry_year: String,
```
ArjunKarthik, line 4018: Why number, expiry_month & expiry_year were kept as String rather than Secret? While logging the connector request it will expose all those infos. Use secret<String> instead

crates/router/src/connector/adyen/transformers.rs
```diff
@@ -4087,19 +4071,41 @@ pub struct AdyenPayoutCancelRequest {
...
+impl TryFrom<&PayoutMethodData> for PayoutCardDetails {
+    type Error = Error;
+    fn try_from(item: &PayoutMethodData) -> Result<Self, Self::Error> {
+        match item {
+            PayoutMethodData::Card(card) => Ok(Self {
+                payment_method_type: "scheme".to_string(), // FIXME: Remove hardcoding
+                number: card.card_number.peek().to_string(),
+                expiry_month: card.expiry_month.peek().to_string(),
+                expiry_year: card.expiry_year.peek().to_string(),
+                holder_name: card
```
ArjunKarthik, line 4084: @kashif-m Are you using dynamic fields for payouts? if so please make changes in `crates/router/src/configs/defaults.rs`

ArjunKarthik (changes requested):

ArjunKarthik (approved):

vspecky (approved):

PR #3165, opened 2023-12-19
refactor: make return_url field as text

Narayanbhat166:
## Type of Change

- [x] Refactoring

## Description

Change the `return_url` type in payment_intent in data models.

### Additional Changes

```sql
ALTER TABLE payment_intent
ALTER COLUMN return_url TYPE TEXT;
```

## Motivation and Context

In order to accomodate return urls of length 255 and more.

## How did you test it?

- Create a payment with return url of more than 255 characters.

Screenshot 2023-12-19 at 12 29 25 PM

## Checklist


Narayanbhat166:
@jarnura @SanchithHegde should we make this field as text or increase the limit to 512 characters? I would suggest increasing the limit to 512/1024 instead of making this a text field as we need not support arbitrary number of query params in the return_url. what are your opinions?

dracarys18 (approved):

Narayanbhat166:
Altering the type directly will cause `cache_plan` error on diesel. Do we store this in a new field?

Narayanbhat166:
Because of security reasons, we will not be making this change. If there is a specific requirement for `return_url` to be more than 255, we can solutionize this separately. Will be closing this issue for now

PR #3210, opened 2023-12-27
feat: add deep health check

Chethan-rao:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

This PR adds support for performing a deeper health check of the application which includes:
* Database connection check with read, write and delete queries (Added support for database transaction too)
* Redis connection check with set, get and delete key (Set key with expiry of 30sec)
* Locker health call from Hyperswitch application

The deep health check API returns a 200 status code if all the components are in good health, but returns 500 if at least one component's health is down

### Additional Changes

## How did you test it?

1. Hit `/health` for getting just the application health.
2. Hit `/health/deep_check` for getting deep report of components health

* Test case 1 (Application not able to connect to database because postgresql is down)

image

* Test case 2 (Database Insert query failing because of duplication)

image

* Test case 3 (Locker is down)

image

* Test case 4 (Locker is up but key custodian locked)

image

* Test case 5 (Locker is up with key custodian unlocked)

image

## Checklist

- [x]  I reviewed the submitted code


crates/router/src/db/health_check.rs
```diff
@@ -0,0 +1,181 @@
...
+    async fn health_check_locker(
+        &self,
+        _: &routes::AppState,
+    ) -> CustomResult<u16, errors::HealthCheckLockerError> {
+        Ok(0)
+    }
+}
+
+#[async_trait::async_trait]
+impl HealthCheckInterface for KafkaStore {
```
SanchithHegde, line 160: We should be calling the same methods on `self.diesel_store` instead, we won't know the actual status of services when `KafkaStore` is used instead of `Store`.

Also, should we move this to `crates/router/src/db/kafka_store.rs`, considering that the other implementations for the type exist there?

crates/router/src/routes/health.rs
```diff
@@ -11,3 +13,47 @@ pub async fn health() -> impl actix_web::Responder {
...
+        }
+        Err(err) => err.to_string(),
+    };
+
+    logger::debug!("Locker health check end");
+
+    actix_web::HttpResponse::Ok().body(format!(
+        "Database - {}\nRedis - {}\nLocker - {}",
+        db_status, redis_status, locker_status
+    ))
```
SanchithHegde, line 58: As an optional change, we can consider returning a JSON instead of a string. It would allow us to extend the deep health check response with more fields/information easily.

crates/router/src/db/health_check.rs
Chethan-rao: makes sense

SanchithHegde (dismissed):

SanchithHegde (approved):

crates/api_models/src/health_check.rs
```diff
@@ -0,0 +1,19 @@
...
+    pub status: String,
+    pub key_custodian_status: KeyCustodianStatus,
+}
+
+#[derive(Debug, Clone, serde::Serialize, serde::Deserialize)]
+pub enum KeyCustodianStatus {
+    Unavailable,
+    Locked,
+    Unlocked,
+}
```
NishantJoshi00, line 19: No need for this, the locker health check provides context around this. We only need to know if locker is available or not

crates/router/src/db/health_check.rs
```diff
@@ -0,0 +1,157 @@
...
+    async fn health_check_db(
+        &self,
+        db: &dyn StorageInterface,
+    ) -> CustomResult<(), errors::HealthCheckDBError> {
+        let conn = connection::pg_connection_write(self)
+            .await
+            .change_context(errors::HealthCheckDBError::DBError)?;
+
+        let _data = conn
+            .transaction_async(|conn| {
```
NishantJoshi00, line 41: might consider going with the [`test_transaction`](https://docs.rs/diesel/latest/diesel/connection/trait.Connection.html#method.test_transaction)

crates/router/src/db/health_check.rs
```diff
@@ -0,0 +1,157 @@
...
+        logger::debug!("Redis get_key was successful");
+
+        redis_conn
+            .delete_key("test_key")
+            .await
+            .change_context(errors::HealthCheckRedisError::DeleteFailed)?;
+
+        logger::debug!("Redis delete_key was successful");
+
+        Ok(())
```
NishantJoshi00, line 108: Not sure, but do we need to follow the same structure as database query for redis? won't [`ping`](https://docs.rs/fred/latest/fred/interfaces/trait.ClientLike.html#method.ping) be enough?.

@inventvenkat @SanchithHegde what do you think?

crates/router/src/db/kafka_store.rs
```diff
@@ -2131,3 +2133,27 @@ impl AuthorizationInterface for KafkaStore {
             .await
     }
 }
+
+#[async_trait::async_trait]
+impl HealthCheckInterface for KafkaStore {
```
NishantJoshi00, line 2138: can we include kafka into the deep health check?

cc: @lsampras @ShivanshMathurJuspay

crates/router/src/routes/app.rs
```diff
@@ -253,9 +253,10 @@ pub struct Health;
 
 impl Health {
     pub fn server(state: AppState) -> Scope {
-        web::scope("")
+        web::scope("health")
             .app_data(web::Data::new(state))
-            .service(web::resource("/health").route(web::get().to(health)))
+            .service(web::resource("").route(web::get().to(health)))
+            .service(web::resource("/deep_check").route(web::post().to(deep_health_check)))
```
NishantJoshi00, line 259: does this need to be post?

crates/router/src/db/health_check.rs
```diff
@@ -0,0 +1,157 @@
...
+            let mut url = locker.host_rs.to_owned();
+            url.push_str("/health");
+            let request = services::Request::new(services::Method::Get, &url);
+            status_code = services::call_connector_api(state, request)
+                .await
+                .change_context(errors::HealthCheckLockerError::FailedToCallLocker)?
+                .map(|resp| resp.status_code)
+                .map_err(|err| err.status_code)
+                .unwrap_or_else(|code| code);
+        }
```
NishantJoshi00, line 127: Include the path into the exclude proxy list. Also, can we define the path as a separate constant?

NishantJoshi00 (changes requested):

crates/router/src/routes/app.rs
Chethan-rao: we r creating and deleting resources right so post might be valid

crates/router/src/db/health_check.rs
Chethan-rao: all the paths in `proxy_bypass_urls`?

NishantJoshi00 (approved):

SanchithHegde (approved):

PR #3237, opened 2024-01-04
fix(connector): [Stripe] capture error message and error code for failed payment, capture, void and refunds

AkshayaFoiger:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [x] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

Resolves [#3737](https://github.com/juspay/hyperswitch-cloud/issues/3737)

## How did you test it?

This following change effects error response handling of auth, refunds, void, and capture.  However for certain calls and payment method. We can't produce an error scenario. Hence, to ensure the changes doesn't break any flow, do the following testing.
1. Create a Bank Transfer payment with stripe. If possible try to fail it. 
2. Create a capture payment and capture it.
3. Create a capture payment and void it.
4. Create and confirm a payment and refund it.
5. Verify all payment methods with capture

## Checklist


crates/router/src/connector/stripe/transformers.rs
```diff
@@ -3829,3 +3917,81 @@ mod test_validate_shipping_address_against_payment_method {
         }
     }
 }
+
+fn build_last_payment_error_response<T>(
```
ArjunKarthik, line 3921: Use try_from instead of method

crates/router/src/connector/stripe/transformers.rs
```diff
@@ -3829,3 +3917,81 @@ mod test_validate_shipping_address_against_payment_method {
...
+                })
+                .or(Some(error_message.clone()))
+        }),
+        status_code: http_code,
+        attempt_status: None,
+        connector_transaction_id: Some(response_id),
+    })
+}
+
+fn build_error_response<T>(
```
ArjunKarthik, line 3957: Use try_from instead of method

crates/router/src/connector/stripe/transformers.rs
```diff
@@ -2355,21 +2359,33 @@ impl<F, T>
         let connector_metadata =
             get_connector_metadata(item.response.next_action.as_ref(), item.response.amount)?;
 
+        let status = enums::AttemptStatus::from(item.response.status);
```
ArjunKarthik, line 2362: The same logic has to be handled for PSync response also

ArjunKarthik (changes requested):

ArjunKarthik (dismissed):

srujanchikke (dismissed):

srujanchikke (approved):

PR #3278, opened 2024-01-08
fix: update amount_capturable based on intent_status and payment flow

hrithikesh026:
## Type of Change

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

wrote logic to update amount_capturable based intent_status and payment flow.

### Additional Changes

## Motivation and Context

This was necessary because connector may send processing status in the authorization response (amount_capturable will be set to 0). This will later change to requires_capture when force sync is done. At this point, with the existing implementation, amount_capturable will remain 0 and will not be updated to authorized_amount. This prevents the capture of the payment. (NMI connector).

## How did you test it?

Manual

Create a payment with manual capture
Screenshot 2024-01-09 at 4 54 16 PM

Force sync
Screenshot 2024-01-08 at 7 00 14 PM

Capture the payment
Screenshot 2024-01-08 at 7 00 19 PM

Force sync
Screenshot 2024-01-08 at 7 00 24 PM

## Checklist

### DIFF
--- a/crates/router/src/connector/utils.rs
+++ b/crates/router/src/connector/utils.rs
@@ -5,7 +5,6 @@ use api_models::{
     payments::{self, BankDebitBilling, OrderDetailsWithAmount},
 };
 use base64::Engine;
-use common_enums::IntentStatus;
 use common_utils::{
     date_time,
     errors::ReportSwitchExt,
@@ -29,9 +28,8 @@ use crate::{
     },
     pii::PeekInterface,
     types::{
-        self, api,
-        transformers::{ForeignFrom, ForeignTryFrom},
-        ApplePayPredecryptData, BrowserInformation, PaymentsCancelData, ResponseId,
+        self, api, transformers::ForeignTryFrom, ApplePayPredecryptData, BrowserInformation,
+        PaymentsCancelData, ResponseId,
     },
     utils::{OptionExt, ValueExt},
 };
@@ -95,9 +93,6 @@ pub trait PaymentResponseRouterData {
     ) -> enums::AttemptStatus
     where
         F: Clone;
-    fn get_amount_capturable<F>(&self, payment_data: &PaymentData<F>) -> Option<i64>
-    where
-        F: Clone;
 }
 
 impl<Flow, Request, Response> PaymentResponseRouterData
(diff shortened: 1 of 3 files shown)


Narayanbhat166 (approved):

crates/router/src/connector/utils.rs
```diff
@@ -130,6 +135,31 @@ where
...
+            IntentStatus::Processing
+            | IntentStatus::RequiresCustomerAction
+            | IntentStatus::RequiresMerchantAction
+            | IntentStatus::RequiresPaymentMethod
+            | IntentStatus::RequiresConfirmation
+            // terminal statuses
+            | IntentStatus::PartiallyCaptured
+            | IntentStatus::Succeeded
+            | IntentStatus::Failed
+            | IntentStatus::Cancelled => Some(0),
```
ArjunKarthik, line 156: Why are we returning 0 incase of Processing, RequiresCustomerAction, RequiresPaymentMethod, RequiresConfirmation?

RequiresCustomerAction - Incase of 3ds manual capture, while going for capture amount capturable will become 0 and capture will fail.

RequiresPaymentMethod, RequiresConfirmation - These are initial status for the payment intent incase of manual capture again the above scenario will happen.

Processing - If Payment attempt has AuthenticationPending, Authorizing, VoidInitiated, CaptureInitiated, Pending, Started, AuthenticationSuccessful, CodInitiated all are mapped to Processing status which again fail capture call for manual capture payment

ArjunKarthik (changes requested):

crates/router/src/connector/utils.rs
hrithikesh026: A payment can be captured only when the intent status is `requires_capture`. Payment can have `RequiresCustomerAction, RequiresPaymentMethod, RequiresConfirmation, Processing` before the status changes to `requires_capture`. In All other cases, `amount_capturable` should be 0 since the payment cannot be captured. Once the status changes to `requires_capture`, we will set `amount_capturable` as `total_authorized_amount()`.

jarnura (dismissed):

crates/router/src/connector/utils.rs
hrithikesh026: > A payment can be captured only when the intent status is `requires_capture`. Payment can have `RequiresCustomerAction, RequiresPaymentMethod, RequiresConfirmation, Processing` before the status changes to `requires_capture`. In All other cases, `amount_capturable` should be 0 since the payment cannot be captured. Once the status changes to `requires_capture`, we will set `amount_capturable` as `total_authorized_amount()`.

This solution was found to be not feasible. So Instead of deciding amount_capturable based only on intent_status. We should calculate amount_capturable based on intent status as well as the payment flow. Have made this change.

crates/router/src/types.rs
```diff
@@ -602,41 +612,149 @@ impl Capturable for PaymentsAuthorizeData {
...
+    ) -> Option<i64>
+    where
+        F: Clone,
+    {
+        Some(self.total_amount)
+    }
+}
 impl Capturable for PaymentsSyncData {
-    fn get_capture_amount<F>(&self, payment_data: &PaymentData<F>) -> Option<i64>
+    fn get_captured_amount<F>(&self, payment_data: &PaymentData<F>) -> Option<i64>
```
ArjunKarthik, line 757: amount_capturable should be updated in PSync also, as payments can reach terminal status in these flows

ArjunKarthik (changes requested):

jarnura (approved):

ArjunKarthik (approved):

PR #3304, opened 2024-01-10
feat: add deep health check for scheduler

Chethan-rao:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

This PR adds support for performing a deeper health check of the scheduler which includes:

Database connection check with read, write and delete queries (Added support for database transaction too)
Redis connection check with set, get and delete key (Set key with expiry of 30sec)

The deep health check API returns a 200 status code if all the components are in good health, but returns 500 if at least one component's health is down

### Additional Changes

## How did you test it?

1. `/health`

image

2. `/health/ready`

image

3. `'health/ready` with some db error

image

Cannot be tested on sandbox as the endpoint is not exposed

## Checklist


NishantJoshi00 (approved):

SanchithHegde (dismissed):

crates/router/src/bin/scheduler.rs
```diff
@@ -37,7 +40,7 @@ async fn main() -> CustomResult<(), ProcessTrackerError> {
     // channel for listening to redis disconnect events
     let (redis_shutdown_signal_tx, redis_shutdown_signal_rx) = oneshot::channel();
     let state = Box::pin(routes::AppState::new(
-        conf,
+        conf.clone(),
```
NishantJoshi00, line 43: why was this clone required?

crates/router/src/bin/scheduler.rs
```diff
@@ -68,6 +71,16 @@ async fn main() -> CustomResult<(), ProcessTrackerError> {
...
 
+    #[allow(clippy::expect_used)]
+    let web_server = Box::pin(start_web_server(
+        state.clone(),
+        scheduler_flow_str.to_string(),
+    ))
+    .await
+    .expect("Failed to create the server");
+
+    tokio::spawn(web_server);
```
NishantJoshi00, line 82: Can we have something like
```suggestion
    tokio::spawn(move async {
    	web_server.await;
    	logger::error!("The health check probe stopped working!");
    });
```

NishantJoshi00 (changes requested):

crates/router/src/bin/scheduler.rs
Chethan-rao: not required. removing

crates/router/src/bin/scheduler.rs
```diff
@@ -76,6 +96,92 @@ async fn main() -> CustomResult<(), ProcessTrackerError> {
     Ok(())
 }
 
+pub async fn start_web_server(
+    state: routes::AppState,
+    service: String,
+) -> errors::ApplicationResult<Server> {
+    let server = state.conf.server.clone();
```
dracarys18, line 103: We should create a seperate config for `scheduler` server.

crates/router/src/bin/scheduler.rs
```diff
@@ -76,6 +96,92 @@ async fn main() -> CustomResult<(), ProcessTrackerError> {
...
+    let web_server = actix_web::HttpServer::new(move || {
+        actix_web::App::new().service(Health::server(state.clone(), service.clone()))
+    })
+    .bind((server.host.as_str(), server.port))?
```
dracarys18, line 107: Make the workers configurable. And keep the default to 1

```rust
workers(conf.scheduler_server.workers)
```

crates/router/src/bin/scheduler.rs
```diff
@@ -76,6 +96,92 @@ async fn main() -> CustomResult<(), ProcessTrackerError> {
...
+
+    Ok(web_server)
+}
+
+pub struct Health;
+
+impl Health {
+    pub fn server(state: routes::AppState, service: String) -> Scope {
+        web::scope("health")
+            .app_data(web::Data::new(state))
```
dracarys18, line 119: Why do we need the whole app state of `router` here. We need a smaller appstate with only necessary stuff.

crates/router/src/bin/scheduler.rs
```diff
@@ -76,6 +96,92 @@ async fn main() -> CustomResult<(), ProcessTrackerError> {
...
+}
+#[instrument(skip_all)]
+pub async fn deep_health_check(
+    state: web::Data<routes::AppState>,
+    service: web::Data<String>,
+) -> impl actix_web::Responder {
+    let report = deep_health_check_func(state, service).await;
+    match report {
+        Ok(response) => {
+            services::http_response_json(serde_json::to_string(&response).unwrap_or_default())
```
dracarys18, line 139: Remove `unwrap_or_default` and handle errors properly

dracarys18, line 139: Instead you can use `ApplicationResponse::Json` as well

dracarys18 (changes requested):

crates/router/src/bin/scheduler.rs
Chethan-rao: The `HealthCheckInterface` is implemented on AppState itself so cannot ignore it. Also it was present in scheduler before

dracarys18 (approved):
LGTM

config/config.example.toml
```diff
@@ -298,6 +298,11 @@ lower_fetch_limit = 1800          # Lower limit for fetching entries from redis
 lock_key = "PRODUCER_LOCKING_KEY" # The following keys defines the producer lock that is created in redis with
 lock_ttl = 160                    # the ttl being the expiry (in seconds)
 
+[scheduler.server]
+port = 3000
+host = "127.0.0.1"
+workers = 1
```
SanchithHegde, line 304: Would be preferable if you can add comments about the same here, and in the `deployment/scheduler/{consumer,producer}.toml` files.

SanchithHegde (commented):

SanchithHegde (approved):

dracarys18 (approved):

NishantJoshi00 (approved):

PR #3360, opened 2024-01-16
refactor(blocklist): separate utility function & kill switch for validating data in blocklist

prajjwalkumar17:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [x] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

Refactoring the way blocklist is written making the code separated into utility functions rather that being on the core flow itself.
Moreover this will allow us to enable and disable the blocklist feature based on each merchant's requirements.
The steps to test out this feature can be found in the mentioned issue.

### Additional Changes

## How did you test it?

Screenshot 2024-01-18 at 17 20 10

## Checklist


crates/router/src/core/blocklist/utils.rs
```diff
@@ -357,3 +361,169 @@ async fn delete_card_bin_blocklist_entry(
...
+        ));
+    }
+
+    let blocklist_lookups = futures::future::join_all(blocklist_futures).await;
+
+    if blocklist_lookups.iter().any(|x| x.is_ok()) {
+        intent_status = storage_enums::IntentStatus::Failed;
+        attempt_status = storage_enums::AttemptStatus::Failure;
+        is_pm_blocklisted = true;
+    }
```
vspecky, line 483: Update this to log any DB errors that might have been returned.
```suggestion
    if blocklist_lookups.iter().any(|x| x.is_ok()) {
        intent_status = storage_enums::IntentStatus::Failed;
        attempt_status = storage_enums::AttemptStatus::Failure;
        is_pm_blocklisted = true;
    }
```

vspecky (changes requested):

crates/data_models/src/payments/payment_attempt.rs
```diff
@@ -308,6 +308,14 @@ pub enum PaymentAttemptUpdate {
         error_message: Option<Option<String>>,
         updated_by: String,
     },
+    BlocklistUpdate {
+        status: storage_enums::AttemptStatus,
+        error_code: Option<Option<String>>,
+        error_message: Option<Option<String>>,
+        updated_by: String,
+        connector: Option<Option<String>>,
+        merchant_connector_id: Option<Option<String>>,
```
vspecky, line 317: You don't need to put the `connector` and `merchant_connector_id` fields in the enum variant for blocklist update. Simply set the fields to `Some(None)` when converting the `PaymentAttemptUpdate` enum to `PaymentAttemptUpdateInternal`.

crates/diesel_models/src/payment_attempt.rs
```diff
@@ -218,6 +218,14 @@ pub enum PaymentAttemptUpdate {
         cancellation_reason: Option<String>,
         updated_by: String,
     },
+    BlocklistUpdate {
+        status: storage_enums::AttemptStatus,
+        error_code: Option<Option<String>>,
+        error_message: Option<Option<String>>,
+        updated_by: String,
+        connector: Option<String>,
+        merchant_connector_id: Option<String>,
```
vspecky, line 227: Most probably not needed as well.

crates/diesel_models/src/payment_attempt.rs
```diff
@@ -565,6 +573,22 @@ impl From<PaymentAttemptUpdate> for PaymentAttemptUpdateInternal {
...
+                error_message,
+                updated_by,
+                merchant_connector_id,
+            } => Self {
+                status: Some(status),
+                error_code,
+                connector: Some(connector),
+                error_message,
+                updated_by,
+                merchant_connector_id: Some(merchant_connector_id),
```
vspecky, line 589:
```suggestion
                connector: Some(None),
                error_message,
                updated_by,
                merchant_connector_id: Some(None),
```
You can hardcode this here since we want to ensure that the `connector` and `merchant_connector_id` fields are set to null in the DB.

crates/router/src/core/blocklist/utils.rs
```diff
@@ -357,3 +365,255 @@ async fn delete_card_bin_blocklist_entry(
...
+    if let Some(extended_card_bin_fingerprint) = extended_card_bin_fingerprint.as_ref() {
+        blocklist_futures.push(db.find_blocklist_lookup_entry_by_merchant_id_fingerprint(
+            merchant_id,
+            extended_card_bin_fingerprint,
+        ));
+    }
+
+    let blocklist_lookups = futures::future::join_all(blocklist_futures).await;
+
+    let mut db_operations_successful = false;
```
vspecky, line 479:
```suggestion
    let mut should_payment_be_blocked = false;
```
Appropriate naming

crates/router/src/core/blocklist/utils.rs
```diff
@@ -357,3 +365,255 @@ async fn delete_card_bin_blocklist_entry(
...
+        .attach_printable(
+            "Failed to update status in Payment Intent to failed due to it being blocklisted",
+        )?;
+
+        // If payment is blocked not showing connector details
+        let attempt_update = storage::PaymentAttemptUpdate::BlocklistUpdate {
+            status: common_enums::AttemptStatus::Failure,
+            error_code: Some(Some("HE-03".to_string())),
+            error_message: Some(Some(
+                "Failed to update status in Payment Attempt to failed, due to it being blocklisted"
```
vspecky, line 513:
```suggestion
                "This payment was blocked"
```
This message will be displayed to the merchant, so have an appropriate error message.

crates/router/src/core/payments.rs
```diff
@@ -1070,6 +1070,33 @@ where
...
+        }
+    };
+
+    if blocklist_guard_enabled {
+        operation
+            .to_domain()?
+            .guard_payment_against_blocklist(state, merchant_account, payment_data)
+            .await?;
+    }
+
```
vspecky, line 1099: Put this in a separate function.

vspecky (changes requested):

crates/api_models/src/errors/actix.rs
```diff
@@ -17,6 +17,7 @@ impl actix_web::ResponseError for ApiErrorResponse {
             Self::MethodNotAllowed(_) => StatusCode::METHOD_NOT_ALLOWED,
             Self::NotFound(_) => StatusCode::NOT_FOUND,
             Self::BadRequest(_) => StatusCode::BAD_REQUEST,
+            Self::GenericError(_) => StatusCode::OK,
```
Narayanbhat166, line 20: maybe rename this to `DomainError`

crates/router/src/core/blocklist/utils.rs
Narayanbhat166: This payment method has been blocked

crates/router/src/core/payments.rs
```diff
@@ -1229,6 +1232,45 @@ where
...
+        }
+    };
+
+    if blocklist_guard_enabled {
+        Ok(operation
+            .to_domain()?
+            .guard_payment_against_blocklist(state, merchant_account, payment_data)
+            .await?)
+    } else {
+        Ok(true)
```
vspecky, line 1270:
```suggestion
        Ok(false)
```
By default, payment is not blocked.

vspecky (changes requested):

vspecky:
Also postman tests are failing. Please check these.

crates/router/src/core/blocklist/utils.rs
```diff
@@ -357,3 +365,250 @@ async fn delete_card_bin_blocklist_entry(
...
+        .to_not_found_response(errors::ApiErrorResponse::PaymentNotFound)
+        .attach_printable(
+            "Failed to update status in Payment Intent to failed due to it being blocklisted",
+        )?;
+
+        // If payment is blocked not showing connector details
+        let attempt_update = storage::PaymentAttemptUpdate::BlocklistUpdate {
+            status: common_enums::AttemptStatus::Failure,
+            error_code: Some(Some("HE-03".to_string())),
+            error_message: Some(Some("This payment was blocked".to_string())),
```
Narayanbhat166, line 512:
```suggestion
            error_message: Some(Some("This payment method was blocked".to_string())),
```

crates/router/src/core/blocklist/utils.rs
```diff
@@ -357,3 +365,250 @@ async fn delete_card_bin_blocklist_entry(
...
+        )
+        .await
+        .to_not_found_response(errors::ApiErrorResponse::PaymentNotFound)
+        .attach_printable(
+            "Failed to update status in Payment Attempt to failed, due to it being blocklisted",
+        )?;
+
+        Err(errors::ApiErrorResponse::PaymentBlockedError {
+            code: 200,
+            message: "This payment was blocked".to_string(),
```
Narayanbhat166, line 528:
```suggestion
            message: "This payment method is blocked".to_string(),
```

Narayanbhat166:
Hey @vspecky, I feel that the error should be handled gracefully by passing a field or a flag saying that the payment method is blocked and all the further steps should read this value and skip the step if payment method is blocked, the status should then be updated in the `payment_response_update_tracker`. What do you feel about this?

vspecky:
> Hey @vspecky, I feel that the error should be handled gracefully by passing a field or a flag saying that the payment method is blocked and all the further steps should read this value and skip the step if payment method is blocked, the status should then be updated in the `payment_response_update_tracker`. What do you feel about this?

@Narayanbhat166 I don't think there's an issue with early return. If we wanted to do it gracefully, we'd have to add guards everywhere in the subsequent flows in the form of `if !blocked { do stuff and return data } else { return some default data }`. I don't see a point in going through all that dev trouble and processing if we know we're gonna return an (irrecoverable) error.

vspecky (approved):

Narayanbhat166:
@vspecky the problem lies in the way this error is raised, in fact this is not an application error, but we are made to handle it as an application error to short circuit the flow and save some of the dev effort.

crates/router/src/core/blocklist/utils.rs
```diff
@@ -357,3 +365,250 @@ async fn delete_card_bin_blocklist_entry(
...
+                    |err| {
+                        logger::error!(error=?err);
+                        None
+                    },
+                    Some,
+                )
+            }
+            _ => None,
+        })
+        .map(hex::encode);
```
Narayanbhat166, line 452: this block of code seems to be repeated, match is being made on the same field and only one match arm is handled, instead create an iterator and apply this logic

vspecky:
> @vspecky the problem lies in the way this error is raised, in fact this is not an application error, but we are made to handle it as an application error to short circuit the flow and save some of the dev effort.

@Narayanbhat166 That makes sense. In the first place, do we even want to send an error? We should probably be sending a `PaymentsResponse` with the status as failed right? (`return Ok()` from `payments_operation_core`)

crates/router/src/core/payments.rs
```diff
@@ -1229,6 +1232,45 @@ where
...
+{
+    let merchant_id = &payment_data.payment_attempt.merchant_id;
+    let blocklist_enabled_key = format!("guard_blocklist_for_{merchant_id}");
+    let blocklist_guard_enabled = state
+        .store
+        .find_config_by_key_unwrap_or(&blocklist_enabled_key, Some("false".to_string()))
+        .await;
+
+    let blocklist_guard_enabled: bool = match blocklist_guard_enabled {
+        Ok(config) => serde_json::from_str(&config.config).unwrap_or(false),
```
Narayanbhat166, line 1253: log this error

crates/router/src/core/payments.rs
```diff
@@ -1229,6 +1232,45 @@ where
...
+
+        // If it is not present in db we are defaulting it to false
+        Err(inner) => {
+            if !inner.current_context().is_db_not_found() {
+                logger::error!("Error fetching guard blocklist enabled config {:?}", inner);
+            }
+            false
+        }
```
Narayanbhat166, line 1261: is this logic doing what you intend to do? There is a negation of the condition, and the return value is outside the if block.

vspecky (approved):

crates/router/src/core/payments.rs
prajjwalkumar17: @Narayanbhat166 we already have major refactorings ongoing for the above feature, which will result in most of the code becoming depricated. So we can possibly do all refactorings in the upcoming PRs, as this is critical to go.

Narayanbhat166 (approved):

PR #3426, opened 2024-01-23
fix: auto retry once for connection closed

inventvenkat:
## Type of Change

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description
Hyper crate has a bug which allows few calls to connect with a dead connections and fails to send request to the server.  This bug of hyper can be tracked with below link. 
`Error: IncompleteMessage: connection closed before message completed` - [Link](https://github.com/hyperium/hyper/issues/2136)

Temporary fix: Retry connection with the server by sending the request again when call fails with this issue. 

Permanent fix would be, whenever reqwest crate release a stable version with hyper's 1.0 version

[Another Link](https://epi052.github.io/feroxbuster-docs/docs/faq/connection-closed/)
   This is just due to the racy nature of networking. hyper has a connection pool of idle connections, and it selected one to send your request. Most of the time, hyper will receive the server’s FIN and drop the dead connection from its pool. But occasionally, a connection will be selected from the pool and written to at the same time the server is deciding to close the connection. Since hyper already wrote some of the request, it can’t really retry it automatically on a new connection, since the server may have acted already

### Additional Changes

## Motivation and Context
Few external api calls fails due to this connection close

## How did you test it?
Running the test case collections

## Checklist


crates/router/src/services/api.rs
```diff
@@ -607,29 +606,83 @@ pub async fn send_request(
...
+                            &[metrics_tag],
+                        )
+                        .await
+                    }
+                    None => Err(error),
+                }
+            } else {
+                Err(error)
+            }
+        }
```
NishantJoshi00, line 684:
```suggestion
        Err(error) if error.current_context() == &errors::ApiClientError::ConnectionClosed => {
                metrics::AUTO_RETRY_CONNECTION_CLOSED.add(&metrics::CONTEXT, 1, &[]);
                match cloned_send_request {
                    Some(cloned_request) => {
                        metrics_request::record_operation_time(
                            cloned_request,
                            &metrics::EXTERNAL_REQUEST_TIME,
                            &[metrics_tag],
                        )
                        .await
                    }
                    None => Err(error),
                }
        },
        err @ Err(_) => err
```

NishantJoshi00 (changes requested):
Other than that, looks good to me

jarnura (changes requested):

crates/router/src/services/api.rs
```diff
@@ -607,29 +606,83 @@ pub async fn send_request(
...
+    // Since hyper already wrote some of the request,
+    // it can’t really retry it automatically on a new connection, since the server may have acted already
+    match response {
+        Ok(response) => Ok(response),
+        Err(error) => {
+            if error.current_context() == &errors::ApiClientError::ConnectionClosed {
+                metrics::AUTO_RETRY_CONNECTION_CLOSED.add(&metrics::CONTEXT, 1, &[]);
+                match cloned_send_request {
+                    Some(cloned_request) => {
+                        metrics_request::record_operation_time(
```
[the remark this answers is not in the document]
inventvenkat: https://github.com/juspay/hyperswitch/pull/3426#issuecomment-1925604426

NishantJoshi00 (dismissed):

jarnura (approved):

crates/router/src/services/api.rs
```diff
@@ -607,32 +606,86 @@ pub async fn send_request(
...
+            match cloned_send_request {
+                Some(cloned_request) => {
+                    metrics_request::record_operation_time(
+                        cloned_request,
+                        &metrics::EXTERNAL_REQUEST_TIME,
+                        &[metrics_tag],
+                    )
+                    .await
+                }
+                None => Err(error),
```
SanchithHegde: Should we maybe log a message that the cloned request was `None`, or something along similar lines?

jarnura (approved):

SanchithHegde (approved):

Narayanbhat166:
@inventvenkat instead of doing this, what if we do not have any `idle_pools`, AFAIK the error is caused mainly because of the idle pools. Would we have any significant performance hit without idle pools?

inventvenkat:
> @inventvenkat instead of doing this, what if we do not have any `idle_pools`, AFAIK the error is caused mainly because of the idle pools. Would we have any significant performance hit without idle pools?

[pool_idle_timeout](https://docs.rs/reqwest/latest/reqwest/struct.ClientBuilder.html#method.pool_idle_timeout) disabling timeout didn't help this case.

PR #3455, opened 2024-01-25
refactor(webhooks): check `is_merchant_notified` field before sending webhook to merchant

### LINKED ISSUE #3408: [BUG] webhooks are not sent again if it fails to send the first time during Psync
### Bug Description

When Payments Sync / Payments Retrieve is called ( by scheduler ), if there is a state change then an outgoing webhook should be triggered notifying the merchant about the change of state. Now the code in place tries to do this
- Generate an idempotent `event_id`, This is to ensure that duplicate webhooks are not sent.
- Try to insert the event in events table, If it already exists then assume that webhook was already sent ( This is where the bug lies )
- If there is no record, then try to send the webhook to merchant.

Now there is a field in the [events table](https://github.com/juspay/hyperswitch/blob/main/crates/diesel_models/src/events.rs#L34) called `is_webhook_notified` which is used to track whether the webhook was already sent to merchant. When creating the event, this is set to false. Once we receive a response from the merchant webhook url, we set it to true if response was `Ok`.

Consider a scenario when Psync was called, event is created and inserted, but sending webhook failed. Now when another Psync is done, it will not send the webhook because the event already exists. It will not check for the `is_webhook_nofitied` field.

### Expected Behavior

The Second Psync should try to resend the webhook if `is_webhook_notified` is false.

### Actual Behavior

The second Psync will not care to resend the webhook

### Steps To Reproduce

- Provide a wrong webhook url when configuring the account.
- Create a payment, sending webhooks will fail.
- Reconfigure the right webhook url and then retrieve the payment again.
- Still the webhook will not be sent.

### Context For The Bug

_No response_

### Environment

Narayanbhat166:
## Type of Change

- [x] Refactoring

## Description

```mermaid
flowchart TD
    event_id --> CheckEventExists{check event exists}

    CheckEventExists --> |yes| CheckMerchantNotified{is_merchant_notified}
    CheckMerchantNotified --> |true| DoNotSendWebhook
    CheckMerchantNotified --> |false| SendWebhook{ send webhook to merchant }

    CheckEventExists --> |no| InsertNewEvent(InsertEvent)
    InsertNewEvent --> SendWebhook
    SendWebhook --> |success| UpdateEvent{ set is_merchant_notified to true}
```
This change will check for the `is_merchant_notified` field and send the webhook only if it is false.

### Additional Changes

## Motivation and Context

closes #3408 

## Checklist


vspecky (changes requested):

sai-harsha-vardhan (approved):

vspecky (approved):

Narayanbhat166:
Closing this PR since automatic webhook retries will handle this.

PR #3519, opened 2024-02-01
feat: add healthcheck for outgoing request

### LINKED ISSUE #3518: Deep health check for outgoing request

dracarys18:
## Type of Change

- [x] New feature

## Description

Closes #3518. Add healthcheck for outgoing calls

## Motivation and Context

Added a check for outgoing requests

## How did you test it?

- `curl --location 'http://localhost:8080/health/ready'`

Screenshot 2024-02-01 at 11 58 57 AM

## Checklist

### DIFF
--- a/crates/router/src/consts.rs
+++ b/crates/router/src/consts.rs
@@ -91,3 +91,6 @@ pub const MAX_SESSION_EXPIRY: u32 = 7890000;
 pub const MIN_SESSION_EXPIRY: u32 = 60;
 
 pub const LOCKER_HEALTH_CALL_PATH: &str = "/health";
+
+// URL for checking the outgoing call
+pub const OUTGOING_CALL_URL: &str = "https://www.google.com";
--- a/crates/router/src/consts.rs
+++ b/crates/router/src/consts.rs
@@ -93,4 +93,4 @@ pub const MIN_SESSION_EXPIRY: u32 = 60;
 pub const LOCKER_HEALTH_CALL_PATH: &str = "/health";
 
 // URL for checking the outgoing call
-pub const OUTGOING_CALL_URL: &str = "https://www.google.com";
+pub const OUTGOING_CALL_URL: &str = "https://api.stripe.com/healthcheck";
--- a/crates/router/src/core/health_check.rs
+++ b/crates/router/src/core/health_check.rs
@@ -4,7 +4,7 @@ use error_stack::ResultExt;
 use router_env::logger;
 
 use crate::{
-    consts::LOCKER_HEALTH_CALL_PATH,
+    consts,
     core::errors::{self, CustomResult},
     routes::app,
     services::api as services,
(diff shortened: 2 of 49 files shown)


Chethan-rao (dismissed):

crates/router/src/consts.rs
```diff
@@ -91,3 +91,6 @@ pub const MAX_SESSION_EXPIRY: u32 = 7890000;
 pub const MIN_SESSION_EXPIRY: u32 = 60;
 
 pub const LOCKER_HEALTH_CALL_PATH: &str = "/health";
+
+// URL for checking the outgoing call
+pub const OUTGOING_CALL_URL: &str = "https://www.google.com";
```
NishantJoshi00, line 96: There is a inherent issue here. What if the infra admin has a blocklist, which blocks https://www.google.com/ but has whitelisted the connector endpoints, then this will fail right?

crates/router/src/core/health_check.rs
```diff
@@ -108,4 +109,17 @@ impl HealthCheckInterface for app::AppState {
...
     }
+
+    async fn health_check_outgoing(&self) -> CustomResult<(), errors::HealthCheckOutGoing> {
+        let request = services::Request::new(services::Method::Get, consts::OUTGOING_CALL_URL);
+        services::call_connector_api(self, request)
+            .await
+            .map_err(|err| errors::HealthCheckOutGoing::OutGoingFailed {
+                message: err.to_string(),
+            })?
+            .ok();
```
NishantJoshi00, line 120: why are we ignoring the `ErrorResponse`?

NishantJoshi00 (changes requested):

crates/router/src/consts.rs
dracarys18: But what if that particular connector is down, we can't just say outgoing call is not working right? That's why I used google.com

crates/router/src/consts.rs
```diff
@@ -91,3 +91,6 @@ pub const MAX_SESSION_EXPIRY: u32 = 7890000;
 pub const MIN_SESSION_EXPIRY: u32 = 60;
 
 pub const LOCKER_HEALTH_CALL_PATH: &str = "/health";
+
+// URL for checking the outgoing call
+pub const OUTGOING_CALL_URL: &str = "https://api.stripe.com/healthcheck";
```
NishantJoshi00, line 96: There is still an edge case, where only this connector is down, we need to come up with a more permanent fix for this

crates/router/src/core/health_check.rs
```diff
@@ -108,4 +109,19 @@ impl HealthCheckInterface for app::AppState {
...
+    async fn health_check_outgoing(&self) -> CustomResult<(), errors::HealthCheckOutGoing> {
+        let request = services::Request::new(services::Method::Get, consts::OUTGOING_CALL_URL);
+        services::call_connector_api(self, request)
+            .await
+            .map_err(|err| errors::HealthCheckOutGoing::OutGoingFailed {
+                message: err.to_string(),
+            })?
+            .map_err(|_| errors::HealthCheckOutGoing::OutGoingFailed {
+                message: "Got a non 200 status while making outgoing request".to_string(),
+            })?;
```
NishantJoshi00, line 122: should be some how indicate what error we got from the connector?

NishantJoshi00 (changes requested):

crates/router/src/consts.rs
NishantJoshi00: This can be resolved as a separate issue

NishantJoshi00 (approved):

PR #3551, opened 2024-02-05
fix(healthcheck): do not return true as response if the check if not applicable

dracarys18:
## Type of Change

- [x] Bugfix

## Description

Don't return true on the places where checks or not applicable

## Motivation and Context

This fixes misconception in health check where it would return true even in case if the case is not checked.

## How did you test it?

- Disable locker in config
```bash
curl --location 'http://localhost:8080/health/ready'
```
Screenshot 2024-02-07 at 2 55 21 PM

**THIS CANNOT BE TESTED IN SANDBOX**

## Checklist


Chethan-rao (approved):

crates/api_models/src/health_check.rs
```diff
@@ -17,4 +17,19 @@ pub struct SchedulerHealthCheckResponse {
...
+}
+
+impl From<HealthState> for bool {
+    fn from(value: HealthState) -> Self {
+        match value {
+            HealthState::Running => true,
+            HealthState::Error | HealthState::NotApplicable => false,
+        }
+    }
+}
```
NishantJoshi00, line 33: Can we handle it in this way:
```
HealthState::Running => true,
HealthState::Error => false,
HealthState::NotApplicable => (the field not visible)
```

NishantJoshi00 (changes requested):

crates/api_models/src/health_check.rs
dracarys18: Done

NishantJoshi00 (approved):

Chethan-rao (approved):

PR #3646, opened 2024-02-13
feat: add cors rules to actix

dracarys18:
## Type of Change

- [x] Enhancement

## Description

Adds cors rules to actix

### Additional Changes

- Added `CorsSettings` config

## Motivation and Context

NA

## How did you test it?

- Configure allowed origins in `ROUTER__CORS__ORIGINS`.
- Add `Origin` header to that configured to your config and the request should return 200
```bash
curl --location 'http://localhost:8080/health' \
--header 'Origin: http://localhost:8080'
```
- And configure origin header to something else like `google.com`. It should return `Bad Request` with Response
`Origin is not allowed to make this request`

```bash
curl --location 'http://localhost:8080/health' \
--header 'Origin: https://google.com'
```

## Checklist

### DIFF
--- a/config/config.example.toml
+++ b/config/config.example.toml
@@ -328,9 +328,10 @@ key_id = "" # The AWS key ID used by the KMS SDK for decrypting data.
 region = "" # The AWS region used by the KMS SDK for decrypting data.
 
 [cors]
-max_age = 36
-origins = "http://localhost:8080"
-allowed_methods = "GET,POST,PUT,DELETE"
+max_age = 30 # Maximum time (in seconds) for which this CORS request may be cached.
+origins = "http://localhost:8080" # List of origins that are allowed to make requests.
+allowed_methods = "GET,POST,PUT,DELETE" # List of methods that are allowed
+wildcard_origin = false # If true, allows any origin to make requests
 
 # EmailClient configuration. Only applicable when the `email` feature flag is enabled.
 [email]
--- a/config/config.example.toml
+++ b/config/config.example.toml
@@ -327,6 +327,11 @@ paypal = { currency = "USD,INR", country = "US" }
 key_id = "" # The AWS key ID used by the KMS SDK for decrypting data.
 region = "" # The AWS region used by the KMS SDK for decrypting data.
 
+[cors]
+max_age = 36
+origins = "http://localhost:8080"
+allowed_methods = "GET,POST,PUT,DELETE"
+
 # EmailClient configuration. Only applicable when the `email` feature flag is enabled.
 [email]
 sender_email = "example@example.com" # Sender email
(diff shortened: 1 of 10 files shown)


config/config.example.toml
```diff
@@ -327,6 +327,11 @@ paypal = { currency = "USD,INR", country = "US" }
 key_id = "" # The AWS key ID used by the KMS SDK for decrypting data.
 region = "" # The AWS region used by the KMS SDK for decrypting data.
 
+[cors]
+max_age = 36
+origins = "http://localhost:8080"
+allowed_methods = "GET,POST,PUT,DELETE"
```
SanchithHegde, line 333: Please document these configuration variables. Also, specify unit for `max_age`.

crates/router/src/configs/settings.rs
```diff
@@ -242,6 +243,15 @@ pub struct DummyConnector {
     pub discord_invite_url: String,
 }
 
+#[derive(Debug, Deserialize, Clone)]
+pub struct CorsSettings {
+    #[serde(deserialize_with = "deserialize_hashset")]
+    pub origins: HashSet<String>,
```
SanchithHegde, line 249: Would this allow us to specify `*`? What if someone self-hosting wants to allow all origins?

crates/router/src/configs/settings.rs
```diff
@@ -242,6 +243,15 @@ pub struct DummyConnector {
...
+    pub max_age: usize,
+    #[serde(deserialize_with = "deserialize_hashset")]
+    pub allowed_methods: HashSet<String>,
```
SanchithHegde, line 252: Include `origins` and `allowed_methods` in `with_list_parse_key()`.

https://github.com/juspay/hyperswitch/blob/774a0322aa4b36d87b122e47cd893383e262de12/crates/router/src/configs/settings.rs#L675-L679

crates/router/src/configs/settings.rs
```diff
@@ -242,6 +243,18 @@ pub struct DummyConnector {
     pub discord_invite_url: String,
 }
 
+#[derive(Debug, Deserialize, Clone)]
+pub struct CorsSettings {
+    #[serde(deserialize_with = "deserialize_hashset")]
+    #[serde(default)]
+    pub origins: HashSet<String>,
```
SanchithHegde, line 250:
```suggestion
    #[serde(default, deserialize_with = "deserialize_hashset")]
    pub origins: HashSet<String>,
```

SanchithHegde (approved):

crates/router/src/cors.rs
```diff
@@ -1,19 +1,21 @@
...
+        .max_age(config.max_age);
 
-    FIXME : I don't know what to put here
-    .allowed_origin_fn(|origin, _req_head| origin.as_bytes().starts_with(b"http://localhost"))
+    if config.wildcard_origin {
+        cors = cors.allow_any_origin()
+    } else {
+        for origin in &config.origins {
+            cors = cors.allowed_origin(origin);
+        }
```
NishantJoshi00, line 17:
```suggestion
    config.wildcard_origin
        .then(|| cors.allow_any_origin())
        .unwrap_or(|| cors.allowed_origin(origin))
```

Can something like this be done?

just a nit

NishantJoshi00 (approved):

PR #3764, opened 2024-02-22
feat(router): add connector mit related columns to the payment methods table

vspecky:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

This PR adds 3 columns to the `payment_methods` table, with the purpose of enabling MIT Transaction Orchestration in the near future.
- `connector_mandate_details`: Intended to store Connector MIT details such as the mandate ID
- `customer_acceptance`: Intended to store customer consent and acceptance details
- `status`: The status of the payment method (Can be `processing` initially for connector MITs that rely on asynchronous flows and/or time-based factors for setup)

#### Flows Affected
- **List Payment Methods for Customer**: This was updated such that only payment methods with status `active` are listed for the customer. To make this backwards compatible, the default status for new saved cards is set to `active`.

### Additional Changes

## Motivation and Context

This change is important for setting up MIT orchestration in the future.

## How did you test it?

Local testing, postman tests. The saved card payments flow needs to work as usual which is covered by Postman tests.

When saving a card, the values of the added columns are set to default for now. This is NULL for `connector_mandate_details` and `customer_acceptance`, and `active` for `status`.
image

Since the default for the status column is set to `active`, the `List Payment Methods for Customer` flow works as intended for saved payment methods.
image

## Checklist


crates/common_enums/src/enums.rs
```diff
@@ -1177,6 +1178,28 @@ pub enum PaymentMethodIssuerCode {
...
+)]
+#[router_derive::diesel_enum(storage_type = "db_enum")]
+#[strum(serialize_all = "snake_case")]
+#[serde(rename_all = "snake_case")]
+pub enum PaymentMethodStatus {
+    Active,
+    Inactive,
+    Processing,
+}
+
```
Chethan-rao, line 1202: Can we derive `Default` and make `Active` status as default variant

vspecky: @Chethan-rao The default when creating a payment method entry in DB is set to `Active` for seamless backwards compatibility. It's not the actual default value for the enum since payment methods can start out as either `Active` (on-session cards) or `Processing` (3DS Cards or MIT setup with any asynchronous flows) which we will handle with checks in the upcoming MIT orchestration feature.

crates/router/src/db/payment_method.rs
```diff
@@ -187,6 +215,33 @@ impl PaymentMethodInterface for MockDb {
...
+                    && pm.status == status
+            })
+            .cloned()
+            .collect();
+
+        if payment_methods_found.is_empty() {
+            Err(errors::StorageError::ValueNotFound(
+                "cannot find payment methods".to_string(),
+            ))
+            .into_report()
```
Narayanbhat166, line 239: does the original implementation ( on diesel ) return this error if it finds an empty vec?

vspecky: This piece of code was copied over verbatim from the list payment methods diesel query.

Chethan-rao (approved):

Chethan-rao (approved):

Narayanbhat166 (approved):

PR #3997, opened 2024-03-06
fix(webhooks): abort outgoing webhook retry task if webhook URL is not available in business profile

### LINKED ISSUE #3995: [BUG] Outgoing webhook retry scheduler tasks remain in pending state if merchant webhook URL is not configured
### Bug Description

In the outgoing webhook retry process tracker (scheduler) workflow introduced in #3842, the process tracker entries added before the webhook is attempted to be delivered remain in the `process_started` status and `Pending` business status, if either:

1. The `webhook_details` field is not provided in the business profile (or merchant account with default profile).
2. The `webhook_details` is provided but the `webhook_url` within the `webhook_details` object is not provided.
3. The business profile initially had a webhoook URL configured but the profile was updated later to remove the webhook URL (or webhook details).

Since in any of the above cases the URL to deliver the webhook to the merchant is unavailable, the task can be aborted. Another optimization is to not create the outgoing webhook event (in the `events` database table) at all, if the merchant webhook URL is not available.

### Expected Behavior

- The event (and corresponding process tracker task) should not be created in the `events` table if the merchant webhook URL has not been configured.
- In case the business profile initially had a webhook URL configured and was later removed from the business profile, the retry tasks for previously failed webhooks should be aborted.

### Actual Behavior

The process tracker task remains in the `process_started` status and `Pending` business status.

### Steps To Reproduce

1. Create a merchant account without the `webhook_details` object or without the `webhook_url` field, using the API.
2. Create an API key, and create a merchant connector account.
3. Make a successful payment.
4. Query the `process_tracker` table in the database to search for workflows with the runner `OUTGOING_WEBHOOK_RETRY_WORKFLOW` and having the payment ID in the name:

   ```sql
   SELECT * FROM process_tracker WHERE runner = 'OUTGOING_WEBHOOK_RETRY_WORKFLOW' AND id LIKE '%pay_1234%';
   ```

   The `status` is `process_started` and the `business_status` of the task is `Pending`.

SanchithHegde:
## Type of Change

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

This PR fixes a bug in the outgoing webhook retry workflow where process tracker tasks remained in pending state if merchant webhook URL was not configured in the business profile. For more information, see: #3995.

In addition, this PR includes these changes:

1. Updates the process tracker task ID format to include the event ID instead of the primary resource ID.
2. Removes unused enum variants from the `WebhooksFlowError` error enum.

The PR can easily be reviewed one commit at a time.

## Motivation and Context

Fixes #3995.

## How did you test it?

Locally, by running `router`, `producer` and `consumer`, and testing for these three cases:

1. For a business profile with webhook URL configured, the behavior should remain as before: webhooks should either be delivered successfully in the first attempt, or scheduled for later retries.
   1. Screenshot of a webhook delivered in the initial attempt (business status is `INITIAL_DELIVERY_ATTEMPT_SUCCESSFUL`):

      Screenshot of webhook delivery attempt successful in the first try

   2. Screenshot of a webhook delivered during a retry attempt (business status is `COMPLETED_BY_PT` and retry count is positive):

      Screenshot of webhook delivery attempt successful during retry

2. For a business profile without `webhook_url` (or `webhook_details`) configured, the event is not created in the `events` table (and no process tracker task), and the `router` service displays a log line about the same:

   Screenshot of log message indicating failure in retrieving merchant webhook URL

   The absence of the event in the `events` table and the process tracker task in the `process_tracker` table can be confirmed using the SQL queries, no rows should be returned in either case:

   ```sql
   SELECT * FROM events WHERE event_id = 'event_id';
   SELECT * FROM process_tracker WHERE runner = 'OUTGOING_WEBHOOK_RETRY_WORKFLOW' AND id LIKE '%event_id%';
   ```

3. For a business profile with a webhook URL configured initially but later updated to not include the `webhook_url` field  or the `webhook_details` object, the task is aborted by the consumer:

   1. The business status is `FAILURE`, and retry count is positive:

      Screenshot of task aborted during a retry attempt

   2. In addition, the `consumer` service displays a log line about the task being aborted.

      Screenshot of log message indicating task being aborted

The process tracker tasks in the above cases can be queried by using the SQL command, substituting the resource ID: 

```sql
SELECT * FROM process_tracker WHERE runner = 'OUTGOING_WEBHOOK_RETRY_WORKFLOW' AND id LIKE '%pay_1234%';
```

## Checklist

### DIFF
--- a/crates/router/src/core/errors.rs
+++ b/crates/router/src/core/errors.rs
@@ -259,7 +259,7 @@ pub enum WebhooksFlowError {
     #[error("Webhook details for merchant not configured")]
     MerchantWebhookDetailsNotFound,
     #[error("Merchant does not have a webhook URL configured")]
-    MerchantWebhookURLNotConfigured,
+    MerchantWebhookUrlNotConfigured,
     #[error("Webhook event updation failed")]
     WebhookEventUpdationFailed,
     #[error("Outgoing webhook body signing failed")]
@@ -276,6 +276,25 @@ pub enum WebhooksFlowError {
     OutgoingWebhookProcessTrackerTaskUpdateFailed,
     #[error("Failed to schedule retry attempt for outgoing webhook")]
     OutgoingWebhookRetrySchedulingFailed,
+}
+
+impl WebhooksFlowError {
+    pub(crate) fn is_webhook_delivery_retryable_error(&self) -> bool {
+        match self {
+            Self::MerchantConfigNotFound
+            | Self::MerchantWebhookDetailsNotFound
+            | Self::MerchantWebhookUrlNotConfigured => false,
+
+            Self::WebhookEventUpdationFailed
+            | Self::OutgoingWebhookSigningFailed
+            | Self::CallToMerchantFailed
+            | Self::NotReceivedByMerchant
+            | Self::DisputeWebhookValidationFailed
+            | Self::OutgoingWebhookEncodingFailed
+            | Self::OutgoingWebhookProcessTrackerTaskUpdateFailed
+            | Self::OutgoingWebhookRetrySchedulingFailed => true,
+        }
+    }
 }
 
 #[derive(Debug, thiserror::Error)]
--- a/crates/router/src/core/errors.rs
+++ b/crates/router/src/core/errors.rs
@@ -260,38 +260,18 @@ pub enum WebhooksFlowError {
     MerchantWebhookDetailsNotFound,
     #[error("Merchant does not have a webhook URL configured")]
     MerchantWebhookURLNotConfigured,
-    #[error("Payments core flow failed")]
-    PaymentsCoreFailed,
-    #[error("Refunds core flow failed")]
-    RefundsCoreFailed,
-    #[error("Dispuste core flow failed")]
-    DisputeCoreFailed,
-    #[error("Webhook event creation failed")]
-    WebhookEventCreationFailed,
     #[error("Webhook event updation failed")]
     WebhookEventUpdationFailed,
     #[error("Outgoing webhook body signing failed")]
     OutgoingWebhookSigningFailed,
-    #[error("Unable to fork webhooks flow for outgoing webhooks")]
-    ForkFlowFailed,
     #[error("Webhook api call to merchant failed")]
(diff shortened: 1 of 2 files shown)


crates/router/src/core/errors.rs
```diff
@@ -259,45 +259,44 @@ pub enum WebhooksFlowError {
     #[error("Webhook details for merchant not configured")]
     MerchantWebhookDetailsNotFound,
     #[error("Merchant does not have a webhook URL configured")]
-    MerchantWebhookURLNotConfigured,
-    #[error("Payments core flow failed")]
-    PaymentsCoreFailed,
```
sahkal, line 264: Are these unused, why are these removed?

SanchithHegde: Yes, the variants were not used anywhere. We can always add new variants as and when required.

Narayanbhat166 (approved):

crates/router/src/core/webhooks.rs
```diff
@@ -817,21 +828,30 @@ async fn trigger_webhook_to_merchant<W: types::OutgoingWebhookType>(
...
-        .map(ExposeInterface::expose)?;
+    let webhook_url = match (
+        get_webhook_url_from_business_profile(&business_profile),
+        process_tracker.clone(),
+    ) {
+        (Ok(webhook_url), _) => Ok(webhook_url),
+        (Err(error), Some(process_tracker)) => {
+            if !error
+                .current_context()
+                .is_webhook_delivery_retryable_error()
```
hrithikesh026, line 839: Is this if condition necessary?
`get_webhook_url_from_business_profile()` will return Err only when merchant has not configured the webhook URL. Why is there a need for calling `is_webhook_delivery_retryable_error()`

SanchithHegde: The current function is only aware that the error returned is `WebhooksFlowError` and nothing more. It's always safer to be checking if the error is something that we can retry webhook deliveries for.

PR #4025, opened 2024-03-08
feat(router): Create a merchant config for enable processor agnostic MIT

ShankarSinghC:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

 This pr adds a merchant config (`pg_agnostic_mit`) to enable usage of Network Transaction/Reference ID in recurring `MIT` payments. This will be a `business profile` level config that a merchant specifies.

### Additional Changes

## How did you test it?

Set processor agnostic MIT config as `true`
image
config status from db
image
Set processor agnostic MIT config as `false`
image
config status from db
image

```
curl --location 'http://localhost:8080/routing/business_profile/{business_profile_id}/configs/pg_agnostic_mit' \
--header 'api-key: api-key' \
--data '{
    "enabled": true
}'
```

## Checklist


crates/router/src/core/routing.rs
```diff
@@ -818,6 +817,37 @@ pub async fn retrieve_linked_routing_config(
     }
 }
 
+pub async fn create_connector_agnostic_mandate_config(
```
vspecky, line 820: Make this an `upsert` logic instead.
```rust
let config = state.find_config();
match config {
    Ok(conf) => { update },
    Err(err) if err.is_db_not_found() => { insert },
    Err(err) => { throw error }
}
```

vspecky: additionally, add logic to the "create business profile" core function to insert a default value to the configs table from the start.

vspecky (changes requested):

crates/router/src/core/routing.rs
ShankarSinghC: This function `find_config_by_key_unwrap_or` has the same logic.

Aprabhat19, line 820: Nit:
```suggestion
pub async fn create_connector_agnostic_mandate_config(
```
pub async fn upsert_connector_agnostic_mandate_config(

Aprabhat19 (approved):

Aprabhat19 (approved):

Chethan-rao (approved):

crates/router/src/routes/app.rs
```diff
@@ -437,6 +437,10 @@ impl Routing {
                         )
                     })),
             )
+            .service(
+                web::resource("/business_profile/{business_profile_id}/configs/pg_agnostic_mit")
```
Narayanbhat166, line 441: this endpoint should not have business profile in the path params

PR #4134, opened 2024-03-19
feat(payment_methods): Client secret implementation in payment method…

Sarthak1799:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

Implemented client secret auth for /payment_methods

### Additional Changes

## How did you test it?

1. Create a customer
```
curl --location --request POST 'http://localhost:8080/customers' \
--header 'api-key: dev_y5XDvbHKvRI6K2dAhTgXvNgQykJiqLXeYtSysCQzPGBGz3RZoUPjPcfD1LW8teYO' \
--data-raw '{
    "email": "guest@example.com",
    "name": "John Doe1",
    "phone": "999999999",
    "phone_country_code": "+65",
    "description": "First customer",
    "address": {
        "line1": "1467",
        "line2": "Harrison Street",
        "line3": "Harrison Street",
        "city": "San Fransico",
        "state": "California",
        "zip": "94122",
        "country": "US",
        "first_name": "John",
        "last_name": "Doe"
    },
    "metadata": {
        "udf1": "value1",
        "new_customer": "true",
        "login_date": "2019-09-10T10:11:12Z"
    }
}'
```
2. Create a payment method
```
curl --location --request POST 'http://localhost:8080/payment_methods' \
--header 'api-key: dev_63yd7Y0i2CvRcJYKfXcIE8wAva0b5YMtKHr6oTpcnggVmWqsK0Q6CvoKB1WiXoTz' \
--data-raw '{
    "customer_id": "cus_RHnmMuG8mKhICwigyUeP"
}'
```
Response - 
```
{
    "merchant_id": "sarthak",
    "customer_id": "cus_RHnmMuG8mKhICwigyUeP",
    "payment_method_id": "pm_DC74n7ocq8SKGvNVai0M",
    "payment_method": null,
    "payment_method_type": null,
    "card": null,
    "recurring_enabled": false,
    "payment_experience": null,
    "metadata": null,
    "created": null,
    "last_used_at": null,
    "client_secret": "pm_DC74n7ocq8SKGvNVai0M_secret_9wUZYKaUcORN13IdGCOo"
}
```
DB entry - 
image

3. Save the payment method with client_secret
```
curl --location --request POST 'http://localhost:8080/payment_methods/pm_DC74n7ocq8SKGvNVai0M/save' \
--header 'api-key: pk_dev_660e47ec672f43849d6b0403fcaf76ac' \
--data-raw '{
    "payment_method": "card",
    "client_secret": "pm_DC74n7ocq8SKGvNVai0M_secret_9wUZYKaUcORN13IdGCOo",
    "customer_id": "cus_RHnmMuG8mKhICwigyUeP",
    "payment_method_data": {
        "card": {
            "card_exp_month": "10",
            "card_exp_year": "2026",
            "card_holder_name": "joseph"
        }
    }
}'
```
Response - 
```
{
    "merchant_id": "sarthak",
    "customer_id": "cus_RHnmMuG8mKhICwigyUeP",
    "payment_method_id": "pm_DC74n7ocq8SKGvNVai0M",
    "payment_method": "card",
    "payment_method_type": null,
    "card": {
        "scheme": null,
        "issuer_country": null,
        "last4_digits": "4242",
        "expiry_month": "10",
        "expiry_year": "2026",
        "card_token": null,
        "card_holder_name": "joseph",
        "card_fingerprint": null,
        "nick_name": null,
        "card_network": null,
        "card_isin": "424242",
        "card_issuer": null,
        "card_type": null,
        "saved_to_locker": true
    },
    "recurring_enabled": false,
    "payment_experience": [
        "redirect_to_url"
    ],
    "metadata": null,
    "created": "2024-04-03T13:02:00.926Z",
    "last_used_at": "2024-04-03T13:02:00.926Z",
    "client_secret": "pm_DC74n7ocq8SKGvNVai0M_secret_9wUZYKaUcORN13IdGCOo"
}
```
DB entry - 
image

## Checklist


crates/api_models/src/payment_methods.rs
```diff
@@ -65,6 +65,12 @@ pub struct PaymentMethodCreate {
     #[cfg(feature = "payouts")]
     #[schema(value_type = Option<Wallet>)]
     pub wallet: Option<payouts::Wallet>,
+
+    /// For Client based calls
+    pub client_secret: Option<String>,
+
+    /// Payment method data to be passed
+    pub payment_method_data: Option<PaymentMethodData>,
```
Chethan-rao, line 73: can we remove the already existing variants at root level? I think its safe to remove it as long as nobody is using this api. Or are u forcing client to pass this field in order to perform client based calls?

crates/router/src/core/payment_methods/cards.rs
```diff
@@ -216,6 +225,216 @@ pub async fn get_or_insert_payment_method(
...
+    key_store: &domain::MerchantKeyStore,
+) -> errors::RouterResponse<api::PaymentMethodResponse> {
+    let db = &*state.store;
+    let merchant_id = &merchant_account.merchant_id;
+    let customer_id = req.customer_id.clone().get_required_value("customer_id")?;
+
+    if req.payment_method_data.is_some() {
+        add_payment_method_data(state, req, merchant_account, key_store).await
+    } else if req.card.is_some() || req.bank_transfer.is_some() || req.wallet.is_some() {
+        add_payment_method(state, req, merchant_account, key_store).await
```
Chethan-rao, line 242: I can create client_secret and pass payment method data in the already existing field `card` in which case it will go to `else if` block creating another payment methods entry in db?

Chethan-rao (changes requested):

crates/api_models/src/payment_methods.rs
Sarthak1799: We wanted to keep the existing interface the API unaltered, while allowing the client based calls. Hence added new fields. For client based calls, client secret has to be passed along with relevant `payment_method_data `

crates/router/src/core/payment_methods/cards.rs
Sarthak1799: the `else if` block is meant for backwards compatibility, so the current version of API can also be supported while also enabling client based calls using `payment_method_data`

crates/router/src/core/payment_methods/cards.rs
```diff
@@ -216,6 +225,216 @@ pub async fn get_or_insert_payment_method(
...
+            if let Ok((mut pm_resp, duplication_check)) = resp {
+                if duplication_check.is_some() {
+                    let pm_update = storage::PaymentMethodUpdate::StatusUpdate {
+                        status: Some(enums::PaymentMethodStatus::Inactive),
+                    };
+
+                    db.update_payment_method(payment_method, pm_update)
+                        .await
+                        .change_context(errors::ApiErrorResponse::InternalServerError)
+                        .attach_printable("Failed to add payment method in db")?;
```
Chethan-rao, line 362: why can't we delete this entry created instead of marking the status to inactive?

Sarthak1799: We ideally do not want to delete any entry, the status is supposed to indicate it's purpose. 
cc: @vspecky

crates/router/src/core/payment_methods/cards.rs
```diff
@@ -216,6 +225,216 @@ pub async fn get_or_insert_payment_method(
...
+                    status: Some(enums::PaymentMethodStatus::Inactive),
+                };
+
+                db.update_payment_method(payment_method, pm_update)
+                    .await
+                    .change_context(errors::ApiErrorResponse::InternalServerError)
+                    .attach_printable("Failed to udpate payment method in db")?;
+
+                return Err(report!(errors::ApiErrorResponse::InternalServerError)
+                    .attach_printable("Failed to add card to locker"));
```
Chethan-rao, line 429: aren't u loosing/ignoring the error stack report?

Sarthak1799: Could you describe a bit on how this should be refactored?

Chethan-rao: instead of using `it let` on `resp`, use match and catch the `Err(err)` and change_context from this `err` to `ApiErrorResponse`. Something like below

`Err(err) => err.change_context(ApiErrorResponse)?`

Sarthak1799: Done. Could you check?

crates/router/src/core/payment_methods/cards.rs
```diff
@@ -216,6 +225,215 @@ pub async fn get_or_insert_payment_method(
...
+        .await
+        .change_context(errors::ApiErrorResponse::PaymentMethodNotFound)
+        .attach_printable("Unable to find payment method")?;
+
+    let stored_client_secret = payment_method
+        .client_secret
+        .clone()
+        .get_required_value("client_secret")
+        .change_context(errors::ApiErrorResponse::PaymentMethodNotFound)
+        .attach_printable("Unable to find payment method")?;
```
Chethan-rao, line 306: Error message should be something like "client secret not found in db"?

Chethan-rao (changes requested):

Chethan-rao (approved):

ThisIsMani (dismissed):
Dashboard specific changes looks fine.

ivor-juspay (dismissed):

crates/router/src/types/api/payment_methods.rs
```diff
@@ -21,15 +21,14 @@ pub(crate) trait PaymentMethodCreateExt {
...
-        if let Some(payment_method_type) = self.payment_method_type {
-            if !validate_payment_method_type_against_payment_method(
-                self.payment_method,
-                payment_method_type,
-            ) {
-                return Err(report!(errors::ApiErrorResponse::InvalidRequestData {
-                    message: "Invalid 'payment_method_type' provided".to_string()
-                })
-                .attach_printable("Invalid payment method type"));
+        if let Some(pm) = self.payment_method {
```
Narayanbhat166, line 24: you can use map here instead of nesting

crates/router/src/core/payment_methods/cards.rs
```diff
@@ -216,6 +225,208 @@ pub async fn get_or_insert_payment_method(
...
+                card: None,
+                recurring_enabled: false,
+                installment_payment_enabled: false,
+                payment_experience: None,
+                metadata: req.metadata,
+                created: None,
+                bank_transfer: None,
+                last_used_at: None,
+                client_secret: res.client_secret,
+            },
```
Narayanbhat166, line 276: we should not be referring to fields from the request in response, implement a From / TryFrom from the Response that is generated in the above step

crates/router/src/core/payment_methods/cards.rs
```diff
@@ -216,6 +225,208 @@ pub async fn get_or_insert_payment_method(
...
+#[instrument(skip_all)]
+pub async fn add_payment_method_data(
+    state: routes::AppState,
+    req: api::PaymentMethodCreate,
+    merchant_account: domain::MerchantAccount,
+    key_store: domain::MerchantKeyStore,
+    pm_id: String,
+) -> errors::RouterResponse<api::PaymentMethodResponse> {
+    let db = &*state.store;
+    let customer_id = req.customer_id.clone().get_required_value("customer_id")?;
```
Narayanbhat166, line 325: We should not be collecting `customer_id` using `client_secret`. `customer_id` should be sent in the `create_payment_method` by the merchant, use the same `customer_id` here

crates/router/src/core/payment_methods/cards.rs
```diff
@@ -216,6 +225,208 @@ pub async fn get_or_insert_payment_method(
...
+    req_client_secret: &String,
+    db: &dyn db::StorageInterface,
+    pm_id: &str,
+) -> errors::CustomResult<diesel_models::PaymentMethod, errors::ApiErrorResponse> {
+    let payment_method = db
+        .find_payment_method(pm_id)
+        .await
+        .change_context(errors::ApiErrorResponse::PaymentMethodNotFound)
+        .attach_printable("Unable to find payment method")?;
+
```
Narayanbhat166, line 292: This database call can be done outside, and pass the `client_secret` to this function

crates/router/src/core/payment_methods/cards.rs
```diff
@@ -216,6 +225,208 @@ pub async fn get_or_insert_payment_method(
...
+        .change_context(errors::ApiErrorResponse::PaymentMethodNotFound)
+        .attach_printable("client secret not found in db")?;
+
+    if req_client_secret != &stored_client_secret {
+        Err((errors::ApiErrorResponse::ClientSecretInvalid).into())
+    } else {
+        let current_timestamp = common_utils::date_time::now();
+        let session_expiry = payment_method
+            .created_at
+            .saturating_add(time::Duration::seconds(consts::DEFAULT_SESSION_EXPIRY));
```
Narayanbhat166, line 306: should we have a provision to change the default session expiry?

crates/router/src/core/payment_methods/cards.rs
```diff
@@ -216,6 +225,208 @@ pub async fn get_or_insert_payment_method(
...
+    pm_id: String,
+) -> errors::RouterResponse<api::PaymentMethodResponse> {
+    let db = &*state.store;
+    let customer_id = req.customer_id.clone().get_required_value("customer_id")?;
+
+    let pmd = req
+        .payment_method_data
+        .clone()
+        .get_required_value("payment_method_data")?;
+    let _payment_method = req.payment_method.get_required_value("payment_method")?;
```
Narayanbhat166, line 331: why are we ignoring the value here?

crates/router/src/core/payment_methods/cards.rs
```diff
@@ -216,6 +225,208 @@ pub async fn get_or_insert_payment_method(
...
+                add_card_to_locker(&state, req.clone(), &card, &customer_id, &merchant_account)
+                    .await
+                    .change_context(errors::ApiErrorResponse::InternalServerError);
+
+            match resp {
+                Ok((mut pm_resp, duplication_check)) => {
+                    if duplication_check.is_some() {
+                        let pm_update = storage::PaymentMethodUpdate::StatusUpdate {
+                            status: Some(enums::PaymentMethodStatus::Inactive),
+                        };
```
Narayanbhat166, line 351: @vspecky we can return an error in this case right? we need not make the payment_method as inactive and send some other payment method.

Narayanbhat166 (changes requested):

crates/router/src/core/payment_methods/cards.rs
Sarthak1799: It's not mandatory in the request but we are mandating it in this particular flow.

crates/router/src/core/payment_methods/cards.rs
Sarthak1799: Could be thought of for future implementation but as of now, we have decided to keep a default 15 min expiry.

crates/router/src/core/payment_methods/cards.rs
```diff
@@ -219,6 +228,214 @@ pub async fn get_or_insert_payment_method(
...
+#[instrument(skip_all)]
+pub fn authenticate_pm_client_secret_and_check_expiry(
+    req_client_secret: &String,
+    payment_method: &diesel_models::PaymentMethod,
+) -> errors::CustomResult<bool, errors::ApiErrorResponse> {
+    let stored_client_secret = payment_method
+        .client_secret
+        .clone()
+        .get_required_value("client_secret")
+        .change_context(errors::ApiErrorResponse::PaymentMethodNotFound)
```
Narayanbhat166, line 280: why are we changing the error to `PaymentMethodNotFound`?

Sarthak1799: What should we change it to?

crates/router/src/core/payment_methods/cards.rs
```diff
@@ -219,6 +228,214 @@ pub async fn get_or_insert_payment_method(
...
+    let customer_id = payment_method.customer_id.clone();
+
+    let client_secret_expired =
+        authenticate_pm_client_secret_and_check_expiry(&client_secret, &payment_method)?;
+
+    if client_secret_expired {
+        let pm_update = storage::PaymentMethodUpdate::StatusUpdate {
+            status: Some(enums::PaymentMethodStatus::Inactive),
+        };
+
```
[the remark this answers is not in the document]
Sarthak1799: We need to deactivate the payment method in case it can't be activated. We also need to add a process tracker task for it, will be adding those in the upcoming of PRs for the issue.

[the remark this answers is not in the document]
Sarthak1799: Since the status can be only updated via `/payment_methods/:pm_id/save` using the `client_secret`, the status should be marked as `inactive` as no action was taken for the particular record. It doesn't seem right to have the same record status be kept as `awaiting_data` forever.

crates/router/src/core/payment_methods/cards.rs
Narayanbhat166: This should be a `missing_required_field` error right?

crates/router/src/core/payment_methods/cards.rs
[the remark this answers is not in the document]
Narayanbhat166: If we are sending a 4xx, it means that this is a client error ( some data is missing / invalid). We should also not be making any updates to the database entities ( payment method ) in this case. So the status should remain in `awaiting_data`. If you change the status of the payment method then the status code should be 200. This is the general practice

crates/router/src/core/payment_methods/cards.rs
Sarthak1799: it was already stored `client_secret` and not something that would be passed in the req for the current API. 
But sure, will change this.

crates/router/src/core/payment_methods/cards.rs
```diff
@@ -3277,11 +3502,16 @@ pub async fn set_default_payment_method(
         .await
         .change_context(errors::ApiErrorResponse::InternalServerError)
         .attach_printable("Failed to update the default payment method id for the customer")?;
+
+    let pm = payment_method
```
Narayanbhat166, line 3506: payment method can be None in this case right?

Sarthak1799: Since payment method is a required field here, we'll throw an error here in case it is None.

Narayanbhat166: you should move this before updating the customer details

Narayanbhat166 (commented):
after saving the payment method, we should set it as default for the customer

Narayanbhat166 (dismissed):

Narayanbhat166 (approved):

Narayanbhat166 (dismissed):

vspecky (dismissed):

vspecky (approved):

ThisIsMani (approved):

PR #4190, opened 2024-03-26
build(docker): add web, control-center in docker-compose

Dhanus3133:
Added hyperswitch-web and hyperswitch-control-center in docker-compose

## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description
Two new services were added to start the SDK, demo, and control center automatically within docker-compose.

Solves #4110 

### Additional Changes

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -122,6 +122,37 @@ services:
     labels:
       logs: "promtail"
 
+  hyperswitch-web:
+    ports:
+      - "9050:9050"
+      - "9060:9060"
+      - "5252:5252"
+    build:
+      context: ./docker
+      dockerfile: web.Dockerfile
+    networks:
+      - router_net
+    environment:
+      - HYPERSWITCH_SERVER_URL=http://localhost:8080
+    depends_on:
+      hyperswitch-server:
+        condition: service_started
+
+  hyperswitch-control-center:
+    image: juspaydotin/hyperswitch-control-center
+    ports:
+      - "9000:9000"
+    networks:
+      - router_net
+    environment:
+      - apiBaseUrl=http://localhost:8080
+      - sdkBaseUrl=http://localhost:9050
+    depends_on:
+      hyperswitch-server:
+        condition: service_started
+      hyperswitch-web:
+        condition: service_started
+
   ### Clustered Redis setup
   redis-cluster:
     image: redis:7
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -132,8 +132,14 @@ services:
       dockerfile: web.Dockerfile
     networks:
       - router_net
+    env_file:
+      - ./.env
     environment:
-      - HYPERSWITCH_SERVER_URL=http://localhost:8080
+      - HYPERSWITCH_PUBLISHABLE_KEY=$HYPERSWITCH_PUBLISHABLE_KEY
+      - HYPERSWITCH_SECRET_KEY=$HYPERSWITCH_SECRET_KEY
+      - HYPERSWITCH_SERVER_URL=${HYPERSWITCH_SERVER_URL:-http://hyperswitch-server:8080}
+      - HYPERSWITCH_CLIENT_URL=${HYPERSWITCH_CLIENT_URL:-http://hyperswitch-web:9050}
+      - SELF_SERVER_URL=${SELF_SERVER_URL:-http://hyperswitch-web:5252}
     depends_on:
       hyperswitch-server:
         condition: service_started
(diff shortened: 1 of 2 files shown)


docker-compose.yml
```diff
@@ -122,6 +122,37 @@ services:
...
+      - "9050:9050"
+      - "9060:9060"
+      - "5252:5252"
+    build:
+      context: ./docker
+      dockerfile: web.Dockerfile
+    networks:
+      - router_net
+    environment:
+      - HYPERSWITCH_SERVER_URL=http://localhost:8080
```
SanchithHegde, line 136: Will this container be able to interact with the server at `localhost:8080`? We would need some more changes and/or the URL might have to be updated?

docker-compose.yml
```diff
@@ -122,6 +122,37 @@ services:
...
+
+  hyperswitch-control-center:
+    image: juspaydotin/hyperswitch-control-center
+    ports:
+      - "9000:9000"
+    networks:
+      - router_net
+    environment:
+      - apiBaseUrl=http://localhost:8080
+      - sdkBaseUrl=http://localhost:9050
```
SanchithHegde, line 149: Same here, the URL would have to be updated here.

SanchithHegde (commented):
Thanks for the PR, @Dhanus3133!

Also, how is the user expected to configure the publishable and API keys, and the remaining URLs?

docker-compose.yml
Dhanus3133: It was working fine when tested. Still Updated with the changes required.

Dhanus3133:
> Thanks for the PR, @Dhanus3133!
> 
> Also, how is the user expected to configure the publishable and API keys, and the remaining URLs?

image

Resolved it.

Dhanus3133:
I noticed that one of the PRs closely mirrors mine. Just wanted to bring this matter to your attention.

docker-compose.yml
```diff
@@ -122,6 +122,43 @@ services:
...
+      - "9050:9050"
+      - "9060:9060"
+      - "5252:5252"
+    build:
+      context: ./docker
+      dockerfile: web.Dockerfile
+    networks:
+      - router_net
+    env_file:
+      - ./.env
```
SanchithHegde, line 136: Could you provide an example of what this `.env` file would contain? Including placeholder values should be good enough.

SanchithHegde (commented):
Have you been able to test this out locally, and verified that the demo app loads correctly?

Dhanus3133:
@SanchithHegde do you know why this happens?
image

SanchithHegde:
> @SanchithHegde do you know why this happens? 

If you're running the web client and demo app within a Docker container, and use `localhost:8080` as server URL, the web client / demo app would try to access a service running at port 8080 within the container itself, not on the host machine. You'd need to understand the intricacies of Docker networking, and then go with a suitable, clean approach for solving this.

docker/web.Dockerfile
```diff
@@ -0,0 +1,18 @@
+FROM node:latest
+
+RUN npm install concurrently -g
+
+WORKDIR /app
+
+RUN git clone https://github.com/juspay/hyperswitch-web .
+
+RUN sed -i '/hot: true,/a \  host: "0.0.0.0",' webpack.dev.js
+RUN sed -i '/hot: true,/a \  host: "0.0.0.0",' Hyperswitch-React-Demo-App/webpack.dev.js
```
lsampras, line 10: might not be needed since after [this](https://github.com/juspay/hyperswitch-web/pull/294).

docker-compose.yml
```diff
@@ -122,6 +122,43 @@ services:
...
+    environment:
+      - HYPERSWITCH_PUBLISHABLE_KEY=$HYPERSWITCH_PUBLISHABLE_KEY
+      - HYPERSWITCH_SECRET_KEY=$HYPERSWITCH_SECRET_KEY
+      - HYPERSWITCH_SERVER_URL=${HYPERSWITCH_SERVER_URL:-http://hyperswitch-server:8080}
+      - HYPERSWITCH_CLIENT_URL=${HYPERSWITCH_CLIENT_URL:-http://hyperswitch-web:9050}
+      - SELF_SERVER_URL=${SELF_SERVER_URL:-http://hyperswitch-web:5252}
+    depends_on:
+      hyperswitch-server:
+        condition: service_started
```
lsampras, line 145: @Dhanus3133 can you remove this since we want the option to run this along with local server.
i.e using the local cargo run along with docker services for hyperswitch-web & hyperswitch-control-center.

docker-compose.yml
```diff
@@ -122,6 +122,43 @@ services:
...
+
+  hyperswitch-control-center:
+    image: juspaydotin/hyperswitch-control-center
+    ports:
+      - "9000:9000"
+    networks:
+      - router_net
+    environment:
+      - apiBaseUrl=http://hyperswitch-server:8080
+      - sdkBaseUrl=http://hyperswitch-web:9050
```
lsampras, line 155: The sdkBaseUrl should point to the js assets, in this case it would be `sdkBaseUrl=http://<SDK_SERVICE>/HyperLoader.js`

PR #4267, opened 2024-04-01
[enhancement] Customer kv impl

akshay-97:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

add kv support for customer

### Additional Changes

## Motivation and Context

support kv flow for customer to mitigate writes going to db

## Checklist


crates/diesel_models/src/query/customers.rs
```diff
@@ -15,6 +15,32 @@ impl CustomerNew {
...
+                        (self.customer_id, self.merchant_id),
+                    )
+                    .await
+                }
+                _ => Err(error),
+            },
+            result => result,
+        }
+    }
+
```
dracarys18, line 43: This look like a redundant function

crates/router/src/core/customers.rs
```diff
@@ -186,10 +187,11 @@ pub async fn delete_customer(
 ) -> errors::CustomerResponse<customers::CustomerDeleteResponse> {
     let db = &state.store;
 
-    db.find_customer_by_customer_id_merchant_id(
+    let _customer = db.find_customer_by_customer_id_merchant_id(
```
dracarys18, line 190: You can remove the underscore here

crates/router/src/core/payment_methods/cards.rs
```diff
@@ -84,7 +84,10 @@ pub async fn create_payment_method(
     payment_method_data: Option<Encryption>,
     key_store: &domain::MerchantKeyStore,
 ) -> errors::CustomResult<storage::PaymentMethod, errors::ApiErrorResponse> {
-    db.find_customer_by_customer_id_merchant_id(customer_id, merchant_id, key_store)
+    let merchant_account = db.find_merchant_account_by_merchant_id(merchant_id, key_store)
+        .await
+        .to_not_found_response(errors::ApiErrorResponse::MerchantAccountNotFound)?;
```
dracarys18, line 89: You can pass the merchant account from the previous function. No need of another DB call here

crates/router/src/core/payments/helpers.rs
```diff
@@ -1157,11 +1157,13 @@ pub async fn get_customer_from_details<F: Clone>(
     match customer_id {
         None => Ok(None),
         Some(c_id) => {
+            let merchant_account = db.find_merchant_account_by_merchant_id(merchant_id , merchant_key_store).await?;
```
dracarys18, line 1160: You can pass it from prev function

crates/router/src/core/payments/helpers.rs
```diff
@@ -1309,13 +1311,15 @@ pub async fn create_customer_if_not_exist<'a, F: Clone, R, Ctx>(
         .customer_id
         .or(payment_data.payment_intent.customer_id.clone());
 
+    let merchant_account = db.find_merchant_account_by_merchant_id(merchant_id, key_store).await?;
```
dracarys18, line 1314: Same here

crates/router/src/db/customers.rs
```diff
@@ -63,9 +70,298 @@ where
...
+            )
+            .await
+            .map_err(Into::into)
+            .into_report()
+        };
+
+        let updated_object = match storage_scheme {
+            MerchantStorageScheme::PostgresOnly => database_call().await,
+            MerchantStorageScheme::RedisKv => {
+                let key = format!("mid_{}_cust_{}", merchant_id, customer_id);
```
dracarys18, line 164: Should the key be merchant_id_customer_id? I dont think this serves the purpose of sharding all the data with single key. Should we think of something else

cc: @jarnura

dracarys18 (changes requested):

crates/router/src/db/customers.rs
dracarys18: Because I think KV is already there for Payment related customer data, which has the shard key as merchant_id_payment_id

crates/router/src/core/customers.rs
akshay-97: sure

crates/router/src/core/payment_methods/cards.rs
akshay-97: sure

crates/router/src/db/customers.rs
akshay-97: i was thinking mandate data can also come under mid_cust_id

dracarys18: But for address it's mid_pid because only payment related address is going to KV

akshay-97: how does that work, would address be changed or created only in payments flow

dracarys18: Yes the addresses which are going through KV flow are only inserted during payments

dracarys18 (approved):

jarnura (approved):

PR #4298, opened 2024-04-04
feat(users): Implemented cookie parsing for auth

racnan:
## Type of Change

- [ ] Bugfix
- [X] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description
This is the second PR for cookie implementation.
In this PR, cookie header is parsed to retrieve JWT that was set during dashboard entry.
This PR does not remove old auth flow where `Authorization` header was used to retrieve JWT from request. Instead, it compares  the JWT retrieved from `Authorization` header and `Cookie` header and logs the result. 

### Additional Changes

## Motivation and Context
Storing JWT in cookies instead of local storage.

## How did you test it?
No testing required, feature is not complete and hence is not enabled.

## Checklist

### DIFF
--- a/crates/router/src/core/errors/api_error_response.rs
+++ b/crates/router/src/core/errors/api_error_response.rs
@@ -262,7 +262,7 @@ pub enum ApiErrorResponse {
     PaymentMethodDeleteFailed,
     #[error(
         error_type = ErrorType::InvalidRequestError, code = "IR_26",
-        message = "Cookie "
+        message = "Invalid Cookie"
     )]
     InvalidCookie,
 }
--- a/crates/router/src/core/errors/api_error_response.rs
+++ b/crates/router/src/core/errors/api_error_response.rs
@@ -260,6 +260,11 @@ pub enum ApiErrorResponse {
     CurrencyConversionFailed,
     #[error(error_type = ErrorType::InvalidRequestError, code = "IR_25", message = "Cannot delete the default payment method")]
     PaymentMethodDeleteFailed,
+    #[error(
+        error_type = ErrorType::InvalidRequestError, code = "IR_26",
+        message = "Cookie "
+    )]
+    InvalidCookie,
 }
 
 impl PTError for ApiErrorResponse {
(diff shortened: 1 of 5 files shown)


crates/router/src/core/errors/api_error_response.rs
```diff
@@ -260,6 +260,11 @@ pub enum ApiErrorResponse {
     CurrencyConversionFailed,
     #[error(error_type = ErrorType::InvalidRequestError, code = "IR_25", message = "Cannot delete the default payment method")]
     PaymentMethodDeleteFailed,
+    #[error(
+        error_type = ErrorType::InvalidRequestError, code = "IR_26",
+        message = "Cookie "
```
ThisIsMani, line 265: Message is different here.

crates/router/src/services/authentication.rs
```diff
@@ -598,6 +599,16 @@ where
...
     let token = get_jwt_from_authorization_header(headers)?;
+    if let Some(token_from_cookies) = get_cookie_from_header(headers)
+        .ok()
+        .map(|cookies| cookies::parse_cookie(cookies).ok())
+        .flatten()
+    {
+        logger::info!(
+            "Cookie header and authorization header JWT comparison result: {}",
+            token == token_from_cookies
+        );
```
ThisIsMani, line 610: Are we going to remove this log in future?

ThisIsMani (changes requested):
Either `auth_type` function should be changed to use `parse_jwt_payload` function or `is_jwt_auth` function should also check in cookies for the jwt.

crates/router/src/services/authentication.rs
racnan: Yess. This is just for testing.

racnan:
> Either `auth_type` function should be changed to use `parse_jwt_payload` function or `is_jwt_auth` function should also check in cookies for the jwt.
> 
> https://github.com/juspay/hyperswitch/blob/63d2b6855acee1adeae2efff10f424e056af0bcb/crates/router/src/services/authentication.rs#L975

@ThisIsMani As mentioned in the description, this pr doesn't use cookies for auth. This is just parsing cookies and checking if it is working as expected. Not changing any of the old flows in this pr. If, by analysing logs, we are happy with cookie parsing then we would move from old flow to new one.

ThisIsMani (approved):

NishantJoshi00 (approved):

apoorvdixit88 (approved):

PR #4357, opened 2024-04-13
fix: added find all support for pm kv

akshay-97:
## Type of Change

- [ x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

Issue scenario faced:
1. create payment method (in /payments)
2. list payment method by customer (find all query that always went to db)
 above api returns empty list
 
 to handle cases where payment method entry present in kv but not drained to db yet, we have added a lookup in kv as well and combining data for findall from both kv and db

### Additional Changes

## Motivation and Context

list payment method by customer and merchant was eventually consistent when payment method followed kv path,
change here is to combine values from db and redis

## How did you test it?

1. create payment method and drain data to db (can skip this step if payment method entries are already present in db but not kv for the customer <> merchant)
2. create another payment method through kv
3. list payment methods should reflect pm created through kv as well
`curl --location 'http://localhost:8080/customers/<cust_id>/payment_methods' \
--header 'api-key:***'`

## Checklist

### DIFF
--- a/crates/router/src/utils/db_utils.rs
+++ b/crates/router/src/utils/db_utils.rs
@@ -37,7 +37,7 @@ where
     }
 }
 
-pub async fn find_all_redis_database<F, RFut, DFut, T>(
+pub async fn find_all_combined_kv_database<F, RFut, DFut, T>(
     redis_fut: RFut,
     database_call: F,
     limit: Option<i64>,
--- a/crates/router/src/utils/db_utils.rs
+++ b/crates/router/src/utils/db_utils.rs
@@ -36,3 +36,73 @@ where
         },
     }
 }
+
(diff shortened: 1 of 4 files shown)


crates/router/src/utils/db_utils.rs
```diff
@@ -36,3 +36,66 @@ where
         },
     }
 }
+
+pub async fn find_all_redis_database<F, RFut, DFut, T>(
```
jarnura, line 40: change the function

jarnura (changes requested):

jarnura (approved):

dracarys18 (approved):

PR #4358, opened 2024-04-15
feat(router): add retrieve poll status api

sai-harsha-vardhan:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description
add retrieve poll status api

This is a generic poll status retrieve API, which can be used in any polling flows. application will set a poll_id in backend as pending, and sends back a poll_id to client to keep polling until the status is completed. once it turns to completed, client will do the required action. This is a light weight alternative to actually polling with the payment status or similar heavy apis.

### Additional Changes

## How did you test it?
Tested Manually
1. Currently we can't test the entire flow, but to check the API we can use the below CURL to check the poll status of some random poll_id, which is expected to give status as 404 Not Found (because it's not a valid one)
```
curl --location --request GET '{{BASE_URL}}/poll/status/poll_id' \
--header 'api-key: {{API_KEY}}'
```
image

Expected 200 response
```
{
    "poll_id": "abc",
    "status": "pending"
}
```

## Checklist

### DIFF
--- a/crates/router/src/core/poll.rs
+++ b/crates/router/src/core/poll.rs
@@ -20,7 +20,9 @@ pub async fn retrieve_poll_status(
         .get_redis_conn()
         .change_context(errors::ApiErrorResponse::InternalServerError)
         .attach_printable("Failed to get redis connection")?;
-    let poll_id = req.poll_id;
+    let request_poll_id = req.poll_id;
+    // prepend 'poll_' to restrict access to only fetching Poll IDs, as this is a freely passed string in the request
+    let poll_id = format!("poll_{}", request_poll_id);
     let redis_value = redis_conn
         .get_key::<Option<String>>(poll_id.as_str())
         .await
@@ -40,6 +42,9 @@ pub async fn retrieve_poll_status(
                 .unwrap_or(PollStatus::NotFound)
         })
         .unwrap_or(PollStatus::NotFound);
-    let poll_response = PollResponse { poll_id, status };
+    let poll_response = PollResponse {
+        poll_id: request_poll_id,
+        status,
+    };
     Ok(ApplicationResponse::Json(poll_response))
 }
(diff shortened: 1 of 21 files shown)


crates/router/src/core/poll.rs
```diff
@@ -0,0 +1,50 @@
...
+                poll_id.clone()
+            )
+        })?;
+    let status: PollStatus = redis_value
+        .map(|value| {
+            value
+                .parse_enum("PollStatus")
+                .map_err(|err| logger::warn!("error while parsing PollStatus: {err}"))
+                .ok()
+                .unwrap_or(PollStatus::NotFound)
```
Narayanbhat166, line 42: this should be `InternalServerError`

crates/router/src/core/poll.rs
```diff
@@ -0,0 +1,50 @@
...
+        })
+        .unwrap_or(PollStatus::NotFound);
```
Narayanbhat166, line 44: we should return a 4xx in this case because we have poll id in the request path, so if the resource is not found then a 404 should be returned

crates/router/src/core/poll.rs
```diff
@@ -0,0 +1,50 @@
...
+    req: crate::types::api::PollId,
+) -> RouterResponse<PollResponse> {
+    let redis_conn = state
+        .store
+        .get_redis_conn()
+        .change_context(errors::ApiErrorResponse::InternalServerError)
+        .attach_printable("Failed to get redis connection")?;
+    let request_poll_id = req.poll_id;
+    // prepend 'poll_' to restrict access to only fetching Poll IDs, as this is a freely passed string in the request
+    let poll_id = format!("poll_{}", request_poll_id);
```
Narayanbhat166, line 25: how are we restricting access control for poll ids? A merchant with a publishable key can access poll status of other merchants?

ivor-juspay (approved):

PR #4425, opened 2024-04-22
refactor(payment_methods): store card_network in locker

Chethan-rao:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [x] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

We have `card_brand` field in locker which is `None` currently for all cards. When a `card_network` is sent in the request, we map it to `card_brand` and store it in locker.

### Additional Changes

## Checklist

### DIFF
--- a/crates/router/src/core/payments/helpers.rs
+++ b/crates/router/src/core/payments/helpers.rs
@@ -1,4 +1,4 @@
-use std::borrow::Cow;
+use std::{borrow::Cow, str::FromStr};
 
 use api_models::{
     mandates::RecurringDetails,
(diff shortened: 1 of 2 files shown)


crates/router/src/core/payments/helpers.rs
```diff
@@ -1780,9 +1780,17 @@ pub async fn retrieve_card_with_permanent_token(
...
-        card_network: None,
+        card_network: card
+            .card_brand
+            .map(|card_brand| enums::CardNetwork::from_str(&card_brand))
+            .transpose()
+            .map_err(|e| {
+                logger::error!("Failed to parse card network {}", e);
+            })
+            .ok()
+            .flatten(),
```
NishantJoshi00, line 1793: Should we do this? I feel ignoring errors isn't a very good approach for solving this

NishantJoshi00 (changes requested):

crates/router/src/core/payments/helpers.rs
Chethan-rao: This error itself won't occur in the first place as we would have already evaluated while storing it

vspecky (approved):

NishantJoshi00 (approved):

PR #4444, opened 2024-04-23
feat: add an api for toggling extended card info feature

Chethan-rao:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

This PR adds an api under `business_profile` domain, to toggle the `extended_card_info` feature with `admin_api_key` as auth.
This PR also includes refactoring of `business_profile` updation approach.

### Additional Changes

## How did you test it?

1. Create a business profile for a merchant (Default one which gets created during merchant account creation works too)
2. Toggle extended card info feature

```
curl --location 'http://localhost:8080/account/:merchant_id/business_profile/:profile_id/toggle_extended_card_info' \
--data '{
    "enabled": true
}'
```
image

3. For verifying, u could check the db if the feature is toggled using below query
```
SELECT is_extended_card_info_enabled FROM business_profile WHERE profile_id = 'profile_id';
```

## Checklist

### DIFF
--- a/crates/diesel_models/src/business_profile.rs
+++ b/crates/diesel_models/src/business_profile.rs
@@ -35,6 +35,7 @@ pub struct BusinessProfile {
     pub payment_link_config: Option<serde_json::Value>,
     pub session_expiry: Option<i64>,
     pub authentication_connector_details: Option<serde_json::Value>,
+    pub is_extended_card_info_enabled: Option<bool>,
 }
 
 #[derive(Clone, Debug, Insertable, router_derive::DebugAsDisplay)]
@@ -61,6 +62,7 @@ pub struct BusinessProfileNew {
     pub payment_link_config: Option<serde_json::Value>,
     pub session_expiry: Option<i64>,
     pub authentication_connector_details: Option<serde_json::Value>,
+    pub is_extended_card_info_enabled: Option<bool>,
 }
 
 #[derive(Clone, Debug, Default, AsChangeset, router_derive::DebugAsDisplay)]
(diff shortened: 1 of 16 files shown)


crates/diesel_models/src/business_profile.rs
```diff
@@ -84,6 +86,84 @@ pub struct BusinessProfileUpdateInternal {
...
+        frm_routing_algorithm: Option<serde_json::Value>,
+        payout_routing_algorithm: Option<serde_json::Value>,
+        is_recon_enabled: Option<bool>,
+        applepay_verified_domains: Option<Vec<String>>,
+        payment_link_config: Option<serde_json::Value>,
+        session_expiry: Option<i64>,
+        authentication_connector_details: Option<serde_json::Value>,
+    },
+    ExtendedCardInfoUpdate {
+        is_extended_card_info_enabled: Option<bool>,
```
vspecky, line 114: Might as well make it mandatory, since it would be a no-op if the parameter wasn't given anyways. Not a blocker tho.
```suggestion
        is_extended_card_info_enabled: bool,
```

vspecky (approved):

Narayanbhat166:
why do we need an api endpoint for this? we already have an endpoint to update business profile, we can add a field there to change this config value

Chethan-rao:
> why do we need an api endpoint for this? we already have an endpoint to update business profile, we can add a field there to change this config value

Because we only want the admin to toggle this feature (`admin_api_key` as auth)

migrations/2024-04-23-132120_add-extended-card-info-to-business-profile/up.sql
```diff
@@ -0,0 +1,3 @@
+-- Your SQL goes here
+
+ALTER TABLE business_profile ADD COLUMN IF NOT EXISTS is_extended_card_info_enabled BOOLEAN DEFAULT FALSE;
```
Narayanbhat166, line 3:
```suggestion
ALTER TABLE business_profile ADD COLUMN IF NOT EXISTS store_extended_card_bin BOOLEAN DEFAULT FALSE;
```

Chethan-rao: This PR is not about card bin. That is being picked up by @Aprabhat19 . This is a different usecase

Narayanbhat166 (approved):

NishantJoshi00 (approved):
Looks good, but the name extended card info, does not convey what this feature actually does, can we consider using something else for that?

PR #4474, opened 2024-04-29
fix(connector): send valid sdk information in authentication flow netcetera

hrithikesh026:
## Type of Change

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

Two required params in SDK information were not serde renamed properly. This was causing SDK based authentication call to fail. Fixed it.

Other changes:
* Changed the type of `error_message` column in `authentication` table to `TEXT` from `VARCHAR(64)`. Because when length of error_message was greater than 64 chars, update operation was failing.
* Pass Device render options in authentication request because it is also required for SDK based authentication.
* Pass address-state as two character state code.
* pass country code in phone number after trimming `+` prefix

### Additional Changes

## How did you test it?

Manual

## Checklist

### DIFF
--- a/crates/router/src/connector/netcetera/netcetera_types.rs
+++ b/crates/router/src/connector/netcetera/netcetera_types.rs
@@ -1615,7 +1615,7 @@ pub enum ThreeDSReqAuthMethod {
 pub struct DeviceRenderingOptionsSupported {
     pub sdk_interface: SdkInterface,
     /// For Native UI SDK Interface accepted values are 01-04 and for HTML UI accepted values are 01-05.
-    pub sdk_ui_type: SdkUiType,
+    pub sdk_ui_type: Vec<SdkUiType>,
 }
 
 #[derive(Serialize, Deserialize, Debug, Clone)]
--- a/crates/router/src/connector/netcetera/netcetera_types.rs
+++ b/crates/router/src/connector/netcetera/netcetera_types.rs
@@ -3,7 +3,11 @@ use std::collections::HashMap;
 use common_utils::pii::Email;
 use serde::{Deserialize, Serialize};
 
-use crate::{connector::utils::AddressDetailsData, errors, types::api::MessageCategory};
+use crate::{
+    connector::utils::{AddressDetailsData, PhoneDetailsData},
+    errors,
+    types::api::MessageCategory,
+};
 
 #[derive(Debug, Deserialize, Serialize, Clone)]
 #[serde(untagged)]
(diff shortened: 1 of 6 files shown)


crates/router/src/connector/utils.rs
```diff
@@ -1228,6 +1233,7 @@ pub trait AddressDetailsData {
     fn get_country(&self) -> Result<&api_models::enums::CountryAlpha2, Error>;
     fn get_combined_address_line(&self) -> Result<Secret<String>, Error>;
     fn to_state_code(&self) -> Result<Secret<String>, Error>;
+    fn to_state_code_option(&self) -> Result<Option<Secret<String>>, Error>;
```
SamraatBansal, line 1236:
```suggestion
    fn to_state_code_as_optional(&self) -> Result<Option<Secret<String>>, Error>;
```

SamraatBansal (changes requested):

crates/router/src/connector/utils.rs
```diff
@@ -1311,6 +1317,12 @@ impl AddressDetailsData for api::AddressDetails {
             _ => Ok(state.clone()),
         }
     }
+    fn to_state_code_option(&self) -> Result<Option<Secret<String>>, Error> {
+        self.state
+            .as_ref()
+            .map(|_| self.to_state_code())
+            .transpose()
+    }
```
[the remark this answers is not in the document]
hrithikesh026: Unfortunately we cannot return None if Err occurs. to_state_code() will return Err when `state` is an invalid string. We cannot ignore this Err.

crates/router/src/connector/netcetera/netcetera_types.rs
```diff
@@ -716,11 +721,24 @@ impl
             bill_addr_state: billing_address
                 .address
                 .as_ref()
-                .and_then(|add| add.state.clone()),
+                .and_then(|add| add.to_state_code_as_optional().transpose())
```
sai-harsha-vardhan, line 724: We need not call `to_state_code_as_optional` here if it's already sent as state code, we should check and call this fn only if required (as `state` in our request is a String and any value can be sent there)

crates/router/src/connector/netcetera/transformers.rs
```diff
@@ -552,6 +554,22 @@ impl TryFrom<&NetceteraRouterData<&types::authentication::ConnectorAuthenticatio
         };
         let browser_information = request.browser_details.map(netcetera_types::Browser::from);
         let sdk_information = request.sdk_information.map(netcetera_types::Sdk::from);
+        let device_render_options = match request.device_channel {
+            api_models::payments::DeviceChannel::App => {
+                Some(netcetera_types::DeviceRenderingOptionsSupported {
+                    // hard-coded until core provides these values.
+                    sdk_interface: netcetera_types::SdkInterface::Both,
```
sai-harsha-vardhan, line 561: This should always be `Native` for us

hrithikesh026: As per netcetera we are supposed to send this value.

SamraatBansal (approved):

PR #4490, opened 2024-04-29
chore: make client certificate and private key secret across codebase

hrithikesh026:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

Made client Certificates and private keys secret across codebase. Because They are merchant secrets and should not be logged.

### Additional Changes

## Motivation and Context

Certificate and private key was getting logged every time a request is made. This would leak the certificate and private key.

## How did you test it?

Manual
1. Integrate a connector that uses Client certificate and private key for api authentication and checked the logs. 
Netcetera can be used.

## Checklist

### DIFF
--- a/crates/masking/src/secret.rs
+++ b/crates/masking/src/secret.rs
@@ -160,6 +160,6 @@ where
     T: ?Sized,
 {
     fn as_ref(&self) -> &T {
-        self.inner_secret.as_ref()
+        self.peek().as_ref()
     }
 }
--- a/crates/masking/src/secret.rs
+++ b/crates/masking/src/secret.rs
@@ -152,3 +152,14 @@ where
         SecretValue::default().into()
     }
 }
+
+impl<SecretValue, MaskingStrategy, T> AsRef<T> for Secret<SecretValue, MaskingStrategy>
+where
+    SecretValue: AsRef<T>,
+    MaskingStrategy: Strategy<SecretValue>,
+    T: ?Sized,
+{
+    fn as_ref(&self) -> &T {
+        self.inner_secret.as_ref()
+    }
+}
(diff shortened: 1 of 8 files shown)


crates/masking/src/secret.rs
```diff
@@ -152,3 +152,14 @@ where
...
+impl<SecretValue, MaskingStrategy, T> AsRef<T> for Secret<SecretValue, MaskingStrategy>
+where
+    SecretValue: AsRef<T>,
+    MaskingStrategy: Strategy<SecretValue>,
+    T: ?Sized,
+{
+    fn as_ref(&self) -> &T {
+        self.inner_secret.as_ref()
+    }
+}
```
SanchithHegde, line 165: Please avoid adding this implementation: we'd prefer knowing where the inner value of the secret is being accessed by `peek()` / `expose()` as compared to implicit conversions such as `Deref` or `AsRef`.

crates/router/src/core/verification.rs
```diff
@@ -42,8 +42,8 @@ pub async fn verify_merchant_creds_for_applepay(
             "application/json".to_string().into(),
         )])
         .set_body(RequestContent::Json(Box::new(request_body)))
-        .add_certificate(Some(cert_data))
-        .add_certificate_key(Some(key_data))
+        .add_certificate(Some(pii::Secret::new(cert_data)))
+        .add_certificate_key(Some(pii::Secret::new(key_data)))
```
SanchithHegde, line 46: We have already imported `ExposeInterface` from the `masking` crate here, might as well import `Secret` from the `masking` crate directly. (Optional change by the way.)

SanchithHegde (approved):

lsampras (approved):

PR #4529, opened 2024-05-03
refactor remove default case handling fiserv

### LINKED ISSUE #4059: [REFACTOR]: [Fiserv] Remove Default Case Handling

- In order to manage payment methods that are not implemented by Hyperswitch or yet to be implemented, the connector transformers file should make use of the NotImplemented ConnectorError enum variant.
- By doing so, we will throw same error message for all the Connector Implementation

:bookmark: Note: All the changes needed should be contained within `hyperswitch/crates/router/src/connector/fiserv/transformer.rs`

ind1goDusk:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [x] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

### Additional Changes

this pull request removes default case handling in `crates/router/src/connector/fiserv/transformers.rs` and handles each case explicitly instead. 

## Motivation and Context

closes #4059

## Checklist

### DIFF
--- a/crates/router/src/connector/fiserv/transformers.rs
+++ b/crates/router/src/connector/fiserv/transformers.rs
@@ -392,7 +392,7 @@ impl<F, T> TryFrom<types::ResponseRouterData<F, FiservSyncResponse, T, types::Pa
     ) -> Result<Self, Self::Error> {
         let gateway_resp = match item.response.sync_responses.first() {
             Some(gateway_response) => gateway_response,
-            _ => Err(errors::ConnectorError::ResponseHandlingFailed)?,
+            None => Err(errors::ConnectorError::ResponseHandlingFailed)?,
         };
 
         Ok(Self {


hyperswitch
SanchithHegde, line 1: Hey @ind1goDusk, could you push the changes made to the repository, instead of adding a submodule?

SanchithHegde (commented):
Also, we would prefer you commenting on the linked issue, so that we can assign it to you.

ind1goDusk:
if this issue is still available, i'd like to be assidned to it. i'm closing this pull request to start from a fresh branch.

SanchithHegde:
> if this issue is still available, i'd like to be assidned to it. i'm closing this pull request to start from a fresh branch.

Yes, @ind1goDusk, the issue is still available. Can you comment on the linked issue so that we can assign it to you?

Issue #4595, opened 2024-05-08
[Feature] Add tenant-ID as a field to all KafkaStore events


lsampras:
### Context
In-order to support multi-tenancy we need to store data partioned via tenant-id.
In order to do that for kafka events we need to pass tenant-id as a field in the generated kafka events

We can use the KafkaStore field added in #4512 to add them to events in Kafka

### Approach
1. Add a new field in `KafkaEvent` struct representing tenant_id

```rust
#[derive(serde::Serialize, Debug)]
struct KafkaEvent<'a, T: KafkaMessage> {
    #[serde(flatten)]
    event: &'a T,
    sign_flag: i32,
    tenant_id: TenantID
}
```

Alternate approaches can be discussed in the issue

subhajit20:
hey @lsampras I am taking this issue.

PR #4598, opened 2024-05-09
refactor(db): Add TenantID field to KafkaEvent struct

subhajit20:
## Type of Change

#4595 

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

- Add TenantID to the KafkaEvent struct and add the extra param to the KafkaEvent constructor
- Made changes to the required placed 

### Additional Changes

## Motivation and Context

- Supporting multi-tenancy

## How did you test it?

image

## Checklist

### DIFF
--- a/crates/router/src/db/kafka_store.rs
+++ b/crates/router/src/db/kafka_store.rs
@@ -1109,7 +1109,7 @@ impl PaymentAttemptInterface for KafkaStore {
 
         if let Err(er) = self
             .kafka_producer
-            .log_payment_attempt(&attempt, None)
+            .log_payment_attempt(&attempt, None, TenantID("default".to_string()))
             .await
         {
             logger::error!(message="Failed to log analytics event for payment attempt {attempt:?}", error_message=?er)
--- a/crates/router/src/db/kafka_store.rs
+++ b/crates/router/src/db/kafka_store.rs
@@ -1,3 +1,5 @@
+use std::sync::Arc;
+
 use common_enums::enums::MerchantStorageScheme;
 use common_utils::{errors::CustomResult, pii};
 use diesel_models::{
@@ -24,7 +26,6 @@ use scheduler::{
     SchedulerInterface,
 };
 use serde::Serialize;
-use std::sync::Arc;
 use storage_impl::redis::kv_store::RedisConnInterface;
 use time::PrimitiveDateTime;
 
--- a/crates/router/src/db/kafka_store.rs
+++ b/crates/router/src/db/kafka_store.rs
@@ -76,7 +76,7 @@ use crate::{
     },
 };
 
-#[derive(Clone, Serialize)]
+#[derive(Debug, Clone, Serialize)]
 pub struct TenantID(pub String);
 
 #[derive(Clone)]
--- a/crates/router/src/db/kafka_store.rs
+++ b/crates/router/src/db/kafka_store.rs
@@ -1,5 +1,3 @@
-use std::sync::Arc;
-
 use common_enums::enums::MerchantStorageScheme;
 use common_utils::{errors::CustomResult, pii};
 use diesel_models::{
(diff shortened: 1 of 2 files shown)


crates/router/src/services/kafka.rs
```diff
@@ -268,26 +272,29 @@ impl KafkaProducer {
         old_attempt: Option<PaymentAttempt>,
     ) -> MQResult<()> {
         if let Some(negative_event) = old_attempt {
-            self.log_event(&KafkaEvent::old(&KafkaPaymentAttempt::from_storage(
-                &negative_event,
-            )))
+            self.log_event(&KafkaEvent::old(
+                &KafkaPaymentAttempt::from_storage(&negative_event),
+                TenantID("default".to_string()),
```
lsampras, line 277: can you use the tenant_id stored in the kafkastore instead?

lsampras (commented):
Also can you share screenshots of the generated events from the kafka-ui ?

crates/router/src/services/kafka.rs
subhajit20: Yes I have used that TenantID here only ... I have imported that at the top.

crates/router/src/services/kafka.rs
```diff
@@ -30,6 +30,7 @@ use crate::types::storage::Dispute;
 
 // Using message queue result here to avoid confusion with Kafka result provided by library
 pub type MQResult<T> = CustomResult<T, KafkaError>;
+use crate::db::kafka_store::TenantID;
```
subhajit20, line 33: Here is the line @lsampras

crates/router/src/services/kafka.rs
lsampras: I mean that can you add a parameter in the log_payment_attempt function to accept a tenant_id and use that instead.

subhajit20: Yes done that!

crates/router/src/db/kafka_store.rs
```diff
@@ -1109,7 +1109,7 @@ impl PaymentAttemptInterface for KafkaStore {
 
         if let Err(er) = self
             .kafka_producer
-            .log_payment_attempt(&attempt, None)
+            .log_payment_attempt(&attempt, None, TenantID("default".to_string()))
```
lsampras, line 1112:
```suggestion
            .log_payment_attempt(&attempt, None, self.tenant_id)
```
Can we use the tenant_id stored in `KafkaStore` struct as such

crates/router/src/services/kafka.rs
```diff
@@ -266,28 +270,32 @@ impl KafkaProducer {
...
         &self,
         delete_old_attempt: &PaymentAttempt,
     ) -> MQResult<()> {
-        self.log_event(&KafkaEvent::old(&KafkaPaymentAttempt::from_storage(
-            delete_old_attempt,
-        )))
+        self.log_event(&KafkaEvent::old(
+            &KafkaPaymentAttempt::from_storage(delete_old_attempt),
+            TenantID("default".to_string()),
+        ))
```
lsampras, line 298: Can we follow the same pattern as we did for `log_payment_attempt` function for every other usage as well?

there shouldn't be any new initializations of `TenantID` in this PR, since it is already defined in KafkaStore which is what we'll be using instead

lsampras (commented):
Some minor refactors needed,

Also can you run the formatting & clippy checks locally there might be some linting problems here.

Would also prefer if you are able to run it locally and share screenshots for the same

crates/router/src/services/kafka.rs
subhajit20: Yes, I this commit solves this issue,
let me know if further changes needed. @lsampras

lsampras (commented):
@subhajit20 the code changes look good, Will you be able to test this out locally & share the screenshots of updated Kafka events?

lsampras:
The kafka events seem to have tenant-ID present

image

lsampras (approved):

lsampras:
Thanks for the contribution @subhajit20.

PR #4626, opened 2024-05-13
Refac/cg inclusion pm

prajjwalkumar17:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

### Additional Changes

## Checklist


crates/router/src/core/payment_methods/cards.rs
```diff
@@ -1876,18 +1877,38 @@ pub async fn list_payment_methods(
...
+            mca.connector_name.clone(),
+            pm_config_mapping,
+            &state.conf.mandates.supported_payment_methods,
+            &state.conf.mandates.update_mandate_supported,
+        ) {
+            logger::error!("Failed to construct constraint graph for list payment methods {e:?}");
+        }
+    }
+
+    let graph = builder.build();
```
vspecky, line 1903: We need to cache the constraint graph. This will require creating a Moka cache for the constraint graph as well as implementing the cache invalidation logic via redis pub-sub.

@Sarthak1799 has implemented the same for routing so you can take his help.

vspecky (changes requested):

crates/router/src/core/payment_methods/cards.rs
```diff
@@ -2757,16 +2874,17 @@ pub async fn call_surcharge_decision_management_for_saved_card(
...
     payment_intent: Option<&storage::PaymentIntent>,
     payment_attempt: Option<&storage::PaymentAttempt>,
     address: Option<&domain::Address>,
     connector: String,
-    config: &settings::ConnectorFilters,
-    supported_payment_methods_for_mandate: &settings::SupportedPaymentMethodsForMandate,
-    supported_payment_methods_for_update_mandate: &settings::SupportedPaymentMethodsForMandate,
+    _config: &settings::ConnectorFilters,
+    _supported_payment_methods_for_mandate: &settings::SupportedPaymentMethodsForMandate,
+    _supported_payment_methods_for_update_mandate: &settings::SupportedPaymentMethodsForMandate,
```
prajjwalkumar17, line 2887: Can be removed

crates/router/src/core/payment_methods/cards.rs
```diff
@@ -1877,51 +1881,139 @@ pub async fn list_payment_methods(
...
+                    billing_address.as_ref(),
+                    mca.connector_name.clone(),
+                    pm_config_mapping,
+                    &state.conf.mandates.supported_payment_methods,
+                    &state.conf.mandates.update_mandate_supported,
+                    &state.conf.saved_payment_methods,
+                )
+                .await?;
+            }
+            if let Err(err) = refresh_pm_filters_cache(&key, graph).await {
```
Sarthak1799, line 1972: How would you get an error here? @prajjwalkumar17

prajjwalkumar17: Removed the logger

crates/router/src/core/payment_methods/cards.rs
```diff
@@ -2903,24 +3021,24 @@ pub async fn filter_payment_methods(
...
-                        && filter3
-                        && filter4
-                        && filter5
-                        && filter6
-                        && filter7
-                        && filter8
-                        && filter9
+                    if filter_pm_based_on_allowed_types
+                        && filter_incorrect_client_secret
+                        && result.is_ok()
```
vspecky, line 3035: We need to check if the result is `true`. Also, log the error if we get one.
```suggestion
                        && matches!(result, Ok(true))
```

crates/router/src/core/payment_methods/utils.rs
```diff
@@ -0,0 +1,857 @@
...
+
+            // Capture Method filter
+            config
+                .0
+                .get(connector.as_str())
+                .or_else(|| config.0.get("default"))
+                .and_then(|inner| match pmt.payment_method_type {
+                    api_enums::PaymentMethodType::Credit | api_enums::PaymentMethodType::Debit => {
+                        if let Ok(Some(capture_method_filter)) =
+                            construct_capture_method_node(builder, inner, &pmt.payment_method_type)
```
vspecky, line 188: Why are we adding capture method filters for cards only?

crates/router/src/core/payment_methods/utils.rs
```diff
@@ -0,0 +1,857 @@
...
+
+    Ok(Some(
+        builder
+            .make_any_aggregator(
+                &agg_nodes,
+                Some("any node for card and non card pm"),
+                None::<()>,
+                None,
+            )
+            .map_err(KgraphError::GraphConstructionError)?,
```
vspecky, line 484: Do we need to construct an `any` aggregator here? Cuz from what I can see in this function, we will construct either a card specific subgraph, or a non-card-specific subgraph, but never both at once. So the `any` aggregator here will probably always have only 1 predecessor, making it redundant. Please check and correct me if I'm wrong.

crates/router/src/core/payment_methods/utils.rs
```diff
@@ -0,0 +1,857 @@
...
+                                .map(|country| {
+                                    dir::DirValue::BillingCountry(
+                                        common_enums::Country::from_alpha2(country),
+                                    )
+                                })
+                                .collect(),
+                            None,
+                            None::<()>,
+                        )
+                        .map_err(KgraphError::GraphConstructionError)?;
```
vspecky, line 598: This node should be created regardless of whether the `connector.payment_method_type` config exists or not right? So this should probably happen outside the `if let`

crates/router/src/core/payment_methods/utils.rs
```diff
@@ -0,0 +1,857 @@
...
```
vspecky, line 667: same here. This should probably be created regardless of whether the config exists or not.

crates/router/src/core/payment_methods/utils.rs
```diff
@@ -0,0 +1,857 @@
...
+                    let pm_object_currency_value_node = builder
+                        .make_in_aggregator(
+                            currency
+                                .into_iter()
+                                .map(dir::DirValue::PaymentCurrency)
+                                .collect(),
+                            None,
+                            None::<()>,
+                        )
+                        .map_err(KgraphError::GraphConstructionError)?;
```
vspecky, line 745: Same here. Probably should be constructed regardless of whether the config exists or not.

crates/router/src/core/payment_methods/utils.rs
```diff
@@ -0,0 +1,857 @@
...
```
vspecky, line 811: Same here. Probably should be constructed regardless of whether the config exists or not.

vspecky (changes requested):

crates/router/src/core/payment_methods/cards.rs
```diff
@@ -2861,35 +2915,99 @@ pub async fn filter_payment_methods(
...
+                        .and_then(|inner| inner.capture_method)
+                        .and_then(|capture_method| {
+                            (capture_method == common_enums::CaptureMethod::Manual).then(|| {
+                                context_values.push(dir::DirValue::CaptureMethod(
+                                    common_enums::CaptureMethod::Manual,
+                                ));
+                            })
+                        });
+
+                    let filter_incorrect_client_secret = req
```
Sarthak1799, line 3010: Can we rename it to something like `saved_payment_methods_filter` or something similar?

crates/router/src/core/payment_methods/cards.rs
```diff
@@ -2861,35 +2915,99 @@ pub async fn filter_payment_methods(
...
+                                        sub_label: None,
+                                    },
+                                )));
+                            };
+                        });
+
+                    payment_attempt
                         .and_then(|attempt| attempt.mandate_data.as_ref())
                         .map(|mandate_detail| {
                             if mandate_detail.update_mandate_id.is_some() {
```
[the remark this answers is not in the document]
prajjwalkumar17: can't use an if statement here, as we are using the `mandate_detail` in the below code block.

crates/router/src/core/payment_methods/cards.rs
```diff
@@ -2861,35 +2915,99 @@ pub async fn filter_payment_methods(
...
-                    let filter9 = req
+                    payment_attempt
+                        .and_then(|inner| inner.capture_method)
+                        .and_then(|capture_method| {
+                            (capture_method == common_enums::CaptureMethod::Manual).then(|| {
+                                context_values.push(dir::DirValue::CaptureMethod(
+                                    common_enums::CaptureMethod::Manual,
+                                ));
+                            })
+                        });
```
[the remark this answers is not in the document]
prajjwalkumar17: We are pushing `CaptureMethod` only if it is `Manual`, because we have `not_applicable_flows = manual` as the only variant in config. There is no place where we can find `not_applicable_flows = automatic`.

Moreover will using `then` in above code will produce some latency or something. Just curious to know why it should be converted to an `if`.

crates/router/src/core/payment_methods/cards.rs
prajjwalkumar17: `key_value_analysis` returns `Result<(), E>`, So matching on `Ok(true)` won't be possible we can anyways match on `Ok(())`.

crates/router/src/core/payment_methods/cards.rs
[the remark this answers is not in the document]
prajjwalkumar17: Moreover it was done in the same way previously :)
Screenshot 2024-05-23 at 17 52 34

crates/router/src/core/payment_methods/utils.rs
prajjwalkumar17: Added it for all PMT

crates/router/src/core/payment_methods/utils.rs
prajjwalkumar17: We can probably discuss on this and can refactor this in upcoming PR. If required can linkup an issue here as well. Anyways we have `update_mandate` flow supported by one connector(cybersource) only. So seeing the timelines if it is not necessary we can ignore it for this PR. We can discuss and change accordingly in the next PR.

crates/router/src/core/payment_methods/cards.rs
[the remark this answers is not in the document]
vspecky: > Moreover will using then in above code will produce some latency or something. Just curious to know why it should be converted to an if.

Using functional style code like this is more appropriate when your code is supposed to return a value (that's what the function signatures of `and_then()` and `then()` denote. Here we're not returning a value, we're performing an action with side effects, which is non-functional i.e. imperative, hence an `if` would be more suitable.

vspecky (approved):

Sarthak1799 (approved):

PR #4708, opened 2024-05-20
feat(users): Create generate recovery codes API

### LINKED ISSUE #4707: refactor: Separate out recovery codes from TOTP
Even though recovery codes is dependent entity on TOTP. But the new findings suggest that TOTP is a part of entity called 2FA and TOTP is depended on the 2FA entity not the TOTP entity.

This will be more clear when we add multiple 2FAs like SMS OTP.

So, we want to separate out recovery codes from the TOTP. This means `begin_totp` should not return the recovery codes anymore.

ThisIsMani:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

Currently recovery code are being generated in `begin_totp` API. We want to make a separate out recovery codes and TOTP, so this PR creates a new API which generates the recovery codes.

This API will also be used to regenerate recovery codes later.

### Additional Changes

## Motivation and Context

This is a sub part of #4707

## How did you test it?

**This can only be tested locally as it requires some changes in the redis.**
```
curl --location 'http://localhost:8080/user/recovery_codes/generate' \
--header 'Authorization: Bearer SPT with purpose as TOTP' \
```
```
{
    "recovery_codes": [
        "w5fM-NLm0",
        "TumH-oyXE",
        "wDIr-zEmu",
        "HZjm-e16M",
        "Q4qB-MupL",
        "5BtN-vRuY",
        "4vG6-URGf",
        "Dbq8-n5V1"
    ]
}
```

## Checklist


crates/router/src/core/errors/user.rs
```diff
@@ -68,8 +68,10 @@ pub enum UserErrors {
     RoleNameAlreadyExists,
     #[error("TOTPNotSetup")]
```
apoorvdixit88, line 69: This also should be `Totp` considering below two

crates/router/src/core/user.rs
```diff
@@ -1631,7 +1631,7 @@ pub async fn begin_totp(
         }));
     }
 
-    let totp = utils::user::generate_default_totp(user_from_db.get_email(), None)?;
+    let totp = utils::user::two_fa::generate_default_totp(user_from_db.get_email(), None)?;
```
apoorvdixit88, line 1634: It would be better if we call this function `two_factor_authentication`?

ThisIsMani: It is a module and I feel it will be too big for a file name.
Opinions @racnan

racnan: `two_factor_auth` ?

apoorvdixit88 (approved):

racnan (approved):

PR #4801, opened 2024-05-29
refactor(core): move router data flow types to hyperswitch domain models crate

hrithikesh026:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [x] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

moved router data flow types to hyperswitch domain models crate

### Additional Changes

## How did you test it?

Checked if the code compiles successfully.

## Checklist


Narayanbhat166 (dismissed):

crates/router/src/connector/adyen.rs
```diff
@@ -367,7 +367,7 @@ impl
     ) -> CustomResult<RequestContent, errors::ConnectorError> {
         let authorize_req = types::PaymentsAuthorizeRouterData::foreign_from((
             req,
-            types::PaymentsAuthorizeData::from(req),
+            types::PaymentsAuthorizeData::foreign_from(req),
```
SanchithHegde, line 370: These can all be `From` implementations itself, since all the types now reside in the `hyperswitch_domain_models` crate, right?

crates/hyperswitch_domain_models/src/router_request_types.rs
```diff
@@ -99,6 +112,38 @@ pub struct ConnectorCustomerData {
...
+            email: data.email,
+            payment_method_data: data.payment_method_data,
+            description: None,
+            phone: None,
+            name: None,
+            preprocessing_id: None,
+        })
+    }
+}
+impl<F> TryFrom<&RouterData<F, PaymentsAuthorizeData, response_types::PaymentsResponseData>>
```
Narayanbhat166, line 128: can we have specialized type here instead of a generic?

crates/router/src/connector/adyen.rs
hrithikesh026: No RouterData alias type definition `SetupMandateRouterData` exists in router_crate. Once that is moved to `hyperswitch_domain_models`, We can move the From trait implementation to `hyperswitch_domain_models`

crates/hyperswitch_domain_models/src/router_request_types.rs
hrithikesh026: Not currently. Because the aliases for RouterData are currently defined in router crate.

Narayanbhat166: The Generic F can be a  specialized type right, in this case it will be `api::Authorize`

crates/router/src/connector/adyen.rs
SanchithHegde: Yeah, can't we move the type aliases as well to the `hyperswitch_domain_models` crate?

hrithikesh026: We can. I would like to take that up in a new PR. Along with moving all From and TryFrom implementations to `hyperswitch_domain_models`.

SanchithHegde (approved):

Narayanbhat166 (approved):

PR #4849, opened 2024-06-03
chore(process_tracker): use `const` instead of `String` for `business_status`

pixincreate:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [x] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

When below command is executed in sandbox DB, we get a set of "distinct" `business_status`es that is specified for process tracker:

```sql
SELECT DISTINCT(business_status) FROM process_tracker;
```

For debugging and fun, execute to learn more about process tracker and its behavior:
```sql
SELECT DISTINCT(name, retry_count, business_status, status), created_at FROM process_tracker order by created_at desc LIMIT 10000;
```

The main goal for us here is to keep all the `business_status`es in a single place, confined for ease of access and increased readability. This also helps us keep track of the statuses we use.

### Additional Changes

## Motivation and Context

We need to keep a track of the statuses that we use in `process_tracker`

## How did you test it?

This should not require any testing as the change done is here just a mere replacement of hardcoded string with `const`s. If needed, the changes can be verified by following #3712's instructions. Reproduced below:

Perform PSync call from connector like `Checkout`. Payment Status should go to `processing` state. Wait for `process_tracker` to kick in and do a `force_sync` on its own. Wait for the status to get updated. I had webhooks set up (invalid configuration), even that is verified to be working.

Command used in DB: 

```sql
SELECT * FROM process_tracker ORDER BY created_at desc LIMIT 2;
```

image

image

image

image

## Checklist

### DIFF
--- a/crates/diesel_models/src/process_tracker.rs
+++ b/crates/diesel_models/src/process_tracker.rs
@@ -229,30 +229,38 @@ mod tests {
 pub mod business_status {
     /// Indicates that an irrecoverable error occurred during the workflow execution.
     pub const GLOBAL_FAILURE: &str = "GLOBAL_FAILURE";
+
     /// Task successfully completed by consumer.
     /// A task that reaches this status should not be retried (rescheduled for execution) later.
     pub const COMPLETED_BY_PT: &str = "COMPLETED_BY_PT";
+
     /// An error occurred during the workflow execution which prevents further execution and
     /// retries.
     /// A task that reaches this status should not be retried (rescheduled for execution) later.
     pub const FAILURE: &str = "FAILURE";
+
     /// The resource associated with the task was removed, due to which further retries can/should
     /// not be done.
     pub const REVOKED: &str = "Revoked";
+
     /// The task was executed for the maximum possible number of times without a successful outcome.
     /// A task that reaches this status should not be retried (rescheduled for execution) later.
     pub const RETRIES_EXCEEDED: &str = "RETRIES_EXCEEDED";
+
     /// The outgoing webhook was successfully delivered in the initial attempt.
     /// Further retries of the task are not required.
     pub const INITIAL_DELIVERY_ATTEMPT_SUCCESSFUL: &str = "INITIAL_DELIVERY_ATTEMPT_SUCCESSFUL";
+
     /// Indicates that an error occurred during the workflow execution.
     /// This status is typically set by the workflow error handler.
     /// A task that reaches this status should not be retried (rescheduled for execution) later.
     pub const GLOBAL_ERROR: &str = "GLOBAL_ERROR";
+
     /// The resource associated with the task has been significantly modified since the task was
     /// created, due to which further retries of the current task are not required.
     /// A task that reaches this status should not be retried (rescheduled for execution) later.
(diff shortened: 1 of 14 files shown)


crates/diesel_models/src/process_tracker.rs
```diff
@@ -227,3 +225,24 @@ mod tests {
...
+    /// Retry deleting tokenized data
+    pub const RETRIES_EXCEEDED: &str = "RETRIES_EXCEEDED";
+    /// Successful webhook first attempt
+    pub const INITIAL_DELIVERY_ATTEMPT_SUCCESSFUL: &str = "INITIAL_DELIVERY_ATTEMPT_SUCCESSFUL";
+    /// Consumer error handler failed to update status
+    pub const GLOBAL_ERROR: &str = "GLOBAL_ERROR";
+    /// Changed webhook resource status after event was created, hence the mismatch
+    pub const RESOURCE_STATUS_MISMATCH: &str = "RESOURCE_STATUS_MISMATCH";
+    /// Business status after a payment is created
+    pub const PENDING: &str = "Pending";
```
SanchithHegde, line 247:
```suggestion
    /// Indicates that an irrecoverable error occurred during the workflow execution.
    pub const GLOBAL_FAILURE: &str = "GLOBAL_FAILURE";

    /// Task successfully completed by consumer.
    /// A task that reaches this status should not be retried (rescheduled for execution) later.
    pub const COMPLETED_BY_PT: &str = "COMPLETED_BY_PT";

    /// An error occurred during the workflow execution which prevents further execution and
    /// retries.
    /// A task that reaches this status should not be retried (rescheduled for execution) later.
    pub const FAILURE: &str = "FAILURE";

    /// The resource associated with the task was removed, due to which further retries can/should
    /// not be done.
    pub const REVOKED: &str = "Revoked";

    /// The task was executed for the maximum possible number of times without a successful outcome.
    /// A task that reaches this status should not be retried (rescheduled for execution) later.
    pub const RETRIES_EXCEEDED: &str = "RETRIES_EXCEEDED";

    /// The outgoing webhook was successfully delivered in the initial attempt.
    /// Further retries of the task are not required.
    pub const INITIAL_DELIVERY_ATTEMPT_SUCCESSFUL: &str = "INITIAL_DELIVERY_ATTEMPT_SUCCESSFUL";

    /// Indicates that an error occurred during the workflow execution.
    /// This status is typically set by the workflow error handler.
    /// A task that reaches this status should not be retried (rescheduled for execution) later.
    pub const GLOBAL_ERROR: &str = "GLOBAL_ERROR";

    /// The resource associated with the task has been significantly modified since the task was
    /// created, due to which further retries of the current task are not required.
    /// A task that reaches this status should not be retried (rescheduled for execution) later.
    pub const RESOURCE_STATUS_MISMATCH: &str = "RESOURCE_STATUS_MISMATCH";

    /// Business status set for newly created tasks.
    pub const PENDING: &str = "Pending";
```

SanchithHegde (commented):

SanchithHegde (approved):

lsampras (approved):

PR #4923, opened 2024-06-10
refactor(storage): remove `id` from payment intent, attempt and remove datamodel ext from payment intent

Narayanbhat166:
## Type of Change

- [x] Refactoring

## Description

This PR includes few changes to the storage models that are necessary in order to support the upcoming changes.
- Remove `id` column from payment intent and payment attempt. This field is a serially increasing integer column. This field may not be required in the storage models, since this will be replaced by a global `resource_id` column.
- The `DataModelExt` trait that was implemented over payment intent has been moved to the `Behaviour` trait. This is in order to have support for encrypted columns in payment intent.

### Additional Changes

## How did you test it?

- Run postman tests locally.

## Checklist


crates/hyperswitch_domain_models/src/type_encryption.rs
```diff
@@ -0,0 +1,237 @@
...
+        crypto::Encryptable::encrypt(inner, key, crypto::GcmAes256),
+        &metrics::ENCRYPTION_TIME,
+        &metrics::CONTEXT,
+        &[],
+    )
+    .await
+}
+
+#[inline]
+pub async fn encrypt_optional<E: Clone, S>(
```
SanchithHegde, line 198: We should either name this function differently (say `encrypt_optional_field`?), or have this implemented as a trait on `Secret<E>` and `Option<Secret<E>>`, so that both have the same method name. Having the name as `encrypt_optional` has caused external folks to think that the encryption process itself happens optionally.

crates/router/src/types/domain/behaviour.rs
```diff
@@ -1,32 +1 @@
...
-    async fn convert(self, key: &Secret<Vec<u8>>) -> CustomResult<SrcType, ValidationError>;
-}
-
-#[async_trait::async_trait]
-impl<T: Send, U: Conversion<DstType = T>> ReverseConversion<U> for T {
-    async fn convert(self, key: &Secret<Vec<u8>>) -> CustomResult<U, ValidationError> {
-        U::convert_back(self, key).await
-    }
-}
+pub use hyperswitch_domain_models::behaviour::{Conversion, ReverseConversion};
```
SanchithHegde, line 1: Can we move this as an inline module within the parent module (`domain.rs`)?

```rust
mod behaviour {
    pub use hyperswitch_domain_models::behaviour::{Conversion, ReverseConversion};
}
```

crates/router/src/types/domain/merchant_key_store.rs
```diff
@@ -1,59 +1 @@
...
-
-    async fn construct_new(self) -> CustomResult<Self::NewDstType, ValidationError> {
-        Ok(diesel_models::merchant_key_store::MerchantKeyStoreNew {
-            merchant_id: self.merchant_id,
-            key: self.key.into(),
-            created_at: date_time::now(),
-        })
-    }
-}
+pub use hyperswitch_domain_models::merchant_key_store::MerchantKeyStore;
```
SanchithHegde, line 1: Same here as well?

crates/router/src/types/domain/types.rs
SanchithHegde, line 1: And here. Even if that is not taken up, please remove the commented code.

crates/hyperswitch_domain_models/src/payments/payment_intent.rs
```diff
@@ -440,6 +446,247 @@ impl From<PaymentIntentUpdate> for PaymentIntentUpdateInternal {
     }
 }
 
+use diesel_models::PaymentIntentUpdate as DieselPaymentIntentUpdate;
+
+impl From<PaymentIntentUpdate> for DieselPaymentIntentUpdate {
```
SanchithHegde, line 451: Can we remove this implementation (and `diesel_models::PaymentIntentUpdate`), and have the necessary implementations on the domain model itself?

crates/router/src/types/domain/behaviour.rs
Narayanbhat166: can you please explain why this change would be necessary?

SanchithHegde: Not a necessary change, only an improvement that we can take up. My reasoning is that we're creating individual files for modules like this which only re-export items from somewhere else and have no other code. I think these can be moved to inline modules instead.

crates/hyperswitch_domain_models/src/payments/payment_intent.rs
SanchithHegde: Yes.

crates/common_utils/Cargo.toml
```diff
@@ -8,16 +8,17 @@ readme = "README.md"
...
 logs = ["dep:router_env"]
+metrics = ["dep:router_env"]
 
 [dependencies]
 async-trait = { version = "0.1.79", optional = true }
 bytes = "1.6.0"
 diesel = "2.1.5"
 error-stack = "0.4.1"
-futures = { version = "0.3.30", optional = true }
+futures = { version = "0.3.30" }
```
SanchithHegde, line 21: Can we make this dependency optional as well, and enable it for the `metrics` feature flag?

crates/hyperswitch_domain_models/Cargo.toml
SanchithHegde, line 1: Nit: Please sort dependencies in each group (first and third party crates)

crates/hyperswitch_domain_models/src/payments/payment_intent.rs
Narayanbhat166: This will be a significant change involving the kv layer. This can be taken up in a separate issue

crates/router/src/types/domain/behaviour.rs
Narayanbhat166: Got it

crates/diesel_models/src/payment_intent.rs
```diff
@@ -233,18 +232,15 @@ pub struct PaymentIntentUpdateInternal {
     #[diesel(deserialize_as = super::OptionalDieselArray<pii::SecretSerdeValue>)]
     pub order_details: Option<Vec<pii::SecretSerdeValue>>,
     pub attempt_count: Option<i16>,
-    pub profile_id: Option<String>,
```
jarnura, line 236: why profile_id got removed?

crates/diesel_models/src/payment_intent.rs
```diff
@@ -233,18 +232,15 @@ pub struct PaymentIntentUpdateInternal {
...
+    pub merchant_decision: Option<String>,
+    pub payment_confirm_source: Option<storage_enums::PaymentSource>,
     pub updated_by: String,
     pub surcharge_applicable: Option<bool>,
     pub incremental_authorization_allowed: Option<bool>,
     pub authorization_count: Option<i32>,
     pub session_expiry: Option<PrimitiveDateTime>,
     pub fingerprint_id: Option<String>,
     pub request_external_three_ds_authentication: Option<bool>,
-    pub charges: Option<pii::SecretSerdeValue>,
```
jarnura, line 247: why charges got removed?

jarnura (changes requested):

crates/diesel_models/src/payment_intent.rs
Narayanbhat166: we should not be able to update `profile_id` in payment intent.

crates/diesel_models/src/payment_intent.rs
Narayanbhat166: This was not being used

jarnura (dismissed):

SanchithHegde (approved):

crates/hyperswitch_domain_models/src/type_encryption.rs
Narayanbhat166: Since there are some other changes as well in this area taken up by @ArjunKarthik, this can be done in a later PR

sai-harsha-vardhan (approved):

jarnura (approved):

sahkal (approved):

SanchithHegde (approved):

PR #4989, opened 2024-06-13
refactor: introduce ConnectorIntegrationNew and add default implementation for each Connector

hrithikesh026:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [x] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

Introduced ConnectorIntegrationNew and added default implementation for each Connector using macros

### Additional Changes

## How did you test it?

Checked if code compiles successfully

## Checklist


crates/hyperswitch_domain_models/src/router_data_v2.rs
```diff
@@ -0,0 +1,27 @@
+pub mod flow_common_types;
+
+use std::{marker::PhantomData, ops::Deref};
+
+pub use flow_common_types::*;
```
SanchithHegde, line 5: Nit: Can we avoid the glob imports everywhere (among all new imports added in this PR)?

hrithikesh026: Sure.
What is the advantage of doing this?

SanchithHegde: The primary reason I'd like not having glob imports is better code navigation. Sure, Ctrl/Cmd + clicking on the type name does take us to the relevant type when we're within the IDE, but if we're quickly checking code outside the IDE, or without rust-analyzer available, such as within the GitHub web UI, finding out where a type is made available from becomes quite difficult.

jarnura (approved):

SanchithHegde (approved):

PR #5006, opened 2024-06-13
refactor: add basic counter metrics for IMC

Chethan-rao:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [x] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

This PR introduces basic counter metrics for IMC -

* hits (successful read)
* misses (failed read)
* manual invalidations (done by applications) - makes use of eviction listener on moka that can help identify strategy based evictions

This PR also moves the `add_attribute` method of metrics to `router_env` crate. This method now takes a list of attributes as an argument as opposed to just one (key,value) pair.

### Additional Changes

## How did you test it?

1. Insert new config key value pair. Data is not cached in IMC during insertion

```
--data '{
    "key": "test_key",
    "value": "test_val_1"
}'
```

2. Retrieve the inserted config so that the data gets cached to IMC. Since this is the first time the data is getting retrieved, there'll be a cache mis

```
curl --location 'http://localhost:8080/configs/test_key' \
--data ''
```

image

3.  Again retrieve the same config. Since in the previous retrieve, the data was cached, this retrieve will be a cache hit

```
curl --location 'http://localhost:8080/configs/test_key' \
--data ''
```

image

4. Update the inserted config. This time the cache data has to be invalidated. This should increment the `IN_MEMORY_CACHE_EVICTION_COUNT` metric

```
curl --location 'http://localhost:8080/configs/test_key' \
--data '{
    "key": "test_key",
    "value": "test_val_2"
}'
```

image

## Checklist


crates/storage_impl/src/redis/cache.rs
```diff
@@ -173,19 +191,38 @@ impl From<CacheKey> for String {
...
+        max_capacity: Option<u64>,
+    ) -> Self {
+        // Record the metrics of manual invalidation of cache entry by the application
+        let eviction_listener = move |_, _, cause| {
+            if let RemovalCause::Explicit = cause {
+                metrics::CACHE_MANUAL_INVALIDATION.add(
+                    &metrics::CONTEXT,
+                    1,
+                    &request::add_attributes([("cache_type", name)]),
+                );
```
SanchithHegde, line 211: What do you think about incrementing this metric in the `remove()` method, and recording the eviction count in a separate metric (say `IN_MEMORY_CACHE_EVICTION_COUNT`) with the `RemovalCause` as an attribute? That way, we can also have information on the number of items removed due to expiration, values updated (`Replaced`) and due to size constraints. 

They can hopefully help us take informed decisions on the configuration values (TTL, TTI, capacity) for each of the in-memory cache instances.

crates/router_env/src/metrics/request.rs
```diff
@@ -0,0 +1,10 @@
+pub fn add_attributes<T, U>(attributes: U) -> Vec<opentelemetry::KeyValue>
```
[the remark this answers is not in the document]
Chethan-rao: let metrics module only have the macro definitions and the helper methods reside separately in this module?

crates/storage_impl/src/redis/cache.rs
Chethan-rao: instead of having this logic in `remove()` and having separate metric, can we have a single generic metric for eviction which has `RemovalCause` as an attribute. And this logic resides in eviction listener alone. this should do that job of counting the eviction for all cause?

SanchithHegde: Sure, that should be fine.

SanchithHegde (approved):

lsampras (approved):

PR #5048, opened 2024-06-19
feat: add mtls for keymanager api requests

dracarys18:
## Type of Change

- [x] Enhancement

## Description

Add mTLS while calling keymanager service

### Additional Changes

## Motivation and Context

This PR adds a new `reqwest::Client`  with Certificates and CA to contact with the key manager

## Checklist


crates/router/src/encryption.rs
```diff
@@ -1,38 +1,116 @@
...
+
+        #[cfg(feature = "keymanager_mtls")]
+        {
+            let ca = state.conf.key_manager.ca.clone();
+            let cert = state.conf.key_manager.cert.clone();
+
+            let identity = reqwest::Identity::from_pem(cert.peek().as_ref())
+                .change_context(errors::ApiClientError::ClientConstructionFailed)?;
+            let ca_cert = reqwest::Certificate::from_pem(ca.peek().as_ref())
+                .change_context(errors::ApiClientError::ClientConstructionFailed)?;
```
Narayanbhat166, line 34: you can use 
```rust
if cfg!(feature = "keymanager_mtls") 
```

you would not need #[allow(unused_mut)] in that case

dracarys18: We cannot do this since ca and cert have compile time feature constraints

ArjunKarthik (approved):
LGTM

crates/router/src/encryption.rs
```diff
@@ -1,38 +1,116 @@
...
 
 pub async fn call_encryption_service<T>(
     state: &SessionState,
     endpoint: &str,
     request_body: T,
-) -> errors::CustomResult<Result<Response, Response>, errors::ApiClientError>
+) -> errors::CustomResult<reqwest::Response, errors::ApiClientError>
 where
-    T: masking::ErasedMaskSerialize + Send + Sync + 'static,
+    T: serde::Serialize + Send + Sync + 'static,
```
Narayanbhat166, line 93: why have we removed masked serialize?

PR #5064, opened 2024-06-20
refactor(redis): spawn one subscriber thread for handling all the published messages to different channel

Chethan-rao:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [x] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

In the current Redis pub-sub implementation, a subscriber spawns a thread to handle all messages published to a channel. As a result, if a subscriber is invoked multiple times (even for subscribing to the same channel), it creates multiple threads for handling messages, which is not the desired behaviour. Ideally, we want to have only one subscriber thread handling all messages published to different channels. To achieve this, we maintain an atomic boolean variable, initially set to `false`, indicating whether the subscriber handler thread has been spawned. This variable will be set to `true` during the first subscriber invocation, thus preventing the creation of multiple threads during subsequent subscriber invocations. 

### Additional Changes

## How did you test it?

In the current implementation, the subscriber has been invoked 2 times thus resulting in 2 threads being spawned. Consequently, cache invalidation attempts occur in both threads, with only one succeeding. With this PR, there'll always be only one subscriber thread responsible for invalidating the cache entry.

* Send a request to the `/configs` endpoint to add a new test config.

```
--data '{
    "key": "test_key",
    "value": "test_val"
}'
```

* This action triggers cache invalidation.
* Check the request logs. You should see only one log entry stating: "Done invalidating test_key"

## Checklist

### DIFF
--- a/crates/storage_impl/src/redis/pub_sub.rs
+++ b/crates/storage_impl/src/redis/pub_sub.rs
@@ -1,3 +1,5 @@
+use std::sync::atomic;
+
 use error_stack::ResultExt;
 use redis_interface::{errors as redis_errors, PubsubInterface, RedisValue};
 use router_env::{logger, tracing::Instrument};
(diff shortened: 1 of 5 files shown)


crates/storage_impl/src/redis/pub_sub.rs
```diff
@@ -61,120 +77,128 @@ impl PubSubInterface for std::sync::Arc<redis_interface::RedisConnectionPool> {
...
-                Ok(value) => value,
-                Err(err) => {
-                    logger::error!(value_conversion_err=?err);
-                    continue;
-                }
-            };
+            let channel_name = message.channel.to_string();
+            match channel_name.as_str() {
+                super::cache::IMC_INVALIDATION_CHANNEL => {
+                    logger::debug!("Invalidating {message:?}");
```
SanchithHegde, line 83: Is it sufficient for us to log only the message value, or do we need to log all the fields (channel, server, etc.)?

SanchithHegde (dismissed):

crates/storage_impl/src/redis/pub_sub.rs
Chethan-rao: We are logging the entire `Message` struct. this does include all the fields (channel, value, server)

SanchithHegde (approved):

dracarys18 (approved):

Narayanbhat166 (approved):

PR #5120, opened 2024-06-25
feat(email): Add `auth_id` in email types and send `auth_id` in email URLs

### LINKED ISSUE #5119: feat(email): Send `auth_id` in email URLs
Currently links in the emails point to public dashboard. For SSO tenants/orgs, they should point to different url, which has `auth_id`.

ThisIsMani:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

This PR will introduce `auth_id` field in email types and also email APIs.

When an email is triggered from a API which gets `auth_id`, then that will be used to construct the email URL as well.

### Additional Changes

## Motivation and Context

Closes #5119.

## How did you test it?

All the following APIs will accept a query param called `auth_id`.
1. /user/connect_account
2. /user/forgot_password
3. /user/user/invite_multiple
4. /user/resend_invite
5. /user/verify_email_request

Example:
```
curl --location 'http://localhost:8080/user/connect_account?auth_id=2d784c51-c6ba-4adf-8d9f-407049ddacd2' \
--data-raw '{
    "email": "email"
}'
```

In the email received from the API, the url in the email will also have the same `auth_id` query param which is the same value as above.

`auth_id` param is optional, and if it is not sent, the email link also will not have any parameter.
## Checklist


crates/router/src/routes/user.rs
```diff
@@ -431,21 +439,31 @@ pub async fn reset_password(
...
     }
 }
+
 pub async fn invite_multiple_user(
     state: web::Data<AppState>,
     req: HttpRequest,
     payload: web::Json<Vec<user_api::InviteUserRequest>>,
-    query: web::Query<user_api::TokenOnlyQueryParam>,
+    token_only_query_param: web::Query<user_api::TokenOnlyQueryParam>,
+    auth_id_query_param: web::Query<user_api::AuthIdQueryParam>,
```
apoorvdixit88, line 448: how we are differentiating between two in request?

crates/router/src/routes/user.rs
```diff
@@ -39,15 +39,19 @@ pub async fn user_signup_with_merchant_id(
...
     let req_payload = json_payload.into_inner();
+    let auth_id = query.into_inner().auth_id;
     Box::pin(api::server_wrap(
         flow.clone(),
         state,
         &http_req,
         req_payload.clone(),
-        |state, _, req_body, _| user_core::signup_with_merchant_id(state, req_body),
+        |state, _, req_body, _| {
+            user_core::signup_with_merchant_id(state, req_body, auth_id.clone())
```
racnan, line 53: `clone()` not required.

racnan (approved):

crates/router/src/routes/user.rs
ThisIsMani: Will pickup in upcoming PRs.

lsampras (approved):

crates/router/src/routes/user.rs
ThisIsMani: Unable to do this, as the `ApiMetricEvent` should be implemented for payload. That means, if I use a tuple like `(req, auth_id)`, then I have to implement `ApiMetricEvent` for `(ReqType, String)`

PR #5161, opened 2024-06-28
fix(auth_methods): Add checks for duplicate `auth_method` in create API

### LINKED ISSUE #5160: bug(auth_methods): Create auth methods is allowing auth methods with same type to be inserted multiple times
Currently create will accept and insert auth methods with same type and name, which shouldn't be possible.

Create should have checks for this and block these kind of requests.

ThisIsMani:
## Type of Change

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

Currently create will accept and insert auth methods with same type and name, which shouldn't be possible. This PR fixes that.

### Additional Changes

## Motivation and Context

Closes #5160. 

## How did you test it?

```
curl --location 'http://localhost:8080/user/auth' \
--data '{
    "owner_id": "org_qaamgpukifSyBG0AxtYA2x",
    "owner_type": "organization",
    "auth_method": {
        "auth_type": "open_id_connect",
        "private_config": {
            "base_url": "https://dev-28418517.okta.com",
            "client_id": "0oahmmwdmuFvv2pFo5d7",
            "client_secret": "-VIrZZeN_A0SdSpFykAUZ0iMJNpSYQyILcfUmYlmZaLaFK7uRayrEuSvhs-Um5IR"
        },
        "public_config": {
            "name": "okta"
        }
    },
    "allow_signup": false
}'
```

If this API is hit with same `auth_type` and `name` in `public_config`, then the API will throw the following error.
```json
{
    "error": {
        "type": "invalid_request",
        "message": "User auth method already exists",
        "code": "UR_43"
    }
}
```

## Checklist


apoorvdixit88 (approved):
LGTM

crates/router/src/utils/user.rs
```diff
@@ -207,16 +207,62 @@ pub fn get_redis_connection(state: &SessionState) -> UserResult<Arc<RedisConnect
...
+            &user_api::AuthConfig::Password => Self::Password,
+            &user_api::AuthConfig::MagicLink => Self::MagicLink,
         }
     }
 }
 
+pub async fn construct_public_and_private_db_configs(
+    auth_config: &user_api::AuthConfig,
+    encryption_key: &[u8],
+) -> UserResult<(Option<Encryption>, Option<serde_json::Value>)> {
```
racnan, line 223: why `(Option<T1>,Option<T2>)` why not `Option<(T1,T2)>` if public and private config are coupled ?

crates/router/src/utils/user.rs
```diff
@@ -207,16 +207,62 @@ pub fn get_redis_connection(state: &SessionState) -> UserResult<Arc<RedisConnect
...
+    match auth_config {
+        user_api::AuthConfig::OpenIdConnect {
+            private_config,
+            public_config,
+        } => {
+            let private_config_value = serde_json::to_value(private_config.clone())
+                .change_context(UserErrors::AuthConfigParsingError)
```
racnan, line 230: why is this `400` ? isnt invalid value caught at api level ?
or someone can pass api model validation (serde) but still provide invalid json ?

crates/router/src/utils/user.rs
```diff
@@ -207,16 +207,62 @@ pub fn get_redis_connection(state: &SessionState) -> UserResult<Arc<RedisConnect
...
+            Ok((
+                Some(encrypted_config.into()),
+                Some(
+                    serde_json::to_value(public_config.clone())
+                        .change_context(UserErrors::AuthConfigParsingError)
+                        .attach_printable("Failed to convert auth config to json")?,
+                ),
+            ))
+        }
+        _ => Ok((None, None)),
```
racnan, line 250: lets match each value here, so that new enum breaks the code at compile time

crates/router/src/utils/user.rs
ThisIsMani: In DB, both are optional, so a case where when one is optional and other is not comes then, we will have to change this everywhere.

crates/router/src/utils/user.rs
```diff
@@ -207,16 +207,62 @@ pub fn get_redis_connection(state: &SessionState) -> UserResult<Arc<RedisConnect
...
+    public_config: Option<serde_json::Value>,
+) -> UserResult<Option<user_api::OpenIdConnectPublicConfig>> {
+    public_config
+        .map(|config| {
+            serde_json::from_value::<user_api::OpenIdConnectPublicConfig>(config)
+                .change_context(UserErrors::InternalServerError)
+                .attach_printable("Unable to parse OpenIdConnectPublicConfig")
+        })
+        .transpose()
+}
```
racnan, line 264: this can be generic instead exclusively for `OpenIdConnectPublicConfig`.

racnan (approved):

apoorvdixit88 (approved):

PR #5166, opened 2024-06-29
added support for reading locally generated file logs via vector inst…

aayushdhiman01:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

Added the code for reading locally generated file logs via vector instance, but need to verify that local logs are coming, can u guide how to check that?

### Additional Changes

## Checklist

### DIFF
--- a/config/vector.yaml
+++ b/config/vector.yaml
@@ -34,6 +34,19 @@ sources:
     address: 0.0.0.0:3103
     encoding: json
 
+  local_file_logs:
+    type: file
+    include:
+      - path: "/var/log/router/**/*.log*"
+    start_at_beginning: true
+    read_from_head: true
+    ignore_older: 0s
+    encoding:
+      multiline:
+        pattern: '^\['
+        negate: true
+        match: after
+
 transforms:
   plus_1_events:
     type: filter
@@ -45,12 +58,14 @@ transforms:
     type: filter
     inputs:
       - app_logs
+      - local_file_logs 
     condition: '.labels."com.docker.compose.service" == "hyperswitch-server"'
 
   parsed_hs_server_logs:
     type: remap
     inputs:
       - app_logs
+      - local_file_logs 
     source: |-
       .message = parse_json!(.message)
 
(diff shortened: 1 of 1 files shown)


lsampras:
Hey @aayushdhiman01 ,
Can you take a look at [this](https://github.com/juspay/hyperswitch/blob/main/monitoring%2Fdocker-compose.yaml)  for reference.

So the logs would be written in a shared local folder and we mount that folder in vector to read them,

You can enable file logging via the config file(development/docker_compose.toml), 

The vector/loki/grafana cotainers can be run via `docker compose --profile monitoring up -d`

aayushdhiman01:
@lsampras  can you check the changes made, mounted the logs folder in vector , in docker-compose file and enabled file logging in development.toml and docker_compose.toml file.

config/vector.yaml
```diff
@@ -45,12 +58,14 @@ transforms:
     type: filter
     inputs:
       - app_logs
+      - local_file_logs
```
lsampras, line 61: This might not be needed since only hyperswitch will write to file

config/vector.yaml
```diff
@@ -100,6 +115,7 @@ sinks:
     type: elasticsearch
     inputs:
       - parsed_hs_server_logs
+      - local_file_logs
```
lsampras, line 118: this won't be needed since the file logs are already present in `parsed_hs_server_logs`

config/vector.yaml
```diff
@@ -131,6 +147,7 @@ sinks:
     type: loki
     inputs:
       - parsed_hs_server_logs
+      - local_file_logs
```
lsampras, line 150: this won't be needed since the file logs are already present in `parsed_hs_server_logs`

lsampras (commented):
We need to add extra labels to distinguish between file & console logs,
can add a label like `source = console/file`

Also can you share screenshots of the file logs in loki

aayushdhiman01:
@lsampras i am not able to connect to loki server in grafana, made the commented changes.

PR #5169, opened 2024-07-01
refactor(migrations): add commands to make file to run migrations for api v2

Narayanbhat166:
## Type of Change

- [x] New feature

## Description

The diesel migration command runs the migrations from the migrations directory. There is no distinction between whether these migrations have to be run before the deployment or after the deployment. 

This PR introduces a separate migrations directory, in order to run backwards incompatible migrations. In order to run the migrations, run the below command

```bash
make migrate
```

In order to revert the migrations
```bash
make revert_migrate
```

## Note on running migrations in higher environments
When running the migrations in higher environments, we need not run the diesel migration run command. It is always better to run the individual migrations to ensure that all the migrations that are run are backwards compatible. It also depends on when to run the migrations, some migrations can be run before the new version is deployed ( migrations which are backwards compatible ), and few migrations have to be run after the new version of application is deployed ( which are not backwards compatible ). The database admin / the user who is running these queries should be aware of what commands are being run and whether this will affect the current application. 

## How did you test it?

Run the commands locally

## Checklist


Makefile
```diff
@@ -113,3 +113,37 @@ precommit : fmt clippy test
...
+	mkdir -p $(ROOT_DIR)/tmp/migrations
+	find $(ROOT_DIR)/tmp/migrations/ -mindepth 1 -delete
+
+	cp -r $(ROOT_DIR)/migrations/. $(ROOT_DIR)/v2_migrations/. $(ROOT_DIR)/tmp/migrations/
+	diesel migration redo --all --migration-dir=$(ROOT_DIR)/tmp/migrations \
+		$(if $(strip $(database-url)),--database-url="$(database-url)",) \
+		$(if $(strip $(call eq,$(locked-schema),yes)),--locked-schema,)
+
+	rm -r $(ROOT_DIR)/tmp/migrations
+	rmdir $(ROOT_DIR)/tmp 2>/dev/null || true
```
SanchithHegde, line 149: Nit: Trailing whitespace.

```suggestion
	rmdir $(ROOT_DIR)/tmp 2>/dev/null || true
```

docker-compose-development.yml
SanchithHegde, line 1: Can you make the same changes in the `docker-compose.yml` file as well?

docs/try_local_system.md
```diff
@@ -476,10 +478,12 @@ Once you're done with setting up the dependencies, proceed with
...
 
-3. Run database migrations using `diesel_cli`:
+3. Run database migrations:
+
+   Note that this requires `diesel_cli` to be installed
 
    ```shell
-   diesel migration --database-url postgres://$DB_USER:$DB_PASS@localhost:5432/$DB_NAME run
+   make migrate database-url=postgres://$DB_USER:$DB_PASS@localhost:5432/$DB_NAME
    ```
```
SanchithHegde, line 487: We need not mention that they'd need `diesel_cli`, since our setup guides already include steps for installing it. However, we'd have to update our guides to include `make`, just in case.

- Update the Ubuntu and Ubuntu on WSL2 guides to install the `make` package using `apt`. (You can edit the existing commands which install `pkg-config` and other packages.
- I _think_ recent versions of MacOS have `make` installed by default.
- For native Windows, we'll have to include a command like so: `winget install ezwinports.make`. (Although from what I remember, we'd have to revisit native Windows setup entirely.)

Narayanbhat166: > Update the Ubuntu and Ubuntu on WSL2 guides to install the make package using apt

The build essential package already includes make, we are installing this on wsl. Have made the changes to ubuntu setup guide to install make along with openssl.

SanchithHegde (approved):

sai-harsha-vardhan (approved):

PR #5178, opened 2024-07-02
refactor(analytics): Create new permissions for Generate Report APIs

### LINKED ISSUE #5177: refactor(reports): Change permissions for report APIs
Currently internal users are unable to use report APIs as some of them require Write permission which internal view only users doesn't have.

We need to change the permission required to access report APIs.

ThisIsMani:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [x] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

Currently, internal users are unable to access payment generate report as it requires `PaymentWrite`.

And other reports are having `Analytics` permission, but in the Front-end, the option is in Operations tab and in Analytics tab.

Because Generate reports option is present in both Analytics and Operations, there will be a new permission, which will be available for both Operations and Analytics groups.

### Additional Changes

## Motivation and Context

Closes #5177.

## How did you test it?

```
curl 'http://localhost:8080/analytics/v1/report/payments' \
  -H 'authorization: Bearer Internal user JWT' \
  --data-raw '{"timeRange":{"startTime":"2024-07-01T11:34:38Z","endTime":"2024-07-02T11:34:38.441Z"},"dimensions":[]}'
```

When this API is hit by a internal user, this API should send email with report.

## Checklist


crates/router/src/analytics.rs
```diff
@@ -592,7 +592,7 @@ pub mod routes {
                 .await
                 .map(ApplicationResponse::Json)
             },
-            &auth::JWTAuth(Permission::Analytics),
+            &auth::JWTAuth(Permission::RefundRead),
```
ivor-juspay, line 595: will users with `RefundWrite` get `RefundRead` permission too?

ivor-juspay (approved):

crates/router/src/analytics.rs
ThisIsMani: User Roles are made up of groups, and those groups is a set of permissions.

So if there is a group names `OperationsWrite`, then in that group, we put all the read and write permissions for operations like payments, refunds, etc..


So, users will never have only `RefundWrite` permission, if they have it, they will also have `RefundRead` as well. As both of them will be in the same group.

lsampras (approved):

ivor-juspay (approved):

racnan (approved):

PR #5219, opened 2024-07-05
feat(payout_link): secure payout links using server side validations and client side headers

kashif-m:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description
This PR includes below changes -
- adding `allowed_domains` field in payout link config
  - this is added at both business profile level and individual payout txns
  - business profile config is used as a fallback in case `allowed_domains` are not passed in payout link's create request
  - this is a required field in production env
- server side validations for incoming render request
  - check `Sec-Fetch-Dest` [[ref](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Sec-Fetch-Dest)] for request source (must be `iframe`)
  - check `Origin / Referer` for the host of the incoming request (must match one in `allowed_domains`)
  - these are skipped if `allowed_domains` is not setup (only in non-production env)
- client side headers
  - attach CSP header - `frame-ancestors` for directing which hosts can embed the links in an iframe [[ref](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/frame-ancestors)]
  - ~~attach X-Frame header - similar to CSP, but for older browsers [[ref](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Frame-Options)]~~ _Removing this as it's outdated_
- update status UI for payout links

feat(generic_link): add allowed_domains in GenericLink response
feat(payout_link): add allowed_domains in payout_link_config (fallback profile config + payout request config)
feat(payout_link): consume meta headers for validating render request for payout links
feat(payout_link): consume CSP and X-Frame headers in the link's response

### Additional Changes

## How did you test it?
**Tested locally using postman**

1. Update business profile's `payout_link_config` to include `allowed_domains`
```
curl --location 'http://localhost:8080/account/merchant_1721206540/business_profile/pro_iMeDAsjCBueHxdjCiZ9I' \
--data '{
    "payout_link_config": {
        "allowed_domains": ["127.0.0.1:5500"]
    }
}'
```

2. Create a payout link
```
curl --location 'http://localhost:8080/payouts/create' \
--header 'api-key: dev_Ew5ejag6iXMc80CDUUflU4daqTvm0wsMc1AgUo174U7iaV1GFvf8kkwt8LLCxDkT' \
--data '{
    "amount": 3000,
    "currency": "EUR",
    "billing": {
        "address": {
            "city": "Hoogeveen",
            "country": "NL",
            "line1": "Raadhuisplein",
            "line2": "92",
            "zip": "7901 BW",
            "state": "FL"
        },
        "phone": {
            "number": "0650242319",
            "country_code": "+31"
        }
    },
    "payout_link": true,
    "payout_link_config": {
        "theme": "#143F1E",
        "logo": "https://upload.wikimedia.org/wikipedia/commons/8/84/Spotify_icon.svg",
        "merchant_name": "Spotify"
    }
}'
```
3. Try opening this link in a new tab (should not open)
4. Open this inside an iframe (should open provided the `allowed_domains` and the iframe's host is same)

**Test cases**
```
cargo test --package common_utils --lib -- link_utils::domain_tests --show-output
```
image

## Checklist


crates/api_models/src/admin.rs
```diff
@@ -1332,11 +1332,39 @@ pub struct BusinessGenericLinkConfig {
...
+        // Validate host domain name
+        let host_domain_valid = self
+            .domain_name
+            .clone()
+            .map(|host_domain| link_utils::validate_strict_domain(&host_domain))
+            .unwrap_or(true);
+        if !host_domain_valid {
+            return Err("Invalid host domain name received");
+        }
+
```
sahkal, line 1354: Why `true` you can have it as false right, by that way you don't have to do reverse if logic

crates/hyperswitch_domain_models/src/errors/api_error_response.rs
```diff
@@ -523,6 +525,9 @@ impl ErrorSwitch<api_models::errors::types::ApiErrorResponse> for ApiErrorRespon
             Self::AddressNotFound => {
                 AER::NotFound(ApiError::new("HE", 4, "Address does not exist in our records", None))
             },
+            Self::GenericConfigurationError { message } => {
```
sahkal, line 528: +1

crates/hyperswitch_domain_models/src/api.rs
```diff
@@ -59,20 +59,26 @@ pub struct PaymentLinkStatusData {
 }
 
 #[derive(Debug, Eq, PartialEq)]
-pub enum GenericLinks {
+pub struct GenericLinks {
+    pub allowed_domains: Option<HashSet<String>>,
```
sahkal, line 63: this shouldn't be optional since hashset is a required field in business profile

crates/router/src/core/generic_link/payout_link/initiate/script.js
```diff
@@ -1,153 +1,171 @@
...
-    })
-    .substring(4)
-    .split(" ")
-    .reduce(function (tz, c) {
-      return tz + c.charAt(0).toUpperCase();
-    }, "");
+  // webpage is iframed, good to load
+} else {
+  var hyper = null;
+  var payoutWidget = null;
```
sahkal, line 29: can you get this checked from @seekshiva

crates/router/src/core/payouts/validator.rs
```diff
@@ -317,3 +331,71 @@ pub async fn create_payout_link_db_entry(
...
+                "Access to payout_link [{}] is forbidden when requested through {}",
+                link_id, requestor
+            )
+        }),
+        None => Err(report!(errors::ApiErrorResponse::AccessForbidden {
+            resource: "payout_link".to_string(),
+        }))
+        .attach_printable_lazy(|| {
+            format!(
+                "Access to payout_link [{}] is forbidden when sec-fetch-dest is not present in request headers",
```
sahkal, line 359: can we have a seperate forked UI for these failures doesn't look good throwing json error in user flow

crates/router/src/core/payouts/validator.rs
```diff
@@ -231,6 +227,16 @@ pub async fn create_payout_link(
...
 
+    // Validate allowed_domains presence
+    let allowed_domains = profile_config
+        .as_ref()
+        .map(|config| config.config.allowed_domains.to_owned())
+        .get_required_value("allowed_domains")
+        .change_context(errors::ApiErrorResponse::GenericConfigurationError {
+            message: "Payout links cannot be used without setting allowed_domains in profile"
+                .to_string(),
+        })?;
```
sahkal, line 238: Why is this fn `create_payout_link` inside validator.rs file

sahkal (commented):
There seems to be a bug in the payout link configs, since we are focusing on json. if one of the field is not sent in payment create and is set in business config, we will make the not sent value as None. Which is not the expected behavior i suppose.

crates/router/src/core/payment_methods.rs
```diff
@@ -302,7 +308,10 @@ pub async fn render_pm_collect_link(
                     html_meta_tags: String::new(),
                 };
                 Ok(services::ApplicationResponse::GenericLinkForm(Box::new(
-                    GenericLinks::PaymentMethodCollect(generic_form_data),
+                    GenericLinks {
+                        allowed_domains: None,
```
sahkal, line 312: can we pass it as empty hashset, if we are going to make it a strict type

crates/hyperswitch_domain_models/src/errors/api_error_response.rs
```diff
@@ -231,6 +231,8 @@ pub enum ApiErrorResponse {
     MissingFilePurpose,
     #[error(error_type = ErrorType::InvalidRequestError, code = "HE_04", message = "File content type not found / valid")]
     MissingFileContentType,
+    #[error(error_type = ErrorType::InvalidRequestError, code = "HE_04", message = "{message}")]
+    GenericConfigurationError { message: String },
```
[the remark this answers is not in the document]
kashif-m: This is not specific to any links. It's specific to any sort of `Configuration` errors, we can make it `LinkConfigurationError` if needed

crates/router/src/core/payouts/validator.rs
```diff
@@ -317,3 +331,71 @@ pub async fn create_payout_link_db_entry(
...
+                "Access to payout_link [{}] is forbidden from requestor - {}",
+                link_id, domain_in_req
+            )
+        })
+    }
+}
+
+fn is_domain_allowed(domain: &str, allowed_domains: HashSet<String>) -> bool {
+    allowed_domains.iter().any(|allowed_domain| {
+        Regex::new(allowed_domain)
```
SanchithHegde, line 396: You cannot do this, allowed domains are not valid regex patterns in the first place. If you consider `*.example.com`, this is an invalid regex pattern.
Secondly, I think accepting regexes as user input is a bad idea.

crates/router/src/types/api/admin.rs
```diff
@@ -171,6 +171,21 @@ impl ForeignTryFrom<(domain::MerchantAccount, BusinessProfileCreate)>
             })
             .transpose()?;
 
+        let payout_link_config = request
+            .payout_link_config
+            .as_ref()
+            .map(|payout_conf| match payout_conf.config.validate() {
+                Ok(_) => Encode::encode_to_value(payout_conf).change_context(
```
SanchithHegde, line 178:
```suggestion
                Ok(_) => payout_conf.encode_to_value().change_context(
```

crates/api_models/src/admin.rs
```diff
@@ -1332,11 +1332,39 @@ pub struct BusinessGenericLinkConfig {
...
+        let are_allowed_domains_valid = self
+            .allowed_domains
+            .clone()
+            .iter()
+            .any(|allowed_domain| link_utils::validate_wildcard_domain(allowed_domain));
```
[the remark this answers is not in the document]
kashif-m: The origin / referer in incoming request needs to be validated against a list of allowed domains, so as long as even one of the configured domain matches, it will be validated

seekshiva (approved):

sahkal (approved):

PR #5262, opened 2024-07-09
fix(payments_create): save the `customer_id` in payments create

Narayanbhat166:
## Type of Change

- [x] Bugfix

## Description

In case we receive a 4xx when creating the payment before calling the connector, we do not save the `customer_id` in payment intent. This PR fixes it

This PR also includes changes to ensure that `customer_id` is always taken from both the places in the PaymentsRequest. The root `customer_id` and `customer.id` field.

## How did you test it?

- Create a merchant account
```bash
--data '{
    "merchant_id": "merchant_1720593206"
}'
```

- Create an api key
```bash
curl --location 'http://localhost:8080/api_keys/merchant_1720593134' \
--data '{
  "name": "API Key 1",
  "description": null,
  "expiration": "2038-01-19T03:14:08.000Z"
}'
```

- Create a payment
```bash
curl --location 'http://localhost:8080/payments' \
--header 'api-key: dev_RMV9h1z6QVqWHW9B7LBRl6peSokwQ7go96pfAjehuGZDGiwlenKAbgtyqUlXny9J' \
--data '{
    "amount": 0,
    "payment_id": "pay_1720593248",
    "customer_id": "cus_123",
    "currency": "USD",
    "confirm": true,
    "payment_type": "setup_mandate",
    "payment_method": "card",
    "payment_method_data": {
        "card": {
            "card_number": "4111111111111111",
            "card_exp_month": "10",
            "card_exp_year": "25",
            "card_holder_name": "joseph Doe",
            "card_cvc": "123"
        }
    }
}'
```
Fails with the error message
```json
{
    "error": {
        "type": "invalid_request",
        "message": "No eligible connector was found for the current payment method configuration",
        "code": "HE_04"
    }
}
```

- Check in the database whether `customer_id` is updated
image

## Checklist

### DIFF
--- a/crates/api_models/src/payments.rs
+++ b/crates/api_models/src/payments.rs
@@ -518,6 +518,62 @@ pub struct PaymentsRequest {
     pub merchant_order_reference_id: Option<String>,
 }
 
+/// Checks if the inner values of two options are equal
+/// Returns true if values are not equal, if values match returns false
+fn are_optional_values_invalid<T: PartialEq>(
+    first_option: Option<&T>,
+    second_option: Option<&T>,
+) -> bool {
+    first_option
+        .zip(second_option)
+        .map(|(value1, value2)| value1 != value2)
+        .unwrap_or(true)
+}
+
+impl PaymentsRequest {
+    /// Get the customer id
+    ///
+    /// First check the id for `customer.id`
+    /// If not present, check for `customer_id` at the root level
+    pub fn get_customer_id(&self) -> Option<&id_type::CustomerId> {
+        self.customer_id
+            .as_ref()
+            .or(self.customer.as_ref().map(|customer| &customer.id))
+    }
+
+    /// Checks if the customer details are passed in both places
+    /// If they are passed in both places, check for both the values to be equal
+    /// Or else, return the field which has inconsistent data
+    pub fn validate_customer_details_in_request(&self) -> Option<&str> {
+        if let Some(customer_details) = self.customer.as_ref() {
+            if are_optional_values_invalid(self.customer_id.as_ref(), Some(&customer_details.id)) {
+                return Some("customer_id");
+            }
+
+            if are_optional_values_invalid(self.email.as_ref(), customer_details.email.as_ref()) {
+                return Some("email");
+            }
+
+            if are_optional_values_invalid(self.name.as_ref(), customer_details.name.as_ref()) {
+                return Some("name");
+            }
+
+            if are_optional_values_invalid(self.phone.as_ref(), customer_details.phone.as_ref()) {
+                return Some("phone");
+            }
+
+            if are_optional_values_invalid(
+                self.phone_country_code.as_ref(),
+                customer_details.phone_country_code.as_ref(),
+            ) {
+                return Some("phone_country_code");
+            }
+        }
+
+        None
+    }
+}
+
 /// Fee information to be charged on the payment being collected
 #[derive(Debug, serde::Deserialize, serde::Serialize, Clone, ToSchema)]
 #[serde(rename_all = "snake_case")]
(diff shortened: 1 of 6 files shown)


crates/api_models/src/payments.rs
```diff
@@ -518,6 +518,62 @@ pub struct PaymentsRequest {
...
+/// Returns true if values are not equal, if values match returns false
+fn are_optional_values_invalid<T: PartialEq>(
+    first_option: Option<&T>,
+    second_option: Option<&T>,
+) -> bool {
+    first_option
+        .zip(second_option)
+        .map(|(value1, value2)| value1 != value2)
+        .unwrap_or(true)
+}
```
hrithikesh026, line 531: This function is supposed to return `true` only when two options are Some and the inner value is different right?

But if one or more options are None, this will return `true` as well.

crates/api_models/src/payments.rs
```diff
@@ -518,6 +518,62 @@ pub struct PaymentsRequest {
...
+        self.customer_id
+            .as_ref()
+            .or(self.customer.as_ref().map(|customer| &customer.id))
+    }
+
+    /// Checks if the customer details are passed in both places
+    /// If they are passed in both places, check for both the values to be equal
+    /// Or else, return the field which has inconsistent data
+    pub fn validate_customer_details_in_request(&self) -> Option<&str> {
+        if let Some(customer_details) = self.customer.as_ref() {
```
hrithikesh026, line 548: You can de-structure `customer_details` so that any new fields introduced to `CustomerDetails` will not affect this logic.

crates/api_models/src/payments.rs
```diff
@@ -518,6 +518,62 @@ pub struct PaymentsRequest {
...
+    pub fn get_customer_id(&self) -> Option<&id_type::CustomerId> {
+        self.customer_id
+            .as_ref()
+            .or(self.customer.as_ref().map(|customer| &customer.id))
+    }
+
+    /// Checks if the customer details are passed in both places
+    /// If they are passed in both places, check for both the values to be equal
+    /// Or else, return the field which has inconsistent data
+    pub fn validate_customer_details_in_request(&self) -> Option<&str> {
```
hrithikesh026, line 547: Wouldn't it be better to return a list of mismatched field names instead.

crates/router/src/core/payments/helpers.rs
```diff
@@ -1536,12 +1494,7 @@ pub fn validate_customer_details_in_request(
 pub fn get_customer_details_from_request(
     request: &api_models::payments::PaymentsRequest,
 ) -> CustomerDetails {
-    let customer_id = request
-        .customer
-        .as_ref()
-        .map(|customer_details| &customer_details.id)
-        .or(request.customer_id.as_ref())
-        .map(ToOwned::to_owned);
+    let customer_id = request.get_customer_id().map(ToOwned::to_owned);
```
hrithikesh026, line 1498: nit pick: 

```suggestion
    let customer_id = request.get_customer_id().cloned();

```

hrithikesh026 (approved):

crates/api_models/src/payments.rs
```diff
@@ -518,6 +518,139 @@ pub struct PaymentsRequest {
...
+    }
+}
+
+impl PaymentsRequest {
+    /// Get the customer id
+    ///
+    /// First check the id for `customer.id`
+    /// If not present, check for `customer_id` at the root level
+    pub fn get_customer_id(&self) -> Option<&id_type::CustomerId> {
+        self.customer_id
```
hrithikesh026, line 539: Since `customer_id` is public, it doesn't stop developers from fetching it directly from `PaymentRequest` in future. Can we make it private so that `get_customer_id()` is the only way to get customer_id from `PaymentRequest`

hrithikesh026 (approved):

sai-harsha-vardhan (approved):

PR #5293, opened 2024-07-11
fix(cypress): validate false positives and fix PML tests

pixincreate:
## Type of Change

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [x] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [x] CI/CD

## Description

TL:Dr: This PR fixes the false positives by adding a check and also fixes PML tests from throwing `422`.

When we get a 4xx response but in configs, we are expecting a 2xx, the tests would pass which is unexpected and a false positive.   
Also, PML connector have been failing due to changed creds contract. This PR also fixes that.

### Additional Changes

## Motivation and Context

@Sakilmostak 

## How did you test it?

PML Tests run without any issues:

image

This fails at PML for connectors other than `Stripe` and `Cybersource` and is outside the scope of this PR. 

CC: @prajjwalkumar17 

## Checklist

### DIFF
--- a/cypress-tests/cypress/e2e/PaymentMethodListTest/00000-PaymentMethodListTests.cy.js
+++ b/cypress-tests/cypress/e2e/PaymentMethodListTest/00000-PaymentMethodListTests.cy.js
@@ -1,18 +1,17 @@
 import apiKeyCreateBody from "../../fixtures/create-api-key-body.json";
 import createConnectorBody from "../../fixtures/create-connector-body.json";
-import getConnectorDetails from "../PaymentMethodListUtils/utils";
 import merchantCreateBody from "../../fixtures/merchant-create-body.json";
-import * as utils from "../PaymentMethodListUtils/Utils";
+import State from "../../utils/State";
 import {
+  bank_redirect_ideal_and_credit_enabled,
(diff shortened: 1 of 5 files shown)


Sakilmostak (approved):

cypress-tests/cypress/e2e/PaymentMethodListTest/00000-PaymentMethodListTests.cy.js
pixincreate, line 1: Lints.

pixincreate, line 1: These are just lints.

cypress-tests/cypress/e2e/PaymentUtils/Utils.js
pixincreate, line 1: When we get a 4xx, but we expect it to be a 2xx, we are supposed to throw an error notifying the user that something is wrong.  

This is a partial implementation of getting things right. This does not address all of false positives. To hav e100% coverage, we need to refactor `commands.js`

likhinbopanna (approved):

PR #5379, opened 2024-07-19
fix: use encrypt api for all encryption and decryption

ArjunKarthik:
## Type of Change

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description
encrypt/decrypt should happen only via encryption service and DEK should be same in application and encryption service

### Additional Changes

## Motivation and Context
encrypt/decrypt should happen only via encryption service and DEK should be same in application and encryption service

## How did you test it?
Changes can be verified only through logs. Look for "Fall back to Application Encryption" or "Fall back to Application Decryption" in the logs post deployment after running below tests. If logs are present encryption/decryption failed with encryption service and fall back to the application encryption which has to be investigated.

Test with merchant created before the changes deployed
1. Create Payment
2. Retrieve Payment
3. Create Refund
4. Retrieve Payment
Test with merchant created after the changes deployed
1. Create Merchant
2. Create API Key
3. Create MCA
4. Create Payment
6. Retrieve Payment
7. Create Refund
8. Retrieve Payment
Requires sanity on all the flows

## Checklist

### DIFF
--- a/crates/router/src/types/domain/user.rs
+++ b/crates/router/src/types/domain/user.rs
@@ -6,8 +6,6 @@ use api_models::{
 use common_enums::TokenPurpose;
 #[cfg(not(feature = "v2"))]
 use common_utils::id_type;
-#[cfg(feature = "keymanager_create")]
-use common_utils::types::keymanager::EncryptionCreateRequest;
 use common_utils::{
     crypto::Encryptable, errors::CustomResult, new_type::MerchantName, pii,
     types::keymanager::Identifier,
@@ -25,6 +23,8 @@ use once_cell::sync::Lazy;
 use rand::distributions::{Alphanumeric, DistString};
 use router_env::env;
 use unicode_segmentation::UnicodeSegmentation;
+#[cfg(feature = "keymanager_create")]
+use {base64::Engine, common_utils::types::keymanager::EncryptionTransferRequest};
 
 use crate::{
     consts,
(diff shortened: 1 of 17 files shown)


dracarys18 (approved):

Narayanbhat166 (approved):

crates/router/src/types/domain/user.rs
```diff
@@ -988,18 +1001,6 @@ impl UserFromStorage {
...
-                        identifier: Identifier::User(key_store.user_id.clone()),
-                    },
-                )
-                .await
-                .change_context(UserErrors::InternalServerError)?;
-            }
-
             state
                 .global_store
                 .insert_user_key_store(key_manager_state, key_store, &master_key.to_vec().into())
```
ThisIsMani, line 1006: What will happen to transferred key if the db insertion fails here? Will it also be deleted?

ArjunKarthik: No that won't be removed.

ThisIsMani: Will that doesn't cause inconsistency?

ArjunKarthik: No, if new user store created for the same user it will be updated in the key store

ThisIsMani: Discussed with @ArjunKarthik, came to a conclusion that this function is temporary and transfer will override the service.

ThisIsMani (approved):

PR #5383, opened 2024-07-21
feat(analytics): populate status_code, initial_attempt_id & delivery_attempt on clickhouse for outgoing webhook events

### LINKED ISSUE #5070: [FEATURE] Populate status code and delivery attempt information in outgoing webhooks ClickHouse events
### Description

As of opening this issue, the outgoing webhooks ClickHouse events contain a few fields:

https://github.com/juspay/hyperswitch/blob/ca61e47585071865cf7df5c05fdbe3f57818ca95/crates/router/src/events/outgoing_webhook_logs.rs#L11-L21

There are `error` and `is_error` fields that indicate whether an error occurred when delivering the webhook to the merchant server, but do not include any information on whether the merchant server even received the webhook.

To address this issue, we must include the following optional fields in the ClickHouse events:

- `status_code`
- `delivery_attempt`
- `initial_attempt_id`

The fields would have to be added in the `OutgoingWebhookEvent` struct, and the corresponding columns must be added in the ClickHouse table. Once the fields have been added in code, the fields must be populated from the [`Event` in the database](https://github.com/juspay/hyperswitch/blob/ca61e47585071865cf7df5c05fdbe3f57818ca95/crates/router/src/types/domain/event.rs#L15-L31):

- The `delivery_attempt` and `initial_attempt_id` are available directly in the `Event` struct.
- The `status_code` field is available within the `response` struct, if the webhook was delivered to the merchant server.

Loaki07:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

Closes #5070.

This PR enhances the outgoing_webhook_events and outgoing_webhook_events_audit tables by adding three new columns: initial_attempt_id, status_code, and delivery_attempt. These additions are essential for tracking the initial_attempt_id, status_code, and delivery_attempt status for outgoing webhook events, allowing for more detailed monitoring and troubleshooting of webhook deliveries to merchant servers.

### Database Schema Changes On Clickhouse

Please run the following `ALTER TABLE` commands to update the database schema:

#### `outgoing_webhook_events` Table
```sql
ALTER TABLE outgoing_webhook_events
ADD COLUMN initial_attempt_id Nullable(String),
ADD COLUMN status_code Nullable(UInt16),
ADD COLUMN delivery_attempt LowCardinality(String);
```

#### `outgoing_webhook_events_audit` Table
```sql
ALTER TABLE outgoing_webhook_events_audit
ADD COLUMN initial_attempt_id Nullable(String),
ADD COLUMN status_code Nullable(UInt16),
ADD COLUMN delivery_attempt LowCardinality(String);
```

### Additional Changes

## How did you test it?
Tested locally. 
outgoing_webhook_events
outgoing_webhook_events_audit

## Checklist


crates/router/src/core/webhooks/outgoing.rs
```diff
@@ -290,7 +290,7 @@ async fn trigger_webhook_to_merchant(
     );
     logger::debug!(outgoing_webhook_response=?response);
 
-    match delivery_attempt {
+    let updated_event: Option<domain::Event> = match delivery_attempt {
```
SanchithHegde, line 293: We're returning `None` from the `match` block only in case of API client errors, after `api_client_error_handler()` is called. And we're returning an error at the end of `api_client_error_handler()`:
https://github.com/juspay/hyperswitch/blob/623cf4c841847f109597ee50017d6cfc0b4d9982/crates/router/src/core/webhooks/outgoing.rs#L736

We should be able to directly return `domain::Event` instead of `Option<domain::Event>` from the match block and this function.

crates/router/src/core/webhooks/outgoing.rs
```diff
@@ -424,46 +430,60 @@ async fn trigger_webhook_to_merchant(
...
-            .ok()
-    } else {
-        None
-    };
+    match trigger_webhook_result {
+        Ok(updated_event) => {
+            if let Some(updated_event) = updated_event {
+                event = updated_event
+            }
+        }
```
SanchithHegde, line 456: Once you update the return type of the function to be `domain::Event` instead of `Option<domain::Event>`, this would have to be suitably updated.

crates/router/src/core/webhooks/outgoing.rs
```diff
@@ -424,46 +430,60 @@ async fn trigger_webhook_to_merchant(
...
+    }
 
     let outgoing_webhook_event_content = content
         .as_ref()
         .and_then(api::OutgoingWebhookContent::get_outgoing_webhook_event_content)
         .or_else(|| get_outgoing_webhook_event_content_from_event_metadata(event.metadata));
 
+    // Extract the status_code field from the webhook response
+    let status_code = utils::extract_value_from_response(&event.response, "status_code");
```
SanchithHegde, line 476: This need not be a utility function, you can call `event.response.parse_value()` and then access the `status_code` field. You can remove the utility function that was added.

crates/router/src/core/webhooks/outgoing.rs
Loaki07: Yeah, makes sense.

crates/router/src/core/webhooks/outgoing.rs
SanchithHegde: Not absolutely necessary, you can call `event.response.peek().parse_value()`.

crates/router/src/core/webhooks/outgoing.rs
```diff
@@ -733,7 +750,7 @@ async fn api_client_error_handler(
         .change_context(errors::WebhooksFlowError::OutgoingWebhookRetrySchedulingFailed)?;
     }
 
-    Err(error)
+    Ok(event)
```
SanchithHegde, line 753: Let's not change the return value here, we'd want this function to return an error itself.

Let me clarify:

We want the `trigger_webhook_to_merchant()` function to return errors in case the client faces an error when sending the webhook (i.e., `api_client_error_handler()`), or if the merchant server returns a non-2xx status code (i.e., `error_response_handler()`). This is because when we are firing the analytics event, we check if the `trigger_webhook_to_merchant()` function returned an error, and mark the webhook delivery attempt as unsuccessful (in the ClickHouse event).

Keeping this is mind, we cannot return the updated event when the merchant server returns a non-2xx status code, because `error_response_handler()` returns an error which is propagated by the `?`, and we just cannot return the updated event after that function call. The same applies to the `api_client_error_handler()` invocation as well. This requires us to fetch the updated event in the `raise_webhooks_analytics_event()` function.

So, let's revert all changes to the `trigger_webhook_to_merchant()` and `api_client_error_handler()` functions, and update the `raise_webhooks_analytics_event()` function to fetch the updated event from the database with the merchant ID and event ID, and populate the fields in the analytics event from this updated event object.

My bad, I should have caught this during my previous review.

crates/router/src/core/webhooks/outgoing.rs
Loaki07: I feel like I'm missing something here. Please clarify. 

In this case `peek()` return `&String` and only available method is `parse_struct()`.
And when attempting to do`event.response.parse_value("status_code")` it requires `ValueExt` trait to be implemented for `Secret<String>`

SanchithHegde: My bad, it's `parse_struct()`.

crates/router/src/core/webhooks/outgoing.rs
Loaki07: Got it, will revert the changes and fetch the updated even from db.

Loaki07:
@SanchithHegde I've made the discussed changes. Kindly, review.

crates/router/src/core/webhooks/outgoing.rs
```diff
@@ -457,15 +473,40 @@ fn raise_webhooks_analytics_event(
...
+        .await
+        .change_context(errors::StorageError::ValueNotFound(debug.clone()))
+        .attach_printable(debug)?;
+
+    // Get status_code from webhook response
+    let webhook_response: Option<OutgoingWebhookResponseContent> = event
+        .response
+        .map(|res| res.peek().to_owned())
+        .and_then(|res| res.parse_struct("OutgoingWebhookResponseContent").ok());
+    let status_code = webhook_response.and_then(|res| res.status_code);
```
SanchithHegde, line 495: 1. Let's not propagate the error here with `?`, but instead log it and populate `None` for the `status_code` in the analytics event if we're unable to retrieve the event from the database. We'd like to raise an analytics event, even if we're not able to obtain the status code that the merchant server returned.
2. Use `attach_printable_lazy()` instead of `attach_printable()` and move the `format!()` call within `attach_printable_lazy()`, since `format!()` involves a heap allocation.
3. Log the error if deserializing the response as `OutgoingWebhookResponseContent` fails.
4. Once these changes have been incorporated, the function no longer would be required to return a `Result`.

SanchithHegde (commented):
Also, you need not frequently pull changes from the `main` branch unless you have conflicts. The merge queue will handle merging your changes with the latest `main` branch and running tests on the resulting branch.

crates/router/src/core/webhooks/outgoing.rs
```diff
@@ -352,7 +364,7 @@ async fn trigger_webhook_to_merchant(
                         delivery_attempt,
                         ScheduleWebhookRetry::WithProcessTracker(process_tracker),
                     )
-                    .await?;
+                    .await?
```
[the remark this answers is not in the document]
Loaki07: Sure, it would now return the error.

crates/router/src/core/webhooks/outgoing.rs
```diff
@@ -457,13 +473,50 @@ fn raise_webhooks_analytics_event(
...
+            &merchant_id,
+            &event_id,
+            merchant_key_store,
+        )
+        .await
+        .attach_printable_lazy(|| format!("event not found for id: {}", &event_id))
+        .map_err(|e| {
+            logger::error!("{:?}", e);
+            e
+        })
```
SanchithHegde, line 490:
```suggestion
        .map_err(|error| {
            logger::error!(?error);
            error
        })
```

We use the `tracing` crate for logging, so we can use the `?` to use the `Debug` impl for logging. Refer to the [`tracing` docs](https://docs.rs/tracing/0.1.40/tracing/#recording-fields) for more information.

crates/router/src/core/webhooks/outgoing.rs
```diff
@@ -457,13 +473,50 @@ fn raise_webhooks_analytics_event(
...
+    let status_code = if let Some(updated_event) = updated_event {
+        let webhook_response: Option<OutgoingWebhookResponseContent> = updated_event
+            .response
+            .map(|res| res.peek().to_owned())
+            .and_then(|res| {
+                res.parse_struct("OutgoingWebhookResponseContent")
+                    .map_err(|e| {
+                        logger::error!("Error deserializing webhook response: {:?}", e);
+                        e
+                    })
```
SanchithHegde, line 503:
```suggestion
                    .map_err(|error| {
                        logger::error!(?error, "Error deserializing webhook response");
                        error
                    })
```

This is so that `error` would be a separate field in the recorded log message / event.

crates/router/src/core/webhooks/outgoing.rs
```diff
@@ -457,13 +473,50 @@ fn raise_webhooks_analytics_event(
...
+        .await
+        .attach_printable_lazy(|| format!("event not found for id: {}", &event_id))
+        .map_err(|e| {
+            logger::error!("{:?}", e);
+            e
+        })
+        .ok();
+
+    // Get status_code from webhook response
+    let status_code = if let Some(updated_event) = updated_event {
```
SanchithHegde, line 494: Since we've used `and_then()` elsewhere in this function, we can use it here as well? Something like:
`let status_code = updated_event.and_then(|...| ...);`

SanchithHegde (commented):

crates/router/src/core/webhooks/outgoing.rs
Loaki07: Nice! Got it!

crates/router/src/core/webhooks/outgoing.rs
Loaki07: This is much better.

SanchithHegde (approved):

lsampras (dismissed):

SanchithHegde:
@Loaki07 Can you please address the failing CI checks (those running on `ubuntu-latest`)? You can allow the `clippy::too_many_arguments` lint in this case.

SanchithHegde (approved):

lsampras (approved):

sai-harsha-vardhan (approved):

PR #5471, opened 2024-07-29
refactor(id_type): use macros for defining ID types and implementing common traits

SanchithHegde:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [x] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

This PR introduces declarative macros to define ID types and implement common traits such as `Debug`, `FromSql`, `ToSql`, etc. on these ID types. In addition, this PR replaces the usages of the `from()` associated function (to construct the ID types) with the `TryFrom` trait implementation.

## Motivation and Context

Helps reduce duplication of boilerplate code.

## How did you test it?

Sanity testing using Postman collection, code _should_ work if it compiles.

## Checklist


crates/router/src/core/admin.rs
```diff
@@ -2016,7 +2016,7 @@ impl MerchantConnectorAccountCreateBridge for api::MerchantConnectorCreate {
                     connector_webhook_details.encode_to_value(
                     )
                     .change_context(errors::ApiErrorResponse::InternalServerError)
-                    .attach_printable(format!("Failed to serialize api_models::admin::MerchantConnectorWebhookDetails for Merchant: {}", business_profile.merchant_id))
+                    .attach_printable(format!("Failed to serialize api_models::admin::MerchantConnectorWebhookDetails for Merchant: {}", business_profile.merchant_id.get_string_repr()))
```
Narayanbhat166, line 2019: We should not have any use of `get_string_repr` functions ( this will be removed later ), So we can use the debug impl here and in other places as well where we just want to display this information

Narayanbhat166 (approved):
Looks Good To Me

crates/router/src/core/admin.rs
SanchithHegde: Fair enough, I've updated the PR to use the `Debug` impl in log messages, while keeping the other usages like cache keys or merchant-facing error messages as is. Or let me know if you would prefer me adding back the `Display` implementation for the other usages.

Narayanbhat166 (dismissed):

srujanchikke (dismissed):
reviewed connector part and LGTM.

Aprabhat19 (approved):

ThisIsMani (dismissed):
Dashboard specific changes looks fine.

lsampras (approved):

jarnura (approved):

jarnura (approved):

Narayanbhat166 (approved):

ThisIsMani (approved):

Aprabhat19 (approved):

PR #5481, opened 2024-07-30
feat(cypress): add corner cases

pixincreate:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [x] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [x] CI/CD

## Description

This PR adds variation cases where Hyperswitch is known to get `4xx` response. This is added as a part of regression test.

Tests include, but not limited to:

- Invalid Card info
  - Invalid card number length
  - Invalid card expiry month
  - Invalid card expiry year
  - Invalid Card CVV
- Confirm a payment without PMD
- Capture greater amount
- Capture successful payment
- Void successful payment

To accommodate the changes for above, slight refactoring has been done to the configs where I've removed `payment_method` from `confirm_body` in `fixtures` and moved them to individual connector level configs.

This also adds a check to `4xx` error body where we validate whether the error response is an `object` or a `string` (required for invalid card length).

This also forced me to fix issues with data trans connector.

This enhancement will be done in multiple PRs as I do not want to make it complicated unnecessarily.

### Additional Changes

## Motivation and Context

NIL

## How did you test it?

Corner cases
image

|Adyen|Bank of America|Bluesnap|Cybersource|Iatapay|
|-|-|-|-|-|
|image|image|image|image|image|
|NMI|PayPal|Stripe|Trustpay||
|image|image|image|image||

There happens to be an issue with NMI:

image

```
expected 'amount_to_capture is greater than amount' to equal 'This Payment could not be captured because it has a capture_method of manual. The expected state is manual_multiple'
```

It should expect `capture_method` to be `manual` and should not be limited to `manual_multiple`

## Checklist


cypress-tests/cypress/e2e/PaymentTest/00020-Variations.cy.js
```diff
@@ -0,0 +1,275 @@
...
+        globalState = new State(state);
+      });
+    });
+
+    after("flush global state", () => {
+      cy.task("setGlobalState", globalState.data);
+    });
+
+    it("[Payment create] Invalid card number", () => {
+      let data = getConnectorDetails(globalState.get("commons"))["card_pm"][
```
pixincreate, line 30: Using `commons` in here is intentional since we expect these to run against regardless of connector

cypress-tests/cypress/fixtures/confirm-body.json
```diff
@@ -2,8 +2,6 @@
   "client_secret": "",
   "return_url": "https://hyperswitch.io",
   "confirm": true,
-  "payment_method": "card",
```
Sakilmostak, line 5: What is the reason for removing payment method from fixtures and adding individually to each configs? Otherwise LGTM

Sakilmostak (changes requested):

cypress-tests/cypress/fixtures/confirm-body.json
pixincreate: For corner cases, especially for `confirm payment w/o pmd`, we're not supposed to pass any `payment_method` info. With `payment_method` being existent in `confirm_body` fixture, it is being supplied in the request resulting in failures.  
To over come this, I had to remove it from fixtures which resulted in other tests to fail forcing me to add at config level.

likhinbopanna (approved):

Gnanasundari24 (approved):

Issue #5487, opened 2024-07-31
[refactor]: add level support for roles


apoorvdixit88:
Determine a given role exists at which level, i.e., determine a entity of each roles which it is associated to.
- This will help users to perform operation accordingly.
- A profile/merchant level user cannot perform org level opertaion.
- Similarly a hierarchy can be maintained where a user (which will be associated to a particular role) can not perform any operations which are for higher entities.

PR #5488, opened 2024-07-31
refactor(role): determine level of role entity

apoorvdixit88:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [x] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description
Add entity type to roles (predefined and customs)
This can be used to determine at which level a current entity exist, and can perform operations accordingly.

### Additional Changes

## Motivation and Context
Closes [#5487](https://github.com/juspay/hyperswitch/issues/5487)

## How did you test it?
The change is at DB level and can be tested with the upcoming application changes PRs.

## Checklist


migrations/2024-07-30-124102_add_entity_type_to_roles/up.sql
```diff
@@ -0,0 +1,2 @@
+-- Your SQL goes here
+ALTER TABLE roles ADD COLUMN entity_type VARCHAR(64);
```
ThisIsMani, line 2: It need not to be nullable, we can put default as merchant.

ThisIsMani:
Schema consistency is failing for `v2`. Run `just migrate_v2 run` and commit the new schema file.

migrations/2024-07-30-124102_add_entity_type_to_roles/up.sql
ThisIsMani: No, we don't have to remove the default at all.

When we insert the roles we are anyways always inserting the entity_type of the role.

Putting it default will remove the burden of running the backfill queries. Like we did for version.

ThisIsMani (approved):
I think it's better to add comments wherever there will be changes when we remove V1.

racnan (approved):

jarnura (commented):
We can make the DB layer also enum instead of varchar.

jarnura (approved):

apoorvdixit88:
> We can make the DB layer also enum instead of varchar.

In `user_roles` table also we have kept it varchar . We can cover both in separate PR.

PR #5492, opened 2024-07-31
feat(auth): Add `profile_id` in `AuthenticationData`

### LINKED ISSUE #5491: feat(auth): Support profile level authentication
- After the dashboard is changed to profile level, there will be 3 types of APIs based on the level
	1. Merchant Level only
		- API Keys
		- Merchant Account
		- Business profile
	2. Profile Level only
		- Connectors
		- Routing
		- Operations (Payment, refund, disputes except list and filter)
		- etc...
	1. Profile + Merchant
		- Operations list and filter
		- etc...

### Merchant Level APIs
- No changes here.

### Profile Level only
- These should start accepting an optional profile_id in the function parameter, which will be passed from auth layer.
- There are some APIs which are currently at pseudo profile level like payments create and refunds create.
	- Currently these kind APIs take profile_id in the payload or query param.
	- From now on, these should not accept these in the request and take those values from the auth layer.
- To support the current API calls, these will fallback to body if profile_id in header is not present.

### Profile + Merchant Level
- These APIs should operate both in merchant level and profile level.
- These APIs accept a vec of profile_ids in the function parameter.
- There should be different routes for these APIs and probably the core function will remain same.
	- The route function will be different for each route but calls the same core function.
	- If the profile_id vec is not empty, then the API will send the data for those specific profile_ids else for the whole merchant.

## Breaking changes
- Current APIs will be changed in a way that the current calls will work without any changes. So there shouldn't be any breaking changes.

## Auth
- Auth layer will now give either
	1. org_id, merchant_id and profile_id
	2. org_id, merchant_id and Vec<profile_id>
- There are currently two types of JWT Auths.
	1. JWTAuth
	2. JWTMerchantFromRoute
- Both of these will have a new implementation which will send the data as mentioned in the first point.
- Both of these have profile_id as optional type for backwards compatibility.

- Once we have a stable version where profile_id is always present in the JWT, then we can have following JWT Auths for more checks.
	1. JWTAuth
	2. JWTMerchantFromRoute
	3. JWTProfileFromRoute
	4. JWTMerchantAndProfileFromRoute
- These will always have profile_id in the payload.

- Routes can choose to pass or not to pass the profile data to the core function.

ThisIsMani:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

Since we are going to change the dashboard to profile level, we need profile level authentication. 
This PR introduces profile_id in the AuthenticationData. Core API will take this into consideration before sending any response.

### Additional Changes

## Motivation and Context

Closes #5491.

## How did you test it?

All the APIs which are using JWT, should not fail in the authentication layer.

## Checklist


crates/router/src/services/authentication.rs
```diff
@@ -46,6 +46,14 @@ pub mod decision;
 pub struct AuthenticationData {
     pub merchant_account: domain::MerchantAccount,
     pub key_store: domain::MerchantKeyStore,
+    pub profile_id: Option<String>,
+}
+
+#[derive(Clone)]
+pub struct AuthenticationDataWithMultipleProfiles {
+    pub org_id: id_type::OrganizationId,
+    pub merchant_id: id_type::MerchantId,
```
hrithikesh026, line 55: This should be merchant_account.
```suggestion
    pub merchant_account: domain::MerchantAccount,
```

crates/router/src/services/authentication.rs
```diff
@@ -852,6 +903,7 @@ where
         let auth = AuthenticationData {
             merchant_account: merchant,
             key_store,
+            profile_id: payload.profile_id,
```
hrithikesh026, line 906: Are we not supposed to validate this profile_id?

crates/router/src/services/authentication.rs
```diff
@@ -684,6 +698,43 @@ where
...
+        authorization::check_authorization(&self.0, &permissions)?;
+
+        Ok((
+            AuthenticationDataWithMultipleProfiles {
+                org_id: payload.org_id,
+                merchant_id: payload.merchant_id.clone(),
+                profile_id: payload
+                    .profile_id
+                    .map(|profile_id| vec![profile_id])
+                    .unwrap_or(vec![]),
```
hrithikesh026, line 737: Are we not supposed to validate this merchant_id and profile_id list?

crates/router/src/services/authentication.rs
```diff
@@ -852,6 +903,7 @@ where
         let auth = AuthenticationData {
             merchant_account: merchant,
             key_store,
+            profile_id: payload.profile_id,
         };
         Ok((
             auth.clone(),
```
hrithikesh026, line 909: I don't think .clone() is required here.

crates/router/src/services/authentication.rs
jarnura: what kind of validation?

crates/router/src/services/authentication.rs
jarnura: what kind of validation?

jarnura (changes requested):

crates/router/src/services/authentication.rs
hrithikesh026: Like check if the MerchantAccount is present in DB, And fetch and return the record.
And also check if all profiles belong to the given merchant.

crates/router/src/services/authentication.rs
hrithikesh026: check if the profile belongs to the given merchant.

crates/router/src/services/authentication.rs
ThisIsMani: No, we don't have to.

Since this JWT is issued and signed by us we can always say that this JWT will have valid profile_ids.

We can do it for APIKey auth where the profile_ids are sent by merchant themselves.

crates/router/src/services/authentication.rs
ThisIsMani: We are using auth again in below lines.

crates/router/src/services/authentication.rs
```diff
@@ -46,6 +46,14 @@ pub mod decision;
 pub struct AuthenticationData {
     pub merchant_account: domain::MerchantAccount,
     pub key_store: domain::MerchantKeyStore,
+    pub profile_id: Option<String>,
+}
+
+#[derive(Clone)]
+pub struct AuthenticationDataWithMultipleProfiles {
+    pub merchant_id: domain::MerchantAccount,
```
hrithikesh026, line 54:
```suggestion
    pub merchant_account: domain::MerchantAccount,
```

crates/router/src/services/authentication.rs
```diff
@@ -46,6 +46,14 @@ pub mod decision;
...
+    pub key_store: domain::MerchantKeyStore,
+    pub profile_id: Vec<String>,
```
hrithikesh026, line 56:
```suggestion
    pub profile_ids: Vec<String>,
```

hrithikesh026: Or `pub profile_id_list: Vec<String>,`

crates/router/src/services/authentication.rs
```diff
@@ -46,6 +46,14 @@ pub mod decision;
...
     pub merchant_account: domain::MerchantAccount,
     pub key_store: domain::MerchantKeyStore,
+    pub profile_id: Option<String>,
+}
+
+#[derive(Clone)]
+pub struct AuthenticationDataWithMultipleProfiles {
+    pub merchant_account: domain::MerchantAccount,
+    pub key_store: domain::MerchantKeyStore,
+    pub profile_id_list: Vec<String>,
```
apoorvdixit88, line 56: we can make this option?

apoorvdixit88 (dismissed):

hrithikesh026 (approved):

crates/router/src/services/authentication.rs
```diff
@@ -684,6 +698,61 @@ where
...
+                &key_store,
+            )
+            .await
+            .change_context(errors::ApiErrorResponse::InvalidJwtToken)?;
+
+        Ok((
+            AuthenticationDataWithMultipleProfiles {
+                key_store,
+                merchant_account: merchant,
+                profile_id_list: payload.profile_id.map(|profile_id| vec![profile_id]),
```
jarnura, line 755:
```suggestion
                profile_id_list: None,
```

apoorvdixit88 (approved):

hrithikesh026 (approved):

jarnura (approved):

PR #5498, opened 2024-07-31
feat(core): accept business profile in core functions for payments, refund, payout and disputes

hrithikesh026:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

Currently Auth layer provides core with only merchant_account after successful authentication(merchant level dashboard).

When control center moves to profile level, we would need to accept profiles id in core functions. Made that change for a few core functions that are currently being used by control center(dashboard).

### Additional Changes

## How did you test it?

This feature can only be tested once profile_id based jwt token changes are done.
For now general sanity test would suffice.

## Checklist


crates/router/src/core/payments.rs
```diff
@@ -788,6 +788,7 @@ pub async fn payments_core<F, Res, Req, Op, FData>(
     state: SessionState,
     req_state: ReqState,
     merchant_account: domain::MerchantAccount,
+    _profile_id_list: Option<Vec<String>>,
```
ThisIsMani, line 791: Why is this `Option<Vec<String>>`?

hrithikesh026: This would give a bit more flexibility in future.

Some core functions are being called from other flows like webhooks, compatibility layer etc. So in these parts it wouldn't make sense to pass hard-coded empty vecs.

ThisIsMani (dismissed):

jarnura (approved):

racnan (approved):

PR #5513, opened 2024-08-02
feat(cypress): add mandates and refunds variation cases

pixincreate:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [x] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [x] CI/CD

## Description

This PR:
- Fixes inconsistencies with respect to naming in configs (bank redirects) -- Make all object names follow `PascalCase` naming convention
- Add mandate validation
- Variation tests:
  - Refunds exceed captured amount
  - Recurring mandate with greater mandate amount
  - Refund unsuccessful amount
  - 3DS with greater capture
  - Void successful payment
- Add basic structural support for `responseCustom`[this provides us the ability to define multiple Responses (limited to `2`) for a single `request`]
  - This opens up possibility for us to handle multiple responses that we receive from connector's end
  - For example, in Adyen when `capture_method` is `automatic`, payment `status` goes to `succeeded` state and when `capture_method` is `manual`, payment `status` goes to `processing` state

### Additional Changes

## Motivation and Context

Why?

## How did you test it?

Tested against `Stripe`:

image

Iatapay, which is more error prone, just works:

image

## Checklist


cypress-tests/cypress/e2e/PaymentUtils/Commons.js
```diff
@@ -959,6 +983,134 @@ export const connectorDetails = {
...
+        payment_method: "card",
+        payment_method_type: "debit",
+        setup_future_usage: "on_session",
+        payment_method_data: {
+          card: {
+            card_number: "4242424242424242",
+            card_exp_month: "01",
+            card_exp_year: "2023",
+            card_holder_name: "joseph Doe",
+            card_cvc: "123456",
```
likhinbopanna, line 998: Why can't we have one variation test case at a time , i can see you are using invalid currency and invalid cvv in one test case , can you clarify this ?

pixincreate: The thing is, errors are thrown based on the priority or the way code is written.  
CVC is checked before expiry year is checked. So, in this case, it does not matter if we use an already expired year.  

Hope that clears things out.

likhinbopanna (approved):

Gnanasundari24 (approved):

Gnanasundari24:
@pixincreate Can you add the tescase result for adyen and paypal as you have introduced custom response

pixincreate:
> @pixincreate Can you add the tescase result for adyen and paypal as you have introduced custom response

As you can see in the changes introduced to Adyen [configs](https://github.com/juspay/hyperswitch/pull/5513/files#diff-1beaad14cecb3e9288cbfffd5c0da99b02d8b23809602160760255a80ba7805c) and [commons](https://github.com/juspay/hyperswitch/pull/5513/files#diff-58576924c69cf304d6bd0918b51bb078f06916c0e8dd2ac5a65c8ec5348eefba) in this PR, I've only added the basic support for `responseCustom` and have **not** added any tests or fixes for respective connectors mentioned. That can be taken up as a separate PR.

PR #5520, opened 2024-08-05
refactor(auth): Pass `profile_id` from the auth to core functions

### LINKED ISSUE #5491: feat(auth): Support profile level authentication

ThisIsMani:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [x] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

Use the profile_id from the authentication data and pass it to core functions.

### Additional Changes

## Motivation and Context

Closes #5491

## How did you test it?

Using local dashboard.
All the payment, disputes, payouts and refund APIs should work without any authentication error.

## Checklist


hrithikesh026 (dismissed):

apoorvdixit88 (dismissed):

crates/router/src/services/authentication.rs
```diff
@@ -968,6 +974,64 @@ where
...
+        let auth = AuthenticationData {
+            merchant_account: merchant,
+            key_store,
+            profile_id: payload.profile_id,
+        };
+        Ok((
+            auth.clone(),
+            AuthenticationType::MerchantJwt {
+                merchant_id: auth.merchant_account.get_id().clone(),
+                user_id: None,
```
racnan, line 1029: shouldnt be `None`

crates/router/src/services/authentication.rs
```diff
@@ -968,6 +974,64 @@ where
...
+        let key_manager_state = &(&state.session_state()).into();
+        let key_store = state
+            .store()
+            .get_merchant_key_store_by_merchant_id(
+                key_manager_state,
+                &payload.merchant_id,
+                &state.store().get_master_key().to_vec().into(),
+            )
+            .await
+            .change_context(errors::ApiErrorResponse::InvalidJwtToken)
```
racnan, line 1007: `500` is not handled.

racnan (approved):

jarnura (approved):

PR #5183, opened 2024-07-03
feat(ci): add cypress tests to github ci

pixincreate:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [x] CI/CD

## Description

This PR makes Cypress tests to run against PRs.
At present, it runs against Stripe.

|Jobs|Time taken to build router|Time taken to run Cypress|Overall time consumption|
|-|-|-|-|
|4|18m 02s|07m 18s|26m 53s|
|6|16m 08s|07m 12s|26m 21s|
|6|16m 11s|07m 26s|26m 16s|
|8|15m 47s|07m 13s|24m 44s|
|8|24m 36s|07m 09s|36m 10s|

6 core should be fine for now. Will be monitoring this over a week.

### Additional Changes

- [ ] This PR modifies the API contract
- [ ] This PR modifies the database schema
- [ ] This PR modifies application configuration/environment variables

## Motivation and Context

NIL

## How did you test it?

CI should run against PRs successfully: https://github.com/juspay/hyperswitch/actions/runs/11089890539/job/30811578393?pr=5183

## Checklist

- [ ] I formatted the code `cargo +nightly fmt --all`
- [ ] I addressed lints thrown by `cargo clippy`
- [x] I reviewed the submitted code
- [ ] I added unit tests for my changes where possible

PR #5524, opened 2024-08-05
feat(cypress): add configs for not overriding screenshots

pixincreate:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [x] CI/CD

## Description

Currently, we do not have the privilege for properly taking screenshots when a test fails. We have to manually run the test and see what went wrong, why it went wrong. That is because upcoming test would override previous / existing screenshot. This is a loss for us.

- screenshots that are taken will **not** be overwritten and will **not** be trashed between runs
- To mask sensitive info, elements with the class .secret-info or any element with the attribute data-hide=true will be blacked out to prevent sensitive information from being visible in the screenshot
- Cypress will take screenshot of entire `runner` window and not just specific `viewport`
- Hooks:
  - `onBeforeScreenshot`:
    - Runs before taking the screenshot
    - It looks for elements with the class `.clock` within the `$el` (the element being captured)
    - If such elements are found, they are hidden to ensure they don't appear in the screenshot
    - Useful to exclude dynamic or changing elements from screenshots
  - `onAfterScreenshot`:
    - Runs after taking the screenshot
    - It finds elements with the class `.clock` that were hidden before the screenshot and makes them visible again
    - This ensures that the visibility of these elements is restored for subsequent tests or interactions

This change is in fact, taken from https://github.com/juspay/hyperswitch/pull/5183

### Additional Changes

## Motivation and Context

Take screenshots.

## How did you test it?

it does takes screenshots with this change.

image
image

## Checklist

### DIFF
--- a/cypress-tests/cypress/support/screenshotConfigs.js
+++ b/cypress-tests/cypress/support/screenshotConfigs.js
@@ -1,0 +1,21 @@
+Cypress.Screenshot.defaults({
+  blackout: [".secret-info", "[data-hide=true]"],
+  capture: "runner",
+  overwrite: false,
+
+  onBeforeScreenshot($el) {
+    const $clock = $el.find(".clock");
+
+    if ($clock) {
+      $clock.hide();
+    }
+  },
+
+  onAfterScreenshot($el, props) {
+    const $clock = $el.find(".clock");
+
+    if ($clock) {
+      $clock.show();
+    }
+  },
+});
--- a/cypress-tests/cypress.config.js
+++ b/cypress-tests/cypress.config.js
@@ -21,4 +21,5 @@ module.exports = {
   chromeWebSecurity: false,
   defaultCommandTimeout: 10000,
   pageLoadTimeout: 20000,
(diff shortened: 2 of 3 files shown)


Sakilmostak (approved):

Gnanasundari24 (approved):

cypress-tests/cypress/support/screenshotConfigs.js
```diff
@@ -0,0 +1,21 @@
+Cypress.Screenshot.defaults({
+  blackout: [".secret-info", "[data-hide=true]"],
```
pixincreate, line 2: `blackout` field can be expanded to include `api_keys` and other stuffs (like sensitive info from connectors(?)) from getting exposed in responses.  
It is not a concern at this point in time as we're no where exposing the field(s). But yes, there is a room for improvement in this region.

CC: @Sakilmostak @SanchithHegde

PR #5527, opened 2024-08-05
refactor(merchant_account_v2): remove routing algorithms from merchant account and add version column

Narayanbhat166:
## Type of Change

- [x] Refactoring

## Description

Remove the routing algorithm from merchant_account for api v2. This PR also adds version column to the merchant account.

### Additional Changes

Remove the `routing_algorithm` field from merchant_account. This is for v2 version only.
For v1 version of merchant account, version column has been added.

## How did you test it?

Test the sanity of merchant account

## Impact
Merchant account

## Checklist


SanchithHegde (approved):

crates/hyperswitch_domain_models/src/merchant_account.rs
```diff
@@ -135,7 +135,6 @@ pub struct MerchantAccountSetter {
     pub publishable_key: String,
     pub storage_scheme: MerchantStorageScheme,
     pub metadata: Option<pii::SecretSerdeValue>,
-    pub routing_algorithm: Option<serde_json::Value>,
     pub frm_routing_algorithm: Option<serde_json::Value>,
```
jarnura, line 138: remove these this algorithm also

jarnura (changes requested):

jarnura (dismissed):

SanchithHegde (dismissed):

crates/hyperswitch_domain_models/src/customer.rs
```diff
@@ -36,7 +35,6 @@ pub struct Customer {
     pub address_id: Option<String>,
     pub default_payment_method_id: Option<String>,
     pub updated_by: Option<String>,
-    pub version: ApiVersion,
```
sahkal, line 39: why are these deleted, please add them back

Narayanbhat166: The idea is to have a single place to set this value, and this should be set at the database layer only. But since we might need this in logs in the future, will add this back

SanchithHegde (approved):

jarnura (approved):

SanchithHegde (approved):

Aprabhat19 (approved):

Issue #5540, opened 2024-08-06
feat: support profile level delete in V1 - V2


apoorvdixit88:
Add support to delete user role 
- Support profile level delete along with merchant level delete.
- In v1v2 version, delete from both V1 and V2

PR #5541, opened 2024-08-06
feat(users): add support for profile user delete

apoorvdixit88:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [x] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description
- Add support to delete profile level users
- Delete is backward compatible
- Delete to support both V1 and V2 operations accordingly

### Additional Changes

## Motivation and Context
Closes [#5540](https://github.com/juspay/hyperswitch/issues/5540)

## How did you test it?
Use the curl to delete user role
```
curl --location --request DELETE 'http://localhost:8080/user/user/delete' \
--header 'Authorization: Bearer JWT' \
--data-raw '{
  "email": "email_to_delete"
}'
```

## Checklist


ThisIsMani (commented):
This might break when org level roles are updatable or deletable.

apoorvdixit88:
> This might break when org level roles are updatable or deletable.

For V1 org_level roles are neither updatable nor deletable. With V2 it will work fine. Also we won't be introducing org_level roles in V1.

crates/router/src/core/user_role.rs
```diff
@@ -342,68 +342,151 @@ pub async fn delete_user_role(
...
             .await
             .change_context(UserErrors::InternalServerError)
-            .attach_printable("Error while deleting user role")?
-    } else {
+            .attach_printable("Error while deleting user role")?;
+    }
+
+    if !user_role_deleted_flag {
+        return Err(report!(UserErrors::InvalidDeleteOperation))
+            .attach_printable("User is not associated with the merchant");
```
racnan, line 470: `500` is not handled.

racnan (changes requested):

crates/router/src/utils/user_role.rs
```diff
@@ -172,3 +172,19 @@ pub async fn get_multiple_role_info_for_user_roles(
...
 }
+
+pub fn entity_level(entity_type: EntityType) -> u8 {
+    match entity_type {
+        EntityType::Internal => 3,
+        EntityType::Organization => 2,
+        EntityType::Merchant => 1,
+        EntityType::Profile => 0,
+    }
+}
```
racnan, line 183: https://doc.rust-lang.org/std/cmp/trait.Ord.html

crates/router/src/core/user_role.rs
```diff
@@ -342,68 +342,163 @@ pub async fn delete_user_role(
...
+
+    // Check if user has any more role associations
+    let user_roles = state
+        .store
+        .list_user_roles_by_user_id(user_from_db.get_user_id(), UserRoleVersion::V1)
+        .await
+        .change_context(UserErrors::InternalServerError)?;
+
+    // If user has no more role associated with him then deleting user
+    if user_roles.is_empty() {
```
ThisIsMani, line 493: You can check if v2 roles are also empty.

ThisIsMani (commented):
Please remove the old delete function.

ThisIsMani (approved):

crates/router/src/db/kafka_store.rs
```diff
@@ -2645,25 +2645,50 @@ impl UserRoleInterface for KafkaStore {
...
+        org_id: &id_type::OrganizationId,
+        merchant_id: &id_type::MerchantId,
+        profile_id: Option<&String>,
         version: enums::UserRoleVersion,
-    ) -> CustomResult<Vec<user_storage::UserRole>, errors::StorageError> {
-        self.diesel_store
-            .list_user_roles_by_user_id(user_id, version)
-            .await
+    ) -> CustomResult<storage::UserRole, errors::StorageError> {
+        self.find_user_role_by_user_id_and_lineage(
```
jarnura, line 2666:
```suggestion
        self.diesel_store.find_user_role_by_user_id_and_lineage(
```

crates/router/src/db/kafka_store.rs
```diff
@@ -2645,25 +2645,50 @@ impl UserRoleInterface for KafkaStore {
...
+
+    async fn delete_user_role_by_user_id_and_lineage(
+        &self,
+        user_id: &str,
+        org_id: &id_type::OrganizationId,
+        merchant_id: &id_type::MerchantId,
+        profile_id: Option<&String>,
+        version: enums::UserRoleVersion,
+    ) -> CustomResult<storage::UserRole, errors::StorageError> {
+        self.delete_user_role_by_user_id_and_lineage(
```
jarnura, line 2684:
```suggestion
        self.diesel_store.delete_user_role_by_user_id_and_lineage(
```

crates/router/src/db/kafka_store.rs
jarnura: In kakfaStore if we can self.functions will causes stack_overflow. So avoid following that.

crates/router/src/db/kafka_store.rs
jarnura: In kakfaStore if we can self.functions will causes stack_overflow. So avoid following that.

jarnura (changes requested):

jarnura (approved):

PR #5553, opened 2024-08-07
feat(core): use profile_id passed from auth layer within core functions

hrithikesh026:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

Currently few core functions accept either a list of profile_ids or a single profile_id from auth layer. 

But the core functions were not using the profile_ids for validation or filtering.

In this PR, I have done the below changes.
1. If the core function returns a single object as response, we would validate if the `profile_id` passed by auth layer is same as the `profile_id` present in the object.
2. If the core function returns a list of objects, we would filter the list such that only objects that has `profile_id` which is present in the `profile_id_list` provided by the auth layer is returned.

### Additional Changes

## How did you test it?

Cannot be tested right now because currently we are not including profile_id in JWT
Made sure existing jwt token auth works fine.

## Checklist


crates/router/src/core/utils.rs
```diff
@@ -1254,3 +1315,64 @@ pub fn get_incremental_authorization_allowed_value(
...
+    profile_id_auth_layer: Option<&String>,
+    profile_id: Option<&String>,
+) -> RouterResult<()> {
+    if profile_id_auth_layer != profile_id {
+        return Err(errors::ApiErrorResponse::PreconditionFailed {
+            message: "Profile id authentication failed. Please use the correct JWT token"
+                .to_string(),
+        }
+        .into());
+    }
```
ThisIsMani, line 1376: Won't this fail if we hit the API with APIKey?

ThisIsMani: We should only check this if auth profile id is present.

crates/router/src/core/utils.rs
```diff
@@ -1254,3 +1315,79 @@ pub fn get_incremental_authorization_allowed_value(
...
+        object_list
+    }
+}
+
+pub(super) fn validate_profile_id_from_auth_layer<T: GetProfileId>(
+    profile_id_auth_layer: Option<&String>,
+    object: &T,
+) -> RouterResult<()> {
+    if profile_id_auth_layer
+        .is_some_and(|auth_profile_id| Some(auth_profile_id) == object.get_profile_id())
```
ThisIsMani, line 1384: This will still break, if we don't find the profile_id in `T`. What are the cases where we will get profile_id as null?

If the profile_id is not supposed to be null, isn't it data integrity issue. And shouldn't we throw 500?

ThisIsMani: If we can do this, then we make the trait function return the profile id directly instead of option.

jarnura (approved):

crates/router/src/routes/payments.rs
```diff
@@ -1015,7 +1015,7 @@ pub async fn payments_list_by_filter(
             payments::apply_filters_on_payments(
                 state,
                 auth.merchant_account,
-                None,
+                auth.profile_id.map(|profile_id| vec![profile_id]),
```
ThisIsMani, line 1018: Shouldn't this be None. This is a list API.

crates/router/src/core/utils.rs
```diff
@@ -1254,3 +1315,97 @@ pub fn get_incremental_authorization_allowed_value(
...
+        object_list
+    }
+}
+
+pub(super) fn validate_profile_id_from_auth_layer<T: GetProfileId>(
+    profile_id_auth_layer: Option<String>,
+    object: &T,
+) -> RouterResult<()> {
+    if profile_id_auth_layer
+        .is_some_and(|auth_profile_id| Some(&auth_profile_id) == object.get_profile_id())
```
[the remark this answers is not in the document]
ThisIsMani: Something like no profile_id found in object

ThisIsMani (approved):

jarnura (approved):

PR #4159, opened 2024-03-20
feat(events): allow listing webhook events and webhook delivery attempts by business profile

SanchithHegde:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

This PR extends the events list APIs added in #4131 to allow specifying either merchant ID or the business profile ID in the URLs. This PR also includes the corresponding OpenAPI specification changes.

The URLs remain the same as described in #4131, with the only difference being that `{merchant_id}` can now be substituted with `{merchant_or_profile_id}`.

## Motivation and Context

This would allow users to list and work with events associated with a specific business profile, which could be useful when users would want to exclude events from other business profiles associated with the same merchant account.

## How did you test it?

1. The three functionalities provided by the endpoints must be unaffected when querying by merchant ID:

   - List events related to a specific object:

     ```shell
     curl -H 'api-key: test_admin' 'http://localhost:8080/events/merchant_1710609960?object_id=pay_9uclJXQboJ0FZPm9Nq2K' 2>/dev/null | jq '.'
     ```

   - Filter events:

     ```shell
     curl -H 'api-key: test_admin' 'http://localhost:8080/events/merchant_1710609960?created_after=2024-03-18T09:00Z&created_before=2024-03-18T11:00Z&limit=2&offset=2' 2>/dev/null | jq '.'
     ```

   - List delivery attempts:

     ```shell
     $ curl -H 'api-key: test_admin' 'http://localhost:8080/events/merchant_1710609960/evt_018e562fcd4571b0b831098bb9ebe230/attempts' 2>/dev/null | jq '.'
     ```

2. The above three functionalities must be unaffected if the merchant ID in the URLs is replaced by the profile ID:

   - List events related to a specific object:

     ```shell
     $ curl -H 'api-key: test_admin' 'http://localhost:8080/events/pro_H7FbvX2pBuJLjlPhkpRa?object_id=pay_9uclJXQboJ0FZPm9Nq2K' 2>/dev/null | jq '.'
     [
       {
         "event_id": "evt_018e484cb6207265a1c715568f889590",
         "merchant_id": "merchant_1710609960",
         "profile_id": "pro_H7FbvX2pBuJLjlPhkpRa",
         "object_id": "pay_9uclJXQboJ0FZPm9Nq2K",
         "event_type": "payment_succeeded",
         "event_class": "payments",
[...]

PR #5566, opened 2024-08-07
refactor: use business profile domain models instead of diesel models

SanchithHegde:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [x] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

This PR contains the following changes:

- Updates the `BusinessProfileInterface` trait methods to use and return business profile domain models instead of the corresponding diesel models. As a result of this, almost all of the code in the `router` crate interacting with business profiles had to be updated to use the domain models, and the key manager state and key store had to be passed around.

- :warning: Due to the above change, we can no longer fetch the business profile with the profile ID alone, we would need the merchant ID additionally (to obtain the merchant key store, which is then used to decrypt the business profile). So, I've temporarily removed the ability to list outgoing webhook events by profile ID, by commenting out the relevant code.
  - This effectively reverts changes introduced in #4159, but since the control center hasn't yet integrated these APIs, I'll address this in a separate PR.

- Addresses compilation failures which arose when the following command was run:

  ```shell
  cargo clippy --package router --features v2,customer_v2,merchant_account_v2,payment_methods_v2,routing_v2,merchant_connector_account_v2
  ```

  **I need help from the reviewers in verifying that these changes are indeed correct.**

- Addresses a Rust 1.80 `clippy` lint by removing an unused type in Braintree connector code. (This would cause `clippy` checks to fail on CI starting August 8th, otherwise.)

## Motivation and Context

Making use of domain models helps ensure that the encrypted data stored in the database is available in plaintext for the developers working with those fields, and helps avoid possible repetitive decryption of the same value.
Additionally, this change should help unblock more changes around the API v2 refactoring activity.

## How did you test it?

1. Running `clippy` and ensuring that it succeeds.

2. Running `cargo clippy -p router --features v2,customer_v2,merchant_account_v2,payment_methods_v2,routing_v2,merchant_connector_account_v2` and ensuring that the command succeeds.

3. Ensuring that the `outgoing_webhook_custom_http_headers` field in business profile is stored and retrieved correctly, as this is the only field involving encryption in the business profile domain model.
   - This was done by providing custom headers in `outgoing_webhook_custom_http_headers` field in the business profile create API, and confirming that the data is returned correctly when the business profile is retrieved.

5. Ensuring that the outgoing webhook custom headers are correctly included in outgoing webhook requests.

   I'd provided the following value for the custom headers:

   ```json
   "outgoing_webhook_custom_http_headers": {
     "foo": "bar",
     "bar": "baz"
   }
   ```

   Screenshot of custom headers included in outgoing webhook request:
   Screenshot of outgoing webhook request headers

## Checklist


jarnura (approved):

Narayanbhat166 (approved):

crates/router/src/routes/webhook_events.rs
```diff
@@ -15,15 +15,15 @@ use crate::{
...
+    path: web::Path<common_utils::id_type::MerchantId>,
     query: web::Query<EventListConstraints>,
 ) -> impl Responder {
     let flow = Flow::WebhookEventInitialDeliveryAttemptList;
     let merchant_id_or_profile_id = path.into_inner();
     let constraints = query.into_inner();
 
     let request_internal = EventListRequestInternal {
-        merchant_id_or_profile_id: merchant_id_or_profile_id.clone(),
+        merchant_id_or_profile_id: merchant_id_or_profile_id.get_string_repr().to_string(),
```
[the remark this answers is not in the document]
SanchithHegde: I'll take this change in a separate PR. For now, I've just gotten things to compile, will get these endpoints to work correctly in a separate PR.

crates/router/src/services/authentication.rs
```diff
@@ -1033,6 +1033,7 @@ where
     }
 }
 
+/*
 pub struct JWTAuthMerchantOrProfileFromRoute {
```
ThisIsMani, line 1037: Can we remove this, or is there a reason why this is kept?

jarnura (approved):

crates/router/src/services/authentication.rs
SanchithHegde: I'll address the changes around the webhook events list endpoints in a separate PR, will revisit it then. I've intentionally included changes related to the webhook events list endpoints in one commit, so that I can revisit changes from that single commit.

Narayanbhat166 (approved):

ThisIsMani (approved):

crates/router/src/services/authentication.rs
ThisIsMani: Not mandatory, but if you are making any changes in this PR, can you add a TODO comment.

Aprabhat19 (approved):

PR #5586, opened 2024-08-10
build: bump MSRV to 1.76.0

SanchithHegde:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [x] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

This PR bumps the Minimum Supported Rust Version (MSRV) from Rust 1.70 to 1.76.

This PR includes the following changes:

- Bumps MSRV to 1.76.0.
- Move lint configuration to workspace level `[lints]` table and removes the `.cargo/config.toml` file.
- Updates CI workflows to handle the removal of the `.cargo/config.toml` file.
- Removes unnecessary `#![forbid(unsafe_code)]` annotations in crates.
- Removes unused `main.rs` file in the `analytics` crate.
- Enables clippy warnings on usages of `println!()` and `eprintln!()` with the [`clippy::print_stdout`](https://rust-lang.github.io/rust-clippy/stable/index.html#print_stdout) and the [`clippy::print_stderr`](https://rust-lang.github.io/rust-clippy/stable/index.html#print_stderr) lints, respectively.
- Removes the unused `static_cache` module in the `common_utils` crate.
- Updates some usages of `map_err()` which returned the original error with `inspect_err()`.

## Motivation and Context

Bumping MSRV allows us to take advantage of the new features introduced and functions / methods stabilized between Rust 1.70 and 1.76. Some of the features / functions include:

- [Lint configuration in `Cargo.toml`](https://blog.rust-lang.org/2023/11/16/Rust-1.74.0.html#lint-configuration-through-cargo): Our previous lint configuration was done via setting `RUSTFLAGS` in `.cargo/config.toml`, which re-compiles all crates if we introduce a new lint.
- [`Result::inspect`](https://doc.rust-lang.org/stable/std/result/enum.Result.html#method.inspect), [`Result::inspect_err`](https://doc.rust-lang.org/stable/std/result/enum.Result.html#method.inspect_err) and [`Option::inspect`](https://doc.rust-lang.org/stable/std/option/enum.Option.html#method.inspect).
- [`type_name_of_val`](https://doc.rust-lang.org/stable/std/any/fn.type_name_of_val.html)

In addition, this would also allow us to update some of our dependencies to newer versions.

## How did you test it?

N/A

## Checklist


crates/router/src/core/payment_methods/cards.rs
```diff
@@ -1557,7 +1557,7 @@ pub async fn add_card_to_locker(
                 card_reference,
             )
             .await
-            .map_err(|error| {
+            .inspect_err(|_| {
```
jarnura, line 1560: instead discarding the error, log there error also. Do the same for the below inspect_err also.

crates/router/src/core/payment_methods/vault.rs
```diff
@@ -977,9 +977,8 @@ pub async fn create_tokenize(
             )
             .await
             .map(|_| lookup_key.clone())
-            .map_err(|err| {
+            .inspect_err(|_| {
```
jarnura, line 980: instead discarding the error, log there error also

crates/storage_impl/src/lib.rs
```diff
@@ -262,9 +262,8 @@ impl<T: DatabaseStore> KVRouterStore<T> {
             )
             .await
             .map(|_| metrics::KV_PUSHED_TO_DRAINER.add(&metrics::CONTEXT, 1, &[]))
-            .map_err(|err| {
+            .inspect_err(|_| {
```
jarnura, line 265: instead discarding the error, log there error also. Do the same for the below inspect_err also.

jarnura (changes requested):

crates/router/src/core/payment_methods/cards.rs
SanchithHegde: These errors won't be discarded exactly, they will be returned implicitly by `inspect_err()`. The errors will only be ignored for the purpose of `inspect_err()`, because the error information is not needed here for raising metrics.

jarnura (approved):

Chethan-rao (approved):

lsampras (approved):

Narayanbhat166 (approved):

PR #5636, opened 2024-08-15
feat(core): Add mTLS certificates for each request

Sakilmostak:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

Itau bank needs mTLS in production for verification of the request. Certificates (rootCA and pubkey) needs to be passed along with each request for Itau to verify. 
Note: To send the certificates, all related certificates needs to be concatenated and put into .pem extenstion and then should be base64 encoded
Private key should also be base64 encoded before passing into the field

### Additional Changes

## How did you test it?

This can only be tested in production since sandbox doesn't support mTLS.

Note: Apple pay (with any enabled connector), Netcetera and normal payment (with any connector) should be tested in sandbox since it affects there flows.

The above flows have been sanity tested on integ environment for this PR

## Checklist


jarnura (dismissed):

crates/router/Cargo.toml
```diff
@@ -93,7 +93,7 @@ rand = "0.8.5"
 rand_chacha = "0.3.1"
 rdkafka = "0.36.2"
 regex = "1.10.4"
-reqwest = { version = "0.11.27", features = ["json", "native-tls", "__rustls", "gzip", "multipart"] }
+reqwest = { version = "0.11.27", features = ["json", "__rustls", "rustls-tls", "gzip", "multipart"] }
```
SanchithHegde, line 96: It should be sufficient to use `rustls-tls` alone, since that internally enables `__rustls`.

crates/router/src/core/payments/helpers.rs
```diff
@@ -105,7 +105,8 @@ pub fn create_identity_from_certificate_and_key(
     let certificate_key = String::from_utf8(decoded_certificate_key)
         .change_context(errors::ApiClientError::CertificateDecodeFailed)?;
 
-    reqwest::Identity::from_pkcs8_pem(certificate.as_bytes(), certificate_key.as_bytes())
+    let key_chain = format!("{}{}", certificate_key, certificate);
```
SanchithHegde, line 108: I think we can move `create_identity_from_certificate_and_key()` and `create_certificate()` to `crates/router/src/services/api/client.rs` file.

SanchithHegde (commented):
There also seems to be another `create_identity_from_certificate_and_key()` defined in `crates/router/src/core/admin.rs` file?

PR #5642, opened 2024-08-19
refactor(core):  Refactor fallback routing behaviour in payments for v2

Aprabhat19:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description
This PR refactors the  fallback routing behaviour in payments routing for v2. Update the function to accept business_profile as a parameter instead of profile_id.

### Additional Changes

## How did you test it?
For v2:
- create a MA and add 2 MCAs, it will be added for default fallback in the order they were cretaed 
- make a payment , it should go through the mca that was added first based on default fallback

## Checklist


crates/router/src/core/admin.rs
```diff
@@ -1918,7 +1924,53 @@ impl<'a> MerchantDefaultConfigUpdate<'a> {
         Ok(())
     }
 }
-
+#[cfg(all(
+    feature = "v2",
+    feature = "routing_v2",
+    feature = "business_profile_v2"
+))]
+struct MerchantDefaultConfigUpdate<'a> {
```
jarnura, line 1932:
```suggestion
struct MerchantDefaultRoutingConfigUpdate<'a> {
```

crates/router/src/core/admin.rs
```diff
@@ -1918,7 +1924,53 @@ impl<'a> MerchantDefaultConfigUpdate<'a> {
...
+    transaction_type: &'a api_enums::TransactionType,
+    key_store: hyperswitch_domain_models::merchant_key_store::MerchantKeyStore,
+}
+#[cfg(all(
+    feature = "v2",
+    feature = "routing_v2",
+    feature = "business_profile_v2"
+))]
+impl<'a> MerchantDefaultConfigUpdate<'a> {
+    async fn update_merchant_default_config(&self) -> RouterResult<()> {
```
jarnura, line 1948:
```suggestion
    async fn update_merchant_default_routing_config(&self) -> RouterResult<()> {
```

crates/router/src/core/admin.rs
jarnura: And the type Name should be profile instead of merchant

crates/router/src/core/admin.rs
jarnura: And the function name should express what it does, like use-case-wise, eg: update_default_fallback_routing_if_routable_connector_exists.

crates/router/src/core/admin.rs
```diff
@@ -2748,6 +2804,157 @@ pub async fn create_connector(
...
+
+    let merchant_account = state
+        .store
+        .find_merchant_account_by_merchant_id(key_manager_state, merchant_id, &key_store)
+        .await
+        .to_not_found_response(errors::ApiErrorResponse::MerchantAccountNotFound)?;
+
+    let profile_id = req
+        .clone()
+        .validate_and_get_profile_id(
```
jarnura, line 2850: is this function doing a find_by profile already?

crates/router/src/core/admin.rs
```diff
@@ -2748,6 +2804,157 @@ pub async fn create_connector(
...
+        })?;
+    let connector_type_and_connector_enum = ConnectorTypeAndConnectorName {
+        connector_type: &req.connector_type,
+        connector_name: &req.connector_name,
+    };
+    let routable_connector = connector_type_and_connector_enum.get_routable_connector()?;
+
+    // The purpose of this merchant account update is just to update the
+    // merchant account `modified_at` field for KGraph cache invalidation
+    state
```
jarnura, line 2885: have a separate wrapper based function which says update_merchant_account_for_kgraph_invalidation

jarnura (changes requested):

jarnura (dismissed):

crates/router/src/core/payments.rs
```diff
@@ -3949,6 +3946,7 @@ where
     feature = "routing_v2",
     feature = "business_profile_v2"
 ))]
+#[allow(clippy::too_many_arguments)]
```
Narayanbhat166, line 3949: is this needed?

crates/router/src/core/payments.rs
```diff
@@ -4053,12 +4059,16 @@ where
     let (profile_id, routing_algorithm_id) = {
         let (profile_id, routing_algorithm) = match &transaction_data {
             TransactionData::Payment(payment_data) => (
-                payment_data.payment_intent.profile_id.clone(),
+                payment_data
```
Narayanbhat166, line 4062: we have profile_id from business_profile which is taken in the arguments

crates/router/src/core/payments.rs
```diff
@@ -4071,11 +4081,18 @@ where
             .unwrap_or_default();
         (profile_id, algorithm_ref.algorithm_id)
     };
+    // Validate that the profile_id being passed in intent should be same as the profile_id passed from the business profile parameter
+    r_utils::when(&business_profile.profile_id != profile_id, || {
```
Narayanbhat166, line 4085: we already validate this and only then payment intent is created, so this check should not be required

crates/router/src/core/payments.rs
Aprabhat19: Yes

Aprabhat19: Yes

crates/router/src/core/payments.rs
Aprabhat19: @kashif-m , can you confirm this for payout attempt's profile id as well. And if so was there any specific reason we decided to pass payment_data in Transaction type enum .

kashif-m: It's the same for payouts as well, we can use `profile_id` in business profile

We are passing payment_data or payout_data in Transaction enum mainly for making dsl_input

crates/router/src/core/admin.rs
```diff
@@ -2713,27 +2743,44 @@ pub async fn create_connector(
...
+        transaction_type: &req.get_transaction_type(),
+    };
+
+    #[cfg(all(
+        feature = "v2",
+        feature = "routing_v2",
+        feature = "business_profile_v2",
+    ))]
+    //update merchant default config
+    let merchant_default_config_update = DefaultFallbackRoutingConfigUpdate {
```
kashif-m, line 2773: Does this update account for payout's default fallback config as well?

crates/router/src/core/admin.rs
```diff
@@ -3928,31 +3975,37 @@ impl BusinessProfileWrapper {
...
+        transaction_data: &'a routing::TransactionData<'_, F>,
+    ) -> RouterResult<(&String, Option<String>)>
     where
         F: Send + Clone,
     {
-        match transaction_data {
+        let (profile_id, routing_algorithm_id) = match transaction_data {
             routing::TransactionData::Payment(payment_data) => (
-                payment_data.payment_intent.profile_id.clone(),
+                payment_data
```
kashif-m, line 3987: We can take `profile_id` from business profile here as well

crates/router/src/core/payment_methods/cards.rs
```diff
@@ -2435,10 +2435,17 @@ pub async fn list_payment_methods(
     )
     .await?;
 
+    let profile_id = profile_id
```
kashif-m, line 2438: business profile can be used here too?

crates/router/src/core/payments/routing.rs
```diff
@@ -273,36 +279,39 @@ pub async fn perform_static_routing_v1<F: Clone>(
...
         )
         .await
         .change_context(errors::RoutingError::FallbackConfigFetchFailed)?;
+        #[cfg(all(
+            feature = "v2",
+            feature = "routing_v2",
+            feature = "business_profile_v2"
+        ))]
+        let fallback_config = admin::BusinessProfileWrapper::new(business_profile.clone())
+            .get_default_fallback_list_of_connector_under_profile()
```
kashif-m, line 305: Does this account for payouts?

crates/router/src/core/admin.rs
Aprabhat19: No , this is for Payments v2 only

crates/router/src/core/payments/routing.rs
Aprabhat19: No it just accounts for payments v2

kashif-m (approved):

crates/router/src/core/payment_methods/cards.rs
Narayanbhat166: @Aprabhat19 can you check whether this api is being used with api key anywhere? Because we will not have context of business profile when used with api key and it would fail at this point

Narayanbhat166 (approved):

crates/router/src/core/payment_methods/cards.rs
Aprabhat19: Its not being used with api -key

jarnura (approved):

prajjwalkumar17 (approved):

PR #5644, opened 2024-08-20
feat(global_id): create a `GlobalId` domain type

Narayanbhat166:
## Type of Change

- [x] New feature

## Description

This PR adds a global id type that can be used to create global ids for other entities.

## Motivation and Context

All the entities henceforth will be identified by a universal identifier, i.e the global id. 

## How did you test it?

unit test cases are added
Screenshot 2024-08-22 at 12 33 04 PM

No testing required as only the types are added

## Checklist

### DIFF
--- a/crates/common_utils/src/id_type/global_id.rs
+++ b/crates/common_utils/src/id_type/global_id.rs
@@ -1,0 +1,71 @@
+#![allow(unused)]
+
+use crate::{
+    consts::{
+        CELL_IDENTIFIER_LENGTH, MAX_ALLOWED_MERCHANT_REFERENCE_ID_LENGTH,
+        MIN_REQUIRED_MERCHANT_REFERENCE_ID_LENGTH,
+    },
+    errors, generate_time_ordered_id,
+    id_type::LengthId,
+};
+use error_stack::ResultExt;
+
+#[derive(Debug, Clone, Hash, PartialEq, Eq)]
+/// A global id that can be used to identify any entity
+/// This id will have information about the entity and cell in a distributed system architecture
+struct GlobalId(
+    LengthId<MAX_ALLOWED_MERCHANT_REFERENCE_ID_LENGTH, MIN_REQUIRED_MERCHANT_REFERENCE_ID_LENGTH>,
+);
+
+/// Entities that can be identified by a global id
+enum GlobalEntity {
+    Customer,
+    Payment,
+}
+
+impl GlobalEntity {
+    fn prefix(&self) -> &'static str {
+        match self {
+            Self::Customer => "cus",
+            Self::Payment => "pay",
+        }
+    }
+}
+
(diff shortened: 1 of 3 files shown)


crates/common_utils/src/id_type/global_id.rs
```diff
@@ -0,0 +1,72 @@
...
+    },
+    errors, generate_time_ordered_id,
+    id_type::LengthId,
+};
+
+#[derive(Debug, Clone, Hash, PartialEq, Eq)]
+/// A global id that can be used to identify any entity
+/// This id will have information about the entity and cell in a distributed system architecture
+struct GlobalId(
+    LengthId<MAX_ALLOWED_MERCHANT_REFERENCE_ID_LENGTH, MIN_REQUIRED_MERCHANT_REFERENCE_ID_LENGTH>,
```
jarnura, line 18: These MAX and MIN length should be different from `MAX_ALLOWED_MERCHANT_REFERENCE_ID_LENGTH` `MIN_REQUIRED_MERCHANT_REFERENCE_ID_LENGTH`

crates/common_utils/src/id_type/global_id.rs
```diff
@@ -0,0 +1,72 @@
...
+);
+
+/// Entities that can be identified by a global id
+enum GlobalEntity {
```
jarnura, line 22: Can we have unit struct instead of having as ENUM?

crates/common_utils/src/id_type/global_id.rs
```diff
@@ -0,0 +1,72 @@
...
+    fn get_string_repr(&self) -> &str {
+        &self.0 .0 .0
+    }
+}
+
+impl GlobalId {
+    /// Create a new global id from entity and cell information
+    /// The entity prefix is used to identify the entity, `cus` for customers, `pay`` for payments etc.
+    pub fn generate(entity: GlobalEntity, cell_id: CellId) -> Self {
+        let prefix = format!("{}_{}", entity.prefix(), cell_id.get_string_repr());
```
jarnura, line 67:
```suggestion
        let prefix = format!("{}_{}", cell_id.get_string_repr(),  entity.prefix());
```

jarnura (changes requested):

crates/common_utils/src/id_type/global_id.rs
Narayanbhat166: This is for generating the prefixes only, pattern matching can be done with enums. How will unit struct help in this case? Are we planning to have some additional functionality for this?

jarnura: Yes, Instead having as Enum, Having Traits and Unit Struct do the same functionality so that we can add more usecase on those unit structs.

jarnura (changes requested):

jarnura (approved):

SanchithHegde (approved):

PR #5589, opened 2024-08-12
feat: add new routes for profile level list apis

hrithikesh026:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

The following list apis currently return records that belong to the merchant. I've created a new equivalent endpoints to retrieve the records based on profiles. 

| Operation | Existing Merchant level list endpoints    | New Profile Level list endpoints   |
|-----------|-------------|-------------|
| Connector List | /accounts/{account_id}/connectors | /accounts/{account_id}/profile/connectors |
| PaymentsList | /payments/list | /payments/profile/list |
| Get Payments Filters | /payments/v2/filter | /payments/v2/profile/filter |
| RefundList  | /refunds/list | /refunds/profile/list |
| Get Refund Filter  | /refunds/v2/filter | /refunds/v2/profile/filter |
| Payout List  | /payouts/list | /payouts/profile/list |
| Disputes List  | /disputes/list | /disputes/profile/list |

The auth layer will provide the list of profile_ids for which we need to fetch the records.

At core we will fetch all records that belongs to the merchant and then we'll filter out the records that doesn't belong to the given business profile. This change is already made in a previous pr.

### Additional Changes

- [x] This PR modifies the API contract
- [ ] This PR modifies the database schema
- [ ] This PR modifies application configuration/environment variables

## How did you test it?

1.  Merchant Connector List
2. Payment List
3. Payment List with filters
4. Get Payment Filters
5. Refund List
6. Get Refund Filters
7. Payouts List
8. Disputes List

All above apis returned 200 status code

## Checklist

- [x] I formatted the code `cargo +nightly fmt --all`
- [x] I addressed lints thrown by `cargo clippy`
- [x] I reviewed the submitted code
- [ ] I added unit tests for my changes where possible

PR #5647, opened 2024-08-20
refactor(euclid): check the authenticity of profile_id being used

prajjwalkumar17:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [x] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

This change will ensure that the profile id being sent in params, belongs to the merchant using them.

### Additional Changes

## How did you test it?

Cannot be tested right now because currently we are not including profile_id in JWT.
Was able to compile the code.

## Checklist


crates/router/src/core/routing.rs
```diff
@@ -88,11 +88,13 @@ impl RoutingAlgorithmUpdate {
 pub async fn retrieve_merchant_routing_dictionary(
     state: SessionState,
     merchant_account: domain::MerchantAccount,
+    authentication_profile_id: Option<String>,
     query_params: RoutingRetrieveQuery,
     transaction_type: &enums::TransactionType,
 ) -> RouterResponse<routing_types::RoutingKind> {
     metrics::ROUTING_MERCHANT_DICTIONARY_RETRIEVE.add(&metrics::CONTEXT, 1, &[]);
 
+    core_utils::validate_profile_id_from_auth_layer(authentication_profile_id, &query_params)?;
```
racnan, line 97: is you have `profile_id` in query params, then you can pass it in authentication layer itself to validate it. Or you can handle in handler function. 
Lets try to not have auth in core function whenever possible.

crates/router/src/services/authentication.rs
```diff
@@ -1033,6 +1033,78 @@ where
...
+                            user_id: Some(payload.user_id),
+                        },
+                    ))
+                }
+            } else {
+                return Err(report!(errors::ApiErrorResponse::InvalidJwtToken));
+            }
+        } else {
+            return Err(report!(errors::ApiErrorResponse::InvalidJwtToken));
+        }
```
hrithikesh026, line 1184: as per this logic, authentication will fail if either of the 2 profile_ids is None.

prajjwalkumar17: This is fixed as now, we are handling it in different way.

crates/router/src/routes/routing.rs
```diff
@@ -41,11 +42,17 @@ pub async fn routing_create_config(
         #[cfg(not(feature = "release"))]
         auth::auth_type(
             &auth::HeaderAuth(auth::ApiKeyAuth),
-            &auth::JWTAuth(Permission::RoutingWrite),
+            &auth::JWTAuthProfileFromRoute {
+                profile_id: json_payload.profile_id,
```
ThisIsMani, line 46: Let's not use the profile_id from the json in this auth. We can use it if the profile_id is in query param or the url.


If the profile_id is optional and is Noen, we can fallback to normal JWTAuth instead of JWTAuthProfileFromRoute.

ThisIsMani (changes requested):

crates/router/src/routes/routing.rs
prajjwalkumar17: >If the profile_id is optional and is Noen, we can fallback to normal JWTAuth instead of JWTAuthProfileFromRoute.

Are you suggesting for the places where the profile_id is Optional in query or request params, we will proceed with the validation in core or no validation at all? 
As you are asking to replace the new authProfile functions with older ones.

racnan: if authentication parameter is coming in request body (profile id in this case) then that check should be done in core function.
if it comes from path then validation should be done before invoking core function.

ThisIsMani:
`retrieve_merchant_routing_dictionary` should have 2 different endpoints.
One at merchant level and other at profile level as it is a list API. Refer #5589 for implementation.

jarnura (approved):

Cargo.lock
```diff
@@ -8,7 +8,7 @@ version = "0.5.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
```
ThisIsMani, line 8: Is this supposed to change?

crates/router/src/core/routing.rs
```diff
@@ -84,12 +84,13 @@ impl RoutingAlgorithmUpdate {
 pub async fn retrieve_merchant_routing_dictionary(
     state: SessionState,
     merchant_account: domain::MerchantAccount,
+    authentication_profile_id: Option<common_utils::id_type::ProfileId>,
```
ThisIsMani, line 87: This should accept a vector of profile_ids.
Ref: https://github.com/juspay/hyperswitch/blob/379831932a9c1e73b669f40af75cd02e682a0ec7/crates/router/src/core/payments.rs#L2917

crates/router/src/core/routing.rs
```diff
@@ -98,7 +99,20 @@ pub async fn retrieve_merchant_routing_dictionary(
...
+            if authentication_profile_id
+                .clone()
+                .is_some_and(|auth_profile| auth_profile == routing_metadata.profile_id)
+            {
+                Some(routing_metadata)
+            } else {
+                None
+            }
+        })
+        .collect();
```
ThisIsMani, line 114: Please check if you can use this.
https://github.com/juspay/hyperswitch/blob/379831932a9c1e73b669f40af75cd02e682a0ec7/crates/router/src/core/payments.rs#L2935

crates/router/src/routes/routing.rs
```diff
@@ -204,6 +214,50 @@ pub async fn list_routing_configs(
...
+    Box::pin(oss_api::server_wrap(
+        flow,
+        state,
+        &req,
+        query.into_inner(),
+        |state, auth: auth::AuthenticationData, query_params, _| {
+            routing::retrieve_merchant_routing_dictionary(
+                state,
+                auth.merchant_account,
+                auth.profile_id,
```
ThisIsMani, line 237: Should pass a vec.

ThisIsMani: https://github.com/juspay/hyperswitch/blob/379831932a9c1e73b669f40af75cd02e682a0ec7/crates/router/src/routes/payments.rs#L1049

crates/router/src/routes/routing.rs
```diff
@@ -636,31 +705,66 @@ pub async fn routing_retrieve_linked_config(
...
+        ))
+        .await
+    } else {
+        Box::pin(oss_api::server_wrap(
+            flow,
+            state,
+            &req,
+            query.clone(),
+            |state, auth: AuthenticationData, query_params, _| {
+                routing::retrieve_linked_routing_config(
```
ThisIsMani, line 748: This function should also take profile_id from the JWT and filter the data. If a profile user didn't send profile_id in the request, then this API will work as merchant level where it shouldn't.

crates/router/src/routes/routing.rs
```diff
@@ -636,31 +705,66 @@ pub async fn routing_retrieve_linked_config(
...
+                    required_permission: Permission::RoutingRead,
+                },
+                req.headers(),
+            ),
+            #[cfg(feature = "release")]
+            &auth::JWTAuthProfileFromRoute {
+                profile_id,
+                required_permission: Permission::RoutingRead,
+            },
+            api_locking::LockAction::NotApplicable,
```
ThisIsMani, line 738: NIT: You can construct this auth type before calling the server_wrap.

prajjwalkumar17: This can't be done as this is a shared reference and will be dropped before being used.
FYI: did this in my first trials.

ThisIsMani (approved):

jarnura (approved):

kashif-m (approved):

crates/router/src/routes/app.rs
```diff
@@ -711,6 +711,17 @@ impl Routing {
                         )
                     })),
             )
+            .service(web::resource("/list/{profile_id}").route(web::get().to(
```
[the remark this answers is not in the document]
prajjwalkumar17: Thia will be taken up in the immediate next PR, considering this PR is in pipeline from a long time.

kashif-m (approved):

crates/router/src/routes/admin.rs
```diff
@@ -892,8 +892,7 @@ pub async fn business_profile_update(
         },
         auth::auth_type(
             &auth::AdminApiAuthWithMerchantIdFromRoute(merchant_id.clone()),
-            &auth::JWTAuthMerchantAndProfileFromRoute {
-                merchant_id: merchant_id.clone(),
+            &auth::JWTAuthProfileFromRoute {
```
ThisIsMani, line 895: This should not be changed.

ThisIsMani (changes requested):

ThisIsMani (approved):

Aprabhat19 (approved):

kashif-m (approved):

SanchithHegde (approved):

sai-harsha-vardhan (approved):

PR #5687, opened 2024-08-25
refactor: introduce a domain type for profile ID

SanchithHegde:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [x] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

This PR introduces a domain type for profile ID and refactors the existing code to use the domain type in place of strings being used today.

In addition, this PR has the following changes:

- Introduces a trait `GenerateId` for generating IDs.
- Removes the `Default` implementation on `PayoutsNew` and `PayoutAttemptNew` types defined in the `hyperswitch_domain_models` crate, since the `ProfileId` type does not implement `Default`.

### Additional Changes

This PR has no new contract changes, this PR only updates the merchant connector account create request in the OpenAPI documentation to include the maximum character length restriction placed on the `profile_id` field.

## Motivation and Context

Using a domain type should help better enforce restrictions like minimum and maximum character restrictions, and should help prevent from incorrect values being passed around (to some extent).

## How did you test it?

Sanity testing with Postman.

## Checklist


crates/router/src/core/admin.rs
```diff
@@ -885,21 +883,16 @@ impl MerchantAccountUpdateBridge for api::MerchantAccountUpdate {
...
-                    key_store,
-                    Some(profile_id),
-                    merchant_id,
-                )
-                .await?
-                .map(|business_profile| Some(business_profile.profile_id))
-            } else {
-                // If empty, Update profile_id to None in the database
-                Some(None)
-            }
```
SanchithHegde, line 902: We won't be able to allow the "if empty set default profile to null" behavior anymore if using the `ProfileId` domain type for the field, since the domain type doesn't allow constructing profile IDs from empty strings.

Narayanbhat166: Thanks for pointing this out!

jarnura: @JeevaRamu0104 Tagging for visibility

jarnura (dismissed):

jarnura (dismissed):

jarnura (approved):

Narayanbhat166 (approved):

Aprabhat19 (approved):

lsampras (approved):

PR #5700, opened 2024-08-26
feat(payment_methods_v2): Implemented Diesel and Domain models for v2

Sarthak1799:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

- Diesel model for payment methods v2
- Domain models for payment methods v1/v2

### Additional Changes

## How did you test it?

- Test all flows (payments/payment_methods) for v1
- v2 cannot be tested as core logic is incomplete

## Checklist


jarnura (changes requested):

crates/diesel_models/src/query/payment_method.rs
```diff
@@ -173,3 +183,146 @@ impl PaymentMethod {
         }
     }
 }
+
+#[cfg(all(feature = "v2", feature = "payment_methods_v2"))]
+impl PaymentMethod {
+    pub async fn delete_by_id(conn: &PgPooledConn, id: &str) -> StorageResult<Self> {
```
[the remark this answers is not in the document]
Sarthak1799: We might not require this query for soft delete. Will remove this then.

crates/diesel_models/src/query/payment_method.rs
```diff
@@ -173,3 +183,146 @@ impl PaymentMethod {
...
+        generics::db_metrics::track_database_call::<<Self as HasTable>::Table, _, _>(
+            filter.get_result_async::<i64>(conn),
+            generics::db_metrics::DatabaseOperation::Count,
+        )
+        .await
+        .change_context(errors::DatabaseError::Others)
+        .attach_printable("Failed to get a count of payment methods")
+    }
+
+    pub async fn find_by_customer_id_merchant_id_status(
```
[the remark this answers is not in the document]
Sarthak1799: In Saved payment method list, we query by (merchant_id, customer_id) to get a list of saved payment methods for the customer.

crates/diesel_models/src/payment_method.rs
```diff
@@ -142,6 +280,63 @@ impl PaymentMethodUpdate {
     }
 }
 
+#[cfg(all(feature = "v2", feature = "payment_methods_v2"))]
+#[derive(
+    Clone, Debug, Default, AsChangeset, router_derive::DebugAsDisplay, Serialize, Deserialize,
+)]
+#[diesel(table_name = payment_methods)]
+pub struct PaymentMethodUpdateInternal {
```
SanchithHegde, line 288: Can we have `last_modified` as a required field in both the `PaymentMethodUpdateInternal` structs, so that we can ensure that the field is always updated with any update to the payment method record? (You would have to remove the `Default` derive on `PaymentMethodUpdateInternal` as a result.)

crates/router/src/core/payment_methods/cards.rs
```diff
@@ -4986,31 +5089,19 @@ where
     not(feature = "payment_methods_v2")
 ))]
 pub async fn get_card_details_with_locker_fallback(
-    pm: &payment_method::PaymentMethod,
+    pm: &domain::PaymentMethod,
     state: &routes::SessionState,
-    key_store: &domain::MerchantKeyStore,
+    _key_store: &domain::MerchantKeyStore,
```
SanchithHegde, line 5094: We can remove function parameters that are not accessed in both v1 and v2 implementations. You can do the same change for other functions in this file that don't access `key_store`.

crates/router/src/core/payment_methods/cards.rs
```diff
@@ -5180,25 +5237,13 @@ pub async fn get_lookup_key_from_locker(
 }
 
 async fn get_masked_bank_details(
-    state: &routes::SessionState,
-    pm: &payment_method::PaymentMethod,
-    key_store: &domain::MerchantKeyStore,
+    _state: &routes::SessionState,
```
SanchithHegde, line 5240: We can remove this parameter as well, from functions which don't access it.

crates/storage_impl/Cargo.toml
```diff
@@ -16,6 +16,7 @@ v1 = ["api_models/v1", "diesel_models/v1", "hyperswitch_domain_models/v1"]
 v2 = ["api_models/v2", "diesel_models/v2", "hyperswitch_domain_models/v2"]
 payment_v2 = ["hyperswitch_domain_models/payment_v2", "diesel_models/payment_v2"]
 customer_v2 = ["api_models/customer_v2", "diesel_models/customer_v2", "hyperswitch_domain_models/customer_v2"]
+payment_methods_v2 = ["diesel_models/payment_methods_v2"]
```
SanchithHegde, line 19: Please update `scripts/ci-checks-v2.sh` to include this feature, wherever `storage_impl` crate is being checked.

v2_migrations/2024-08-23-112510_payment_methods_v2_db_changes/up.sql
```diff
@@ -0,0 +1,34 @@
...
+
+ALTER TABLE payment_methods DROP COLUMN IF EXISTS payment_method_issuer;
+
+ALTER TABLE payment_methods DROP COLUMN IF EXISTS payment_method_issuer_code;
+
+DROP TYPE IF EXISTS "PaymentMethodIssuerCode";
+
+ALTER TABLE payment_methods DROP CONSTRAINT IF EXISTS payment_methods_pkey;
+ALTER TABLE payment_methods DROP COLUMN IF EXISTS id;
+ALTER TABLE payment_methods ADD COLUMN IF NOT EXISTS id VARCHAR(64) PRIMARY KEY;
```
SanchithHegde, line 30: We cannot be adding a primary key column to an existing database though, since the value in the column for all existing records would be `NULL`?

This would have to be done along the lines of:

- Drop existing `id` column.
- Add new `id` column, without a `NOT NULL` or a `PRIMARY KEY` constraint.
- Copy over data from `payment_method_id` column to `id` column.
- Drop primary key constraint on `payment_method_id` column.
- Add primary key constraint on `id` column.
- Drop `payment_method_id` column (eventually).

v2_migrations/2024-08-23-112510_payment_methods_v2_db_changes/down.sql
```diff
@@ -0,0 +1,44 @@
...
+    'jp_bacs'
+);
+
+ALTER TABLE payment_methods ADD COLUMN IF NOT EXISTS payment_method_issuer_code "PaymentMethodIssuerCode";
+
+ALTER TABLE payment_methods DROP CONSTRAINT IF EXISTS payment_methods_pkey;
+
+ALTER TABLE payment_methods ADD COLUMN IF NOT EXISTS payment_method_id VARCHAR(64) PRIMARY KEY;
+
+UPDATE payment_methods SET payment_method_id = id;
```
SanchithHegde, line 41: Same applies here, we'd have to copy over data to the `payment_method_id` column first and then add the primary key constraint.

crates/diesel_models/src/query/payment_method.rs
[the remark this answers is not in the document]
jarnura: Any one query is enough right, keeping status and use that to solve other cases also

crates/diesel_models/src/query/payment_method.rs
```diff
@@ -173,3 +187,108 @@ impl PaymentMethod {
...
+                .eq(customer_id.to_owned())
+                .and(dsl::merchant_id.eq(merchant_id.to_owned())),
+            limit,
+            None,
+            Some(dsl::last_used_at.desc()),
+        )
+        .await
+    }
+
+    pub async fn get_count_by_customer_id_merchant_id_status(
```
jarnura, line 216: If this is only for OLAP add feature gate

jarnura (changes requested):

crates/diesel_models/src/query/payment_method.rs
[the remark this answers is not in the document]
Sarthak1799: have removed the other one.

crates/diesel_models/src/query/payment_method.rs
Sarthak1799: This query was introduced for default payment method flow, we can decide later on if this is needed for v2, have removed it for now.

jarnura (approved):

crates/router/src/core/pm_auth.rs
```diff
@@ -328,19 +322,39 @@ async fn store_bank_details_in_payment_methods(
...
     let payment_methods = db
         .find_payment_method_by_customer_id_merchant_id_list(
+            &((&state).into()),
+            &key_store,
             &customer_id,
             merchant_account.get_id(),
             None,
         )
         .await
         .change_context(ApiErrorResponse::InternalServerError)?;
```
SanchithHegde, line 338: Attach context messages wherever internal server errors are raised, here and elsewhere in this file.

crates/router/src/db/payment_method.rs
```diff
@@ -359,36 +551,155 @@ mod storage {
...
+        ) -> CustomResult<domain::PaymentMethod, errors::StorageError> {
+            let payment_method = Conversion::convert(payment_method)
+                .await
+                .change_context(errors::StorageError::DecryptionError)?;
+
+            let merchant_id = payment_method.merchant_id.clone();
+            let customer_id = payment_method.customer_id.clone();
+            let key = PartitionKey::MerchantIdCustomerId {
+                merchant_id: &merchant_id,
+                customer_id: customer_id.get_string_repr(),
```
SanchithHegde, line 583: Why are we using a string representation here, can't we use the domain type itself?

SanchithHegde (approved):

crates/router/src/core/pm_auth.rs
Sarthak1799: We'll take it up in a separate PM auth PR

crates/router/src/core/payment_methods/cards.rs
```diff
@@ -186,6 +186,7 @@ pub async fn create_payment_method(
                 last_used_at: current_time,
                 payment_method_billing_address,
                 updated_by: None,
+                version: enums::ApiVersion::V1,
```
SanchithHegde, line 189: Can we use constants instead?

jarnura (dismissed):

SanchithHegde (dismissed):

Aprabhat19 (dismissed):

SanchithHegde (approved):

Aprabhat19 (approved):

jarnura (approved):

Issue #5776, opened 2024-09-02
feat(user_role): add support to get user role details v2


apoorvdixit88:
Add support to return user role details v2
Request will take email and response will return something like
```
[
		{
			role_id: "r1",
			role_name: "name",
			org: {
				name: "Juspay",
				id: "o1"
			},
			merchant: {                       // Optional
				name: "Juspay merchant",
				id: "m1"
			},
			profile: {                        // Optional
				name: "Juspay profile",
				id: "p1"
			},
			status: "status",
			entity_type: "type"
		}
	]
```

PR #5777, opened 2024-09-02
feat(user_roles): get user role details

apoorvdixit88:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description
Add support to get user role details

### Additional Changes

## Motivation and Context
Closes [#5776](https://github.com/juspay/hyperswitch/issues/5776)

## How did you test it?
Request:
```
curl --location 'http://localhost:8080/user/user/v2' \
--header 'Authorization: Bearer JWT' \
--data-raw '{
    "email": "some_email"
}'
```
Response
```
[
    {
        "role_id": "org_admin",
        "org": {
            "name": null,
            "id": "org_SDzPXs082bpEQt3jbPlE"
        },
        "merchant": null,
        "profile": null,
        "status": "Active",
        "entity_type": "organization"
    }
]
```

## Checklist


crates/router/src/core/user.rs
```diff
@@ -1320,6 +1321,177 @@ pub async fn get_user_details_in_merchant_account(
...
+        })
+        .await
+        .change_context(UserErrors::InternalServerError)
+        .attach_printable("Failed to fetch user roles")?
+        .into_iter()
+        .collect();
+
+    let mut role_details_list = Vec::new();
+
+    for user_role in user_roles_set {
```
ThisIsMani, line 1354: Parallelize.

crates/router/src/core/user.rs
```diff
@@ -1320,6 +1321,177 @@ pub async fn get_user_details_in_merchant_account(
...
+                        .merchant_id
+                        .ok_or(UserErrors::InternalServerError)
+                        .attach_printable("Merchant id not found for custom role")?,
+                    &user_from_token.org_id,
+                )
+                .await
+                .change_context(UserErrors::InternalServerError)
+                .attach_printable("Failed to get custom role")?
+                .into()
+        };
```
ThisIsMani, line 1479: Why not `from_role_id()`?

crates/router/src/routes/user.rs
```diff
@@ -336,6 +336,24 @@ pub async fn get_user_role_details(
     .await
 }
 
+pub async fn list_user_roles_details(
+    state: web::Data<AppState>,
+    req: HttpRequest,
+    payload: web::Query<user_api::GetUserRoleDetailsRequest>,
+) -> HttpResponse {
+    let flow = Flow::GetUserRoleDetails;
```
ThisIsMani, line 344: New flow?

crates/router/src/core/user.rs
apoorvdixit88: coz merchant id is non option param in that, he we may or may not get it.

crates/router/src/routes/user.rs
apoorvdixit88: Yes we can have, but this flow was not getting used anywhere

crates/router/src/core/user.rs
ThisIsMani: Make a function in roles which takes user role and gives back a role info.

If the role_id is present in the predefined roles it will not do any db calls.
Else if the entity type is org, then throw internal server error (As we don't have any org level custom roles.)
Else if the entity type is merchant, then get merchant id from the user role (if it is not there throw 500) and find the custom role.
Else if the entity type is merchant or profile, then get merchant id from the user role (if it is not there throw 500) and find the custom role.

You can change the `from_role_id()` function if it is possible.

crates/router/src/core/user.rs
```diff
@@ -1320,6 +1321,210 @@ pub async fn get_user_details_in_merchant_account(
...
+                            name: profile_map
+                                .get(profile_id)
+                                .ok_or(UserErrors::InternalServerError)?
+                                .to_owned(),
+                        })
+                    } else {
+                        return Err(UserErrors::InternalServerError);
+                    }
+                }
+            };
```
ThisIsMani, line 1512: Can be done in single match statement.

ThisIsMani (approved):
Some suggestions.

ThisIsMani (approved):

crates/router/src/core/user.rs
```diff
@@ -1407,6 +1407,237 @@ pub async fn get_user_details_in_merchant_account(
...
+        .attach_printable("Org id not found")?
+        .get_organization_name();
+
+    let org = NameIdUnit {
+        id: user_from_token.org_id.clone(),
+        name: org_name,
+    };
+
+    let (merchant_ids, merchant_profile_ids) = user_roles_set.iter().try_fold(
+        (Vec::new(), Vec::new()),
```
SanchithHegde, line 1467: You can probably use something like `Vec::with_capacity(user_roles_set.len())` here, if the size of the `Vec` is known beforehand, and is same as `user_roles_set`.

apoorvdixit88: No the size is not fix, we will be only considering those use roles, where the entity type is merchant and profile.
And ignoring the internal and organisation in this use case

apoorvdixit88: No the size is not fix, we will be only considering those use roles, where the entity type is merchant and profile.
And ignoring the internal and organisation in this use case

SanchithHegde (approved):

PR #5825, opened 2024-09-06
refactor: Add a GAT Data to Operation trait

jarnura:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [x] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

Refactor the `Operation` trait to have different types of session tracking object for every implementation

## Motivation and Context

- Operation trait, this trait groups `ValidateRequest`, `Domain`, `GetTracker`, `UpdateTrackers`, these traits members function uses `PaymentData` to tracker session related objects in it.
- `PaymentData` is not same for all requests and other verticals, because of this we need to add field is irreverent to a Operation. So introducing a `GAT` type called Data to `Operation` trait to support, each instance of Operation can have different session object.
- The implementation details, database calls related to the implementation and mutating `PaymentData` were in helper function which is moved backed to implementation block of the member function for `get_or_create_customer_details`.
- Unwanted usage of PaymentData in many places, when the function requires only some fields or a object, this adds friction to use small modules in other places of code because it required `PaymentData` and `PaymentData` can't be constructed easily.
- There are lots of places `PaymentData` mutual reference is passed when it is not required.
- Mutability of PaymentData logic is everywhere when it needs to be only in core flow only.
- There are many unwanted clones for `ProfileId`
-  Payouts using F, were F is generic which tells which payment flow the current handler is, which is not required for payouts.
- F is not required for Routing DSL related functions.
- Payout and Payments routing logic is different but how to decide the routing, those function are reused but not by using traits for polymorphism, which cause more friction to changes for payments and vice-versa.
- The same problem exists for FrmOperation also.
- Changed &String to &str

### Additional Changes

- Introduced a GAT type called Data to Operation trait to handle different session tracker object for each handler.
- A two new trait is introduced namely `OperationSessionGetters`, `OperationSessionSetters`, the GAT type has trait bound on these, so that the member function of these  `ValidateRequest`, `Domain`, `GetTracker`, `UpdateTrackers` can access Operation's Data.
- Made some and structs to use `ProfileId` as reference to avoid multiple clones of the same data.
- Moving the database call and mutation logic of `PaymentData` for `get_or_create_customer_details` back to trait implementation instead of helpers
- The F is not required for payouts, the dependencies is removed for some of the functions.
- The F is not required for some functions, the dependencies is removed for those functions.
	- `get_routing_algorithm_id`
	- `perform_decision_management`
	- `make_dsl_input`
	- `perform_eligibility_analysis`
	- `perform_fallback_routing`
	- `perform_eligibility_analysis_with_fallback`
- The routing function for payouts and payments where separated.
- Added a generic in FraudOperation to handle multiple sessions object of payments
- Remove mut PaymentData in some of the places where mutability not required
- Unwanted usage of PaymentData in multiple places also removed to reduce friction in using small modules.

## How did you test it?

- There are types related changes, so compilation is running successfully for both v1 and v2.
- Adding to that then tests in postman in CI/CD also running successfully. 
  

## Checklist


crates/router/src/core/payments/transformers.rs
```diff
@@ -633,48 +644,37 @@ where
...
-                .collect(),
-        )
-    };
+    let refunds_response = payment_data.get_refunds().is_empty().then(|| {
+        payment_data
+            .get_refunds()
+            .into_iter()
+            .map(ForeignInto::foreign_into)
+            .collect()
+    });
```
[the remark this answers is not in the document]
jarnura: Yes, my bad. Fixing it.

SanchithHegde (approved):

crates/router/src/core/payments.rs
```diff
@@ -3567,17 +3652,29 @@ where
...
         let (mut connectors, check_eligibility) = routing::perform_straight_through_routing(
             &routing_algorithm,
-            payment_data.creds_identifier.clone(),
+            payment_data.get_creds_identifier(),
         )
         .change_context(errors::ApiErrorResponse::InternalServerError)
         .attach_printable("Failed execution of straight through routing")?;
 
         if check_eligibility {
+            // let setup_mandate = payment_data.get_setup_mandate();
```
sai-harsha-vardhan, line 3661: Can we please remove these if not required?

crates/router/src/core/payments.rs
```diff
@@ -3614,17 +3711,31 @@ where
...
         let (mut connectors, check_eligibility) = routing::perform_straight_through_routing(
             routing_algorithm,
-            payment_data.creds_identifier.clone(),
+            payment_data.get_creds_identifier(),
         )
         .change_context(errors::ApiErrorResponse::InternalServerError)
         .attach_printable("Failed execution of straight through routing")?;
 
         if check_eligibility {
+            // let setup_mandate = payment_data.get_setup_mandate();
```
sai-harsha-vardhan, line 3720: Same here

sai-harsha-vardhan (approved):
Core changes LGTM

Aprabhat19 (approved):

crates/router/src/core/payments.rs
jarnura: Sure will remove these, the changes are addressed in this https://github.com/juspay/hyperswitch/pull/5851

crates/router/src/core/payments.rs
jarnura: Sure will remove these, the changes are addressed in this https://github.com/juspay/hyperswitch/pull/5851

crates/router/src/core/payments/operations.rs
```diff
@@ -41,32 +41,33 @@ use crate::{
...
 
-pub type BoxedOperation<'a, F, T> = Box<dyn Operation<F, T> + Send + Sync + 'a>;
+pub type BoxedOperation<'a, F, T, D> = Box<dyn Operation<F, T, Data = D> + Send + Sync + 'a>;
 
 pub trait Operation<F: Clone, T>: Send + std::fmt::Debug {
-    fn to_validate_request(&self) -> RouterResult<&(dyn ValidateRequest<F, T> + Send + Sync)> {
+    type Data;
+    fn to_validate_request(
+        &self,
+    ) -> RouterResult<&(dyn ValidateRequest<F, T, Self::Data> + Send + Sync)> {
```
Narayanbhat166, line 50: should validate request also have the PaymentData associated with it? The lifecycle of PaymentData starts from getTrackers right?

jarnura: We are returning `BoxedOperation` from `ValidateRequest` which needs the type information. So the type is added to `ValidateRequest` too. This is only from type stitching

PR #5783, opened 2024-09-03
feat(payments_v2): payment intent diesel and domain models changes v2

Narayanbhat166:
## Type of Change

- [x] Refactoring

## Description

This PR refactors the codebase to include all the payment related operations under the v1 feature flag. 

This also includes changes to the payment intent diesel and domain models to add new fields that are relevant to the v2 version of application.

### Additional Changes

- [ ] This PR modifies the API contract
- [x] This PR modifies the database schema
- [ ] This PR modifies application configuration/environment variables

## Checklist

- [x] I formatted the code `cargo +nightly fmt --all`
- [x] I addressed lints thrown by `cargo clippy`
- [x] I reviewed the submitted code

PR #5880, opened 2024-09-13
refactor(payment_intent_v2): payment intent fields refactoring

Narayanbhat166:
## Type of Change

- [x] Refactoring

## Description

This PR addresses some comments that were left on #5783. Few fields have been mademandatory and renamed
- renamed `statement_descriptor_name` to `statement_descriptor`
- make `currecy`, `client_secret` and `profile_id` as a mandatory field in payment intent
- group amount related fields to a struct to be used in payment intent domain model
```rust
pub struct AmountDetails {
    /// The amount of the order in the lowest denomination of currency
    order_amount: MinorUnit,
    /// The currency of the order
    currency: common_enums::Currency,
    /// The shipping cost of the order. This has to be collected from the merchant
    shipping_cost: Option<MinorUnit>,
    /// Tax details related to the order. This will be calculated by the external tax provider
    tax_details: Option<TaxDetails>,
    /// The action to whether calculate tax by calling external tax provider or not
    skip_external_tax_calculation: TaxCalculationOverride,
    /// The action to whether calculate surcharge or not
    skip_surcharge_calculation: SurchargeCalculationOverride,
    /// The surcharge amount to be added to the order, collected from the merchant
    surcharge_amount: Option<MinorUnit>,
    /// tax on surcharge amount
    tax_on_surcharge: Option<MinorUnit>,
}
```

- Change all the boolean values in the domain models to enums indicating relevant actions that has to be taken.
- Add doc comments to all the fields in payment intent domain model.

This PR also move some other parts of the code under v1 flag, so that when the v2 code is written, these functions will be revisited and then they with either be refactored to be reused / rewritten.

This PR also adds the `PaymentGlobalId` support for `PaymentIntent`

### Additional Changes

This affects the v2 payment intent schema

## Motivation and Context

Refactor the fields in v1 that will be removed / not supported in v2 and refactor the code for general improvements in code quality.

## How did you test it?

This PR does not add any feature. Only v2 code is affected. A sanity of payments should ensure that this PR does not introduce any bugs.

## Impact
payment intent v2 

## Checklist


crates/diesel_models/src/schema_v2.rs
```diff
@@ -890,12 +882,15 @@ diesel::table! {
...
-        amount_to_capture -> Nullable<Int8>,
         prerouting_algorithm -> Nullable<Jsonb>,
         surcharge_amount -> Nullable<Int8>,
         tax_on_surcharge -> Nullable<Int8>,
         #[max_length = 64]
         frm_merchant_decision -> Nullable<Varchar>,
+        #[max_length = 255]
+        statement_descriptor -> Nullable<Varchar>,
+        enable_payment_link -> Nullable<Bool>,
+        apply_mit_exemption -> Nullable<Bool>,
```
jarnura, line 893: Add the below fields,
- customer_present
- payment_link_config
- routing_algorithm_id

jarnura (changes requested):

crates/hyperswitch_domain_models/src/payments.rs
```diff
@@ -82,70 +82,179 @@ impl PaymentIntent {
...
+
+#[cfg(feature = "v2")]
+impl From<Option<bool>> for TaxCalculationOverride {
+    fn from(value: Option<bool>) -> Self {
+        match value {
+            Some(true) => TaxCalculationOverride::Calculate,
+            _ => TaxCalculationOverride::Skip,
+        }
+    }
+}
```
hrithikesh026, line 116: I don't think it is a good idea to map None and Some(false) to TaxCalculationOverride::Skip.
During conversion from domain model to diesel model, we will end up populating surcharge_applicable field with Some(false) even if it was None in DB.

hrithikesh026: In KV store

Narayanbhat166: yes that is expected. We should always have a value, either `true` or `false` in the database

crates/diesel_models/src/schema_v2.rs
Narayanbhat166: These fields have been added

crates/hyperswitch_domain_models/src/payments.rs
hrithikesh026: Since `surcharge_applicable` in `payment_intent` is an inferred field. It should be in `payment_attempt`.

`surcharge_applicable` is set to true during payment_method_list call. If any surcharge is calculated through the surcharge rules, we'll set `surcharge_applicable` to `true`. 
Here you have considered it as flag to skip surcharge calculation. 
The decision to skip surcharge calculation is made through `surcharge_amount` field in payment_intent. If the merchant sends some `surcharge_amount` through payments create call, we'll will skip surcharge calculation and send whatever surcharge sent by the merchant to the connector.

Narayanbhat166: All the fields in `PaymentIntent` are about the intent of the merchant. In this case we cannot use this field for our internal use. This behaviour will need to change in v2. This field will only tell about whether merchant wants us to calculate the surcharge or not.

crates/hyperswitch_domain_models/src/payments.rs
```diff
@@ -82,70 +82,179 @@ impl PaymentIntent {
...
+    /// The description of the order. This will be passed to connectors which support description.
     pub description: Option<String>,
+    /// The return url for the payment. This is the url to which the user will be redirected after the payment is completed.
     pub return_url: Option<String>,
-    pub metadata: Option<serde_json::Value>,
-    pub statement_descriptor_name: Option<String>,
+    /// The metadata for the payment intent. This is the metadata that will be passed to the connectors.
+    pub metadata: Option<pii::SecretSerdeValue>,
+    /// The statement descriptor for the order, this will be displayed in the user's bank statement.
+    pub statement_descriptor: Option<String>,
```
jarnura, line 191: This can be a `Option<StatementDescriptor>`, type can have a max length validation on new.

crates/hyperswitch_domain_models/src/payments.rs
```diff
@@ -82,70 +82,179 @@ impl PaymentIntent {
...
-    pub off_session: Option<bool>,
-    pub client_secret: Option<String>,
+    /// The client secret that is generated for the payment. This is used to authenticate the payment from client facing apis.
+    pub client_secret: String,
+    /// The active attempt for the payment intent. This is the payment attempt that is currently active for the payment intent.
     pub active_attempt: RemoteStorageObject<PaymentAttempt>,
     pub order_details: Option<Vec<pii::SecretSerdeValue>>,
+    /// This is the list of payment method types that are allowed for the payment intent.
+    /// This field allows the merchant to restrict the payment methods that can be used for the payment intent.
     pub allowed_payment_method_types: Option<serde_json::Value>,
```
jarnura, line 208:
```suggestion
    pub allowed_payment_method_types: Option<pii::SecretSerdeValue>,
```

crates/hyperswitch_domain_models/src/payments.rs
```diff
@@ -82,70 +82,179 @@ impl PaymentIntent {
...
-    pub profile_id: Option<id_type::ProfileId>,
+    /// The profile id for the payment.
+    pub profile_id: id_type::ProfileId,
+    /// The payment link id for the payment. This is generated only if `enable_payment_link` is set to true.
     pub payment_link_id: Option<String>,
-    // Denotes the action(approve or reject) taken by merchant in case of manual review.
-    // Manual review can occur when the transaction is marked as risky by the frm_processor, payment processor or when there is underpayment/over payment incase of crypto payment
+    /// This Denotes the action(approve or reject) taken by merchant in case of manual review.
+    /// Manual review can occur when the transaction is marked as risky by the frm_processor, payment processor or when there is underpayment/over payment incase of crypto payment
     pub frm_merchant_decision: Option<String>,
```
jarnura, line 220: Decision can be a enum

crates/hyperswitch_domain_models/src/payments.rs
```diff
@@ -82,70 +82,179 @@ impl PaymentIntent {
...
+    /// The shipping address for the order in a denormalized form.
     pub shipping_address: Option<Encryptable<Secret<serde_json::Value>>>,
+    /// Capture method for the payment
     pub capture_method: Option<storage_enums::CaptureMethod>,
-    pub id: id_type::PaymentId,
+    /// Authentication type that is requested by the merchant for this payment.
     pub authentication_type: Option<common_enums::AuthenticationType>,
-    pub amount_to_capture: Option<MinorUnit>,
+    /// This contains the pre routing results that are done when routing is done during listing the payment methods.
     pub prerouting_algorithm: Option<serde_json::Value>,
```
jarnura, line 247: Is this for tracking purpose then can we make this as prerouting_algorithm_id

jarnura (changes requested):

jarnura (dismissed):

Aprabhat19 (dismissed):

Aprabhat19 (approved):

jarnura (approved):

PR #5885, opened 2024-09-13
feat(v2): Add cell id in config for v2

Sarthak1799:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

Added Cell Id in SessionState config

### Additional Changes

## Checklist

### DIFF
--- a/config/development.toml
+++ b/config/development.toml
@@ -734,3 +734,6 @@ connector_list = ""
 
 [recipient_emails]
 recon = "recon@example.com"
+
+[global_cell_id]
+cell_id = "Random_id"
--- a/crates/common_utils/src/id_type.rs
+++ b/crates/common_utils/src/id_type.rs
@@ -21,6 +21,8 @@ use diesel::{
     serialize::{Output, ToSql},
     sql_types,
 };
+#[cfg(feature = "v2")]
+pub use global_id::CellId;
 pub use merchant::MerchantId;
 pub use merchant_connector_account::MerchantConnectorAccountId;
 pub use organization::OrganizationId;
--- a/crates/common_utils/src/id_type/global_id.rs
+++ b/crates/common_utils/src/id_type/global_id.rs
@@ -32,8 +32,8 @@ impl GlobalEntity {
     }
 }
 
-#[derive(Debug, Clone, Hash, PartialEq, Eq)]
-pub(crate) struct CellId(LengthId<CELL_IDENTIFIER_LENGTH, CELL_IDENTIFIER_LENGTH>);
+#[derive(Debug, Clone, Hash, PartialEq, Eq, serde::Deserialize)]
+pub struct CellId(LengthId<CELL_IDENTIFIER_LENGTH, CELL_IDENTIFIER_LENGTH>);
 
 #[derive(Debug, Error, PartialEq, Eq)]
 pub enum CellIdError {
--- a/crates/router/src/configs/settings.rs
+++ b/crates/router/src/configs/settings.rs
@@ -47,8 +47,7 @@ pub struct CmdLineConf {
     pub config_path: Option<PathBuf>,
 }
 
-#[derive(Debug, Deserialize, Clone, Default)]
-#[serde(default)]
+#[derive(Debug, Deserialize, Clone)]
 pub struct Settings<S: SecretState> {
     pub server: Server,
     pub proxy: Proxy,
@@ -121,6 +120,8 @@ pub struct Settings<S: SecretState> {
     pub decision: Option<DecisionConfig>,
     pub locker_based_open_banking_connectors: LockerBasedRecipientConnectorList,
     pub recipient_emails: RecipientMails,
+    #[cfg(feature = "v2")]
+    pub global_cell_id: GlobalCellId,
 }
 
 #[derive(Debug, Deserialize, Clone, Default)]
@@ -906,6 +907,11 @@ pub struct ServerTls {
     pub certificate: PathBuf,
 }
 
+#[cfg(feature = "v2")]
+#[derive(Debug, Clone, Deserialize)]
+pub struct GlobalCellId {
+    pub cell_id: common_utils::id_type::CellId,
+}
 #[derive(Debug, Deserialize, Clone, Default)]
 pub struct RecipientMails {
     pub recon: Email,
--- a/crates/router/src/configs/secrets_transformers.rs
+++ b/crates/router/src/configs/secrets_transformers.rs
@@ -459,5 +459,7 @@ pub(crate) async fn fetch_raw_secrets(
         decision: conf.decision,
         locker_based_open_banking_connectors: conf.locker_based_open_banking_connectors,
         recipient_emails: conf.recipient_emails,
+        #[cfg(feature = "v2")]
+        global_cell_id: conf.global_cell_id
     }
 }


crates/common_utils/src/id_type/global_id.rs
```diff
@@ -32,8 +32,8 @@ impl GlobalEntity {
     }
 }
 
-#[derive(Debug, Clone, Hash, PartialEq, Eq)]
-pub(crate) struct CellId(LengthId<CELL_IDENTIFIER_LENGTH, CELL_IDENTIFIER_LENGTH>);
+#[derive(Debug, Clone, Hash, PartialEq, Eq, serde::Deserialize)]
```
Narayanbhat166, line 35: do not derive Deserialization directly, implement custom deserialization which also validates the constraints. You can refer to the Deser implementation of `AlphanumericId`

config/development.toml
```diff
@@ -734,3 +734,6 @@ connector_list = ""
 
 [recipient_emails]
 recon = "recon@example.com"
+
+[global_cell_id]
+cell_id = "Random_id"
```
Narayanbhat166, line 739:
```suggestion
[cell_information]
id = "abcdef"
```

Narayanbhat166: The cell_id should be of 5 characters only

Narayanbhat166 (changes requested):

jarnura (approved):

crates/common_utils/src/id_type/global_id.rs
```diff
@@ -79,6 +79,23 @@ impl CellId {
     }
 }
 
+impl<'de> serde::Deserialize<'de> for CellId {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: serde::Deserializer<'de>,
+    {
+        let deserialized_length_id = LengthId::deserialize(deserializer)?;
```
Narayanbhat166, line 87: can you deserialize to string, and then call `from_str` function?

Narayanbhat166: You can refer to the similar implementation here 
https://github.com/juspay/hyperswitch/blob/a31dcea30ded6360a0db952b828f188fd05a60d9/crates/common_utils/src/id_type.rs#L182

crates/common_utils/src/id_type/global_id.rs
```diff
@@ -79,6 +79,23 @@ impl CellId {
...
+        Ok(Self(deserialized_length_id))
+    }
+}
+
+impl Default for CellId {
```
Narayanbhat166, line 92: do we need the default impl?

crates/router/src/configs/settings.rs
```diff
@@ -906,6 +912,11 @@ pub struct ServerTls {
     pub certificate: PathBuf,
 }
 
+#[cfg(feature = "v2")]
+#[derive(Debug, Clone, Deserialize, Default)]
+pub struct CellInformation {
```
Narayanbhat166, line 917: can you impl default on this struct rather than on the inner field struct?

crates/common_utils/src/id_type/global_id.rs
Sarthak1799: As the parent type requires it, we will need a default impl for this.

config/config.example.toml
```diff
@@ -731,18 +731,21 @@ connector_list = ""
 [recipient_emails]
 recon = "test@example.com"
 
+[cell_information]
+id = "12345"
```
SanchithHegde, line 735: Can we have documentation comments for these? Please add documentation comments in `config/deployments/env_specific.toml` as well.

crates/common_utils/src/id_type.rs
```diff
@@ -50,6 +52,13 @@ fn get_invalid_input_character(input_string: Cow<'static, str>) -> Option<char>
 /// A type for alphanumeric ids
 pub(crate) struct AlphaNumericId(String);
 
+impl Default for AlphaNumericId {
```
SanchithHegde, line 55: Let's avoid implementing / deriving `Default` on `AlphaNumericId` and `LengthId`.

crates/common_utils/src/id_type/global_id.rs
SanchithHegde, line 1: Can we update `CellId::from_str()` to accept `impl AsRef<str>` instead of `&str` and remove `CellId::from_string()` entirely?

crates/router/src/configs/settings.rs
Sarthak1799: Cannot be done as the internal types are private

crates/common_utils/src/id_type.rs
Sarthak1799: In order to derive Default on CellId, we need to be able to construct the internal types with some default value.
Since the internal types are defined as private, we can't use them directly, unless we use some conversion like `LengthId::from_alphanumeric_id` which performs checks and returns a result @SanchithHegde

crates/router/src/configs/settings.rs
```diff
@@ -941,6 +945,20 @@ pub struct ServerTls {
...
+#[derive(Debug, Clone, Deserialize, PartialEq, Eq)]
+pub struct CellInformation {
+    pub id: common_utils::id_type::CellId,
+}
+
+#[cfg(feature = "v2")]
+impl Default for CellInformation {
+    fn default() -> Self {
+        let cell_id = common_utils::id_type::CellId::from_str("abcde")
+            .expect("Failed to create a default for Cell Id");
```
SanchithHegde, line 958: 1. Nit: Can we have the default value to something along the lines of `default_cell_id`? To be more clear that it's a default...
2. Add `#[allow(clippy::expect_used)]` above this statement.

crates/router/src/configs/settings.rs
```diff
@@ -941,6 +945,21 @@ pub struct ServerTls {
...
+    pub id: common_utils::id_type::CellId,
+}
+
+#[cfg(feature = "v2")]
+impl Default for CellInformation {
+    fn default() -> Self {
+        #[allow(clippy::expect_used)]
+        let cell_id = common_utils::id_type::CellId::from_str("default_cell_id") // This is to provide a default cell id for application settings.
+            .expect("Failed to create a default for Cell Id"); // Ideally we're not meant to have a default for Cell Id, so a panic here is acceptable here.
+        Self { id: cell_id }
```
SanchithHegde, line 960: Nit:

```suggestion
        // We provide a static default cell id for constructing application settings.
        // This will only panic at application startup if we're unable to construct the default,
        // around the time of deserializing application settings.
        // And a panic at application startup is considered acceptable.
        #[allow(clippy::expect_used)]
        let cell_id = common_utils::id_type::CellId::from_str("default_cell_id") 
            .expect("Failed to create a default for Cell Id");
        Self { id: cell_id }
```

SanchithHegde (approved):

crates/router/src/configs/settings.rs
```diff
@@ -940,6 +944,26 @@ pub struct ServerTls {
...
+
+#[cfg(feature = "v2")]
+impl Default for CellInformation {
+    fn default() -> Self {
+        // We provide a static default cell id for constructing application settings.
+        // This will only panic at application startup if we're unable to construct the default,
+        // around the time of deserializing application settings.
+        // And a panic at application startup is considered acceptable.
+        #[allow(clippy::expect_used)]
+        let cell_id = common_utils::id_type::CellId::from_str("default_cell_id")
```
Narayanbhat166, line 961: This will fail the application startup because this value violates length constraints

SanchithHegde (approved):

Narayanbhat166 (approved):

PR #5932, opened 2024-09-18
feat(opensearch): restrict search view access based on user roles and permissions

tsdk02:
## Type of Change

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

Fixes the issue raised here: [https://github.com/juspay/hyperswitch-cloud/issues/6759](https://github.com/juspay/hyperswitch-cloud/issues/6759)

- Currently, a profile level user is able to see the other profile payments list when using global-search, irrespective of whether he has permissions or not. But the user will not be able to view the details related to the payments upon clicking the particular payment.

- This PR fixes this behaviour by restricting the profiles / merchants to be searched based on the user roles associated with the role_id and the permissions associated with the user role

- Now, only if the use role has the necessary READ permissions to access the indexes, he would be able to search the payments related to that particular profile/merchant.

- The `search_params` will now be constructed with only those ProfileLevel / MerchantLevel / OrgLevel entities which will be searched through the opensearch query.
### Additional Changes

## Motivation and Context

To preserve confidentiality of the payment details

## How did you test it?

Try to search payments of a profile from another profile which doesn't have access to the profile where the payments are present. Results should not show up while using global search.

## Checklist


crates/router/src/analytics.rs
```diff
@@ -1838,25 +1843,104 @@ pub mod routes {
...
+                        let role_id = user_role.role_id.clone();
+                        let org_id = user_role.org_id.clone().unwrap_or_default();
+                        async move {
+                            RoleInfo::from_role_id_in_org_scope(&state, &role_id, &org_id)
+                                .await
+                                .change_context(UserErrors::InternalServerError)
+                                .change_context(OpenSearchError::UnknownError)
+                                .map(|role_info| (role_id, role_info))
+                        }
+                    })
```
ThisIsMani, line 1882: Please evaluate futures::join_all. You can solve this issues because of move.

crates/router/src/analytics.rs
```diff
@@ -1838,25 +1843,104 @@ pub mod routes {
...
                     .collect();
 
-                let merchant_id: MerchantId = auth.merchant_id;
-                let org_id: OrganizationId = auth.org_id;
-                let search_params: Vec<AuthInfo> = vec![AuthInfo::MerchantLevel {
-                    org_id: org_id.clone(),
-                    merchant_ids: vec![merchant_id.clone()],
-                }];
+                let mut search_params: Vec<AuthInfo> = Vec::new();
+                for user_role in &filtered_user_roles {
```
ThisIsMani, line 1903:
```suggestion
                for user_role in filtered_user_roles.iter() {
```

ThisIsMani: You can also use `iter().map()` instead.

lsampras (approved):

PR #5971, opened 2024-09-19
feat(router): add api_models and openapi changes for payments create intent api for v2

sai-harsha-vardhan:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

add api_models and openapi changes for payments create intent api for v2

### Additional Changes

## How did you test it?

API model changes only - Tested mintlify docs locally
image

## Checklist


crates/api_models/src/payments.rs
```diff
@@ -114,6 +114,286 @@ pub struct CustomerDetailsResponse {
...
+    /// Additional data that might be required by hyperswitch based on the requested features by the merchants.
+    pub feature_metadata: Option<FeatureMetadata>,
+
+    /// Whether to generate the payment link for this payment or not (if applicable)
+    #[schema(value_type = Option<EnablePaymentLinkRequest>)]
+    #[serde(default)]
+    pub payment_link_enabled: Option<common_enums::EnablePaymentLinkRequest>,
+
+    #[schema(value_type = Option<PaymentCreatePaymentLinkConfig>)]
+    pub payment_link_config: Option<PaymentCreatePaymentLinkConfig>,
```
SanchithHegde, line 218: Can we add doc comments for all fields without any?

crates/api_models/src/payments.rs
```diff
@@ -114,6 +114,286 @@ pub struct CustomerDetailsResponse {
...
+
+    /// Custom payment link config id set at business profile, send only if business_specific_configs is configured
+    pub payment_link_config_id: Option<String>,
```
SanchithHegde, line 221: Do we need both `payment_link_config` and `payment_link_config_id` to be specified?
Also, there doesn't seem to be a `business_specific_configs` field?

crates/api_models/src/payments.rs
```diff
@@ -114,6 +114,286 @@ pub struct CustomerDetailsResponse {
...
+    #[serde(default, deserialize_with = "amount::deserialize")]
+    pub amount: Amount,
+
+    /// The three letter ISO currency code in uppercase. Eg: 'USD' to charge US Dollars
+    #[schema(example = "USD", value_type = Currency)]
+    pub currency: api_enums::Currency,
+
+    /// The business profile to be used for this payment, if not passed the default business profile associated with the merchant account will be used. It is mandatory in case multiple business profiles have been set up.
+    #[schema(value_type = String)]
+    pub profile_id: id_type::ProfileId,
```
SanchithHegde, line 132: We require a profile ID to be always passed, right? We don't consider default business profile in v2?

crates/api_models/src/payments.rs
```diff
@@ -114,6 +114,286 @@ pub struct CustomerDetailsResponse {
...
+
+    /// Unique identifier for the payment. This ensures idempotency for multiple payments
+    /// that have been done by a single merchant.
+    #[schema(
+        value_type = Option<String>,
+        min_length = 30,
+        max_length = 30,
+        example = "pay_mbabizu24mvu3mela5njyhpit4"
+    )]
+    pub merchant_reference_id: Option<id_type::PaymentId>,
```
jarnura, line 142: We should name this as `PaymentReferenceId` to avoid confusion, since payment.id and PaymentId very similar but does different purpose

crates/api_models/src/payments.rs
```diff
@@ -114,6 +114,286 @@ pub struct CustomerDetailsResponse {
...
+
+    /// Details of the routing configuration for that payment
+    #[schema(value_type = Option<String>)]
+    pub routing_algorithm_id: Option<String>,
+
+    #[schema(value_type = Option<CaptureMethod>, example = "automatic")]
+    pub capture_method: Option<api_enums::CaptureMethod>,
```
jarnura, line 149: Add doc comments

crates/api_models/src/payments.rs
```diff
@@ -114,6 +114,286 @@ pub struct CustomerDetailsResponse {
...
+
+    #[schema(value_type = Option<AuthenticationType>, example = "no_three_ds", default = "three_ds")]
+    pub authentication_type: Option<api_enums::AuthenticationType>,
```
jarnura, line 152: if we have default value then we can make this non-optional value

crates/api_models/src/payments.rs
```diff
@@ -114,6 +114,286 @@ pub struct CustomerDetailsResponse {
...
+
+    /// The billing details of the payment. This address will be used for invoicing.
+    pub billing: Option<Address>,
+
+    /// The shipping address for the payment
+    pub shipping: Option<Address>,
+
+    /// The identifier for the customer
+    #[schema(value_type = Option<String>, max_length = 64, min_length = 1, example = "cus_y3oqhf46pyzuxjbcn2giaqnb44")]
+    pub customer_id: Option<id_type::CustomerId>,
```
jarnura, line 162: Make this a Enum,

```
enum CustomerId {
	Guest,
	Customer(Box<id_type::CustomerId>)
}
```

So here None, means it guest so, make this a explicit information in type rather implicitly handling these in code

crates/api_models/src/payments.rs
```diff
@@ -114,6 +114,286 @@ pub struct CustomerDetailsResponse {
...
+
+    /// Set to true to indicate that the customer is not in your checkout flow during this payment, and therefore is unable to authenticate. This parameter is intended for scenarios where you collect card details and charge them later. When making a recurring payment by passing a mandate_id, this parameter is mandatory
```
jarnura, line 164: in v2 we are not going to handle mandate_id related flow, fix the comment.

If we make customer_present has a default value and the default is true, so that we can remove Option

crates/api_models/src/payments.rs
```diff
@@ -114,6 +114,286 @@ pub struct CustomerDetailsResponse {
...
+    #[schema(example = true, value_type = PresenceOfCustomerDuringPayment)]
+    #[serde(default)]
+    pub customer_present: Option<common_enums::PresenceOfCustomerDuringPayment>,
+
+    /// A description for the payment
+    #[schema(example = "It's my first payment request")]
+    pub description: Option<String>,
```
jarnura, line 171: We have domain type for description, `Description`, Use `Description` instead of string

crates/api_models/src/payments.rs
```diff
@@ -114,6 +114,286 @@ pub struct CustomerDetailsResponse {
...
+    pub setup_future_usage: Option<api_enums::FutureUsage>,
+
+    /// Apply MIT exemption for a payment
+    #[schema(value_type = MitExemptionRequest)]
+    #[serde(default)]
+    pub apply_mit_exemption: common_enums::MitExemptionRequest,
+
+    /// For non-card charges, you can use this value as the complete description that appears on your customers’ statements. Must contain at least one letter, maximum 22 characters.
+    #[schema(max_length = 255, example = "Hyperswitch Router")]
+    pub statement_descriptor: Option<String>,
```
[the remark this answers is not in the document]
jarnura: Create a Domain type like Description, basically a `StatementDescriptor`.

crates/api_models/src/payments.rs
```diff
@@ -114,6 +114,286 @@ pub struct CustomerDetailsResponse {
...
+    }]"#)]
+    pub order_details: Option<Vec<OrderDetailsWithAmount>>,
+
+    /// Use this parameter to restrict the Payment Method Types to show for a given PaymentIntent
+    #[schema(value_type = Option<Vec<PaymentMethodType>>)]
+    pub allowed_payment_method_types: Option<Vec<api_enums::PaymentMethodType>>,
+
+    /// You can specify up to 50 keys, with key names up to 40 characters long and values up to 500 characters long. Metadata is useful for storing additional, structured information on an object.
+    #[schema(value_type = Option<Object>, example = r#"{ "udf1": "some-value", "udf2": "some-value" }"#)]
+    pub metadata: Option<serde_json::Value>,
```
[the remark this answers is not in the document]
jarnura: Make Secret Serde Value for `unstructured` types

crates/api_models/src/payments.rs
```diff
@@ -114,6 +114,286 @@ pub struct CustomerDetailsResponse {
...
+
+    ///Request an incremental authorization, i.e., increase the authorized amount on a confirmed payment before you capture it.
+    #[schema(value_type = RequestIncrementalAuthorization)]
+    #[serde(default)]
+    pub request_incremental_authorization: common_enums::RequestIncrementalAuthorization,
+
+    ///Will be used to expire client secret after certain amount of time to be supplied in seconds
+    ///(900) for 15 mins
+    #[schema(example = 900)]
+    pub session_expiry: Option<u32>,
```
jarnura, line 234: Instead of Option we can handle this with default value

jarnura (changes requested):

crates/api_models/src/payments.rs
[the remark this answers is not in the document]
sai-harsha-vardhan: Documentation is already available for AuthenticationType enum, which would be used while constructing openapi_spec

crates/api_models/src/payments.rs
```diff
@@ -114,6 +114,286 @@ pub struct CustomerDetailsResponse {
...
+    /// A description for the payment
+    #[schema(example = "It's my first payment request")]
+    pub description: Option<String>,
+
+    /// The URL to which you want the user to be redirected after the completion of the payment operation
+    #[schema(value_type = Option<String>, example = "https://hyperswitch.io")]
+    pub return_url: Option<Url>,
+
+    #[schema(value_type = Option<FutureUsage>, example = "off_session")]
+    pub setup_future_usage: Option<api_enums::FutureUsage>,
```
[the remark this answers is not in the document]
sai-harsha-vardhan: Documentation is already available for AuthenticationType enum, which would be used while constructing openapi_spec

crates/api_models/src/payments.rs
sai-harsha-vardhan: this will have default a value which can taken from profiles, so in api_models it is optional but in domain models it's mandatory

crates/api_models/src/payments.rs
sai-harsha-vardhan: We can have this in domain model right?

jarnura: yes

crates/api_models/src/payments.rs
```diff
@@ -114,6 +114,298 @@ pub struct CustomerDetailsResponse {
...
+    pub billing: Option<Address>,
+
+    /// The shipping address for the payment
+    pub shipping: Option<Address>,
+
+    /// The identifier for the customer
+    #[schema(value_type = Option<String>, max_length = 64, min_length = 1, example = "cus_y3oqhf46pyzuxjbcn2giaqnb44")]
+    pub customer_id: Option<id_type::CustomerId>,
+
+    /// Set to true to indicate that the customer is not in your checkout flow during this payment, and therefore is unable to authenticate. This parameter is intended for scenarios where you collect card details and charge them later. When making a recurring payment by passing a mandate_id, this parameter is mandatory
```
jarnura, line 156:
```suggestion
    /// Set to true to indicate that the customer is in your checkout flow during this payment, and therefore is able to authenticate. This parameter should be false when merchant's doing merchant initiated payments and customer is not present while doing the payment.
```

crates/api_models/src/payments.rs
jarnura: Make this mandatory, with the doc comment saying what we take as default if the merchant configured in profile and if he is not configured default is 15mins.

jarnura (changes requested):

jarnura (approved):

ShankarSinghC (approved):

crates/api_models/src/payments.rs
```diff
@@ -114,6 +114,298 @@ pub struct CustomerDetailsResponse {
...
+    pub payment_method_type: Option<PaymentMethodTypeTax>,
+}
+
+#[derive(Clone, Debug, PartialEq, Serialize, Deserialize, ToSchema)]
+pub struct PaymentMethodTypeTax {
+    /// The order tax amount for the payment method type
+    pub order_tax_amount: MinorUnit,
+    /// The payment method type
+    #[schema(value_type = PaymentMethodType, example = "google_pay")]
+    pub pmt: common_enums::PaymentMethodType,
```
SanchithHegde, line 400: Can we please avoid abbreviating things in API models? Can we name this as `payment_method_type`?

CC: @jarnura @Narayanbhat166

sai-harsha-vardhan: will take this up in next PR

PR #6109, opened 2024-09-26
feat(worldpay): migrate to v7

kashif-m:
## Type of Change

- [x] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description
-

### Additional Changes

## Checklist

### DIFF
--- a/crates/connector_configs/toml/development.toml
+++ b/crates/connector_configs/toml/development.toml
@@ -3384,9 +3384,10 @@ merchant_secret="Source verification key"
   payment_method_type = "google_pay"
 [[worldpay.wallet]]
   payment_method_type = "apple_pay"
-[worldpay.connector_auth.BodyKey]
+[worldpay.connector_auth.SignatureKey]
 api_key="Username"
 key1="Password"
+api_secret="Merchant Identifier"
 [worldpay.connector_webhook_details]
 merchant_secret="Source verification key"
 
--- a/crates/router/src/connector/worldpay.rs
+++ b/crates/router/src/connector/worldpay.rs
@@ -42,10 +42,16 @@ where
         req: &types::RouterData<Flow, Request, Response>,
         _connectors: &settings::Connectors,
     ) -> CustomResult<Vec<(String, request::Maskable<String>)>, errors::ConnectorError> {
-        let mut headers = vec![(
-            headers::CONTENT_TYPE.to_string(),
-            self.get_content_type().to_string().into(),
-        )];
+        let mut headers = vec![
+            (
+                headers::ACCEPT.to_string(),
+                self.get_content_type().to_string().into(),
+            ),
+            (
+                headers::CONTENT_TYPE.to_string(),
+                self.get_content_type().to_string().into(),
+            ),
+        ];
         let mut api_key = self.get_auth_header(&req.connector_auth_type)?;
         headers.append(&mut api_key);
         Ok(headers)
@@ -62,7 +68,7 @@ impl ConnectorCommon for Worldpay {
     }
 
     fn common_get_content_type(&self) -> &'static str {
-        "application/vnd.worldpay.payments-v6+json"
+        "application/vnd.worldpay.payments-v7+json"
     }
 
     fn base_url<'a>(&self, connectors: &'a settings::Connectors) -> &'a str {
@@ -86,10 +92,14 @@ impl ConnectorCommon for Worldpay {
         res: Response,
         event_builder: Option<&mut ConnectorEvent>,
     ) -> CustomResult<ErrorResponse, errors::ConnectorError> {
-        let response: WorldpayErrorResponse = res
-            .response
-            .parse_struct("WorldpayErrorResponse")
-            .change_context(errors::ConnectorError::ResponseDeserializationFailed)?;
+        router_env::logger::info!(raw_response=?res);
+        let response = if res.status_code != 404 {
+            res.response
+                .parse_struct("WorldpayErrorResponse")
+                .change_context(errors::ConnectorError::ResponseDeserializationFailed)?
+        } else {
+            WorldpayErrorResponse::default()
+        };
 
         event_builder.map(|i| i.set_error_response_body(&response));
         router_env::logger::info!(connector_response=?response);
@@ -183,7 +193,7 @@ impl ConnectorIntegration<api::Void, types::PaymentsCancelData, types::PaymentsR
     ) -> CustomResult<String, errors::ConnectorError> {
         let connector_payment_id = req.request.connector_transaction_id.clone();
         Ok(format!(
-            "{}payments/settlements/{}",
+            "{}payments/authorizations/cancellations/{}",
             self.base_url(connectors),
             connector_payment_id
         ))
@@ -361,6 +371,28 @@ impl ConnectorIntegration<api::Capture, types::PaymentsCaptureData, types::Payme
         self.common_get_content_type()
     }
 
+    fn get_url(
+        &self,
+        req: &types::PaymentsCaptureRouterData,
+        connectors: &settings::Connectors,
+    ) -> CustomResult<String, errors::ConnectorError> {
+        let connector_payment_id = req.request.connector_transaction_id.clone();
+        Ok(format!(
+            "{}payments/settlements/partials/{}",
+            self.base_url(connectors),
+            connector_payment_id
+        ))
+    }
+
+    fn get_request_body(
+        &self,
+        req: &types::PaymentsCaptureRouterData,
+        _connectors: &settings::Connectors,
+    ) -> CustomResult<RequestContent, errors::ConnectorError> {
+        let connector_req = WorldpayPartialRequest::try_from(req)?;
+        Ok(RequestContent::Json(Box::new(connector_req)))
+    }
+
     fn build_request(
         &self,
         req: &types::PaymentsCaptureRouterData,
@@ -372,6 +404,9 @@ impl ConnectorIntegration<api::Capture, types::PaymentsCaptureData, types::Payme
                 .url(&types::PaymentsCaptureType::get_url(self, req, connectors)?)
                 .attach_default_headers()
                 .headers(types::PaymentsCaptureType::get_headers(
+                    self, req, connectors,
+                )?)
+                .set_body(types::PaymentsCaptureType::get_request_body(
                     self, req, connectors,
                 )?)
                 .build(),
@@ -409,19 +444,6 @@ impl ConnectorIntegration<api::Capture, types::PaymentsCaptureData, types::Payme
             }
             _ => Err(errors::ConnectorError::ResponseHandlingFailed)?,
         }
-    }
-
-    fn get_url(
-        &self,
-        req: &types::PaymentsCaptureRouterData,
-        connectors: &settings::Connectors,
-    ) -> CustomResult<String, errors::ConnectorError> {
-        let connector_payment_id = req.request.connector_transaction_id.clone();
-        Ok(format!(
-            "{}payments/settlements/{}",
-            self.base_url(connectors),
-            connector_payment_id
-        ))
     }
 
     fn get_error_response(
@@ -463,7 +485,7 @@ impl ConnectorIntegration<api::Authorize, types::PaymentsAuthorizeData, types::P
         connectors: &settings::Connectors,
     ) -> CustomResult<String, errors::ConnectorError> {
         Ok(format!(
-            "{}payments/authorizations",
+            "{}cardPayments/customerInitiatedTransactions",
             self.base_url(connectors)
         ))
     }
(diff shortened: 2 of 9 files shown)


crates/router/src/connector/worldpay.rs
```diff
@@ -86,10 +92,14 @@ impl ConnectorCommon for Worldpay {
...
-            .response
-            .parse_struct("WorldpayErrorResponse")
-            .change_context(errors::ConnectorError::ResponseDeserializationFailed)?;
+        router_env::logger::info!(raw_response=?res);
+        let response = if res.status_code != 404 {
+            res.response
+                .parse_struct("WorldpayErrorResponse")
+                .change_context(errors::ConnectorError::ResponseDeserializationFailed)?
+        } else {
+            WorldpayErrorResponse::default()
```
srujanchikke, line 101: Can we not default in status mappings?

crates/router/src/connector/worldpay.rs
```diff
@@ -183,7 +193,7 @@ impl ConnectorIntegration<api::Void, types::PaymentsCancelData, types::PaymentsR
     ) -> CustomResult<String, errors::ConnectorError> {
         let connector_payment_id = req.request.connector_transaction_id.clone();
         Ok(format!(
-            "{}payments/settlements/{}",
+            "{}payments/authorizations/cancellations/{}",
             self.base_url(connectors),
             connector_payment_id
         ))
```
srujanchikke, line 199:
```suggestion
            "{}payments/authorizations/cancellations/{connector_payment_id}",
            self.base_url(connectors)
        ))
```

crates/router/src/connector/worldpay/response.rs
```diff
@@ -305,14 +260,24 @@ impl PaymentsResponseScheme {
...
+impl Default for WorldpayErrorResponse {
+    fn default() -> Self {
+        Self {
+            error_name: "Not found".to_string(),
+            message: "Resource not found".to_string(),
+            validation_errors: None,
+        }
+    }
+}
+
```
srujanchikke, line 280: why do we need this ?

crates/router/src/connector/worldpay/transformers.rs
```diff
@@ -47,11 +47,12 @@ fn fetch_payment_instrument(
...
     match payment_method {
         domain::PaymentMethodData::Card(card) => Ok(PaymentInstrument::Card(CardPayment {
-            card_expiry_date: CardExpiryDate {
+            expiry_date: ExpiryDate {
                 month: utils::CardData::get_expiry_month_as_i8(&card)?,
                 year: utils::CardData::get_expiry_year_as_i32(&card)?,
             },
             card_number: card.card_number,
+            cvc: card.card_cvc,
             ..CardPayment::default()
```
srujanchikke, line 56: can we not have default values for payment method data ?

crates/router/src/connector/worldpay/transformers.rs
```diff
@@ -271,11 +294,24 @@ impl TryFrom<types::PaymentsResponseRouterData<WorldpayPaymentsResponse>>
...
 
-impl<F> TryFrom<&types::RefundsRouterData<F>> for WorldpayRefundRequest {
+impl TryFrom<&types::PaymentsCaptureRouterData> for WorldpayPartialRequest {
+    type Error = error_stack::Report<errors::ConnectorError>;
+    fn try_from(item: &types::PaymentsCaptureRouterData) -> Result<Self, Self::Error> {
+        Ok(Self {
+            reference: item.payment_id.clone().replace("_", "-"),
+            value: PaymentValue {
+                amount: item.request.amount_to_capture,
+                currency: item.request.currency.to_string(),
```
srujanchikke, line 304: please change the type of currency to Currency enums

crates/router/src/connector/worldpay/transformers.rs
```diff
@@ -271,11 +294,24 @@ impl TryFrom<types::PaymentsResponseRouterData<WorldpayPaymentsResponse>>
...
 }
 
-impl<F> TryFrom<&types::RefundsRouterData<F>> for WorldpayRefundRequest {
+impl TryFrom<&types::PaymentsCaptureRouterData> for WorldpayPartialRequest {
+    type Error = error_stack::Report<errors::ConnectorError>;
+    fn try_from(item: &types::PaymentsCaptureRouterData) -> Result<Self, Self::Error> {
+        Ok(Self {
+            reference: item.payment_id.clone().replace("_", "-"),
+            value: PaymentValue {
+                amount: item.request.amount_to_capture,
```
srujanchikke, line 303: amount should be consumed for connector routerdata. Please convert the amount before passing it to transformers.
```suggestion
                amount: item.amount_to_capture,
```

migrations/2024-09-25-113851_increase_connector_transaction_id_length_in_payment_and_refund/down.sql
```diff
@@ -0,0 +1,8 @@
+ALTER TABLE payment_attempt
+ALTER COLUMN connector_transaction_id TYPE VARCHAR(128);
+
+ALTER TABLE refund
+ALTER COLUMN connector_transaction_id TYPE VARCHAR(128);
+
+ALTER TABLE refund
+ALTER COLUMN connector_refund_id TYPE VARCHAR(128);
```
srujanchikke, line 8: Could you please check this ?
cc : @SanchithHegde  @jarnura

srujanchikke (changes requested):
Can we also add pm filters & cypress for worldpay ?

crates/connector_configs/toml/development.toml
```diff
@@ -3384,9 +3384,10 @@ merchant_secret="Source verification key"
   payment_method_type = "google_pay"
 [[worldpay.wallet]]
   payment_method_type = "apple_pay"
-[worldpay.connector_auth.BodyKey]
+[worldpay.connector_auth.SignatureKey]
```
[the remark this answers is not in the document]
kashif-m: Not required as of now! I have handled both `BodyKey` and `SignatureKey` in application

crates/router/src/connector/worldpay.rs
kashif-m: They send an empty body for 404 status codes .. we need to conditionally parse due to this behaviour or else it would throw derserialization errors leading to 5xx on our end

Making every field optional in error response also wouldn't work since it would expect `{}` but WorldPay empty body ``

srujanchikke (approved):
reviewed connector code.

crates/diesel_models/src/query/capture.rs
```diff
@@ -67,3 +67,22 @@ impl Capture {
         .await
     }
 }
+
+impl ConnectorTransactionIdTrait for Capture {
```
kashif-m, line 71: move to diesel_models/capture

crates/common_utils/src/types.rs
```diff
@@ -1247,3 +1247,125 @@ where
...
+                id
+            ))),
+        }
+    }
+}
+
+impl From<String> for ConnectorTransactionId {
+    fn from(src: String) -> Self {
+        // ID already hashed
+        if src.starts_with("hash_") {
```
kashif-m, line 1305: Is this good enough?

apoorvdixit88 (approved):
Dashboard changes look fine.

migrations/2024-09-25-113851_increase_connector_transaction_id_length_in_payment_and_refund/up.sql
```diff
@@ -0,0 +1,11 @@
+ALTER TABLE payment_attempt
+ADD COLUMN IF NOT EXISTS connector_transaction_data VARCHAR(512);
+
+ALTER TABLE refund
+ADD COLUMN IF NOT EXISTS connector_refund_data VARCHAR(512);
+
+ALTER TABLE refund
```
Narayanbhat166, line 7: why do we need to store this? isn't this already stored in payment attempt?

Narayanbhat166:
Hey @kashif-m, what do you think about this change not leaking from the domain model? This is only so that we have limitations with the data modelling with postgres right? once we move to a different approach, we can try optimizing the storage layer, but the application logic should not be affected right.

So I was thinking if we can have this change made only at the database level, i.e postgres for now. we can have two columns in payment attempt, but the domain model will only have one column called `connector_transaction_id`. What is your opinion on this?

migrations/2024-09-25-113851_increase_connector_transaction_id_length_in_payment_and_refund/up.sql
kashif-m: Capture and Refund IDs for WP follows the same format. These IDs can differ across different resources (payments, refunds, captures) - they don't for WP, but better to store both the data and hashed values in the tables.

kashif-m:
> Hey @kashif-m, what do you think about this change not leaking from the domain model? This is only so that we have limitations with the data modelling with postgres right? once we move to a different approach, we can try optimizing the storage layer, but the application logic should not be affected right.
> 
> So I was thinking if we can have this change made only at the database level, i.e postgres for now. we can have two columns in payment attempt, but the domain model will only have one column called `connector_transaction_id`. What is your opinion on this?

I was thinking the same! We can introduce a new struct for the application (let's say, `PaymentAttempt`), and transform the struct bound to the table (let's say `PaymentAttemptTable`). Operations can only be performed on `PaymentAttempt`, which would internally convert to `PaymentAttemptTable` for querying the DB.

Currently, it uses getter functions to fetch the correct IDs (`get_connector_payment_id` - present in PaymentAttempt or `get_optional_connector_transaction_id` - can be used on domain type `ConnectorTransactionId`).

config/deployments/sandbox.toml
```diff
@@ -339,7 +339,9 @@ upi_collect = {country = "IN", currency = "INR"}
 [pm_filters.plaid]
 open_banking_pis = {currency = "EUR,GBP"}
 
-[pm_filters.worldpay]
+[pm_filters]
```
pixincreate, line 342: this is supposed to be `[pm_filters.worldpay]`?

pixincreate (commented):
cypress changes looks good.
deployment env changes, i've a doubt w.r.t sandbox.toml

crates/hyperswitch_domain_models/src/router_request_types.rs
```diff
@@ -439,6 +441,8 @@ pub struct PaymentsSyncData {
 
     pub amount: MinorUnit,
     pub integrity_object: Option<SyncIntegrityObject>,
+    pub attempt_status: AttemptStatus,
+    pub intent_status: IntentStatus,
```
Narayanbhat166, line 445: why do we need these details at the connector level? we already have attempt status in router_data.status

kashif-m: Removed!

crates/router/src/core/payments/operations/payment_response.rs
```diff
@@ -1496,12 +1497,23 @@ async fn payment_response_update_tracker<F: Clone, T: types::Capturable>(
...
+                                                    ConnectorTransactionId::form_id_and_data(id);
+                                                (Some(txn_id), txn_data)
+                                            }
+                                        };
                                     let capture_update = storage::CaptureUpdate::ResponseUpdate {
                                         status: enums::CaptureStatus::foreign_try_from(
                                             router_data.status,
                                         )?,
-                                        connector_capture_id: connector_transaction_id.clone(),
+                                        connector_capture_id: connector_capture_id.clone(),
```
Narayanbhat166, line 1514: we are not storing connector transaction id here anymore?

Narayanbhat166 (approved):

crates/router/src/core/payments/operations/payment_response.rs
kashif-m: It's the same ID, added a new variable since we're storing ID + optional hashed data

crates/common_utils/src/types.rs
```diff
@@ -1247,3 +1247,129 @@ where
...
+        match txn_id {
+            Self::TxnId(_) => (txn_id, None),
+            Self::HashedData(_) => (txn_id, Some(src)),
+        }
+    }
+
+    /// Implementation for retrieving
+    pub fn get_txn_id<'a>(
+        &'a self,
+        txn_data: Option<&'a String>,
```
NishantJoshi00, line 1286: nit: this could have been `Option<&'a str>`

crates/common_utils/src/types.rs
```diff
@@ -1247,3 +1247,129 @@ where
...
+    Self: FromSql<sql_types::Text, DB>,
+{
+    type Row = Self;
+
+    fn build(row: Self::Row) -> deserialize::Result<Self> {
+        Ok(row)
+    }
+}
+
+impl<DB> FromSql<sql_types::Text, DB> for ConnectorTransactionId
```
NishantJoshi00, line 1334: There are macros fro this right @Narayanbhat166

NishantJoshi00 (approved):

crates/common_utils/src/types.rs
kashif-m: Will update in the subsequent PRs

srujanchikke (approved):

Aprabhat19 (approved):

pixincreate (approved):

racnan (approved):

PR #6252, opened 2024-10-07
refactor(router): modify `net_amount` to be a struct in the domain model of payment_attempt and handle amount changes across all flows

sai-harsha-vardhan:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [x] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

modify `net_amount` to be a struct in the domain model of payment_attempt and handle amount changes across all flows
1. modify `net_amount` to be a struct and move order_amount, shipping_cost, order_tax_amount, surcharge_amount and tax_on_surcharge into NetAmount
2. use total_amount of net_amount instead of using payment_attempt.amount in all the subsequent flows like captures, refunds etc.
3. use net_amount.total_amount instead of intent amount wherever it's needed

### Additional Changes

## How did you test it?

Tested Manually

## Checklist


crates/hyperswitch_domain_models/src/payments/payment_attempt.rs
```diff
@@ -334,8 +331,95 @@ pub struct PaymentAttempt {
...
+        self.order_amount = order_amount;
+    }
+
+    pub fn set_order_tax_amount(&mut self, order_tax_amount: Option<MinorUnit>) {
+        self.order_tax_amount = order_tax_amount;
+    }
+
+    pub fn set_surcharge_details(
+        &mut self,
+        surcharge_details: Option<crate::router_request_types::SurchargeDetails>,
```
jarnura, line 416:
```suggestion
        surcharge_details: Option<router_request_types::SurchargeDetails>,
```

crates/diesel_models/src/payment_attempt.rs
```diff
@@ -303,9 +290,10 @@ impl PaymentAttemptNew {
     /// returns amount + surcharge_amount + tax_amount (surcharge) + shipping_cost + order_tax_amount
     pub fn calculate_net_amount(&self) -> MinorUnit {
```
jarnura, line 291: Why this function is required when we have `get_total_amount` from `NetAmount`

crates/hyperswitch_domain_models/src/router_request_types.rs
```diff
@@ -546,14 +546,16 @@ impl
...
-            original_amount: payment_attempt.amount,
+            original_amount: payment_attempt.net_amount.get_order_amount(),
             surcharge: common_utils::types::Surcharge::Fixed(
                 request_surcharge_details.surcharge_amount,
             ),
             tax_on_surcharge: None,
             surcharge_amount,
             tax_on_surcharge_amount,
-            final_amount: payment_attempt.amount + surcharge_amount + tax_on_surcharge_amount,
+            final_amount: payment_attempt.net_amount.get_order_amount()
```
jarnura, line 556: Why we are not considering `get_total_amount` here? Reason for again doing calculation here.

crates/router/src/core/payment_methods/cards.rs
```diff
@@ -2915,7 +2915,7 @@ pub async fn list_payment_methods(
         .transpose()?;
     let setup_future_usage = payment_intent.as_ref().and_then(|pi| pi.setup_future_usage);
     let payment_type = payment_attempt.as_ref().map(|pa| {
-        let amount = api::Amount::from(pa.amount);
+        let amount = api::Amount::from(pa.net_amount.get_order_amount());
```
jarnura, line 2918: Only order amount needs to be used here, we can refer intent amount itself.

crates/router/src/core/payments.rs
```diff
@@ -854,12 +860,17 @@ pub async fn call_surcharge_decision_management_for_session_flow(
...
 ) -> RouterResult<Option<api::SessionSurchargeDetails>> {
-    if let Some(surcharge_amount) = payment_attempt.surcharge_amount {
-        let tax_on_surcharge_amount = payment_attempt.tax_amount.unwrap_or_default();
-        let final_amount = payment_attempt.amount + surcharge_amount + tax_on_surcharge_amount;
+    if let Some(surcharge_amount) = payment_attempt.net_amount.get_surcharge_amount() {
+        let tax_on_surcharge_amount = payment_attempt
+            .net_amount
+            .get_tax_on_surcharge()
+            .unwrap_or_default();
+        let final_amount = payment_attempt.net_amount.get_order_amount()
```
jarnura, line 868: Why we are calculation net_amount instead of using `get_total_amount` function?

crates/router/src/core/payments/helpers.rs
```diff
@@ -900,22 +917,32 @@ pub fn validate_amount_to_capture_and_capture_method(
...
             })
             .unwrap_or_default();
-        let total_capturable_amount =
-            original_amount.map(|original_amount| original_amount + surcharge_amount);
+        let shipping_cost = request
+            .shipping_cost
+            .or(payment_attempt
+                .and_then(|payment_attempt| payment_attempt.net_amount.get_shipping_cost()))
+            .unwrap_or_default();
+        let total_capturable_amount = original_amount
```
jarnura, line 944: Why we are calculating this instead of using `get_total_amount`?

crates/router/src/core/payments/operations/payments_incremental_authorization.rs
```diff
@@ -106,6 +100,12 @@ impl<F: Send + Clone>
             .await
             .to_not_found_response(errors::ApiErrorResponse::PaymentNotFound)?;
 
+        if payment_attempt.get_total_amount() > request.amount {
```
jarnura, line 103: This can't be true with `get_total_amount`,  because `request.amount` is a `order_amount` which can be lower than the get_total_amount. Or In incremental_authotization they can pass total amount in request?

If genuine cases also this can throw errors.

crates/router/src/core/payments/routing.rs
```diff
@@ -232,7 +232,7 @@ pub fn make_dsl_input(
     };
 
     let payment_input = dsl_inputs::PaymentInput {
-        amount: payments_dsl_input.payment_intent.amount,
+        amount: payments_dsl_input.payment_attempt.get_total_amount(),
```
jarnura, line 235: @Aprabhat19  Is routing need total_amount or just order_amount? In other places dsl_inputs do we need total amount?

crates/router/src/core/payments/transformers.rs
```diff
@@ -703,7 +703,12 @@ where
         .as_ref()
         .get_required_value("currency")?;
     let amount = currency
```
jarnura, line 705: Is this is total amount case or order_amount, why we are sending order amount, is this should be total_amount?

jarnura (changes requested):

crates/diesel_models/src/payment_attempt.rs
sai-harsha-vardhan: This `impl` is on diesel models of PaymentAttemptNew, in which net_amount is still a MinorUnit

crates/hyperswitch_domain_models/src/router_request_types.rs
sai-harsha-vardhan: We are trying to calculate surcharge_details from the RequestSurchargeDetails, net_amount is not computed yet. That's why we are getting surcharge details from RequestSurchargeDetails and calculating SurchargeDetails

crates/router/src/core/payment_methods/cards.rs
sai-harsha-vardhan: yes, but here it's inside payment_attempt's map block. payment_intent is also option here

crates/router/src/core/payments/operations/payment_confirm.rs
```diff
@@ -1340,8 +1331,11 @@ impl<F: Clone> UpdateTracker<F, PaymentData<F>, api::PaymentsRequest> for Paymen
                         client_source,
                         client_version,
                         customer_acceptance: payment_data.payment_attempt.customer_acceptance,
-                        shipping_cost,
-                        order_tax_amount,
+                        shipping_cost: payment_data.payment_intent.shipping_cost,
```
Narayanbhat166, line 1334: we need not have these fields in the flattened format in update right? can we use the amount struct in update as well?

crates/router/src/core/payments/helpers.rs
sai-harsha-vardhan: This is being called in validate_request flow also, at that point payment_attempt is not present yet. So, we have to take details from request and calculate the amount

crates/router/src/core/payments/operations/payments_incremental_authorization.rs
sai-harsha-vardhan: In incremental_authotization, request.amount is total_amount which is on top of net_amount (not order_amount here)

crates/router/src/core/payments/operations/payment_create.rs
```diff
@@ -694,6 +694,10 @@ impl<F: Clone + Send> Domain<F, api::PaymentsRequest, PaymentData<F>> for Paymen
                 }),
                 payment_method_type: None,
             });
+            payment_data
```
Narayanbhat166, line 697: we should not update payment attempt when calculating the tax, since this operation is not dependent on payment attempt. @swangi-kumari can you please confirm this once?

crates/router/src/core/payments/routing.rs
sai-harsha-vardhan: Ideally we should be doing on net_amount right? @Aprabhat19

crates/router/src/core/payments/operations/payment_create.rs
sai-harsha-vardhan: Yes ideally it shouldn't depend on payment attempt but in v1, we are storing tax calculation details in net_amount of payment_attempt. In v2, we can have different way to handle this

crates/router/src/core/payments/operations/payment_create.rs
```diff
@@ -1163,16 +1166,20 @@ impl PaymentCreate {
...
                     .and_then(|inner| inner.mandate_type.clone().map(Into::into)),
                 external_three_ds_authentication_attempted: None,
                 mandate_data,
                 payment_method_billing_address_id,
-                net_amount: MinorUnit::new(i64::default()),
+                net_amount: hyperswitch_domain_models::payments::payment_attempt::NetAmount::new(
+                    MinorUnit::from(amount),
+                    request.shipping_cost,
+                    None,
+                    surcharge_amount,
```
Narayanbhat166, line 1180: since all the fields here are MinorUnit, there might be chances where one amount is passed as another amount. Can we instead have this as a from function from the request instead of the new function, we can preserve the context of fields in that case when converting.

crates/router/src/core/payments/transformers.rs
sai-harsha-vardhan: This is order_amount only, we are sending order amount and currency in RedirectionFormData

crates/router/src/core/payments/operations/payment_confirm.rs
sai-harsha-vardhan: Based on the flow, we would need to restrict the updating of fields. Let's say in payments update flow, we don't want to update surcharge and tax amount.

jarnura (changes requested):

crates/router/src/core/payments/operations/payment_create.rs
jarnura: There is separate field storing this data in intent itself, @swangi-kumari can you confirm this once?

crates/router/src/core/payments/operations/payments_incremental_authorization.rs
jarnura: Add comments explain the case

crates/router/src/core/payments/types.rs
```diff
@@ -212,8 +212,9 @@ impl ForeignTryFrom<(&SurchargeDetails, &PaymentAttempt)> for SurchargeDetailsRe
                 .tax_on_surcharge_amount
                 .get_amount_as_i64(),
         )?;
-        let display_final_amount = currency
-            .to_currency_base_unit_asf64(surcharge_details.final_amount.get_amount_as_i64())?;
+        let display_final_amount = currency.to_currency_base_unit_asf64(
```
jarnura, line 215: @hrithikesh026 Is this final_amount is total_amount including all amount related calculation or only order_amount + surcharge?

crates/hyperswitch_domain_models/src/router_request_types.rs
```diff
@@ -528,6 +526,10 @@ impl SurchargeDetails {
     pub fn get_total_surcharge_amount(&self) -> MinorUnit {
         self.surcharge_amount + self.tax_on_surcharge_amount
     }
+
+    pub fn get_final_amount(&self) -> MinorUnit {
```
jarnura, line 530: Do we really needed this get_final_amount with just order_amount?

jarnura (changes requested):

crates/router/src/core/payments/operations/payment_create.rs
swangi-kumari: We store tax_details in payment_intent
```
pub struct PaymentIntent {
    pub tax_details: Option<TaxDetails>,
}
```


```
pub struct TaxDetails {
    pub default: Option<DefaultTax>,
    pub payment_method_type: Option<PaymentMethodTypeTax>,
}
```

sai-harsha-vardhan: Yes we store tax_details in payment_intent, but we would need to update payment_attempt.net_amount with order_tax_amount too so that all the subsequent flows would use the updated net_amount which includes order_tax_amount

crates/router/src/core/payments/types.rs
sai-harsha-vardhan: Currently `display_final_amount` is not being used, removing it

crates/router/src/core/payments/operations/payment_create.rs
jarnura: After confirm only we need to compute total_amount using intent, for create the related information will be stored on intent and the computation also on intent.

jarnura (changes requested):

jarnura (approved):
LGTM

jarnura (commented):
@Gnanasundari24 This is a critical change, please test all payment, refund and dispute flows.

crates/router/src/core/payments/routing.rs
Aprabhat19: Yes we would do it on total_amount, cause he could have a rule saying `amount`>1000, route it to `connector_b`. So total_amount would be a more accurate field for making such decisions

Aprabhat19 (approved):

PR #6253, opened 2024-10-07
fix(users): Allow accepting invites for `org_admin`s

ThisIsMani:
## Type of Change

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

Currently accepting invites for org_admin will not work. This PR fixes it.

### Additional Changes

## Motivation and Context

Closes [#6254](https://github.com/juspay/hyperswitch/issues/6254)

## How did you test it?

```
curl 'http://localhost:8080/user/accept_invite_from_email' \
  -H 'authorization: Bearer SPT with accept_invite_from_email' \
  --data-raw '{"token":"email token"}'
```
Response:
```json
{
    "token": "JWT",
    "token_type": "user_info"
}
```

## Checklist

### DIFF
--- a/crates/router/src/utils/user_role.rs
+++ b/crates/router/src/utils/user_role.rs
@@ -261,8 +261,11 @@ pub async fn get_lineage_for_user_id_and_entity_for_accepting_invite(
 > {
     match entity_type {
         EntityType::Organization => {
-            let org_id = id_type::OrganizationId::try_from(std::borrow::Cow::from(entity_id))
-                .change_context(UserErrors::InternalServerError)?;
+            let Ok(org_id) =
+                id_type::OrganizationId::try_from(std::borrow::Cow::from(entity_id.clone()))
+            else {
+                return Ok(None);
+            };
 
             let user_roles = state
                 .store
@@ -282,7 +285,7 @@ pub async fn get_lineage_for_user_id_and_entity_for_accepting_invite(
                 .collect::<HashSet<_>>();
 
             if user_roles.len() > 1 {
-                return Ok(None);
+                return Err(UserErrors::InternalServerError.into());
             }
 
             if let Some(user_role) = user_roles.into_iter().next() {
@@ -326,7 +329,7 @@ pub async fn get_lineage_for_user_id_and_entity_for_accepting_invite(
                 .collect::<HashSet<_>>();
 
             if user_roles.len() > 1 {
-                return Ok(None);
+                return Err(UserErrors::InternalServerError.into());
             }
 
             if let Some(user_role) = user_roles.into_iter().next() {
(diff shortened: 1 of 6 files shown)


crates/router/src/utils/user_role.rs
```diff
@@ -255,12 +255,54 @@ pub async fn get_lineage_for_user_id_and_entity_for_accepting_invite(
...
+                    version: None,
+                    status: Some(UserStatus::InvitationSent),
+                    limit: None,
+                })
+                .await
+                .change_context(UserErrors::InternalServerError)?
+                .into_iter()
+                .collect::<HashSet<_>>();
+
+            if user_roles.len() > 1 {
```
apoorvdixit88, line 284: We can evaluate this?

apoorvdixit88 (dismissed):

crates/router/src/utils/user_role.rs
```diff
@@ -255,12 +255,54 @@ pub async fn get_lineage_for_user_id_and_entity_for_accepting_invite(
...
-        id_type::MerchantId,
+        Option<id_type::MerchantId>,
         Option<id_type::ProfileId>,
     )>,
 > {
     match entity_type {
-        EntityType::Organization => Err(UserErrors::InvalidRoleOperation.into()),
+        EntityType::Organization => {
+            let org_id = id_type::OrganizationId::try_from(std::borrow::Cow::from(entity_id))
+                .change_context(UserErrors::InternalServerError)?;
```
SanchithHegde, line 265: Can we have a context message here (with `attach_printable` / `attach_printable_lazy`) and other places where we're raising internal server errors?

crates/router/src/utils/user_role.rs
ThisIsMani: This is correct as for a particular entity_id, there can only be 1 invite and not more than that.

crates/router/src/utils/user_role.rs
ThisIsMani: - I've gone through the code and checked wherever we are throwing `InternalServerError`. All of the places are mostly
  1. Data problems, which we can find using DB efficiently as we don't have the whole data of the object.
  2. Logical errors or Unreachable errors, which can be found out if we see the code at the line number at which the error is logged.
- In my opinion, adding `attach_printable`s without any important debug information will pollute the code. So, that's why I don't prefer adding them unless they are adding any value to the log.
- In the place where you commented, there was no need to throw 500 and I changed that code.

SanchithHegde (approved):

racnan (approved):
lets remove `Entity` type from email and use one `Entity` across repo.

apoorvdixit88 (approved):

Riddhiagrawal001 (approved):

Issue #2438, opened 2023-10-03
[FEATURE] Implement `MerchantAccountInterface` for `MockDb` 


SanchithHegde:
Spin off from #172. Refer to the parent issue for more information

dalprahcd:
Hello @SanchithHegde, I would like to help with this issue.

VedantKhairnar:
Hey @dalprahcd  ,
Let us know if you face any issues. Happy to help! : )

imSanko:
@dalprahcd I'm looking up with adding few changes in this directory if you can kindly assign this issue to me or let me know if a PR will be better to review my work or not!!

dalprahcd:
> Hey @dalprahcd , Let us know if you face any issues. Happy to help! : )

Thanks! I will let you know if I face any difficulties.

> @dalprahcd I'm looking up with adding few changes in this directory if you can kindly assign this issue to me or let me know if a PR will be better to review my work or not!!

Hello @imSanko, I want to help with this implementation, you are welcome to do any changes you want (as long as you don't implement `MerchantAccountInterface` for `MockDb`), and I can always merge with your changes.

VedantKhairnar:
Hey @dalprahcd  ,
Just checking in - are you working on it?
Let us know if you need any help.

dalprahcd:
Hello @SanchithHegde and @VedantKhairnar,

Looks like I will not be able to help with this implementation. My knowledge of Rust still not there.
I could only implement `delete_merchant_account_by_merchant_id`:
``` rust
    async fn delete_merchant_account_by_merchant_id(
        &self,
        merchant_id: &str,
    ) -> CustomResult<bool, errors::StorageError> {
        let mut accounts = self.merchant_accounts.lock().await;
        let index = accounts
            .iter()
            .position(|account| account.merchant_id == merchant_id)
            .ok_or::<errors::StorageError>(errors::StorageError::ValueNotFound(format!(
                "No merchant account found for merchant_id = {merchant_id}"
            )))?;
        accounts.remove(index);
        Ok(true)
    }
```

Anyone is welcome to take over from now on.

Sangamesh26:
Hey @dalprahcd, please let me know, if you want any kind of help.
If you wanna continue with this implementation, I will help you out.
Please join our [discord server](https://discord.gg/6XDUfB4w29) and ping in the channel (#dev-help-oss-contrib)

Happy to help :)

shivansh-bhatnagar18:
[...]

PR #6283, opened 2024-10-10
feat(db): implement `MerchantAccountInteraface` for `Mockdb`

akhildevelops:
## Type of Change

- [] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

Mockdb is a stub to avoid costly interactions with db during testing. This PR powers mockdb to have [MerchantAccountInterface](https://github.com/juspay/hyperswitch/blob/3d70b1a31703c5316183adaa9312a38e598fce01/crates/router/src/db/merchant_account.rs#L26) for adding, deleting, searching merchant accounts.

### Additional Changes

## Motivation and Context

Fixes https://github.com/juspay/hyperswitch/issues/2438

## Checklist

- [x] addressed lints thrown by `cargo clippy`


akhildevelops:
@gorakhnathy7 I'm done in implementing the interface. Please review and provide if there are any comments.

akhildevelops:
It's been quite some time since this pr is reviewed.

crates/router/src/db/merchant_account.rs
```diff
@@ -512,8 +505,13 @@ impl MerchantAccountInterface for MockDb {
         _merchant_account: storage::MerchantAccountUpdate,
         _merchant_key_store: &domain::MerchantKeyStore,
     ) -> CustomResult<domain::MerchantAccount, errors::StorageError> {
-        // [#172]: Implement function for `MockDb`
-        Err(errors::StorageError::MockDbError)?
+        self.update_specific_fields_in_merchant(
+            _state,
+            _this.get_id(),
+            _merchant_account,
+            _merchant_key_store,
```
SanchithHegde, line 512: You can remove the underscore prefix from the function params as well as here.

crates/router/src/db/merchant_account.rs
```diff
@@ -524,7 +522,19 @@ impl MerchantAccountInterface for MockDb {
         _merchant_key_store: &domain::MerchantKeyStore,
     ) -> CustomResult<domain::MerchantAccount, errors::StorageError> {
         // [#TODO]: Implement function for `MockDb`
```
SanchithHegde, line 524: Can you remove this comment, wherever you've added the implementation?

crates/diesel_models/src/merchant_account.rs
```diff
@@ -197,6 +197,57 @@ impl MerchantAccount {
...
+        &mut self,
+        update: MerchantAccountUpdateInternal,
+    ) -> Result<(), &'static str> {
+        macro_rules! update {
+            ([$(($attr:ident,)),*]) => {
+
+                $(self.$attr = update.$attr;)*
+            };
+            ([$(($attr:ident,$error:literal)),*])=>{
+                $(self.$attr = update.$attr.ok_or(concat!($error,stringify!($attr)))?;)*
```
SanchithHegde, line 212: This would cause it to throw an error for all fields that are `None`?

That is not the intended behavior here, fields being `None` only signify that they would not be included in the `UPDATE` query, and thus would not be updated.

From [the diesel docs](https://diesel.rs/guides/all-about-updates.html):

> If the struct has any optional fields on it, these will also have special behavior. By default, `#[derive(AsChangeset)]` will assume that `None` means that you don’t wish to assign that field.

crates/router/src/db/merchant_account.rs
```diff
@@ -524,7 +522,19 @@ impl MerchantAccountInterface for MockDb {
...
+        let mut accounts = self.merchant_accounts.lock().await;
+        let account = accounts
+            .iter_mut()
+            .find(|account| account.get_id() == _merchant_id)
+            .ok_or(errors::StorageError::ValueNotFound(format!(
+                "Merchant ID: {:?} not found",
+                _merchant_id
+            )))?;
+        account
+            .from_update(_merchant_account.into())
```
SanchithHegde, line 534: Considering my comment on the `from_update()` method, can we avoid the macro implementation, and do something similar to the other implementations?

Here's the update implementation on profiles, for your reference:
https://github.com/juspay/hyperswitch/blob/4ef48c39b3ed7c1fcda9c850da766a0bdb701335/crates/router/src/db/business_profile.rs#L325-L364

crates/router/src/db/merchant_account.rs
```diff
@@ -533,22 +543,62 @@ impl MerchantAccountInterface for MockDb {
...
         // [#172]: Implement function for `MockDb`
-        Err(errors::StorageError::MockDbError)?
+        let accounts = self.merchant_accounts.lock().await;
+        let account = accounts
+            .iter()
+            .find(|account| {
+                account
+                    .publishable_key
+                    .as_ref()
+                    .is_some_and(|key| key == _publishable_key)
```
SanchithHegde, line 553: You can remove the underscore prefix from function params, wherever you're accessing them in your implementations.

akhildevelops:
@SanchithHegde Can you check now, I've found an innovative way without cloning and skipping lot of boilerplate to update merchant account by implementing a special trait to Option<T>

crates/diesel_models/src/merchant_account.rs
```diff
@@ -197,6 +197,60 @@ impl MerchantAccount {
...
+    fn edit_value_from_self(self, value:&mut T){
+        if let Some(x)=self{
+            *value=x;
+        }
+    }
+}
+
+impl MerchantAccount {
+    #[cfg(feature = "v1")]
+    pub fn from_update(&mut self, update: MerchantAccountUpdateInternal) {
```
SanchithHegde, line 225: One of the primary reasons we use the [struct expression syntax](https://doc.rust-lang.org/reference/expressions/struct-expr.html) in our `apply_changeset` implementations (and explicitly return `MerchantAccount`, `Profile`, etc.) is so that the compiler can throw errors if we forget to include the new fields here. With this implementation, we'd lose out on that benefit.

akhildevelops: Cool will refactor.

akhildevelops: Designed with reference to profile.

akhildevelops:
@SanchithHegde, can you review now re-designed merchantaccount updates for mockdb

.gitignore
SanchithHegde, line 1: Nit: Can we revert changes to this file?

SanchithHegde (commented):

SanchithHegde (approved):

SanchithHegde:
@akhildevelops Could you please address the failing CI checks?

akhildevelops:
@SanchithHegde Can you please check now by running workflows.

SanchithHegde (approved):

Chethan-rao (approved):

PR #6367, opened 2024-10-18
ci: introduce batch-migrate tests

pixincreate:
taken from #5306## Type of Change

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [x] CI/CD

## Description

this pr adds tests and framework changes related to batch-migrate feature from 
https://github.com/juspay/hyperswitch/pull/5306

### Additional Changes

## Motivation and Context

Jagan.

## How did you test it?

Tests fail as the merchant account does not exist.

- batch-migrate expects you to have an account setup already.
cypress is a unified tool and you cannot exclude certain tests from running within a service that easily in an environment.
- at present, we've added a csv file in `.github/data` dir to pass the file in the request.
the blocker is that `mid`s and other unique values in the file needs to be existent in certain environments.
- this forces us to keep `2` separate files (one for integ and the other for sandbox).
- additionally, we do cypress tests in github ci. for that, we need to inject / fill-in data into the `hyperswitch_db` every time before the tests are run (we're already doing it for adding `cards_info` now)

## Checklist


pixincreate:
closing since this cannot be achieved, at least with the current setup.

PR #6380, opened 2024-10-21
feat(users): Add profile level custom role

Riddhiagrawal001:
## Type of Change

- [ ] Bugfix
- [X] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

### Additional Changes

## Checklist

### DIFF
--- a/crates/api_models/src/user_role/role.rs
+++ b/crates/api_models/src/user_role/role.rs
@@ -6,6 +6,7 @@ pub struct CreateRoleRequest {
     pub role_name: String,
     pub groups: Vec<PermissionGroup>,
     pub role_scope: RoleScope,
+    pub entity_type: Option<EntityType>,
 }
 
 #[derive(Debug, serde::Deserialize, serde::Serialize)]
@@ -20,6 +21,7 @@ pub struct RoleInfoWithGroupsResponse {
     pub groups: Vec<PermissionGroup>,
     pub role_name: String,
     pub role_scope: RoleScope,
+    pub entity_type: Option<EntityType>,
 }
 
 #[derive(Debug, serde::Deserialize, serde::Serialize)]
--- a/crates/diesel_models/src/query/role.rs
+++ b/crates/diesel_models/src/query/role.rs
@@ -106,6 +106,7 @@ impl Role {
         conn: &PgPooledConn,
         org_id: id_type::OrganizationId,
         merchant_id: Option<id_type::MerchantId>,
+        profile_id: Option<id_type::ProfileId>,
         entity_type: Option<common_enums::EntityType>,
         limit: Option<u32>,
     ) -> StorageResult<Vec<Self>> {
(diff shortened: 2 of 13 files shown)


crates/diesel_models/src/query/role.rs
```diff
@@ -131,6 +132,97 @@ impl Role {
...
+                };
+                query = query
+                    .filter(dsl::org_id.eq(org_id))
+                    .filter(
+                        dsl::scope
+                            .eq(RoleScope::Organization)
+                            .or(dsl::scope.eq(RoleScope::Merchant))
+                            .or(dsl::scope.eq(RoleScope::Profile)),
+                    )
+                    .filter(dsl::entity_type.eq_any(entity_in_vec))
```
apoorvdixit88, line 177: Just for clarity, In case when `is_lineage data_required` `true` case this query will return same result as we will getting without applying this `eq_any` filter?

crates/router/src/core/user_role/role.rs
```diff
@@ -276,28 +300,47 @@ pub async fn list_roles_with_info(
...
+                        user_from_token.org_id,
+                        user_from_token.merchant_id,
+                    ),
+                    request.entity_type.is_none(),
                     None,
                 )
                 .await
                 .change_context(UserErrors::InternalServerError)
                 .attach_printable("Failed to get roles")?,
             // TODO: Populate this from Db function when support for profile id and profile level custom roles is added
```
apoorvdixit88, line 324: We can remove this todo now?

migrations/2024-10-17-073555_add-profile-id-to-roles/up.sql
```diff
@@ -0,0 +1,2 @@
+-- Your SQL goes here
+ALTER TABLE roles ADD profile_id Varchar(64);
```
apoorvdixit88, line 2: nitpick:
- To keep drop.sql and up.sql consistent we can either use COLUMN keyword in both, or drop it for both, since it is optional.
- We can also use IF EXISTs and IF NOT EXISTs (Optional change)

crates/api_models/src/user_role/role.rs
```diff
@@ -21,6 +22,7 @@ pub struct RoleInfoWithGroupsResponse {
     pub groups: Vec<PermissionGroup>,
     pub role_name: String,
     pub role_scope: RoleScope,
+    pub entity_type: Option<EntityType>,
```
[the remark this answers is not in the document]
Riddhiagrawal001: Because in the request its optional , so thought in response also it should be optional

crates/diesel_models/src/query/role.rs
Riddhiagrawal001: Yes

ThisIsMani (commented):
Are you sure that old db functions will work without any issues with the new data?

ThisIsMani:
1. Please add indices similar to merchant and org.
2. As this PR affects the older queries (new data will be read by old queries), if a rollback happens, there is a high chance that application will break. Please check if the newer data breaks the old application. If yes, the older queries should be changed and deployed first.

Issue #6316, opened 2024-10-15
[REFACTOR] Worldpay payments integration


kashif-m:
### Feature Description

WorldPay payments integration currently makes use of modular APIs which gives more control over the integration. This approach is not best compatible with HyperSwitch due to WP's 3DS flows - which requires sending PAN (card numbers) in multiple places. For `on_session` transactions, this would contribute to the added complexity since data is not stored in such transactions (or stored for ~15mins in redis cache).

Migrating to WP's standard payments API follows payment resource creation in an uniform way.

### Possible Implementation

Migrate from modular APIs [[ref](https://developer.worldpay.com/products/access/card-payments)] to standard payment APIs [[ref](https://developer.worldpay.com/products/access/payments/openapi)].

### Have you spent some time checking if this feature request has been raised before?

- [X] I checked and didn't find a similar issue

### Have you read the Contributing Guidelines?

- [X] I have read the [Contributing Guidelines](https://github.com/juspay/hyperswitch/blob/main/docs/CONTRIBUTING.md)

### Are you willing to submit a PR?

Yes, I am willing to submit a PR!

kashif-m:
Propagate the refusal codes as error codes in HyperSwitch.

kashif-m:
# 3

- Make payment_method_type non-mandatory
- Fix PM filters

kashif-m:
# # 4

- Add cypress test cases

kashif-m:
# # 5

- Use automatic capture by default

kashif-m:
# # 6

- Mandates
- Webhooks

kashif-m:
# # 7 Bugs

- Use 4 digit expiry year

kashif-m:
Re-opening for https://github.com/juspay/hyperswitch/pull/6468

kashif-m:
# # 10

- Add zip code to dynamic fields
- Update `connector_transaction_data` and `connector_refund_data` columns to 1024

kashif-m:
Closing this as everything is resolved here.

PR #6420, opened 2024-10-24
feat(cypress-test): include worldpay's request / response structure for test suite

kashif-m:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [x] CI/CD

## Description
Described in #6316 point # 4 (read comments)

### Additional Changes

## Motivation and Context
Helps in sanity testing of Worldpay connector.

## How did you test it?
Locally using `npm run cypress:payments`
Screenshot 2024-10-24 at 11 17 11 AM

## Checklist

### DIFF
--- a/cypress-tests/cypress/e2e/PaymentTest/00006-VoidPayment.cy.js
+++ b/cypress-tests/cypress/e2e/PaymentTest/00006-VoidPayment.cy.js
@@ -67,7 +67,7 @@ describe("Card - NoThreeDS Manual payment flow test", () => {
 
     it("void-call-test", () => {
       let data = getConnectorDetails(globalState.get("connectorId"))["card_pm"][
-        "Void"
+        "VoidAfterConfirm"
       ];
       let req_data = data["Request"];
       let res_data = data["Response"];
@@ -178,7 +178,7 @@ describe("Card - NoThreeDS Manual payment flow test", () => {
       it("void-call-test", () => {
         let data = getConnectorDetails(globalState.get("connectorId"))[
           "card_pm"
-        ]["Void"];
+        ]["VoidAfterConfirm"];
         let req_data = data["Request"];
         let res_data = data["Response"];
         cy.voidCallTest(fixtures.voidBody, req_data, res_data, globalState);
--- a/cypress-tests/cypress/e2e/PaymentTest/00013-SaveCardFlow.cy.js
+++ b/cypress-tests/cypress/e2e/PaymentTest/00013-SaveCardFlow.cy.js
@@ -37,7 +37,7 @@ describe("Card - SaveCard payment flow test", () => {
           "card_pm"
         ]["SaveCardUseNo3DSAutoCapture"];
         let req_data = data["Request"];
-        let res_data = data["Response"];
+        let res_data = data["ResponseCustom"] ? data["ResponseCustom"] : data["Response"];
         cy.createConfirmPaymentTest(
           fixtures.createConfirmPaymentBody,
           req_data,
@@ -115,7 +115,7 @@ describe("Card - SaveCard payment flow test", () => {
           "card_pm"
         ]["SaveCardUseNo3DSAutoCapture"];
         let req_data = data["Request"];
-        let res_data = data["Response"];
+        let res_data = data["ResponseCustom"] ? data["ResponseCustom"] : data["Response"];
         cy.createConfirmPaymentTest(
           fixtures.createConfirmPaymentBody,
           req_data,
@@ -214,7 +214,7 @@ describe("Card - SaveCard payment flow test", () => {
           "card_pm"
         ]["SaveCardUseNo3DSAutoCapture"];
         let req_data = data["Request"];
-        let res_data = data["Response"];
+        let res_data = data["ResponseCustom"] ? data["ResponseCustom"] : data["Response"];
         cy.createConfirmPaymentTest(
           fixtures.createConfirmPaymentBody,
           req_data,
@@ -312,7 +312,7 @@ describe("Card - SaveCard payment flow test", () => {
           "card_pm"
         ]["SaveCardUseNo3DSAutoCaptureOffSession"];
         let req_data = data["Request"];
-        let res_data = data["Response"];
+        let res_data = data["ResponseCustom"] ? data["ResponseCustom"] : data["Response"];
         cy.createConfirmPaymentTest(
           fixtures.createConfirmPaymentBody,
           req_data,
--- a/cypress-tests/cypress/e2e/PaymentTest/00013-SaveCardFlow.cy.js
+++ b/cypress-tests/cypress/e2e/PaymentTest/00013-SaveCardFlow.cy.js
@@ -47,10 +47,11 @@ describe("Card - SaveCard payment flow test", () => {
           globalState
         );
         if (should_continue) {
-          // Don't continue if payment status is processing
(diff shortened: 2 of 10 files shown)


cypress-tests/cypress/e2e/PaymentTest/00006-VoidPayment.cy.js
pixincreate, line 1: any changes done to the `PaymentTest` will affect all the connectors. please make sure to test with other connectors if this change is needed.

cypress-tests/cypress/e2e/PaymentTest/00020-Variations.cy.js
```diff
@@ -331,7 +331,8 @@ describe("Corner cases", () => {
     });
 
     it("Capture call", () => {
-      let data = getConnectorDetails(globalState.get("commons"))["card_pm"][
+      let connectorId = globalState.get("connectorId");
+      let data = getConnectorDetails(connectorId === "worldpay" ? connectorId : globalState.get("commons"))["card_pm"][
```
pixincreate, line 335: is there any other way to achieve the same?
asking so because we do not want to hardcode `tests` to a specific connector and prefer to have tests as generic as possible. a specific should not be an exception.

same applies to the changes done in file.

also, i highly doubt this will not work. it should be `connectorId["card_pm"]["CaptureCapturedAmount"]` instead of just `connectorId`

cypress-tests/cypress/e2e/PaymentUtils/Commons.js
```diff
@@ -666,6 +666,27 @@ export const connectorDetails = {
...
+        body: {
+          error: {
+            type: "invalid_request",
+            message:
+              "You cannot cancel this payment because it has status succeeded",
+            code: "IR_16",
+          },
+        },
+      },
+    }),
```
pixincreate, line 689: again, this will affect all the connectors since you've modified `PaymentTest`

cypress-tests/cypress/e2e/PaymentUtils/WorldPay.js
```diff
@@ -0,0 +1,508 @@
...
+    //         body: {
+    //             error: {
+    //                 type: "invalid_request",
+    //                 message: "Payment method type not supported",
+    //                 code: "IR_19",
+    //                 reason: "debit mandate payment is not supported by worldpay"
+    //             }
+    //         },
+    //     },
+    // },
```
pixincreate, line 341: commented code can be removed

cypress-tests/cypress/e2e/PaymentUtils/WorldPay.js
```diff
@@ -0,0 +1,508 @@
...
+        body: {
+          error: {
+            type: "invalid_request",
+            message:
+              "You cannot confirm this payment because it has status processing",
+            code: "IR_16",
+          },
+        },
+      },
+    },
```
pixincreate, line 389: these are handled by `commons.js`?

cypress-tests/cypress/e2e/PaymentUtils/Commons.js
kashif-m: This is handled by fallbackData which is fetched from `Commons.js`

cypress-tests/cypress/e2e/PaymentTest/00020-Variations.cy.js
kashif-m: Connector specific data should always be fetched from respective connectors, and the fallback logic can handle fetching the data from `Commons.js` wherever it's needed

cypress-tests/cypress/e2e/PaymentUtils/WorldPay.js
kashif-m: They're present for a different status (`succeeded`)

cypress-tests/cypress/e2e/PaymentTest/00020-Variations.cy.js
pixincreate: on variations, most of the tests never hit a connector until now and hence we directly used `commons`. now that i think of, even if you remove the hardcoded `commons` and replace that with `connectorId`, it should work. the reason being, the `Utils.js` in `PaymentUtils` merges the `<connector_name>.js` with `commons.js` before the test is run.

cypress-tests/cypress/e2e/PaymentTest/00013-SaveCardFlow.cy.js
```diff
@@ -37,7 +37,7 @@ describe("Card - SaveCard payment flow test", () => {
           "card_pm"
         ]["SaveCardUseNo3DSAutoCapture"];
         let req_data = data["Request"];
-        let res_data = data["Response"];
+        let res_data = data["ResponseCustom"] ? data["ResponseCustom"] : data["Response"];
```
pixincreate, line 40: I feel this wont work as what we expect.

We, at least, need to make this depend on the response status, not a boolean value while giving priority to `Response` and not `ResponseCustom`. 

Main use case of using `ResponseCustom` is that when you're expecting the connector to throw 2 or more different responses for a payment.

cypress-tests/cypress/e2e/PaymentTest/00013-SaveCardFlow.cy.js
```diff
@@ -312,7 +312,7 @@ describe("Card - SaveCard payment flow test", () => {
           "card_pm"
         ]["SaveCardUseNo3DSAutoCaptureOffSession"];
```
pixincreate, line 313: this config does not exist in `WorldPay.js` either. So, it will be fetched from `commons.js` and by default, it will expect a 5xx or 4xx.

kashif-m: Yes! that's the expectation at the moment (since mandates are not implemented yet)

cypress-tests/cypress/e2e/PaymentTest/00020-Variations.cy.js
kashif-m: Yes!

pixincreate (approved):
since this affects almost every connector out there, it would be great if you could run the test against all the connectors once to confirm. yes, multiple connectors do fail due to blockers that cypress cannot handle (at least, until now).

you can make use of `scripts/execute_cypress.sh` script and run against multiple connectors in parallel.

please refer to README to know more about running in parallel / multiple connectors at once.

cypress-tests/cypress/e2e/PaymentUtils/Utils.js
```diff
@@ -39,13 +40,19 @@ const connectorDetails = {
...
 export default function getConnectorDetails(connectorId) {
   let x = mergeDetails(connectorId);
   return x;
 }
 
+export function getConnectorFlowDetails(connectorData, commonData, key) {
+  let data = connectorData[key] === undefined ? commonData[key] : connectorData[key];
+  return data;
+}
+
```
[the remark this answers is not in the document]
kashif-m: It does have a very limited scope as of now, we can remove this once we introduce custom deep merge (merge everything except `Response`). But we can let it be for now, since it's out of scope of this PR

likhinbopanna (approved):

PR #5877, opened 2024-09-13
refactor: Rename business profile to profiles in api, diesel, domain, interface and error types

jarnura:
- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [x] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

Rename business profile to profiles in api, diesel, domain, interface and error types

### Additional Changes

- [x] This PR modifies the API contract
- [ ] This PR modifies the database schema
- [ ] This PR modifies application configuration/environment variables

- API Model types related to `BusinessProfiles` changed to `Profiles` both v1 and v2
- Diesel Model types related to `BusinessProfiles` changed to `Profiles` for both v1 and v2 since almost all the places we use common functions, and the table name is not changed for both versions. Only the transit types in the application changed
- Domain Model types related to `BusinessProfiles` changed to `Profiles` for both v1 and v2 since this is transit type in application
- Error variant for Profile where changed
	- `BusinessProfileNotFound` to `ProfileNotFound`
- In Openapi spec of v1 the function name used is changed
- In Openapi spec of v2 both function name and types used where changed
- In `admin.rs` all private struct for both v1 and v2 is changes because those where transit types
- Database interface where changed to `ProfileInterface`
- Profile related route handler wrapper where moved to a separate file `src/routes/profile.rs` from `src/routes/admin.rs`

## Motivation and Context

Currently the usage of Profile terms and BusinessProfile terms messed up and there is no consistency in usage of these terms.

This pull request covers only a part of renaming the business profile relates changes.

## How did you test it?

These are only types changes, so if the compilation is successful then everything is good. 

## Checklist

- [x] I formatted the code `cargo +nightly fmt --all`
- [x] I addressed lints thrown by `cargo clippy`
- [x] I reviewed the submitted code
- [x] I added unit tests for my changes where possible

PR #6433, opened 2024-10-24
refactor(payment_methods): refactor customer payment methods list v2 code to follow better code practices

SanchithHegde:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [x] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

This PR includes refactors around the customer payment methods list v2 code to follow some of the practices we've started enforcing in the recent few months. Additionally, I've picked up some minor refactors along the way. 

This PR contains the following changes:

- Adds a `should_collect_cvv_during_payment` column to the business profile v2 table to replace the `{merchant_id}_requires_cvv` config we used to read previously (and still do in v1) from the `configs` table.
- Renames the `business_profile` field to `profile` in the `SavedPMLPaymentsInfo` struct. (This follows the changes done in #5877.)
- Renames the `requires_cvv` field to `collect_cvv_during_payment` in `SavedPMLPaymentsInfo` struct, to follow the new column added in the business profile v2 table.
- Passes profile to the `form_payments_info()` method instead of fetching it within the method. Eventually, we'd want to obtain the profile from the authentication layer instead, which would be passed forward.
- Moves the `PaymentsMandateReference` from the `router` crate to the `diesel_models` crate, and updates the `connector_mandate_details` field in the payment methods table to use `PaymentsMandateReference` instead of `serde_json::Value` in v2 code. (This should help reduce / prevent invalid data being stored in this column of the database.)
- Refactors the `get_mca_status()` function to avoid a possible database call (to fetch merchant connector accounts associated with a merchant / profile) happening within an iterator.
- Drops the `metadata` column from the payment methods table and API models in v2 code, since it isn't being actively used in v1 code anyway. Neither do any of our hosted environments have any useful data stored in that column.

The other minor refactors I took up along the way include:

- Improves the doc comment for the `is_network_tokenization_enabled` field in profile API models.
- Update the `ExtendedCardInfoUpdate`, `ConnectorAgnosticMitUpdate` and `NetworkTokenizationUpdate` enum variants of the `ProfileUpdate` domain model enum to have required fields instead of optional ones. (This is because these variants either enable something or not, and we always have the value to be updated in all cases.)
- Uses `serialize_and_set_key_with_expiry()` when storing payment method token in Redis instead of explicit JSON serialization happening in business logic.
- Replace usages of `create_payment_method()` (which only seemed to be updating the metadata field, and was thus confusing) with a correct `apply_changeset()` implementation. This change only affects MockDb implementation.

The PR can be reviewed one commit at a time, if that helps ease reviews.

### Additional Changes

- This PR adds a new column in the business profile v2 database table and corresponding field in the business profile v2 API models.
- This PR drops the metadata column from the payment methods v2 database table and corresponding payment methods v2 API models.

## Motivation and Context

This PR contains only a subset of the changes / improvements that can help us in arriving at the final implementation of the customer payment methods list v2 API. There are still a few more refactors pending, which would be taken up in subsequent PRs.

## How did you test it?

Haven't been able to test these changes, there are still a couple of panics remaining in the customer payment methods list v2 code.

## Checklist


crates/router/src/core/payment_methods.rs
```diff
@@ -1456,7 +1453,7 @@ pub async fn list_customer_payment_method_util(
 ))]
 pub async fn list_customer_payment_method(
     state: &SessionState,
-    merchant_account: domain::MerchantAccount,
+    merchant_account: &domain::MerchantAccount,
```
jarnura, line 1456: fetch profile also from middleware

crates/router/src/core/payment_methods/cards.rs
```diff
@@ -4669,25 +4653,64 @@ pub async fn get_mca_status(
...
+    merchant_id: &id_type::MerchantId,
+    is_connector_agnostic_mit_enabled: bool,
+    connector_mandate_details: Option<&payment_method::PaymentsMandateReference>,
+    network_transaction_id: Option<&String>,
+    merchant_connector_accounts: Vec<domain::MerchantConnectorAccount>,
+) -> bool {
+    if is_connector_agnostic_mit_enabled && network_transaction_id.is_some() {
+        return true;
+    }
+    if let Some(connector_mandate_details) = connector_mandate_details {
```
jarnura, line 4692: use match instead of if let Some pattern

crates/router/src/core/payment_methods/cards.rs
```diff
@@ -4669,25 +4653,64 @@ pub async fn get_mca_status(
...
+        let mca_ids = merchant_connector_accounts
```
jarnura, line 4693: create a private `struct MerchantConnectorAccounts(Vec<MerchantConnectorAccount>)`

and write these as impl function, use this impl function above too.

jarnura (changes requested):

crates/diesel_models/src/payment_method.rs
```diff
@@ -838,3 +842,37 @@ impl From<&PaymentMethodNew> for PaymentMethod {
         }
     }
 }
+
+#[derive(Debug, Clone, serde::Serialize, serde::Deserialize)]
+pub struct PaymentsMandateReferenceRecord {
+    pub connector_mandate_id: String,
+    pub payment_method_type: Option<common_enums::PaymentMethodType>,
```
jarnura, line 849:
```suggestion
    pub payment_method_sub_type: Option<common_enums::PaymentMethodType>,
```
In v2 the PaymentMethodType is referred as payment_method_sub_type and PaymentMethod is referred as payment_method_type

crates/diesel_models/src/payment_method.rs
```diff
@@ -838,3 +842,37 @@ impl From<&PaymentMethodNew> for PaymentMethod {
...
+    pub original_payment_authorized_amount: Option<i64>,
+    pub original_payment_authorized_currency: Option<common_enums::Currency>,
+    pub mandate_metadata: Option<serde_json::Value>,
```
jarnura, line 852:
```suggestion
    pub mandate_metadata: make this secret value,
```

jarnura (changes requested):

crates/diesel_models/src/payment_method.rs
SanchithHegde: We'd need to update the API models, diesel and domain models as well. I'll take this up in a separate PR.

jarnura (dismissed):

ThisIsMani (dismissed):

Sarthak1799 (dismissed):

jarnura (approved):

Sarthak1799 (approved):

ThisIsMani (approved):

PR #6441, opened 2024-10-25
feat: Added grpc based health check

Sarthak1799:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

Implemented gRPC based health check

### Additional Changes

## How did you test it?

Tested locally
```
curl --location --request GET 'http://localhost:8080/health/ready' \
--header 'api-key: dev_Li5BLbzNIHjf2t0J6kYgS7sQurtmLubs1sGAxTE6MVo59aMBEMDSK6Qt8me7kJZQ'
```
Response
```
{
    "database": true,
    "redis": true,
    "analytics": true,
    "opensearch": true,
    "outgoing_request": true,
    "grpc_health_check": {
        "dynamic_routing_service": true
    }
}
```
image

Response when runtime config is disabled
```
{
    "database": true,
    "redis": true,
    "analytics": true,
    "opensearch": true,
    "outgoing_request": true,
    "grpc_health_check": {}
}
```

## Checklist


crates/api_models/src/health_check.rs
```diff
@@ -9,6 +9,8 @@ pub struct RouterHealthCheckResponse {
     #[cfg(feature = "olap")]
     pub opensearch: bool,
     pub outgoing_request: bool,
+    #[cfg(feature = "dynamic_routing")]
+    pub grpc_health_check: bool,
```
NishantJoshi00, line 13: instead, this should be a map of services that should be healthy right?

crates/external_services/src/grpc_client/health_check_client.rs
```diff
@@ -0,0 +1,100 @@
...
+        let uri = format!("http://{}:{}", host, port);
+        let channel = tonic::transport::Endpoint::new(uri)
+            .map_err(|err| HealthCheckError::ConnectionError(err.to_string()))?
+            .connect()
+            .await
+            .map_err(|err| HealthCheckError::ConnectionError(err.to_string()))?;
+        let mut client = HealthClient::new(channel);
+
+        let request = tonic::Request::new(HealthCheckRequest {
+            service: "dynamo".to_string(), // check this in review
```
NishantJoshi00, line 87: better to pass this as an input to this function. and generate if from the config (i.e. `DynamicRoutingClientConfig` in this case)

NishantJoshi00 (changes requested):

crates/api_models/src/health_check.rs
Sarthak1799: We only had one service so did not include it atm.
You mean a map of (service -> bool) right?

crates/external_services/src/grpc_client/health_check_client.rs
```diff
@@ -0,0 +1,112 @@
...
+        host: String,
+        port: u16,
+        service: String,
+    ) -> HealthCheckResult<HealthCheckResponse> {
+        let uri = format!("http://{}:{}", host, port);
+        let channel = tonic::transport::Endpoint::new(uri)
+            .map_err(|err| HealthCheckError::ConnectionError(err.to_string()))?
+            .connect()
+            .await
+            .map_err(|err| HealthCheckError::ConnectionError(err.to_string()))?;
```
Chethan-rao, line 97: Should we consider opening this channel only once during app start and using this channel whenever health check is hit? Because currently for every health check api hit, this channel will be opened right?

crates/external_services/src/grpc_client/health_check_client.rs
```diff
@@ -0,0 +1,112 @@
...
+                "error calling dynamic routing service".to_string(),
+            ))
+            .ok()
+            .flatten()
+            .is_some_and(|resp| resp.status == expected_status);
+
+        service_map.insert(
+            HealthCheckServices::DynamicRoutingService,
+            health_check_succeed,
+        );
```
Chethan-rao, line 81: If dynamic_routing is not enabled in runtime, still this will be inserted to map right?

Chethan-rao (changes requested):

crates/external_services/src/grpc_client/health_check_client.rs
Sarthak1799: The entire health check module is gated under `dynamic_routing` feature flag. Would this still be an issue?

Chethan-rao: That's compile time. What if it's disabled in runtime

crates/external_services/src/grpc_client/health_check_client.rs
Sarthak1799: Missed this, will refactor.

crates/external_services/src/grpc_client/health_check_client.rs
```diff
@@ -0,0 +1,112 @@
...
+        let mut service_map = HealthCheckMap::new();
+
+        let health_check_succeed = connection
+            .async_map(|conn| self.get_response_from_grpc_service(conn.0, conn.1, conn.2))
+            .await
+            .transpose()
+            .change_context(HealthCheckError::ConnectionError(
+                "error calling dynamic routing service".to_string(),
+            ))
+            .ok()
```
Chethan-rao, line 74: Error is dropped after ok and is not propagated or logged?

NishantJoshi00 (approved):

crates/external_services/src/grpc_client/health_check_client.rs
```diff
@@ -0,0 +1,149 @@
...
+    /// status is invalid
+    #[error("Invalid Status from server")]
+    InvalidStatus,
+}
+
+/// Health Check Client type
+#[derive(Debug, Clone)]
+pub struct HealthCheckClient {
+    /// Health clients for all gRPC based services
+    pub clients: HashMap<HealthCheckServices, HealthClient<Client>>,
```
Chethan-rao, line 45: `HealthClient` is specific to dynamic routing health check client right? other grpc services's health client cannot be added to this HashMap as its HealthClient will be different

crates/external_services/src/grpc_client.rs
```diff
@@ -1,18 +1,38 @@
...
 /// Struct contains all the gRPC Clients
 #[derive(Debug, Clone)]
 pub struct GrpcClients {
     /// The routing client
     #[cfg(feature = "dynamic_routing")]
     pub dynamic_routing: RoutingStrategy,
+    /// Health Check client for all gRPC services
+    #[cfg(feature = "dynamic_routing")]
+    pub health_client: HealthCheckClient,
 }
```
Chethan-rao, line 36: According to this type, you will still have to build 2 connections (one for actual service communication and other for health check)? Should this health check client be tied to the actual gRPC services types?
cc: @Aprabhat19

Chethan-rao (changes requested):

crates/external_services/src/grpc_client/health_check_client.rs
Sarthak1799: If I understand this correctly, the `HealthClient` type is borrowed from proto types, in which case the type for the `HealthClient` for all other gRPC services should be the same (protofile being the same). The `HealthClient` instance however, would be different for each service in the hashmap.

Chethan-rao (approved):

Issue #6457, opened 2024-10-28
feat(users): add tenant_id in user roles


apoorvdixit88:
- Use global interface for user roles
- Add column tenant_id in user_roles
- Handle insertion of tenant_id
- Hadle token validation, a token issued in one tenant should not work in other tenancy

PR #6458, opened 2024-10-28
feat(users): add global support in user roles

apoorvdixit88:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description
This PR includes
- Use of global interface for user roles table
- Add tenant_id column in user_roles table
- Handle user_role insertions with correct value for tenant_id
- Auth Changes: Token validation, token issued in one tenancy should not be valid in other

### Additional Changes

## Motivation and Context
Closes [#6457](https://github.com/juspay/hyperswitch/issues/6457)

## How did you test it?
With the multi -tenancy feature flag enabled, when trying to signup by sending different headers for different tenancies>
```
curl --location 'http://localhost:8080/user/signup?token_only=true' \
--header 'x-tenant-id: test' \
--data-raw '{
    "email": "global_test@juspay.in",
    "password": "Hyper@123"
}'
```
```
curl --location 'http://localhost:8080/user/signup?token_only=true' \
--data-raw '{
    "email": "global_public@juspay.in",
    "password": "Hyper@123"
}'
```
The user role is getting inserted properly, with correct values of tenant_id.
Screenshot 2024-10-29 at 2 42 08 AM

When Trying to use one JWT token with some other header that doesn't match with origin tenancy where it is issued it is authentication part is throwing error. Which is correct.

## Checklist


crates/router/src/services/authentication.rs
```diff
@@ -186,6 +187,7 @@ pub struct SinglePurposeToken {
     pub origin: domain::Origin,
     pub path: Vec<TokenPurpose>,
     pub exp: u64,
+    pub tenant_id: Option<String>,
```
ThisIsMani, line 190: What are the problems that this `tenant_id` in the JWT solving?

Is it only because to check whether this token is issued in this `tenant_id` or not?

Is there anything else?

crates/router/src/types/domain/user.rs
```diff
@@ -1104,6 +1106,8 @@ pub struct NewUserRole<E: Clone> {
     pub created_at: PrimitiveDateTime,
     pub last_modified: PrimitiveDateTime,
     pub entity: E,
+    //TODO: This to be removed and a new type TenantLevel to be created
+    pub tenant_id: String,
```
ThisIsMani, line 1110: Add `tenant_id` in the level structs. This should not be added here.

ThisIsMani: There is no need for tenant levels as of now. 

Unless we are not adding tenant level user roles.

crates/router/src/core/user.rs
```diff
@@ -644,6 +644,8 @@ async fn handle_existing_user_invitation(
         created_at: now,
         last_modified: now,
         entity: domain::NoLevel,
+        // TODO: To be sent from token after option is removed
+        tenant_id: state.tenant.tenant_id.clone(),
```
ThisIsMani, line 648: Can we do `unwrap_or` instead.

This way compiler tells us to change once we remove `tenant_id` from the JWT.

ThisIsMani (changes requested):

crates/router/src/services/authentication.rs
apoorvdixit88: Just to check whether the token belong to given tenancy or not.
Token of one tenancy should not be valid in other.

crates/router/src/core/user.rs
apoorvdixit88: Yes, that would be better. Thanks

crates/router/src/types/domain/user.rs
apoorvdixit88: I thought of covering in the with the next PR, where we will give support for tenant levels user roles. We then should add tenant_id in all the level structs.

crates/router/src/services/authentication.rs
ThisIsMani: Won't this cause conflict the tenant_id in the header?

apoorvdixit88: No it will be matched with state, header and state should be same.

ThisIsMani (commented):
If there is no tenant_id in the JWT, we are using tenant_id from the state anyways. So, can't we remove the option from the UserFromToken and fill it with state's tenant_id if it is not present in the token?

crates/router/src/core/user.rs
```diff
@@ -1142,6 +1167,7 @@ pub async fn create_internal_user(
             UserStatus::Active,
         )
         .add_entity(domain::MerchantLevel {
+            tenant_id: common_utils::consts::DEFAULT_TENANT.to_string(),
```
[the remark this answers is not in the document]
apoorvdixit88: this is for internal user.

crates/router/src/services/authorization.rs
```diff
@@ -112,6 +112,15 @@ pub fn check_permission(
         )
 }
 
+pub fn check_tenant(token_tenant_id: Option<String>, header_tenant_id: &str) -> RouterResult<()> {
+    if let Some(tenant_id) = token_tenant_id {
+        if tenant_id != header_tenant_id {
+            return Err(ApiErrorResponse::InvalidTenant { tenant_id }.into());
```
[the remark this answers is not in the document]
apoorvdixit88: It was 500, 
I changed the error to 401, Invalid JWT.

apoorvdixit88:
> If there is no tenant_id in the JWT, we are using tenant_id from the state anyways. So, can't we remove the option from the UserFromToken and fill it with state's tenant_id if it is not present in the token?

I guess i did it for some reason, we will be removing Option from JWT. Keeping Option will help in stagger release, we don't need to check if tenant_id is not present in token, we can keep treat it as a valid token.

crates/router/src/core/user.rs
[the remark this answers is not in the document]
ThisIsMani: But if I hit this API in a particular tenant deployment, it will create a internal user in public tenant, which doesn't make sense.

And we may not have control over the admins of other tenants. If you don't want application to create any internal user in other tenants than public, then have a check instead but don't create a user in public tenant if the admin didn't say otherwise.

ThisIsMani:
> I guess i did it for some reason, we will be removing Option from JWT. Keeping Option will help in stagger release, we don't need to check if tenant_id is not present in token, we can keep treat it as a valid token.

Yeah, you can keep the type in `AuthToken` optional, I was talking about the `UserFromToken` which doesn't affect stagger. Anyways I thought I'd just let you know, you don't have to change it.

ThisIsMani (approved):

tsdk02 (approved):

jagan-jaya (approved):

crates/router/src/services/authentication.rs
```diff
@@ -189,6 +190,7 @@ pub struct SinglePurposeToken {
...
```
dracarys18, line 193: Are we ensuring length this is below 64 anywhere?

apoorvdixit88: For now we are just ensuing it from DB, if the length is more than 64 charater  we will get error while insertion (signing up, creating tenant_admin etc) 
`value too long for type character varying(64)`

dracarys18 (approved):

SanchithHegde (approved):

PR #6523, opened 2024-11-10
fix(webhooks): add support for updating mandate details in webhooks flow 

srujanchikke:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

For Fiuu, the mandate ID is updated via webhooks in the case of 3DS payments. The payment webhook body was not consumed when the payment was successful, causing intermittent issues in the sandbox environment. This PR adds support to address this issue.

**more detailed issue description :** 
- Fiuu mandate id is consumed from webhooks
- Incase of 3ds payment, on returning to hs return url we do a psync call
- If psync happends before we get notification from connector, we don't consume the webhook. Hence we will never have connector mandate id in our system.
- To consume the webhook, we can add logic in webhooks core that after psync check for the source  verification & webhook event based on that get mandate details from webhook and update it in the payment method & payment attempt.

### Additional Changes

## How did you test it?

Cannot be tested directly this is an intermediate issue in sandbox. So please test this flow multiple time and mandate id should be stored in CIT transaction against both payment attempt & payment method for the transaction.

## Checklist


AkshayaFoiger (dismissed):

crates/hyperswitch_connectors/src/connectors/fiuu.rs
```diff
@@ -890,4 +890,46 @@ impl webhooks::IncomingWebhook for Fiuu {
...
+    {
+        let header = utils::get_header_key_value("content-type", request.headers)?;
+        let payload: FiuuWebhooksResponse = if header == "application/x-www-form-urlencoded" {
+            serde_urlencoded::from_bytes::<FiuuWebhooksResponse>(request.body)
+                .change_context(errors::ConnectorError::WebhookResourceObjectNotFound)?
+        } else {
+            request
+                .body
+                .parse_struct("fiuu::FiuuWebhooksResponse")
+                .change_context(errors::ConnectorError::WebhookResourceObjectNotFound)?
```
sai-harsha-vardhan, line 907:
```suggestion
                .change_context(errors::ConnectorError::WebhookBodyDecodingFailed)?
```

crates/hyperswitch_connectors/src/connectors/fiuu.rs
```diff
@@ -890,4 +890,46 @@ impl webhooks::IncomingWebhook for Fiuu {
...
+        };
+        match payload.clone() {
+            FiuuWebhooksResponse::FiuuWebhookPaymentResponse(webhook_payment_response) => {
```
sai-harsha-vardhan, line 910: We can directly parse it into FiuuWebhookPaymentResponse

crates/router/src/core/webhooks/incoming.rs
```diff
@@ -1716,3 +1741,135 @@ async fn fetch_optional_mca_and_connector(
...
+                .find_payment_method(
+                    key_manager_state,
+                    key_store,
+                    payment_method_id,
+                    merchant_account.storage_scheme,
+                )
+                .await
+                .to_not_found_response(errors::ApiErrorResponse::PaymentMethodNotFound)?;
+
+            if payment_method_info.connector_mandate_details.is_none() {
```
sai-harsha-vardhan, line 1771: `is_none` won't work here, we should check the existence of mca_id in the hashmap

jarnura (approved):

sai-harsha-vardhan (dismissed):

jarnura (approved):

sai-harsha-vardhan (approved):

deepanshu-iiitu (approved):

PR #6533, opened 2024-11-11
feat(themes): Setup themes table

### LINKED ISSUE #6532: feat(themes): Setup DB for themes
There should a separate table for themes for storing the `theme_id`s for a particular lineage. This should be the schema:

| org_id | merchant_id | profile_id | theme_id |
| ------ | ----------- | ---------- | -------- |
| o1     | m1          | p1         | t1       |
| o1     | m1          | p2         | t2       |

This will be used for changing theme automatically when user switches any entities.

ThisIsMani:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

This PR creates a new table named `themes` in the DB. This will be used to store the `theme_id` for any lineage of `tenant_id`, `org_id`, `merchant_id` and `profile_id`.

### Additional Changes

## Motivation and Context

Closes #6532

## How did you test it?

This is an internal change and there is no API to test this currently.

I've tested the read query by creating a temporary API which uses the query created in this PR.

## Checklist


crates/common_utils/src/types.rs
```diff
@@ -3,6 +3,8 @@ pub mod keymanager;
 
 /// Enum for Authentication Level
 pub mod authentication;
+/// Enum for Theme Lineage
+pub mod theme;
```
SanchithHegde, line 7: I'd prefer to avoid having this in `common_utils` crate if possible, since this crate is either being pulled as a direct or a transitive dependency by a good number of our repositories. Can we move this to `common_enums` instead?

migrations/2024-11-06-121933_setup-themes-table/up.sql
```diff
@@ -0,0 +1,17 @@
+-- Your SQL goes here
+CREATE TABLE IF NOT EXISTS themes (
+    theme_id VARCHAR(64) PRIMARY KEY,
+    tenant_id VARCHAR(64) NOT NULL,
+    org_id VARCHAR(64),
+    merchant_id VARCHAR(64),
+    profile_id VARCHAR(64),
+    created_at TIMESTAMP NOT NULL DEFAULT now(),
+    last_modified_at TIMESTAMP NOT NULL DEFAULT now()
```
SanchithHegde, line 9: Let's avoid the defaults on the database side, and have the application always provide it. This would also help prevent issues with the timestamps being generated based on the database server's time zone configuration.

crates/diesel_models/src/user/theme.rs
```diff
@@ -0,0 +1,29 @@
...
+    pub theme_id: String,
+    pub tenant_id: String,
+    pub org_id: Option<id_type::OrganizationId>,
+    pub merchant_id: Option<id_type::MerchantId>,
+    pub profile_id: Option<id_type::ProfileId>,
+    pub created_at: PrimitiveDateTime,
+    pub last_modified_at: PrimitiveDateTime,
+}
+
+#[derive(router_derive::Setter, Clone, Debug, Insertable, router_derive::DebugAsDisplay)]
```
SanchithHegde, line 19: Do we need the `Setter` derive here?

crates/diesel_models/src/query/user/theme.rs
```diff
@@ -0,0 +1,95 @@
...
+            Self::lineage_filter(lineage),
+        )
+        .await
+    }
+
+    pub async fn delete_by_theme_id_and_lineage(
+        conn: &PgPooledConn,
+        theme_id: String,
+        lineage: ThemeLineage,
+    ) -> StorageResult<Self> {
```
SanchithHegde, line 86: Since `theme_id` is the primary key, do we necessarily need the other information though? Can't we delete it using the `theme_id` alone?

crates/common_utils/src/types.rs
ThisIsMani: Yeah, I'd also like to put in `common_enums` and tried this first but `id_type` is defined in `common_utils` and it is not available in `common_enums`.

I want this enum to be available in DB files and API files and only this is the common dependency where I can add something like this.

Let me know, if there is any other crate where I can place this.

apoorvdixit88 (approved):
LGTM ⭐

crates/diesel_models/src/user/theme.rs
ThisIsMani: Not needed, removed it.

crates/diesel_models/src/query/user/theme.rs
SanchithHegde: Fair enough.

SanchithHegde (approved):

crates/diesel_models/src/user/theme.rs
```diff
@@ -0,0 +1,29 @@
...
+    pub org_id: Option<id_type::OrganizationId>,
+    pub merchant_id: Option<id_type::MerchantId>,
+    pub profile_id: Option<id_type::ProfileId>,
+    pub created_at: PrimitiveDateTime,
+    pub last_modified_at: PrimitiveDateTime,
+}
+
+#[derive(Clone, Debug, Insertable, router_derive::DebugAsDisplay)]
+#[diesel(table_name = themes)]
+pub struct ThemeNew {
```
racnan, line 21: do we need this type or can we use `Theme` type above for insert as well.

racnan (approved):

crates/diesel_models/src/user/theme.rs
ThisIsMani: Keeping this separate for consistency.

PR #6546, opened 2024-11-12
refactor(core): add profile_id for default_fallback api

prajjwalkumar17:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [x] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

This will refactor the route `routing/default/profile`
Previously the output for this route was the default fallback connectors from all the profiles, 
After this refactor it will have the fallback connectors only for the mentioned profile. If the `profile_id` is provided by the authentication layer, if not it will behave as previous by taking the `merchant_id` from auth layer.

Moreover we have removed 2 routes from backed as these were not being used(Dahboard as well as internal apis), hence were redundant.
GET `routing/default`
POST `routing/default`

### Additional Changes

## Motivation and Context

Required by our Dashboard to move the api's under profile level

## How did you test it?

```
curl --location 'http://127.0.0.1:8080/routing/default/profile' \
--header 'Authorization: Bearer token
```
This should have the connectors from the mentioned profile only.
```
[
    {
        "connector": "bankofamerica",
        "merchant_connector_id": "mca_ADICIF9zh09TjnpUgusU"
    }
]
```

## Checklist

### DIFF
--- a/crates/router/src/core/routing.rs
+++ b/crates/router/src/core/routing.rs
@@ -917,10 +917,9 @@ pub async fn retrieve_default_routing_config(
 ) -> RouterResponse<Vec<routing_types::RoutableConnectorChoice>> {
     metrics::ROUTING_RETRIEVE_DEFAULT_CONFIG.add(&metrics::CONTEXT, 1, &[]);
     let db = state.store.as_ref();
-    let id = profile_id.map(|profile_id| 
-        profile_id.get_string_repr().to_owned())
-        .unwrap_or_else(|| 
-            merchant_account.get_id().get_string_repr().to_string());
+    let id = profile_id
+        .map(|profile_id| profile_id.get_string_repr().to_owned())
+        .unwrap_or_else(|| merchant_account.get_id().get_string_repr().to_string());
 
     helpers::get_merchant_default_config(db, &id, transaction_type)
         .await
--- a/crates/router/src/core/routing.rs
+++ b/crates/router/src/core/routing.rs
@@ -912,15 +912,13 @@ pub async fn retrieve_default_fallback_algorithm_for_profile(
 pub async fn retrieve_default_routing_config(
     state: SessionState,
     profile_id: Option<common_utils::id_type::ProfileId>,
+    merchant_account: domain::MerchantAccount,
     transaction_type: &enums::TransactionType,
 ) -> RouterResponse<Vec<routing_types::RoutableConnectorChoice>> {
-    let profile_id = profile_id.ok_or(errors::ApiErrorResponse::GenericNotFoundError {
-        message: "profile_id not found".to_string(),
-    })?;
     metrics::ROUTING_RETRIEVE_DEFAULT_CONFIG.add(&metrics::CONTEXT, 1, &[]);
     let db = state.store.as_ref();
-
-    helpers::get_merchant_default_config(db, profile_id.get_string_repr(), transaction_type)
+    if let Some(profile_id) = profile_id {
+        helpers::get_merchant_default_config(db, profile_id.get_string_repr(), transaction_type)
         .await
         .map(|conn_choice| {
             metrics::ROUTING_RETRIEVE_DEFAULT_CONFIG_SUCCESS_RESPONSE.add(
@@ -930,6 +928,18 @@ pub async fn retrieve_default_routing_config(
             );
             service_api::ApplicationResponse::Json(conn_choice)
         })
+    } else {
+        helpers::get_merchant_default_config(db, merchant_account.get_id().get_string_repr(), transaction_type)
+        .await
+        .map(|conn_choice| {
+            metrics::ROUTING_RETRIEVE_DEFAULT_CONFIG_SUCCESS_RESPONSE.add(
+                &metrics::CONTEXT,
+                1,
+                &[],
+            );
+            service_api::ApplicationResponse::Json(conn_choice)
+        })
+    }
 }
 
 #[cfg(feature = "v2")]
--- a/crates/router/src/core/routing.rs
+++ b/crates/router/src/core/routing.rs
@@ -909,26 +909,26 @@ pub async fn retrieve_default_fallback_algorithm_for_profile(
 }
 
 #[cfg(feature = "v1")]
-
 pub async fn retrieve_default_routing_config(
     state: SessionState,
-    merchant_account: domain::MerchantAccount,
+    profile_id: common_utils::id_type::ProfileId,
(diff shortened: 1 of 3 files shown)


crates/router/src/routes/routing.rs
```diff
@@ -557,27 +557,31 @@ pub async fn routing_retrieve_default_config(
...
+    let path = path.into_inner();
     Box::pin(oss_api::server_wrap(
         Flow::RoutingRetrieveDefaultConfig,
         state,
         &req,
-        (),
-        |state, auth: auth::AuthenticationData, _, _| {
-            routing::retrieve_default_routing_config(state, auth.merchant_account, transaction_type)
+        path.clone(),
+        |state, _, profile_id, _| {
```
ThisIsMani, line 569: Use JWT for profile id.

ThisIsMani (changes requested):

crates/router/src/routes/app.rs
```diff
@@ -759,22 +759,14 @@ impl Routing {
...
-                    .route(web::post().to(|state, req, payload| {
-                        routing::routing_update_default_config(
-                            state,
-                            req,
-                            payload,
-                            &TransactionType::Payment,
-                        )
-                    })),
+                web::resource("/default").route(web::post().to(|state, req, payload| {
+                    routing::routing_update_default_config(
```
Aprabhat19, line 763: Can we check if this is also not being used then maybe we can remove this

prajjwalkumar17: @ThisIsMani are we using this route in Dashboard as internally there is no dependency on this ?

crates/router/src/core/routing.rs
```diff
@@ -909,26 +909,29 @@ pub async fn retrieve_default_fallback_algorithm_for_profile(
...
-
 pub async fn retrieve_default_routing_config(
     state: SessionState,
-    merchant_account: domain::MerchantAccount,
+    profile_id: Option<common_utils::id_type::ProfileId>,
     transaction_type: &enums::TransactionType,
 ) -> RouterResponse<Vec<routing_types::RoutableConnectorChoice>> {
+    let profile_id = profile_id.ok_or(errors::ApiErrorResponse::GenericNotFoundError {
+        message: "profile_id not found".to_string(),
+    })?;
```
ThisIsMani, line 919: Can we also handle the case where `profile_id` is not present.

Is can be if...else, and if `profile_id` is not present it will work like before.

crates/router/src/routes/app.rs
ThisIsMani: Please check with @JeevaRamu0104 once.

crates/router/src/routes/routing.rs
```diff
@@ -565,17 +565,8 @@ pub async fn routing_retrieve_default_config(
         &req,
         (),
         |state, auth: auth::AuthenticationData, _, _| {
-            routing::retrieve_default_routing_config(state, auth.merchant_account, transaction_type)
+            routing::retrieve_default_routing_config(state, auth.profile_id, transaction_type)
```
ThisIsMani, line 568: Also pass `auth.merchant_account`. It will be used if `profile_id` is not present.

crates/router/src/routes/app.rs
prajjwalkumar17: This is checked out with @JeevaRamu0104 , we don't have these endpoints being used in dashboard.

crates/router/src/core/routing.rs
```diff
@@ -909,26 +909,43 @@ pub async fn retrieve_default_fallback_algorithm_for_profile(
...
+                    &metrics::CONTEXT,
+                    1,
+                    &[],
+                );
+                service_api::ApplicationResponse::Json(conn_choice)
+            })
+    } else {
+        helpers::get_merchant_default_config(
+            db,
+            merchant_account.get_id().get_string_repr(),
```
Aprabhat19, line 934: Why are we having support for merchant_id, I think we might not need that

prajjwalkumar17: Can you check these two:
So basically we are doing this, to make dashboard backwards compatible.
https://github.com/juspay/hyperswitch/pull/6546#discussion_r1839852600
https://github.com/juspay/hyperswitch/pull/6546#discussion_r1839848057

ThisIsMani (dismissed):
Can you update the description with all the APIs that are changed because of this PR. Let me know once done.

crates/router/src/core/routing.rs
```diff
@@ -909,26 +909,43 @@ pub async fn retrieve_default_fallback_algorithm_for_profile(
...
-    .await
-    .map(|conn_choice| {
-        metrics::ROUTING_RETRIEVE_DEFAULT_CONFIG_SUCCESS_RESPONSE.add(&metrics::CONTEXT, 1, &[]);
-        service_api::ApplicationResponse::Json(conn_choice)
-    })
+    if let Some(profile_id) = profile_id {
+        helpers::get_merchant_default_config(db, profile_id.get_string_repr(), transaction_type)
+            .await
+            .map(|conn_choice| {
+                metrics::ROUTING_RETRIEVE_DEFAULT_CONFIG_SUCCESS_RESPONSE.add(
```
Aprabhat19, line 924: Can we move this outside the if block ?

JeevaRamu0104 (dismissed):

crates/router/src/core/routing.rs
```diff
@@ -909,26 +909,43 @@ pub async fn retrieve_default_fallback_algorithm_for_profile(
...
+                    &metrics::CONTEXT,
+                    1,
+                    &[],
```
Aprabhat19, line 927: Can we handle the functions in the if and else block in a common manner .

prajjwalkumar17: Okay so let's consider this situation when the function `get_merchant_default_config`, returns error.
So the metrics(`ROUTING_RETRIEVE_DEFAULT_CONFIG_SUCCESS_RESPONSE`) will be pushed even when we are returning an error, which shouldn't be the case.

So considering this situation the metrics is being pushed only when we have a success confirmation.

Aprabhat19: No I am talking about the function `get_merchant_default_config` move it out of the if else block

Aprabhat19 (approved):

NishantJoshi00 (approved):

PR #6555, opened 2024-11-13
refactor(payment_methods_v2): rename `payment_method` and `payment_method_type` fields and use concrete type for `payment_method_data`

SanchithHegde:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [x] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

This PR involves the following changes in payment methods v2 code:

- Renames the `payment_method` field to `payment_method_type`, and `payment_method_type` to `payment_method_subtype`. This change affect the database columns itself, the diesel models, the domain models and the API models.
- Updates the `payment_method_data` field in the domain model to have the concrete type wrapped by `EncryptedJsonType`, instead of the existing `serde_json::Value` type.
  - The `EncryptedJsonType` was also introduced in this PR. It adds support to use a concrete type in domain models, while handling the JSON serialization and deserialization before encryption and after decryption of data, respectively. 

In addition, this PR removes the `PaymentMethodList` struct from the API models, which isn't being used anywhere.

By the way, I'm open to better name suggestions for the newly introduced things!

### Additional Changes

- This PR renames database columns in payment methods v2 code.
- This PR renames fields in payment methods v2 API models.
- This PR removes an unused type from the API models crate.

## Motivation and Context

- The renaming of columns was suggested in https://github.com/juspay/hyperswitch/pull/6433#discussion_r1828880058. This also follows the changes done in the payment attempt v2 types for these two fields.
- Using of the concrete type in the domain models helps prevent JSON deserialization happening multiple times in different parts of the codebase.

## How did you test it?

This PR mostly affects the payment methods v2 code, with focus on the customer payment methods list v2 code. I haven't been able to test these changes, there are still a couple of panics remaining in the customer payment methods list v2 code. 

Ideally, nothing in the payment methods v1 code should be affected and existing Postman and Cypress tests should pass.

## Checklist


crates/api_models/src/payment_methods.rs
```diff
@@ -1541,18 +1541,6 @@ pub struct PaymentMethodListResponse {
...
-        {
-            "payment_method": "wallet",
-            "payment_experience": null,
-            "payment_method_issuers": [
-                "labore magna ipsum",
-                "aute"
-            ]
-        }
-    ]
-    ))]
```
SanchithHegde, line 1555: Let me know if we need an example here.

crates/router/src/core/customers.rs
```diff
@@ -621,7 +621,7 @@ impl CustomerDeleteBridge for customers::GlobalId {
             // check this in review
             Ok(customer_payment_methods) => {
                 for pm in customer_payment_methods.into_iter() {
-                    if pm.payment_method == Some(enums::PaymentMethod::Card) {
+                    if pm.payment_method_type_v2 == Some(enums::PaymentMethod::Card) {
```
SanchithHegde, line 624: This field is being accessed from the diesel model, which doesn't have methods to access these fields (only domain models do as of now). Let me know if I should add methods to the diesel models to access these fields as well.

crates/hyperswitch_domain_models/src/type_encryption.rs
```diff
@@ -562,14 +565,242 @@ mod encrypt {
         }
     }
 
+    impl<T> EncryptedJsonType<T>
+    where
+        T: std::fmt::Debug + Clone + serde::Serialize + serde::de::DeserializeOwned,
```
jarnura, line 570: What are case T required Debug? is that necessary.

crates/hyperswitch_domain_models/src/type_encryption.rs
```diff
@@ -562,14 +565,242 @@ mod encrypt {
...
+                .as_slice()
+                .parse_struct::<T>(std::any::type_name::<T>())
+                .map(|result| Secret::new(Self::from(result)))
+                .attach_printable("Failed to JSON deserialize data after decryption")
+        }
+    }
+
+    #[async_trait]
+    impl<
+            T: std::fmt::Debug + Clone + serde::Serialize + serde::de::DeserializeOwned + Send,
```
jarnura, line 596: What are case T required Debug? is that necessary.

jarnura (changes requested):

crates/hyperswitch_domain_models/src/type_encryption.rs
SanchithHegde: The usages of the `Encode` trait methods (in this case `encode_to_vec()`) needs us to implement `Debug`.
Moreover, if the `PaymentMethod` domain model derives `Debug` and one of it's fields has a `EncryptedJsonType<T>`, we'd also need `Debug` to be implemented on `EncryptedJsonType` and thus `T`.

jarnura (dismissed):

crates/api_models/src/events/payment.rs
```diff
@@ -196,13 +196,26 @@ impl ApiEventMetric for PaymentsResponse {
...
             payment_method: self.payment_method,
             payment_method_type: self.payment_method_type,
         })
     }
+
+    #[cfg(all(feature = "v2", feature = "payment_methods_v2"))]
+    fn get_api_event_type(&self) -> Option<ApiEventsType> {
+        Some(ApiEventsType::PaymentMethod {
+            payment_method_id: self.payment_method_id.clone(),
+            payment_method: self.payment_method_type,
```
Narayanbhat166, line 215: are we not changing the field name her?

crates/api_models/src/payment_methods.rs
```diff
@@ -852,6 +853,7 @@ pub struct CardDetailFromLocker {
 #[cfg(all(feature = "v2", feature = "payment_methods_v2"))]
 #[derive(Debug, serde::Deserialize, serde::Serialize, Clone, ToSchema)]
 pub struct CardDetailFromLocker {
+    #[schema(value_type = CountryAlpha2)]
```
Narayanbhat166, line 856:
```suggestion
    #[schema(value_type = Option<CountryAlpha2>)]
```

crates/api_models/src/events/payment.rs
SanchithHegde: Since renaming the field affects the analytics pipeline, we'd have to include ClickHouse migrations as well.
For now, I'm not including v2 changes in the analytics pipeline.

CC: @ShivanshMathurJuspay @tsdk02

crates/api_models/src/payment_methods.rs
SanchithHegde: Good catch, thanks!

crates/hyperswitch_domain_models/src/type_encryption.rs
```diff
@@ -562,14 +565,242 @@ mod encrypt {
...
+            V: crypto::DecodeMessage + crypto::EncodeMessage + Send + 'static,
+            S: masking::Strategy<EncryptedJsonType<T>> + Send + Sync,
+        > TypeEncryption<EncryptedJsonType<T>, V, S>
+        for crypto::Encryptable<Secret<EncryptedJsonType<T>, S>>
+    {
+        #[instrument(skip_all)]
```
dracarys18, line 602: PR looks good, can we add a comment somewhere to not remove this skip_all here?

jarnura (approved):

crates/api_models/src/events/payment.rs
ShivanshMathurJuspay: Are we renaming payment_method to payment_method_type ?

dracarys18 (approved):

crates/hyperswitch_domain_models/src/payment_methods.rs
```diff
@@ -77,9 +78,11 @@ pub struct PaymentMethod {
     pub merchant_id: common_utils::id_type::MerchantId,
     pub created_at: PrimitiveDateTime,
     pub last_modified: PrimitiveDateTime,
-    pub payment_method: Option<storage_enums::PaymentMethod>,
-    pub payment_method_type: Option<storage_enums::PaymentMethodType>,
-    pub payment_method_data: OptionalEncryptableValue,
+    pub payment_method_type: Option<storage_enums::PaymentMethod>,
+    pub payment_method_subtype: Option<storage_enums::PaymentMethodType>,
+    pub payment_method_data: Option<
```
Sarthak1799, line 83: Would it be better to use a type alias here?

Sarthak1799 (approved):

crates/hyperswitch_domain_models/src/payment_methods.rs
SanchithHegde: I did add a type alias for it, I'm taking it up in my next PR.

ShivanshMathurJuspay (approved):

jarnura (approved):

tsdk02 (approved):

PR #6559, opened 2024-11-13
ci(cypress): Add Checkout Connector

likhinbopanna:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [x] CI/CD

## Description
Add Checkout Connector for Cypress Automation 

### Additional Changes

## Motivation and Context
Checkout connector was not available in the cypress automation 

## How did you test it?
Cypress Tests 

- Checkout 
image

- NMI 
image

## Checklist

### DIFF
--- a/cypress-tests/cypress/e2e/PaymentTest/00014-SaveCardFlow.cy.js
+++ b/cypress-tests/cypress/e2e/PaymentTest/00014-SaveCardFlow.cy.js
@@ -47,11 +47,7 @@ describe("Card - SaveCard payment flow test", () => {
           globalState
         );
         if (should_continue) {
-          // Don't continue if payment status is processing during auto capture
-          // Payment data is tokenized only after payment is successful
-          let notProcessing = res_data?.body?.status != "processing";
-          should_continue =
-            notProcessing && utils.should_continue_further(res_data);
+          should_continue = utils.should_continue_further(res_data);
         }
       });
 
--- a/cypress-tests/cypress/e2e/PaymentUtils/Checkout.js
+++ b/cypress-tests/cypress/e2e/PaymentUtils/Checkout.js
@@ -1,0 +1,315 @@
+const successfulNo3DSCardDetails = {
+  card_number: "4242424242424242",
+  card_exp_month: "01",
+  card_exp_year: "25",
+  card_holder_name: "joseph Doe",
+  card_cvc: "123",
+};
+
+const successfulThreeDSTestCardDetails = {
+  card_number: "4242424242424242",
+  card_exp_month: "01",
+  card_exp_year: "25",
+  card_holder_name: "joseph Doe",
+  card_cvc: "123",
+};
+
+export const connectorDetails = {
+  card_pm: {
+    PaymentIntent: {
+      Request: {
+        currency: "USD",
+        customer_acceptance: null,
+        setup_future_usage: "on_session",
+      },
+      Response: {
+        status: 200,
+        body: {
+          status: "requires_payment_method",
+        },
+      },
+    },
+    "3DSManualCapture": {
+      Request: {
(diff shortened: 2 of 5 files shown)


cypress-tests/cypress/e2e/PaymentUtils/Checkout.js
```diff
@@ -0,0 +1,315 @@
...
+        },
+      },
+    },
+    "3DSManualCapture": {
+      Request: {
+        payment_method: "card",
+        payment_method_data: {
+          card: successfulThreeDSTestCardDetails,
+        },
+        currency: "USD",
```
pixincreate, line 38: remove currency field from all confirm calls.

cypress-tests/cypress/e2e/PaymentUtils/Checkout.js
```diff
@@ -0,0 +1,315 @@
...
+      Response: {
+        status: 501,
+        body: {
+          error: {
+            type: "invalid_request",
+            message: "Setup Mandate flow for Checkout is not implemented",
+            code: "IR_00",
+          },
+        },
+      },
```
pixincreate, line 230: should we move this to `commons` and remove the same from all configs?
we can directly use `connectorName` within `commons` to dynamically modify the error message.

cypress-tests/cypress/e2e/PaymentUtils/Checkout.js
```diff
@@ -0,0 +1,315 @@
...
+        currency: "USD",
+        setup_future_usage: "on_session",
+        customer_acceptance: {
+          acceptance_type: "offline",
+          accepted_at: "1963-05-03T04:07:52.723Z",
+          online: {
+            ip_address: "127.0.0.1",
+            user_agent: "amet irure esse",
+          },
+        },
```
pixincreate, line 305: this can be moved to a constant object so that it can be reused.

cypress-tests/cypress/e2e/PaymentTest/00014-SaveCardFlow.cy.js
```diff
@@ -47,11 +47,7 @@ describe("Card - SaveCard payment flow test", () => {
           globalState
         );
         if (should_continue) {
-          // Don't continue if payment status is processing during auto capture
-          // Payment data is tokenized only after payment is successful
-          let notProcessing = res_data?.body?.status != "processing";
-          should_continue =
-            notProcessing && utils.should_continue_further(res_data);
+          should_continue = utils.should_continue_further(res_data);
```
pixincreate, line 50: why was this reverted? this was added by @kashif-m to add support for WorldPay. Removing this would result in WorldPay tests failing.

likhinbopanna: If status is processing after the payment confirm we cannot skip the next test cases because Psync call would make payment success and save card flow to work , we need to handle worldpay in a different way.

cypress-tests/cypress/e2e/PaymentUtils/Checkout.js
Gnanasundari24: @pixincreate Let's add this in the internal thread, and we can handle changes for all connectors in a separate PR. Updating all connectors at once would require testing each one individually, so we'll treat this as a separate task

pixincreate: yes, that's what even i want to do. lets have it as a separate pr so that only that area will be focused.

pixincreate (approved):

Gnanasundari24 (approved):

PR #6569, opened 2024-11-14
ci(cypress): Add list and revoke for zero auth mandate payments

Gnanasundari24:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [x] CI/CD

## Description
 Added the list mandate and revoke mandates for zero auth mandate payment
 Removed the fiuu hardcoded checks which is invalid for few connectors

### Additional Changes

## Motivation and Context

To cover the scenarios for cypress

## How did you test it?

Screenshot 2024-11-14 at 4 21 05 PM

## Checklist

### DIFF
--- a/cypress-tests/cypress/support/commands.js
+++ b/cypress-tests/cypress/support/commands.js
@@ -1964,9 +1964,7 @@ Cypress.Commands.add(
           } else if (response.body.authentication_type === "no_three_ds") {
             if (response.body.connector === "fiuu") {
               expect(response.body.status).to.equal("failed");
-            } else {
-              expect(response.body.status).to.equal("succeeded");
-            }
+            } 
           } else {
             throw new Error(
               `Invalid authentication type ${response.body.authentication_type}`
@@ -2046,9 +2044,7 @@ Cypress.Commands.add(
           } else if (response.body.authentication_type === "no_three_ds") {
             if (response.body.connector === "fiuu") {
               expect(response.body.status).to.equal("failed");
-            } else {
-              expect(response.body.status).to.equal("succeeded");
-            }
+            } 
           } else {
             throw new Error(
               `Invalid authentication type ${response.body.authentication_type}`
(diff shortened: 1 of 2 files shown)


likhinbopanna (approved):

cypress-tests/cypress/support/commands.js
```diff
@@ -1964,9 +1964,7 @@ Cypress.Commands.add(
           } else if (response.body.authentication_type === "no_three_ds") {
             if (response.body.connector === "fiuu") {
               expect(response.body.status).to.equal("failed");
-            } else {
-              expect(response.body.status).to.equal("succeeded");
-            }
+            }
```
pixincreate, line 1967: what will be the expected response? looks like we're not at all validating this going forward...?

pixincreate (commented):
other than that, lgtm

cypress-tests/cypress/support/commands.js
Gnanasundari24: For fiuu it will be failed and for other connector we have to make this function more generic so that it will take from the config.js file

Gnanasundari24: That's why I have removed that check. It shouldn't work if the status is processing

pixincreate: Can that be added back as well along side `succeeded` for safety?

Gnanasundari24: When i add the succeeded state other connectors which are going processing are failing which is not correct. So we have to fix it fully

pixincreate: Yes, that's what I said.

For now, add a check there instead of only having an `else` condition.

Something like:

```js
else if (response.body.status === "succeeded")
// assertion
else if (response.body.status === "processing")
// assertion
else
// throw an error?
```

Gnanasundari24: @pixincreate Its wrong know. We have to remove that fiuu condition only in the future. I ran all the connector and it wont break anything.

pixincreate (approved):

PR #6588, opened 2024-11-17
feat(cypress): add multiple creds and flags support

### LINKED ISSUE #6645: [FEAT] Multiple credential support for a connector in Cypress
## Requirement

Execute cypress tests for a connector having **multiple credentials** which can be used for different scenarios (one api key for `cards`, another for `bank redirects` and etc.,).

### Introduce new format for `creds.json`

`creds.json` should include both the formats. if a connector requires us passing **multiple credentials**, then that connector api key object **should** include multiple `connector_account_details` object all named under a generic name `connector_<number>` where `number` is an `integer`.

Refer example given below for more clarity:

```json
// connector having multiple credentials
<connector_name_1>: {
    "connector_1": {
            "auth_type": "KeyType",
            "api_key": "",
            "api_secret": "",
            "key1": "",
            "key2": ""
        },
        "metadata": {}
    },
    "connector_2": {
            "auth_type": "KeyType",
            "api_key": "",
            "api_secret": "",
            "key1": "",
            "key2": ""
        },
        "metadata": {}
    },
},
// connector with a single credential
<connector_name_2>: {
		"auth_type": "KeyType",
		"api_key": "",
		"api_secret": "",
		"key1": "",
		"key2": ""
	}
}
```

### Introduce new `object` (`Configs`) for `connector.js`

Within `connector.js`, introduce new object called as `Configs` alongside to `Request` and `Response` where user can define `flags` to achieve granular control over what test is being run.

An example implementation given below for reference:

```js
getCustomExchange({
	Configs: {
		TRIGGER_SKIP: true, // skips redirection flow from running. takes in a boolean
		DELAY: {
			STATUS: true, // flag to turn delay feature on or off. takes in a boolean
			TIMEOUT: 5000, // timeout in milliseconds
		},
		CONNECTOR_CREDENTIAL: connector_1 / connector_2 // flag to route tests to a specific profile
		// etc.,
	},
	Requst: {},
	Response: {}
}),
```

### Modify `getValueByKey` function in `Utils.js`

Validate if `connector_account_details` dot exist within `<connector>` object in creds.
If it does not, start a loop and see if `connector_account_details` exist within every object. If true, return the `connector_account_details` while setting an object as a Cypress environment flag (`MULTIPLE_CONNECTORS`) with status `true` and length. If any validation fails, directly return the object (`data[key]`).

An example implementation given below for reference:

```js
if (data && typeof data === "object" && key in data) {
	// Connector object has multiple keys
	if (typeof data[key].connector_account_details === "undefined") {
		const keys = Object.keys(data[key]);

		for (let i = 0; i < keys.length; i++) {
			const currentItem = data[key][keys[i]];

			if (currentItem.hasOwnProperty("connector_account_details")) {
				Cypress.env("MULTIPLE_CONNECTORS", {
					status: true,
					count: keys.length,
				});

				return currentItem;
			}
		}
	}

	return data[key];
} else {
	return null;
}

```

### Add a new test to MCA create call

If `MULTIPLE_CONNECTORS.status` is `TRUE`. Check `MULTIPLE_CONNECTORS.count` and create `profile` and `mca` beneath that `profile` based on the number of `count`.

An example of possible implementation given below for reference:

```js
if (Cypress.env("MULTIPLE_CONNECTORS")?.status) {
	for (let i = 0; i < Cypress.env("MULTIPLE_CONNECTORS").count; i++) {
		cy.createBusinessProfileTest(
			createBusinessProfileBody,
			globalState,
			"profile" + i // new optional fields
		);
		cy.createConnectorCallTest(
			"payment_processor",
			createConnectorBody,
			payment_methods_enabled,
			globalState,
			"profile" + i, // new optional fields
			"merchantConnector" + i // new optional fields
		);
	}
}
```

Store these created `profile_id`s and `mca`s in `globalState` for future usage.
Pass `CONNECTOR_CREDENTIAL` value as `connector_1` or `connector_2` in `<connector>.js`
In `commands.js`, `execute` these configs before a `request` has been made.
`execute` here means to make these configs work.

Preferably, make this execution of configs a function and pass the values accordingly along with trace to find from where the function has been called.

## Limitations

- Cypress cannot call itself unless a wrapper around Cypress is written (`Rustpress`?)
  - This stops us from running the entire test suite against every credentials
  - One possible work around for this is to have multiple keys (`cybersource_1`, `cybersource_2`) but this will result in a lot of confusion
- Current implementation **requires** the user to mandatorily pass a `CONNECTOR_CREDENTIAL` config to pass a different `profileId`
  - Hence, the tests will only run once but, specific tests like `incremental auth` will be forced to run against different credential
- `00025-IncrementalAuth.cy.js` can be stopped from execution based on `MULTIPLE_CONNECTORS` environment variable as it is set during run time and the file is read asynchronously when the execution starts (to address this, we will have to make a function call to see if it is set during run time and it is, then skip)

pixincreate:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [x] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [x] CI/CD

## Description

This PR introduces 2 new features:

- Multiple credentials for a connector
- Configuration flags

closes #6645 (check this issue for detailed documentation / flow / explanation)

### Additional Changes

## Motivation and Context

- We need to run tests against multiple connector api keys for a single connector
- We also need configuration flag support to selectively grant a test certain features / options

## How did you test it?

Only Cybersource with `multiple creds` (takes only from first `connector_1`):

image

Cybersource with `multiple creds` (along with `incremental auth`):

image

(incremental auth fails from the connector's end and hence 4 failures)

Stripe:

image

Wise:

image

PML:

image

have disabled incremental auth from executing:

image

trustpay refunds (multiple creds):

image

i commented iDEAL redirections to check, error is being thrown from connector's end and there's nothing much i can do:

image
image

image

## Checklist


cypress-tests/cypress/e2e/RoutingTest/00000-PriorityRouting.cy.js
```diff
@@ -109,12 +110,13 @@ describe("Priority Based Routing Test", () => {
...
         globalState
       );
+
       if (should_continue)
         should_continue = utils.should_continue_further(res_data);
     });
 
     it("retrieve-payment-call-test", () => {
-      cy.retrievePaymentCallTest(globalState);
+      cy.retrievePaymentCallTest(globalState, null);
```
pixincreate, line 119: i'm not okay with passing `null` for `configs` as it is unsupported here.

cypress-tests/cypress/e2e/PaymentTest/00015-ZeroAuthMandate.cy.js
```diff
@@ -122,39 +142,38 @@ describe("Card - SingleUse Mandates flow test", () => {
...
 
     it("Retrieve Payment Call Test", () => {
-      cy.retrievePaymentCallTest(globalState);
+      let data = getConnectorDetails(globalState.get("connectorId"))["card_pm"][
+        "ZeroAuthConfirmPayment"
+      ];
+
+      let configs = validateConfig(data["Configs"]);
+
+      cy.retrievePaymentCallTest(globalState, configs);
```
jagan-jaya, line 176:
```suggestion
      let data = getConnectorDetails(globalState.get("connectorId"))["card_pm"][
        "ZeroAuthConfirmPayment"
      ];
      cy.retrievePaymentCallTest(globalState, data);
```
here also pass only the `data` to  retrievePaymentCallTest, you are already doing validateConfig inside

Change it in other places also

jagan-jaya (changes requested):
Changes looks good to me, 
Please do the minor refactoring mentioned and check cypress failure

cypress-tests/cypress/e2e/PaymentTest/00015-ZeroAuthMandate.cy.js
pixincreate: oh, i guess i missed it here. will get it done and update the flow in the attached issue.

pixincreate:
to-do:

- update credentials
- disable incremental auth (it is being failed from connector's end and there's nothing much i can do about it)
- add ci check to check formatting

jagan-jaya (dismissed):

jagan-jaya (approved):

cypress-tests/.husky/pre-commit
```diff
@@ -0,0 +1,3 @@
+# .husky/pre-commit
+npm run format:check && npm run lint
+git update-index --again
```
SanchithHegde, line 3: Why do we need to run `git update-index`?

.github/workflows/cypress-tests-runner.yml
```diff
@@ -60,6 +60,41 @@ jobs:
...
+        with:
+          node-version: '20'
+          cache: 'npm'
+
+      - name: Install Cypress and dependencies
+        if: ${{ env.RUN_TESTS == 'true' }}
+        run: |
+          npm ci --prefix ./cypress-tests
+
+      - name: Check formatting
```
SanchithHegde, line 75: I'd prefer this being run in a separate job, which is not included as a required check for PR merges.

.github/workflows/cypress-tests-runner.yml
```diff
@@ -60,6 +60,41 @@ jobs:
...
+        if: ${{ env.RUN_TESTS == 'true' }}
+        run: |
+          npm run format:check --prefix cypress-tests && npm run lint --prefix cypress-tests
+
+          if [ $? -ne 0 ]; then
```
SanchithHegde, line 80: I doubt this would work as expected: I think it would exit the step on first failure.

.github/workflows/cypress-tests-runner.yml
```diff
@@ -60,6 +60,41 @@ jobs:
...
+
+            # Run prettier to fix formatting
+            npm run format --prefix cypress-tests
+
+            # Run eslint to auto-fix linting issues
+            # This will not fix logical issues, but will fix formatting issues
+            npm run lint --prefix cypress-tests -- --fix
+
+            # Add changes to git and commit
+            git add .
```
SanchithHegde, line 93:
```suggestion
            git add cypress-tests
```

.github/workflows/cypress-tests-runner.yml
```diff
@@ -60,6 +60,41 @@ jobs:
...
+            git commit --message 'chore: run formatter and lint fixes'
```
SanchithHegde, line 94:
```suggestion
            git commit --message 'chore(cypress): run formatter and address lints'
```

.github/workflows/cypress-tests-runner.yml
```diff
@@ -60,6 +60,41 @@ jobs:
...
+            git config --local user.email '148525504+hyperswitch-bot[bot]@users.noreply.github.com'
+
+            # Run prettier to fix formatting
+            npm run format --prefix cypress-tests
+
+            # Run eslint to auto-fix linting issues
+            # This will not fix logical issues, but will fix formatting issues
+            npm run lint --prefix cypress-tests -- --fix
+
+            # Add changes to git and commit
```
SanchithHegde, line 92: Committing changes would require you to checkout the branch correctly, and not a merge commit of the PR branch and base branch. The checkout step would have to be updated to handle this.

.github/workflows/cypress-tests-runner.yml
```diff
@@ -22,6 +22,45 @@ env:
   RUST_MIN_STACK: 10485760
 
 jobs:
+  formatter:
```
SanchithHegde, line 25: This job should only run if we're running against a PR (we should not be modifying anything in case of merge queue). Can add an `if` attribute on the entire job.

.github/workflows/cypress-tests-runner.yml
```diff
@@ -22,6 +22,45 @@ env:
...
+        shell: bash
+        run: |
+          npm run format --prefix cypress-tests
+          npm run lint --prefix cypress-tests -- --fix
+
+          if ! git diff --exit-code --quiet -- cypress-tests; then
+            echo "::notice::Cypress formatting and lint check failed"
+
+            git config --local user.name 'hyperswitch-bot[bot]'
+            git config --local user.email '148525504+hyperswitch-bot[bot]@users.noreply.github.com'
```
SanchithHegde, line 57: We shouldn't be using the bot name and email always, we can use that only in case the PR is opened from the same repo.

.github/workflows/cypress-tests-runner.yml
```diff
@@ -22,6 +22,45 @@ env:
...
+
+            git add cypress-tests
+            git commit --message 'chore(cypress): run formatter and address lints'
+            git push
```
SanchithHegde, line 61: Have you confirmed if the commit and push steps work correctly?

pixincreate: Yes, this [commit](https://github.com/juspay/hyperswitch/pull/6588/commits/01f41c68b8477b848c788b6b164255c62cc0ab52) was actually done by that (from local, not from action).

SanchithHegde: I'd prefer the steps be tested from the action as well.

.github/workflows/cypress-tests-runner.yml
pixincreate: added a check alongside https://github.com/juspay/hyperswitch/pull/6588#discussion_r1864937159

.github/workflows/cypress-tests-runner.yml
```diff
@@ -22,6 +22,46 @@ env:
...
+    if: ${{ ((github.event_name == 'pull_request') && (github.event.pull_request.head.repo.full_name == github.event.pull_request.base.repo.full_name)) }}
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+        with:
+          ref: ${{ github.head_ref || github.ref }}
+
+      - name: Install Node.js
+        if: ${{ env.RUN_TESTS == 'true' }}
```
SanchithHegde, line 36: I don't think you need this `if` condition since you've added the check on the job level.

.github/workflows/cypress-tests-runner.yml
```diff
@@ -22,6 +22,46 @@ env:
...
+        uses: actions/setup-node@v4
+        with:
+          node-version: '20'
+          cache: 'npm'
```
SanchithHegde, line 40: Given the choice, I'd prefer disabling this in favor of caching Rust dependencies.

jagan-jaya (dismissed):

.github/workflows/cypress-tests-runner.yml
```diff
@@ -22,6 +22,66 @@ env:
...
+        run: |
+          npm ci --prefix ./cypress-tests
+
+      - name: Check formatting for forked pull requests
+        if: ${{ github.event.pull_request.head.repo.full_name != github.event.pull_request.base.repo.full_name }}
+        shell: bash
+        run: |
+          # Checks return exit code 1 if failed and hence we use `|| true` to avoid failing the check
+          npm run format:check --prefix cypress-tests || true
+          npm run lint --prefix cypress-tests || true
```
SanchithHegde, line 65: So this step would always pass and we wouldn't know of failures at all?

SanchithHegde (dismissed):

jagan-jaya (approved):

cypress-tests/cypress/e2e/RoutingTest/00000-PriorityRouting.cy.js
```diff
@@ -157,53 +147,48 @@ describe("Priority Based Routing Test", () => {
...
         globalState
       );
       if (should_continue)
-        should_continue = utils.should_continue_further(res_data);
+        should_continue = utils.should_continue_further(data);
     });
 
     it("retrieve-payment-call-test", () => {
-      cy.retrievePaymentCallTest(globalState);
+      cy.retrievePaymentCallTest(globalState, null);
```
likhinbopanna, line 191: Why are you passing `null` here ?

cypress-tests/cypress/e2e/PaymentTest/00004-NoThreeDSAutoCapture.cy.js
```diff
@@ -25,46 +25,43 @@ describe("Card - NoThreeDS payment flow test", () => {
...
+      cy.confirmCallTest(fixtures.confirmBody, data, true, globalState);
+
       if (should_continue)
-        should_continue = utils.should_continue_further(res_data);
+        should_continue = utils.should_continue_further(data);
     });
 
     it("retrieve-payment-call-test", () => {
-      cy.retrievePaymentCallTest(globalState);
+      const data = getConnectorDetails(globalState.get("connectorId"))[
```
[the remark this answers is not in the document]
pixincreate: glad you asked. data exist only to fetch the `Configs` and nothing else. in the `support/commands.js` (image attached below), we're extracting `Configs` from `data`. this is done specifically keeping `CONNECTOR_CREDENTIAL` in mind so that we can pass `mca_id` directly which is used in verification (assertion check). 

image

the entire field can be removed if we stop doing `mca_id` check.

cypress-tests/cypress/e2e/PaymentTest/00025-ConfigTest.cy.js
```diff
@@ -19,10 +19,10 @@ describe("Config Tests", () => {
   context(
     "Update collect_billing_details_from_wallet_connector to true and verifying in payment method list, this config should be true",
     () => {
-      let should_continue = true;
+      let shouldContinue = true;
```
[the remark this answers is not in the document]
pixincreate: implementing `camelCase` format just to align with [javascript standards](https://www.w3schools.com/js/js_conventions.asp). will do the same changes everywhere going forward and encourage everyone to adopt the same.

cypress-tests/cypress/e2e/PaymentUtils/Cybersource.js
```diff
@@ -330,7 +396,26 @@ export const connectorDetails = {
         },
       },
     },
+    IncrementalAuth: {
+      Request: {
```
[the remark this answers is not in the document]
pixincreate: incremental auth enpoint accepts neither `profileid`, nor `mca_id`. so i found it unnecessary have `CONNECTOR_CREDENTIAL` config added there. also, there exist no need for us to have any other configs at the moment

cypress-tests/cypress/e2e/PaymentTest/00025-ConfigTest.cy.js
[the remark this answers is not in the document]
likhinbopanna: Yes , because everywhere else is snake_case , you could've changed everything , if you really wanted to change

cypress-tests/cypress/e2e/PaymentTest/00004-NoThreeDSAutoCapture.cy.js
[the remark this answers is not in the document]
likhinbopanna: Are we not verifying the payment status ?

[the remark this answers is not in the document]
pixincreate: yes. remove this mca_id check?

likhinbopanna (approved):

jagan-jaya (approved):

SanchithHegde (approved):

Gnanasundari24 (approved):

PR #6589, opened 2024-11-17
refactor(payments_v2): use batch encryption for intent create and confirm intent

Narayanbhat166:
## Type of Change

- [x] Refactoring

## Description

This PR adds batch encryption for the payment intent create and payment intent confirm apis. This also refactors the `to_encryptable` operation to fix a small bug which considers the inner type and ignores the annotated types in certain conditions.

This PR also adds a domain type for address which will be used across the application instead of the api models ( which is the current scenerio ). This will be used in the subsequent PRs.

## Motivation and Context

In higher environments, the encryption and decryption happens via the key manager service which is an external api call. If we are encrypting every field of a domain model individually, this would take a lot of network calls and add latency. So we are batch encrypting them.

## How did you test it?

- Check the sanity of payments v2
- Create a payment
- Confirm the payment
- Retrieve the payment

All the above should return 200 status code

## Checklist


dracarys18 (dismissed):

hrithikesh026 (approved):
LGTM

crates/hyperswitch_domain_models/src/payments/payment_attempt.rs
```diff
@@ -1744,13 +1753,39 @@ impl behaviour::Conversion for PaymentAttempt {
...
                 .cloned();
 
+            let decrypted_data = crypto_operation(
+                state,
+                common_utils::type_name!(Self::DstType),
+                CryptoOperation::BatchDecrypt(EncryptedPaymentAttempt::to_encryptable(
+                    EncryptedPaymentAttempt {
+                        payment_method_billing_address: storage_model
+                            .payment_method_billing_address,
+                    },
```
dracarys18, line 1768: Why do we need batch encryption for a single field tho?

dracarys18 (approved):

crates/hyperswitch_domain_models/src/payments/payment_attempt.rs
Narayanbhat166: For now there is only one field, later there might be other fields as well which can be added over here

dracarys18 (approved):

hrithikesh026 (approved):

Issue #5156, opened 2024-06-28
[FEATURE] Add support for SMTP email server


erikvaliyev:
### Feature Description

Add support for sending emails with a custom SMTP server. (Currently only AWS SES is supported)

### Possible Implementation

Could be implemented using `lettre` crate.
Steps to implement:
- Create the email building and sending logic
- Impl the EmailClient trait
- Add a new arm to the match statement in `create_email_client()` function

### Have you spent some time checking if this feature request has been raised before?

- [X] I checked and didn't find a similar issue

### Have you read the Contributing Guidelines?

- [X] I have read the [Contributing Guidelines](https://github.com/juspay/hyperswitch/blob/main/docs/CONTRIBUTING.md)

### Are you willing to submit a PR?

Yes, I am willing to submit a PR!

PR #6617, opened 2024-11-19
feat(email): Add SMTP support to allow mails through self hosted/custom SMTP server

jagan-jaya:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [X] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

Add SMTP support to allow mails through self-hosted/custom SMTP server
#5156 

### Additional Changes

## Motivation and Context

Hyperswitch currently supports only AWS SES to send emails, other open users who want to use some custom SMTP server will have no way to configure it. With this implementation users can change below config to use their own SMTP server.

```
[email]
active_email_client = "SMTP"

[email.smtp]
host = "127.0.0.1"
port = 1025
connection = "plaintext" #currently plaintext and starttls are supported
```

## How did you test it?

Setup mailhog to test emails locally
```docker run -p 1025:1025 -p 8025:8025 mailhog/mailhog```

1. Check if is possible to disable email with the env config
```
[email]
active_email_client = "NO_EMAIL_CLIENT"
```
Start the server with `cargo r` and there should be no error

2. Check the SMTP server without authentication and No TLS with the above configs
Start the server with `cargo r`, and there should be no error
Start the dashboard and change the below config
```
[user]
base_url = "http://localhost:9000"
```

- Check if signup email is sent correctly 
- Check if reset password email is sent correctly: Profile -> Reset Password 
- Check if Forgot password email is sent correctly
- Check if magic link is working
- Check if invite user email is sent correctly
Screenshot 2024-11-20 at 01 24 02
Screenshot 2024-11-20 at 01 24 28

## Checklist

### DIFF
--- a/crates/external_services/src/email/smtp.rs
+++ b/crates/external_services/src/email/smtp.rs
@@ -1,0 +1,181 @@
+use crate::email::{EmailClient, EmailError, EmailResult, EmailSettings, IntermediateString};
+use common_utils::{errors::CustomResult, pii};
+use error_stack::ResultExt;
+use lettre::{
+    address::AddressError,
+    error,
+    message::{header::ContentType, Mailbox},
+    transport::smtp::{self, authentication::Credentials},
+    Message, SmtpTransport, Transport,
+};
+use masking::PeekInterface;
+use std::time::Duration;
+
+/// Client for SMTP server operation
+#[derive(Debug, Clone, Default, serde::Deserialize)]
+pub struct SmtpServer {
+    /// sender email id
+    pub sender: String,
+    /// SMTP server specific configs
+    pub smtp_config: SmtpServerConfig,
+}
+
+impl SmtpServer {
+    /// A helper function to create SMTP server client
+    pub fn create_client(&self) -> Result<SmtpTransport, SmtpError> {
+        let host = self.smtp_config.host.clone();
+        let port = self.smtp_config.port;
+        let timeout = Some(Duration::from_secs(10));
+        let credentials = self
+            .smtp_config
+            .username
+            .clone()
+            .zip(self.smtp_config.password.clone())
+            .map(|(username, password)| Credentials::new(username.to_owned(), password.to_owned()));
+        match &self.smtp_config.connection {
+            SmtpConnection::StartTls => match credentials {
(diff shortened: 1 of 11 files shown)

### CODE REVIEW

File: crates/external_services/src/email/smtp.rs [hunk 1]
```diff
@@ -0,0 +1,187 @@
...
+    /// hostname of the SMTP server eg: smtp.google.com
+    pub host: String,
+    /// portname of the SMTP server eg: 25
+    pub port: u16,
+    /// timeout for the SMTP server connection eg: 10
+    pub timeout: u64,
+    /// Username name of the SMTP server
+    pub username: Option<String>,
+    /// Password of the SMTP server
+    pub password: Option<String>,
```
SanchithHegde, line 95: We can make these two fields as `Secret`, not sure if we should make `host` a `Secret` as well.

crates/external_services/src/email/smtp.rs
```diff
@@ -0,0 +1,187 @@
...
+/// Struct that contains the SMTP server specific configs required
+#[derive(Debug, Clone, Default, serde::Serialize, serde::Deserialize)]
+#[serde(rename_all = "snake_case")]
+pub struct SmtpServerConfig {
+    /// hostname of the SMTP server eg: smtp.google.com
+    pub host: String,
+    /// portname of the SMTP server eg: 25
+    pub port: u16,
+    /// timeout for the SMTP server connection eg: 10
+    pub timeout: u64,
```
SanchithHegde, line 91: Can we mention in the doc comment that this is in seconds?

crates/external_services/src/email/smtp.rs
```diff
@@ -0,0 +1,187 @@
...
+    async fn send_email(
+        &self,
+        recipient: pii::Email,
+        subject: String,
+        body: Self::RichText,
+        _proxy_url: Option<&String>,
+    ) -> EmailResult<()> {
+        // Create a client every time when the email is being sent
+        let email_client =
+            Self::create_client(self).change_context(EmailError::EmailSendingFailure)?;
```
SanchithHegde, line 153: As a future enhancement, we can probably maintain a connection pool instead.

Narayanbhat166 (approved):

PR #6636, opened 2024-11-21
refactor(core): add error handling wrapper to wehbook

Sakilmostak:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [x] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

In webhook, for specific connectors, we want to consume every webhook that arrives even if a error occurs while processing. This is done to keep the webhook channel open with the connector. 
In this pr, a error wrapper is introduced to handle this logic where based on connector, we choose whether to send error or not.

### Additional Changes

## How did you test it?

Tested through Postman:
- Create a MCA (Adyenplatform):
- Setup webhook endpoint in dashboard (Adyenplatform):
- Create a payout:
```
{
    "amount": 100,
    "currency": "EUR",
    "customer_id": "sakilcust",
    "email": "payout_customer@example.com",
    "phone": "999999999",
    "phone_country_code": "+65",
    "description": "Its my first payout request",
    "payout_type": "bank",
    "priority": "instant",
    "payout_method_data": {
        "bank": {
            "iban": "DE89370400440532013000"
        }
    },
    "billing": {
        "address": {
            "line1": "Raadhuisplein",
            "line2": "92",
            "city": "Hoogeveen",
            "state": "FL",
            "zip": "7901 BW",
            "country": "NL",
            "first_name": "John",
            "last_name": "Doe"
        },
        "phone": {
            "number": "0650242319",
            "country_code": "+31"
        }
    },
    "entity_type": "Individual",
    "recurring": true,
    "confirm": true,
    "auto_fulfill": true
}
```
- You should receive webhook for the following payout to the endpoint mentioned during merchant creation

Case 2:
- Setup `ngrok` to intercept webhook
- Create a payout (same body as above)
- delete the `payout_attempt` from the payout_attempt table against the generated `payout_id`
- See the log in ngrok for response sent to the adyenplatform
- The status should be 200 with `x-http-code` present in headers

## Checklist

### DIFF
--- a/crates/api_models/src/connector_enums.rs
+++ b/crates/api_models/src/connector_enums.rs
@@ -282,6 +282,9 @@ impl Connector {
     pub fn is_pre_processing_required_before_authorize(&self) -> bool {
         matches!(self, Self::Airwallex)
     }
+    pub fn should_acknowledge_webhook_for_resource_not_found_errors(&self) -> bool {
+        matches!(self, Self::Adyenplatform)
+    }
     #[cfg(feature = "dummy_connector")]
     pub fn validate_dummy_connector_enabled(
         &self,
--- a/crates/router/src/connector/adyenplatform.rs
+++ b/crates/router/src/connector/adyenplatform.rs
@@ -13,6 +13,8 @@ use error_stack::report;
 use error_stack::ResultExt;
 #[cfg(feature = "payouts")]
 use http::HeaderName;
+use hyperswitch_interfaces::webhooks::IncomingWebhookFlowError;
+use masking::Maskable;
 #[cfg(feature = "payouts")]
 use masking::Secret;
 #[cfg(feature = "payouts")]
@@ -27,11 +29,7 @@ use crate::{
     configs::settings,
     core::errors::{self, CustomResult},
     headers,
-    services::{
-        self,
-        request::{self, Mask},
-        ConnectorValidation,
-    },
+    services::{self, request::Mask, ConnectorValidation},
     types::{
         self,
         api::{self, ConnectorCommon},
@@ -67,7 +65,7 @@ impl ConnectorCommon for Adyenplatform {
     fn get_auth_header(
         &self,
         auth_type: &types::ConnectorAuthType,
-    ) -> CustomResult<Vec<(String, request::Maskable<String>)>, errors::ConnectorError> {
+    ) -> CustomResult<Vec<(String, Maskable<String>)>, errors::ConnectorError> {
         let auth = adyenplatform::AdyenplatformAuthType::try_from(auth_type)
             .change_context(errors::ConnectorError::FailedToObtainAuthType)?;
         Ok(vec![(
@@ -209,7 +207,7 @@ impl services::ConnectorIntegration<api::PoFulfill, types::PayoutsData, types::P
         &self,
         req: &types::PayoutsRouterData<api::PoFulfill>,
         _connectors: &settings::Connectors,
-    ) -> CustomResult<Vec<(String, request::Maskable<String>)>, errors::ConnectorError> {
+    ) -> CustomResult<Vec<(String, Maskable<String>)>, errors::ConnectorError> {
         let mut header = vec![(
             headers::CONTENT_TYPE.to_string(),
             types::PayoutFulfillType::get_content_type(self)
@@ -398,6 +396,25 @@ impl api::IncomingWebhook for Adyenplatform {
         #[cfg(not(feature = "payouts"))]
         {
             Err(report!(errors::ConnectorError::WebhooksNotImplemented))
+        }
+    }
+
+    fn get_webhook_api_response(
+        &self,
+        _request: &api::IncomingWebhookRequestDetails<'_>,
+        error_kind: Option<IncomingWebhookFlowError>,
+    ) -> CustomResult<services::api::ApplicationResponse<serde_json::Value>, errors::ConnectorError>
+    {
+        if error_kind.is_some() {
+            Ok(services::api::ApplicationResponse::JsonWithHeaders((
+                serde_json::Value::Null,
+                vec![(
+                    "x-http-code".to_string(),
+                    Maskable::Masked(Secret::new("404".to_string())),
+                )],
+            )))
+        } else {
+            Ok(services::api::ApplicationResponse::StatusOk)
         }
     }
 
--- a/crates/router/src/connector/adyenplatform.rs
+++ b/crates/router/src/connector/adyenplatform.rs
@@ -13,6 +13,7 @@ use error_stack::report;
 use error_stack::ResultExt;
 #[cfg(feature = "payouts")]
 use http::HeaderName;
+use hyperswitch_interfaces::webhooks::IncomingWebhookFlowError;
 use masking::Maskable;
 #[cfg(feature = "payouts")]
 use masking::Secret;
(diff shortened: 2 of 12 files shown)


crates/router/src/utils.rs
```diff
@@ -1344,3 +1344,7 @@ pub async fn trigger_refund_outgoing_webhook(
 ) -> RouterResult<()> {
     todo!()
 }
+
+pub fn is_webhook_not_found_error(err: &errors::ApiErrorResponse) -> bool {
+    matches!(err, errors::ApiErrorResponse::WebhookResourceNotFound)
```
sai-harsha-vardhan, line 1349: Please add all the other NotFound errors which can occur like PaymentNotFound, RefundNotFound etc.

crates/router/src/core/webhooks/incoming.rs
```diff
@@ -117,14 +117,104 @@ pub async fn incoming_webhooks_wrapper<W: types::OutgoingWebhookType>(
     Ok(application_response)
 }
 
+async fn incoming_webhooks_error_wrapper<W: types::OutgoingWebhookType>(
```
sai-harsha-vardhan, line 120: can we try to eliminate this wrapper?

crates/api_models/src/connector_enums.rs
```diff
@@ -282,6 +282,9 @@ impl Connector {
     pub fn is_pre_processing_required_before_authorize(&self) -> bool {
         matches!(self, Self::Airwallex)
     }
+    pub fn should_connector_bypass_error_if_webhook_not_found(&self) -> bool {
```
sai-harsha-vardhan, line 285:
```suggestion
    pub fn should_acknowledge_webhook_for_resource_not_found_errors(&self) -> bool {
```

crates/router/src/connector/adyenplatform.rs
```diff
@@ -401,6 +398,29 @@ impl api::IncomingWebhook for Adyenplatform {
...
+            Ok(services::api::ApplicationResponse::JsonWithHeaders((
+                serde_json::json!({ "message": "accepted" }),
+                vec![(
+                    "x-http-code".to_string(),
+                    Maskable::Masked(Secret::new("404".to_string())),
+                )],
+            )))
+        } else {
+            Ok(services::api::ApplicationResponse::TextPlain(
+                "[accepted]".to_string(),
```
sai-harsha-vardhan, line 419: There's no need to send `"[accepted]"` back in the response / any content body right?

crates/router/src/core/webhooks/incoming.rs
```diff
@@ -64,17 +116,36 @@ pub async fn incoming_webhooks_wrapper<W: types::OutgoingWebhookType>(
...
+    let (application_response, webhooks_response_tracker, serialized_req) = match Box::pin(
+        incoming_webhooks_core::<W>(
             state.clone(),
             req_state,
             req,
             merchant_account.clone(),
             key_store,
-            connector_name_or_mca_id,
+            merchant_connector_account,
+            connector.clone(),
```
sai-harsha-vardhan, line 139: you can try passing references here

crates/router/src/core/webhooks/incoming.rs
```diff
@@ -64,17 +116,36 @@ pub async fn incoming_webhooks_wrapper<W: types::OutgoingWebhookType>(
     body: actix_web::web::Bytes,
 ) -> RouterResponse<serde_json::Value> {
     let start_instant = Instant::now();
-    let (application_response, webhooks_response_tracker, serialized_req) =
-        Box::pin(incoming_webhooks_core::<W>(
+
+    // Fetch the merchant connector account to get the webhooks source secret
```
sai-harsha-vardhan, line 120: Can you also please add the comment why the mca and connector are being fetched in wrapper instead of core?

crates/router/src/utils.rs
```diff
@@ -1344,3 +1344,15 @@ pub async fn trigger_refund_outgoing_webhook(
...
+
+pub fn is_webhook_not_found_error(err: &errors::ApiErrorResponse) -> bool {
+    matches!(
+        err,
+        errors::ApiErrorResponse::WebhookResourceNotFound
+            | errors::ApiErrorResponse::DisputeNotFound { .. }
+            | errors::ApiErrorResponse::PayoutNotFound
+            | errors::ApiErrorResponse::MandateNotFound
+            | errors::ApiErrorResponse::PaymentNotFound
+            | errors::ApiErrorResponse::RefundNotFound
```
sai-harsha-vardhan, line 1356: Should add `errors::ApiErrorResponse::AuthenticationNotFound` error too

crates/router/src/utils.rs
```diff
@@ -1344,3 +1344,15 @@ pub async fn trigger_refund_outgoing_webhook(
 ) -> RouterResult<()> {
     todo!()
 }
+
+pub fn is_webhook_not_found_error(err: &errors::ApiErrorResponse) -> bool {
```
SanchithHegde, line 1348: Can we have this as a method on `ApiErrorResponse`?

crates/router/src/core/webhooks/incoming.rs
```diff
@@ -52,6 +52,58 @@ use crate::{
...
+            .get_webhook_api_response(&request_details)
+            .switch()
+            .attach_printable("Could not get incoming webhook api response from connector")?;
+
+        Ok((
+            response,
+            WebhookResponseTracker::NoEffect,
+            serde_json::Value::Null,
+        ))
+    } else {
```
SanchithHegde, line 102: This seems unnecessarily complicated in my opinion, took me some time to wrap my head around it:

- At this point, we already know that there's an error, we know the type of error, and that we should be ignoring object not found errors for this specific connector.
- Yet, we modify the request headers, construct an incoming webhook request (again) and then call `get_webhook_api_response()`.
- In this specific case, we are modifying request headers to indicate that it was a object not found error.

I believe we also have a fair idea about where in the incoming webhooks core code we'd get an object not found error. Given that we know where we'd get an object not found error, can't we handle it suitably in the incoming webhooks core itself and return a connector-specific response?

For generating a connector-specific response, we can do one of two things:
- Add a new trait method along the lines of `get_webhook_object_not_found_api_response()`.
- Add a new parameter to the `get_webhook_api_response()` that either indicates the type of error (if applicable), or specifies whether it is an object not found error, and the connector code can choose to ignore the error or not (and generate a suitable response) based on this newly introduced parameter.

Do let me know if I'm missing something here.

crates/router/src/connector/adyenplatform.rs
```diff
@@ -401,6 +399,25 @@ impl api::IncomingWebhook for Adyenplatform {
...
+        error_kind: Option<IncomingWebhookFlowError>,
+    ) -> CustomResult<services::api::ApplicationResponse<serde_json::Value>, errors::ConnectorError>
+    {
+        if error_kind.is_some() {
+            Ok(services::api::ApplicationResponse::JsonWithHeaders((
+                serde_json::Value::Null,
+                vec![(
+                    "x-http-code".to_string(),
+                    Maskable::Masked(Secret::new("404".to_string())),
+                )],
```
SanchithHegde, line 414: Do we still need this custom header? If yes, then where would be monitoring this (Grafana dashboard, etc.)? Or can we have metrics instead?

SanchithHegde (approved):

sai-harsha-vardhan (approved):

srujanchikke (approved):

Issue #6642, opened 2024-11-22
refactor(tenant): use tenant id type


apoorvdixit88:
Refactor: Use tenant_id type instead of string for better type safety

PR #6643, opened 2024-11-22
refactor(tenant): use tenant id type

apoorvdixit88:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [x] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description
Use tenant_id type for better type safety

### Additional Changes

## Motivation and Context
Closes [#6642](https://github.com/juspay/hyperswitch/issues/6642)

## How did you test it?
It refactoring PR, its compiling and checks are passing
Tested flows like signup/signin with tenancy featue flag enabled and disabled in local

## Checklist


ThisIsMani (dismissed):

ThisIsMani (approved):

jagan-jaya (dismissed):

racnan (dismissed):

crates/common_utils/src/id_type/tenant.rs
```diff
@@ -0,0 +1,30 @@
...
+crate::impl_debug_id_type!(TenantId);
+crate::impl_try_from_cow_str_id_type!(TenantId, "tenant_id");
+
+crate::impl_serializable_secret_id_type!(TenantId);
+crate::impl_queryable_id_type!(TenantId);
+crate::impl_to_sql_from_sql_id_type!(TenantId);
+
+impl TenantId {
+    /// Get tenant id from String
+    pub fn wrap(tenant_id: String) -> CustomResult<Self, ValidationError> {
```
SanchithHegde, line 20: Nit: Can we use a better name which indicates what it's being constructed from, say `try_from_string()` maybe?

crates/common_utils/src/id_type/tenant.rs
```diff
@@ -0,0 +1,30 @@
...
+        Self::try_from(std::borrow::Cow::from(tenant_id))
+    }
+
+    /// Create a tenant ID without check
+    pub fn new_unchecked(tenant_id: &str) -> Self {
```
SanchithHegde, line 25: Do we necessarily need the unchecked, or can we avoid it?

My concern is that keeping a public unchecked constructor allows the possibility of invalid data entering our system.

crates/router/src/db/events.rs
```diff
@@ -732,7 +732,10 @@ mod tests {
         ))
         .await;
         let state = &Arc::new(app_state)
-            .get_session_state("public", || {})
+            .get_session_state(
+                &common_utils::id_type::TenantId::wrap("public".to_string()).unwrap(),
```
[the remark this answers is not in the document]
apoorvdixit88: Its test, i guess we can hard code the value, as it was before.

crates/router/src/core/user.rs
```diff
@@ -1105,7 +1105,8 @@ pub async fn create_internal_user(
             }
         })?;
 
-    let default_tenant_id = common_utils::consts::DEFAULT_TENANT.to_string();
+    let default_tenant_id =
+        common_utils::id_type::TenantId::new_unchecked(common_utils::consts::DEFAULT_TENANT);
```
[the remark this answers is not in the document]
apoorvdixit88: Unchecked we can use for internal purpose like for the const we will be declaring. With this we don't need to handle the errors. But its not required we can change.

crates/common_utils/src/id_type/tenant.rs
apoorvdixit88: Yes we can, for org id we follow the `wrap()`, i just thought to keep things consistent. Shall i change it there as well?

crates/common_utils/src/id_type/tenant.rs
apoorvdixit88: We don't necessary need unchecked, its just that it can be used for internal use cases. We don't need to handle the errors for the constants we will be defining. 
But yes it can add up to possibility of invalid data, i will try to remove this.

crates/common_utils/src/id_type/tenant.rs
SanchithHegde: Let it remain for now, don't take that change in this PR.

Also looping in @sai-harsha-vardhan and @Narayanbhat166 about this.

SanchithHegde (approved):

dracarys18 (approved):

srujanchikke (approved):
Reviewed connector changes

crates/router/src/services/api.rs
```diff
@@ -722,21 +722,24 @@ where
...
-        .change_context(
-            errors::ApiErrorResponse::InvalidRequestData {
-                message: format!("`{}` header is invalid", headers::X_TENANT_ID),
-            }
-            .switch(),
-        )?;
+        let request_tenant_id = incoming_request_header
+            .get(TENANT_HEADER)
+            .and_then(|value| value.to_str().ok())
+            .ok_or_else(|| errors::ApiErrorResponse::MissingTenantId.switch())
```
racnan, line 732: Header parsing error and header missing error shouldn't be the same.

racnan (approved):

prajjwalkumar17 (approved):

crates/router/src/services/api.rs
apoorvdixit88: Getting parsing here is very unlikely here, we only will be defining the tenant_id in configs. And it will be a valid UTF-8 characterset.

PR #6726, opened 2024-12-03
ci: update cypress creds

### LINKED ISSUE #6645: [FEAT] Multiple credential support for a connector in Cypress

pixincreate:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [x] CI/CD

## Description

Updates creds. closes #6645

### Additional Changes

## Motivation and Context

Me

## How did you test it?

NA

## Checklist


SanchithHegde (dismissed):

likhinbopanna (dismissed):

pixincreate:
once https://github.com/juspay/hyperswitch/pull/6588 gets merged, this pr will get conflicts that needs to be resolved. better to wait until then.

likhinbopanna (approved):

SanchithHegde (approved):

PR #6761, opened 2024-12-05
feat(routing): Contract based routing integration 

Sarthak1799:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

- Built an interface for Contract based routing
- Integrated Contract based routing with hyperswitch
- Built new APIs for contract routing config setup
- Refactored some existing dynamic routing code

### Additional Changes

## How did you test it?

1. Enable Contract routing config
```
curl --location --request POST 'http://localhost:8080/account/sarthak2/business_profile/pro_YozTgS8HebvBlk0UaeWW/dynamic_routing/contracts/toggle?enable=dynamic_connector_selection' \
--header 'api-key: dev_1aD8YuFd6Ovanf3oOtAYobqI6qw4ZBRSfw6BgYbifWHpaopkisBtm8obNXkFbVJn' \
--data-raw '{
    "config": {
        "constants": [0.7,0.35],
        "time_scale": "day"
    },
    "label_info": [{
        "label": "stripe",
        "target_count": 10000,
        "incremental_count": 0,
        "target_time": 1780486655,
        "mca_id": "mca_5Mk0Qcum2tnbmKoyqeuL"
    }]
}'
```
Response - 
```
{
    "id": "routing_FlQ49V71J8ZCXEwPzFy9",
    "profile_id": "pro_YozTgS8HebvBlk0UaeWW",
    "name": "Contract based dynamic routing algorithm",
    "kind": "dynamic",
    "description": "",
    "created_at": 1734009361,
    "modified_at": 1734009361,
    "algorithm_for": "payment"
}
```
2. Set Volume split for dynamic routing 
```
curl --location --request POST 'http://localhost:8080/account/sarthak2/business_profile/pro_YozTgS8HebvBlk0UaeWW/dynamic_routing/set_volume_split?split=80' \
--header 'api-key: dev_1aD8YuFd6Ovanf3oOtAYobqI6qw4ZBRSfw6BgYbifWHpaopkisBtm8obNXkFbVJn'
```
3. Create a Payment
Metrics Population post payment - 
image

## Checklist


crates/router/src/core/routing.rs
```diff
@@ -1420,6 +1427,269 @@ pub async fn success_based_routing_update_configs(
...
+        .clone()
+        .dynamic_routing
+        .contract_based_client
+        .clone()
+        .async_map(|ct_client| async move {
+            ct_client
+                .invalidate_contracts(prefix_of_dynamic_routing_keys)
+                .await
+                .change_context(errors::ApiErrorResponse::GenericNotFoundError {
+                    message: "Failed to invalidate the routing keys".to_string(),
```
Chethan-rao, line 1684:
```suggestion
                    message: "Failed to invalidate the contract based routing keys".to_string(),
```

Chethan-rao (changes requested):

crates/router/src/core/routing/helpers.rs
```diff
@@ -866,6 +954,238 @@ pub async fn push_metrics_with_update_window_for_success_based_routing(
...
+        let (first_contract_based_connector, connector_score, current_payment_cnt) = (first_contract_based_connector.label
+            .split_once(':')
+            .ok_or(errors::ApiErrorResponse::InternalServerError)
+            .attach_printable(format!(
+                "unable to split connector_name and mca_id from the first connector {:?} obtained from dynamic routing service",
+                first_contract_based_connector
+            ))?
+            .0, first_contract_based_connector.score, first_contract_based_connector.current_count );
+
+        core_metrics::DYNAMIC_CONTRACT_BASED_ROUTING.add(
```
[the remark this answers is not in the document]
prajjwalkumar17: I think Prometheus should be suffice, 
we can add something like for this particular `mca_id`, these many transactions went through this particular connector.

Chethan-rao (changes requested):

crates/api_models/src/routing.rs
```diff
@@ -860,6 +954,87 @@ impl CurrentBlockThreshold {
...
+        if let Some(new_label_info) = new.label_info {
+            new_label_info.iter().for_each(|new_label_info| {
+                if let Some(existing_label_infos) = &mut self.label_info {
+                    for existing_label_info in existing_label_infos {
+                        if existing_label_info.mca_id == new_label_info.mca_id {
+                            existing_label_info.update(new_label_info.clone());
+                        }
+                    }
+                } else {
+                    self.label_info = Some(vec![new_label_info.clone()]);
```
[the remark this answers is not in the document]
Sarthak1799: This case would only arise when we don't have any existing entries in the vec

crates/router/src/core/payments/routing.rs
```diff
@@ -1279,34 +1282,83 @@ pub fn make_dsl_input_for_surcharge(
...
+        .change_context(errors::RoutingError::DeserializationError {
+            from: "JSON".to_string(),
+            to: "DynamicRoutingAlgorithmRef".to_string(),
+        })
+        .attach_printable("unable to deserialize DynamicRoutingAlgorithmRef from JSON")?
+        .ok_or(errors::RoutingError::GenericNotFoundError {
+            field: "dynamic_routing_algorithm".to_string(),
+        })?;
+
+    let mut connector_list = routable_connectors.clone();
```
Chethan-rao, line 1310: no need to assign here since you are anyways doing `unwrap_or(routable_connectors);` at the end

crates/router/src/core/payments/routing.rs
```diff
@@ -1279,34 +1282,83 @@ pub fn make_dsl_input_for_surcharge(
...
+                profile,
+                dynamic_routing_config_params_interpolator.clone(),
+                algorithm.clone(),
+            )
+        })
+        .await
+        .transpose()
+        .attach_printable("unable to perform success_based_routing")
+        .ok()
+        .flatten()
```
Chethan-rao, line 1328: why are you ignoring error here?

crates/router/src/core/payments/routing.rs
```diff
@@ -1279,34 +1282,83 @@ pub fn make_dsl_input_for_surcharge(
...
+                profile,
+                dynamic_routing_config_params_interpolator,
+                algorithm.clone(),
+            )
+        })
+        .await
+        .transpose()
+        .attach_printable("unable to perform contract_based_routing")
+        .ok()
+        .flatten()
```
Chethan-rao, line 1347: why are you ignoring error here?

crates/router/src/core/payments/routing.rs
```diff
@@ -1395,3 +1449,97 @@ pub async fn perform_success_based_routing(
...
+        )
+        .await
+        .change_context(errors::RoutingError::ContractBasedRoutingConfigError)
+        .attach_printable("unable to fetch contract based dynamic routing configs")?;
+
+        let contract_based_connectors: CalContractScoreResponse = client
+            .calculate_contract_score(
+                business_profile.get_id().get_string_repr().into(),
+                contract_based_routing_configs,
+                "".to_string(),
```
Chethan-rao, line 1499: will this be empty? makes sense since contract will be associated to id and label and not param but still just making sure

crates/router/src/core/payments/operations/payment_response.rs
```diff
@@ -2010,14 +2010,25 @@ async fn payment_response_update_tracker<F: Clone, T: types::Capturable>(
...
                         &payment_attempt,
                         routable_connectors,
                         &business_profile,
-                        success_based_routing_config_params_interpolator,
+                        dynamic_routing_config_params_interpolator,
                     )
                     .await
-                    .map_err(|e| logger::error!(dynamic_routing_metrics_error=?e))
+                    .map_err(|e| logger::error!(contract_based_routing_metrics_error=?e))
                     .ok();
```
Chethan-rao, line 2032: This will be called when

```
if payment_intent.status.is_in_terminal_state()
            && business_profile.dynamic_routing_algorithm.is_some()
```
`business_profile.dynamic_routing_algorithm.is_some()` is this enough to decide on whether to call sr and cr? should we parse the type outside before calling individual push metrics functions

Chethan-rao (changes requested):

crates/api_models/src/routing.rs
```diff
@@ -563,6 +610,9 @@ impl DynamicRoutingAlgorithmRef {
         if let Some(success_based_algorithm) = new.success_based_algorithm {
             self.success_based_algorithm = Some(success_based_algorithm)
         }
+        if let Some(contract_based_routing) = new.contract_based_routing {
+            self.contract_based_routing = Some(contract_based_routing)
+        }
     }
 
     pub fn update_specific_ref(
```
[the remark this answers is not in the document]
Sarthak1799: The selected code updates the entire algorithm.

crates/router/src/core/payments/routing.rs
Sarthak1799: We're not sending params for now, so this remains empty

crates/api_models/src/routing.rs
```diff
@@ -563,6 +610,9 @@ impl DynamicRoutingAlgorithmRef {
         if let Some(success_based_algorithm) = new.success_based_algorithm {
             self.success_based_algorithm = Some(success_based_algorithm)
         }
+        if let Some(contract_based_routing) = new.contract_based_routing {
+            self.contract_based_routing = Some(contract_based_routing)
+        }
```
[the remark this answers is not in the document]
Sarthak1799: This is updating the algorithm right?

crates/router/src/core/payments/operations/payment_response.rs
```diff
@@ -1974,11 +1974,21 @@ async fn payment_response_update_tracker<F: Clone, T: types::Capturable>(
...
+                    .dynamic_routing_algorithm
+                    .clone()
+                    .map(|val| val.parse_value("DynamicRoutingAlgorithmRef"))
+                    .transpose()
+                    .change_context(errors::ApiErrorResponse::InternalServerError)
+                    .attach_printable("unable to deserialize DynamicRoutingAlgorithmRef from JSON")?
+                    .unwrap_or_default();
+
             let state = state.clone();
             let business_profile = business_profile.clone();
```
Chethan-rao, line 1988: you can only clone the `profile_id` instead of whole profile object since below methods only require id
```suggestion
            let profile_id = business_profile.get_id().clone();
```

crates/router/src/core/payments/routing.rs
```diff
@@ -1281,35 +1284,84 @@ pub fn make_dsl_input_for_surcharge(
...
+                state,
+                routable_connectors.clone(),
+                profile,
+                dynamic_routing_config_params_interpolator.clone(),
+                algorithm.clone(),
+            )
+        })
+        .await
+        .transpose()
+        .inspect_err(|e| logger::error!("Dynamic_routing error {:?}", e))
```
Chethan-rao, line 1326: can we follow the same logger fields everywhere for dynamic routing errors for quick and easy finding. Also since we are anyways dropping the error after this log and not propagating it, we can use map_err instead of inspect_err

```suggestion
        .map_err(|e| logger::error!(dynamic_routing_error=?e))
```

crates/router/src/core/payments/routing.rs
```diff
@@ -1281,35 +1284,84 @@ pub fn make_dsl_input_for_surcharge(
...
+                state,
+                routable_connectors.clone(),
+                profile,
+                dynamic_routing_config_params_interpolator,
+                algorithm.clone(),
+            )
+        })
+        .await
+        .transpose()
+        .inspect_err(|e| logger::error!("Dynamic_routing error {:?}", e))
```
Chethan-rao, line 1346:
```suggestion
        .map_err(|e| logger::error!(dynamic_routing_error=?e))
```

crates/router/src/core/payments/routing.rs
```diff
@@ -1281,35 +1284,84 @@ pub fn make_dsl_input_for_surcharge(
...
+                routable_connectors.clone(),
+                profile,
+                dynamic_routing_config_params_interpolator.clone(),
+                algorithm.clone(),
+            )
+        })
+        .await
+        .transpose()
+        .inspect_err(|e| logger::error!("Dynamic_routing error {:?}", e))
+        .attach_printable("unable to perform success_based_routing")
```
Chethan-rao, line 1327: this `attach_printable` won't have any effect since we are not propagating the error but instead we are calling `.ok()` on Result. So we can remove this

crates/router/src/core/payments/routing.rs
```diff
@@ -1281,35 +1284,84 @@ pub fn make_dsl_input_for_surcharge(
...
+                routable_connectors.clone(),
+                profile,
+                dynamic_routing_config_params_interpolator,
+                algorithm.clone(),
+            )
+        })
+        .await
+        .transpose()
+        .inspect_err(|e| logger::error!("Dynamic_routing error {:?}", e))
+        .attach_printable("unable to perform contract_based_routing")
```
Chethan-rao, line 1347: Same here. remove this

crates/router/src/core/payments/routing.rs
```diff
@@ -1281,35 +1284,84 @@ pub fn make_dsl_input_for_surcharge(
...
+        .attach_printable("unable to perform success_based_routing")
+        .ok()
+        .flatten()
+        .unwrap_or(routable_connectors.clone());
+
+    connector_list = dynamic_routing_algo_ref
+        .contract_based_routing
+        .as_ref()
+        .async_map(|algorithm| {
+            perform_contract_based_routing(
```
Chethan-rao, line 1336: also do we want to go forward with performing contract based routing even if success based routing failed? because I see that you are only logging the error after success based failed and doing unwrap_or to original connector list

crates/router/src/core/payments/routing.rs
```diff
@@ -1281,35 +1284,84 @@ pub fn make_dsl_input_for_surcharge(
...
+    Ok(connector_list)
+}
+
 /// success based dynamic routing
 #[cfg(all(feature = "v1", feature = "dynamic_routing"))]
 #[instrument(skip_all)]
 pub async fn perform_success_based_routing(
     state: &SessionState,
     routable_connectors: Vec<api_routing::RoutableConnectorChoice>,
     business_profile: &domain::Profile,
```
Chethan-rao, line 1361: No need to receive whole profile object because I see that only profile_id is getting used in this method

crates/router/src/core/payments/routing.rs
```diff
@@ -1398,3 +1452,97 @@ pub async fn perform_success_based_routing(
         Ok(routable_connectors)
     }
 }
+
+#[cfg(all(feature = "v1", feature = "dynamic_routing"))]
+pub async fn perform_contract_based_routing(
+    state: &SessionState,
+    routable_connectors: Vec<api_routing::RoutableConnectorChoice>,
+    business_profile: &domain::Profile,
```
Chethan-rao, line 1460: same here, only get profile_id as argument

crates/router/src/core/routing.rs
```diff
@@ -1419,6 +1463,292 @@ pub async fn success_based_routing_update_configs(
...
+                .find_by_merchant_connector_account_merchant_id_merchant_connector_id(
+                    key_manager_state,
+                    merchant_account.get_id(),
+                    &info.mca_id,
+                    &key_store,
+                )
+                .await
+                .change_context(errors::ApiErrorResponse::MerchantConnectorAccountNotFound {
+                    id: info.mca_id.get_string_repr().to_owned(),
+                })?;
```
Chethan-rao, line 1504: can we do tokio join?

crates/router/src/core/routing.rs
```diff
@@ -1419,6 +1463,292 @@ pub async fn success_based_routing_update_configs(
...
+            feature_to_enable,
+            routing_types::DynamicRoutingType::ContractBasedRouting,
+        );
+        algo.disable_algorithm_id(routing_types::DynamicRoutingType::SuccessRateBasedRouting);
+        algo
+    } else {
+        let contract_algo = routing_types::ContractRoutingAlgorithm {
+            algorithm_id_with_timestamp: routing_types::DynamicAlgorithmWithTimestamp {
+                algorithm_id: Some(algorithm_id.clone()),
+                timestamp: common_utils::date_time::now_unix_timestamp(),
```
Chethan-rao, line 1585: call `DynamicAlgorithmWithTimestamp::new`

Chethan-rao (changes requested):

crates/router/src/core/routing.rs
```diff
@@ -1419,6 +1463,292 @@ pub async fn success_based_routing_update_configs(
     Ok(service_api::ApplicationResponse::Json(new_record))
 }
 
+#[cfg(all(feature = "v1", feature = "dynamic_routing"))]
+pub async fn contract_based_dynamic_routing_setup(
```
Chethan-rao, line 1467: why didn't you make use of existing `toggle_specific_dynamic_routing` function to activate

crates/router/src/core/payments/routing.rs
Sarthak1799: Can you give more reasoning into why `map_err` instead?

crates/router/src/core/routing.rs
Sarthak1799: Because CR cannot be set to default, we can use that for routing types that can be always be toggled to some default config.

crates/router/src/core/payments/operations/payment_response.rs
```diff
@@ -1974,11 +1974,21 @@ async fn payment_response_update_tracker<F: Clone, T: types::Capturable>(
...
         {
+            let dynamic_routing_algo_ref: api_models::routing::DynamicRoutingAlgorithmRef =
+                business_profile
+                    .dynamic_routing_algorithm
+                    .clone()
+                    .map(|val| val.parse_value("DynamicRoutingAlgorithmRef"))
+                    .transpose()
+                    .change_context(errors::ApiErrorResponse::InternalServerError)
+                    .attach_printable("unable to deserialize DynamicRoutingAlgorithmRef from JSON")?
+                    .unwrap_or_default();
```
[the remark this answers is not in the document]
Sarthak1799: let expressions in the same condition are unstable. We'd have to have two nested loops then. 
I'll try to find a way better way of doing this

crates/router/src/core/payments/routing.rs
Sarthak1799: Success routing can only fail here if,
It's set up but call fails, in this case contract won't be activated so the call won't happen.
It's not set up, in which case call won't happen and contract call would take place next.

crates/router/src/core/routing/helpers.rs
```diff
@@ -573,78 +579,149 @@ pub fn get_default_config_key(
...
-/// Checked fetch of success based routing configs
+/// Cfetch dynamic routing configs
 #[cfg(all(feature = "v1", feature = "dynamic_routing"))]
 #[instrument(skip_all)]
-pub async fn fetch_success_based_routing_configs(
+pub async fn fetch_dynamic_routing_configs<T>(
     state: &SessionState,
-    business_profile: &domain::Profile,
+    profile_id: &id_type::ProfileId,
     success_based_routing_id: id_type::RoutingId,
```
Chethan-rao, line 678: since this is generic function now rename this parameter to `routing_id`

crates/router/src/core/routing/helpers.rs
```diff
@@ -869,6 +938,217 @@ pub async fn push_metrics_with_update_window_for_success_based_routing(
...
+                final_label_info.label.clone(),
+                final_label_info.mca_id.get_string_repr()
+            ),
+            target_count: final_label_info.target_count,
+            target_time: final_label_info.target_time,
+            mca_id: final_label_info.mca_id.to_owned(),
+        };
+
+        let payment_status_attribute =
+            get_desired_payment_status_for_success_routing_metrics(payment_attempt.status);
```
Chethan-rao, line 1023: name this method generic as this is being called in both sr and cr

crates/router/src/core/routing/helpers.rs
```diff
@@ -869,6 +938,217 @@ pub async fn push_metrics_with_update_window_for_success_based_routing(
...
+        let (first_contract_based_connector, connector_score, current_payment_cnt) = (first_contract_based_connector.label
+            .split_once(':')
+            .ok_or(errors::ApiErrorResponse::InternalServerError)
+            .attach_printable(format!(
+                "unable to split connector_name and mca_id from the first connector {:?} obtained from dynamic routing service",
+                first_contract_based_connector
+            ))?
+            .0, first_contract_based_connector.score, first_contract_based_connector.current_count );
+
+        let outcome = get_success_based_metrics_outcome_for_payment(
```
Chethan-rao, line 1072: same here, name it generic

crates/router/src/core/routing/helpers.rs
```diff
@@ -869,6 +938,217 @@ pub async fn push_metrics_with_update_window_for_success_based_routing(
...
+            payment_status_attribute,
+            payment_connector.to_string(),
+            first_contract_based_connector.to_string(),
+        );
+
+        core_metrics::DYNAMIC_CONTRACT_BASED_ROUTING.add(
```
Chethan-rao, line 1078: Should this metric be pushed only if `payment_status_attribute` is `charged`?

crates/router/src/core/routing/helpers.rs
```diff
@@ -1077,15 +1390,18 @@ pub async fn enable_dynamic_routing_algorithm(
...
+            dynamic_routing
+                .disable_algorithm_id(routing_types::DynamicRoutingType::ContractBasedRouting);
+
             enable_specific_routing_algorithm(
                 state,
                 key_store,
                 business_profile,
                 feature_to_enable,
-                dynamic_routing_algo_ref,
+                dynamic_routing.clone(), // check this in review
```
Chethan-rao, line 1404: comment?

crates/router/src/core/routing/helpers.rs
```diff
@@ -1231,6 +1552,13 @@ pub async fn default_specific_dynamic_routing_setup(
                 algorithm_for: common_enums::TransactionType::Payment,
             }
         }
+        // Should we provide a default for this?
```
Chethan-rao, line 1555: comment?

Chethan-rao (changes requested):

crates/router/src/core/routing/helpers.rs
Sarthak1799: Won't go ahead with default

crates/router/src/core/routing/helpers.rs
Sarthak1799: Was to check the updation logic above, removing this.

Chethan-rao (approved):

tsdk02 (dismissed):

crates/router/src/core/payments/routing.rs
```diff
@@ -1281,41 +1284,92 @@ pub fn make_dsl_input_for_surcharge(
...
-            .change_context(errors::RoutingError::DeserializationError {
-                from: "JSON".to_string(),
-                to: "DynamicRoutingAlgorithmRef".to_string(),
-            })
-            .attach_printable("unable to deserialize DynamicRoutingAlgorithmRef from JSON")?
-            .unwrap_or_default();
+    logger::debug!(
+        "performing dynamic_routing for profile {}",
+        profile.get_id().get_string_repr()
+    );
```
prajjwalkumar17, line 1297: Can we possible log this once the `dynamic_routing_algorithm` is found in `profile` and is de-serialized, So that we will be certain that dynamic_routing will be performed for sure. Right now if the algo is not found, the action won't happen itself.
Basically after line 1310.

crates/router/src/core/payments/routing.rs
```diff
@@ -1398,3 +1452,95 @@ pub async fn perform_success_based_routing(
         Ok(routable_connectors)
     }
 }
+
+#[cfg(all(feature = "v1", feature = "dynamic_routing"))]
+pub async fn perform_contract_based_routing(
+    state: &SessionState,
+    routable_connectors: Vec<api_routing::RoutableConnectorChoice>,
+    profile_id: &common_utils::id_type::ProfileId,
+    _dynamic_routing_config_params_interpolator: routing::helpers::DynamicRoutingConfigParamsInterpolator,
```
prajjwalkumar17, line 1461: Can we remove this, if redundant ?

prajjwalkumar17 (dismissed):

crates/router/src/core/payments/routing.rs
Sarthak1799: Had kept this as we had decided to use the params in the upcoming PRs,

Chethan-rao (dismissed):

NishantJoshi00 (commented):
Looks good, some questions:
1. Is this enabled by default?
2. Have you tested if this works for a contract?
3. Where are the state for this being stored?
4. Where is the flow this is being called?
5. Are any interactions with dynamic routing service in the critical path (if so, how do we handle network failures)
6. I was able to see some caching changes, will this affect the state management for managing contract.


I think its imperative to follow micro PR model to avoid such confusions; 
1. it becomes difficult to follow the code changes.
2. Such massive changes can hide bugs which are difficult to find due to its sheer size.

Sarthak1799:
- We have to enable contract based routing using setup APIs introduced in this PR
- have tested the flow locally
- The actual config is stored in `routing_algorithm` with it's ref stored in profile, like all other DR algorithms
- Similar to other routing types, this will be called in the payment flow, if the contract based algo is enabled.
- In case of Dynamic routing failure, the list of eligible of connectors are defaulted to the list from static routing
- The caching changes here are to make the caching code generic to all categories of Dynamic routing (SR, ER, Contracts) and making use of specific functions from redis interface (like `get_or_populate_in_memory`)
cc: @Chethan-rao 


Agreed on the micro PR model @NishantJoshi00, had thought about breaking up in several PRs but much of the reviews were already completed at the time.

NishantJoshi00 (dismissed):

tsdk02 (dismissed):

NishantJoshi00 (approved):

Chethan-rao (approved):

tsdk02 (approved):

PR #6772, opened 2024-12-06
feat(users): Incorporate themes in user APIs

ThisIsMani:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

todo!()

### Additional Changes

todo!()

## Motivation and Context

todo!()

## How did you test it?

todo!()

## Checklist


crates/router/src/core/user.rs
```diff
@@ -1056,6 +1118,15 @@ pub async fn resend_invite(
         .get_entity_id_and_type()
         .ok_or(UserErrors::InternalServerError)?;
 
+    let role_info = user_from_token.get_role_info_from_db(&state).await?;
+
+    let theme = theme_utils::get_most_specific_theme_using_token_and_min_entity(
+        &state,
+        &user_from_token,
+        role_info.get_entity_type(),
```
apoorvdixit88, line 1126: Shouldn't it be receiver's role info?

crates/router/src/configs/secrets_transformers.rs
```diff
@@ -545,6 +545,6 @@ pub(crate) async fn fetch_raw_secrets(
             .network_tokenization_supported_card_networks,
         network_tokenization_service,
         network_tokenization_supported_connectors: conf.network_tokenization_supported_connectors,
-        theme_storage: conf.theme_storage,
+        theme: conf.theme,
```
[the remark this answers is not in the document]
ThisIsMani: Yeah, I kept the old configs as well.

crates/router/src/core/user.rs
ThisIsMani: Yeah, changed this.

apoorvdixit88 (dismissed):

migrations/2024-12-05-131123_add-email-theme-data-in-themes/up.sql
SanchithHegde, line 1: Nit: Would prefer to avoid including such defaults on the database side. 😬
Do we have any other option?

ThisIsMani: Those defaults are present only to fill the old data if anything is present. 

Application will never insert null in the DB. This is the `NewTheme` struct in the last commit.
https://github.com/juspay/hyperswitch/blob/f61ae14a9f1a7500b4bb9313d2b2fb3b3e0404a9/crates/diesel_models/src/user/theme.rs#L42

If we have to remove the default, then I would have to make that optional. But that will cause more problems later...

SanchithHegde (approved):

apoorvdixit88 (approved):

jarnura (dismissed):

jarnura (approved):

SanchithHegde (approved):

PR #6827, opened 2024-12-12
feat(redis-interface): add redis interface command to set multiple the keys in redis and increment if the key already exists

Aprabhat19:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description
add redis interface command to set multiple the keys in redis and increment if the key already exists

### Additional Changes

## How did you test it?
 
 - Wrote a unit test to insert the keys
Screenshot 2024-12-13 at 1 05 33 PM

## Checklist


crates/redis_interface/src/commands.rs
```diff
@@ -852,12 +852,28 @@ impl super::RedisConnectionPool {
             .await
             .change_context(errors::RedisError::ConsumerGroupClaimFailed)
     }
+
+    #[instrument(level = "DEBUG", skip(self))]
+    pub async fn set_keys_using_script(
```
NishantJoshi00, line 857: this is not a set interface, rather an `incr` interface.

crates/redis_interface/src/commands.rs
```diff
@@ -852,12 +852,28 @@ impl super::RedisConnectionPool {
...
+        &self,
+        keys: Vec<String>,
+        values: Vec<String>,
```
NishantJoshi00, line 860: Have we evaluated the case if `keys` and `values` have different length?
in that case, better write this as `key_values: Vec<(String, usize)>`

NishantJoshi00 (changes requested):

crates/redis_interface/src/commands.rs
Aprabhat19: Sure will change the name

NishantJoshi00 (approved):

crates/redis_interface/src/commands.rs
```diff
@@ -852,12 +852,32 @@ impl super::RedisConnectionPool {
             .await
             .change_context(errors::RedisError::ConsumerGroupClaimFailed)
     }
+
+    #[instrument(level = "DEBUG", skip(self))]
+    pub async fn incr_keys_using_script(
+        &self,
+        lua_script: &str,
+        keys_and_values: HashMap<String, usize>,
```
NishantJoshi00, line 860: Separate this if the script is generic

NishantJoshi00 (changes requested):

crates/redis_interface/src/commands.rs
```diff
@@ -852,12 +852,31 @@ impl super::RedisConnectionPool {
             .await
             .change_context(errors::RedisError::ConsumerGroupClaimFailed)
     }
+
+    #[instrument(level = "DEBUG", skip(self))]
+    pub async fn incr_keys_using_script<V>(
+        &self,
+        lua_script: &str,
+        key: Vec<String>,
```
NishantJoshi00, line 860: Can't this be multiple keys?

NishantJoshi00 (dismissed):

crates/redis_interface/src/commands.rs
Aprabhat19: Can be but do we also want to make the keys generic,I mean should we allow that? As today in our redis-interface module there's no other key than a string

Sarthak1799 (dismissed):

NishantJoshi00 (approved):

Sarthak1799 (approved):

PR #6855, opened 2024-12-16
fix(core): Populate off_session based on payments request

deepanshu-iiitu:
## Type of Change

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

### Additional Changes

## Checklist


Aprabhat19 (commented):
Can we add test cases?

crates/router/src/core/payments/transformers.rs
```diff
@@ -2619,7 +2619,10 @@ impl<F: Clone> TryFrom<PaymentAdditionalData<'_, F>> for types::PaymentsAuthoriz
             payment_method_data: (payment_method_data.get_required_value("payment_method_data")?),
             setup_future_usage: payment_data.payment_intent.setup_future_usage,
             mandate_id: payment_data.mandate_id.clone(),
-            off_session: payment_data.mandate_id.as_ref().map(|_| true),
+            off_session: payment_data
```
sai-harsha-vardhan, line 2622: What would happen in a case where `setup_future_usage` is set as `off_session` by merchant in payments create call and `off_sessions` is `true` in confirm call, I see that currently we don't have any validations in stripe and we send the data as it is. What's the expected behaviour here? @deepanshu-iiitu

deepanshu-iiitu: Stripe does not want setup_future_usage and off_session to come in the same payment request.
This validation has to be handled by the merchant only. 
We will throw the error message from stripe if merchant sends both these values

Aprabhat19: @sai-harsha-vardhan  the case that you have mentioned might not happen because `off_session:true` is just supported in case of S2S calls, where there won't be an intent, it would be one single confirm

Narayanbhat166 (approved):

Aprabhat19 (approved):

sai-harsha-vardhan (approved):

PR #6869, opened 2024-12-18
feat(email): Add mailhog by default in docker-compose for local smtp server

jagan-jaya:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [X] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

Add support to run local email server in docker-compose. User can see the email we send even in local setup

### Additional Changes

## Checklist


docker-compose.yml
```diff
@@ -51,10 +51,18 @@ services:
...
+    image: mailhog/mailhog
+    networks:
+      - router_net
+    ports:
+      - "1025:1025"
+      - "8025:8025"
 
   ### Application services
   hyperswitch-server:
-    image: juspaydotin/hyperswitch-router:standalone
```
SanchithHegde, line 57: Could you revert this change?

JeevaRamu0104 (approved):

ThisIsMani (dismissed):

docker-compose.yml
SanchithHegde: No, we consider nightly images to be unstable, we can't use them in Docker Compose setup.

jagan-jaya: can we use hyperswitch-router:latest? this should be built based on last stable release right?

SanchithHegde: Yes, it is based on latest stable release. 

But does the `latest` image tag work with the Docker Compose setup, considering that it would also have the encryption service related features enabled? Have we tried using it in our Docker Compose setup?

jagan-jaya: Yes, working as expected in local

SanchithHegde (approved):

JeevaRamu0104 (approved):

docker-compose.yml
JeevaRamu0104: Resolving this conversation

PR #6957, opened 2024-12-29
refactor(proxy): specify hosts for proxy exclusion instead of complete URLs

### LINKED ISSUE #1039: [FEATURE] Use proxy exclusion instead of a separate proxied client
### Feature Description

Currently we provide the option for proxying all external requests made to Connectors. We also use reqwest clients to make API calls to basilisk/locker or other utilities.

Recently we started reusing the `reqwest::Client` (via once_cell) for performance reasons
Our current approach involves maintaining 2 clients:
- Proxied Client for any requests going outside our system
- Non-Proxied client for internal services

Currently we need to maintain a list of the services that should bypass proxy & choose a client accordingly. This adds a bit of a boilerplate whenever we try to make requests.

### Possible Implementation

Expected behavior:

Instead we want to push this behavior down to the reqwest library…
we believe the [no_proxy](https://docs.rs/reqwest/latest/reqwest/struct.Proxy.html#method.no_proxy) function should help us avoid this maintenance.

We would ideally accept a list of `proxy_bypass_urls` as part of our config which would then be used to create a single `reqwest::Client`  which can be used for internal + external services

No, I don't have time to work on this right now

SanchithHegde:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [x] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

This PR refactors the logic for excluding specific HTTP / HTTPS traffic from being proxied via the outgoing proxy. Specifically, the changes included are:

1. We no longer need two clients (a proxied client and a non-proxied client), we can have the same client handle traffic as required.
2. We need not specify complete URLs for proxy exclusion, we can just specify the domain names or IP addresses of the hosts to be excluded. This also handles excluding traffic from a subdomain if a parent domain is excluded.

This is achieved with [`reqwest::Proxy::no_proxy()`](https://docs.rs/reqwest/0.12.11/reqwest/struct.Proxy.html#method.no_proxy) and [`reqwest::NoProxy::from_string()`](https://docs.rs/reqwest/0.12.11/reqwest/struct.NoProxy.html#method.from_string).

### Additional Changes

This PR introduces a config option `bypass_proxy_hosts` instead of `bypass_proxy_urls` under the `[proxy]` config section.

## Motivation and Context

This would help significantly reduce maintenance efforts on our side with respect to which URLs we wanted to be excluded from being proxied, since we previously had some of these URLs being specified in code. The PR helps ensure that all the domains to be excluded would be specified via configuration alone, while also allowing us to easily exclude subdomains (for example, excluding `cluster.local` would exclude all traffic to services within a Kubernetes cluster from being proxied).

Closes #1039.

## How did you test it?

Locally, by setting up [`mitmproxy`](https://mitmproxy.org) as a proxy for outgoing HTTP and HTTPS traffic. I also set up our card vault locally to simulate an internal service running close to the application.

I ran `mitmproxy` using the command:

```shell
mitmweb --listen-port 8090 --web-port 8091
```

Application was configured in the `development.toml` file to use the proxy:

```toml
[proxy]
http_url = "http://localhost:8090"
https_url = "http://localhost:8090"
bypass_proxy_hosts = "localhost, cluster.local"
```

1. Creating a payment via Postman:

   Screenshot of 2 requests being proxied

   Two requests are listed on the `mitmproxy` web interface, one to `api.stripe.com` and one to `webhook.site`, indicating that they were proxied. In this case, all HTTPS traffic is being proxied.

2. Saving a payment method in the locker:

   The locker host is initially configured as:

   ```toml
   [locker]
   host = "http://127.0.0.1:3001"
   mock_locker = false
   ```

   Since the locker host was `127.0.0.1` and only `localhost` was excluded from being proxied, the locker request is also being proxied:

   Screenshot of card vault request being proxied

   On updating the locker host to `http://localhost:3001` in the `development.toml` file, the locker request is no longer proxied, the `mitmproxy` web interface is blank:

   Screenshot of a blank mitmproxy web interface

   As another confirmation that the locker call actually happened, I see logs emitted by the locker application with the same request ID as that returned by the `router` in its response headers.

   One more thing to note here is that we did not have to include the port number (`localhost:3001`) in the `bypass_proxy_hosts` list, only `localhost` was sufficient.

3. Excluding `stripe.com` from being proxied and creating a payment:

   I updated the `development.toml` file to include `stripe.com` in `bypass_proxy_hosts`:

   ```toml
   bypass_proxy_hosts = "localhost, cluster.local, stripe.com"
   ```

   Even though we only excluded `stripe.com` and did not explicitly exclude `api.stripe.com`, the request to `api.stripe.com` was excluded as well. This can be compared with the screenshot in (1), the request to `api.stripe.com` is not listed:

   Screenshot of a request being proxied

(2) and (3) indicate that proxy exclusion is happening as expected.

### Testing on our hosted environments

In case of our hosted environments, we would have to ensure that nothing with respect to outgoing traffic from the `router` application remains affected: if an outgoing proxy has been configured (correctly), all of the external HTTP(S) traffic (to connectors, etc.) must be proxied, while traffic to our internal services such as our locker or encryption service must be excluded from being proxied.

## Checklist


crates/router/src/configs/settings.rs
```diff
@@ -623,7 +623,7 @@ pub struct Proxy {
     pub http_url: Option<String>,
     pub https_url: Option<String>,
     pub idle_pool_connection_timeout: Option<u64>,
-    pub bypass_proxy_urls: Vec<String>,
+    pub bypass_proxy_hosts: Option<String>,
```
SanchithHegde, line 626: Not using a `Vec` or `HashSet` here (and splitting the string on commas), we pass the string directly to [`NoProxy::from_string()`](https://docs.rs/reqwest/0.12.11/reqwest/struct.NoProxy.html#method.from_string), which handles the splitting by comma and trimming whitespaces.

crates/router/src/services/api/client.rs
```diff
@@ -31,16 +29,16 @@ fn get_client_builder(
                 .unwrap_or_default(),
         ));
 
-    if should_bypass_proxy {
-        return Ok(client_builder);
-    }
+    let proxy_exclusion_config =
+        reqwest::NoProxy::from_string(&proxy_config.bypass_proxy_hosts.clone().unwrap_or_default());
```
SanchithHegde, line 33: Using `unwrap_or_default()` here, since `NoProxy::from_string()` returns `None` if the string is empty, and since `Proxy::no_proxy()` accepts `Option<NoProxy>`.

crates/router/src/services/api/client.rs
```diff
@@ -218,68 +179,41 @@ where
...
 }
 
 dyn_clone::clone_trait_object!(ApiClient);
 
 #[derive(Clone)]
 pub struct ProxyClient {
-    proxy_client: reqwest::Client,
-    non_proxy_client: reqwest::Client,
-    whitelisted_urls: Vec<String>,
+    proxy_config: Proxy,
```
SanchithHegde, line 192: Added the `proxy_config` field, which is required when constructing a client where certificates are specified.

crates/router/src/services/api/client.rs
```diff
@@ -218,68 +179,41 @@ where
...
-        if let Some(url) = proxy_config.http_url.as_ref() {
-            proxy_builder = proxy_builder.proxy(
-                reqwest::Proxy::http(url)
-                    .change_context(ApiClientError::InvalidProxyConfiguration)?,
-            );
-        }
-
-        let proxy_client = proxy_builder
+    pub fn new(proxy_config: &Proxy) -> CustomResult<Self, ApiClientError> {
+        let client = get_client_builder(proxy_config)?
```
SanchithHegde, line 199: Updated this to call `get_client_builder()`, since most of the code was duplicated.

crates/router/src/services/api/client.rs
```diff
@@ -218,68 +179,41 @@ where
...
         &self,
-        base_url: String,
         client_certificate: Option<masking::Secret<String>>,
         client_certificate_key: Option<masking::Secret<String>>,
     ) -> CustomResult<reqwest::Client, ApiClientError> {
         match (client_certificate, client_certificate_key) {
             (Some(certificate), Some(certificate_key)) => {
-                let client_builder =
-                    reqwest::Client::builder().redirect(reqwest::redirect::Policy::none());
+                let client_builder = get_client_builder(&self.proxy_config)?;
```
SanchithHegde, line 216: Updated this to call `get_client_builder()` as well, since the `get_reqwest_client()` method did not previously configure proxies for the `ClientBuilder`.

Chethan-rao (approved):

dracarys18 (approved):

NishantJoshi00 (approved):

PR #6961, opened 2024-12-30
test(cypress): Add Test for In Memory Cache

sumitdahiya125:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [x] CI/CD

## Description

Add Test Cases for testing In Memory Cache

### Additional Changes

## Motivation and Context

Currently test cases were not present.

## How did you test it?

Screenshot 2024-12-30 at 3 28 02 PM
Screenshot 2024-12-30 at 3 28 10 PM

## Checklist


cypress-tests/cypress/support/commands.js
```diff
@@ -3331,3 +3331,95 @@ Cypress.Commands.add("incrementalAuth", (globalState, data) => {
...
+    },
+    failOnStatusCode: false,
+  }).then((response) => {
+    logRequestId(response.headers["x-request-id"]);
+
+    expect(response.status).to.equal(200);
+    expect(response.body).to.have.property("key").to.equal(key);
+    expect(response.body).to.have.property("value").to.equal(value);
+  });
+});
```
pixincreate, line 3425: i would suggest you remove this part of code and instead, update the `updateConfig` command at [`L3219`](https://github.com/juspay/hyperswitch/blob/003cf1c1ff60ae642c347fd63d8224da51e082c3/cypress-tests/cypress/support/commands.js#L3219) as previously discussed.

also, make sure you add name to the assertion instead of having it plain. example:

```js
expect(
        response.body.incremental_authorization_allowed,
        "incremental_authorization_allowed" // this is the name to the assertion and having this makes it more meaningful
      ).to.be.true;
```

cypress-tests/cypress/support/commands.js
```diff
@@ -3331,3 +3331,95 @@ Cypress.Commands.add("incrementalAuth", (globalState, data) => {
...
+Cypress.Commands.add("deleteConfigs", (globalState, key, value) => {
+  const base_url = globalState.get("baseUrl");
+  const api_key = globalState.get("adminApiKey");
+
+  cy.request({
+    method: "DELETE",
+    url: `${base_url}/configs/${key}`,
+    headers: {
+      "Content-Type": "application/json",
+      "api-key": api_key,
```
pixincreate, line 3415: make sure you follow conventions:

- declare constants and variables before passing it to the `request` / `response` (exchange)
- use `camelCase` (there are a ton of violations in the code, we are slowly moving towards using `camelCase` conventions as it is the standard for `js`)

```suggestion
  const apiKey = globalState.get("adminApiKey");
  const baseUrl = globalState.get("baseUrl");
  const url: `${baseUrl}/configs/${key}`;

  cy.request({
    method: "DELETE",
    url: url,
    headers: {
      "Content-Type": "application/json",
      "api-key": apiKey,
```

cypress-tests/cypress/support/commands.js
```diff
@@ -3331,3 +3331,95 @@ Cypress.Commands.add("incrementalAuth", (globalState, data) => {
...
+    },
+    failOnStatusCode: false,
+  }).then((response) => {
+    logRequestId(response.headers["x-request-id"]);
+
+    expect(response.status).to.equal(200);
+    expect(response.body).to.have.property("key").to.equal(key);
+    expect(response.body).to.have.property("value").to.equal(value);
+  });
+});
```
pixincreate, line 3358: when creating configs, we're not passing the key as a param in the `url` but as a kv pair in the body.

i feel, we can update the existing `updateConfig` function to support creating configs as well (by passing another value to the function, something like `create`, `update`, `delete` and based on the check, either create a new config or update the existing one)? we can also pass relevant `method` based on this check.

but yeah, this check has to be done the beginning of the function, even before declaring the constants and variables.

cypress-tests/cypress/e2e/PaymentTest/00028-MemoryCacheConfigs.cy.js
```diff
@@ -0,0 +1,35 @@
...
+  });
+
+  after("flush global state", () => {
+    cy.task("setGlobalState", globalState.data);
+  });
+
+  context("Config flows", () => {
+    const key = "test-key";
+    const value = "test value";
+    const newValue = "new test value";
```
pixincreate, line 19: can these be made dynamic? you can use [`generateRandomString`](https://github.com/juspay/hyperswitch/blob/8092c1fd8fbe115ce7b3f4329758800e506bccae/cypress-tests/cypress/utils/RequestBodyUtils.js#L13) function from `RequestBodyUtils`.

Asking so because, i believe, `newValue` can be updated only once unless a different value is passed and this will cause errors when tests are run back to back without creating another merchant.

SanchithHegde: As part of the test, we're creating a config first, updating it and finally deleting it. So, if the test runs successfully, the config should not exist and the next run should not be a concern.

That said, since these configs are NOT associated with any merchant or profile (but have a "global" scope), if there are two Cypress processes running this test case simultaneously, then the test case _could_ fail because both test cases would read and write to the same config key.

If we're running this test case for every connector and there's a possibility that we'd run tests for connectors in parallel, then we should either consider moving this test case out of connector-specific test cases, or consider suffixing a random string to reduce the possibility of collisions. I'm personally leaning towards the former, since having tests that are not associated with a merchant or a profile is a valid use case.

cypress-tests/cypress/support/commands.js
```diff
@@ -3331,3 +3331,95 @@ Cypress.Commands.add("incrementalAuth", (globalState, data) => {
...
+
+  cy.request({
+    method: "POST",
+    url: `${base_url}/configs/${key}`,
+    headers: {
+      "Content-Type": "application/json",
+      "api-key": api_key,
+    },
+    body: {
+      key: key,
```
SanchithHegde, line 3393: We don't need the `key` to be provided in the request body, right? I see a `#[serde(skip_deserializing)]` annotation on the field:

https://github.com/juspay/hyperswitch/blob/7d00583a8076ef2f996345549b5e81c7f90361dc/crates/router/src/types/api/configs.rs#L7-L12

CC: @dracarys18

cypress-tests/cypress/support/commands.js
SanchithHegde: In my opinion, this test case is rather simple and even though there may be some code duplication, it's minimal. I think we need not consider optimizing it now.

cypress-tests/cypress/support/commands.js
dracarys18: Yes we dont need to send it in the body

pixincreate (approved):

pixincreate:
i've addressed the changes in #6992

cypress-tests/cypress/support/commands.js
pixincreate: addressed in [`06f3c42` (#6992)](https://github.com/juspay/hyperswitch/pull/6992/commits/06f3c42a0b903dde3b09ed2efb39accbadf162a5)

Gnanasundari24 (approved):

PR #6968, opened 2025-01-02
chore(keymanager): add tenant-id to keymanager requests

dracarys18:
## Type of Change

- [x] Enhancement

## Description

Adds tenant-id for every requests to keymanager

## Motivation and Context

Added tenant id header for the keymanager service to classify key ids based on tenants.

## How did you test it?

- Create Merchant Account with x-tenant-id as `public`
```bash
--header 'x-feature: integ-custom' \
--data-raw '{
  "merchant_id": "1735815159",
  "locker_id": "m0010",
  "merchant_details": {
    "website": "www.example.com",
    "address": {
      "line1": "1467",
      "line2": "Harrison Street",
      "line3": "Harrison Street",
      "city": "San Fransico",
      "state": "California",
      "zip": "94122",
      "country": "US"
    }
  },
  "webhook_details": {
    "webhook_version": "1.0.1",
  },
  "metadata": {
    "city": "NY",
    "unit": "245"
  },
  "primary_business_details": [
    {
      "country": "US",
      "business": "default"
    }
  ]
}'
```
- Get the `x-request-id` for the request and query it in grafana and see the if the tenant_id is `public`
Screenshot 2025-01-02 at 4 24 01 PM
## Checklist

### DIFF
--- a/config/development.toml
+++ b/config/development.toml
@@ -794,7 +794,7 @@ sdk_eligible_payment_methods = "card"
 
 [multitenancy]
 enabled = false
-global_tenant = { tenant_id = "global" ,schema = "public", redis_key_prefix = "", clickhouse_database = "default"}
+global_tenant = { tenant_id = "global" ,schema = "public", redis_key_prefix = "global", clickhouse_database = "default"}
 
 [multitenancy.tenants.public]
 base_url = "http://localhost:8080"
--- a/config/development.toml
+++ b/config/development.toml
@@ -794,7 +794,7 @@ sdk_eligible_payment_methods = "card"
 
 [multitenancy]
 enabled = false
-global_tenant = { schema = "public", redis_key_prefix = "", clickhouse_database = "default"}
+global_tenant = { tenant_id = "global" ,schema = "public", redis_key_prefix = "", clickhouse_database = "default"}
 
 [multitenancy.tenants.public]
 base_url = "http://localhost:8080"
--- a/crates/common_utils/src/id_type/tenant.rs
+++ b/crates/common_utils/src/id_type/tenant.rs
@@ -1,5 +1,7 @@
 use crate::errors::{CustomResult, ValidationError};
 
+const DEFAULT_GLOBAL_TENANT_ID: &str = "global";
+
 crate::id_type!(
     TenantId,
     "A type for tenant_id that can be used for unique identifier for a tenant"
@@ -15,10 +17,10 @@ crate::impl_queryable_id_type!(TenantId);
 crate::impl_to_sql_from_sql_id_type!(TenantId);
 
 impl TenantId {
-    /// Construct TenantID without checking for length constraints
-    pub fn new_unchecked(input_string: String) -> Self {
+    /// Get the default global tenant ID
+    pub fn get_default_global_tenant_id() -> Self {
         Self(super::LengthId::new_unchecked(
-            super::AlphaNumericId::new_unchecked(input_string),
+            super::AlphaNumericId::new_unchecked(DEFAULT_GLOBAL_TENANT_ID.to_string()),
         ))
     }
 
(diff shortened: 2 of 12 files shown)


config/development.toml
```diff
@@ -794,7 +794,7 @@ sdk_eligible_payment_methods = "card"
 
 [multitenancy]
 enabled = false
-global_tenant = { schema = "public", redis_key_prefix = "", clickhouse_database = "default"}
+global_tenant = { tenant_id = "global" ,schema = "public", redis_key_prefix = "", clickhouse_database = "default"}
```
SanchithHegde, line 797: Should we consider setting `schema` and `redis_key_prefix` to be `global` here?

crates/router/src/configs/defaults.rs
```diff
@@ -138,6 +139,17 @@ impl Default for super::settings::KvConfig {
     }
 }
 
+impl Default for super::settings::GlobalTenant {
+    fn default() -> Self {
+        Self {
+            tenant_id: id_type::TenantId::new_unchecked(String::new()),
```
SanchithHegde, line 145: I feel passing in empty string here could cause issues when this value is being serialized/deserialized.

@Narayanbhat166 Correct me if I'm wrong.

config/development.toml
```diff
@@ -794,7 +794,7 @@ sdk_eligible_payment_methods = "card"
 
 [multitenancy]
 enabled = false
-global_tenant = { schema = "public", redis_key_prefix = "", clickhouse_database = "default"}
+global_tenant = { tenant_id = "global" ,schema = "public", redis_key_prefix = "global", clickhouse_database = "default"}
```
SanchithHegde, line 797: Could you replicate similar changes in the other TOML config files which have `global_tenant` configuration?

crates/common_utils/src/id_type/tenant.rs
```diff
@@ -15,6 +15,13 @@ crate::impl_queryable_id_type!(TenantId);
 crate::impl_to_sql_from_sql_id_type!(TenantId);
 
 impl TenantId {
+    /// Construct TenantID without checking for length constraints
+    pub fn new_unchecked(input_string: String) -> Self {
```
Narayanbhat166, line 19: This function should not be public

dracarys18: We need this public to be used in router crate :/

crates/router/src/configs/defaults.rs
dracarys18: I believe the `new_unchecked` does not validate the length

SanchithHegde: But deserialization would. So, even if the application does start running with the empty string as the tenant ID, and say the tenant ID is serialized / deserialized when writing to / reading from Redis, then the serialization may go through but the deserialization would fail, thus possibly causing an internal server error.

crates/common_utils/src/id_type/tenant.rs
Narayanbhat166: then we can create the function which will create the tenant id with specified name, we cannot accept any generic string for unchecked functions.

Reference: https://github.com/juspay/hyperswitch/blob/1f20a9577ccdbe794faa498c8b250f7f6ad05703/crates/common_utils/src/id_type/merchant.rs#L56

crates/router/src/configs/defaults.rs
Narayanbhat166: why can't we have the default tenant id as well? isn't it `global`?

crates/common_utils/src/id_type/tenant.rs
dracarys18: Okay

dracarys18: Done

Narayanbhat166 (approved):

NishantJoshi00 (approved):

crates/router/src/configs/settings.rs
```diff
@@ -195,8 +195,10 @@ impl storage_impl::config::TenantConfig for Tenant {
     }
 }
 
-#[derive(Debug, Deserialize, Clone, Default)]
+#[derive(Debug, Deserialize, Clone)]
 pub struct GlobalTenant {
+    #[serde(default = "id_type::TenantId::get_default_global_tenant_id")]
+    pub tenant_id: id_type::TenantId,
```
SanchithHegde, line 201: Why did we need this `serde(default)`?

SanchithHegde (approved):

PR #6985, opened 2025-01-03
chore(roles): remove redundant variant from PermissionGroup

kashif-m:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [x] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description
This removes the redundant `ReconOps` variant from `PermissionGroup` enum which is stored in DB. This was kept back in application for backwards compatibility. Migrations are added for this and this will not be needed anymore.

### Additional Changes

## Motivation and Context
Helps in keeping a clean permission view on HS dashboard.

## How did you test it?
Locally by running migrations and viewing permission view in dashboard.

## Checklist


migrations/2025-01-03-104019_migrate_permission_group_for_recon/down.sql
```diff
@@ -0,0 +1 @@
+SELECT 1;
```
ThisIsMani, line 1: Should we replace the groups back here? If someone rolls back to older version, new variant wouldn't work.

kashif-m: We cannot revert the groups back, this is a one time activity for renaming `recon_ops` to `recon_ops_manage`. Once this is done, we will lose context of which groups were `recon_ops`, and that's expected since application expects `recon_ops_manage`

If someone rolls back to an older version of the application, it will work regardless since we've had the new variant (ReconOpsManage) for a while.

ThisIsMani: But I don't see any issue in replacing either... If someone reverts to the version where `ReconOpsManage` is not present, then it wouldn't work without replacing. 

I would not expect anyone to do so. But, this is something I can think of which will give us advantage without any downsides...

jarnura (approved):

PR #7041, opened 2025-01-15
fix: fallback to key without prefix if with prefix fails

dracarys18:
## Type of Change

- [x] Bugfix

## Description

Fallback to key without prefix if with prefix fails

## Motivation and Context

To keep the tenant data isolated from each other, We are using a prefix for Redis keys i.e A key which would be saved as card123 without multitenancy would be saved as public:card123 with Multitenancy.

If the user makes a payment on the older pod where MultiTenancy is disabled and  client_secret is saved without a prefix, Now if the confirm calls goes to the pod where Multitenancy is enabled the Payment would fail with Unauthorized. This is just one of the instances where the payments would fail.

https://github.com/juspay/hyperswitch/pull/4630

## How did you test it?

NA

## Checklist

### DIFF
--- a/crates/redis_interface/src/commands.rs
+++ b/crates/redis_interface/src/commands.rs
@@ -65,7 +65,13 @@ impl super::RedisConnectionPool {
         V::Error: Into<fred::error::RedisError> + Send + Sync,
     {
         self.pool
-            .set(key, value, Some(Expiration::KEEPTTL), None, false)
+            .set(
+                self.add_prefix(key),
+                value,
+                Some(Expiration::KEEPTTL),
+                None,
+                false,
+            )
             .await
             .change_context(errors::RedisError::SetFailed)
     }
@@ -165,10 +171,19 @@ impl super::RedisConnectionPool {
     where
         V: FromRedis + Unpin + Send + 'static,
     {
-        self.pool
+        match self
+            .pool
             .get(self.add_prefix(key))
             .await
             .change_context(errors::RedisError::GetFailed)
+        {
+            Ok(v) => Ok(v),
+            Err(_) => self
+                .pool
+                .get(key)
+                .await
+                .change_context(errors::RedisError::GetFailed),
+        }
     }
 
     #[instrument(level = "DEBUG", skip(self))]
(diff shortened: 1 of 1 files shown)


crates/redis_interface/src/commands.rs
```diff
@@ -165,21 +171,39 @@ impl super::RedisConnectionPool {
...
     {
-        self.pool
+        match self
+            .pool
             .get(self.add_prefix(key))
             .await
             .change_context(errors::RedisError::GetFailed)
+        {
+            Ok(v) => Ok(v),
+            Err(_) => self
```
NishantJoshi00, line 181: can you log this error and then do the next get?

Also, this approach will have a performance implication if the key really doesn't exist. Will we remove this later?

crates/redis_interface/src/commands.rs
```diff
@@ -165,21 +171,39 @@ impl super::RedisConnectionPool {
...
     {
-        self.pool
+        match self
+            .pool
             .exists(self.add_prefix(key))
             .await
             .change_context(errors::RedisError::GetFailed)
+        {
+            Ok(v) => Ok(v),
+            Err(_) => self
```
NishantJoshi00, line 201: Same here

crates/redis_interface/src/commands.rs
```diff
@@ -497,10 +521,19 @@ impl super::RedisConnectionPool {
...
     {
-        self.pool
+        match self
+            .pool
             .hget(self.add_prefix(key), field)
             .await
             .change_context(errors::RedisError::GetHashFieldFailed)
+        {
+            Ok(v) => Ok(v),
+            Err(_) => self
```
NishantJoshi00, line 531: And here

crates/redis_interface/src/commands.rs
dracarys18: Yes this is just temporary we won't merge it to main this will be deployed on prod as a hotfix

NishantJoshi00 (approved):
This is a temporary fix, makes sense. Cool to go forward with it

crates/redis_interface/src/commands.rs
jagan-jaya: Don't we need it for delete_key?

jarnura (approved):

crates/redis_interface/src/commands.rs
jarnura: We might need this in main too, or we need to give SOP for open-source users, how to release this version. If the changes are not in main then, they might face issue.

dracarys18:
Closing this PR #7044

PR #7043, opened 2025-01-16
fix(multitenancy): add a fallback for get commands in redis

dracarys18:
## Type of Change

- [x] Enhancement

## Description

- Add a fallback for Redis gets with prefix
- Add a Domain type for Redis for the correctness of adding a prefix

## Motivation and Context

To keep the tenant data isolated from each other, We are using a prefix for Redis keys i.e A key which would be saved as card123 without multitenancy would be saved as public:card123 with Multitenancy.
If the user makes a payment on the older pod where MultiTenancy is disabled and  client_secret is saved without a prefix, Now if the confirm calls goes to the pod where Multitenancy is enabled the Payment would fail with Unauthorized. This is just one of the instances where the payments would fail.

## How did you test it?

** THIS CANNOT BE TESTED ON HIGHER ENVIRONMENTS **

## Checklist


jarnura (approved):

racnan (dismissed):

crates/redis_interface/src/commands.rs
```diff
@@ -201,19 +243,39 @@ impl super::RedisConnectionPool {
...
+                {
+                    Err(_err)
+                }
+
+                #[cfg(feature = "multitenancy_fallback")]
+                {
+                    self.pool
+                        .del(key.tenant_unaware_key(self))
+                        .await
+                        .change_context(errors::RedisError::GetFailed)
```
Chethan-rao, line 265:
```suggestion
                        .change_context(errors::RedisError::DeleteFailed)
```

crates/redis_interface/src/commands.rs
```diff
@@ -201,19 +243,39 @@ impl super::RedisConnectionPool {
...
     #[instrument(level = "DEBUG", skip(self))]
     pub async fn delete_multiple_keys(
         &self,
-        keys: &[String],
+        keys: &[RedisKey],
     ) -> CustomResult<Vec<DelReply>, errors::RedisError> {
-        let futures = keys.iter().map(|key| self.pool.del(self.add_prefix(key)));
+        let futures = keys
+            .iter()
+            .map(|key| self.pool.del(key.tenant_aware_key(self)));
```
Chethan-rao, line 278: why are we not matching on result and if its error, we should be retrying the delete with `tenant_unaware_key`??
Also I think we can make use of `delete_key` method itself 
```
let futures = keys.iter().map(|key| self.delete_key(key));
```

And then `delete_key` would take care of tenant aware and tenant unaware deletion

Chethan-rao (changes requested):
Also why are none of the stream redis commands not having this fallback logic?

dracarys18:
> Also why are none of the stream redis commands not having this fallback logic?

Retrying streams is a little tricky because `xread` will not throw any error if the stream is empty obviously. And in this case the streams with prefix will mostly be empty. And only way to retry it is to drain from both the streams. And I am not sure what sort of problems that could come with it. So I think it's better to do that consciously rather than trying it in the code.

Chethan-rao (approved):

racnan (approved):

jarnura (approved):

PR #7102, opened 2025-01-24
chore: bump cypress to `v14.0.0`

### LINKED ISSUE #7103: [CHORE] Bump Cypress to v14
bump cypress version to `v14` from `v13.17.0`. there's a ton of under the hood changes and improvements.

check: [docs.cypress.io/app/references/changelog](https://docs.cypress.io/app/references/changelog)

cypress v14 has significant impact on the way we do the redirection flow. check pr description for detailed information.

pixincreate:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [x] Dependency updates
- [ ] Documentation
- [x] CI/CD

## Description

closes #7103

- video recording
  - video recording gets around the race condition that resulted in parallel tests to fail when a payment failed at the same time
  - connector specific recording
  - store only failed payment recording
- version bump
  - cypress to `v14.0.0` (changelog: https://docs.cypress.io/app/references/changelog)
    - this is a major change. explained below*.
  - eslint to `v9.19.0`
- fix multiple credential wrongly picking `connector_1` for trustpay resulting in `refund` test to fail. logic is as follows:
  - `determineConnectorConfig` is where the problem existed
  - we have 2 objects and 5 values
    - connector credential takes in 3 values -- either next connector (boolean) or value with or without spec name
    - if connector credential is null or it does not exist, default to `connector_1`
    - if connector credential has `next connector` boolean set to `true`, check if multiple connector status is `true` and count is greater than `1`. if `true`, return `connector_2`, else default to using `connector_1`
    - if connector credential has spec name with value, return value only for that spec name (this will affect config only in that respective `spec`), else, just return value that affects the config in every flow
- add security headers
- cypress v14 specific changes*
  - use camelCase in redirection flow
  - sub domains will now require us to be passed within `cy.origin()` effectively making it impossible to work with redirection tests without below mentioned changes
  - once the url is visited, wait until redirection is completed (wait until domain is changed to connector's domain)
  - use `cy.url()` to get redirected url (connector's url), which is a promise and it cannot be accessed outside of that promise, and is then passed to `cy.origin()` that enables us to interact with the ui elements
  - in verify url function, wait until redirection is completed and we're back to `hyperswitch.io` website by verifying the host
  - once, on the returned url, access the present url (returned url) with `cy.url` which is then passed to `cy.origin()` so that we can access the elements just so that we can verify that there's no errors like `4xx`
  - access the url and then verify the terminal payment status
- bug fixes
  - fix przelwey24 and eps bank redirect in adyen
  - fix upi in adyen
  - uncaught exception resulting in abrupt failure in adyen tests (happens only in ui)
    ```js
	__all:1 Uncaught (in promise) 
	Event {isTrusted: true, Symbol(ERR_PREPARED_FOR_SERIALIZATION): true, type: 'error', target: null, currentTarget: null, …}
	isTrusted: true
	Symbol(ERR_PREPARED_FOR_SERIALIZATION): true
	bubbles: false
	cancelBubble: false
	cancelable: false
	composed: false
	currentTarget: null
	defaultPrevented: false
	eventPhase: 0
	returnValue: true
	srcElement: null
	target: null
	timeStamp: 298220.2000000002
	type: "error"
	```

### Additional Changes

## Motivation and Context

just keeping up-to-date.

## How did you test it?

ci should pass. especially the redirections.

image

## Checklist

- [x] I formatted the code `npm run format`
- [x] I addressed lints thrown by `npm run lint`


Sakilmostak (approved):

pixincreate:
marking this pr as draft since ci is failing only in github ci after upgrading to v14 and no where else.

likhinbopanna (approved):

Gnanasundari24 (approved):

PR #7178, opened 2025-02-04
feat(process_tracker): Task implementation for psync payments

Aprabhat19:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

- This PR consists of the process tracker flow for Psync and Review Tasks

-  **Psync Task** is created if a payment’s status is Processing with an active attempt_id
	> Action: match attempt status and update the process tracker status based on that.
	> Finish the current task
- This PR also includes changes in the execute_payment workflow where , we record an attempt first and then call the proxy api to perform the payments
Screenshot 2025-03-12 at 6 25 37 PM

- if the payment fails it would retry in a given interval for given number of default retries
Screenshot 2025-03-13 at 2 07 42 PM

- If thePayment goes to processing , a PSYNC_WORKFLOW Is Scheduled

### Additional Changes

## How did you test it?
- Create a MA and a MCA for payments and an MCA for billing connectors with a threshold to retry the payments
- When a failure webhook  is received a payment intent is created and multiple attempts are created for the external failed payments
- Once the threshold is reached the payment is inserted to the Process tracker, which is picked after the schedule time
- A payment attempt is created, which could either be  Succeed or Failed 
- iF payment is processing a `Psync-Workflow` is to be created

## Checklist


crates/hyperswitch_connectors/src/connectors/chargebee/transformers.rs
```diff
@@ -488,13 +488,13 @@ impl TryFrom<ChargebeeWebhookBody> for revenue_recovery::RevenueRecoveryAttemptD
         Ok(Self {
             amount,
             currency,
-            merchant_reference_id,
+            merchant_reference_ids: Some(merchant_reference_id),
```
srujanchikke, line 491: Why should we make this an optional field ? We will get this field in payment intent. Making this optional  will lead to create an attempt without merchant reference id and this field is necessary in record back flow where we actually update the invoice in billing connector.

crates/hyperswitch_connectors/src/connectors/chargebee/transformers.rs
```diff
@@ -488,13 +488,13 @@ impl TryFrom<ChargebeeWebhookBody> for revenue_recovery::RevenueRecoveryAttemptD
...
             connector_transaction_id,
             error_code,
             error_message,
             processor_payment_method_token: connector_mandate_details.mandate_id,
             connector_customer_id: connector_mandate_details.customer_id,
-            connector_account_reference_id,
+            connector_account_reference_ids: Some(connector_account_reference_id),
```
srujanchikke, line 497: Same for this field as well, we can find this in billing merchant connector account.

crates/hyperswitch_domain_models/src/payments.rs
```diff
@@ -675,6 +675,41 @@ impl PaymentIntent {
...
 
+    pub fn create_revenue_recovery_attempt_data(
+        &self,
+        revenue_recovery_metadata: api_models::payments::PaymentRevenueRecoveryMetadata,
+        billing_connector_account: &merchant_connector_account::MerchantConnectorAccount,
+    ) -> revenue_recovery::RevenueRecoveryAttemptData {
+        revenue_recovery::RevenueRecoveryAttemptData {
+            amount: self.amount_details.order_amount,
+            currency: self.amount_details.currency,
+            merchant_reference_ids: self.merchant_reference_id.clone(), // is the intent id
```
srujanchikke, line 686: This is not intent id. This is invoice id at billing connector which we will use it to update the invoice status in record back flow.

crates/router/src/core/payments/operations/proxy_payments_intent.rs
```diff
@@ -185,26 +185,40 @@ impl<F: Send + Clone + Sync> GetTracker<F, PaymentConfirmData<F>, ProxyPaymentsR
...
+                    active_attempt_id,
+                    storage_scheme,
+                )
+                .await
+                .change_context(errors::ApiErrorResponse::PaymentNotFound)
+                .attach_printable("Could not find payment attempt")?,
+
+            None => {
+                let payment_attempt_domain_model: hyperswitch_domain_models::payments::payment_attempt::PaymentAttempt =
+                hyperswitch_domain_models::payments::payment_attempt::PaymentAttempt::proxy_create_domain_model(
```
srujanchikke, line 202: Previously we were creating payment attempt and now if active attempt is not present then we are creating payment attempt. Is it added to support normal proxy flow(None case)?

Aprabhat19: Yes

crates/hyperswitch_connectors/src/connectors/chargebee/transformers.rs
Aprabhat19: Done

crates/hyperswitch_domain_models/src/payments.rs
Aprabhat19: Removed the commented line

srujanchikke (approved):

crates/router/src/core/revenue_recovery.rs
```diff
@@ -45,57 +30,78 @@ pub const PSYNC_WORKFLOW: &str = "PSYNC_WORKFLOW";
...
+                    Box::pin(action.execute_payment_task_response_handler(
+                        state,
+                        payment_intent,
+                        execute_task_process,
+                        revenue_recovery_payment_data,
+                        &mut revenue_recovery_metadata,
+                    ))
+                    .await?;
+                }
+                Err(_) => {
```
jarnura, line 90: log the error don't discard

jarnura (approved):

jarnura (approved):

srujanchikke (dismissed):

jarnura (approved):

PR #7199, opened 2025-02-05
refactor(core): filter default routing config response based on connector type

prajjwalkumar17:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

### Additional Changes

## Checklist

### DIFF
--- a/crates/router/src/core/routing.rs
+++ b/crates/router/src/core/routing.rs
@@ -932,16 +932,54 @@ pub async fn retrieve_default_routing_config(
 ) -> RouterResponse<Vec<routing_types::RoutableConnectorChoice>> {
     metrics::ROUTING_RETRIEVE_DEFAULT_CONFIG.add(1, &[]);
     let db = state.store.as_ref();
+    let key_manager_state = &(&state).into();
+    let key_store = state
+        .store
+        .get_merchant_key_store_by_merchant_id(
+            key_manager_state,
+            merchant_account.get_id(),
+            &state.store.get_master_key().to_vec().into(),
+        )
+        .await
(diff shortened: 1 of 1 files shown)


crates/router/src/core/routing.rs
```diff
@@ -932,16 +932,54 @@ pub async fn retrieve_default_routing_config(
...
-            service_api::ApplicationResponse::Json(conn_choice)
-        })
+    let mut connectors = Vec::new();
+    let conn_choice = helpers::get_merchant_default_config(db, &id, transaction_type).await?;
+    for connector in conn_choice.iter() {
+        if let Some(mca_id) = &connector.merchant_connector_id {
+            let mca = state
+                .store
+                .find_by_merchant_connector_account_merchant_id_merchant_connector_id(
+                    key_manager_state,
```
Aprabhat19, line 957: Instead of doing a db call for every mca , can you do it once and retrieve all the mca's and then filter it based on the connector type,  you can use the following function to do so `find_merchant_connector_account_by_merchant_id_and_disabled_list`, in v1

Aprabhat19:
Also , can we add this filtering logic in v2 as well?

crates/router/src/core/routing.rs
prajjwalkumar17: For v2 the changes will be taken up in a separate PR. Here is the [issue](https://github.com/juspay/hyperswitch/issues/7305).

Aprabhat19 (dismissed):

kashif-m (approved):

kashif-m (approved):

PR #7308, opened 2025-02-19
ci: minor refactors and improvements

SanchithHegde:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [x] CI/CD

## Description

This PR includes minor refactors and improvements to our CI workflows:

1. The Docker build workflow files are not being used, and have been removed.
2. The connector sanity and UI sanity workflows are not being actively used. I didn't want to remove them completely, so instead "archived" them for later reference.
3. Dependency updates: The versions of the actions that we use have been updated.
4. Replaced usages of `baptiste0928/cargo-install` with `taiki-e/install-action` to install tools. This is done because if the binary is not available in cache, the `baptiste0928/cargo-install` action builds it from source, which can take a longer amount of time, as compared to `taiki-e/install-action`, which should be deterministic since it pulls pre-built binaries from releases.
5. We were using the `CARGO_REGISTRIES_CRATES_IO_PROTOCOL` environment variable to enable sparse index, but since it is enabled by default on recent Rust versions, I've removed it.
6. Replaced usages of `postgres:latest` and `redis:latest` with `postgres:alpine` and `redis:alpine` respectively, to slightly reduce bandwidth used when pulling images.
7. The Postman workflow has a `workflow_dispatch` trigger, which hasn't been used in quite a while, so I've removed it.
8. The `CI-push` workflow was missing some environment variables that allowed us to handle flaky CI failures due to network issues, so I copied them from the `CI-pr` workflow.

## Motivation and Context

Minor improvements and dependency updates to the CI workflows.

## How did you test it?

N/A

## Checklist


pixincreate (commented):
other than that, lgtm

pixincreate (approved):

prasunna09 (approved):
Dockerfile' changes lgtm

PR #6798, opened 2024-12-11
fix(webhooks): mask custom outgoing webhook headers in profile response

srujanchikke:
## Type of Change

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

Mask custom outgoing webhook headers in profile response. This changes is required to mask the custom outgoing webhook headers in the profile dashboard. This also requires control center changes, BE changes will only be merged and 
deployed after dashboard changes.

### Additional Changes

- [ ] This PR modifies the API contract
- [ ] This PR modifies the database schema
- [ ] This PR modifies application configuration/environment variables

## How did you test it?

Test Case 1 : Profile create 

```
curl --location 'http://localhost:8080/account/merchant_1733900220/business_profile' \
--header 'Content-Type: application/json' \
--header 'api-key: ••••••' \
--data-raw '{
    "profile_name" : "test_profile1",
    "webhook_details": {
        "webhook_url": "https://webhook.site/672f0697-9e14-4a49-abef-b15769c19962",
        "webhook_version": "1.0.1",
        "webhook_username": "ekart_retail",
        "webhook_password": "password_ekart@123",
        "payment_created_enabled": true,
        "payment_succeeded_enabled": true,
        "payment_failed_enabled": true
    },
    "outgoing_webhook_custom_http_headers": {
        "authentication": "this is test key header",
        "test_key": "asd",
        "client_key": "asdkfjhakdjh"
    }
}'
```
Response 
```
{
    "merchant_id": "merchant_1733900220",
    "profile_id": "pro_nvV4ljCmGfPnUiRbjjMW",
    "profile_name": "test_profile1",
    "return_url": "https://google.com/success",
    "enable_payment_response_hash": true,
    "payment_response_hash_key": "IurVIL6k065SRjO2TWHcx6cf2mQCWO1SKf1tep6AZ79XBFBcLJFvsfpZ3JWRkxHy",
    "redirect_to_merchant_with_http_post": false,
    "webhook_details": {
        "webhook_version": "1.0.1",
        "webhook_username": "ekart_retail",
        "webhook_password": "password_ekart@123",
        "webhook_url": "https://webhook.site/672f0697-9e14-4a49-abef-b15769c19962",
        "payment_created_enabled": true,
        "payment_succeeded_enabled": true,
        "payment_failed_enabled": true
    },
    "metadata": null,
    "routing_algorithm": null,
[...]

PR #7309, opened 2025-02-19
refactor(payment_methods_v2): refactor network tokenization flow for v2

prasunna09:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [x] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

Refactor network tokenization flow for v2 payment method. 
1. moved the network tokenization req, response types from domain to router types, since the types are not generic for         Hyperswitch but req and response types of token service.
2. replaced hanging functions to impl based in the flow
3. Update ListCustomerPaymentMethod response and Payment Method response with network token details.

### Additional Changes

## Checklist


crates/router/src/core/payment_methods/network_tokenization.rs
```diff
@@ -30,7 +30,7 @@ use crate::{
     routes::{self, metrics},
     services::{self, encryption},
     settings,
-    types::{api, domain},
+    types::{api, domain, payment_methods as payment_method_types},
```
Narayanbhat166, line 33: in domain payment methods is exported for public use right? why do we need to import here again from payment_methods?

Narayanbhat166 (approved):

crates/router/src/core/payment_methods/network_tokenization.rs
prasunna09: actually this is router types not domain payment methods

crates/router/src/types/domain.rs
```diff
@@ -54,8 +63,12 @@ pub use event::*;
 pub use merchant_connector_account::*;
 pub use merchant_key_store::*;
 pub use network_tokenization::*;
+pub use payment_method_data::*;
 pub use payment_methods::*;
+#[cfg(all(feature = "v2", feature = "payment_methods_v2"))]
 pub use payments::*;
```
SanchithHegde, line 69: Why is the `payments` module exports behind the `payment_methods_v2` feature flag?

crates/hyperswitch_domain_models/src/vault.rs
```diff
@@ -0,0 +1,56 @@
+use std::fmt::Debug;
```
SanchithHegde, line 1: Why was this import added?

crates/router/src/types/payment_methods.rs
```diff
@@ -1,9 +1,20 @@
+use std::fmt::Debug;
```
SanchithHegde, line 1: Why was this import added?

crates/router/src/core/payment_methods/transformers.rs
```diff
@@ -1068,16 +1090,18 @@ impl
...
             original_payment_authorized_amount,
             original_payment_authorized_currency,
             metadata,
-            token: connector_token,
+            token: Secret::new(connector_token),
             // Token that is derived from payments mandate reference will always be multi use token
             token_type: common_enums::TokenizationType::MultiUse,
-        }
+        };
+        res.set_masked_token();
```
SanchithHegde, line 1104: Won't this break the multi-use PSP token flow? What's the point of returning a masked PSP token if the PSP token is meant to be used by the merchant?

CC: @Narayanbhat166

crates/api_models/src/payment_methods.rs
```diff
@@ -901,7 +904,37 @@ pub struct ConnectorTokenDetails {
...
+                } else {
+                    masked_val.push('*'); // Mask the middle characters
+                }
+            }
+            masked_val
+        };
+        masking::Secret::new(masked_val)
+    }
+
+    pub fn set_masked_token(&mut self) {
```
SanchithHegde, line 935: This suffers from the same concerns pointed out in https://github.com/juspay/hyperswitch/pull/6798#discussion_r1880781752, and allows us to construct a `ConnectorTokenDetails` without a masked value for the `token` field.

If providing a masked token in the payment method response is still intentional, then can we ensure that the `token` field would always be masked? 

Maybe we can define a newtype like this, which would be used for the `token` field?
https://github.com/juspay/hyperswitch/blob/d1f53036c75771d8387a9579a544c1e2b3c17353/crates/common_utils/src/new_type.rs#L78-L91

crates/hyperswitch_domain_models/src/vault.rs
```diff
@@ -0,0 +1,56 @@
...
+    fn get_vaulting_data_key(&self) -> String;
+}
+
+impl VaultingDataInterface for PaymentMethodVaultingData {
+    fn get_vaulting_data_key(&self) -> String {
+        match &self {
+            Self::Card(card) => card.card_number.to_string(),
+            Self::NetworkToken(network_token) => network_token.network_token.to_string(),
+        }
+    }
```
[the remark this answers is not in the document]
prasunna09: Currently this Vaulting Interface is not used anywhere, but will take this up in separate pr

crates/router/src/types/payment_methods.rs
prasunna09: request response types are present in this file which uses Debug

crates/router/src/core/payment_methods/transformers.rs
prasunna09: yes, will takes these changes in separate pr, for now reverting this change

Narayanbhat166 (dismissed):

crates/router/src/types/payment_methods.rs
SanchithHegde: The `Debug` derive macro has been [in the Rust prelude since Rust 1.0](https://doc.rust-lang.org/std/prelude/v1/index.html), due to which we don't have to explicitly import it, if using the derive macro. However, we are required to import it when implementing the `Debug` trait manually, if I remember correctly.

SanchithHegde (dismissed):

ShankarSinghC (approved):

PR #5557, opened 2024-08-07
fix(core): Update pm_status accordingly for the respective attempt status

Aprabhat19:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description
Update pm_status accordingly for the respective attempt status. Earlier we were mapping the AttemptStatus to Active in all the cases other than Failure. In this PR we must update the pm_status as Active only if it's charged / Authorised by the connector. 

### Additional Changes

- [ ] This PR modifies the API contract
- [ ] This PR modifies the database schema
- [ ] This PR modifies application configuration/environment variables

## How did you test it?

- Create a 3ds payment
>  
```
Create
curl --location 'http://localhost:8080/payments' \
--header 'Content-Type: application/json' \
--header 'Accept: application/json' \
--header 'api-key: dev_uaSNBro089ifFDTOyfoTqV6fEZOIkjtEUCDeHMDbwolILfq885p0xt0Nu9pwGLW9' \
--data-raw '{
    "amount": 10000,
    "currency": "USD",
    "confirm": false,
    
    "customer_id": "CustomerX7777",
    "capture_method": "automatic",
    "capture_on": "2022-09-10T10:11:12Z",
    "authentication_type": "three_ds",
    "return_url": "https://google.com",
     "email": "something@gmail.com",
    "name": "Joseph Doe",
    "phone": "999999999",
    "phone_country_code": "+65",
    "description": "Its my first payment request",   
"setup_future_usage":"off_session",
    "billing": {
        "address": {
            "line1": "1467",
            "line2": "Harrison Street",
            "line3": "Harrison Street",
            "city": "San Fransico",
            "state": "California",
            "zip": "94122",
            "country": "US",
            "first_name": "joseph",
            "last_name": "Doe"
        },
        "phone": {
            "number": "8056594427",
            "country_code": "+91"
        }
    },
    

    
    "shipping": {
        "address": {
            "line1": "1467",
            "line2": "Harrison Street",
            "line3": "Harrison Street",
            "city": "San Fransico",
            "state": "California",
            "zip": "94122",
            "country": "US",
            "first_name": "joseph",
            "last_name": "Doe"
        },
        "phone": {
            "number": "8056594427",
            "country_code": "+91"
        }
    },
    "connector_metadata": {
        "noon": {
            "order_category": "pay"
        }
    }
   
    
}'

```

>
  ```
[...]

PR #7496, opened 2025-03-12
ci(cypress): verify `payment_method_status` for respective attempt status

pixincreate:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [x] CI/CD

## Description

This PR adds additional validation for existing tests to check the payment method status.

### Additional Changes

## Motivation and Context

https://github.com/juspay/hyperswitch/pull/5557

## How did you test it?

ci should pass.

## Checklist

- [x] I formatted the code `npm run format`
- [x] I addressed lints thrown by `npm run lint`


cypress-tests/cypress/support/commands.js
```diff
@@ -2155,13 +2155,41 @@ Cypress.Commands.add(
                 "nextActionUrl",
                 response.body.next_action.redirect_to_url
               );
+
+              if (
+                typeof response.body.payment_method_id !== "undefined" ||
```
likhinbopanna, line 2160: Why do you want to use undefined , can't we use ?

```javascript
if (response.body.payment_method_id){
expect(response.body.payment_method_status,"payment_method_status").to.equal("active");
}

File: cypress-tests/cypress/support/commands.js [hunk 2]
```diff
@@ -2339,6 +2368,33 @@ Cypress.Commands.add(
             ).to.equal(merchant_connector_id);
           }
 
+          if (
+            response.body.payment_method_id &&
+            typeof response.body.payment_method_id === "string"
```
likhinbopanna, line 2373: Why do you want to verify `typeof` the payment method id ?

cypress-tests/cypress/support/commands.js
```diff
@@ -2339,6 +2368,33 @@ Cypress.Commands.add(
...
+            // Validate the payment_method_id format
+            expect(
+              response.body.payment_method_id,
+              "payment_method_id"
+            ).to.include("pm_").and.to.not.be.null;
+
+            // Determine expected status based on payment status
+            const expectedStatus = [
+              "succeeded",
+              "processing",
```
likhinbopanna, line 2384: For `processing` status ,  payment_method_status would be `inactive` , can you make the change here ?

cypress-tests/cypress/support/commands.js
```diff
@@ -2155,13 +2155,41 @@ Cypress.Commands.add(
...
+              }
             } else if (response.body.authentication_type === "no_three_ds") {
               for (const key in resData.body) {
                 expect(resData.body[key]).to.equal(response.body[key]);
               }
               expect(response.body.customer_id).to.equal(
                 globalState.get("customerId")
               );
+              if (
+                ["succeeded", "requires_capture"].includes(response.body.status)
```
likhinbopanna, line 2176: Can you include partially_captured status as well ?

cypress-tests/cypress/support/commands.js
pixincreate: if the `body` does not have the field `payment_method_id`, then we can directly use `response.body?.payment_method_id` where the field itself is missing. since this is a confirm, we will indeed have the field with data in it being `null`! So yeah, this check can be tweaked a bit.

cypress-tests/cypress/support/commands.js
pixincreate: we're getting status as `active`. at least for cybersource. i did confirm with @Gnanasundari24 the other day. only after that, this change is incorporated.

cypress-tests/cypress/support/commands.js
pixincreate: we explicitly want the data in it to be a string and not otherwise (`null`)

cypress-tests/cypress/support/commands.js
likhinbopanna: I don’t think this is correct. For processing, the payment method status should be `inactive`. By including the processing status here, we are adding invalid test case.
By the way, why does `Cybersource` have an `active` status ,that's because You get a processing status only after a manual capture payment, where the payment is already captured. Hence, it is `active` for the `requires_capture` status.

cypress-tests/cypress/support/commands.js
likhinbopanna: You still didn't add partially capture

pixincreate: [`d3935fe` (#7496)](https://github.com/juspay/hyperswitch/pull/7496/commits/d3935fe288c04c016fcc3e7d5de59afbc5699477)

was added in this commit

likhinbopanna (approved):

Gnanasundari24 (approved):

PR #6613, opened 2024-11-19
feat(router): add outgoing payment webhooks for v2

sai-harsha-vardhan:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

This PR adds outgoing payment webhooks for v2. This PR improves the webhook delivery flow by introducing traits to enhance extensibility and modularity. Additionally, it optimizes the raise_webhooks_analytics_event process by removing an unnecessary database fetch.

### Additional Changes

- [ ] This PR modifies the API contract
- [ ] This PR modifies the database schema
- [ ] This PR modifies application configuration/environment variables

## How did you test it?

Tested Manually
Setup incoming webhooks and trigger a successful payment, to receive an outgoing webhook

 - The incoming webhook to be setup to {base_url}/v2/webhooks/(merchant_id}/{profile_id}/{mca_id}
 - Create Payment 
```
curl --location 'http://localhost:8080/v2/payments/create-intent' \
--header 'api-key: dev_gkzPh8C1IaaBlf1Vo31QNoKUkhRuKUDYfm6zoe1LLiZL3rpOyDppUVI2nrLT9KIP' \
--header 'Content-Type: application/json' \
--header 'x-profile-id: pro_7MjgLAahgfPro5dOWlHk' \
--header 'Authorization: api-key=dev_gkzPh8C1IaaBlf1Vo31QNoKUkhRuKUDYfm6zoe1LLiZL3rpOyDppUVI2nrLT9KIP' \
--data-raw '{
    "amount_details": {
        "order_amount": 100,
        "currency": "USD"
    },
    
    "capture_method":"automatic",
    "authentication_type": "no_three_ds",
    "billing": {
        "address": {
            "first_name": "John",
            "last_name": "Dough"
        },
        "email": "example@example.com"
    },
    "shipping": {
        "address": {
            "first_name": "John",
            "last_name": "Dough",
            "city": "Karwar",
            "zip": "581301",
            "state": "Karnataka"
        },
        "email": "example@example.com"
    }
}'
```
- Confirm Payment
```
curl --location 'http://localhost:8080/v2/payments/12345_pay_0196803f8a8e7840810bf8dcc6b0fa74/confirm-intent' \
--header 'x-client-secret: cs_0196803f8aa376c282d881ceb57c6aac' \
[...]

PR #7578, opened 2025-03-19
feat(core): Support for multiple outgoing webhooks added

Sweta-Kumari-Sharma:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

### Additional Changes

## Checklist


crates/api_models/src/admin.rs
```diff
@@ -52,7 +52,7 @@ pub struct MerchantAccountCreate {
     pub return_url: Option<url::Url>,
 
     /// Webhook related details
-    pub webhook_details: Option<WebhookDetails>,
+    pub webhook_details: Option<Vec<Option<WebhookDetails>>>,
```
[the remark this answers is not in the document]
Sweta-Kumari-Sharma: After the migration, the `schema.rs` file generates the `webhook_details` field as `Nullable<Array<Nullable<Json>>>`.
The reason for having both options  `Option<Vec<Option<WebhookDetails>>>` is to align with the schema generated after the migration.

crates/api_models/src/admin.rs
```diff
@@ -639,31 +639,19 @@ pub struct PrimaryBusinessDetails {
...
-    pub webhook_version: Option<String>,
-
-    ///The user name for Webhook login
-    #[schema(max_length = 255, max_length = 255, example = "ekart_retail")]
-    pub webhook_username: Option<String>,
-
-    ///The password for Webhook login
-    #[schema(value_type = Option<String>, max_length = 255, example = "ekart@123")]
-    pub webhook_password: Option<Secret<String>>,
+    pub webhook_endpoint_id: Option<id_type::WebhookEndpointId>,
```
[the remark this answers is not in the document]
Sweta-Kumari-Sharma: Yes, the change is backwards compatible. Since `webhook_endpoint_id` is `nullable`, for the previous `webhook_details`, its value will be `None`.

crates/api_models/src/admin.rs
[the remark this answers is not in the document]
jarnura: This is API related change in api-model, we shouldn't modify this for database related things, and changing a type of existing argument violate backward compatibility.

crates/diesel_models/src/merchant_account.rs
```diff
@@ -30,7 +30,7 @@ pub struct MerchantAccount {
     pub redirect_to_merchant_with_http_post: bool,
     pub merchant_name: Option<Encryption>,
     pub merchant_details: Option<Encryption>,
-    pub webhook_details: Option<crate::business_profile::WebhookDetails>,
+    pub webhook_details: Option<Vec<Option<crate::business_profile::WebhookDetails>>>,
```
jarnura, line 33:
```suggestion
    pub webhook_details: Option<Vec<Option<business_profile::WebhookDetails>>>,
```

crates/diesel_models/src/merchant_account.rs
```diff
@@ -65,7 +65,7 @@ pub struct MerchantAccountSetter {
...
```
jarnura, line 68:
```suggestion
    pub webhook_details: Option<Vec<Option<business_profile::WebhookDetails>>>,
```

crates/diesel_models/src/merchant_account.rs
```diff
@@ -219,7 +219,7 @@ pub struct MerchantAccountNew {
     pub merchant_name: Option<Encryption>,
     pub merchant_details: Option<Encryption>,
     pub return_url: Option<String>,
-    pub webhook_details: Option<crate::business_profile::WebhookDetails>,
+    pub webhook_details: Option<Vec<Option<crate::business_profile::WebhookDetails>>>,
```
jarnura, line 222:
```suggestion
    pub webhook_details: Option<Vec<Option<business_profile::WebhookDetails>>>,
```

crates/diesel_models/src/merchant_account.rs
```diff
@@ -322,7 +322,7 @@ pub struct MerchantAccountUpdateInternal {
...
```
jarnura, line 325:
```suggestion
    pub webhook_details: Option<Vec<Option<business_profile::WebhookDetails>>>,
```

jarnura (changes requested):

crates/router/src/core/webhooks/outgoing.rs
```diff
@@ -47,70 +52,102 @@ use crate::{
 
 const OUTGOING_WEBHOOK_TIMEOUT_SECS: u64 = 5;
 
+pub(crate) async fn get_webhook_details_for_event_type(
```
SanchithHegde, line 55: Nit: Does this function need to be `async`?

crates/router/src/core/webhooks/outgoing.rs
```diff
@@ -579,6 +680,76 @@ async fn raise_webhooks_analytics_event(
     state.event_handler().log_event(&webhook_event);
 }
 
+pub(crate) async fn add_bulk_outgoing_webhook_task_to_process_tracker(
```
SanchithHegde, line 683: Nit: Please add `#[instrument(skip_all)]` on this function.

crates/router/src/core/webhooks/incoming.rs
```diff
@@ -725,18 +721,20 @@ async fn payments_incoming_webhook_flow(
...
-                    business_profile,
+                    &business_profile,
+                    payment_id.get_string_repr(),
                     outgoing_event_type,
-                    enums::EventClass::Payments,
-                    payment_id.get_string_repr().to_owned(),
-                    enums::EventObjectType::PaymentDetails,
-                    api::OutgoingWebhookContent::PaymentDetails(Box::new(payments_response)),
+                    diesel_models::enums::EventClass::Payments,
+                    diesel_models::enums::EventObjectType::PaymentDetails,
```
SanchithHegde, line 731: Nit: Can just qualify them as `enums::...`?


```suggestion
                    enums::EventClass::Payments,
                    enums::EventObjectType::PaymentDetails,
```

crates/router/src/core/webhooks/incoming.rs
```diff
@@ -725,18 +721,20 @@ async fn payments_incoming_webhook_flow(
...
                     primary_object_created_at,
                 ))
-                .await?;
+                .await
+                .map_err(|e| {
+                    e.change_context(errors::ApiErrorResponse::WebhookProcessingFailure)
+                })?;
```
SanchithHegde, line 737: Nit: Unnecessary `map_err()`.


```suggestion
                .change_context(errors::ApiErrorResponse::WebhookProcessingFailure)?;
```

crates/router/src/core/webhooks/incoming.rs
```diff
@@ -844,28 +842,17 @@ async fn payouts_incoming_webhook_flow(
...
-                enums::EventObjectType::PayoutDetails,
-                api::OutgoingWebhookContent::PayoutDetails(Box::new(payout_create_response)),
-                Some(updated_payout_attempt.created_at),
+                diesel_models::enums::EventClass::Payouts,
+                diesel_models::enums::EventObjectType::PayoutDetails,
+                Some(payout_attempt.created_at),
             ))
-            .await?;
+            .await
+            .map_err(|e| e.change_context(errors::ApiErrorResponse::WebhookProcessingFailure))?;
```
SanchithHegde, line 855: Same comments as before, about the `enums::` qualification and the unnecessary `map_err()`.

crates/router/src/core/webhooks/incoming.rs
```diff
@@ -1052,20 +1039,17 @@ async fn refunds_incoming_webhook_flow(
...
-            enums::EventObjectType::RefundDetails,
-            api::OutgoingWebhookContent::RefundDetails(Box::new(refund_response)),
-            Some(updated_refund.created_at),
+            diesel_models::enums::EventClass::Refunds,
+            diesel_models::enums::EventObjectType::RefundDetails,
+            Some(refund.created_at),
         ))
-        .await?;
+        .await
+        .map_err(|e| e.change_context(errors::ApiErrorResponse::WebhookProcessingFailure))?;
```
SanchithHegde, line 1052: Same comments as before, about the `enums::` qualification and the unnecessary `map_err()`.

crates/router/src/core/webhooks/incoming.rs
```diff
@@ -1456,34 +1439,27 @@ async fn mandates_incoming_webhook_flow(
...
-                enums::EventObjectType::MandateDetails,
-                api::OutgoingWebhookContent::MandateDetails(mandates_response),
-                Some(updated_mandate.created_at),
+                diesel_models::enums::EventClass::Mandates,
+                diesel_models::enums::EventObjectType::MandateDetails,
+                primary_object_created_at,
             ))
-            .await?;
+            .await
+            .map_err(|e| e.change_context(errors::ApiErrorResponse::WebhookProcessingFailure))?;
```
SanchithHegde, line 1462: Same comments as before, about the `enums::` qualification and the unnecessary `map_err()`.

crates/router/src/core/webhooks/incoming.rs
```diff
@@ -1646,21 +1623,20 @@ async fn disputes_incoming_webhook_flow(
...
-            dispute_object.dispute_id.clone(),
-            enums::EventObjectType::DisputeDetails,
-            api::OutgoingWebhookContent::DisputeDetails(disputes_response),
+            diesel_models::enums::EventClass::Disputes,
+            diesel_models::enums::EventObjectType::DisputeDetails,
             Some(dispute_object.created_at),
         ))
-        .await?;
+        .await
+        .map_err(|e| e.change_context(errors::ApiErrorResponse::WebhookProcessingFailure))?;
```
SanchithHegde, line 1638: Same comments as before, about the `enums::` qualification and the unnecessary `map_err()`.

crates/router/src/core/webhooks/incoming.rs
```diff
@@ -1646,21 +1623,20 @@ async fn disputes_incoming_webhook_flow(
...
-        let event_type: enums::EventType = dispute_object.dispute_status.foreign_into();
 
-        Box::pin(super::create_event_and_trigger_outgoing_webhook(
+        let event_type: enums::EventType = dispute_object.dispute_status.foreign_into();
+        Box::pin(super::add_bulk_outgoing_webhook_task_to_process_tracker(
             state,
-            merchant_context,
-            business_profile,
+            &business_profile,
+            &dispute_object.dispute_id.clone(),
```
SanchithHegde, line 1631: Why do we need the clone if we're anyway passing a reference to the dispute ID?

crates/router/src/utils.rs
```diff
@@ -1143,7 +1143,7 @@ where
 #[cfg(feature = "v1")]
 #[allow(clippy::too_many_arguments)]
 pub async fn trigger_payments_webhook<F, Op, D>(
-    merchant_context: domain::MerchantContext,
+    _merchant_context: domain::MerchantContext,
```
SanchithHegde, line 1146: Why is this variable / parameter no longer being read?

crates/router/src/utils.rs
```diff
@@ -1272,26 +1292,45 @@ pub async fn trigger_refund_outgoing_webhook(
...
+                            diesel_models::enums::EventClass::Refunds,
+                            diesel_models::enums::EventObjectType::RefundDetails,
+                            primary_object_created_at,
+                        ),
+                    )
                     .await
                 }
                 .in_current_span(),
             );
+        // if let Some(outgoing_event_type) = event_type {
```
SanchithHegde, line 1314: Can you remove the commented code?

SanchithHegde (commented):
Also, please include more test cases in the PR description:

- Ensuring that webhooks are correctly being sent to all webhook endpoints in case multiple endpoints are configured in the business profile.
- Ensuring that automatic and manual retries works for all webhook endpoints.
- Ensuring that no change in behavior is observed for old business profiles with old data: webhook delivery of initial attempts, automatic retries and manual retries must happen as they used to previously.
- Ensuring that analytics events are being correctly populated in the analytics pipeline, for all webhook endpoints. (Please set up Kafka and ClickHouse locally using Docker Compose, test it out and include screenshots.)

crates/router/src/utils.rs
Sweta-Kumari-Sharma: We were calling fn `create_event_and_trigger_outgoing_webhook` previously in `trigger_payments_webhook` fn, which required `merchant_context` to be passed as a parameter. But now we are calling `add_bulk_outgoing_webhook_task_to_process_tracker` from `trigger_payments_webhook`, which is why we don't need this parameter anymore.

crates/diesel_models/src/business_profile.rs
```diff
@@ -702,6 +702,16 @@ pub struct WebhookDetails {
     pub payment_created_enabled: Option<bool>,
     pub payment_succeeded_enabled: Option<bool>,
     pub payment_failed_enabled: Option<bool>,
+    pub multiple_webhooks_list: Option<Vec<MultipleWebhookDetail>>,
+}
+
+#[derive(Clone, Debug, serde::Deserialize, serde::Serialize, diesel::AsExpression)]
+#[diesel(sql_type = diesel::sql_types::Json)]
+pub struct MultipleWebhookDetail {
```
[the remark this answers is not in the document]
Sweta-Kumari-Sharma: Although `created_at` and `updated_at` are not required for the current logic, but  we can store them to support potential future use cases and for reference.

crates/api_models/src/admin.rs
```diff
@@ -688,6 +688,18 @@ pub struct WebhookDetails {
...
+}
+
+#[derive(Clone, Debug, Deserialize, ToSchema, Serialize)]
+#[serde(deny_unknown_fields)]
+pub struct MultipleWebhookDetail {
+    pub webhook_endpoint_id: Option<id_type::WebhookEndpointId>,
+    pub webhook_url: Option<Secret<String>>,
+    pub events: Vec<common_enums::EventType>,
+    pub status: Option<common_enums::OutgoingWebhookEndpointStatus>,
 }
```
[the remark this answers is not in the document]
swetasharma03: `webhook_endpoint_id` is optional because it will be generated internally when we create the `webhook_detail`, so it will be `None` at creation time.
`status` is also optional, as we treat newly created webhook endpoints as `active` by default, so we don't expect this value to be explicitly provided.
However, `webhook_url` is a required field and has been made mandatory.
Also, I have made the `events` field a `HashSet` instead of a `Vec`

crates/diesel_models/src/business_profile.rs
```diff
@@ -702,6 +702,16 @@ pub struct WebhookDetails {
...
+    pub webhook_endpoint_id: Option<common_utils::id_type::WebhookEndpointId>,
+    pub webhook_url: Option<Secret<String>>,
+    pub events: Vec<common_enums::EventType>,
+    pub status: Option<common_enums::OutgoingWebhookEndpointStatus>,
```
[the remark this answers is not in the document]
swetasharma03: For the Diesel model, I’ve made the fields mandatory:
`webhook_endpoint_id` is generated internally and always present once created, so it’s non-optional there.
`webhook_url` is a required field and is marked mandatory.
`status` defaults to active when creating a new webhook endpoint, so it’s also non-optional in the model.

Also, I’ve updated the `events` field to use a `HashSet` instead of a `Vec`.

crates/router/src/types/transformers.rs
```diff
@@ -2130,6 +2131,39 @@ impl ForeignFrom<api_models::admin::WebhookDetails>
...
+
+        if let Some(list) = item.multiple_webhooks_list {
+            normalized_list.extend(
+                list.into_iter()
+                    .map(ForeignFrom::foreign_from)
+                    .collect::<Vec<_>>(),
+            );
+        }
+
         Self {
```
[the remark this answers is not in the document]
swetasharma03: Yes, this was done to maintain backward compatibility with `api_models`, which doesn’t include `multiple_webhook_list` in older version. So, we convert the available `webhook_url` into a `multiple_webhook_list`.

crates/router/src/types/transformers.rs
```diff
@@ -2138,6 +2172,26 @@ impl ForeignFrom<api_models::admin::WebhookDetails>
...
+            webhook_endpoint_id: Some(
+                item.webhook_endpoint_id.unwrap_or_else(|| {
+                    common_utils::generate_webhook_endpoint_id_of_default_length()
+                }),
+            ),
+            webhook_url: item.webhook_url,
+            events: item.events,
+            status: item
+                .status
+                .or(Some(common_enums::OutgoingWebhookEndpointStatus::Active)),
```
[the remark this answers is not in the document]
Sweta-Kumari-Sharma: We use this transformer when creating the `webhook_detail`. I have removed `item.webhook_endpoint_id ` because the `webhook_endpoint_id` is generated internally and is not expected to be provided externally. Additionally, the `status` field defaults to `active`.

When updating the `business_profile`, we do not create a new webhook endpoint—that is handled separately in the update_business_profile endpoint. During the update, we expect the entire `multiple_webhook_list` to be provided, not just the single webhook object being updated. This ensures the merchant’s existing webhook data is preserved and not accidentally lost.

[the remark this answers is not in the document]
Sweta-Kumari-Sharma: We use this transformer when creating the `webhook_detail`. I have removed `item.webhook_endpoint_id ` because the `webhook_endpoint_id` is generated internally and is not expected to be provided externally. Additionally, the `status` field defaults to `active`.

When updating the `business_profile`, we do not create a new webhook endpoint—that is handled in the update_business_profile endpoint. During the update, we expect the entire `multiple_webhook_list` to be provided, not just the single webhook object being updated. This ensures the merchant’s existing webhook data is preserved and not accidentally lost.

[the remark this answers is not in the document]
swetasharma03: We use this transformer when creating the `webhook_detail`. I have removed `item.webhook_endpoint_id ` because the `webhook_endpoint_id` is generated internally and is not expected to be provided externally. Additionally, the `status` field defaults to `active`.

When updating the `business_profile`, we have not created a new webhook update endpoint—that is handled in the update_business_profile endpoint. During the update, we expect the entire `multiple_webhook_list` to be provided, not just the single webhook object being updated. This ensures the merchant’s existing webhook data is preserved and not accidentally lost.

crates/router/src/core/admin.rs
```diff
@@ -1116,6 +1131,21 @@ pub async fn merchant_account_update(
         .await
         .to_not_found_response(errors::ApiErrorResponse::MerchantAccountNotFound)?;
 
+    #[cfg(feature = "v1")]
+    if let Some(ref webhook_details) = &req.webhook_details {
```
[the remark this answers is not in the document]
swetasharma03: When updating the account, we have not created a new webhook update endpoint—that is handled in the account update endpoint. During the update, we expect the entire multiple_webhook_list to be provided, not just the single webhook object being updated. This ensures the merchant’s existing webhook data is preserved and not accidentally lost.

crates/router/src/core/admin.rs
```diff
@@ -4509,6 +4553,19 @@ pub async fn update_profile(
 ) -> RouterResponse<api::ProfileResponse> {
     let db = state.store.as_ref();
     let key_manager_state = &(&state).into();
+    if let Some(ref webhook_details) = &request.webhook_details {
```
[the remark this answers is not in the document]
swetasharma03: When updating the profile, we have not created a new webhook update endpoint—that is handled in the profile update endpoint. During the update, we expect the entire multiple_webhook_list to be provided, not just the single webhook object being updated. This ensures the merchant’s existing webhook data is preserved and not accidentally lost.

crates/router/src/core/admin.rs
```diff
@@ -4171,11 +4215,11 @@ pub async fn retrieve_profile(
...
-    Ok(service_api::ApplicationResponse::Json(
-        api_models::admin::ProfileResponse::foreign_try_from(business_profile)
-            .change_context(errors::ApiErrorResponse::InternalServerError)
-            .attach_printable("Failed to parse business profile details")?,
-    ))
+    let profile_response = api_models::admin::ProfileResponse::foreign_try_from(business_profile)
+        .change_context(errors::ApiErrorResponse::InternalServerError)
+        .attach_printable("Failed to parse business profile details")?;
+
+    Ok(service_api::ApplicationResponse::Json(profile_response))
```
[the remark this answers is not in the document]
swetasharma03: You're right, that change was made by mistake. I've corrected it and removed the unintended modification.

crates/common_utils/src/id_type/webhook_endpoint.rs
```diff
@@ -0,0 +1,24 @@
+use crate::errors::{CustomResult, ValidationError};
+
+crate::id_type!(
+    WebhookEndpointId,
+    "A type for webhook_endpoint_id that can be used for unique identifier for a webhook_endpoint"
+);
+crate::impl_id_type_methods!(WebhookEndpointId, "webhook_endpoint_id");
+crate::impl_generate_id_id_type!(WebhookEndpointId, "web");
+crate::impl_default_id_type!(WebhookEndpointId, "web");
```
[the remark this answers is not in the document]
swetasharma03: We need this `Default` implementation to handle cases where `webhook_endpoint_id` is missing in older data. It allows us to convert the available data into a `multiple_webhook_list`, particularly in functions like `get_webhook_details_for_event_type`.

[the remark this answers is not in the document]
SanchithHegde: Can't we make use of the `generate_webhook_endpoint_id_of_default_length()` function instead of using the `Default` trait? 

The reason I don't prefer implementing the `Default` trait is because the implementation generates a random ID by default, and is not a static value. While `Default` trait implementations typically involve static values.

crates/diesel_models/src/business_profile.rs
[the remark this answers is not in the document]
SanchithHegde: Could you please add the two fields then? Won't hurt to have this information I feel.

crates/api_models/src/admin.rs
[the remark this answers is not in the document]
SanchithHegde: If we don't need them to provide specific fields in the request, can we consider using two different types, one for request and one for response (say `MultipleWebhookDetailRequest` and `MultipleWebhookDetailResponse` or something similar)?

The response type could have fields similar to the diesel model type (because all of the data in the response would be populated from the diesel model anyway), while the request need not have fields like `webhook_endpoint_id` at all, while `status` may be optionally provided.

crates/router/src/core/webhooks/utils.rs
```diff
@@ -137,21 +137,41 @@ pub async fn construct_webhook_router_data(
...
+    const MAX_PREFIX_LEN: usize = 64 - EVENT_ID_SUFFIX_LENGTH;
+
+    let input = match webhook_endpoint_id {
+        Some(id) => format!("{primary_object_id}_{event_type}_{id}"),
+        None => format!("{primary_object_id}_{event_type}"),
+    };
+
+    // Hash and encode the input using SHA-256 and URL-safe base64 (without padding)
+    let hash = Sha256::digest(input.as_bytes());
+    let encoded = URL_SAFE_NO_PAD.encode(hash);
```
SanchithHegde, line 162: 1. Please use `Sha256` from the `common_utils` crate instead.
2. And for base64 encoding, add a const in `common_utils` crate called `BASE64_ENGINE_URL_SAFE_NO_PAD`:
   https://github.com/juspay/hyperswitch/blob/dee5d0c552b382ba7c2db9cf672cc050c17ee3b9/crates/common_utils/src/consts.rs#L94-L99
3. Also, please add a comment explaining why we're doing this hashing + encoding (and why we're not using the concatenated value directly).

crates/router/src/core/webhooks/utils.rs
```diff
@@ -137,21 +137,41 @@ pub async fn construct_webhook_router_data(
...
+
+    // Truncate to MAX_PREFIX_LEN (56) if needed
+    let common_prefix = if encoded.len() > MAX_PREFIX_LEN {
+        &encoded[..MAX_PREFIX_LEN]
+    } else {
+        &encoded
+    };
```
SanchithHegde, line 169: Truncating most likely won't be needed, considering SHA256 output would be 256 bits (or 32 bytes), and base64 encoding adds ~33% overhead, so encoded output should be of size ~43 characters (32 * 1.33), well within our limit of 56 characters.

crates/router/src/core/webhooks/utils.rs
```diff
@@ -137,21 +137,41 @@ pub async fn construct_webhook_router_data(
...
     const EVENT_ID_SUFFIX_LENGTH: usize = 8;
+    const MAX_PREFIX_LEN: usize = 64 - EVENT_ID_SUFFIX_LENGTH;
+
+    let input = match webhook_endpoint_id {
+        Some(id) => format!("{primary_object_id}_{event_type}_{id}"),
+        None => format!("{primary_object_id}_{event_type}"),
+    };
+
+    // Hash and encode the input using SHA-256 and URL-safe base64 (without padding)
+    let hash = Sha256::digest(input.as_bytes());
```
SanchithHegde, line 161: Also, because we're now hashing the concatenated value, would this break any of the flows (initial attempt, automatic retry and manual retry) for old business profiles having the old data format?
Can we please test this out?

crates/router/src/types/transformers.rs
```diff
@@ -2130,6 +2131,54 @@ impl ForeignFrom<api_models::admin::WebhookDetails>
...
+                common_enums::EventType::MandateActive,
+                common_enums::EventType::MandateRevoked,
+                common_enums::EventType::PayoutSuccess,
+                common_enums::EventType::PayoutFailed,
+                common_enums::EventType::PayoutInitiated,
+                common_enums::EventType::PayoutProcessing,
+                common_enums::EventType::PayoutCancelled,
+                common_enums::EventType::PayoutExpired,
+                common_enums::EventType::PayoutReversed,
+            ]);
```
SanchithHegde, line 2163: If you're going to hardcode this list today, we won't know when new variants are added to `EventType`, as the compiler won't throw errors about it.

We can instead derive [`strum::VariantArray`](https://docs.rs/strum/latest/strum/derive.VariantArray.html) on `EventType`, and you should be able to then do something like:

```suggestion
            let events: std::collections::HashSet<_> = std::collections::HashSet::from_iter(common_enums::EventType::VARIANTS);
```

This should provide us with all variants available for that enum.

And you can import `HashSet` it should be fine.

crates/router/src/types/transformers.rs
```diff
@@ -2130,6 +2131,54 @@ impl ForeignFrom<api_models::admin::WebhookDetails>
...
+            normalized_list.push(legacy_entry);
+        }
+
+        if let Some(list) = item.multiple_webhooks_list {
+            normalized_list.extend(
+                list.into_iter()
+                    .map(ForeignFrom::foreign_from)
+                    .collect::<Vec<_>>(),
+            );
+        }
```
SanchithHegde, line 2180: Since we're adding both the legacy entry and those present in the `multiple_webhooks_list` field to the list, how are we handling the case where the same URL is present in both the legacy entry and as an item in the `multiple_webhooks_list` field? What would be the behavior in that case?

Moreover, do you think we should allow sending both the `webhook_url` and the `multiple_webhooks_list` fields in the request?

crates/router/src/types/transformers.rs
```diff
@@ -2138,6 +2187,24 @@ impl ForeignFrom<api_models::admin::WebhookDetails>
...
+    fn foreign_from(item: api_models::admin::MultipleWebhookDetail) -> Self {
+        Self {
+            webhook_endpoint_id: item
+                .webhook_endpoint_id
+                .unwrap_or_else(|| common_utils::generate_webhook_endpoint_id_of_default_length()),
+
+            webhook_url: item.webhook_url,
+            events: item.events,
+
+            status: common_enums::OutgoingWebhookEndpointStatus::Active,
```
SanchithHegde, line 2207: Is there any specific reason for hardcoding the status to be `Active`?

Would this transformation be used when we're updating the webhook endpoints as well? If yes, how is the merchant supposed to disable a specific webhook endpoint?

crates/router/src/core/webhooks/types.rs
```diff
@@ -78,6 +78,8 @@ pub(crate) struct OutgoingWebhookTrackingData {
     pub(crate) primary_object_id: String,
     pub(crate) primary_object_type: enums::EventObjectType,
     pub(crate) initial_attempt_id: Option<String>,
+    pub(crate) webhook_endpoint_id: Option<common_utils::id_type::WebhookEndpointId>,
+    pub(crate) primary_object_created_at: Option<time::PrimitiveDateTime>,
```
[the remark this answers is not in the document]
SanchithHegde: Could you please point me to the code snippet where this field is being used to construct idempotent ID? (And I'm guessing you meant `idempotent_event_id`?)

crates/router/src/core/webhooks/utils.rs
```diff
@@ -137,21 +137,41 @@ pub async fn construct_webhook_router_data(
...
 
+use base64::{engine::general_purpose::URL_SAFE_NO_PAD, Engine as _};
+use sha2::{Digest, Sha256};
+
 #[inline]
 pub(crate) fn get_idempotent_event_id(
     primary_object_id: &str,
     event_type: types::storage::enums::EventType,
     delivery_attempt: types::storage::enums::WebhookDeliveryAttempt,
+    webhook_endpoint_id: Option<&str>,
```
SanchithHegde, line 148: Let us accept `Option<&WebhookEndpointId>` here itself instead of `Option<&str>`.

crates/router/src/core/webhooks/outgoing.rs
```diff
@@ -46,6 +47,83 @@ use crate::{
...
+        .change_context(errors::WebhooksFlowError::MerchantWebhookDetailsNotFound)?;
+
+    match webhook_endpoint_id {
+        // Legacy case: no endpoint ID, fall back to single `webhook_url`
+        None => {
+            if let Some(webhook_url) = webhook_details.webhook_url.clone() {
+                Ok(MultipleWebhookDetail {
+                    webhook_endpoint_id: common_utils::id_type::WebhookEndpointId::default(),
+                    webhook_url: webhook_url,
+                    events: HashSet::new(), // legacy data might not have events specified
```
SanchithHegde, line 105: Should we consider calling `HashSet::from_iter(enums::EventType::VARIANTS)` here as well, as suggested in one of my previous comments? This would be consistent with the API model to diesel model conversion we're doing in `crates/router/src/types/transformers.rs`.

crates/router/src/core/webhooks/outgoing.rs
```diff
@@ -46,6 +47,83 @@ use crate::{
...
+            } else {
+                Err(errors::WebhooksFlowError::MerchantWebhookDetailsNotFound.into())
+            }
+        }
+
+        // Modern case: look up by ID from the multiple list
+        Some(endpoint_id) => {
+            let webhooks = webhook_details
+                .multiple_webhooks_list
+                .clone()
```
SanchithHegde, line 117: Nit: Do we necessarily need the clone?

crates/router/src/core/admin.rs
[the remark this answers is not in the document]
SanchithHegde: What would happen in such a scenario?

Let's say both the app server and the control center are deployed with support for the new API contract, so new merchant accounts / profiles that are created now have details stored in the `multiple_webhooks_list` field. In the very rare case that the control center is rolled back to a previous version without support for the `multiple_webhooks_list` field (but the app server stays on the newer version), when updating the merchant account / profile, what would happen?

If I understand correctly, the app server would notice that the `multiple_webhooks_list` field is not provided in the request, so it assumes it to be `None` by default. So, in the eyes of the app server, the `webhook_details` field in the request could look like:

```rust
webhook_details: WebHookDetails {
    webhook_url: Some("some-url"),
    // Other fields ...
    multiple_webhooks_list: None,
}
```

And because we're updating the entire `webhook_details` field during the merchant account / profile update operation, we would end up overwriting the previously existing list of webhook endpoints and lose all previous data, isn't it?

How are we planning to handle such situations?

crates/router/src/core/webhooks/outgoing.rs
```diff
@@ -46,6 +47,83 @@ use crate::{
...
+    let webhook_list =
+        if let Some(multiple_webhooks) = webhook_details.multiple_webhooks_list.clone() {
+            multiple_webhooks
+        } else if let Some(webhook_url) = webhook_details.webhook_url.clone() {
+            vec![MultipleWebhookDetail {
+                webhook_endpoint_id: common_utils::id_type::WebhookEndpointId::default(),
+                webhook_url: webhook_url,
+                events: HashSet::from([event_type]), // legacy data might not have events specified
+                status: common_enums::OutgoingWebhookEndpointStatus::Active,
+            }]
```
SanchithHegde, line 72: Now that I think about it, I feel we can write a function to construct the `MultipleWebhookDetail` for the "legacy" case where we only have the `webhook_url`. Something along the lines of:

```rust
#[inline]
pub(crate) fn get_multiple_webhook_detail_for_legacy_url(webhook_url: Secret<String>) -> MultipleWebhookDetail {
    MultipleWebhookDetail {
        webhook_endpoint_id: generate_webhook_endpoint_id_of_default_length(),
        webhook_url,
        events: HashSet::from_iter(enums::EventType::VARIANTS),
        status: enums::OutgoingWebhookEndpointStatus::Active,
    }
}
```

And this can be called everywhere we're handling the "legacy" case.

crates/router/src/core/webhooks/outgoing.rs
```diff
@@ -46,6 +47,83 @@ use crate::{
...
+        } else {
+            return Ok(vec![]);
+        };
+
+    let matching_details: Vec<MultipleWebhookDetail> = webhook_list
+        .into_iter()
+        .filter(|detail| {
+            detail.events.contains(&event_type)
+                && detail.status == common_enums::OutgoingWebhookEndpointStatus::Active
```
SanchithHegde, line 81: Nit:

We can implement methods on `MultipleWebhookDetail` called `is_endpoint_active()` and `is_event_enabled()` which can provide us this information:

```rust
impl MultipleWebhookDetail {
    pub fn is_endpoint_active(&self) -> bool {
        self.status == enums::OutgoingWebhookEndpointStatus::Active
    }

    pub fn is_event_enabled(&self, event_type: EventType) -> bool {
        self.events.contains(&event_type)
    }
}
```

Which could then simplify this code to something like:
```suggestion
            detail.is_event_supported(event_type) && detail.is_endpoint_active()
```

crates/router/src/core/webhooks/outgoing.rs
```diff
@@ -46,6 +47,83 @@ use crate::{
...
+    match webhook_endpoint_id {
+        // Legacy case: no endpoint ID, fall back to single `webhook_url`
+        None => {
+            if let Some(webhook_url) = webhook_details.webhook_url.clone() {
+                Ok(MultipleWebhookDetail {
+                    webhook_endpoint_id: common_utils::id_type::WebhookEndpointId::default(),
+                    webhook_url: webhook_url,
+                    events: HashSet::new(), // legacy data might not have events specified
+                    status: common_enums::OutgoingWebhookEndpointStatus::Active,
+                })
```
SanchithHegde, line 107: We can call the `get_multiple_webhook_detail_for_legacy_url()` function here.

crates/router/src/core/webhooks/outgoing.rs
```diff
@@ -58,59 +136,32 @@ pub(crate) async fn create_event_and_trigger_outgoing_webhook(
...
-            "Outgoing webhooks are disabled in application configuration, or merchant webhook URL \
-             could not be obtained; skipping outgoing webhooks for event"
-        );
-        return Ok(());
-    }
 
-    let event_id = utils::generate_event_id();
-    let merchant_id = business_profile.merchant_id.clone();
-    let now = common_utils::date_time::now();
+    if webhook_detail.events.contains(&event_type) {
```
SanchithHegde, line 144: Nit: We can call `webhook_detail.is_event_enabled(event_type)` once we've implemented it.

crates/router/src/core/webhooks/outgoing.rs
```diff
@@ -137,110 +188,148 @@ pub(crate) async fn create_event_and_trigger_outgoing_webhook(
...
 
-    Ok(())
+        Ok(())
+    } else {
+        logger::debug!(
+            business_profile_id=?business_profile.get_id(),
+            webhook_endpoint_id=?webhook_detail.webhook_endpoint_id,
+            events=?webhook_detail.events,
+            "Webhook detail doesn't match the event type; skipping outgoing webhook"
+        );
```
SanchithHegde, line 330: Should we consider logging the current event type as well? So that we can at least know _which_ event type is being skipped.

crates/router/src/core/webhooks/outgoing.rs
```diff
@@ -285,6 +376,13 @@ pub(crate) async fn trigger_webhook_and_raise_event(
     .await;
 }
 
+fn get_webhook_url_from_webhook_detail(
+    webhook_detail: MultipleWebhookDetail,
+) -> CustomResult<String, errors::WebhooksFlowError> {
+    Ok(webhook_detail.webhook_url.peek().clone())
+}
```
SanchithHegde, line 383: Nit: This can be a method implemented on `MultipleWebhookDetail`, and need not return a `Result`.

crates/router/src/core/webhooks/outgoing.rs
```diff
@@ -579,6 +677,68 @@ async fn raise_webhooks_analytics_event(
...
+        runner,
+        tag,
+        tracking_data,
+        None,
+        schedule_time,
+        common_types::consts::API_VERSION,
+    )
+    .map_err(errors::StorageError::from)?;
+
+    let attributes = router_env::metric_attributes!(("flow", "OutgoingWebhookRetry"));
```
SanchithHegde, line 729: Nit: Do we wanna maybe indicate that this is a bulk retry task?


```suggestion
    let attributes = router_env::metric_attributes!(("flow", "OutgoingWebhookRetryBulk"));
```

crates/router/src/workflows/outgoing_webhook_retry.rs
```diff
@@ -34,202 +34,370 @@ use crate::{
...
+            .await
+        }
+        Some("OUTGOING_WEBHOOK_RETRY") => {
+            Box::pin(trigger_outgoing_webhook_retry_workflow(
+                state,
+                refund_tracker,
+            ))
+            .await
+        }
+        _ => Err(errors::ProcessTrackerError::JobNotFound),
```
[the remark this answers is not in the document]
SanchithHegde: Nit: This would be `None` case, right?
```suggestion
        None => Err(errors::ProcessTrackerError::JobNotFound),
```

crates/router/src/workflows/outgoing_webhook_retry.rs
```diff
@@ -34,202 +38,379 @@ use crate::{
...
+        db,
+        key_manager_state,
+        &merchant_key_store,
+        Some(&profile_id),
+        &merchant_id,
+    )
+    .await?
+    .ok_or(errors::ApiErrorResponse::ProfileNotFound {
+        id: profile_id.get_string_repr().to_string(),
+    })?;
```
SanchithHegde, line 113: Why are we checking at this point in the flow if the profile ID belongs to the merchant account or not?

crates/router/src/workflows/outgoing_webhook_retry.rs
```diff
@@ -34,202 +38,379 @@ use crate::{
...
+    let webhook_details = webhooks_core::get_webhook_details_for_event_type(
+        &business_profile,
+        event_type,
+    )
+    .map_err(|err| {
+        logger::error!(original_webhook_error=?err, business_profile_id=?business_profile.get_id(), "Failed to get webhook details for event type");
+        errors::ProcessTrackerError::ResourceFetchingFailed {
+            resource_name: "webhook_details".to_string(),
+        }
+    })?;
```
SanchithHegde, line 133: Can we return `CustomResult` from this function and call `change_context()` instead of `map_err()` here?

crates/router/src/workflows/outgoing_webhook_retry.rs
```diff
@@ -34,202 +38,379 @@ use crate::{
...
+        return Ok(());
+    }
+
+    let key_store = db
+        .get_merchant_key_store_by_merchant_id(
+            key_manager_state,
+            &tracking_data.merchant_id,
+            &db.get_master_key().to_vec().into(),
+        )
+        .await?;
```
SanchithHegde, line 150: We already have the `merchant_key_store` variable, why do we need to fetch it again?

crates/router/src/workflows/outgoing_webhook_retry.rs
```diff
@@ -34,202 +38,379 @@ use crate::{
...
+                    event_type,
+                    event_class,
+                    payment_id,
+                    primary_object_type,
+                    content,
+                    primary_object_created_at,
+                    webhook_detail_clone,
+                )
+                .await
+                .unwrap_or(())
```
SanchithHegde, line 177: Why the `unwrap_or(())`?
And we seem to be ignoring the error, can we log it instead?

crates/router/src/workflows/outgoing_webhook_retry.rs
```diff
@@ -34,202 +38,379 @@ use crate::{
...
+            }
+            .in_current_span(),
+        );
+    }
+    let _ = state
+        .store
+        .as_scheduler()
+        .finish_process_with_business_status(process.clone(), business_status::COMPLETED_BY_PT)
+        .await
+        .change_context(errors::WebhooksFlowError::OutgoingWebhookProcessTrackerTaskUpdateFailed);
```
SanchithHegde, line 187: Same here, we're ignoring the error here due to the `let _`, can we log the error instead?

crates/router/src/workflows/outgoing_webhook_retry.rs
```diff
@@ -34,202 +38,379 @@ use crate::{
...
+            &key_store,
+            &tracking_data.business_profile_id,
+        )
+        .await?;
+
+    let event_id = webhooks_core::utils::generate_event_id();
+    // let webhook_endpoint_id = tracking_data
+    //     .webhook_endpoint_id
+    //     .as_ref()
+    //     .ok_or(errors::ApiErrorResponse::WebhookBadRequest)?;
```
SanchithHegde, line 224: Nit: Remove commented code.

crates/router/src/workflows/outgoing_webhook_retry.rs
```diff
@@ -34,202 +38,379 @@ use crate::{
...
+        event_id,
+        event_type: initial_event.event_type,
+        event_class: initial_event.event_class,
+        is_webhook_notified: false,
+        primary_object_id: initial_event.primary_object_id,
+        primary_object_type: initial_event.primary_object_type,
+        created_at: now,
+        merchant_id: Some(business_profile.merchant_id.clone()),
+        business_profile_id: Some(business_profile.get_id().to_owned()),
+        primary_object_created_at: tracking_data.primary_object_created_at,
```
SanchithHegde, line 273: This could have been read from `initial_event` as well?

crates/router/src/core/webhooks/outgoing.rs
```diff
@@ -137,110 +174,162 @@ pub(crate) async fn create_event_and_trigger_outgoing_webhook(
...
+                .await;
+            }
+            .in_current_span(),
         );
-    })
-    .ok();
-
-    let cloned_key_store = merchant_context.get_merchant_key_store().clone();
-    // Using a tokio spawn here and not arbiter because not all caller of this function
-    // may have an actix arbiter
```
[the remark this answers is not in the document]
SanchithHegde: Please add this comment in `trigger_outgoing_webhook_bulk_workflow()` as well, before the `tokio::spawn()`.

crates/router/src/workflows/outgoing_webhook_retry.rs
```diff
@@ -34,202 +34,370 @@ use crate::{
...
+    let webhook_detail = match webhooks_core::get_webhook_detail_by_webhook_endpoint_id(
+        &business_profile,
+        &new_event.webhook_endpoint_id,
+    )
+    .await
+    {
+        Ok(detail) => detail,
+        Err(_e) => {
+            return Err(errors::ProcessTrackerError::UnexpectedFlow);
+        }
```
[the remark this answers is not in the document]
SanchithHegde: Could you please answer my previous questions?

> Why are we ignoring the error here?
>
> Can we use `change_context`() to propagate the error instead?

crates/router/src/workflows/outgoing_webhook_retry.rs
```diff
@@ -34,202 +38,379 @@ use crate::{
...
+
+    let key_store = db
+        .get_merchant_key_store_by_merchant_id(
+            key_manager_state,
+            &tracking_data.merchant_id,
+            &db.get_master_key().to_vec().into(),
+        )
+        .await?;
+    for webhook_detail in webhook_details.iter() {
+        let webhook_detail_clone = webhook_detail.clone();
```
SanchithHegde, line 152: Nit: You should be able to do something like:


```suggestion
    for webhook_detail in webhook_details {
```

And you can avoid the `webhook_detail.clone()`.

crates/router/src/workflows/outgoing_webhook_retry.rs
```diff
@@ -34,202 +38,379 @@ use crate::{
...
+        idempotent_event_id: Some(idempotent_event_id),
+        initial_attempt_id: Some(initial_event.event_id.clone()),
+        request: initial_event.request,
+        response: None,
+        delivery_attempt: Some(delivery_attempt),
+        metadata: initial_event.metadata,
+        webhook_endpoint_id: initial_event.webhook_endpoint_id,
+        is_overall_delivery_successful: Some(false),
+    };
+    let webhook_id = new_event.webhook_endpoint_id.clone();
```
SanchithHegde, line 283: Nit: Naming.


```suggestion
    let webhook_endpoint_id = new_event.webhook_endpoint_id.clone();
```

crates/router/src/core/webhooks/incoming.rs
```diff
@@ -4,8 +4,9 @@ use actix_web::FromRequest;
 #[cfg(feature = "payouts")]
 use api_models::payouts as payout_models;
 use api_models::webhooks::{self, WebhookResponseTracker};
+use common_enums::EventClass;
 use common_utils::{errors::ReportSwitchExt, events::ApiEventsType, ext_traits::AsyncExt};
-use diesel_models::ConnectorMandateReferenceId;
+use diesel_models::{enums::EventObjectType, ConnectorMandateReferenceId};
```
SanchithHegde, line 9: Let's avoid these direct imports please, we can qualify them with `enums::` instead.

crates/router/src/workflows/outgoing_webhook_retry.rs
```diff
@@ -34,202 +38,379 @@ use crate::{
...
+
+    let cloned_state = state.clone();
+    let (content, _) = Box::pin(get_outgoing_webhook_content_and_event_type(
+        state.clone(),
+        state.get_req_state(),
+        merchant_account.clone(),
+        merchant_key_store.clone(),
+        &tracking_data,
+    ))
+    .await?;
```
SanchithHegde, line 123: `get_outgoing_webhook_content_and_event_type()` isn't meant to be used as the primary way of obtaining the outgoing webhook content: it was added for backward compatibility reasons.

Another reason I would prefer to completely avoid it is that it has the caveat that the resource could have transitioned to a completely different status (as compared to what we're notifying about) by the time the workflow would be executed.

We previously had checks for it:
https://github.com/juspay/hyperswitch/blob/0a06ad91252d76d0c937344d8c76c9a2bbbc84d7/crates/router/src/workflows/outgoing_webhook_retry.rs#L175-L228

crates/router/src/core/webhooks/incoming.rs
```diff
@@ -1456,34 +1436,27 @@ async fn mandates_incoming_webhook_flow(
                 merchant_context.get_merchant_account().get_id(),
                 &mandate_id,
                 storage::MandateUpdate::StatusUpdate { mandate_status },
-                mandate,
+                mandate.clone(),
```
SanchithHegde, line 1439: Why the clone?

crates/router/src/core/webhooks/outgoing.rs
```diff
@@ -10,15 +10,16 @@ use common_utils::{
     type_name,
     types::keymanager::{Identifier, KeyManagerState},
 };
-use diesel_models::process_tracker::business_status;
-use error_stack::{report, ResultExt};
+use diesel_models::{business_profile::MultipleWebhookDetail, process_tracker::business_status};
+use error_stack::{report, Report, ResultExt};
 use hyperswitch_domain_models::type_encryption::{crypto_operation, CryptoOperation};
 use hyperswitch_interfaces::consts;
 use masking::{ExposeInterface, Mask, PeekInterface, Secret};
```
Aishwariyaa-Anand, line 17:
```suggestion
use masking;
```
Avoid using direct imports

crates/router/src/core/webhooks/outgoing.rs
```diff
@@ -46,6 +47,83 @@ use crate::{
 };
 
 const OUTGOING_WEBHOOK_TIMEOUT_SECS: u64 = 5;
+pub const OUTGOING_WEBHOOK_BULK_TASK: &str = "OUTGOING_WEBHOOK_BULK";
+pub const OUTGOING_WEBHOOK_RETRY_TASK: &str = "OUTGOING_WEBHOOK_RETRY";
```
Aishwariyaa-Anand, line 51: Can we move these constants to crates/router/src/core/webhooks/types.rs ?
They are used across v1 and v2 files.

crates/router/src/core/webhooks/outgoing.rs
```diff
@@ -46,6 +47,83 @@ use crate::{
...
+
+pub(crate) fn get_webhook_details_for_event_type(
```
Aishwariyaa-Anand, line 53: Can we reorder the functions to reflect the logical order of execution? That would help improve readability and make the flow easier to follow. Something like:
> add_bulk_outgoing_webhook_task_to_process_tracker
> create_event_and_trigger_outgoing_webhook
> trigger_webhook_and_raise_event
> trigger_webhook_to_merchant
> raise_webhooks_analytics_event

crates/router/src/core/webhooks/outgoing.rs
```diff
@@ -1,4 +1,4 @@
-use std::collections::HashMap;
+use std::collections::{HashMap, HashSet};
```
Aishwariyaa-Anand, line 2: 1. In `trigger_webhook_to_merchant`, where we match on `delivery_attempt`, we could introduce a **trait** to handle successful and failed deliveries for InitialAttempt, AutomaticRetry, and ManualRetry. This will help decouple the logic and improve code readability.
2. Functions like `increment_webhook_outgoing_received_count` and `increment_webhook_outgoing_not_received_count` can be reused from crates/router/src/core/webhooks/utils.rs instead of redefining.
3. In raise_webhooks_analytics_event, we could pass the trigger_webhook_result directly to help avoid an extra DB fetch for the updated event:
 ```
trigger_webhook_result: CustomResult<
        (domain::Event, Option<Report<errors::WebhooksFlowError>>),
        errors::WebhooksFlowError,
    >,
``` 

You can refer to this PR https://github.com/juspay/hyperswitch/pull/6613
These changes can be taken up in a separate PR

crates/router/src/core/admin.rs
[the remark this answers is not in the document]
Aishwariyaa-Anand: How about we introduce a separate set of webhook endpoints for create, update and delete.
This would make the code logic simple and also would make it easier for the merchant in case of webhook updation.

swetasharma03:
https://github.com/juspay/hyperswitch/pull/7578#discussion_r2189318409
As adding new endpoints will increase complexity of the PR, we will be breaking out this PR into 5 new fresh PRs.
CC: @SanchithHegde @Aishwariyaa-Anand

PR #7579, opened 2025-03-20
feat(core): add redis lock during insertion of event in event table during initial attempt of outgoing webhook delivery

cookieg13:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [X] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description
The current implementation of insertion of an event in `event` table does not handle the case where multiple requests with the same idempotent_event_id might be processed concurrently, leading to multiple unique constraint violation errors in the database.

This was reported by a merchant who was facing high volumes of unique constraint violation errors.

NOTE: this unique constraint log error issue is only applicable for the`initial` attempt of outgoing webhook delivery, not for retry attempts

**Why does this happen?**
1. Multiple requests with the same `idempotent_event_id` might reach the database at the same time.
2. Each request checks for the event, finds nothing (since it’s not committed yet), and proceeds to insert.
3. The first request succeeds, while the others fail due to the unique constraint on idempotent_event_id.

**Solution (Using Redis Lock)**
Before querying the database, acquire a lock in Redis for the given **idempotent_event_id**.
If there are multiple processes with same **idempotent_event_id** only one process will hold the lock, others will leave early.

 If a lock is acquired:
- Check if the event already exists in the database.
- If it exists, return early
-  Otherwise, insert the event.
- Once done, release the lock

### Additional Changes

## How did you test it?

Cannot be directly tested.

## Checklist

### DIFF
--- a/config/deployments/env_specific.toml
+++ b/config/deployments/env_specific.toml
@@ -154,6 +154,10 @@ bg_metrics_collection_interval_in_secs = 15           # Interval for collecting 
 delay_between_retries_in_milliseconds = 500 # Delay between retries in milliseconds
 redis_lock_expiry_seconds = 180             # Seconds before the redis lock expires
 
+[webhooks]
+outgoing_enabled = true
+redis_lock_expiry_seconds = 180            # Seconds before the redis lock expires
+
 # Main SQL data store credentials
 [master_database]
 username = "db_user"      # DB Username
--- a/config/deployments/env_specific.toml
+++ b/config/deployments/env_specific.toml
@@ -154,10 +154,6 @@ bg_metrics_collection_interval_in_secs = 15           # Interval for collecting 
 delay_between_retries_in_milliseconds = 500 # Delay between retries in milliseconds
 redis_lock_expiry_seconds = 180             # Seconds before the redis lock expires
 
-[webhooks]
-outgoing_enabled = true
-redis_lock_expiry_seconds = 180            # Seconds before the redis lock expires
-
 # Main SQL data store credentials
 [master_database]
 username = "db_user"      # DB Username
--- a/crates/router/src/core/webhooks/outgoing.rs
+++ b/crates/router/src/core/webhooks/outgoing.rs
@@ -136,6 +136,41 @@ pub(crate) async fn create_event_and_trigger_outgoing_webhook(
         is_overall_delivery_successful: Some(false),
     };
 
+    let lock_value = utils::perform_redis_lock(
+        &state,
+        &idempotent_event_id,
+        merchant_account.get_id().to_owned(),
+    )
+    .await?;
+
+    if lock_value.is_none() {
+        return Ok(());
+    }
+
+    if (state
+        .store
+        .find_event_by_merchant_id_event_id(
+            key_manager_state,
+            &merchant_id,
+            &event_id,
+            merchant_key_store,
+        )
+        .await)
+        .is_ok()
+    {
+        logger::debug!(
+            "Event with idempotent ID `{idempotent_event_id}` already exists in the database"
+        );
+        utils::free_redis_lock(
+            &state,
+            &idempotent_event_id,
+            merchant_account.get_id().to_owned(),
+            lock_value,
+        )
+        .await?;
+        return Ok(());
+    }
+
     let event_insert_result = state
         .store
         .insert_event(key_manager_state, new_event, merchant_key_store)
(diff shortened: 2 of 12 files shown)


crates/router/src/core/webhooks/utils.rs
```diff
@@ -154,3 +158,98 @@ pub(crate) fn get_idempotent_event_id(
...
+        WEBHOOK_LOCK_PREFIX,
+        merchant_id.get_string_repr(),
+        unique_locking_key
+    );
+    let redis_lock_expiry_seconds = state.conf().lock_settings.redis_lock_expiry_seconds;
+
+    let redis_lock_result = redis_conn
+        .set_key_if_not_exists_with_expiry(
+            &redis_locking_key.as_str().into(),
+            state.get_request_id(),
```
SanchithHegde, line 186: Why do we need the request ID here? And what would happen if the request ID is `None`, considering `get_request_id()` returns an `Option<String>`?

crates/router/src/core/webhooks/utils.rs
```diff
@@ -154,3 +158,98 @@ pub(crate) fn get_idempotent_event_id(
...
+        )
+        .await;
+
+    match redis_lock_result {
+        Ok(redis::SetnxReply::KeySet) => {
+            logger::info!("Lock acquired for {:?}", redis_locking_key);
+            Ok(true)
+        }
+        Ok(redis::SetnxReply::KeyNotSet) => {
+            logger::info!("Lock already held for {:?}", redis_locking_key);
```
SanchithHegde, line 197: Since `redis_locking_key` is a `String`, you can use the `Display` impl itself here.

```suggestion
            logger::info!("Lock acquired for {redis_locking_key}");
            Ok(true)
        }
        Ok(redis::SetnxReply::KeyNotSet) => {
            logger::info!("Lock already held for {redis_locking_key}");
```

crates/router/src/core/webhooks/utils.rs
```diff
@@ -154,3 +158,98 @@ pub(crate) fn get_idempotent_event_id(
...
+            Ok(false)
+        }
+        Err(err) => Err(err.change_context(errors::ApiErrorResponse::InternalServerError)),
+    }
+}
+
+pub const WEBHOOK_LOCK_PREFIX: &str = "WEBHOOK_LOCK";
+
+pub async fn free_redis_lock<A>(
```
SanchithHegde, line 206: Attach context message wherever you're raising internal server errors in this function.

crates/router/src/core/webhooks/utils.rs
```diff
@@ -154,3 +158,98 @@ pub(crate) fn get_idempotent_event_id(
...
+        .await
+    {
+        Ok(val) => {
+            if val == state.get_request_id() {
+                match redis_conn
+                    .delete_key(&redis_locking_key.as_str().into())
+                    .await
+                {
+                    Ok(redis::types::DelReply::KeyDeleted) => {
+                        logger::info!("Lock freed {:?}", redis_locking_key);
```
SanchithHegde, line 236: Can use `Display` impl here.

```suggestion
                        logger::info!("Lock freed {redis_locking_key}");
```

crates/router/src/core/webhooks/utils.rs
```diff
@@ -154,3 +158,98 @@ pub(crate) fn get_idempotent_event_id(
...
+        .get_redis_conn()
+        .change_context(errors::ApiErrorResponse::InternalServerError)?;
+
+    let redis_locking_key = format!(
+        "{}_{}_{}",
+        WEBHOOK_LOCK_PREFIX,
+        merchant_id.get_string_repr(),
+        unique_locking_key
+    );
+    let redis_lock_expiry_seconds = state.conf().lock_settings.redis_lock_expiry_seconds;
```
SanchithHegde, line 181: @jarnura Are we okay with using the same TTL as that used for API locking?

crates/router/src/core/webhooks/utils.rs
```diff
@@ -154,3 +158,98 @@ pub(crate) fn get_idempotent_event_id(
 pub(crate) fn generate_event_id() -> String {
     common_utils::generate_time_ordered_id("evt")
 }
+
+pub async fn perform_redis_lock<A>(
```
SanchithHegde, line 162: Nit: Can keep the visibility of the `perform_redis_lock()` and `free_redis_lock()` functions as `pub(super)`, so that they're only visible within the `webhooks` module.

crates/router/src/core/webhooks/outgoing.rs
```diff
@@ -136,6 +136,33 @@ pub(crate) async fn create_event_and_trigger_outgoing_webhook(
...
+        .await)
+        .is_ok()
+    {
+        logger::debug!(
+            "Event with idempotent ID `{idempotent_event_id}` already exists in the database"
+        );
+        return Ok(());
+    }
+
     let event_insert_result = state
```
SanchithHegde, line 166: I guess we can remove the `if error.current_context().is_db_unique_violation()` check we're performing in the following block.

crates/router/src/core/webhooks/utils.rs
cookieg13: @SanchithHegde the `get_request_id()` is used as redis value in api_locking, so I had followed the same. But if request_id is None, it could lead to unintended behavior in the redis locking mechanism.

I have changed it to `uuid` now, shouldn't we also do the same change for the api_locking usecases?

crates/router/src/core/webhooks/utils.rs
jarnura: Have a separate locking expiry time

jarnura (dismissed):

config/deployments/env_specific.toml
```diff
@@ -154,6 +154,10 @@ bg_metrics_collection_interval_in_secs = 15           # Interval for collecting
 delay_between_retries_in_milliseconds = 500 # Delay between retries in milliseconds
 redis_lock_expiry_seconds = 180             # Seconds before the redis lock expires
 
+[webhooks]
+outgoing_enabled = true
+redis_lock_expiry_seconds = 180            # Seconds before the redis lock expires
```
SanchithHegde, line 159: Keep this section in either:

- `env_specific.toml` file
- `integration_test.toml`, `sandbox.toml` and `production.toml` files

Don't keep it in both sets of files, it would cause conflicts.

SanchithHegde (commented):

SanchithHegde (approved):

jarnura (approved):

PR #7622, opened 2025-03-25
ci(cypress): do not overwrite metadata during mca update call

arindam-sahoo:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [x] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [x] CI/CD

## Description

The Connector Update Body is being overwritten, which included the metadata as well. Due to this, if there were any required metadata fields of a connector, then the Connector Update Test would fail.

For example, `Braintree` connector has a required metadata field, and while testing the Connector Update, just because it was being overwritten by the `update-connector-body.json` fixture, the test fails as shown in the following screenshot.
image

### Additional Changes

## Motivation and Context

While adding the Card test Flows for the `Braintree` connector, the Connector Update Test Failed.

## How did you test it?

Running the `0000-CoreFlows.cy.js` for `Braintree` connector

image

image

## Checklist


pixincreate:
how does this change actually fixes things?

when creating the mca, auth is read that includes metadata. and you're passing it. unless the update call nullifies the metadata that is passed in the create call, update call shouldn't fail at all.

i can see that we're overwriting the metadata while making the update call. 

try removing the metadata object from `cypress-tests/cypress/fixtures/update-connector-body.json` and see if that helps.

arindam-sahoo:
Yup, right. My bad, didn't even think that it may be because of the fixtures, making changes.

arindam-sahoo:
Changes are done in 0f3bb51391b951c8507a347eedf4ffae352f6180 and reverted back the wrongly made changes in 9e61292bf79e081e5f6cf21d01cd6c0e48127a3c

pixincreate (approved):

likhinbopanna (approved):

PR #7626, opened 2025-03-25
refactor(accounts): move accounts related tables to accounts schema

hrithikesh026:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

* moved merchant_account, business_profile and merchant_connector_account under accounts_schema.
* added `id` as varchar in business_profile and merchant_connector_account and added logic to populate it during creation in v1 itself.
* added migration to create index for every table after running compatible migrations. 

### Additional Changes

## How did you test it?

Create a merchant_account and MCA for that merchant account
and check if id column is populated for business_profile and mca.

MCA DB query:
```
select id, merchant_connector_id from merchant_connector_account
where merchant_connector_id = '<mca_id>';
```

image

Profile DB query:
```
select id, profile_id from business_profile
where profile_id = '<profile_id>';
```
image

## Checklist


crates/common_types/src/payment_methods.rs
```diff
@@ -1,15 +1,19 @@
...
+    AsExpression, Queryable,
 };
 use serde::{Deserialize, Serialize};
 use utoipa::ToSchema;
 
 /// Details of all the payment methods enabled for the connector for the given merchant account
 #[derive(Debug, Clone, Serialize, Deserialize, ToSchema, AsExpression)]
 #[serde(deny_unknown_fields)]
-#[diesel(sql_type = Jsonb)]
+#[diesel(sql_type = Json)]
```
jarnura, line 16: Why we moving away from Jsonb?

jarnura (changes requested):

crates/common_types/src/payment_methods.rs
hrithikesh026: because `payment_methods_enabled` column in MCA is `json` in v1. If we change it to `jsonb` in v2, then v1 will not be able `deserialise` that field.

hrithikesh026: One main advantage of jsonb is that the json data is validated before insertion. we can be sure that this json value be a valid one since application is writing it. No need for validation at DB level.

jarnura: okay

jarnura (approved):

v2_compatible_migrations/2024-08-28-081757_drop_not_null_constraints_on_v1_columns/up.sql
```diff
@@ -1,43 +1,62 @@
--- Drop not null constraint on org_id in organization table
--- Drop not null constraint on org_id in organization table
+-- Drop not null constraints
+------------------------ Organization -----------------------
 ALTER TABLE organization 
     DROP CONSTRAINT organization_pkey,
     ALTER COLUMN org_id DROP NOT NULL;
+-- Create index on org_id in organization table
+CREATE INDEX idx_organization_org_id ON organization (org_id);
```
Narayanbhat166, line 7: why are we creating these indexes again in v2? can you add appropriate comments?

Narayanbhat166 (approved):

jarnura (approved):

PR #7692, opened 2025-04-02
chore(payment_methods_v2): fingerprint changes for card

su-shivanshmathur:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

We were creating fingerprint on vault side for a card based on card number and expire data , issue was arriving when same card number was there but different expire information for the same customer , now a new fingerprint ID was getting generated.
Hence was being added as a new payment method in database for the same card number but changed metadata ( exp month, exp year )

To handle this , a new field is being added as seconday_fingerprint_id for maintaining the fingerprint based on card number and handle the updation of the metadata information based on the field

We can check for both the primary and secondary fingerprint and based on the result of both the fingerprint can create a new Payment Method ID or update an already existing one.

### Additional Changes

## Checklist


Narayanbhat166:
@ShivanshMathurJuspay , can you add a clear description as to why this change is required? you can update the content in motivation and context subheading.

ShivanshMathurJuspay:
> @ShivanshMathurJuspay , can you add a clear description as to why this change is required? you can update the content in motivation and context subheading.

@Narayanbhat166 have added the context of the change in the issue attached to the PR.

NishantJoshi00 (approved):
Looks good overall.

I am not able to figure out when it will be set and how are we planning to use it.

Specifically the **motivation** for this.

prasunna09:
> Looks good overall.
> 
> I am not able to figure out when it will be set and how are we planning to use it.
> 
> Specifically the **motivation** for this.

@NishantJoshi00 , Currently the fingerprint is generated in card number + expiry. this secondary fingerprint is generated only on card number. So if there is any change in expiry for same card number, secondary finger print will be same and primary fingerprint will be modified, this way the deduplication logic is handled by application.

jarnura (approved):

NishantJoshi00 (dismissed):

NishantJoshi00 (approved):

jarnura (approved):

prasunna09 (approved):

PR #7766, opened 2025-04-08
feat(router): add debit routing retry logic for alternate networks

chaitak-gorai:
## Type of Change

- [ ] Bugfix
- [ x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

Added retry logic for debit routing for alternate networks. 
Note:
- Just the logic is implemented keeping `alternate_network_retry_possible = true` 
- db/enum changes will be done through separate pr.

### Additional Changes

## Motivation and Context

[Debit Routing Retries Solution Doc](https://docs.google.com/document/d/1DjM1wETUu3eOOPVRdiT9yLbCcbeQsBOYi9ySdW8WqFg/edit?tab=t.0)

## How did you test it?

NA

## Checklist

- [x ] I addressed lints thrown by `cargo clippy`
- [x ] I reviewed the submitted code

### DIFF
--- a/crates/router/src/core/payments.rs
+++ b/crates/router/src/core/payments.rs
@@ -698,6 +698,7 @@ where
                                 #[cfg(not(feature = "frm"))]
                                 None,
                                 &business_profile,
+                                is_debit_routing_performed,
                             )
                             .await?;
                         };
@@ -1466,6 +1467,34 @@ pub fn find_connector_with_networks(
     })
 }
 
+pub fn get_next_connector_with_global_network(
+    connectors: &mut IntoIter<api::ConnectorRoutingData>,
+) -> RouterResult<(api::ConnectorData, enums::CardNetwork)> {
+    connectors
+        .find_map(|connector_data| {
+            connector_data
+                .network
+                .filter(|n| n.is_global_network())
+                .map(|network| (connector_data.connector_data, network))
+        })
+        .ok_or(errors::ApiErrorResponse::InternalServerError)
+        .attach_printable("Connector with global network not found in connectors iterator")
+}
+
+pub fn get_next_connector_with_diff_network(
+    connectors: &mut IntoIter<api::ConnectorRoutingData>,
+    prev_network: Option<enums::CardNetwork>,
+) -> RouterResult<(api::ConnectorData, enums::CardNetwork)> {
+    connectors
+        .find_map(|connector_data| {
+            connector_data
+                .network
+                .filter(|n| Some(n) != prev_network.as_ref())
+                .map(|network| (connector_data.connector_data, network))
+        })
+        .ok_or(errors::ApiErrorResponse::InternalServerError)
+        .attach_printable("Connector with different network not found in connectors iterator")
+}
 
 #[cfg(feature = "v2")]
 #[instrument(skip_all)]
--- a/crates/router/src/core/payments/retry.rs
+++ b/crates/router/src/core/payments/retry.rs
@@ -1,5 +1,6 @@
 use std::{str::FromStr, vec::IntoIter};
 
+use common_enums::enums::CardNetwork;
 use common_utils::{ext_traits::Encode, types::MinorUnit};
 use diesel_models::enums as storage_enums;
 use error_stack::{report, ResultExt};
@@ -45,6 +46,7 @@ pub async fn do_gsm_actions<F, ApiRequest, FData, D>(
     schedule_time: Option<time::PrimitiveDateTime>,
     frm_suggestion: Option<storage_enums::FrmSuggestion>,
     business_profile: &domain::Profile,
+    _is_debit_routing_performed: bool,
 ) -> RouterResult<types::RouterData<F, FData, types::PaymentsResponseData>>
 where
     F: Clone + Send + Sync,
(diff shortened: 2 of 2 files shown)


crates/router/src/core/payments/retry.rs
```diff
@@ -65,138 +67,267 @@ where
...
+                            true,
+                            frm_suggestion,
+                            business_profile,
+                            false,
+                        )
+                        .await?;
+                    }
+                } else {
+                    //existing logic of retries if alternate_network_retry_possible = false
+                    if step_up_retry_possible {
```
ShankarSinghC, line 230: Previously if a setup_retry payment fails the flow use to break as this block of code was outside the loop. Now how do we ensure that the flow breaks after the step_up_retry ?

This case will arise  when : - 
1. The retry count is set to 3 for a profile.
2. First payment was `no_three_ds` and resulted in authentication error.
3. Payment is retried with `three_ds` and let's consider it again resulted in some error.
4. This payment is not supposed to be retried as per the existing implementation. But since step up retry is inside the loop we again check gsm based on the error

crates/router/src/core/payments.rs
```diff
@@ -1466,6 +1467,34 @@ pub fn find_connector_with_networks(
...
+            connector_data
+                .network
+                .filter(|n| n.is_global_network())
+                .map(|network| (connector_data.connector_data, network))
+        })
+        .ok_or(errors::ApiErrorResponse::InternalServerError)
+        .attach_printable("Connector with global network not found in connectors iterator")
+}
+
+pub fn get_next_connector_with_diff_network(
```
ShankarSinghC, line 1484: If the retry payment also results in a network network error there is a chance that this function to return the first connector with which the initial payment was done.

This case will arise  when : - 
1. The retry count is set to 3 for a profile and debit routing is enabled.
2. Connector list after the routing is  performed `[(adyen: star), (adyen: pulse), (adyen: visa), (stripe: visa), (cybersource: visa)]`
3. According to the above list the first payment will be done through `(adyen: star)`
4. Consider the first payment resulted in network error.
5. For the retry this above function will be triggered and `(adyen: pulse)` will be chosen.
6. If the retry payment also fails with network error this function will be triggered again as the retry count is 3.
7. Now there is a possibility that the function might Some(n) != prev_network give `(adyen: star)` as we are only checking for `Some(n) != prev_network`. 

To fix this we have to eliminate the element from the list with which we have already retried.

ShankarSinghC (commented):
Consider using more descriptive variable names in the map closures to improve readability and maintainability

crates/router/src/core/payments/retry.rs
chaitak-gorai: If we don't want to retry after the step_up_retry is done then we can break from the loop in the step_up_retry block 
```
                        router_data = do_retry(
                            &state.clone(),
                            req_state.clone(),
                            original_connector_data,
                            operation,
                            customer,
                            merchant_account,
                            key_store,
                            payment_data,
                            router_data,
                            validate_result,
                            schedule_time,
                            true,
                            frm_suggestion,
                            business_profile,
                            false, //should_retry_with_pan is not applicable for step-up
                        )
                        .await?;
                        break; 
```

crates/router/src/core/payments.rs
chaitak-gorai: Got this edge case. 

So should we eliminate items from the list or should we keep a list of the (connector,network) we have retried and pass that list to these functions?

ShankarSinghC: We currently use `get_connector_data` for the retries which consumes the already retried connector (or elements) so that in the next retry the same connector won't be chosen.

crates/router/src/core/payments/retry.rs
ShankarSinghC: @sai-harsha-vardhan can you confirm this

crates/router/src/core/payments.rs
chaitak-gorai: As discussed,
we are going to remove the failed network form the `connector_routing_data` list for the network error cases,

Modified the `get_next_connector_with_diff_network` function to delete the elements having the `prev_netwok` then return the new connector network pair.

crates/router/src/core/payments.rs
```diff
@@ -1466,6 +1467,62 @@ pub fn find_connector_with_networks(
...
+    *connectors = filtered_prev_connector_newtwork_pairs.into_iter();
+
+    connectors
+    .find_map(|connector_data| {
+        connector_data
+            .network
+            .map(|network| (connector_data.connector_data, network))
+    })
+    .ok_or(errors::ApiErrorResponse::InternalServerError)
+    .attach_printable("Connector with different network not found in connectors iterator")
```
ShankarSinghC, line 1524: Update the error message. We are trying to find different connector with same network

crates/router/src/core/payments/retry.rs
```diff
@@ -65,138 +67,289 @@ where
...
+        Some(domain::PaymentMethodData::Card(card)) => card.card_network.clone(),
+        _ => None,
+    };
 
-    #[cfg(feature = "v1")]
-    let is_no_three_ds_payment = matches!(
-        payment_data.get_payment_attempt().authentication_type,
-        Some(storage_enums::AuthenticationType::NoThreeDs)
-    );
+    let mut previous_connector: api::ConnectorData = original_connector_data.clone();
```
ShankarSinghC, line 75: `original_connector_data` will always have the 1st connector with which payment was performed.
During the 2nd or 3rd retry we need the latest / recent connector with which the payment was done. For this we would have to get the previous connector from payment attempt.

@sai-harsha-vardhan can you confirm this ?

chaitak-gorai: For the first time only we are using it. In the loop Whenever we are getting a new connector we are updating the `previous_connector` to that.

crates/router/src/core/payments/retry.rs
```diff
@@ -27,6 +28,209 @@ use crate::{
...
+    match (
+        context.alternate_network_retry,
+        context.step_up_retry,
+        context.pan_retry,
+        context.prev_network_type,
+        context.debit_routing_status,
+    ) {
+        (AlternateNetworkRetryPossible, StepUpRetryPossible, _, PrevNetworkIsGlobal, _) => (
+            RetryAction::StepUpRetry,
+            ConnectorSelectionAction::SelectSameConnectorNetwork(
```
sai-harsha-vardhan, line 150: Should we blindly take this decisions and fail over, without considering connector list? It could be a case where our decision suitable is not available, then we should fallback to other retry right?

crates/router/src/core/payments.rs
```diff
@@ -1466,6 +1467,77 @@ pub fn find_connector_with_networks(
...
 }
 
+pub fn get_next_connector_with_global_network(
+    connectors: &mut IntoIter<api::ConnectorRoutingData>,
+) -> RouterResult<(api::ConnectorData, enums::CardNetwork)> {
+    connectors
+        .find_map(|connector_data| {
+            connector_data
+                .network
+                .filter(|n| n.is_global_network())
```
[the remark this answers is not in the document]
chaitak-gorai: done

crates/router/src/core/payments.rs
```diff
@@ -1466,6 +1467,77 @@ pub fn find_connector_with_networks(
...
+
+pub fn select_next_connector_network(
+    connectors: &mut IntoIter<api::ConnectorRoutingData>,
+) -> RouterResult<(api::ConnectorData, enums::CardNetwork)> {
+    let connector_data = connectors
+        .next()
+        .ok_or(errors::ApiErrorResponse::InternalServerError)
+        .attach_printable("No next connector found in the connectors iterator")?;
+
+    let network = connector_data
```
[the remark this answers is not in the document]
chaitak-gorai: done, added a extra match case where it is called

crates/router/src/core/payments/retry.rs
```diff
@@ -27,6 +28,209 @@ use crate::{
...
+                previous_connector
+                .expect("Previous connector must be available"),
```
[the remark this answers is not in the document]
chaitak-gorai: throwing exception error now

crates/router/src/core/payments/retry.rs
```diff
@@ -27,6 +28,209 @@ use crate::{
...
+                PrevNetworkType::PrevNetworkIsLocal
+            },
+            debit_routing_status: if is_debit_routing_performed {
+                DebitRoutingStatus::DebitRoutingApplied
+            } else {
+                DebitRoutingStatus::DebitRoutingNotApplied
+            },
+        }
+    }
+}
```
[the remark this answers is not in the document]
chaitak-gorai: as suggested by @jarnura for better readability we are using enums instead of bools

crates/router/src/core/payments/retry.rs
```diff
@@ -27,6 +28,209 @@ use crate::{
...
+impl RetryContext {
+    pub fn new(
+        alternate_network_retry_possible: bool,
+        step_up_retry_possible: bool,
+        should_retry_with_pan: bool,
+        prev_is_global_network: bool,
+        is_debit_routing_performed: bool,
+    ) -> Self {
+        RetryContext {
+            alternate_network_retry: if alternate_network_retry_possible {
```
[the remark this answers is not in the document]
chaitak-gorai: done, Implemented `from` for each enums.

```
impl From<bool> for AlternateNetworkRetry {
    fn from(value: bool) -> Self {
        if value {
            Self::AlternateNetworkRetryPossible
        } else {
            Self::AlternateNetworkRetryNotPossible
        }
    }
}

        RetryContext {
            alternate_network_retry: alternate_network_retry_possible.into(),
```

crates/router/src/core/payments/retry.rs
```diff
@@ -27,6 +28,209 @@ use crate::{
...
+            RetryAction::StepUpRetry,
+            ConnectorSelectionAction::SelectConnectorWithGlobalNetwork,
+        ),
+        (AlternateNetworkRetryPossible, StepUpRetryNotPossible, _, _, _) => (
+            RetryAction::NetworkRetry,
+            ConnectorSelectionAction::RemoveCurrentNetworkAndChooseAnother(
+                previous_network.expect("Previous network must be available"),
+            ),
+        ),
+        (AlternateNetworkRetryNotPossible, StepUpRetryPossible, _, _, _) => (
```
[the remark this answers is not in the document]
chaitak-gorai: we should check the debit routing performed case, added this in the match.

crates/router/src/core/payments/retry.rs
```diff
@@ -27,6 +28,209 @@ use crate::{
...
+            ),
+        ),
+        (
+            AlternateNetworkRetryNotPossible,
+            StepUpRetryNotPossible,
+            PanRetryNotPossible,
+            _,
+            DebitRoutingApplied,
+        ) => (
+            RetryAction::AutoRetry,
```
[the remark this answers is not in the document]
chaitak-gorai: this will handle the default case and the present retry case when no step up and pan retry is done.

crates/router/src/core/payments/retry.rs
```diff
@@ -65,105 +270,116 @@ where
...
-            state,
-            merchant_account.get_id(),
-            original_connector_data.connector_name,
-        )
-        .await
-    } else {
-        false
-    };
+        match get_gsm_decision(gsm) {
+            api_models::gsm::GsmDecision::Retry => {
```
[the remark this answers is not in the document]
chaitak-gorai: what should be the value for the decision field in when the `step_up_retry = true`. I assumed it should be Retry only. 

as in the present logic, here also we are priortising the step up retry ( in the decision match fn) and also exiting the loop if step_up retry is done.

crates/router/src/core/payments.rs
```diff
@@ -1466,6 +1467,62 @@ pub fn find_connector_with_networks(
...
+        })
+        .ok_or(errors::ApiErrorResponse::InternalServerError)
+        .attach_printable("Connector with global network not found in connectors iterator")
+}
+
+pub fn remove_current_network_and_choose_another(
+    connectors: &mut IntoIter<api::ConnectorRoutingData>,
+    prev_network: enums::CardNetwork,
+) -> RouterResult<(api::ConnectorData, enums::CardNetwork)> {
+    let filtered_connectors: Vec<_> = connectors
```
sai-harsha-vardhan, line 1488: Can we directly use filter map and avoid collecting into vec?

crates/router/src/core/payments/retry.rs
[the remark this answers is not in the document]
sai-harsha-vardhan: Decision may not be Retry, because it is like any other retries which is independent. step up retry can be independently enabled, without enabling auto retry (decision = retry signifies retry with different connector)

crates/router/src/core/payments/retry.rs
```diff
@@ -27,6 +28,289 @@ use crate::{
     types::{self, api, domain, storage},
 };
 
+#[derive(Debug, Clone, Copy)]
+pub enum RetryAction {
```
sai-harsha-vardhan, line 32: Could we move all of this to domain models?

PR #7798, opened 2025-04-11
chore(users): add hubspot tracking to prod intent

Riddhiagrawal001:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

### Additional Changes

## Checklist

### DIFF
--- a/crates/api_models/src/user/dashboard_metadata.rs
+++ b/crates/api_models/src/user/dashboard_metadata.rs
@@ -105,6 +105,7 @@ pub struct ProdIntent {
     pub is_completed: bool,
     #[serde(default)]
     pub product_type: MerchantProductType,
+    pub business_country_name: Option<String>,
 }
 
 #[derive(Debug, serde::Deserialize, serde::Serialize, Clone)]
@@ -178,4 +179,26 @@ pub enum GetMetaDataResponse {
     IsChangePasswordRequired(bool),
     OnboardingSurvey(Option<OnboardingSurvey>),
     ReconStatus(Option<ReconStatus>),
+}
+
+#[derive(Clone, Debug, serde::Deserialize, serde::Serialize)]
+pub struct HubspotRequest {
+    #[serde(rename = "useHubspot")]
+    pub use_hubspot: bool,
+    pub country: String,
+    #[serde(rename = "hubspotFormId")]
+    pub hubspot_form_id: String,
+    pub firstname: String,
+    pub lastname: String,
+    pub email: String,
+    #[serde(rename = "companyName")]
+    pub company_name: String,
+    pub lead_source: String,
+    pub website: String,
+    pub phone: String,
+    pub role: String,
+    #[serde(rename = "monthlyGMV")]
+    pub monthly_gmv: String,
+    pub bd_notes: String,
+    pub message: String,
 }
--- a/crates/router/src/configs/settings.rs
+++ b/crates/router/src/configs/settings.rs
@@ -685,6 +685,13 @@ pub struct FallbackMerchantIds {
 }
 
 #[derive(Debug, Clone, Default, Deserialize)]
+pub struct HubSpot {
+    pub enabled: bool,
+    pub base_url: String,
+    pub form_id: String,
+}
+
+#[derive(Debug, Clone, Default, Deserialize)]
 pub struct UserSettings {
     pub password_validity_in_days: u16,
     pub two_factor_auth_expiry_in_secs: i64,
@@ -692,6 +699,7 @@ pub struct UserSettings {
     pub base_url: String,
     pub force_two_factor_auth: bool,
     pub force_cookies: bool,
+    pub hubspot: HubSpot,
 }
 
 #[derive(Debug, Deserialize, Clone)]
--- a/crates/router/src/core/user/dashboard_metadata.rs
+++ b/crates/router/src/core/user/dashboard_metadata.rs
@@ -3,11 +3,15 @@ use std::str::FromStr;
 use api_models::user::dashboard_metadata::{self as api, GetMultipleMetaDataPayload};
 #[cfg(feature = "email")]
 use common_enums::EntityType;
-use common_utils::pii;
+use common_utils::{
+    pii,
+    request::{Method, RequestBuilder, RequestContent},
+};
 use diesel_models::{
     enums::DashboardMetadata as DBEnum, user::dashboard_metadata::DashboardMetadata,
 };
 use error_stack::{report, ResultExt};
+use http::header;
 #[cfg(feature = "email")]
 use masking::ExposeInterface;
 use masking::PeekInterface;
@@ -17,7 +21,7 @@ use router_env::logger;
 use crate::{
     core::errors::{UserErrors, UserResponse, UserResult},
     routes::{app::ReqState, SessionState},
-    services::{authentication::UserFromToken, ApplicationResponse},
+    services::{authentication::UserFromToken, send_request, ApplicationResponse},
     types::domain::{self, user::dashboard_metadata as types, MerchantKeyStore},
     utils::user::dashboard_metadata as utils,
 };
@@ -501,7 +505,7 @@ async fn insert_metadata(
                     .await?;
                     let email_contents = email_types::BizEmailProd::new(
                         state,
-                        data,
+                        data.clone(),
                         theme.as_ref().map(|theme| theme.theme_id.clone()),
                         theme
                             .map(|theme| theme.email_config())
@@ -517,6 +521,51 @@ async fn insert_metadata(
                         .await;
                     logger::info!(prod_intent_email=?send_email_result);
                 }
+            }
+
+            if state.conf.user.hubspot.enabled {
+                let hubspotbody = api::HubspotRequest {
+                    use_hubspot: true,
+                    country: data.business_country_name.unwrap_or_default(),
+                    hubspot_form_id: state.conf.user.hubspot.form_id.to_string(),
+                    firstname: data.poc_name.unwrap_or_default(),
+                    lastname: "".to_string(),
+                    email: data
+                        .poc_email
+                        .clone()
+                        .unwrap_or_default()
+                        .peek()
+                        .to_string(),
+                    company_name: data.legal_business_name.unwrap_or_default(),
+                    lead_source: "Hyperswitch Dashboard".to_string(),
+                    website: data.business_website.unwrap_or_default(),
+                    phone: "".to_string(),
+                    role: "".to_string(),
+                    monthly_gmv: "".to_string(),
+                    bd_notes: "".to_string(),
+                    message: "".to_string(),
+                };
+
+                let base_url = email_types::get_base_url(&state);
+
+                let hubspot_request = RequestBuilder::new()
+                    .method(Method::Post)
+                    .url(&state.conf.user.hubspot.base_url)
+                    .set_body(RequestContent::FormUrlEncoded(Box::new(hubspotbody)))
+                    .attach_default_headers()
+                    .headers(vec![(
+                        header::ORIGIN.to_string(),
+                        format!("{base_url}/dashboard").into(),
+                    )])
+                    .build();
+
+                send_request(&state, hubspot_request, None)
+                    .await
+                    .change_context(UserErrors::InternalServerError)
+                    .attach_printable(format!(
+                        "Failed to send data to hubspot for user_id {}",
+                        user.user_id.clone(),
+                    ))?;
             }
 
             metadata
--- a/config/development.toml
+++ b/config/development.toml
@@ -403,6 +403,12 @@ base_url = "http://localhost:8080"
 force_two_factor_auth = false
 force_cookies = true
 
+[user.hubspot]
+enabled = false
+base_url = ""
+form_id=""
+
+
 [bank_config.eps]
 stripe = { banks = "arzte_und_apotheker_bank,austrian_anadi_bank_ag,bank_austria,bankhaus_carl_spangler,bankhaus_schelhammer_und_schattera_ag,bawag_psk_ag,bks_bank_ag,brull_kallmus_bank_ag,btv_vier_lander_bank,capital_bank_grawe_gruppe_ag,dolomitenbank,easybank_ag,erste_bank_und_sparkassen,hypo_alpeadriabank_international_ag,hypo_noe_lb_fur_niederosterreich_u_wien,hypo_oberosterreich_salzburg_steiermark,hypo_tirol_bank_ag,hypo_vorarlberg_bank_ag,hypo_bank_burgenland_aktiengesellschaft,marchfelder_bank,oberbank_ag,raiffeisen_bankengruppe_osterreich,schoellerbank_ag,sparda_bank_wien,volksbank_gruppe,volkskreditbank_ag,vr_bank_braunau" }
 adyen = { banks = "bank_austria,bawag_psk_ag,dolomitenbank,easybank_ag,erste_bank_und_sparkassen,hypo_tirol_bank_ag,posojilnica_bank_e_gen,raiffeisen_bankengruppe_osterreich,schoellerbank_ag,sparda_bank_wien,volksbank_gruppe,volkskreditbank_ag" }


crates/api_models/src/user/dashboard_metadata.rs
```diff
@@ -179,3 +180,25 @@ pub enum GetMetaDataResponse {
...
+    pub company_name: String,
+    pub lead_source: String,
+    pub website: String,
+    pub phone: String,
+    pub role: String,
+    #[serde(rename = "monthlyGMV")]
+    pub monthly_gmv: String,
+    pub bd_notes: String,
+    pub message: String,
+}
```
ThisIsMani, line 204: You can serialize these to CamelCase instead of renaming these manually.

https://serde.rs/attr-rename.html

crates/router/src/core/user/dashboard_metadata.rs
```diff
@@ -519,6 +523,51 @@ async fn insert_metadata(
                 }
             }
 
+            if state.conf.hubspot.enabled {
+                let hubspotbody = api::HubspotRequest {
+                    use_hubspot: true,
+                    country: data.business_country_name.unwrap_or_default(),
+                    hubspot_form_id: state.conf.hubspot.form_id.to_string(),
+                    firstname: data.poc_name.unwrap_or_default(),
+                    lastname: "".to_string(),
```
ThisIsMani, line 532: This the also the default value for the string.

You can use default implementation here as well using `Default::default()`.

I think it's better if the type is in domain and there's a constructor which gives this struct directly.

crates/router/src/configs/settings.rs
```diff
@@ -684,6 +685,13 @@ pub struct FallbackMerchantIds {
     pub merchant_ids: HashSet<id_type::MerchantId>,
 }
 
+#[derive(Debug, Clone, Default, Deserialize)]
+pub struct HubSpotSettings {
+    pub enabled: bool,
+    pub base_url: String,
```
ThisIsMani, line 691: Is this only storing the base URL for the Hubspot or the whole request URL?

crates/router/src/core/user/dashboard_metadata.rs
```diff
@@ -519,6 +523,51 @@ async fn insert_metadata(
...
+                    email: data
+                        .poc_email
+                        .clone()
+                        .unwrap_or_default()
```
ThisIsMani, line 536: What will be the default in this case?

crates/router/src/core/user.rs
```diff
@@ -96,7 +96,7 @@ pub async fn signup_with_merchant_id(
     let send_email_result = state
         .email_client
         .compose_and_send_email(
-            email_types::get_base_url(&state),
+            get_base_url(&state),
```
ThisIsMani, line 99: (NIT) We generally use util functions like this: `utils::some_function()`. Can you check the other util functions and keep the consistency.

crates/router/src/configs/settings.rs
```diff
@@ -684,6 +685,13 @@ pub struct FallbackMerchantIds {
     pub merchant_ids: HashSet<id_type::MerchantId>,
 }
 
+#[derive(Debug, Clone, Default, Deserialize)]
+pub struct HubSpotSettings {
```
ThisIsMani, line 689: (NIT) Hubspot is named `HubSpot` in some places and `Hobspot` in others. Can we make this consistent. Looking at the request it's probably the latter.

crates/api_models/src/user/dashboard_metadata.rs
Riddhiagrawal001: Not all the keys needs to be passed as Camelcase for eg : `company_name` should be passed as `companyName`
but `lead_source` should be passed as `lead_source`

crates/router/src/configs/settings.rs
Riddhiagrawal001: This is the whole request_url to which we will be doing the s2s call

crates/router/src/core/user/dashboard_metadata.rs
Riddhiagrawal001: Ideally the email should never be empty . May be we can replace default value as email from token

crates/router/src/configs/settings.rs
ThisIsMani: Can we rename this to `request_url` then? `base_url` doesn't make sense here.

config/config.example.toml
```diff
@@ -424,6 +424,11 @@ base_url = ""                        # Base url used for user specific redirects
 force_two_factor_auth = false        # Whether to force two factor authentication for all users
 force_cookies = true                 # Whether to use only cookies for JWT extraction and authentication
 
+[hubspot]
+enabled = false  # Whether to enable hubspot tracking
+base_url = ""    # Base url for hubspot tracking
+form_id=""       # HubSpot-specific form ID
```
ThisIsMani, line 430:
```suggestion
form_id=""       # Hubspot-specific form ID
```

crates/api_models/src/user/dashboard_metadata.rs
```diff
@@ -179,3 +180,53 @@ pub enum GetMetaDataResponse {
...
+            email,
+            company_name,
+            lead_source: "Hyperswitch Dashboard".to_string(),
+            website,
+            phone: Default::default(),
+            role: Default::default(),
+            monthly_gmv: Default::default(),
+            bd_notes: Default::default(),
+            message: Default::default(),
+        }
```
ThisIsMani, line 230:
```suggestion
        Self {
            use_hubspot: true,
            country,
            hubspot_form_id,
            firstname,
            email,
            company_name,
            lead_source: "Hyperswitch Dashboard".to_string(),
            website,
            ..Default::default()
        }
```

crates/api_models/src/user/dashboard_metadata.rs
```diff
@@ -179,3 +180,53 @@ pub enum GetMetaDataResponse {
...
+    ) -> Self {
+        Self {
+            use_hubspot: true,
+            country,
+            hubspot_form_id,
+            firstname,
+            lastname: Default::default(),
+            email,
+            company_name,
+            lead_source: "Hyperswitch Dashboard".to_string(),
```
ThisIsMani, line 223: Can we put this in a const.

crates/api_models/src/user/dashboard_metadata.rs
```diff
@@ -179,3 +180,53 @@ pub enum GetMetaDataResponse {
...
+    pub phone: String,
+    pub role: String,
+    #[serde(rename = "monthlyGMV")]
+    pub monthly_gmv: String,
+    pub bd_notes: String,
+    pub message: String,
+}
+
+impl HubspotRequest {
+    pub fn new(
```
ThisIsMani, line 207: (NIT) We don't generally write constructors in the api models, we can move this into domain models, since this is not a request coming to Hyperswitch, doesn't need to be here.

crates/router/src/core/user/dashboard_metadata.rs
```diff
@@ -52,7 +56,7 @@ pub async fn get_multiple_metadata(
     for key in metadata_keys {
         let data = metadata.iter().find(|ele| ele.data_key == key);
         let resp;
-        if data.is_none() && utils::is_backfill_required(key) {
+        if data.is_none() && user_utils::dashboard_metadata::is_backfill_required(key) {
```
ThisIsMani, line 59: We can import `utils::dashboard_metadata` again with a different name.

crates/router/src/workflows/api_key_expiry.rs
```diff
@@ -74,7 +74,7 @@ impl ProcessTrackerWorkflow<SessionState> for ApiKeyExpiryWorkflow {
             )
             .ok_or(errors::ProcessTrackerError::EApiErrorResponse)?;
 
-        let theme = theme_utils::get_most_specific_theme_using_lineage(
+        let theme = user_utils::theme::get_most_specific_theme_using_lineage(
```
ThisIsMani, line 77: (NIT) We can import `utils::user::theme` again and use it here instead of qualifying it here. Since this is used only once, not necessary here.

config/config.example.toml
```diff
@@ -424,6 +424,11 @@ base_url = ""                        # Base url used for user specific redirects
 force_two_factor_auth = false        # Whether to force two factor authentication for all users
 force_cookies = true                 # Whether to use only cookies for JWT extraction and authentication
 
+[hubspot]
+enabled = false  # Whether to enable hubspot tracking
+request_url = ""    # Request url for hubspot tracking
```
ThisIsMani, line 429: (NIT) Please align this.

crates/router/src/core/user/dashboard_metadata.rs
```diff
@@ -519,6 +522,25 @@ async fn insert_metadata(
                 }
             }
 
+            if state.conf.hubspot.enabled {
+                let hubspotbody = HubspotRequest::new(
```
ThisIsMani, line 526: (NIT)
```suggestion
                let hubspot_body = HubspotRequest::new(
```

crates/router/src/consts/user.rs
```diff
@@ -41,3 +41,5 @@ pub const REDIS_SSO_TTL: i64 = 5 * 60; // 5 minutes
 pub const DEFAULT_PROFILE_NAME: &str = "default";
 pub const DEFAULT_PRODUCT_TYPE: common_enums::MerchantProductType =
     common_enums::MerchantProductType::Orchestration;
+
+pub const HUBSPOT_LEAD_SOURCE: &str = "Hyperswitch Dashboard"; // hubspot lead source
```
ThisIsMani, line 45: Do we need a comment here? It is the same as the variable name.

crates/router/src/core/user/dashboard_metadata.rs
ThisIsMani: So, what happens if this is empty?

config/config.example.toml
```diff
@@ -424,6 +424,11 @@ base_url = ""                        # Base url used for user specific redirects
...
+form_id=""       # Hubspot-specific form ID
```
ThisIsMani, line 430: (NIT)
Can we do this in all the configs.
```suggestion
form_id = ""       # Hubspot-specific form ID
```

ThisIsMani (dismissed):

crates/external_services/src/crm.rs
```diff
@@ -0,0 +1,145 @@
...
+use reqwest;
+use router_env::logger;
+
+use crate::{
+    http_client, http_client::request::InvalidCRMConfig, hubspot_proxy::core::HubspotRequest,
+};
+
+/// Hubspot CRM configuration
+#[derive(Debug, Clone, serde::Deserialize)]
+pub struct HubspotSettings {
```
racnan, line 25: use proxy in name.

crates/external_services/src/crm.rs
```diff
@@ -0,0 +1,145 @@
...
+            Self::HubspotProxy { hubspot_proxy } => Arc::new(hubspot_proxy.clone()),
+            Self::NoCrm => Arc::new(NoCrm),
+        }
+    }
+}
+
+#[async_trait::async_trait]
+impl CRMInterface for NoCrm {
+    async fn make_body(&self, _details: CRMPayload) -> RequestContent {
+        RequestContent::Json(Box::new(serde_json::json!({})))
```
racnan, line 89: no need to import `serde_json` just for default/empty values.

crates/external_services/src/crm.rs
```diff
@@ -0,0 +1,145 @@
...
+    }
+
+    async fn send_request(
+        &self,
+        _proxy: &Proxy,
+        _request: Request,
+    ) -> CustomResult<reqwest::Response, HttpClientError> {
+        logger::info!("NO CRM manager is not configured, resolving with a 200 OK response");
+        Err(HttpClientError::UnexpectedState)
+            .attach_printable("NO CRM manager is not configured, resolving with a 200 OK response")
```
racnan, line 103: you are not returning `200` response as mentioned in the logs.

crates/external_services/src/http_client/request.rs
```diff
@@ -0,0 +1,60 @@
...
+#[derive(Debug, Clone)]
+pub struct InvalidCRMConfig(pub &'static str);
+
+impl std::error::Error for InvalidCRMConfig {}
+
+impl Display for InvalidCRMConfig {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        write!(f, "crm: {}", self.0)
+    }
+}
```
racnan, line 60: put this in CRM module.

crates/external_services/src/hubspot_proxy/core.rs
```diff
@@ -0,0 +1,76 @@
+/// Lead source constant for Hubspot
+pub const HUBSPOT_LEAD_SOURCE: &str = "Hyperswitch Dashboard"; // hubspot lead source
```
racnan, line 2: no need to create separate file for this.

crates/router/src/core/user/dashboard_metadata.rs
```diff
@@ -519,6 +519,42 @@ async fn insert_metadata(
...
+            let hubspot_request = state
+                .crm_client
+                .make_request(hubspotbody, base_url.to_string())
+                .await;
+
+            let _ = state
+                .crm_client
+                .send_request(&state.conf.proxy, hubspot_request)
+                .await
+                .inspect_err(|_| {
```
jarnura, line 551: Don't discard error, and do error/warn logging instead of info

crates/external_services/src/crm.rs
```diff
@@ -0,0 +1,157 @@
...
+
+        when(self.form_id.is_default_or_empty(), || {
+            Err(InvalidCRMConfig("form_id must not be empty"))
+        })
+    }
+}
+
+/// Error thrown when the crm config is invalid
+#[derive(Debug, Clone)]
+pub struct InvalidCRMConfig(pub &'static str);
```
jarnura, line 51:
```suggestion
pub struct InvalidCrmConfig(pub &'static str);
```
Lets not use full CAPITAL in snake-case, do the same in other places too

jarnura (changes requested):

crates/router/src/core/user/dashboard_metadata.rs
Riddhiagrawal001: > Don't discard error
We do not want the prodIntent flow to stop if the crm fails hence discarding the error

crates/hyperswitch_interfaces/src/crm.rs
```diff
@@ -0,0 +1,69 @@
...
+    request::{Request, RequestContent},
+};
+use masking::Secret;
+
+use super::types::Proxy;
+use crate::errors::HttpClientError;
+
+/// CRM Payload structure
+#[derive(Clone, Debug, serde::Serialize, Default)]
+pub struct CRMPayload {
```
ThisIsMani: We can move this to a trait type.

ThisIsMani (approved):

crates/hyperswitch_interfaces/src/crm.rs
Riddhiagrawal001: Will take this change in a separate PR

ThisIsMani (dismissed):

jarnura (dismissed):

racnan (approved):
@ShivanshMathurJuspay can you review metrics changes.
Also, how can we test if the metrics are correct after this is deployed.

ThisIsMani (dismissed):

config/deployments/env_specific.toml
SanchithHegde, line 1: Can consider adding this in `config.example.toml` as well.

ThisIsMani (approved):

racnan (dismissed):

crates/external_services/src/crm.rs
```diff
@@ -0,0 +1,157 @@
...
+#[derive(Debug, Clone)]
+pub struct InvalidCrmConfig(pub &'static str);
+
+impl std::error::Error for InvalidCrmConfig {}
+
+impl Display for InvalidCrmConfig {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        write!(f, "crm: {}", self.0)
+    }
+}
```
SanchithHegde, line 59: Would prefer `Display` and `Formatter` being qualified here with `std::fmt`, instead of being imported.

crates/external_services/src/crm.rs
```diff
@@ -0,0 +1,157 @@
...
+
+#[derive(Debug, Clone)]
+/// NoCrm struct
+pub struct NoCrm;
```
SanchithHegde, line 63: Nit: Can also derive `Copy`, if that helps.

crates/external_services/src/http_client.rs
```diff
@@ -0,0 +1,182 @@
...
+pub mod metrics;
+/// request module
+pub mod request;
+use std::{error::Error, time::Duration};
+
+use common_utils::request::RequestContent;
+pub use common_utils::request::{ContentType, Method, RequestBuilder};
+use error_stack::ResultExt;
+
+#[allow(missing_docs, missing_debug_implementations)]
```
SanchithHegde, line 17: Nit: We don't have to allow `missing_debug_implementations` for a function, clippy shouldn't complain about it.

jarnura (approved):

crates/external_services/src/http_client/metrics.rs
```diff
@@ -0,0 +1,9 @@
+use router_env::{counter_metric, global_meter, histogram_metric_f64};
+
+global_meter!(GLOBAL_METER, "ROUTER_API");
+
+counter_metric!(REQUEST_BUILD_FAILURE, GLOBAL_METER);
+
+histogram_metric_f64!(EXTERNAL_REQUEST_TIME, GLOBAL_METER);
+
+counter_metric!(AUTO_RETRY_CONNECTION_CLOSED, GLOBAL_METER);
```
SanchithHegde, line 9: Please remove these from the `router` crate if they are not being used there, and if not removed already.

crates/external_services/src/http_client/request.rs
```diff
@@ -0,0 +1,45 @@
...
+
+use common_utils::request::Headers;
+pub use common_utils::{errors::CustomResult, request::ContentType};
+use error_stack::ResultExt;
+use hyperswitch_interfaces::errors::HttpClientError;
+pub use masking::{Mask, Maskable};
+use router_env::{instrument, tracing};
+
+#[allow(missing_docs, missing_debug_implementations)]
+pub trait HeaderExt {
```
SanchithHegde, line 11: Can remove the `allow(missing_debug_implementations)` on traits as well.

crates/external_services/src/hubspot_proxy.rs
```diff
@@ -0,0 +1,76 @@
...
+    pub hubspot_form_id: String,
+
+    /// The first name of the user.
+    pub firstname: String,
+
+    /// The last name of the user.
+    pub lastname: String,
+
+    /// The email address of the user.
+    pub email: String,
```
SanchithHegde, line 25: Can consider making them `Secret<String>`.

crates/external_services/src/hubspot_proxy.rs
```diff
@@ -0,0 +1,76 @@
...
+    pub company_name: String,
+
+    /// The source of the lead, typically set to "Hyperswitch Dashboard".
+    pub lead_source: String,
+
+    /// The website URL of the company.
+    pub website: String,
+
+    /// The phone number of the user.
+    pub phone: String,
```
SanchithHegde, line 38: Can make this optional maybe? And let's wrap this within a `Secret` as well.

crates/external_services/src/hubspot_proxy.rs
```diff
@@ -0,0 +1,76 @@
...
+
+    /// Notes from the business development team.
+    pub bd_notes: String,
+
+    /// Additional message or comments.
+    pub message: String,
+}
+
+#[allow(missing_docs, missing_debug_implementations)]
+impl HubspotRequest {
```
SanchithHegde, line 55:
```suggestion
#[allow(missing_docs)]
impl HubspotRequest {
```

crates/hyperswitch_interfaces/src/crm.rs
```diff
@@ -0,0 +1,69 @@
...
+    pub business_identifier: Option<String>,
+
+    /// The website URL of the business.
+    pub business_website: Option<String>,
+
+    /// The name of the point of contact (POC) for the business.
+    pub poc_name: Option<String>,
+
+    /// The contact number of the point of contact (POC) for the business.
+    pub poc_contact: Option<String>,
```
SanchithHegde, line 42: Let's wrap these within `Secret`.

crates/external_services/src/hubspot_proxy.rs
Riddhiagrawal001: We cannot make this optional as this is a required field in hubspot_proxy

crates/external_services/src/http_client/metrics.rs
Riddhiagrawal001: Removed EXTERNAL_REQUEST_TIME & AUTO_RETRY_CONNECTION_CLOSED from router as those were not getting used

SanchithHegde (dismissed):

racnan (dismissed):

jarnura (approved):

ThisIsMani (dismissed):

racnan (approved):

ThisIsMani (approved):

jarnura (dismissed):

SanchithHegde (approved):

jarnura (approved):

PR #7803, opened 2025-04-12
refactor(middleware): add middleware to record metrics for request count and duration

SanchithHegde:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [x] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

This PR moves the metrics that record incoming request count, status code and duration to the middleware, so that all requests that the server receives would be accounted for. As part of this change, the following two metrics are recorded as part of the middleware rather than within `server_wrap()`:

- `REQUESTS_RECEIVED`: Records the number of requests that the server received. Has request path and HTTP method as attributes.
- `REQUEST_TIME`: Records the time it took to respond to the request. Also helps us obtain a count of the number of requests to which the server responded successfully. Has request path, HTTP method and HTTP status code as attributes.

Additionally, this PR removes the `REQUEST_STATUS` metric, since it becomes redundant with the `REQUEST_TIME` metric helping us obtain a count of requests that the server responded to, along with tracking the status code. This PR also removes some duplicate code from a couple of places.

## Motivation and Context

This change is being as part of a bigger change to improve the metrics emitted by the application. This change would help account for all the requests that the server received, record the complete time taken by the server when responding to requests, and possibly help with monitoring the server better.

## How did you test it?

Locally, by running the OpenTelemetry Collector, Prometheus and Grafana with our Docker Compose setup.

1. Using `REQUESTS_RECEIVED` to count the number of requests received by the server:

   Screenshot of graph displaying number of requests received

   The graph can be obtained by running a query like:

   ```text
   sum by(method, path) (increase(REQUESTS_RECEIVED_total[$__rate_interval]))
   ```

   Notice that the metric has `method` and `path` as attributes.

2. Using `REQUEST_TIME` to count the number of requests that the server responded to:

   
   Screenshot of graph displaying number of requests the server responded to

   The graph can be obtained by running a query like:

   ```text
   sum by(method, path, status_code) (increase(REQUEST_TIME_count[$__rate_interval]))
   ```

   Notice that the metric has `method`, `path` and `status_code` as attributes. We are able to use `REQUEST_TIME` to count requests (even though it is a histogram metric) because `REQUEST_TIME_count` is automatically generated which functions like a counter. 

   Ideally, the number recorded by `REQUESTS_RECEIVED` and `REQUEST_TIME` should be the same but in case of application restarts mid-request, the number recorded by `REQUEST_TIME` may be slightly lesser.

3. Using `REQUEST_TIME` to record the request duration:

   The 99th, 90th and 50th percentile values can be plotted:

   Screenshot of percentiles for request duration

   The graph can be obtained by running a query like:

   ```text
   histogram_quantile(0.99, sum by(le, method, path, status_code) (rate(REQUEST_TIME_bucket{path=~"/payments.*"}[$__rate_interval])))
   ```

   This would provide us the 99th percentile values. The 90th and 50th percentile values can be obtained by replacing `0.99` in the above query with `0.9` and `0.5`, respectively.

   The same data can be plotted as a heatmap:
   
   Screenshot of heatmap of request duration

   The heatmap can be obtained by running a query like:

   ```text
   sum by(le) (rate(REQUEST_TIME_bucket{path=~"/payments.*"}[$__rate_interval]))
   ```

## Checklist


crates/analytics/src/metrics/request.rs
```diff
@@ -22,14 +20,3 @@ where
...
-#[inline]
-pub async fn time_future<F, R>(future: F) -> (R, time::Duration)
-where
-    F: futures::Future<Output = R>,
-{
-    let start = time::Instant::now();
-    let result = future.await;
-    let time_spent = start.elapsed();
-    (result, time_spent)
-}
```
SanchithHegde, line 35: Deleted this in favor of the same function defined in the `common_utils` crate:

https://github.com/juspay/hyperswitch/blob/2123f63bc54030adf1d8f42f08240c71b03fc09c/crates/common_utils/src/metrics/utils.rs#L7-L17

crates/router/src/routes/metrics/utils.rs
```diff
@@ -1,12 +0,0 @@
...
-#[inline]
-pub async fn time_future<F, R>(future: F) -> (R, time::Duration)
-where
-    F: futures::Future<Output = R>,
-{
-    let start = time::Instant::now();
-    let result = future.await;
-    let time_spent = start.elapsed();
-    (result, time_spent)
-}
```
SanchithHegde, line 12: Deleted this in favor of the same function defined in the `common_utils` crate:

https://github.com/juspay/hyperswitch/blob/2123f63bc54030adf1d8f42f08240c71b03fc09c/crates/common_utils/src/metrics/utils.rs#L7-L17

crates/router/src/routes/metrics/request.rs
```diff
@@ -1,38 +1,5 @@
...
-) {
-    super::REQUEST_STATUS.add(
-        1,
-        router_env::metric_attributes!(
-            ("status_code", status_code),
-            ("flow", flow),
-            ("merchant_id", merchant_id.clone()),
-        ),
-    )
-}
```
SanchithHegde, line 34: Removed these two since these were only being called within `server_wrap()`.

ShivanshMathurJuspay (approved):
lgtm!

crates/router/src/middleware.rs
```diff
@@ -396,3 +397,88 @@ where
...
+        let request_method = Cow::<'static, str>::from(req.method().as_str().to_owned());
+
+        Box::pin(async move {
+            let mut attributes =
+                router_env::metric_attributes!(("path", request_path), ("method", request_method))
+                    .to_vec();
+
+            let response_fut = svc.call(req);
+
+            crate::routes::metrics::REQUESTS_RECEIVED.add(1, &attributes);
```
jarnura, line 467:
```suggestion
            metrics::REQUESTS_RECEIVED.add(1, &attributes);
```

crates/router/src/middleware.rs
```diff
@@ -396,3 +397,88 @@ where
...
+            let (response_result, request_duration) =
+                common_utils::metrics::utils::time_future(response_fut).await;
+            let response = response_result?;
+
+            attributes.extend_from_slice(router_env::metric_attributes!((
+                "status_code",
+                i64::from(response.status().as_u16())
+            )));
+
+            crate::routes::metrics::REQUEST_TIME
```
jarnura, line 478:
```suggestion
            metrics::REQUEST_TIME
```

jarnura (changes requested):

jarnura (approved):

NishantJoshi00 (approved):

ShivanshMathurJuspay (approved):

PR #7812, opened 2025-04-15
fix(connector): handle edge case where currency comes as empty upon payment decline

pixincreate:
## Type of Change

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

This PR fixes an edge case with GlobalPay where, upon payment decline, we get a deserialization error and that is because `currency` field is not expected to be empty but rather a valid `Currency` enum. 

### Additional Changes

## Checklist

### DIFF
--- a/crates/hyperswitch_connectors/src/connectors/globalpay/response.rs
+++ b/crates/hyperswitch_connectors/src/connectors/globalpay/response.rs
@@ -4,6 +4,7 @@ use masking::Secret;
 use serde::{Deserialize, Serialize};
 
 use super::requests;
+use crate::utils::deserialize_optional_currency;
 
 #[derive(Debug, Serialize, Deserialize)]
 pub struct GlobalpayPaymentsResponse {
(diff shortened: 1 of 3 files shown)


swangi-kumari (approved):

crates/hyperswitch_connectors/src/connectors/globalpay/response.rs
```diff
@@ -31,6 +32,7 @@ pub struct GlobalpayPaymentsResponse {
     /// The country in ISO-3166-1(alpha-2 code) format.
     pub country: Option<String>,
     /// The currency of the amount in ISO-4217(alpha-3)
+    #[serde(deserialize_with = "deserialize_optional_currency")]
```
deepanshu-iiitu, line 35: Should we do this change for amount as well?

deepanshu-iiitu (changes requested):

deepanshu-iiitu (approved):

PR #7871, opened 2025-04-22
feat(users): add support for caching and resolving last used lineage context

tsdk02:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

This PR introduces lineage context caching to enhance login and account-switch flows with better continuity and reduced DB load.

###  Key Changes

- **During Login:**
  - Attempt to fetch the last used `LineageContext` from Redis (`LINEAGE_CONTEXT_<user_id>`).
  - If context is found:
    - Validate whether a matching user role exists (using both `V1` and `V2` role versions).
    - If a valid user_role is found → use this lineage context to issue the JWT token.
    - If not → fallback to the default user_role-based resolution logic.
  - If context is **not found in cache**, fallback to the existing login flow entirely.

- **During Account Switch:**
  - After switching org/merchant/profile context and issuing the new JWT token,
    the new `LineageContext` is saved to Redis (`LINEAGE_CONTEXT_<user_id>`).
  - This allows restoring the last used org/merchant/profile on future logins.

- **Redis Cache Expiry:**
  - Cached lineage context entries are stored with a TTL of **7 days**.
  - If the user is inactive beyond this duration, the system reverts to default logic.

### Additional Changes

## Motivation and Context

- Improves user experience by remembering their last used account context across sessions.
- Reduces unnecessary DB lookups during login for returning users.
- Ensures fallback safety in case of invalid or expired lineage context.

## How did you test it?

Tested locally with TTL as 1 minute, keys are getting stored in redis
Screenshot 2025-04-22 at 7 40 32 PM

image

Screenshot 2025-04-22 at 7 41 43 PM

### Test 1:
- Log in to the merchant account. 
- Check redis using the key:  ```GET LINEAGE_CONTEXT_<user_id>``` The value should contain the last used lineage context for the user.
- Switch to another org/merchant/profile.
- Recheck redis with the same key: ```GET LINEAGE_CONTEXT_<user_id>```. The lineage should now be changed.

### Test 2:
- After the above test, log out and log in back, within the TTL for the lineage_context, which is 7 days.
- You should be logged in to exactly the same lineage which you last used.

### Test 3 (optional):
- Have a note of the lineage.
- Do not use this account for 7 days.
- Try logging in now.
- You should be logged in to a random account (possibly the lineage which you have not used last)
- This proves, that the key has expired and the default fallback login is now used.

## Checklist


crates/api_models/src/user.rs
```diff
@@ -396,3 +396,13 @@ pub struct ListProfilesForUserInOrgAndMerchantAccountResponse {
...
+
+#[derive(Debug, serde::Serialize, serde::Deserialize)]
+pub struct LineageContext {
+    pub user_id: String,
+    pub merchant_id: id_type::MerchantId,
+    pub role_id: String,
+    pub org_id: id_type::OrganizationId,
+    pub profile_id: id_type::ProfileId,
+    pub tenant_id: Option<id_type::TenantId>,
+}
```
ThisIsMani, line 408: Why is this in api_models?

crates/api_models/src/user.rs
```diff
@@ -396,3 +396,13 @@ pub struct ListProfilesForUserInOrgAndMerchantAccountResponse {
...
 }
+
+#[derive(Debug, serde::Serialize, serde::Deserialize)]
+pub struct LineageContext {
+    pub user_id: String,
+    pub merchant_id: id_type::MerchantId,
+    pub role_id: String,
+    pub org_id: id_type::OrganizationId,
+    pub profile_id: id_type::ProfileId,
+    pub tenant_id: Option<id_type::TenantId>,
```
ThisIsMani, line 407: Why is this option? If this is because of token type, we can insert default if the token is None, like we do in other places. 

++ @apoorvdixit88

crates/router/src/types/domain/user/decision_manager.rs
```diff
@@ -123,21 +128,118 @@ impl JWTFlow {
...
+                    utils::user_role::get_single_profile_id(state, user_role, &merchant_id).await?;
+                (
+                    user_id,
+                    org_id,
+                    merchant_id,
+                    profile_id,
+                    user_role.role_id.clone(),
+                    Some(user_role.tenant_id.clone()),
+                )
+            }
```
ThisIsMani, line 217: These 2 are same. Code can be done better if we move these outside of match.

crates/router/src/types/domain/user/decision_manager.rs
```diff
@@ -123,21 +128,118 @@ impl JWTFlow {
         state: &SessionState,
         next_flow: &NextFlow,
         user_role: &UserRole,
+        lineage_context: Option<LineageContext>,
```
ThisIsMani, line 131: Why do we need this as a parameter? Can't we get this inside the function itself?

crates/router/src/utils/user.rs
```diff
@@ -339,3 +342,48 @@ pub async fn validate_email_domain_auth_type_using_db(
     .then_some(())
     .ok_or(UserErrors::InvalidUserAuthMethodOperation.into())
 }
+
+pub async fn set_lineage_context_in_cache(
+    state: &SessionState,
+    lineage_context: user_api::LineageContext,
+    user_id: String,
```
ThisIsMani, line 349: (NIT) Revert the positions of the arguments, key should come before value.

crates/router/src/utils/user.rs
```diff
@@ -339,3 +342,48 @@ pub async fn validate_email_domain_auth_type_using_db(
...
+        .change_context(UserErrors::InternalServerError)
+        .attach_printable("Failed to get lineage context from redis")?;
+
+    match lineage_context {
+        Some(json_str) => match serde_json::from_str::<user_api::LineageContext>(&json_str) {
+            Ok(ctx) => Ok(Some(ctx)),
+            Err(err) => {
+                logger::error!("Failed to deserialize LineageContext: {:?}", err);
+                Ok(None)
+            }
```
ThisIsMani, line 385: If we really don't want this to mess up our flows, we should throw all the errors here like we did for connection failure above and the callers should log and ignore the errors, which we are already doing. So, we can throw the actual error here and there's no need to fallback.

crates/router/src/types/domain/user/decision_manager.rs
```diff
@@ -339,7 +442,29 @@ impl NextFlow {
...
+                            "No lineage context found in cache for user {}",
+                            user_id.to_string()
+                        );
+                        None
+                    }
+                    Err(err) => {
+                        logger::error!("Failed to fetch lineage context from Redis: {:?}", err);
+                        None
+                    }
+                };
```
ThisIsMani, line 463: (NIT) We can minimize this by logging the result if it's not Ok(Some(_)), we don't need separate logging if we do that. Would make the function less bloaty. We can move this to a util as well if needed.

crates/api_models/src/user.rs
tsdk02: moved to domain/user

crates/api_models/src/user.rs
apoorvdixit88: Yes, this should not be option, even in token type we need to change, for the time we can use default.

crates/router/src/types/domain/user/decision_manager.rs
apoorvdixit88: I think there is some another function, that is generating JWT for user_role, irrespective of lineage context @tsdk02 ?

crates/router/src/types/domain/user/decision_manager.rs
```diff
@@ -123,21 +128,118 @@ impl JWTFlow {
...
+            user_id: user_id.clone(),
+            merchant_id: merchant_id.clone(),
+            role_id: role_id.clone(),
+            org_id: org_id.clone(),
+            profile_id: profile_id.clone(),
+            tenant_id: tenant_id.clone(),
+        };
+
+        if let Err(e) =
+            utils::user::set_lineage_context_in_cache(state, lineage_context, user_id.clone()).await
```
apoorvdixit88, line 230: Are we setting context even if we received LineageContext as `None` in param?

crates/router/src/types/domain/user.rs
```diff
@@ -1458,3 +1458,13 @@ where
...
 }
+
+#[derive(Debug, serde::Serialize, serde::Deserialize)]
+pub struct LineageContext {
+    pub user_id: String,
+    pub merchant_id: id_type::MerchantId,
+    pub role_id: String,
+    pub org_id: id_type::OrganizationId,
+    pub profile_id: id_type::ProfileId,
+    pub tenant_id: Option<id_type::TenantId>,
```
apoorvdixit88, line 1469: Please omit the Option here.

crates/router/src/types/domain/user/decision_manager.rs
tsdk02: Yes, we are basically setting the context in the sign-in and sign-up flows

crates/router/src/types/domain/user/decision_manager.rs
tsdk02: Made the changes and moved to util function

crates/router/src/types/domain/user/decision_manager.rs
tsdk02: The other function is basically the sign-up flow, so yes, we can move this redis call inside the generate_jwt function.

crates/api_models/src/user.rs
tsdk02: Made `TenantId` mandatory, and updated the code to sue default tenant_id, if `None`

crates/router/src/utils/user.rs
tsdk02: Though the helper functions are throwing errors, the error gets propagated and eventually logged, whenever the set_lineage_context and get_lineage_context are called.

We do not mean to disrupt the login and switch functionalities because of this feature

crates/router/src/utils/user.rs
```diff
@@ -339,3 +342,49 @@ pub async fn validate_email_domain_auth_type_using_db(
     .then_some(())
     .ok_or(UserErrors::InvalidUserAuthMethodOperation.into())
 }
+
+pub async fn get_lineage_context_from_cache(
+    state: &SessionState,
+    user_id: String,
```
ThisIsMani, line 348: (NIT) Use `&str` here.

crates/router/src/utils/user.rs
```diff
@@ -339,3 +342,49 @@ pub async fn validate_email_domain_auth_type_using_db(
...
+                .attach_printable("Failed to deserialize LineageContext from JSON")?;
+            Ok(Some(ctx))
+        }
+        None => Ok(None),
+    }
+}
+
+pub async fn set_lineage_context_in_cache(
+    state: &SessionState,
+    user_id: String,
```
ThisIsMani, line 371: Here as well.

crates/router/src/types/domain/user.rs
```diff
@@ -1458,3 +1458,54 @@ where
...
+            utils::user::set_lineage_context_in_cache(state, user_id.clone(), self.clone()).await
+        {
+            logger::error!(
+                "Failed to set lineage context in Redis for user {}: {:?}",
+                user_id,
+                e
+            );
+        } else {
+            logger::debug!("Lineage context cached for user {}", user_id);
+        }
```
ThisIsMani, line 1509: Could have used `map_err` here.

crates/router/src/types/domain/user/decision_manager.rs
```diff
@@ -124,23 +128,85 @@ impl JWTFlow {
...
+                        state.global_store.find_user_role_by_user_id_and_lineage(
+                            &ctx.user_id,
+                            &tenant_id,
+                            &ctx.org_id,
+                            &ctx.merchant_id,
+                            &ctx.profile_id,
+                            UserRoleVersion::V2,
+                        ),
+                    )
+                    .ok()
```
ThisIsMani, line 160: We could separate out lineage context get and user role get. Would increase readability.

ThisIsMani (approved):

apoorvdixit88 (approved):

crates/router/src/types/domain/user/decision_manager.rs
tsdk02: Tried to optimize, removed use of `futures::executor::block_on`

PR #7892, opened 2025-04-24
feat(user): use global redis key prefix for user lineage context

tsdk02:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

This PR updates the lineage context Redis caching logic to use a dedicated global Redis key prefix (`global`) instead of the default tenant-scoped prefixes (`public_`, `public_v2_`), causing inconsistencies between deployments (i.e., v1 and v2).

Introduced a new helper: `get_redis_connection_for_global_tenant`
  - Fetches Redis connection with `key_prefix` set to `global`.
- Updated lineage context caching (`get_lineage_context_from_cache`, `set_lineage_context_in_cache`) to use the global Redis connection

This will enable [https://github.com/juspay/hyperswitch/pull/7871](https://github.com/juspay/hyperswitch/pull/7871) to work with both v1 and v2 merchant accounts.

The redis-key should now be in the format of global:LINEAGE_CONTEXT_<user_id>

### Additional Changes

## Motivation and Context

Previously, lineage context was being cached with environment-specific prefixes (`public_` in V1 and `public_v2_` in V2), leading to inconsistencies in user experience during login/logout flows across deployments. Switching to a consistent global prefix resolves this by isolating user lineage data under a unified Redis namespace.

## How did you test it?

Use test cases from this PR: [https://github.com/juspay/hyperswitch/pull/7871](https://github.com/juspay/hyperswitch/pull/7871)

The key should now be in the format of global:LINEAGE_CONTEXT_<user_id>
image

## Checklist

### DIFF
--- a/crates/router/src/utils/user.rs
+++ b/crates/router/src/utils/user.rs
@@ -143,6 +143,27 @@ pub fn get_redis_connection(state: &SessionState) -> UserResult<Arc<RedisConnect
         .attach_printable("Failed to get redis connection")
 }
 
+pub fn get_redis_connection_for_global_tenant(
+    state: &SessionState,
+) -> UserResult<Arc<RedisConnectionPool>> {
+    let redis_connection_pool = state
+        .store
+        .get_redis_conn()
+        .change_context(UserErrors::InternalServerError)
+        .attach_printable("Failed to get redis connection")?;
+
+    let global_tenant_prefix = &state.conf.multitenancy.global_tenant.redis_key_prefix;
+
+    Ok(Arc::new(RedisConnectionPool {
+        pool: Arc::clone(&redis_connection_pool.pool),
+        key_prefix: global_tenant_prefix.to_string(),
+        config: Arc::clone(&redis_connection_pool.config),
+        subscriber: Arc::clone(&redis_connection_pool.subscriber),
+        publisher: Arc::clone(&redis_connection_pool.publisher),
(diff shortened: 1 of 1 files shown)


apoorvdixit88 (approved):
We also need to check for `get_redis_connection` at other places apart from this lineage functionality in user module

jarnura (approved):

crates/router/src/utils/user.rs
```diff
@@ -143,6 +143,27 @@ pub fn get_redis_connection(state: &SessionState) -> UserResult<Arc<RedisConnect
         .attach_printable("Failed to get redis connection")
 }
 
+pub fn get_redis_connection_for_global_tenant(
```
dracarys18, line 146: Can you make the changes in the same function? Preferably taking tenant as a function parameter.

That will mandate the developer creating redis connection to think about the tenancy.

dracarys18 (changes requested):

dracarys18 (approved):
Approving this for now as this is merchant critical

crates/router/src/utils/user.rs
tsdk02: Sure, will take this up in a separate PR, as this will affect all the redis functions which exist currently.

PR #7940, opened 2025-04-29
feat(users): store and retrieve lineage_context from DB instead of Redis

tsdk02:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [x] Enhancement
- [x] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

Previously, the user's `lineage_context` (userid, org_id, merchant_id, profile_id, role_id, tenant_id) was cached in Redis.
This PR moves the storage and retrieval of `lineage_context` to the PostgreSQL `users` table instead of Redis.

This ensures permanent persistence of lineage_context in DB, as opposed to the 7 days TTL when stored in Redis.

### Key Changes

- **Schema Change:**
  - Added `lineage_context` as a `JSONB` column in the `users` table.

- **During Login:**
  - Attempt to fetch `lineage_context` from DB (`users.lineage_context`).
  - If present:
    - Validate the context by checking if a matching user role exists (checks both v1 and v2 user roles).
    - If valid → use it for JWT generation.
    - If not → fallback to default DB-based role resolution.
  - If not present → fallback directly to default role resolution.
  - DB update for the new lineage context is now done via `tokio::spawn` to avoid blocking response.

- **During Org/Merchant/Profile Switch:**
  - After switching, the new `lineage_context` is serialized and persisted in the `users` table.
  - This enables restoring the last used context during the next login.
  - The update is done in the background using `tokio::spawn` with success and failure logs.

- **Error Handling:**
  - All DB failures or deserialization errors are logged.
  - Failures do not interrupt the login/switch flow — safe fallback ensures continuity.

### Additional Changes

## Motivation and Context

By moving the `lineage_context` to the `users` table in DB:

- We ensure the context is consistently available regardless of deployment environment.
- The last used account context becomes part of the user's persistent record.
- It simplifies logic and avoids prefix mismatches across tenants.

## How did you test it?

Screenshot 2025-04-29 at 11 45 43 PM

image

#### Test Case 1: Lineage Context Caching on Login
1. Log in as a user with an assigned role.
2. Check that `lineage_context` is stored in the `users` table (PostgreSQL).
3. Log out and log back in → verify that the JWT reflects the same org/merchant/profile context without role fallback.

#### Test Case 2: Lineage Context Update on Org/Profile Switch
1. Log in as a user and switch to another org/merchant/profile.
2. Confirm that the updated context is persisted in the `lineage_context` field in the DB.
3. Log out and log in again → JWT should reflect the newly switched context.

#### Test Case 3: Fallback to Role Resolution upon clearing lineage_context in db (Tested on local)
1. Manually removed `lineage_context` from the DB for a user.
2. Logged in again → verified that the system correctly falls back to the default role resolution logic.
3. Confirm that the context is repopulated in the DB after login.

#### Test Case 4: Fallback to Role Resolution upon deleting user_role (Tested on local)
1. Manually deleted `user_role` from the DB for a user (user_role corresponds to user invited to another account)
2. Tried to log in again → verified that the system correctly falls back to the default role resolution logic.
3. Confirm that the context is repopulated in the DB after login.

## Checklist


crates/diesel_models/src/user.rs
```diff
@@ -72,6 +75,9 @@ pub enum UserUpdate {
     PasswordUpdate {
         password: Secret<String>,
     },
+    LineageContextUpdate {
+        lineage_context: Option<serde_json::Value>,
```
ThisIsMani, line 79: Why do we need option here?

crates/diesel_models/src/user.rs
```diff
@@ -121,6 +130,20 @@ impl From<UserUpdate> for UserUpdateInternal {
...
                 totp_secret: None,
                 totp_recovery_codes: None,
+                lineage_context: None,
+            },
+            UserUpdate::LineageContextUpdate { lineage_context } => Self {
+                name: None,
+                password: None,
+                is_verified: None,
+                // updating last_modified_at because of lineage context update will make it difficult to
+                // track the last modified time of other operations done by the user
```
ThisIsMani, line 140: I don't think this comment is required. We anyways update `last_modified_at` whenever we update any data. Did you want it to be not updated?

crates/diesel_models/src/user.rs
```diff
@@ -24,6 +24,7 @@ pub struct User {
     #[diesel(deserialize_as = OptionalDieselArray<Secret<String>>)]
     pub totp_recovery_codes: Option<Vec<Secret<String>>>,
     pub last_password_modified_at: Option<PrimitiveDateTime>,
+    pub lineage_context: Option<serde_json::Value>,
```
ThisIsMani, line 27: Can't we use the `LineageContext` struct here instead of JSON?

Refer this: https://github.com/juspay/hyperswitch/blob/b04ccafe48394fad86bf072b2e3a74626cc26550/crates/diesel_models/src/merchant_account.rs#L33

crates/router/src/core/user.rs
```diff
@@ -3182,9 +3182,26 @@ pub async fn switch_org_for_user(
...
+            },
+        )
+        .await
+        .map_err(|e| {
+            logger::error!(
+                "Failed to update lineage context for user {}: {:?}",
+                user_from_token.user_id,
+                e
+            )
+        });
```
ThisIsMani, line 3204: We can implement this on `LineageContext` or the `UserFromStorage` type if it is still big after the other changes.

crates/router/src/db/user.rs
```diff
@@ -239,6 +240,12 @@ impl UserInterface for MockDb {
                         last_password_modified_at: Some(common_utils::date_time::now()),
                         ..user.to_owned()
                     },
+                    storage::UserUpdate::LineageContextUpdate { lineage_context } => {
+                        storage::User {
+                            lineage_context: lineage_context.clone(),
+                            ..user.to_owned()
+                        }
```
ThisIsMani, line 247: We are not updating `last_modified_at` here? Are we not supposed to?

ThisIsMani (commented):
Stop the redis writes. From this PR only reads should happen from the redis.

crates/router/src/core/user.rs
tsdk02: But generally, db calls are not supposed to be isolated right, so added the code in the main function itself. Do you recommend having this in the impl on lineageContext?

crates/diesel_models/src/user.rs
tsdk02: Yes, we generally except the `last_modified_at` to be changed when the user performs actions like rotating password, setting totp etc, as it would help us in debugging. But updating `lineage_context` is a frequent activity and I feel modifiying `last_modified_at` every time will defeat the purpose of tracking significant user updates.

crates/router/src/db/user.rs
tsdk02: We are not updating last_modified_at in any kind of update for MockDb

tsdk02: Updated last_modified_at for all kinds of updates in MockDb now.

crates/router/src/core/user.rs
```diff
@@ -3182,9 +3185,29 @@ pub async fn switch_org_for_user(
             .clone(),
     };
 
-    lineage_context
-        .try_set_lineage_context_in_cache(&state, user_from_token.user_id.as_str())
-        .await;
+    tokio::spawn({
```
ThisIsMani, line 3188: Can we abstract these out.

crates/router/src/core/user.rs
```diff
@@ -3182,9 +3185,29 @@ pub async fn switch_org_for_user(
...
+                .await
+            {
+                logger::error!(
+                    "Failed to update lineage context for user {}: {:?}",
+                    user_id,
+                    e
+                );
+            } else {
+                logger::debug!("Successfully updated lineage context for user {}", user_id);
+            }
```
ThisIsMani, line 3208: Could use match instead.

crates/router/src/core/user.rs
```diff
@@ -3394,9 +3417,27 @@ pub async fn switch_merchant_for_user_in_org(
...
+                .await
+                .map_err(|e| {
+                    logger::error!(
+                        "Failed to update lineage context for user {}: {:?}",
+                        user_id,
+                        e
+                    );
+                });
+        }
+    });
```
ThisIsMani, line 3440: Why is this different from the update call from `switch_org`?

crates/router/src/db/user.rs
```diff
@@ -208,17 +209,20 @@ impl UserInterface for MockDb {
...
     ) -> CustomResult<storage::User, errors::StorageError> {
         let mut users = self.users.lock().await;
+        let last_modified_at = common_utils::date_time::now();
         users
             .iter_mut()
             .find(|user| user.user_id == user_id)
             .map(|user| {
                 *user = match &update_user {
                     storage::UserUpdate::VerifyUser => storage::User {
+                        last_modified_at,
```
ThisIsMani, line 219: 💯

crates/router/src/core/user.rs
tsdk02: Moved to a util function

jarnura (dismissed):

ThisIsMani (approved):

racnan (approved):

ThisIsMani (dismissed):

ThisIsMani (approved):

jarnura (approved):

PR #8006, opened 2025-05-12
feat(core): add a procedural macro for validating schema attributes for a struct

kashif-m:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description
This PR adds automatic validation for string length constraints defined in schema attributes. The validation prevents 5xx errors that occur when strings exceeding database column length limits are passed to the database.

The implementation:
1. Creates a proc macro `ValidateSchema` that reads `min_length` and `max_length` parameters from schema attributes
2. Generates validation code that is invoked at runtime
3. Returns appropriate 4xx errors with helpful messages when validation fails

### Additional Changes

## Motivation and Context
When API requests contain fields that exceed database column length limits, the application currently returns 5xx errors when those values are stored. This is incorrect behavior - it should be a 4xx error (invalid request) since the client provided invalid data.

This change implements the correct behavior, and also uses the same schema attributes for both OpenAPI spec generation and validation giving us a single source of truth. It also improves error handling - returning 4xx instead of 5xx errors for invalid requests

## How did you test it?
<details>
    <summary>1. Added unit test cases for the macro</summary>

Screenshot 2025-05-12 at 5 50 33 PM

</details>

<details>
    <summary>2. Pass invalid return_url (length exceeding 255 characters)</summary>

cURL

    curl --location --request POST 'http://localhost:8080/payments' \
        --header 'Accept-Language: es-ES' \
        --header 'api-key: dev_tizBo0XFMZBNZ3jWBlkrc6C81CiCdliXrLXFoAg2fTgoTmKvOIk2oAP43Gwb3AGK' \
        --data-raw '{"customer_id":"cus_Wz8ND2QMGUTQCZBgO4Jx_1","profile_id":"pro_DC4DZ899zblYKajNmd0K","customer_acceptance":{"acceptance_type":"online","accepted_at":"1963-05-03T04:07:52.723Z","online":{"ip_address":"127.0.0.1","user_agent":"amet irure esse"}},"amount":12,"currency":"SGD","confirm":false,"payment_link":true,"setup_future_usage":"off_session","capture_method":"automatic","billing":{"address":{},"email":"random@juspay.in"},"return_url":"https://example.com?q=example.com&q=example.com&q=example.com&q=example.com&q=example.com&q=example.com&q=example.com&q=example.com&q=example.com&q=example.com&q=example.com&q=example.com&q=example.com&q=example.com&q=example.com&q=example.com&q=example.abc","payment_link_config":{"theme":"#111111","is_setup_mandate_flow":true,"background_image":{"url":"https://img.freepik.com/free-photo/abstract-blue-geometric-shapes-background_24972-1841.jpg","position":"bottom","size":"cover"}}}'

Response

    {
        "error": {
            "type": "invalid_request",
            "message": "return_url must be at most 255 characters long. Received 258 characters",
            "code": "IR_06"
        }
    }

</details>

## Checklist

### DIFF
--- a/crates/api_models/src/payments.rs
+++ b/crates/api_models/src/payments.rs
@@ -847,6 +847,7 @@ impl AmountDetailsUpdate {
     Clone,
     ToSchema,
     router_derive::PolymorphicSchema,
+    router_derive::ValidateSchema,
 )]
 #[generate_schemas(PaymentsCreateRequest, PaymentsUpdateRequest, PaymentsConfirmRequest)]
 #[serde(deny_unknown_fields)]
@@ -963,7 +964,7 @@ pub struct PaymentsRequest {
     pub description: Option<String>,
 
     /// The URL to which you want the user to be redirected after the completion of the payment operation
-    #[schema(value_type = Option<String>, example = "https://hyperswitch.io")]
+    #[schema(value_type = Option<String>, example = "https://hyperswitch.io", max_length = 255)]
     pub return_url: Option<Url>,
 
     #[schema(value_type = Option<FutureUsage>, example = "off_session")]
@@ -1018,7 +1019,7 @@ pub struct PaymentsRequest {
     pub customer_acceptance: Option<CustomerAcceptance>,
 
     /// A unique identifier to link the payment to a mandate. To do Recurring payments after a mandate has been created, pass the mandate_id instead of payment_method_data
-    #[schema(max_length = 255, example = "mandate_iwer89rnjef349dni3")]
+    #[schema(max_length = 64, example = "mandate_iwer89rnjef349dni3")]
     #[remove_in(PaymentsUpdateRequest)]
     pub mandate_id: Option<String>,
 
--- a/crates/router/src/routes/payments.rs
+++ b/crates/router/src/routes/payments.rs
@@ -37,6 +37,12 @@ pub async fn payments_create(
 ) -> impl Responder {
     let flow = Flow::PaymentsCreate;
     let mut payload = json_payload.into_inner();
+    if let Err(err) = payload
+        .validate()
+        .map_err(|message| errors::ApiErrorResponse::InvalidRequestData { message })
+    {
+        return api::log_and_return_error_response(err.into());
+    };
 
     if let Some(api_enums::CaptureMethod::Scheduled) = payload.capture_method {
         return http_not_implemented();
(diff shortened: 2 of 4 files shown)


jarnura (approved):

crates/router_derive/src/macros/schema.rs
```diff
@@ -0,0 +1,88 @@
...
+        }
+
+        let min_check = min_length.map(|min_val| {
+            quote! {
+                if value_len < #min_val {
+                    return Err(format!("{} must be at least {} characters long. Received {} characters", 
+                        stringify!(#field_name), #min_val, value_len));
+                }
+            }
+        }).unwrap_or_else(|| quote! {});
```
SanchithHegde, line 49: Do you think we should cause a compile time error if we are unable to generate code? Otherwise, we'd end up not knowing about the error and no validations would end up happening.

This applies to here, and other places where you're using `unwrap_or_else()` with an empty `quote! {}` block.

kashif-m: We're returning an empty `quote! {}` block when we don't find `max_length` or `min_length` parameters in the attribute. So this should be expected.

SanchithHegde (approved):

cypress-tests/cypress/e2e/configs/Payment/Commons.js
```diff
@@ -1611,4 +1611,53 @@ export const connectorDetails = {
       },
     },
   },
+  return_url_variations: {
+    "return_url_too_long": getCustomExchange({
+      Request: {
+        customer_id: "customer_1234567890",
+        return_url: "http://example.com/" + "a".repeat(237),
```
pixincreate, line 1618: although this does the job, can we have function call done here?

you can introduce below function in `cypress-tests/cypress/utils/RequestBodyUtils.js` (AI generated BTW). Saying so because it will make this look a lot cleaner and reusable in future as well.

```js
/**
 * Generates a random alphanumeric string of a specified length.
 * An optional prefix can be added, resulting in "prefix_randomstring".
 *
 * @param {number} length - The desired length of the random part of the string.
 * @param {string} [prefix] - An optional prefix for the string.
 * @returns {string} The generated random string.
 *
 * @example
 * generateRandomAlphanumericString(10); // "aK8iPqXzL0" (example output)
 * generateRandomAlphanumericString(5, "user"); // "user_b3Xy1" (example output)
 */
export const generateRandomAlphanumericString = (length, prefix) => {
  let result = '';
  const characters = 'ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789';
  const charactersLength = characters.length;
  for (let i = 0; i < length; i++) {
    result += characters.charAt(Math.floor(Math.random() * charactersLength));
  }
  if (prefix) {
    return `${prefix}_${result}`;
  }
  return result;
};
```

pixincreate: feel free to ignore if you think this is unnecessary to be considered in this pr.

kashif-m: We can do this refactoring in subsequent PRs and issues.

kashif-m: Resolving it for now

PR #8016, opened 2025-05-13
feat(business_profile): add account cache

NishantJoshi00:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

Implemented account cache for the business profile interface, similar to the merchant account caching. This involved modifying the `ProfileInterface` implementation in `crates/router/src/db/business_profile.rs` to use caching for `find_business_profile_by_profile_id`, `update_profile_by_profile_id`, and `delete_profile_by_profile_id_merchant_id` functions. A new helper function `publish_and_redact_business_profile_cache` was added to handle cache invalidation. Added `#[derive(serde::Serialize, serde::Deserialize)]` to the `Profile` struct in `crates/diesel_models/src/business_profile.rs` to enable caching of the diesel model.

### Additional Changes

## Motivation and Context

This change is required to improve the performance of fetching and updating business profile data by utilizing caching. It solves the problem of repeatedly querying the database for frequently accessed business profile information.

## How did you test it?

Verified that the code compiles successfully by running `cargo check`. Further testing with unit/integration tests is recommended to fully verify the caching functionality.

## Checklist

### DIFF
--- a/crates/router/src/db/business_profile.rs
+++ b/crates/router/src/db/business_profile.rs
@@ -1,6 +1,8 @@
 use common_utils::{ext_traits::AsyncExt, types::keymanager::KeyManagerState};
 use error_stack::{report, ResultExt};
 use router_env::{instrument, tracing};
+#[cfg(feature = "accounts_cache")]
+use storage_impl::redis::cache::{self, CacheKind, ACCOUNTS_CACHE};
 
 use super::Store;
 use crate::{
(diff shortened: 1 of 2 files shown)


crates/router/src/db/business_profile.rs
```diff
@@ -106,17 +111,43 @@ impl ProfileInterface for Store {
...
                 key_manager_state,
                 merchant_key_store.key.get_inner(),
                 merchant_key_store.merchant_id.clone().into(),
             )
             .await
             .change_context(errors::StorageError::DecryptionError)
+        }
     }
 
     async fn find_business_profile_by_merchant_id_profile_id(
```
SanchithHegde, line 153: Would we add cache support for this as well, or is it not required?

crates/router/src/db/business_profile.rs
```diff
@@ -147,17 +178,48 @@ impl ProfileInterface for Store {
...
+        #[cfg(feature = "accounts_cache")]
+        {
+            cache::get_or_populate_in_memory(
+                self,
+                &format!(
+                    "{}_{}_{}",
+                    CACHE_KEY_PREFIX,
+                    profile_name,
+                    merchant_id.get_string_repr()
+                ),
```
SanchithHegde, line 197: Nit: We can probably extract this to a function.

Also, are we only using the cache key prefix for this function only, and not when fetching profile by profile ID? Any specific reasons for doing so?

akshay-97: figured profile id should be unique, profile_name and merchant_id could be present in other context as well

crates/router/src/db/business_profile.rs
akshay-97: didnt see this used in oltp flow, can add it here as well

jarnura (approved):

PR #8030, opened 2025-05-14
build(deps): migrate usages of `once_cell` crate to standard library equivalents

SanchithHegde:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [x] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

This PR updates the usages of the `once_cell` crate and its types with their standard library equivalents. This PR replaces:

- [`once_cell::sync::OnceCell`](https://docs.rs/once_cell/latest/once_cell/sync/struct.OnceCell.html) with [`std::sync::OnceLock`](https://doc.rust-lang.org/std/sync/struct.OnceLock.html).
- [`once_cell::sync::Lazy`](https://docs.rs/once_cell/latest/once_cell/sync/struct.Lazy.html) with [`std::sync::LazyLock`](https://doc.rust-lang.org/std/sync/struct.LazyLock.html).

However, this is not a complete migration, since [`once_cell::sync::OnceCell::get_or_try_init()`](https://docs.rs/once_cell/latest/once_cell/sync/struct.OnceCell.html#method.get_or_try_init) cannot be replaced with [`std::sync::OnceLock::get_or_try_init()`](https://doc.rust-lang.org/std/sync/struct.OnceLock.html#method.get_or_try_init) as it is not available in the stable toolchain yet, as of Rust 1.86. Due to this reason, `once_cell` is still kept as a dependency in the `common_utils`, `external_services` and `router` crates, only where the `get_or_try_init()` method is being used.

P.S.: This code was mostly written by AI, which I've also reviewed along the way.

## Motivation and Context

This PR just replaces usages of one of our dependencies to the standard library equivalents.

## How did you test it?

1. Performed sanity testing via Postman. The expected result is that there are no behavior changes due to this PR.

2. I have run unit tests on the `euclid` and `common_utils` crates, all of them (from these two crates) passed. Some of these unit tests make use of the static variables that were migrated in this PR:

   - https://github.com/juspay/hyperswitch/blob/5fd6df73c9adaa454402c9b79ba37eff9f4c3ffc/crates/common_utils/src/validation.rs#L111-L120
     - This test exercises the `EMAIL_REGEX` static variable.
   - https://github.com/juspay/hyperswitch/blob/5fd6df73c9adaa454402c9b79ba37eff9f4c3ffc/crates/euclid/src/dssa/analyzer.rs#L258-L377
     - These tests call `analyze` without providing a specific graph, thus using `ANALYSIS_GRAPH`.

   Screenshot of unit test run:

   Screenshot of unit test run

3. I have confirmed that cache system works correctly by adding a log line in case of in-memory cache hits (this log line was added only for testing and is not included in the PR). The in-memory caches also make use of static variables and `LazyLock`, so their correct functioning indicates that the migration hasn't broken the cache system.

   Screenshot of log lines indicating in-memory cache hits

## Checklist


jarnura (approved):

ThisIsMani (approved):

tsdk02 (approved):

pixincreate (approved):
Cypress, LGTM.

ShankarSinghC (approved):

PR #8100, opened 2025-05-21
refactor(cypress): fix `braintree` redirection handling assertions

arindam-sahoo:
## Type of Change

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [x] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [x] CI/CD

## Description

Handling Redirection was failing on Cypress after Confirming 3DS Payments for `Braintree` Connector. It was looking for a iframe, and the assertion was failing. Actually, the `currentHost` and `originalHost` being the same in the handleFlow function (which should not have been the case), it asserts that an iframe exists which does not.

### Additional Changes

## Motivation and Context

Cypress Tests for `Braintree` were failing.

### Before
<details><summary>Click to see Image</summary>

image

</details> 

### After
image

## Checklist


cypress-tests/cypress/e2e/configs/Payment/Braintree.js
```diff
@@ -143,6 +143,9 @@ export const connectorDetails = {
       },
     },
     "3DSAutoCapture": {
+      Configs: {
+        TRIGGER_SKIP: true,
+      },
```
pixincreate, line 148: why skipping? unless it is really not possible to achieve the redirection this should be used. i would suggest you modify the redirectionHandler instead and add support for braintree 3ds instead.

PR #8166, opened 2025-05-28
refactor(router): Separate fallback from success based routing

GauravRawat369:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

Remove `perform_eligibility_analysis` and  `perform_fallback_routing` from `perform_eligibility_analysis_with_fallback`
Added fallback routing after getting connectors from **Dynamo**

### Additional Changes

## How did you test it?

1. Created 2 mac and added them to the fallback
2. Took Cybersource and created a static rule (where `country = NE`)
3. And checked if the fallback is sent to Dynamo or not

- Added Stripe as a fallback. 

Screenshot 2025-05-29 at 3 36 17 PM

- Fallback has not been sent to Dynamo.

Screenshot 2025-05-29 at 3 34 52 PM

- Created 2 mca accounts, only 1 mca received from `fetch_sr`.

Screenshot 2025-06-09 at 12 34 50 PM

- What we received from Dynamo in `route_connector_v1_for_payments`

Screenshot 2025-06-09 at 12 30 15 PM

- List of connectors after adding a fallback to Dynamo connectors

Screenshot 2025-06-09 at 2 29 04 PM

## Checklist


prajjwalkumar17 (commented):
Changes seems fine, can you once add the test case for fetch success rate. As in update we always send only one connector, let's check for fetch_success_rate and what connectors are sent by dynamo. So the flow will be:
1. Create 2 mca accounts.
2. Create a static rule and toggle SR, with something like.
>>> country = NE -> mca1
3. Check what is recieved from fetch_sr. (technically it should only be mca1 and not both)

PR #8178, opened 2025-05-29
feat(gRPC): build gRPC client interface to initiate communication with recovery-decider service

AdityaKumaar21:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

I have implemented a gRPC client that facilitates communication with the recovery-decider gRPC service. The client should be capable of initializing and managing the connection to the service, sending requests, and handling responses in accordance with the defined service contract (protobuf definitions).

### Additional Changes

## How did you test it?

Decider gRPC server:-
Screenshot 2025-06-11 at 01 16 34

Decider gRPC client(HS):-
Screenshot 2025-06-05 at 02 05 20

Predictor gRPC server:-
Screenshot 2025-06-05 at 02 03 06

## Checklist

## Summary by CodeRabbit

- **New Features**
  - Introduced support for a gRPC client: Recovery Decider, enabling external retry decisioning.
  - Integrated dynamic scheduling for smart retry logic based on external gRPC service responses.
  - Added new protobuf definition for Recovery Decider.

- **Improvements**
  - Enhanced revenue recovery payment flow with richer payment intent context and smarter retry scheduling.
  - Expanded logging and API flow tracking with new flow types.

- **Configuration**
  - Added configuration options for new gRPC clients in the development environment.

- **Bug Fixes**
  - Improved error handling and logging for gRPC client interactions.


config/development.toml
```diff
@@ -1145,6 +1145,10 @@ host = "localhost"
 port = 8000
 service = "dynamo"
 
+[grpc_client.recovery_decider_client]
+host = "127.0.0.1"
+port = 8000
```
Aprabhat19: Can you change the port number

crates/common_utils/Cargo.toml
```diff
@@ -41,6 +41,7 @@ nutype = { version = "0.4.3", features = ["serde"] }
 once_cell = "1.21.3"
 openssl = {version = "0.10.72", optional = true}
 phonenumber = "0.3.7"
+prost-types = { version = "0.13" }
```
Aprabhat19:
```suggestion
prost-types  = "0.13"
```

crates/external_services/build.rs
```diff
@@ -1,29 +1,47 @@
...
+        proto_files_to_compile.push(proto_base_path.join("elimination_rate.proto"));
+        proto_files_to_compile.push(proto_base_path.join("health_check.proto"));
+    }
+
+    #[cfg(feature = "v2")]
+    {
+        proto_files_to_compile.push(proto_base_path.join("recovery_decider.proto"));
+    }
+
+    if !proto_files_to_compile.is_empty() {
```
Aprabhat19: Can you build the files separately , doing that we might not require checks

crates/external_services/src/grpc_client.rs
```diff
@@ -37,13 +43,20 @@ pub struct GrpcClients {
     /// Health Check client for all gRPC services
     #[cfg(feature = "dynamic_routing")]
     pub health_client: HealthCheckClient,
+    #[cfg(feature = "v2")]
+    #[allow(missing_docs)]
```
Aprabhat19: Add documentations

crates/external_services/src/grpc_client.rs
```diff
@@ -68,15 +82,29 @@ impl GrpcClientSettings {
...
+        let health_client = HealthCheckClient::build_connections(self, client.clone())
             .await
             .expect("Failed to build gRPC connections");
 
+        #[cfg(feature = "v2")]
+        let recovery_decider_client = self
+            .recovery_decider_client
+            .get_recovery_decider_connection(client.clone())
+            .map(|client| {
+                #[allow(clippy::as_conversions)]
```
Aprabhat19: Remove this

crates/external_services/src/grpc_client.rs
```diff
@@ -129,12 +157,12 @@ impl<T> AddHeaders for tonic::Request<T> {
     }
 }
 
-#[cfg(feature = "dynamic_routing")]
+#[cfg(any(feature = "dynamic_routing", feature = "v2"))]
```
Aprabhat19: We don't need to pass headers, so can you revert the changes

crates/external_services/src/grpc_client/recovery_decider_client.rs
```diff
@@ -0,0 +1,171 @@
...
+    tonic::include_proto!("decider");
+}
+
+use common_utils::custom_serde::prost_timestamp::SerializableTimestamp;
+use decider::decider_client::DeciderClient;
+pub use decider::DeciderRequest;
+
+/// Represents the response from the decider service, suitable for HTTP JSON serialization.
+#[derive(Debug, Clone, serde::Serialize, serde::Deserialize)]
+pub struct DeciderResponseForSerde {
```
Aprabhat19: Why do we need this?

crates/external_services/src/grpc_client/recovery_decider_client.rs
```diff
@@ -0,0 +1,171 @@
...
+}
+
+#[async_trait::async_trait]
+#[allow(missing_docs)]
+#[allow(clippy::too_many_arguments)]
+pub trait RecoveryDeciderClientInterface:
+    dyn_clone::DynClone + Send + Sync + std::fmt::Debug
+{
+    #[allow(missing_docs)]
+    #[allow(clippy::too_many_arguments)]
```
Aprabhat19: Remove these , and add docs instead

crates/external_services/src/grpc_client/recovery_decider_client.rs
```diff
@@ -0,0 +1,171 @@
...
+    async fn decide_on_retry(
+        &mut self,
+        first_error_message: String,
+        billing_state: String,
+        card_funding: String,
+        card_network: String,
+        card_issuer: String,
+        start_time: Option<prost_types::Timestamp>,
+        end_time: Option<prost_types::Timestamp>,
```
Aprabhat19: Try to move these into a struct and send the structure as a parameter.

crates/external_services/src/grpc_client/recovery_decider_client.rs
```diff
@@ -0,0 +1,171 @@
...
+            return Err(Report::new(RecoveryDeciderError::ConfigError(
+                "Host is not configured for Recovery Decider client".to_string(),
+            )));
+        }
+
+        let uri_string = format!("http://{}:{}", host, port);
+        let uri = uri_string
+            .parse::<tonic::transport::Uri>()
+            .map_err(Report::from)
+            .change_context_lazy(|| {
```
Aprabhat19: Use change context

crates/external_services/src/grpc_client/recovery_decider_client.rs
```diff
@@ -0,0 +1,171 @@
...
+            first_error_message,
+            billing_state,
+            card_funding,
+            card_network,
+            card_issuer,
+            start_time,
+            end_time,
+            retry_count,
+        };
+        let request = grpc_client::create_grpc_request(request_data, headers);
```
Aprabhat19: We do not need to pass any headers now,  and even if we do require header our requirement might differ, so can you not reuse that and build a separate structure to construct headers and send in the request

crates/external_services/src/grpc_client/recovery_decider_client.rs
```diff
@@ -0,0 +1,171 @@
...
+            .map_err(|status| {
+                logger::error!(grpc_error =?status, "Decider service call failed");
+                RecoveryDeciderError::ServiceError(status.message().to_string())
+            })?
+            .into_inner();
+
+        logger::debug!(grpc_decider_response =?grpc_response);
+
+        // Map to our Serde-compatible struct
+        let response_for_serde = DeciderResponseForSerde {
```
Aprabhat19: Why do we need to do this? Can you not send the response built by tonic

crates/external_services/src/grpc_client/recovery_decider_client.rs
```diff
@@ -0,0 +1,171 @@
...
+            retry_flag: grpc_response.retry_flag,
+            retry_time: grpc_response.retry_time.map(SerializableTimestamp::from),
+        };
+
+        Ok(response_for_serde)
+    }
+}
+
+impl common_utils::events::ApiEventMetric for DeciderResponseForSerde {
```
Aprabhat19: You might not need this

crates/external_services/src/grpc_client/recovery_decider_client.rs
```diff
@@ -0,0 +1,171 @@
...
+    fn get_api_event_type(&self) -> Option<common_utils::events::ApiEventsType> {
+        Some(common_utils::events::ApiEventsType::Miscellaneous)
+    }
+}
+
+impl common_utils::events::ApiEventMetric for DeciderRequest {
```
Aprabhat19: Same for this

crates/external_services/src/grpc_client/recovery_decider_client.rs
```diff
@@ -0,0 +1,171 @@
...
+    clippy::as_conversions,
+    clippy::use_self
+)]
+pub mod decider {
+    tonic::include_proto!("decider");
+}
+
+use common_utils::custom_serde::prost_timestamp::SerializableTimestamp;
+use decider::decider_client::DeciderClient;
+pub use decider::DeciderRequest;
```
Aprabhat19: Can you use the decider request built by tonic, instead of creating a new one

crates/router_env/src/logger/types.rs
```diff
@@ -180,6 +180,8 @@ pub enum Flow {
     PaymentsConfirmIntent,
     /// Payments create and confirm intent flow
     PaymentsCreateAndConfirmIntent,
+    /// Payments Recovery Should Retry Test
+    RecoveryDeciderShouldRetry,
```
Aprabhat19: why do we have this type?

crates/external_services/src/grpc_client/recovery_decider_client.rs
AdityaKumaar21: there was issue in the serialization of Timestamp,so thats why i have created this to handle the retry_time field, which is a prost_types::Timestamp in the gRPC response. DeciderResponseForSerde uses a custom SerializableTimestamp type to ensure this timestamp is consistently serialized as a JSON object with seconds and nanos fields,

crates/external_services/src/grpc_client/recovery_decider_client.rs
AdityaKumaar21: The gRPC response is mapped to `DeciderResponseForSerde` primarily to ensure the retry_time (`prost_types::Timestamp`) is consistently serialized into a specific JSON object with seconds and nanos fields, using the SerializableTimestamp wrapper.

Aprabhat19: You can convert it wherever you are  consuming timestamps

crates/api_models/src/payments.rs
```diff
@@ -8713,7 +8713,7 @@ pub struct PaymentRevenueRecoveryMetadata {
 
 #[cfg(feature = "v2")]
 #[derive(Debug, Clone, PartialEq, Deserialize, Serialize, ToSchema)]
-#[serde(rename_all = "snake_case", tag = "type", content = "value")]
```
Aprabhat19: Why are we removing it from snake_case

crates/common_utils/src/lib.rs
```diff
@@ -203,6 +204,28 @@ pub mod date_time {
             f.write_str(&output)
         }
     }
+
+    /// Converts a `time::PrimitiveDateTime` to a `prost_types::Timestamp`.
+    /// Assumes UTC for the conversion.
+    pub fn convert_to_prost_timestamp(dt: PrimitiveDateTime) -> Timestamp {
```
Aprabhat19: Can we do this conversion in external_services itself?

crates/external_services/build.rs
```diff
@@ -1,29 +1,55 @@
 #[allow(clippy::expect_used)]
 fn main() -> Result<(), Box<dyn std::error::Error>> {
-    #[cfg(feature = "dynamic_routing")]
+    #[cfg(any(feature = "v2", feature = "dynamic_routing"))]
```
Aprabhat19: Can we separate out the flags v2 and dynamic routing changes?

crates/external_services/build.rs
```diff
@@ -1,29 +1,55 @@
...
+        // Compilation for dynamic_routing protos
+        #[cfg(feature = "dynamic_routing")]
+        {
+            let dr_proto_files = [
+                proto_base_path.join("success_rate.proto"),
+                proto_base_path.join("contract_routing.proto"),
+                proto_base_path.join("elimination_rate.proto"),
+                proto_base_path.join("health_check.proto"),
+            ];
+            if !dr_proto_files.is_empty() {
```
Aprabhat19: Why do we need this?

crates/external_services/build.rs
```diff
@@ -1,29 +1,49 @@
...
-                &[proto_path],
+            .out_dir(&out_dir)
+            .compile_well_known_types(true)
+            .extern_path(".google.protobuf.Timestamp", "::prost_types::Timestamp")
+            .type_attribute(
+                "google.protobuf.Timestamp",
+                "#[derive(serde::Serialize, serde::Deserialize)]",
             )
-            .expect("Failed to compile proto files");
+            .compile(&dr_proto_files, &[&proto_base_path])?;
```
Aprabhat19: Lets not propagate an error here, we want these files to be compiled to set up a client connection

crates/external_services/Cargo.toml
```diff
@@ -26,6 +36,7 @@ dynamic_routing = [
     "dep:hyper-util",
     "dep:http-body-util",
 ]
+revenue_recovery = []
```
Aprabhat19: We already have created a feature flag named revenue_recovery can you use that?

crates/external_services/build.rs
```diff
@@ -1,29 +1,49 @@
...
-                    elimination_proto_file,
-                    contract_routing_proto_file,
-                ],
-                &[proto_path],
+            .out_dir(&out_dir)
+            .compile_well_known_types(true)
+            .extern_path(".google.protobuf.Timestamp", "::prost_types::Timestamp")
+            .type_attribute(
+                "google.protobuf.Timestamp",
+                "#[derive(serde::Serialize, serde::Deserialize)]",
```
Aprabhat19: Can you remove this from here

Aprabhat19:
Can you add the configs in other toml files as well

crates/external_services/src/lib.rs
```diff
@@ -28,6 +28,35 @@ pub mod managers;
...
 pub mod crm;
 
+/// date_time module
+pub mod date_time {
+    use prost_types::Timestamp;
+    // use masking::{Deserialize, Serialize};
+    use time::{OffsetDateTime, PrimitiveDateTime};
+
+    /// Converts a `time::PrimitiveDateTime` to a `prost_types::Timestamp`.
+    /// Assumes UTC for the conversion.
```
Aprabhat19: Remove this line of comment

crates/external_services/src/grpc_client/revenue_recovery/recovery_decider_client.rs
```diff
@@ -0,0 +1,122 @@
...
+impl RecoveryDeciderClientInterface for DeciderClient<Client> {
+    /// collects the request from HS and sends it to recovery decider gRPC service
+    async fn decide_on_retry(
+        &mut self,
+        request_payload: DeciderRequest,
+        recovery_headers: super::common::GrpcRecoveryHeaders,
+    ) -> RecoveryDeciderResult<DeciderResponse> {
+        let request =
+            super::common::create_revenue_recovery_grpc_request(request_payload, recovery_headers);
+        logger::info!(message_sent_to_recovery_decider_server=?request); // Client-specific logging
```
Aprabhat19: Remove this as you are logging  request twice

crates/common_utils/Cargo.toml
```diff
@@ -41,6 +41,7 @@ nutype = { version = "0.4.3", features = ["serde"] }
 once_cell = "1.21.3"
 openssl = {version = "0.10.72", optional = true}
 phonenumber = "0.3.7"
+prost-types = "0.13"
```
SanchithHegde: Why did we add this dependency here? I don't see any other changes to the `common_utils` crate?

crates/external_services/Cargo.toml
```diff
@@ -47,6 +58,8 @@ serde = { version = "1.0.219", features = ["derive"] }
 thiserror = "1.0.69"
 vaultrs = { version = "0.7.4", optional = true }
 prost = { version = "0.13", optional = true }
+prost-types = "0.13"
```
SanchithHegde: Can we make this dependency an optional one?

crates/external_services/Cargo.toml
```diff
@@ -13,6 +13,16 @@ email = ["dep:aws-config"]
...
 v1 = ["hyperswitch_interfaces/v1", "common_utils/v1"]
+v2 = [
+    "dep:prost",
+    "dep:tonic",
+    "dep:tonic-build",
+    "dep:router_env",
+    "tokio/macros",
+    "tokio/rt-multi-thread",
+    "dep:hyper-util",
+    "dep:http-body-util"
```
SanchithHegde: Why are we enabling all of these when `v2` feature is enabled, shouldn't we be enabling these only when the `revenue_recovery` feature is enabled?

crates/external_services/build.rs
```diff
@@ -1,29 +1,43 @@
...
-        let success_rate_proto_file = proto_path.join("success_rate.proto");
-        let contract_routing_proto_file = proto_path.join("contract_routing.proto");
-        let elimination_proto_file = proto_path.join("elimination_rate.proto");
-        let health_check_proto_file = proto_path.join("health_check.proto");
+        let proto_base_path = router_env::workspace_path().join("proto");
         let out_dir = std::path::PathBuf::from(std::env::var("OUT_DIR")?);
-
-        // Compile the .proto file
+        let v2_proto_files = [proto_base_path.join("recovery_decider.proto")];
+        println!("Compiling v2 proto files: {:?}", v2_proto_files);
```
SanchithHegde: Does the build script even print out any output that's visible to us? I believe we can't see this line being printed (unless we enable verbose cargo logs)?

crates/external_services/build.rs
```diff
@@ -1,29 +1,43 @@
...
-                    elimination_proto_file,
-                    contract_routing_proto_file,
-                ],
-                &[proto_path],
+            .out_dir(&out_dir)
+            .compile_well_known_types(true)
+            .extern_path(".google.protobuf.Timestamp", "::prost_types::Timestamp")
+            .type_attribute(
+                "google.protobuf.Timestamp",
+                "#[derive(serde::Serialize, serde::Deserialize)]",
```
SanchithHegde: What format would the timestamp be serialized with: ISO 8601 or something else? Asking because we've used ISO 8601 format throughout our codebase.

crates/external_services/src/grpc_client.rs
```diff
@@ -130,7 +169,7 @@ pub(crate) fn create_grpc_request<T: Debug>(message: T, headers: GrpcHeaders) ->
     let mut request = tonic::Request::new(message);
     request.add_headers_to_grpc_request(headers);
 
-    logger::info!(dynamic_routing_request=?request);
+    logger::info!(request=?request);
```
SanchithHegde: Nit: This can be simplified.

```suggestion
    logger::info!(?request);
```

crates/external_services/src/grpc_client/revenue_recovery.rs
```diff
@@ -0,0 +1,6 @@
+#![cfg(all(feature = "revenue_recovery", feature = "v2"))]
```
SanchithHegde: Nit: Why do we need this annotation, considering you've added a similar one in `crates/external_services/src/grpc_client.rs` above the module declaration?

crates/external_services/src/grpc_client/revenue_recovery.rs
```diff
@@ -0,0 +1,6 @@
...
+
+/// common file for revenue recovery
+pub mod common;
```
SanchithHegde: You could have kept the common code in this file (`revenue_recovery.rs`) itself, instead of in a separate module.

crates/external_services/src/grpc_client/revenue_recovery/common.rs
```diff
@@ -0,0 +1,56 @@
+#![cfg(all(feature = "revenue_recovery", feature = "v2"))]
```
SanchithHegde: Again, why is this annotation needed?

crates/external_services/src/grpc_client/revenue_recovery/common.rs
```diff
@@ -0,0 +1,56 @@
...
+    fn add_recovery_headers(&mut self, headers: GrpcRecoveryHeaders);
+}
+
+impl<T> AddRecoveryHeaders for tonic::Request<T> {
+    #[track_caller]
+    fn add_recovery_headers(&mut self, headers: GrpcRecoveryHeaders) {
+        headers.request_id.map(|request_id| {
+            request_id
+                .parse()
+                .map(|request_id_val| { // Renamed to avoid conflict with outer scope
```
SanchithHegde: Nit: Can we remove this comment?

crates/external_services/src/grpc_client/revenue_recovery/recovery_decider_client.rs
```diff
@@ -0,0 +1,121 @@
+#![cfg(all(feature = "revenue_recovery", feature = "v2"))]
```
SanchithHegde: Again, why the annotation?

crates/external_services/src/grpc_client/revenue_recovery/recovery_decider_client.rs
```diff
@@ -0,0 +1,121 @@
...
+
+use std::fmt::Debug;
+
+use common_utils::errors::CustomResult;
+use error_stack::{Report, ResultExt};
+use router_env::logger;
+use serde;
```
SanchithHegde: Nit: This import is useless, can be removed.

crates/external_services/src/grpc_client/revenue_recovery/recovery_decider_client.rs
```diff
@@ -0,0 +1,121 @@
...
+
+use decider::decider_client::DeciderClient;
+pub use decider::{DeciderRequest, DeciderResponse};
+
+/// Recovery Decider result
+pub type RecoveryDeciderResult<T> = CustomResult<T, RecoveryDeciderError>;
+
+#[allow(missing_docs)]
+#[derive(Debug, Clone, thiserror::Error)]
+pub enum RecoveryDeciderError {
```
SanchithHegde: You've already added doc comments for all enum variants, might as well add a doc comment for the enum itself and remove the `allow(missing_docs)` annotation.

crates/external_services/src/grpc_client/revenue_recovery/recovery_decider_client.rs
```diff
@@ -0,0 +1,121 @@
...
+dyn_clone::clone_trait_object!(RecoveryDeciderClientInterface);
+
+/// Configuration for the Recovery Decider gRPC client.
+#[derive(Debug, Default, Clone, serde::Deserialize, serde::Serialize)]
+pub struct RecoveryDeciderClientConfig {
+    /// Host
+    pub host: String,
+    /// port number
+    pub port: u16,
+}
```
SanchithHegde: Let's please accept a base URL alone, we can still send requests without specifying a port number. I've discussed this with @Aprabhat19 in a separate communication channel.

crates/external_services/src/lib.rs
```diff
@@ -28,6 +28,44 @@ pub mod managers;
 /// crm module
 pub mod crm;
 
+#[cfg(feature = "v2")]
+/// date_time module
+pub mod date_time {
```
SanchithHegde: This module is no longer specific to the `v2` feature flag, considering that you're importing the `grpc_client::revenue_recovery` module here?

Can we use correct feature flags here please?

crates/external_services/src/lib.rs
```diff
@@ -28,6 +28,44 @@ pub mod managers;
...
+    use prost_types::Timestamp;
+    // use masking::{Deserialize, Serialize};
```
SanchithHegde: Remove commented code.

crates/external_services/src/lib.rs
```diff
@@ -28,6 +28,44 @@ pub mod managers;
...
+
+        OffsetDateTime::from_unix_timestamp_nanos(timestamp_nanos)
+            .map(|offset_dt| PrimitiveDateTime::new(offset_dt.date(), offset_dt.time()))
+            .map_err(|original_error| {
+                router_env::logger::debug!(
+                    "Prost timestamp conversion failed: {:?}",
+                    original_error
+                );
+                common::DateTimeConversionError::TimestampOutOfRange
+            })
```
SanchithHegde: 1. Nit: Inline format args, clippy with Rust 1.88 warns about it.

   ```suggestion
               .map_err(|original_error| {
                   router_env::logger::debug!(
                       "Prost timestamp conversion failed: {original_error:?}"
                   );
                   common::DateTimeConversionError::TimestampOutOfRange
               })
   ```

2. Can we consider logging it at `error` level instead?

OR

3. Can we consider using `change_context()` here instead of `map_err()` and return an `error_stack::Result` / `CustomResult` from this function?

crates/router/Cargo.toml
```diff
@@ -59,6 +59,7 @@ blake3 = "1.8.2"
 bytes = "1.10.1"
 clap = { version = "4.5.38", default-features = false, features = ["std", "derive", "help", "usage"] }
 config = { version = "0.14.1", features = ["toml"] }
+prost-types = "0.13"
```
SanchithHegde: Can you make this dependency optional and enable it when the necessary feature is enabled?

crates/router/src/core/revenue_recovery/types.rs
```diff
@@ -473,13 +477,13 @@ impl Action {
...
         revenue_recovery_payment_data: &storage::revenue_recovery::RevenueRecoveryPaymentData,
-        global_payment_id: &id_type::GlobalPaymentId,
+        payment_intent: PaymentIntent,
         process: &storage::ProcessTracker,
         payment_attempt: payment_attempt::PaymentAttempt,
     ) -> RecoveryResult<Self> {
         let response = revenue_recovery_core::api::call_psync_api(
             state,
-            global_payment_id,
+            payment_intent.get_id(),
```
SanchithHegde: Why was this change made at all? Can we revert it?

crates/router/src/core/revenue_recovery/types.rs
```diff
@@ -684,15 +690,24 @@ impl Action {
     }
 
     pub(crate) async fn decide_retry_failure_action(
+        state: &SessionState,
         db: &dyn StorageInterface,
         merchant_id: &id_type::MerchantId,
         pt: storage::ProcessTracker,
         revenue_recovery_payment_data: &storage::revenue_recovery::RevenueRecoveryPaymentData,
         payment_attempt: &payment_attempt::PaymentAttempt,
+        payment_intent: &PaymentIntent,
```
SanchithHegde: Nit: There's inconsistency in imports, either fully import both payment attempt and payment intent, or qualify both. I'd prefer qualifying both.

crates/router/src/types/storage/revenue_recovery.rs
```diff
@@ -29,9 +34,12 @@ pub struct RevenueRecoveryPaymentData {
 impl RevenueRecoveryPaymentData {
     pub async fn get_schedule_time_based_on_retry_type(
         &self,
+        state: &SessionState,
         db: &dyn StorageInterface,
```
SanchithHegde: Why do you need both `state` and `db` to be passed?

Either accept only `state`, or accept `db` and the gRPC client as parameters.

crates/router/src/workflows/revenue_recovery.rs
```diff
@@ -200,3 +214,159 @@ pub(crate) async fn get_schedule_time_to_retry_mit_payments(
...
+        .error
+        .as_ref()
+        .map_or(String::new(), |error| error.message.clone());
+
+    let billing_state = payment_intent
+        .billing_address
+        .as_ref()
+        .and_then(|addr_enc| addr_enc.get_inner().address.as_ref())
+        .and_then(|details| details.state.as_ref())
+        .map(|state_from_address| state_from_address.peek().clone())
```
SanchithHegde: We need to consider using `Secret` for gRPC request fields with sensitive information, instead of using raw strings. We can refer to how the connector-service folks are integrating it.

crates/router/src/workflows/revenue_recovery.rs
```diff
@@ -200,3 +214,159 @@ pub(crate) async fn get_schedule_time_to_retry_mit_payments(
...
+            _ => None,
+        })
+        .map(|cn| cn.to_string());
+
+    let card_network_str = match card_network_opt {
+        Some(cn_str) => cn_str,
+        None => {
+            return None;
+        }
+    };
```
SanchithHegde: Just use `?` instead of matching on `card_network_opt`?

crates/router/src/workflows/revenue_recovery.rs
```diff
@@ -200,3 +214,159 @@ pub(crate) async fn get_schedule_time_to_retry_mit_payments(
...
+            payment_metadata
+                .billing_connector_payment_method_details
+                .as_ref()
+        })
+        .and_then(|details| match details {
+            BillingConnectorPaymentMethodDetails::Card(card_info) => card_info.card_issuer.clone(),
+            _ => None,
+        })
+        // will remove this after we are getting the card issuer from Stripe
+        .unwrap_or("CHASE".to_string());
```
SanchithHegde: Ummm, hardcoding things like this is a bad idea, if we forget to remove them later.

Can we consider returning an error from this function maybe?

crates/router/src/workflows/revenue_recovery.rs
```diff
@@ -200,3 +214,159 @@ pub(crate) async fn get_schedule_time_to_retry_mit_payments(
...
+                .as_ref()
+        })
+        .map(|payment_metadata| payment_metadata.payment_method_subtype.to_string());
+
+    let card_funding_str = match card_funding_opt {
+        Some(cf_str) => cf_str,
+        None => {
+            return None;
+        }
+    };
```
SanchithHegde: Same here, you can just use `?`.

crates/router/src/workflows/revenue_recovery.rs
```diff
@@ -200,3 +214,159 @@ pub(crate) async fn get_schedule_time_to_retry_mit_payments(
...
+    }
+
+    let card_network_opt = payment_intent
+        .feature_metadata
+        .as_ref()
+        .and_then(|revenue_recovery_data| {
+            revenue_recovery_data
+                .payment_revenue_recovery_metadata
+                .as_ref()
+        })
```
SanchithHegde: You seem to be accessing `payment_intent.feature_metadata.payment_revenue_recovery_metadata` multiple times in this function. Consider assigning it to a variable once and using it later?

crates/router/src/workflows/revenue_recovery.rs
```diff
@@ -200,3 +214,159 @@ pub(crate) async fn get_schedule_time_to_retry_mit_payments(
...
+
+    let start_time_primitive = payment_intent.created_at;
+    // 1 hr after the actual start time
+    let modified_start_time_primitive =
+        start_time_primitive.saturating_add(time::Duration::hours(1));
+    let start_time_proto = date_time::convert_to_prost_timestamp(modified_start_time_primitive);
+
+    // Calculate end_time as start_time + 30 days
+    let end_time_primitive = start_time_primitive.saturating_add(time::Duration::days(30));
+    let end_time_proto = date_time::convert_to_prost_timestamp(end_time_primitive);
```
SanchithHegde: Instead of hardcoding the start and end times, consider reading this from the `configs` table, and falling back to a default config otherwise?

crates/router/src/workflows/revenue_recovery.rs
```diff
@@ -200,3 +214,159 @@ pub(crate) async fn get_schedule_time_to_retry_mit_payments(
...
+
+    logger::debug!(
+        payment_intent_id = ?payment_intent.get_id(),
+        attempt_id = ?payment_attempt.get_id(),
+        message = "Card details for Recovery DeciderRequest",
+        first_error_message = %first_error_message,
+        billing_state = %billing_state,
```
SanchithHegde: State _may_ be considered PII, considering that it was behind a `Secret`...

crates/router/src/workflows/revenue_recovery.rs
```diff
@@ -200,3 +214,159 @@ pub(crate) async fn get_schedule_time_to_retry_mit_payments(
...
+        card_funding: card_funding_str,
+        card_network: card_network_str,
+        card_issuer: card_issuer_str,
+        start_time: Some(start_time_proto),
+        end_time: Some(end_time_proto),
+        retry_count: retry_count.into(),
+    };
+
+    // Clone the gRPC client into a mutable local variable, this is necessary because `decide_on_retry` requires `&mut self`,
+    // and we need an owned, mutable instance to call it.
```
SanchithHegde: Why does `decide_on_retry` require `&mut self`?

crates/router/src/workflows/revenue_recovery.rs
```diff
@@ -200,3 +214,159 @@ pub(crate) async fn get_schedule_time_to_retry_mit_payments(
...
+            .then_some(())
+            .and(grpc_response.retry_time)
+            .and_then(|prost_ts| {
+                match date_time::convert_from_prost_timestamp(&prost_ts) {
+                    Ok(pdt) => Some(pdt),
+                    Err(e) => {
+                        logger::error!(
+                            "Failed to convert retry_time from prost::Timestamp: {:?}",
+                            e
+                        );
```
SanchithHegde: Nit: Clippy 1.88 would throw a warning.

```suggestion
                        logger::error!(
                            "Failed to convert retry_time from prost::Timestamp: {e:?}"
                        );
```

crates/router/src/workflows/revenue_recovery.rs
```diff
@@ -200,3 +214,159 @@ pub(crate) async fn get_schedule_time_to_retry_mit_payments(
...
+                        None // If conversion fails, treat as no valid retry time
+                    }
+                }
+            }),
+
+        Err(e) => {
+            logger::error!("Recovery decider gRPC call failed: {:?}", e);
```
SanchithHegde: Nit: Clippy 1.88 would throw a warning.

```suggestion
            logger::error!("Recovery decider gRPC call failed: {e:?}");
```

crates/external_services/src/grpc_client.rs
```diff
@@ -33,13 +52,20 @@ pub struct GrpcClients {
     /// Health Check client for all gRPC services
     #[cfg(feature = "dynamic_routing")]
     pub health_client: HealthCheckClient,
+    /// Recovery Decoder Client
```
SanchithHegde: Nit: Typo

```suggestion
    /// Recovery Decider Client
```

crates/external_services/src/grpc_client.rs
```diff
@@ -33,13 +52,20 @@ pub struct GrpcClients {
...
+    #[cfg(all(feature = "revenue_recovery", feature = "v2"))]
+    pub recovery_decider_client: Box<dyn RecoveryDeciderClientInterface>,
```
SanchithHegde: @srujanchikke @Aprabhat19 How are we planning to include this feature in our release builds / images?

Seeing that the `revenue_recovery` feature is enabled when `router`'s `v2` feature is enabled, this feature would be included in our release builds. Because this feature is enabled in our v2 release builds, we would also need a way to toggle this feature at runtime, in case we want to disable it for certain environments of our own, or for anyone else self hosting Hyperswitch.

And if we have a runtime toggle, this client would have to be an optional field...

config/development.toml
SanchithHegde, line 1: Please replicate these changes in the `config/config.example.toml` and `config/deployments/env_specific.toml` files, with documentation comments in both files.

crates/external_services/src/grpc_client.rs
Aprabhat19: If this feature has been enabled we expect the server to be up and running , that is why we placed it under the compile time feature flag for the first time so.

crates/router/src/core/revenue_recovery/types.rs
AdityaKumaar21: i needed `payment_intent` in `decide_retry_failure_action`, so thats why removed the `global_payment_id` and propagated the pi from top, also saved the additional db call to get pi from the `global_payment_id` for `decide_retry_failure_action`.

crates/external_services/build.rs
AdityaKumaar21: it will serialize into a JSON object with seconds and nanos fields...

crates/router/src/workflows/revenue_recovery.rs
AdityaKumaar21: i am not hardcoding them, for the start time, i am getting it from pi.created_at and just adding 1 hr to it, and end_time for our use case will be 30 days from start time.

config/deployments/env_specific.toml
```diff
@@ -354,6 +354,9 @@ host = "localhost" # Client Host
 port = 7000        # Client Port
 service = "dynamo" # Service name
 
+[grpc_client.recovery_decider_client]
```
SanchithHegde: Could you add a comment explaining what this section is for? Please add the same comment in `config/config.example.toml` as well.

crates/external_services/Cargo.toml
```diff
@@ -13,6 +13,16 @@ email = ["dep:aws-config"]
...
 v1 = ["hyperswitch_interfaces/v1", "common_utils/v1"]
+v2 = ["hyperswitch_interfaces/v2", "common_utils/v2"]
+revenue_recovery = [
+    "dep:prost",
+    "dep:router_env",
+    "tokio/macros",
+    "tokio/rt-multi-thread",
+    "dep:hyper-util",
+    "dep:http-body-util",
+    "prost-types",
```
SanchithHegde: Nit:

```suggestion
    "dep:prost-types",
```

crates/external_services/build.rs
```diff
@@ -1,9 +1,27 @@
 #[allow(clippy::expect_used)]
 fn main() -> Result<(), Box<dyn std::error::Error>> {
+    // Compilation for v2 protos
+    #[cfg(feature = "revenue_recovery")]
```
SanchithHegde: Nit:


```suggestion
    // Compilation for revenue recovery protos
    #[cfg(feature = "revenue_recovery")]
```

crates/external_services/build.rs
```diff
@@ -1,9 +1,27 @@
...
+    {
+        let proto_base_path = router_env::workspace_path().join("proto");
+        let out_dir = std::path::PathBuf::from(std::env::var("OUT_DIR")?);
+        let v2_proto_files = [proto_base_path.join("recovery_decider.proto")];
```
SanchithHegde: Nit: Naming. Use either `revenue_recovery_proto_files` or `recovery_proto_files` or just `proto_files`, don't incorrectly name it as `v2_proto_files`.

crates/external_services/build.rs
SanchithHegde: No, my question is about the use / derive of `serde::Serialize` and `serde::Deserialize`, why do we need them? And if we need them, what serialization format would be used over there: would it be ISO 8601 (like the rest of the codebase), or something else?

crates/external_services/src/grpc_client.rs
```diff
@@ -14,19 +18,26 @@ use common_utils::consts;
...
+#[cfg(any(feature = "dynamic_routing", feature = "revenue_recovery"))]
 use tonic::body::Body;
 
+#[cfg(feature = "revenue_recovery")]
+pub use self::revenue_recovery::common::GrpcRecoveryHeaders;
+#[cfg(feature = "revenue_recovery")]
+pub use self::revenue_recovery::recovery_decider_client::{
+    DeciderRequest, DeciderResponse, RecoveryDeciderClientConfig, RecoveryDeciderClientInterface,
+    RecoveryDeciderError, RecoveryDeciderResult,
+};
```
SanchithHegde: Nit: You can combine these imports.

crates/external_services/src/grpc_client.rs
```diff
@@ -39,15 +50,22 @@ pub struct GrpcClients {
     /// Health Check client for all gRPC services
     #[cfg(feature = "dynamic_routing")]
     pub health_client: HealthCheckClient,
+    /// Recovery Decoder Client
```
SanchithHegde: Nit: Typo.


```suggestion
    /// Recovery Decider Client
```

crates/external_services/src/grpc_client.rs
```diff
@@ -59,7 +77,8 @@ impl GrpcClientSettings {
     /// This function will be called at service startup.
     #[allow(clippy::expect_used)]
     pub async fn get_grpc_client_interface(&self) -> Arc<GrpcClients> {
-        #[cfg(feature = "dynamic_routing")]
+        // Define the hyper client if any gRPC feature is enabled
```
SanchithHegde: You can remove such obvious (possibly AI) comments.

crates/external_services/src/lib.rs
```diff
@@ -28,6 +28,40 @@ pub mod managers;
 /// crm module
 pub mod crm;
 
+#[cfg(feature = "revenue_recovery")]
+/// date_time module
+pub mod date_time {
+    use prost_types::Timestamp;
+    use time::{OffsetDateTime, PrimitiveDateTime};
```
SanchithHegde: Please avoid such absolute imports if possible.

crates/external_services/src/lib.rs
```diff
@@ -28,6 +28,40 @@ pub mod managers;
...
+        ts: &Timestamp,
+    ) -> error_stack::Result<PrimitiveDateTime, common::DateTimeConversionError> {
+        let timestamp_nanos = i128::from(ts.seconds) * 1_000_000_000 + i128::from(ts.nanos);
+
+        OffsetDateTime::from_unix_timestamp_nanos(timestamp_nanos)
+            .map(|offset_dt| PrimitiveDateTime::new(offset_dt.date(), offset_dt.time()))
+            .map_err(|original_error| {
+                router_env::logger::error!("Prost timestamp conversion failed: {original_error:?}");
+                error_stack::report!(common::DateTimeConversionError::TimestampOutOfRange)
+            })
```
SanchithHegde: Why not call `change_context()` instead of `map_err()`?

crates/external_services/src/lib.rs
```diff
@@ -28,6 +28,40 @@ pub mod managers;
...
+            // This conversion is safe as nanoseconds (0..999_999_999) always fit within an i32.
+            #[allow(clippy::as_conversions)]
+            nanos: odt.nanosecond() as i32,
+        }
+    }
+
+    /// Converts a `prost_types::Timestamp` to an `time::PrimitiveDateTime`.
+    pub fn convert_from_prost_timestamp(
+        ts: &Timestamp,
+    ) -> error_stack::Result<PrimitiveDateTime, common::DateTimeConversionError> {
```
SanchithHegde: Can we move the `DateTimeConversionError` outside the revenue recovery module, into this module maybe? I'm asking because these utility functions are in the a `date_time` module within the crate and not within a revenue recovery module, so the error can also reside here itself.

crates/external_services/src/grpc_client/revenue_recovery/common.rs
```diff
@@ -0,0 +1,54 @@
...
+
+/// Creates a tonic::Request with recovery headers added.
+pub(crate) fn create_revenue_recovery_grpc_request<T: Debug>(
+    message: T,
+    recovery_headers: GrpcRecoveryHeaders,
+) -> tonic::Request<T> {
+    let mut request = tonic::Request::new(message);
+    request.add_recovery_headers(recovery_headers);
+    request
+}
```
SanchithHegde: Is there any specific reason why you're not just calling `create_grpc_request()` and have implemented it here? Same for the `AddRecoveryHeaders` trait, a similar one exists.

config/development.toml
AdityaKumaar21: Done

crates/external_services/src/grpc_client/revenue_recovery/recovery_decider_client.rs
```diff
@@ -0,0 +1,113 @@
...
+    /// create a connection
+    pub fn get_recovery_decider_connection(
+        &self,
+        hyper_client: Client,
+    ) -> Result<DeciderClient<Client>, Report<RecoveryDeciderError>> {
+        if self.base_url.is_empty() {
+            return Err(Report::new(RecoveryDeciderError::ConfigError(
+                "Base URL is not configured for Recovery Decider client".to_string(),
+            )));
+        }
```
SanchithHegde: This would cause a deployment failure if the feature flag is enabled but the URL is not provided, right? We can't be doing that, we need another runtime toggle.

crates/external_services/src/grpc_client/revenue_recovery/recovery_decider_client.rs
```diff
@@ -0,0 +1,113 @@
...
+
+        logger::debug!(decider_request =?request);
+
+        let grpc_response = self
+            .decide(request)
+            .await
+            .map_err(|status| {
+                logger::error!(grpc_error =?status, "Decider service call failed");
+                RecoveryDeciderError::ServiceError(status.message().to_string())
+            })?
```
SanchithHegde: Nit: You can consider using `change_context()` instead of `map_err()`.

crates/external_services/src/grpc_client.rs
SanchithHegde: Nit: You can combine these imports.

crates/external_services/Cargo.toml
```diff
@@ -43,8 +53,10 @@ serde = { version = "1.0.219", features = ["derive"] }
 thiserror = "1.0.69"
 vaultrs = { version = "0.7.4", optional = true }
 prost = { version = "0.13", optional = true }
+prost-types = { version = "0.13", optional = true }
+time = { version = "0.3.41", features = ["serde", "serde-well-known", "std"] }
 tokio = "1.45.1"
-tonic = "0.13.1"
+tonic =  "0.13.1"
```
SanchithHegde: Nit: Can you revert this (whitespace) change?

crates/external_services/Cargo.toml
```diff
@@ -68,7 +80,7 @@ api_models = { version = "0.1.0", path = "../api_models", optional = true }
 
 
 [build-dependencies]
-tonic-build = "0.13.1"
+tonic-build =  "0.13.1"
```
SanchithHegde: Same here.

crates/router/Cargo.toml
SanchithHegde: Can you make this dependency optional and enable it when the necessary feature is enabled?

crates/router/src/core/revenue_recovery/types.rs
```diff
@@ -568,13 +570,13 @@ impl Action {
     pub async fn payment_sync_call(
         state: &SessionState,
         revenue_recovery_payment_data: &storage::revenue_recovery::RevenueRecoveryPaymentData,
-        global_payment_id: &id_type::GlobalPaymentId,
+        payment_intent: PaymentIntent,
```
SanchithHegde: Nit: Can't we accept a reference here instead?

SanchithHegde: And like I've stated [before](https://github.com/juspay/hyperswitch/pull/8178#discussion_r2191012394):

> Nit: There's inconsistency in imports, either fully import both payment attempt and payment intent, or qualify both. I'd prefer qualifying both.

crates/router/src/types/storage/revenue_recovery.rs
```diff
@@ -1,13 +1,18 @@
...
-use common_utils::id_type;
+use common_utils::{ext_traits::ValueExt, id_type};
+use external_services::grpc_client::{self as external_grpc_client, GrpcHeaders};
 use hyperswitch_domain_models::{
     business_profile, merchant_account, merchant_connector_account, merchant_key_store,
+    payment_method_data::{Card, PaymentMethodData},
+    payments::{payment_attempt::PaymentAttempt, PaymentIntent},
 };
+use masking::PeekInterface;
+use prost_types::Timestamp;
```
SanchithHegde: Nit: Prefer qualifying this import instead.

crates/router/src/types/storage/revenue_recovery.rs
```diff
@@ -40,21 +47,34 @@ impl RevenueRecoveryPaymentData {
...
 pub struct RevenueRecoverySettings {
     pub monitoring_threshold_in_seconds: i64,
     pub retry_algorithm_type: enums::RevenueRecoveryAlgorithmType,
+    pub recovery_timestamp: RecoveryTimestamp,
+}
+
+#[derive(Debug, serde::Deserialize, Clone, Default)]
+pub struct RecoveryTimestamp {
+    pub initial_timestamp_in_hours: i64,
+    pub final_timestamp_in_days: i64,
```
SanchithHegde: Should we consider keeping both of these numbers in terms of hours maybe, if it helps with testing? Or do we have other ways to quickly test it out without having to wait for the specified number of days?

What do you say, @Aprabhat19 @srujanchikke?

crates/router/src/workflows/revenue_recovery.rs
```diff
@@ -6,9 +6,23 @@ use common_utils::{
     id_type,
 };
 #[cfg(feature = "v2")]
+use diesel_models::types::BillingConnectorPaymentMethodDetails;
+#[cfg(feature = "v2")]
 use error_stack::ResultExt;
+#[cfg(all(feature = "revenue_recovery", feature = "v2"))]
+use external_services::date_time;
+#[cfg(all(feature = "revenue_recovery", feature = "v2"))]
+use external_services::grpc_client::revenue_recovery::recovery_decider_client as external_grpc_client;
```
SanchithHegde: Nit: You can combine these imports.

crates/router/src/workflows/revenue_recovery.rs
```diff
@@ -200,3 +214,156 @@ pub(crate) async fn get_schedule_time_to_retry_mit_payments(
...
+pub(crate) async fn get_schedule_time_for_smart_retry(
+    state: &SessionState,
+    payment_attempt: &PaymentAttempt,
+    payment_intent: &PaymentIntent,
+    retry_count: i32,
+) -> Option<time::PrimitiveDateTime> {
+    let first_error_message = payment_attempt
+        .error
+        .as_ref()
+        .map_or(String::new(), |error| error.message.clone());
```
SanchithHegde: Why are we using the empty string here?

crates/router/src/workflows/revenue_recovery.rs
```diff
@@ -200,3 +214,156 @@ pub(crate) async fn get_schedule_time_to_retry_mit_payments(
...
+        .and_then(|payment_metadata| {
+            payment_metadata
+                .billing_connector_payment_method_details
+                .as_ref()
+        });
+
+    let card_network_str = payment_revenue_recovery_metadata
+        .and_then(|details| match details {
+            BillingConnectorPaymentMethodDetails::Card(card_info) => card_info.card_network.clone(),
+            _ => None,
```
SanchithHegde: Should we consider explicitly handling all variants instead of using a wildcard `match` arm?
@Aprabhat19 What do you think?

crates/router/src/workflows/revenue_recovery.rs
```diff
@@ -200,3 +214,156 @@ pub(crate) async fn get_schedule_time_to_retry_mit_payments(
...
+        })
+        .map(|cn| cn.to_string())?;
+
+    let card_issuer_str = payment_revenue_recovery_metadata.and_then(|details| match details {
+        BillingConnectorPaymentMethodDetails::Card(card_info) => card_info.card_issuer.clone(),
+        _ => None,
```
SanchithHegde: Same here, need your opinions @Aprabhat19.

crates/router/src/workflows/revenue_recovery.rs
```diff
@@ -200,3 +214,156 @@ pub(crate) async fn get_schedule_time_to_retry_mit_payments(
...
+    })?;
+
+    // will remove this after we are getting the card issuer from Stripe
+    // .unwrap_or("CHASE".to_string());
```
SanchithHegde: What about this code, can we remove this?

crates/router/src/workflows/revenue_recovery.rs
```diff
@@ -200,3 +214,156 @@ pub(crate) async fn get_schedule_time_to_retry_mit_payments(
...
+                .as_ref()
+        })
+        .map(|payment_metadata| payment_metadata.payment_method_subtype.to_string())?;
+
+    let start_time_primitive = payment_intent.created_at;
+    let recovery_timestamp_config = &state.conf.revenue_recovery.recovery_timestamp;
+    // 1 hr after the actual start time
+    let modified_start_time_primitive = start_time_primitive.saturating_add(time::Duration::hours(
+        recovery_timestamp_config.initial_timestamp_in_hours,
+    ));
```
SanchithHegde: Can we now remove the comment, considering that we're reading that value from config, and it may no longer be one hour now?

crates/router/src/workflows/revenue_recovery.rs
```diff
@@ -200,3 +214,156 @@ pub(crate) async fn get_schedule_time_to_retry_mit_payments(
...
+    let start_time_proto = date_time::convert_to_prost_timestamp(modified_start_time_primitive);
+
+    // Calculate end_time as start_time + 30 days
+    let end_time_primitive = start_time_primitive.saturating_add(time::Duration::days(
+        recovery_timestamp_config.final_timestamp_in_days,
+    ));
```
SanchithHegde: Same here, this value may not be 30 days anymore, since we're reading it from config.

crates/router/src/workflows/revenue_recovery.rs
```diff
@@ -200,3 +214,156 @@ pub(crate) async fn get_schedule_time_to_retry_mit_payments(
...
+                .retry_flag
+                .then_some(())
+                .and(grpc_response.retry_time)
+                .and_then(|prost_ts| {
+                    match date_time::convert_from_prost_timestamp(&prost_ts) {
+                        Ok(pdt) => Some(pdt),
+                        Err(e) => {
+                            logger::error!(
+                                "Failed to convert retry_time from prost::Timestamp: {e:?}"
+                            );
```
SanchithHegde, line 408: How are you planning to know about these errors? Do you plan to have some sort of alerting setup?

config/development.toml
```diff
@@ -1256,6 +1259,10 @@ ucs_only_connectors = [
 monitoring_threshold_in_seconds = 2592000
 retry_algorithm_type = "cascading"
 
+[revenue_recovery.recovery_timestamp]
+initial_timestamp_in_hours = 1
+final_timestamp_in_days = 30
```
SanchithHegde: Any specific reason this config hasn't been added in the other config files?

crates/router/src/workflows/revenue_recovery.rs
```diff
@@ -200,3 +214,156 @@ pub(crate) async fn get_schedule_time_to_retry_mit_payments(
...
+        .and_then(|revenue_recovery_data| {
+            revenue_recovery_data
+                .payment_revenue_recovery_metadata
+                .as_ref()
+        })
+        .and_then(|payment_metadata| {
+            payment_metadata
+                .billing_connector_payment_method_details
+                .as_ref()
+        });
```
SanchithHegde: Nit: This variable is supposed to be payment method details I believe? 

You can keep two different variables, one for payment revenue recovery metadata, one for billing connector payment method details.

crates/router/src/workflows/revenue_recovery.rs
AdityaKumaar21: yes, we are planning to have a monitoring system in future

config/config.example.toml
```diff
@@ -1146,13 +1146,20 @@ host = "localhost"                      # Unified Connector Service Client Host
...
 [billing_connectors_invoice_sync]
 billing_connectors_which_requires_invoice_sync_call = "recurly" # List of billing connectors which has invoice sync api call
 
 [revenue_recovery]
 monitoring_threshold_in_seconds = 2592000 # 30*24*60*60 secs , threshold for monitoring the retry system
 retry_algorithm_type = "cascading" # type of retry algorithm
 
+[revenue_recovery.recovery_timestamp] # Timestamp configuration for Revenue Recovery
+initial_timestamp_in_hours = 1          # number of hours added to start time for Decider service of Revenue Recovery
+final_timestamp_in_hours = 720          # number of days added to start time for ent time in Revenue Recovery
```
SanchithHegde: This should have been number of hours?

And I feel something is wrong here: what's "ent time"

> added to start time for ent time in Revenue Recovery

config/deployments/env_specific.toml
```diff
@@ -383,6 +386,10 @@ host = "localhost"                      # Unified Connector Service Client Host
 port = 8000                             # Unified Connector Service Client Port
 connection_timeout = 10                 # Connection Timeout Duration in Seconds
 
+[revenue_recovery.recovery_timestamp] # Timestamp configuration for Revenue Recovery
+initial_timestamp_in_hours = 1        # number of hours added to start time for Decider service of Revenue Recovery
+final_timestamp_in_hours = 30          # number of days added to start time for ent time in Revenue Recovery
```
SanchithHegde: Again, same comments here. And should the number have been 720 instead of 30?

config/development.toml
```diff
@@ -1256,6 +1259,10 @@ ucs_only_connectors = [
 monitoring_threshold_in_seconds = 2592000
 retry_algorithm_type = "cascading"
 
+[revenue_recovery.recovery_timestamp]
+initial_timestamp_in_hours = 1
+final_timestamp_in_hours = 30
```
SanchithHegde: Same here, should this number have been 720?

crates/external_services/src/grpc_client.rs
```diff
@@ -14,19 +18,27 @@ use common_utils::consts;
 use dynamic_routing::{DynamicRoutingClientConfig, RoutingStrategy};
 #[cfg(feature = "dynamic_routing")]
 use health_check_client::HealthCheckClient;
-#[cfg(feature = "dynamic_routing")]
+#[cfg(any(feature = "dynamic_routing", feature = "revenue_recovery"))]
 use hyper_util::client::legacy::connect::HttpConnector;
-#[cfg(feature = "dynamic_routing")]
+#[cfg(any(feature = "dynamic_routing", feature = "revenue_recovery"))]
 use router_env::logger;
 use serde;
```
SanchithHegde: Remove this `use serde`, this is a useless import.

crates/external_services/src/lib.rs
```diff
@@ -28,6 +28,43 @@ pub mod managers;
 /// crm module
 pub mod crm;
 
+#[cfg(feature = "revenue_recovery")]
+/// date_time module
+pub mod date_time {
+    use error_stack::ResultExt;
+    use prost_types::Timestamp;
```
SanchithHegde: > Please avoid such absolute imports if possible.

I meant this for the import of `prost_types::Timestamp` as well. 😅

crates/external_services/src/grpc_client/revenue_recovery/recovery_decider_client.rs
```diff
@@ -0,0 +1,110 @@
...
+
+impl RecoveryDeciderClientConfig {
+    /// create a connection
+    pub fn get_recovery_decider_connection(
+        &self,
+        hyper_client: Client,
+    ) -> Result<Option<DeciderClient<Client>>, Report<RecoveryDeciderError>> {
+        if self.base_url.is_empty() {
+            return Ok(None);
+        }
```
SanchithHegde: My bad (for my earlier comment), since `recovery_decider_client` field in `GrpcClientSettings` is an optional field, I think you can error out if the base URL is empty. But please confirm that it works as expected:

- If no `recovery_decider_client` related configs are provided, application should still run.
- If `recovery_decider_client` is provided with an empty base URL, then it can error out.

crates/external_services/src/grpc_client/revenue_recovery/recovery_decider_client.rs
```diff
@@ -0,0 +1,110 @@
...
+
+        let uri = self
+            .base_url
+            .parse::<tonic::transport::Uri>()
+            .map_err(Report::from)
+            .change_context(RecoveryDeciderError::ConfigError(format!(
+                "Invalid URI: {}",
+                self.base_url
+            )))?;
```
SanchithHegde, line 85: Nit: You can call `change_context_lazy()`, since you're calling `format!()` here, which does an allocation.

crates/external_services/src/grpc_client/revenue_recovery/common.rs
AdityaKumaar21: in `create_grpc_request`() the Header is having  `tenant_id` as mandatory field and it is not required for our use case, so thats why i have created another Header with `request_id` as optional field and no `tenant_id`.

crates/external_services/src/grpc_client/revenue_recovery/recovery_decider_client.rs
```diff
@@ -0,0 +1,110 @@
...
+        let service_client = DeciderClient::with_origin(hyper_client, uri);
+
+        Ok(Some(service_client))
+    }
+}
+
+#[async_trait::async_trait]
+impl RecoveryDeciderClientInterface for DeciderClient<Client> {
+    /// collects the request from HS and sends it to recovery decider gRPC service
+    async fn decide_on_retry(
```
SanchithHegde: HS?

(By the way, even if you don't add documentation comments for the implementation here, it will consider documentation comments provided where the trait is defined.)

crates/router/src/types/storage/revenue_recovery.rs
```diff
@@ -40,21 +47,34 @@ impl RevenueRecoveryPaymentData {
...
     pub monitoring_threshold_in_seconds: i64,
     pub retry_algorithm_type: enums::RevenueRecoveryAlgorithmType,
+    pub recovery_timestamp: RecoveryTimestamp,
+}
+
+#[derive(Debug, serde::Deserialize, Clone, Default)]
+pub struct RecoveryTimestamp {
+    pub initial_timestamp_in_hours: i64,
+    pub final_timestamp_in_hours: i64,
 }
```
SanchithHegde: Because you're deriving `Default` here, the values that would be populated would be zeroes, in case none of the fields in `RevenueRecoverySettings` are provided. @Aprabhat19 @srujanchikke are we okay with that, or do you folks think we need to have some custom defaults?

crates/router/src/types/storage/revenue_recovery.rs
```diff
@@ -1,13 +1,18 @@
...
-use common_utils::id_type;
+use common_utils::{ext_traits::ValueExt, id_type};
+use external_services::grpc_client::{self as external_grpc_client, GrpcHeaders};
 use hyperswitch_domain_models::{
     business_profile, merchant_account, merchant_connector_account, merchant_key_store,
+    payment_method_data::{Card, PaymentMethodData},
+    payments::{payment_attempt::PaymentAttempt, PaymentIntent},
 };
+use masking::PeekInterface;
+// use prost_types::Timestamp;
```
SanchithHegde: Nit: Remove commented code.

crates/router/src/workflows/revenue_recovery.rs
```diff
@@ -200,3 +214,151 @@ pub(crate) async fn get_schedule_time_to_retry_mit_payments(
...
+        .and_then(|details| details.state.as_ref())
+        .cloned()?;
+
+    // Check if payment_method_data itself is None
+    if payment_attempt.payment_method_data.is_none() {
+        logger::debug!(
+            payment_intent_id = ?payment_intent.get_id(),
+            attempt_id = ?payment_attempt.get_id(),
+            message = "payment_attempt.payment_method_data is None"
+        );
```
SanchithHegde, line 250: Nit: You don't have to explicitly pass the `message` field.

```suggestion
        logger::debug!(
            payment_intent_id = ?payment_intent.get_id(),
            attempt_id = ?payment_attempt.get_id(),
            "payment_attempt.payment_method_data is None"
        );
```

crates/router/src/workflows/revenue_recovery.rs
```diff
@@ -200,3 +214,151 @@ pub(crate) async fn get_schedule_time_to_retry_mit_payments(
...
+        })
+        .and_then(|payment_metadata| {
+            payment_metadata
+                .billing_connector_payment_method_details
+                .as_ref()
+        });
+
+    let card_network_str = billing_connector_payment_method_details
+        .and_then(|details| match details {
+            BillingConnectorPaymentMethodDetails::Card(card_info) => card_info.card_network.clone(),
```
SanchithHegde, line 274: Nit: Can't we avoid the clone here? I think you should be able to, by deriving `Copy` on the `common_enums::enums::CardNetwork` enum.

crates/router/src/workflows/revenue_recovery.rs
```diff
@@ -200,3 +214,151 @@ pub(crate) async fn get_schedule_time_to_retry_mit_payments(
...
+        })
+        .map(|cn| cn.to_string())?;
+
+    let card_issuer_str =
+        billing_connector_payment_method_details.and_then(|details| match details {
+            BillingConnectorPaymentMethodDetails::Card(card_info) => card_info.card_issuer.clone(),
+        })?;
+
+    let card_funding_str = payment_intent
```
SanchithHegde, line 283: Card funding? Where are we obtaining this information from?

crates/router/src/workflows/revenue_recovery.rs
```diff
@@ -200,3 +214,151 @@ pub(crate) async fn get_schedule_time_to_retry_mit_payments(
...
+    } else {
+        logger::debug!("Recovery decider client is not configured");
+        None
+    }
+}
+
+#[cfg(feature = "v2")]
+#[derive(Debug)]
+struct InternalDeciderRequest {
+    first_error_message: String,
```
SanchithHegde, line 428: I don't know how this is being used on the decider side, but do you need this to be a required field or can this be optional (considering that you're populating this with `NO_ERROR_MESSAGE` in case there's no error in the payment attempt)?

crates/router/src/workflows/revenue_recovery.rs
```diff
@@ -200,3 +214,151 @@ pub(crate) async fn get_schedule_time_to_retry_mit_payments(
...
+    end_time: Option<prost_types::Timestamp>,
+    retry_count: f64,
+}
+
+#[cfg(feature = "v2")]
+impl From<InternalDeciderRequest> for external_grpc_client::DeciderRequest {
+    fn from(internal_request: InternalDeciderRequest) -> Self {
+        Self {
+            first_error_message: internal_request.first_error_message,
+            billing_state: internal_request.billing_state.peek().to_string(),
```
SanchithHegde: @Aprabhat19 @srujanchikke Ideally you folks should consider using an equivalent of `Secret` on the gRPC server side as well, similar to how the connector-service folks are doing...

crates/router/src/workflows/revenue_recovery.rs
AdityaKumaar21: yes, this is a mandatory field in decider end to have a scheduled time

SanchithHegde: How would you handle it if you have `No error message`? What do you think is better to handle on your side: a fixed value (`No error message`) or no value at all (`null` / `None`)?

crates/router/src/workflows/revenue_recovery.rs
AdityaKumaar21: This is basically card type that we are fetching from payment intent's `PaymentRevenueRecoveryMetadata`-> `payment_method_subtype`

SanchithHegde: Why the "funding" in the variable name?

config/config.example.toml
```diff
@@ -1148,12 +1148,18 @@ url = "http://localhost:8080"           # Open Router URL
...
+[grpc_client.recovery_decider_client] # Revenue recovery client base url
+base_url = "http://127.0.0.1:8080" #Base URL
+
 [billing_connectors_invoice_sync]
 billing_connectors_which_requires_invoice_sync_call = "recurly" # List of billing connectors which has invoice sync api call
 
 [revenue_recovery]
-monitoring_threshold_in_seconds = 2592000 # 30*24*60*60 secs , threshold for monitoring the retry system
-retry_algorithm_type = "cascading" # type of retry algorithm
+monitoring_threshold_in_seconds = 60  # 30*24*60*60 secs , threshold for monitoring the retry system
```
SanchithHegde: The number and the comment don't match?

SanchithHegde: The number (60) and the comment don't match?

crates/external_services/src/grpc_client.rs
```diff
@@ -75,18 +93,48 @@ impl GrpcClientSettings {
...
+
+                    // Create the client
+                    let client = config
+                        .get_recovery_decider_connection(client.clone())
+                        .expect(
+                            "Failed to establish a connection with the Recovery Decider Server",
+                        );
+
+                    logger::info!("Recovery Decider gRPC client successfully initialized");
+                    // Some(Box::new(client) as Box<dyn RecoveryDeciderClientInterface>)
```
SanchithHegde: Remove commented code.

crates/router/src/core/revenue_recovery/types.rs
```diff
@@ -97,7 +99,7 @@ impl RevenueRecoveryPaymentsAttemptStatus {
         payment_intent: &PaymentIntent,
         process_tracker: storage::ProcessTracker,
         revenue_recovery_payment_data: &storage::revenue_recovery::RevenueRecoveryPaymentData,
-        payment_attempt: payment_attempt::PaymentAttempt,
+        payment_attempt: PaymentAttempt,
```
SanchithHegde, line 102: Nit: Can't we accept a reference here? Do you necessarily need an owned value?

crates/router/src/core/revenue_recovery/types.rs
```diff
@@ -547,13 +549,13 @@ impl Action {
     pub async fn payment_sync_call(
         state: &SessionState,
         revenue_recovery_payment_data: &storage::revenue_recovery::RevenueRecoveryPaymentData,
-        global_payment_id: &id_type::GlobalPaymentId,
+        payment_intent: &PaymentIntent,
         process: &storage::ProcessTracker,
-        payment_attempt: payment_attempt::PaymentAttempt,
+        payment_attempt: PaymentAttempt,
```
SanchithHegde, line 554: Same here, can't we accept a reference?

crates/router/src/types/storage/revenue_recovery.rs
```diff
@@ -40,21 +46,41 @@ impl RevenueRecoveryPaymentData {
...
 #[derive(Debug, serde::Deserialize, Clone, Default)]
 pub struct RevenueRecoverySettings {
     pub monitoring_threshold_in_seconds: i64,
     pub retry_algorithm_type: enums::RevenueRecoveryAlgorithmType,
+    pub recovery_timestamp: RecoveryTimestamp,
+}
+
+#[derive(Debug, serde::Deserialize, Clone)]
+pub struct RecoveryTimestamp {
+    pub initial_timestamp_in_hours: i64,
```
SanchithHegde, line 77: Why do we need an `i64`, can't we use something like `u16` or `u32`? This is anyway a duration, we don't need these to be negative.

crates/router/src/types/storage/revenue_recovery.rs
```diff
@@ -40,21 +46,41 @@ impl RevenueRecoveryPaymentData {
...
+                )
+                .await
             }
         }
     }
 }
+
 #[derive(Debug, serde::Deserialize, Clone, Default)]
 pub struct RevenueRecoverySettings {
     pub monitoring_threshold_in_seconds: i64,
```
SanchithHegde, line 70: Same here, we don't need an `i64`?

srujanchikke (approved):

crates/router/src/workflows/revenue_recovery.rs
AdityaKumaar21: If i derive `Copy` on `common_enums::enums::CardNetwork` , i have to make changes in all the place where we are using `CardNetwork`, should i do that ?

SanchithHegde: Yes, please.

PR #8184, opened 2025-05-30
refactor(routing):  optimize database calls with futures for dynamic algorithms

GauravRawat369:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

We changed the database lookups from running one-by-one (sequentially) to running all at once (concurrently). This was done by:

1. Collecting all database queries as asynchronous tasks (futures).
2. Using `join_all` to execute these tasks in parallel.
3. Processing the combined results.

This speeds up the function by overlapping the waiting times for database responses.

### Additional Changes

## Checklist


crates/router/src/core/routing.rs
```diff
@@ -1266,16 +1279,32 @@ pub async fn retrieve_linked_routing_config(
...
-            let record = db
-                .find_routing_algorithm_metadata_by_algorithm_id_profile_id(
-                    &algorithm_id,
-                    profile_id,
-                )
-                .await
-                .to_not_found_response(errors::ApiErrorResponse::ResourceIdNotFound)?;
-            active_algorithms.push(record.foreign_into());
+            let store = state.store.clone();
+            let profile_id_clone = profile_id.clone();
```
prajjwalkumar17, line 1296:
```suggestion
            let profile = profile_id.clone();
```
Nit: we don't name something with a suffix or prefix considering the type, So we don't name like currency_vector, instead we can say something like currencies.

crates/router/src/core/routing.rs
```diff
@@ -1266,16 +1279,32 @@ pub async fn retrieve_linked_routing_config(
...
+            db_call_futures.push(Box::pin(async move {
+                store
+                    .find_routing_algorithm_metadata_by_algorithm_id_profile_id(
+                        &algorithm_id,
+                        &profile_id_clone,
+                    )
+                    .await
+                    .to_not_found_response(errors::ApiErrorResponse::ResourceIdNotFound)
```
prajjwalkumar17, line 1304: We don't need to await it here, as we are going for a `join_all`, So basically we want to achieve something like this(collect all the futures and then awaiting on them collectively):
```
    let retrieve_config_futures = all_profiles
        .iter()
        .map(|prof| {
            helpers::get_merchant_default_config(
                db,
                prof.get_id().get_string_repr(),
                transaction_type,
            )
        })
        .collect::<Vec<_>>();

    let configs = futures::future::join_all(retrieve_config_futures)
        .await
        .into_iter()
        .collect::<Result<Vec<_>, _>>()?;
```

PR #8191, opened 2025-05-31
refactor(dynamic_routing): add logic for creating merchant account in decision engine

Chethan-rao:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

### Additional Changes

## Checklist

### DIFF
--- a/crates/router/src/core/admin.rs
+++ b/crates/router/src/core/admin.rs
@@ -264,26 +264,6 @@ pub async fn create_merchant_account(
         .await
         .to_duplicate_response(errors::ApiErrorResponse::DuplicateMerchantAccount)?;
 
-    // Call to DE here
-    // Check if creation should be based on default profile
-    #[cfg(all(feature = "dynamic_routing", feature = "v1"))]
-    {
-        if state.conf.open_router.enabled {
-            merchant_account
-                .default_profile
-                .as_ref()
-                .async_map(|profile_id| {
-                    routing::helpers::create_decision_engine_merchant(&state, profile_id)
-                })
-                .await
-                .transpose()
-                .map_err(|err| {
-                    crate::logger::error!("Failed to create merchant in Decision Engine {err:?}");
-                })
-                .ok();
-        }
-    }
-
     let merchant_context = domain::MerchantContext::NormalMerchant(Box::new(domain::Context(
         merchant_account.clone(),
         key_store.clone(),
--- a/crates/router/src/core/routing/helpers.rs
+++ b/crates/router/src/core/routing/helpers.rs
@@ -1535,7 +1535,7 @@ pub async fn disable_dynamic_routing_algorithm(
     let db = state.store.as_ref();
     let key_manager_state = &state.into();
     let profile_id = business_profile.get_id().clone();
-    let (algorithm_id, dynamic_routing_algorithm, cache_entries_to_redact) =
+    let (algorithm_id, mut dynamic_routing_algorithm, cache_entries_to_redact) =
         match dynamic_routing_type {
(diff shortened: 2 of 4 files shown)


crates/router/src/core/admin.rs
```diff
@@ -264,26 +264,6 @@ pub async fn create_merchant_account(
...
-                })
-                .await
-                .transpose()
-                .map_err(|err| {
-                    crate::logger::error!("Failed to create merchant in Decision Engine {err:?}");
-                })
-                .ok();
-        }
-    }
-
```
Chethan-rao, line 286: I don't think we should create merchant_account in decision engine for every profile creation on our side as we are not sure whether SR will be toggled for that profile. Instead I have moved the merchant creation call in DE to the place where SR is being toggled. From this we are taking of all the below cases - 
* New profile creation: Can use SR toggle API which creates both merchant and rules in DE
* Old profile: for both the cases where SR is already toggled or not toggled, `is_merchant_created_in_decision_engine` flag in profile table will indicate whether merchant is created on DE. if not, we'll create it.

cc: @Sarthak1799  @jagan-jaya

jagan-jaya (approved):

prajjwalkumar17 (approved):

crates/router/src/core/routing/helpers.rs
```diff
@@ -2190,6 +2213,31 @@ pub async fn disable_decision_engine_dynamic_routing_setup(
...
+
+        create_decision_engine_merchant(state, profile_id)
+            .await
+            .map_err(|err| {
+                logger::warn!("Merchant creation error in decision_engine: {err:?}");
+            })
+            .ok();
+
+        // TODO: Update the status based on the status code or error message from the API call
+        dynamic_routing_algo_ref.update_merchant_creation_status_in_decision_engine(true);
```
Sarthak1799, line 2425: Rare edge case but if DE response does not come through while the operation still being a success on DE, We'd again make a call for merchant creation next time around, which would fail and this status would never get updated.

A workaround would be to do a GET call on the merchant-account instead of maintaining state at Hyperswitch.

crates/router/src/core/routing/helpers.rs
```diff
@@ -2176,6 +2195,10 @@ pub async fn disable_decision_engine_dynamic_routing_setup(
         },
     };
 
+    // Create merchant in Decision Engine if it is not already created
+    create_merchant_in_decision_engine_if_not_exists(state, profile_id, dynamic_routing_algo_ref)
```
ShankarSinghC, line 2387: Can we move this function to a common location in the routing logic? We need to create a merchant whenever debit routing or dynamic routing is enabled.

crates/router/src/core/routing/helpers.rs
Chethan-rao: And when does this GET call be executed?

Sarthak1799: Instead of looking up the `is_merchant_created_in_decision_engine`, we can do a GET call instead. These are functionally equivalent right?

Chethan-rao: We didn't want to do this extra call everytime something is toggled. Also your case won't even occur if we switch the `is_merchant_created_in_decision_engine` flag based on status from create merchant call which will be picked up by @jagan-jaya

jagan-jaya (approved):

jagan-jaya (dismissed):

PR #8209, opened 2025-06-02
ci(cypress): fix fiuu, fiservemea , paybox and worldpay connector

likhinbopanna:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [x] CI/CD

## Description
Fix fiuu, ferservmea ,paybox and worldpay connector in cypress tests 
Handled 3DS test cases for connectors that do not support 3DS.

### Additional Changes

## How did you test it?
- Worldpay
image

Issue : 1. After creating 3ds payment but without complete the redirection , When psync is done , throwing this error

- Nexixpay
image

- Fiuu

image

Issue : 1. Zero auth mandate goes to processing state and test cases are failing
2. Merchant Payment method list is `null`

- Paybox

image

Issue : 1. zero auth mandate goes to processing state and test cases are failing

## Checklist


cypress-tests/cypress/support/commands.js
```diff
@@ -2577,6 +2591,7 @@ Cypress.Commands.add(
     const merchant_connector_id = globalState.get(
       `${configInfo.merchantConnectorPrefix}Id`
     );
+    const setupFutureUsage = globalState.get("setupFutureUsage");
```
Gnanasundari24: Fix the lints

Gnanasundari24 (dismissed):

cypress-tests/cypress/support/commands.js
```diff
@@ -2246,6 +2256,11 @@ Cypress.Commands.add(
                 "nextActionUrl",
                 response.body.next_action.redirect_to_url
               );
+              for (const key in resData.body) {
+                expect(resData.body[key], [key]).to.deep.equal(
+                  response.body[key]
+                );
+              }
```
pixincreate: why is this required? i have a hunch that adding a loop here will have adverse effects with other connectors and flows. I remember removing it few months ago for save card flow due to that.

cypress-tests/cypress/support/commands.js
```diff
@@ -2138,6 +2144,10 @@ Cypress.Commands.add(
     saveCardConfirmBody.payment_token = globalState.get("paymentToken");
     saveCardConfirmBody.profile_id = profile_id;
 
+    for (const key in reqData) {
+      saveCardConfirmBody[key] = reqData[key];
+    }
+
```
pixincreate: this will cause card info to be passed during the confirm call in save card flow and the cypress behavior changes. we have faced this before (i remember removing it few months ago). 

better to remove it.

cypress-tests/cypress/support/commands.js
```diff
@@ -50,6 +50,12 @@ function validateErrorMessage(response, resData) {
   }
 }
 
+//skip MIT using PMId if connector does not support MIT only
+export function shouldSkipMitUsingPMId(connectorId) {
+  const skipConnectors = ["fiuu"];
+  return skipConnectors.includes(connectorId);
+}
+
```
pixincreate, line 58: may be, you can use `CONNECTOR_LIST` const from `configs/Payments/Utils`? Introduce new const within `EXCLUDE`?

cypress-tests/cypress/e2e/spec/Payment/00019-MandatesUsingPMID.cy.js
```diff
@@ -26,6 +26,10 @@ describe("Card - Mandates using Payment Method Id flow test", () => {
         }
       });
 
+      it("customer-create-call-test", () => {
+        cy.createCustomerCallTest(fixtures.customerCreateBody, globalState);
+      });
+
```
pixincreate, line 32: Please verify that it does not affect other connectors. Saying so because you've touched the spec files and it affects all connectors.

cypress-tests/cypress/e2e/configs/Payment/Fiservemea.js
```diff
@@ -153,6 +379,248 @@ export const connectorDetails = {
...
+        mandate_data: singleUseMandateData,
+        billing: billingAddress,
+      },
+      Response: {
+        status: 200,
+        body: {
+          status: "requires_capture",
+        },
+      },
+    },
```
pixincreate, line 401: since the `setup_future_usage` does a fallback to `on_session` when mandate is not supported, we can start moving this config to `Commons` (refactor commons) such that, instead of expecting the `501` we expect a `2xx` response which can be overriden by connectors if they expect a different status in the response with `TRIGGER_SKIP` set to `true` just because mandate payments shouldn't happen.

yes, this can be considered in a separate pr given the complexity and the work involved.

cypress-tests/cypress/e2e/configs/Payment/Paybox.js
```diff
@@ -51,13 +51,33 @@ const captureNotSupported = {
     "Capture Not allowed in case of Creating the Subscriber is not supported by Paybox",
 };
 
+const billing_data = {
```
pixincreate: make sure it is `camelCase`.

cypress-tests/cypress/e2e/configs/Payment/WorldPay.js
```diff
@@ -1340,170 +875,6 @@ export const connectorDetails = {
...
-                    required_field: "payment_method_data.billing.address.city",
-                    display_name: "city",
-                    field_type: "user_address_city",
-                    value: null,
-                  },
-                  "billing.address.line1": {
-                    required_field: "payment_method_data.billing.address.line1",
-                    display_name: "line1",
-                    field_type: "user_address_line1",
-                    value: null,
```
pixincreate, line 1359: why was these removed?

likhinbopanna: Because worldpay doesn't have all these as a required field

cypress-tests/cypress/e2e/configs/Payment/Fiservemea.js
pixincreate: i would want this to be kept unresolved for now. if not asking for much, can you please open an issue on this topic so that it can be picked in future and not left as a tech debt?

Gnanasundari24 (approved):

PR #8213, opened 2025-06-02
refactor(connector): update add connector script with new connector features

swetasharma03:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [x] Enhancement
- [x] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

- Removed the unwanted changes (like addition of conector_name in RoutableConnectors enum, etc)
- Amount convertor code is generated in the template
- Connector Specification is also added in the template
- Network error codes added
- validate_mandate_payment, validate_psync_reference_id, toml files doesn't directly implements cards as supported payment method
- The unit test file now generated directly to crates/router/tests/connectors etc
- Derive macro copy added for all unit types like enums which only have 1 type of data (eg payment status, refund status) and structs which only have primitive types like i64, u64 , bool, char or unit structs.
- Connector getting added in all the files in alphabetical order
- In default_implmentation.rs, if connector previous to the new connector doesn't exist in any default_imp list, the new_connector will be added on the top of the list, else it will be in alphabetically sorted.
- Code getting compiled directly after template generation.
Screenshot 2025-06-12 at 8 10 59 PM

### Additional Changes

## How did you test it?

This PR only updates the script to generate template code for a connector, hence no API testing required.

Script used to test the template  generation.
`sh scripts/add_connector.sh testConnector https://www.google.com/`

Screenshot 2025-06-27 at 4 40 33 PM

## Checklist


connector-template/mod.rs
```diff
@@ -39,11 +39,17 @@ use hyperswitch_interfaces::{
     types::{self, Response},
     webhooks,
 };
+use std::sync::LazyLock;
```
deepanshu-iiitu, line 42: The script is not generating the connector_name.rs and the transformers.rs file.

swetasharma03: Screenshot 2025-06-03 at 4 10 21 PM
I tested the connector template generation locally, and it seems to be working fine on my end. Could you please check on your side and let me know if you're still facing the issue?

awasthi21 (approved):

scripts/add_connector.sh
```diff
@@ -54,40 +54,129 @@ previous_connector_camelcase="$(tr '[:lower:]' '[:upper:]' <<< ${previous_connec
...
+
+default_impl_files=(
+  "crates/hyperswitch_connectors/src/default_implementations.rs"
+  "crates/hyperswitch_connectors/src/default_implementations_v2.rs"
+)
+
+for file in "${default_impl_files[@]}"; do
+  tmpfile="${file}.tmp"
+
+  awk -v prev="$previous_connector_camelcase" -v new="$payment_gateway_camelcase" '
```
SanchithHegde, line 79: Could we at least have some comments as to what is happening here?

scripts/add_connector.sh
```diff
@@ -54,40 +54,129 @@ previous_connector_camelcase="$(tr '[:lower:]' '[:upper:]' <<< ${previous_connec
...
-                Ok(())\\
-            },\\
-" crates/router/src/core/admin.rs
+
 
 
 # Remove temporary files created in above step
 rm $conn.rs-e $src/types/api.rs-e $src/configs/settings.rs-e config/development.toml-e config/docker_compose.toml-e config/config.example.toml-e loadtest/config/development.toml-e crates/api_models/src/connector_enums.rs-e crates/euclid/src/enums.rs-e crates/api_models/src/routing.rs-e $src/core/payments/flows.rs-e crates/common_enums/src/connector_enums.rs-e $src/types/transformers.rs-e $src/core/admin.rs-e crates/hyperswitch_connectors/src/default_implementations.rs-e crates/hyperswitch_connectors/src/default_implementations_v2.rs-e crates/hyperswitch_interfaces/src/configs.rs-e $src/connector.rs-e config/deployments/integration_test.toml-e config/deployments/production.toml-e config/deployments/sandbox.toml-e temp crates/connector_configs/src/connector.rs-e crates/router/tests/connectors/main.rs-e crates/router/src/core/payments/connector_integration_v2_impls.rs-e
+
+sed -i'' -e "/pub ${previous_connector}: ConnectorParams,/a\\
```
SanchithHegde, line 177: Isn't this going to generate more temporary files, but we' have a step before this which says "Remove temporary files created in above step".

Can we move this line above the `rm` command?

scripts/add_connector.sh
```diff
@@ -109,9 +200,9 @@ echo "\n\n[${payment_gateway}]\napi_key=\"API Key\"" >> ${tests}/sample_auth.tom
 # Remove temporary files created in above step
 rm ${tests}/main.rs-e ${test_utils}/connector_auth.rs-e
 cargo +nightly fmt --all
-cargo check
+cargo run
```
SanchithHegde: Do we necessarily need this command to be `cargo run`, considering that later in the script, we also try to run tests? And that `cargo run` would prevent any further commands in the script from running, unless the server is stopped...?

swetasharma03: I initially used `cargo check`, but noticed some `unexpected behavior`: it reported a successful compilation even though there were 45+ errors in the diesel_models crate (which is unrelated to the generated connector). This seemed odd, as I expected cargo check to fail in such a case.

Here’s a screenshot of the terminal output with `cargo check`:

Screenshot 2025-06-27 at 5 08 04 PM


On the other hand, switching to `cargo run` ensured that the project was fully compiled without any error. Despite concerns that `cargo run` might block further commands due to the running server, it actually completed and allowed `cargo test` to run right afterward without requiring manual server termination.

Here’s a screenshot of the terminal output with `cargo run`:

Screenshot 2025-06-27 at 4 40 33 PM

scripts/add_connector.sh
swetasharma03: We’re already running the `rm` command immediately after the sed edit for `crates/hyperswitch_domain_models/src/configs.rs`, so the temp file (.rs-e) is cleaned up right away. I’ve tested the script multiple times, and no temporary files are left behind.

```
sed -i'' -e "/pub ${previous_connector}: ConnectorParams,/a\\
    pub ${payment_gateway}: ConnectorParams,
" crates/hyperswitch_domain_models/src/configs.rs
rm crates/hyperswitch_domain_models/src/configs.rs-e
```

What are your thoughts? Do we still need to move this code above, or is the current placement okay?

scripts/add_connector.sh
swetasharma03: Previously, in `default_implementations.rs`,`default_implementations_v2.rs`, the new connector was always added at the top of the macro list, without maintaining alphabetical order.

Also using a single `sed` command to insert the new connector depends on the presence of the `previous_connector` in the macro block. This approach fails when the `previous_connector` is missing from one or more lists, leading to skipped insertions.

To handle this correctly, the current logic does the following:

- If the `previous_connector` exists in the macro, the `new_connector` is added after it, preserving alphabetical order.
- If the `previous_connector` is not found, the `new_connector` is inserted at the top of the macro list.

I've also added comments in the code.

deepanshu-iiitu (dismissed):

awasthi21 (dismissed):

scripts/add_connector.sh
SanchithHegde: Can we then group all the `sed` commands together, followed by one `rm` command? So move this `sed` command before the big `rm` command?

scripts/add_connector.sh
SanchithHegde: Can you try with `cargo check --features v1`?

swetasharma03: Yes this worked, I have updated in the code.

deepanshu-iiitu (approved):

awasthi21 (approved):

PR #8320, opened 2025-06-11
fix: correct error for env not found

Shailesh-714:
## Type of Change

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

PR : #[8238](https://github.com/juspay/hyperswitch/pull/8238) calls for a env variable before the .env file is sourced in the scripts/setup.sh file.
This PR fixes the above mentioned issue and handles the health checks using flags instead of exit codes, enabling the script to proceed further without getting terminated.

### Additional Changes

## How did you test it?

1. Run `scripts/setup.sh`.
2. If it prompts `ports already in use`, enter `y` to proceed.
3. Choose `1) Standard Setup` in setup options.
4. If setup is successful, you will get this final output:
Screenshot 2025-06-16 at 5 18 19 PM

5. Try signing in to `control center` using the default credentials provided.

## Checklist

## Summary by CodeRabbit

- **Chores**
	- Improved the setup process to ensure environment variables are loaded after configuration is updated.

### DIFF
--- a/scripts/setup.sh
+++ b/scripts/setup.sh
@@ -299,24 +299,26 @@ check_services_health() {
     HYPERSWITCH_BASE_URL="http://localhost:8080"
     HYPERSWITCH_HEALTH_URL="${HYPERSWITCH_BASE_URL}/health"
     HYPERSWITCH_DEEP_HEALTH_URL="${HYPERSWITCH_BASE_URL}/health/ready"
-
+    local is_success=true
     # Basic health check
-    health_response=$(curl --silent --fail "${HYPERSWITCH_HEALTH_URL}") || exit 0
+    health_response=$(curl --silent --fail "${HYPERSWITCH_HEALTH_URL}") || is_success=false
     if [ "${health_response}" != "health is good" ]; then
-        exit 0
+        is_success=false
     fi
 
     # Deep health check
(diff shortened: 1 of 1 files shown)


inventvenkat (approved):

scripts/setup.sh
```diff
@@ -299,24 +299,26 @@ check_services_health() {
...
 
     # Extract version
-    VERSION=$(curl --silent --output /dev/null --request GET --write-out '%header{x-hyperswitch-version}' "${HYPERSWITCH_BASE_URL}" | sed 's/-dirty$//')
-    INSTALLATION_STATUS="success"
+    if [ "${is_success}" = true ]; then
+        VERSION=$(curl --silent --output /dev/null --request GET --write-out '%header{x-hyperswitch-version}' "${HYPERSWITCH_BASE_URL}" | sed 's/-dirty$//')
+        INSTALLATION_STATUS="success"
+        scarf_call
+    fi
     print_access_info
```
SanchithHegde, line 321: Won't this mean that we would print access info even if `is_success` is `false`? Is that the intended / expected behavior?

Shailesh-714: Yeah, its intended because checks can fail due to soft errors and these will be later altered when port allocation flow is added to it.

PR #8329, opened 2025-06-12
feat: add nix shell environments

kashif-m:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description
This PR extends nix shell environments by including
1. Default minimal shell
2. Development shell
3. QA shell

### Additional Changes

## How did you test it?
1. Minimal shell
```
nix develop
```
2. Dev shell
```
nix develop .#dev
```
3. QA shell
```
nix develop .#qa
```

## Checklist

### DIFF
--- a/docs/try_local_system.md
+++ b/docs/try_local_system.md
@@ -147,7 +147,7 @@ Hyperswitch offers three distinct Nix shells tailored for different activities:
       nix develop .#dev
     ```
    * **Purpose:** Extends the default shell with packages required during active development of the project.
-   * **Usage:** Within this shell, you can run clippy checks and validate OpenAPI specs. This shell uses the latest version of `rust`.
+   * **Usage:** Within this shell, you can run clippy checks and validate OpenAPI specs. This shell uses the latest version of `rust` at the time of commit.
 
 3. **QA Shell (`hyperswitch-qa-shell`):**
    ```shell
--- a/flake.lock
+++ b/flake.lock
@@ -192,11 +192,11 @@
         "nixpkgs": "nixpkgs_3"
       },
       "locked": {
-        "lastModified": 1749609482,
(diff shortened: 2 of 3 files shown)


docs/try_local_system.md
```diff
@@ -127,17 +127,34 @@ This will start the following services using `process-compose`
...
    ```
+   * **Purpose:** Minimal shell for compiling and running the server.
+   * **Usage:** Within this shell, you can run DB migrations, compile and run different components. It uses MSRV of `rust` defined for the project.
+
+2. **Development Shell (`hyperswitch-dev-shell`):**
+    ```shell
+      nix develop .#dev
+    ```
+   * **Purpose:** Extends the default shell with packages required during active development of the project.
+   * **Usage:** Within this shell, you can run clippy checks and validate OpenAPI specs. This shell uses the latest version of `rust`.
```
SanchithHegde: > This shell uses the latest version of `rust`.

But we seem to be using a pinned version of 1.84?

kashif-m: We need to use a pinned version when we directly download the rust binaries.

A way around this would be to use `rustup` - but this is against nix principles around reproducible builds. Imagine checking out to a tag / commit which is relatively older, you wouldn't wanna use the latest `rust` version for compiling the project at that point in time.

We can modify the docs to say something like -

```
Within this shell, you can run clippy checks and validate OpenAPI specs. This shell uses the latest version of `rust` at the time of commit.
```

Updating this will always be a manual effort. It can be a periodic activity (once in a quarter - based on the usage)

swangi-kumari (approved):

PR #8393, opened 2025-06-19
feat(revenue_recovery): Invoke attempt list instead of payment get in recovery webhooks flow

srujanchikke:
## Type of Change

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

Previously we were using payment get to find whether attempt is present in db or not, but for the first invoice webhook there won't be any attempt present in db and recently we added a validation to use payment get only in case of attempt is present.
This PR solves this issue by invoking payment get attempt list instead of payment get. 

Also for stripe billing / stripe connector_transaction_id remains same instead charge id keeps changing , so we added a support for this by checking if charge id is present we compare charge id if not we check for connector transaction id.

### Additional Changes

## How did you test it?

Tested complete revenue recovery flow with stripebilling
1) Set up merchant and mca for billing and payment connector for stripebilling and stripe
2) Make sure merchant have retry_algorithm_type to cascading in business profile
3) Trigger payment invoice from stripe, after threshold external attempts , pt entry should be done in and start triggering internal payments.

## Checklist


crates/hyperswitch_domain_models/src/payments.rs
```diff
@@ -746,6 +746,7 @@ impl PaymentIntent {
             invoice_next_billing_time: None,
             card_isin: None,
             card_network: None,
+            charge_id: None,
```
Aprabhat19, line 750: Why are we populating this to none here?

Aprabhat19 (approved):

Aprabhat19:
Also can you fix the spellcheck

crates/hyperswitch_domain_models/src/payments.rs
srujanchikke: No charge id is present here since it is a internal payment, we didn't call connector yet.

swangi-kumari (commented):
Connector files look good to me.

swangi-kumari (approved):

jagan-jaya (approved):

Aprabhat19 (approved):

hrithikesh026 (approved):
LGTM

Issue #8392, opened 2025-06-19
[FEATURE] allow payout_id to be configurable


kashif-m:
### Feature Description

`payout_id` is an unique identifier for payout resource which is auto generated during payout creation. This field needs to be configurable during payout creation by accepting in the API.

Along with this, a `merchant_order_reference_id` needs to be added which is an unique identifier for the consumer of the API.

### Possible Implementation

Below changes are needed in payout flow across different components -

- Add `payout_id` to API models
- Add `PayoutId` domain type
- Add `merchant_order_reference_id` to API and diesel models
- Add DB queries for fetching orders using `merchant_order_reference_id`
- Add `merchant_order_reference_id` as a filter for OLAP APIs
- Use uuidv4 for payout ID creation for Wise

### Have you spent some time checking if this feature request has been raised before?

- [x] I checked and didn't find a similar issue

### Have you read the Contributing Guidelines?

- [x] I have read the [Contributing Guidelines](https://github.com/juspay/hyperswitch/blob/main/docs/CONTRIBUTING.md)

### Are you willing to submit a PR?

None

kashif-m:
Re-opening this for a bug introduced in the above PR.

Bug report:
- Client secret returned in the API response is incorrect
- Payout links generated in the API response is incorrect

ToDos:
- Use `get_string_repr()` for serializing the payout IDs in all the places - rather than the default formatter.

Issue #1321, opened 2023-05-31
[refactoring/enhancement] update payout_id to PayoutIdType similar to PaymentIdType


kashif-m:
Type of `payout_id` in PayoutCreateRequest needs to be updated to a more constrained type similar to `PaymentIdType`

kashif-m:
Closing this as it was done in pulls/8395

PR #8395, opened 2025-06-20
feat(payouts): add domain type for PayoutId

kashif-m:
feat(payouts): allow payout_id to be configurable in API request
feat(payouts): add merchant_order_reference_id for referencing payouts

## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [x] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description
This pull request introduces two main features for payouts:

1.  **Configurable `payout_id` in API Requests**:
    *   The `payout_id` field in payout creation API requests can now be optionally provided by the merchant. If provided, this ID will be used for the payout. If not provided, the system will generate one.
    *   Internally, the usage of `payout_id` has been standardized to use the strongly-typed `common_utils::id_type::PayoutId` instead of `String`. This change has been propagated throughout relevant structs and functionalities like `PayoutsData`, data handling logic in `crates/router/src/core/utils.rs`, test utilities and DB queries.

2.  **`merchant_order_reference_id` for Payout Referencing**:
    *   The `merchant_order_reference_id` field has been added/updated in payout API models (e.g., `PayoutListFilterConstraints` in `crates/api_models/src/payouts.rs`). This allows merchants to use their own unique identifier to reference payouts.

**Refactoring and Connector Updates**:
*   To support the `id_type::PayoutId` change and ensure correct data flow, connector integrations (Adyen, Adyenplatform, Cybersource, Nomupay) have been refactored.
*   Connectors now consistently use `RouterData.connector_request_reference_id` (which typically holds the `payout_attempt_id` as a `String`) when a string-based reference is required by the external service, rather than converting the internal `PayoutsData.payout_id`. This ensures that connectors receive the appropriate string identifier (e.g., `payout_attempt_id`) they expect.

### Additional Changes

- [x] This PR modifies the API contract.
- [x] This PR modifies the database schema.

## Motivation and Context
These changes address the need for greater flexibility and type safety in payout processing, as outlined in GitHub issues #8392 and #1321.

*   **Configurable `payout_id`**: Allows merchants more control over their payout identifiers within the Hyperswitch.
*   **`merchant_order_reference_id`**: Provides a dedicated field for merchants to use their internal order or reference IDs.
*   **`PayoutId` Type Standardization**: Using `common_utils::id_type::PayoutId` enhances code robustness and maintainability by leveraging Rust's type system to prevent errors related to ID handling.
*   **Connector Refactoring**: Ensures that connectors use the correct string-based identifiers (like `payout_attempt_id` via `connector_request_reference_id`) required by external payout processors, maintaining correctness of integrations.

Links to issues:
*   https://github.com/juspay/hyperswitch/issues/8392
*   https://github.com/juspay/hyperswitch/issues/1321

## How did you test it?

To be added

## Checklist


crates/common_utils/src/id_type/payout.rs
```diff
@@ -0,0 +1,20 @@
+crate::id_type!(
+    PayoutId,
+    "A type for payout_id that can be used for payout ids"
+);
+crate::impl_id_type_methods!(PayoutId, "payout_id");
+
+// This is to display the `PayoutId` as PayoutId(abcd)
+crate::impl_debug_id_type!(PayoutId);
+crate::impl_default_id_type!(PayoutId, "payout");
```
SanchithHegde: Do we necessarily need the `Default` impl, considering that the `Default` impl generated by the macro generates a random payout ID, and not a static value like an empty string?

https://github.com/juspay/hyperswitch/blob/a31745d2a4d2009c32759813d547e296f54b0d81/crates/common_utils/src/macros.rs#L249-L257

I'd recommend against it...

Can we use `impl_generate_id_id_type!()` instead, for generating a random ID?

crates/common_utils/src/id_type/payout.rs
```diff
@@ -0,0 +1,20 @@
...
+
+// Database related implementations so that this field can be used directly in the database tables
+crate::impl_queryable_id_type!(PayoutId);
+crate::impl_to_sql_from_sql_id_type!(PayoutId);
+
+impl std::fmt::Display for PayoutId {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{}", self.get_string_repr())
+    }
+}
```
SanchithHegde: Why do we need this implementation, considering that we haven't implemented `Display` for the other ID types like `PaymentId`?

crates/hyperswitch_connectors/src/connectors/wise/transformers.rs
```diff
@@ -507,10 +509,16 @@ impl<F> TryFrom<&PayoutsRouterData<F>> for WisePayoutCreateRequest {
                         field_name: "profile",
                     }
                 })?;
+
+                let customer_transaction_id =
+                    uuid::Uuid::parse_str(&item.connector_request_reference_id)
+                        .map(|parsed_uuid| parsed_uuid.to_string())
+                        .unwrap_or_else(|_| uuid::Uuid::new_v4().to_string());
```
SanchithHegde: Why are we trying to parse this as UUID?

crates/storage_impl/src/payouts/payouts.rs
```diff
@@ -498,14 +492,13 @@ impl<T: DatabaseStore> PayoutsInterface for crate::RouterStore<T> {
                     query = query.filter(po_dsl::profile_id.eq(profile_id.clone()));
                 }
 
-                query = match (params.starting_at, &params.starting_after_id) {
+                query = match (params.starting_at, params.starting_after_id.as_ref()) {
                     (Some(starting_at), _) => query.filter(po_dsl::created_at.ge(starting_at)),
-                    (None, Some(starting_after_id)) => {
-                        // TODO: Fetch partial columns for this query since we only need some columns
```
SanchithHegde, line 504: Why was this TODO comment removed, considering that you've not addressed it?

migrations/2025-06-19-170656_alter_payout_primary_key/up.sql
```diff
@@ -0,0 +1,7 @@
+ALTER TABLE payout_attempt DROP CONSTRAINT payout_attempt_pkey;
+ALTER TABLE payout_attempt ADD PRIMARY KEY (payout_attempt_id, merchant_id);
```
SanchithHegde: Nit: This would have to match the order of the columns used in the query as well, as in, the DB queries would also have to query using (payout_attempt_id, merchant_id), not (merchant_id, payout_attempt_id), as the index would not be used in the latter case (as far as I understand).

crates/api_models/src/payouts.rs
```diff
@@ -37,13 +37,17 @@ pub struct PayoutCreateRequest {
         example = "187282ab-40ef-47a9-9206-5099ba31e432"
     )]
     #[remove_in(PayoutsCreateRequest, PayoutUpdateRequest, PayoutConfirmRequest)]
-    pub payout_id: Option<String>, // TODO: #1321 https://github.com/juspay/hyperswitch/issues/1321
+    pub payout_id: Option<id_type::PayoutId>,
```
jarnura, line 40: If we are planning to move like paymentId, then this id merchant shouldn't send, this id will be created by use only and not by anyone outside.



There is another field called merchant_reference_id, which merchant can pass their id's

kashif-m: We have both,

For V2 we can proceed with this approach. For V1, we will need to keep the behaviour same as payments resource - which is to allow using custom payout IDs - this shouldn't be a problem since we don't run any validation on this other than the max length check.

crates/hyperswitch_connectors/src/connectors/wise/transformers.rs
kashif-m: Wise always needs to have an UUID as it's unique resource ID*, so

1. Try to parse the incoming `payout_id` as UUID (in case merchants also pass it in that format); if not
2. Create a new UUID

migrations/2025-06-19-170656_alter_payout_primary_key/up.sql
kashif-m: Thanks for pointing this one out!

crates/common_utils/src/id_type/payout.rs
kashif-m: It was being used in some places, removed it completely.

su-shivanshmathur (dismissed):

Sarthak1799 (dismissed):
Routing file changes looks okay

crates/hyperswitch_connectors/src/connectors/wise/transformers.rs
jarnura: Connector side we should not create any uuid's

kashif-m:
https://github.com/juspay/hyperswitch/pull/8395#discussion_r2175139028

Not possible currently @jarnura

Adding an issue for this - https://github.com/juspay/hyperswitch/issues/8505

crates/hyperswitch_connectors/src/connectors/wise/transformers.rs
kashif-m: Added an issue for this - https://github.com/juspay/hyperswitch/issues/8505

kashif-m: Removed it for now, will pick up https://github.com/juspay/hyperswitch/issues/8505

jarnura (approved):

su-shivanshmathur (approved):

su-shivanshmathur (approved):

jarnura (approved):

Sarthak1799 (approved):

PR #8434, opened 2025-06-23
fix(recovery): Populate connector request reference id in revenue recovery record attempt flow.

srujanchikke:
## Type of Change

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

Previously there was no restriction in confirmData to connector request reference id to be mandatory. Recovery internally uses proxy flow, which internally uses confirmData where connector request reference id was hardcoded to none. This Pr adds supports to generate connector request reference to merchant reference id. 

This Pr also has changes of adding support for jwt auth to payment get for dashboard use case. Previously payment sync was used in dashboard to retrieve payment details, but this would fail now in v2 because if there are no attempts created payment get would return 4xx since active attempt is mandotory. To solve this now dashboard is relying on payment get and payment get attempt list api's .

### Additional Changes

## How did you test it?

Test case 1 : 
**step 1 :** create a profile in hyperswitch and wait for 8 mins to move it from monitering to cascading , this would only get updated when webhooks are triggered .
**step 2 :** create payment mca
```
curl --location 'http://localhost:8080/v2/connector-accounts' \
--header 'x-merchant-id: test_seller_AGZCJiHwUNV1seyoepOE' \
--header 'x-profile-id: pro_kYmv4lQjIZFbAqoOVY4M' \
--header 'Authorization: admin-api-key=test_admin' \
--data '{
  "connector_name": "stripe",
        "auth_type": "HeaderKey",
        "api_key": "stripe-api-key"
    },
  "profile_id": "pro_kYmv4lQjIZFbAqoOVY4M"
}'
``` 
**step 4 : ** configure stripe billing connector. 
```
curl --location 'http://localhost:8080/v2/connector-accounts' \
--header 'x-merchant-id: test_seller_AGZCJiHwUNV1seyoepOE' \
--header 'x-profile-id: pro_kYmv4lQjIZFbAqoOVY4M' \
--header 'Authorization: admin-api-key=test_admin' \
--data '{
    "connector_type": "billing_processor",
    "connector_name": "stripebilling",
        "auth_type": "HeaderKey",
        "api_key": "{{stripe-api-key}}"
    },
    "connector_webhook_details": {
        "merchant_secret": "{{stripe-webhook-key}}",
        "additional_secret": "password"
    },
    "feature_metadata": {
        "revenue_recovery": {
            "max_retry_count": 7,
            "billing_connector_retry_threshold": 2,
            "billing_account_reference": {
                "{{payment_connector_mca_created_in_last_step}}": "stripebilling"
            }
        }
    },
    "profile_id": "pro_kYmv4lQjIZFbAqoOVY4M"
}'
```
step 5 : Trigger failed invoice from stripe billing by creating new subscription and move the test clock to 1 week.
This would trigger few external payments. once it exhausts retry threshold it will create process tracker entry. Follwed by 2 mins it would create all retries by hyperswitch as shown below.

Screenshot 2025-06-23 at 7 00 57 PM

## Checklist

### DIFF
--- a/crates/router/src/services/authentication.rs
+++ b/crates/router/src/services/authentication.rs
@@ -2698,7 +2698,27 @@ where
         api_auth
     }
 }
-
+#[cfg(feature = "v2")]
+pub fn api_or_client_or_jwt_auth<'a, T, A>(
+    api_auth: &'a dyn AuthenticateAndFetch<T, A>,
+    client_auth: &'a dyn AuthenticateAndFetch<T, A>,
+    jwt_auth: &'a dyn AuthenticateAndFetch<T, A>,
+    headers: &HeaderMap,
+) -> &'a dyn AuthenticateAndFetch<T, A>
+where
+{
+    if let Ok(val) = HeaderMapStruct::new(headers).get_auth_string_from_header() {
+        if val.trim().starts_with("api-key=") {
+            api_auth
+        } else if is_jwt_auth(headers) {
+            jwt_auth
+        } else {
+            client_auth
+        }
+    } else {
+        api_auth
+    }
+}
 #[derive(Debug)]
 pub struct PublishableKeyAuth;
 
--- a/crates/hyperswitch_domain_models/src/payments/payment_attempt.rs
+++ b/crates/hyperswitch_domain_models/src/payments/payment_attempt.rs
@@ -646,8 +646,10 @@ impl PaymentAttempt {
         let payment_method_type_data = payment_intent.get_payment_method_type();
 
         let payment_method_subtype_data = payment_intent.get_payment_method_sub_type();
-
         let authentication_type = payment_intent.authentication_type.unwrap_or_default();
+        let connector_request_reference_id = payment_intent
+.merchant_reference_id.clone()
+.map(|id| id.get_string_repr().to_owned());
         Ok(Self {
             payment_id: payment_intent.id.clone(),
             merchant_id: payment_intent.merchant_id.clone(),
@@ -698,7 +700,7 @@ impl PaymentAttempt {
             card_discovery: None,
             processor_merchant_id: payment_intent.merchant_id.clone(),
             created_by: None,
-            connector_request_reference_id: None,
+            connector_request_reference_id,
         })
     }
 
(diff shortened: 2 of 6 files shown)


crates/router/src/services/authentication.rs
```diff
@@ -2698,7 +2698,27 @@ where
...
+pub fn api_or_client_or_jwt_auth<'a, T, A>(
+    api_auth: &'a dyn AuthenticateAndFetch<T, A>,
+    client_auth: &'a dyn AuthenticateAndFetch<T, A>,
+    jwt_auth: &'a dyn AuthenticateAndFetch<T, A>,
+    headers: &HeaderMap,
+) -> &'a dyn AuthenticateAndFetch<T, A>
+where
+{
+    if let Ok(val) = HeaderMapStruct::new(headers).get_auth_string_from_header() {
+        if val.trim().starts_with("api-key=") {
```
hrithikesh026, line 2711: nit: `"api-key="` can be a const.

crates/router/src/services/authentication.rs
```diff
@@ -2698,7 +2698,27 @@ where
...
+            api_auth
+        } else if is_jwt_auth(headers) {
+            jwt_auth
+        } else {
+            client_auth
+        }
+    } else {
+        api_auth
+    }
+}
```
tsdk02, line 2721: Are you giving the highest priority to api-key?
What if both `JWT` and `api-key` headers are present in the request body? Which authentication type should be considered in that case?

Or is it the same header used for all authentications in v2?

hrithikesh026: This can never happen right. 
Value for `Authorization` header cannot start with Both` Bearer_` and `api-key=`

tsdk02: Got it, so unlike v1, where there are different headers for different authentication types, in v2 there is just one Authorization header with different values for different types of authentication, correct?

ShankarSinghC (approved):

PR #8458, opened 2025-06-25
feat(core): Added prost::Message trait implementation for Secret<String>

Debarshi-Gupta:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

### Additional Changes

## Checklist

### DIFF
--- a/crates/masking/Cargo.toml
+++ b/crates/masking/Cargo.toml
@@ -13,6 +13,8 @@ alloc = ["zeroize/alloc"]
 serde = ["dep:serde", "dep:serde_json"]
 time = ["dep:time"]
 cassandra = ["dep:scylla"]
+tonic = ["dep:bytes", "dep:prost"]
+bytes = ["dep:bytes"]
 
 [package.metadata.docs.rs]
 all-features = true
@@ -29,6 +31,7 @@ time = { version = "0.3.41", optional = true, features = ["serde-human-readable"
 url = { version = "2.5.4", features = ["serde"] }
 zeroize = { version = "1.8", default-features = false }
 scylla = { git = "https://github.com/juspay/scylla-rust-driver.git",rev = "5700aa2847b25437cdd4fcf34d707aa90dca8b89", optional = true}
+prost = { version = "0.13", optional = true }
 
 [dev-dependencies]
 serde_json = "1.0.140"
--- a/crates/masking/src/secret.rs
+++ b/crates/masking/src/secret.rs
@@ -430,3 +430,36 @@ mod tests {
         );
     }
 }
+
+
+#[cfg(feature = "tonic")]
+impl prost::Message for Secret<String, crate::WithType> {
+    fn encode_raw(&self, buf: &mut impl bytes::BufMut) {
+        prost::encoding::string::encode(1, self.peek(), buf);
+    }
+
+    fn merge_field(
+        &mut self,
+        tag: u32,
+        wire_type: prost::encoding::WireType,
+        buf: &mut impl bytes::Buf,
+        ctx: prost::encoding::DecodeContext,
+    ) -> Result<(), prost::DecodeError> {
+        if tag == 1 {
+            let mut value = String::new();
+            prost::encoding::string::merge(wire_type, &mut value, buf, ctx)?;
+            *self = Secret::new(value);
+            Ok(())
+        } else {
+            prost::encoding::skip_field(wire_type, tag, buf, ctx)
+        }
+    }
+
+    fn encoded_len(&self) -> usize {
+        prost::encoding::string::encoded_len(1, self.peek())
+    }
+
+    fn clear(&mut self) {
+        *self = Secret::new(String::new());
+    }
+}
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4682,6 +4682,7 @@ dependencies = [
  "bytes 1.10.1",
  "diesel",
  "erased-serde 0.4.6",
+ "prost",
  "scylla",
  "serde",
  "serde_json",


crates/masking/Cargo.toml
```diff
@@ -13,6 +13,8 @@ alloc = ["zeroize/alloc"]
 serde = ["dep:serde", "dep:serde_json"]
 time = ["dep:time"]
 cassandra = ["dep:scylla"]
+proto_tonic = ["dep:bytes", "dep:prost"]
+bytes = ["dep:bytes"]
```
SanchithHegde, line 17: Nit: Declaring the `bytes` feature isn't absolutely necessary.

crates/masking/src/secret.rs
```diff
@@ -248,6 +248,41 @@ impl Strategy<serde_json::Value> for JsonMaskStrategy {
...
 }
 
+#[cfg(feature = "proto_tonic")]
+impl<T> prost::Message for Secret<T, crate::WithType>
+where
+    T: prost::Message + Default + Clone,
+{
+    fn encode_raw(&self, buf: &mut impl bytes::BufMut) {
+        prost::Message::encode_raw(self.peek(), buf);
+    }
```
SanchithHegde: Nit: Can we call
```suggestion
    fn encode_raw(&self, buf: &mut impl bytes::BufMut) {
        self.peek().encode_raw(buf);
    }
```
or would the compiler complain about `prost::Message` not being in scope?

crates/masking/src/secret.rs
```diff
@@ -248,6 +248,41 @@ impl Strategy<serde_json::Value> for JsonMaskStrategy {
...
+
+    fn merge_field(
+        &mut self,
+        tag: u32,
+        wire_type: prost::encoding::WireType,
+        buf: &mut impl bytes::Buf,
+        ctx: prost::encoding::DecodeContext,
+    ) -> Result<(), prost::DecodeError> {
+        if tag == 1 {
```
SanchithHegde, line 267: Do we have a reason why we're checking `tag == 1`, other than being based on existing implementations of the trait?

Debarshi-Gupta: Yes, if we are trying to use a proto message where we have multiple fields instead of one, then only that field which has a serialiser value of 1 will be used, other fields will be skipped. If we don't have this condition, then it becomes mandatory to have proto messages with only a single field, if a proto has multiple fields then it will lead to error.

crates/masking/src/secret.rs
Debarshi-Gupta: I have resolved it, thanks.

crates/masking/Cargo.toml
Debarshi-Gupta: If I don't declare this feature, then I am getting warnings. I believe that declaring `bytes` as feature is not necessary only if `dep:bytes` was only used in `bytes` feature, but since I am using `dep:bytes` in `proto_tonic` now, adding `bytes` as feature is now a thing. Absolutely not sure about this, just guessed. 

image

SanchithHegde: I think I have an idea why the warning was thrown only after including `dep:bytes` in `proto_tonic` feature.

From the [docs](https://doc.rust-lang.org/cargo/reference/features.html#optional-dependencies):
> By default, this optional dependency implicitly defines a feature [...]

and

> If you specify the optional dependency with the `dep:` prefix anywhere in the `[features]` table, that disables the implicit feature.

No worries, will be resolving this conversation.

crates/masking/src/secret.rs
knutties: @Debarshi-Gupta - `merge_field` is not one of the functions of the [Message trait in version 0.13.5](https://docs.rs/prost/0.13.5/prost/trait.Message.html).  Who will call it actually?

SanchithHegde: @knutties `merge_field` and `encode_raw` have the `#[doc(hidden)]` attribute, which doesn't include them in the docs. But neither of them have a provided default implementation, so implementors would have to implement these two methods as well.

https://docs.rs/prost/0.13.5/src/prost/message.rs.html#23-40

crates/masking/src/secret.rs
```diff
@@ -248,6 +248,41 @@ impl Strategy<serde_json::Value> for JsonMaskStrategy {
...
+        }
+    }
+
+    fn encoded_len(&self) -> usize {
+        prost::Message::encoded_len(self.peek())
+    }
+
+    fn clear(&mut self) {
+        *self = Self::new(T::default());
+    }
```
SanchithHegde: You should be able to do something like this for `encoded_len()`. And can we call `clear()` on the underlying type, instead of assigning it to `T::default()`?

```suggestion
    fn encoded_len(&self) -> usize {
        self.peek().encoded_len()
    }

    fn clear(&mut self) {
        self.peek_mut().clear()
    }
```

crates/masking/src/strong_secret.rs
```diff
@@ -126,3 +126,41 @@ impl StrongEq for Vec<u8> {
...
+        }
+    }
+
+    fn encoded_len(&self) -> usize {
+        self.inner_secret.encoded_len()
+    }
+
+    fn clear(&mut self) {
+        self.inner_secret = T::default();
+    }
```
SanchithHegde: Can we offload all of the implementations on underlying `Secret`, i.e., the `inner_secret` field? Both for `merge_secret()` and `clear()`.

SanchithHegde (commented):
Have we tried out these implementations by integrating them in the connector-service repo, running the gRPC server and confirming that data is being transmitted and received correctly?

crates/masking/src/strong_secret.rs
Debarshi-Gupta: I have resolved it, thanks.

crates/masking/src/secret.rs
Debarshi-Gupta: I have resolved it, thanks.

crates/masking/src/strong_secret.rs
```diff
@@ -126,3 +126,32 @@ impl StrongEq for Vec<u8> {
...
+
+    fn merge_field(
+        &mut self,
+        tag: u32,
+        wire_type: prost::encoding::WireType,
+        buf: &mut impl bytes::Buf,
+        ctx: prost::encoding::DecodeContext,
+    ) -> Result<(), prost::DecodeError> {
+        self.peek_mut().merge_field(tag, wire_type, buf, ctx)?;
+        Ok(())
```
SanchithHegde: Nit:
```suggestion
        self.peek_mut().merge_field(tag, wire_type, buf, ctx)
```

crates/masking/src/secret.rs
```diff
@@ -248,6 +248,35 @@ impl Strategy<serde_json::Value> for JsonMaskStrategy {
...
+
+    fn merge_field(
+        &mut self,
+        tag: u32,
+        wire_type: prost::encoding::WireType,
+        buf: &mut impl bytes::Buf,
+        ctx: prost::encoding::DecodeContext,
+    ) -> Result<(), prost::DecodeError> {
+        self.peek_mut().merge_field(tag, wire_type, buf, ctx)?;
+        Ok(())
```
SanchithHegde: Nit:

```suggestion
        self.peek_mut().merge_field(tag, wire_type, buf, ctx)
```

crates/masking/src/strong_secret.rs
```diff
@@ -126,3 +126,32 @@ impl StrongEq for Vec<u8> {
...
+    }
+
+    fn encoded_len(&self) -> usize {
+        self.peek().encoded_len()
+    }
+
+    fn clear(&mut self) {
+        self.peek_mut().clear();
```
SanchithHegde, line 158: Nit: Either keep the trailing semicolon in all methods returning unit (`()`) (such as `encoded_len()` and `encode_raw()`), or don't add it in any of them.

crates/masking/src/secret.rs
```diff
@@ -248,6 +248,39 @@ impl Strategy<serde_json::Value> for JsonMaskStrategy {
...
+    fn merge_field(
+        &mut self,
+        tag: u32,
+        wire_type: prost::encoding::WireType,
+        buf: &mut impl bytes::Buf,
+        ctx: prost::encoding::DecodeContext,
+    ) -> Result<(), prost::DecodeError> {
+        if tag == 1 {
+            self.peek_mut().merge_field(tag, wire_type, buf, ctx)?;
+            Ok(())
```
SanchithHegde: Nit:
```suggestion
            self.peek_mut().merge_field(tag, wire_type, buf, ctx)
```

crates/masking/src/strong_secret.rs
```diff
@@ -126,3 +126,36 @@ impl StrongEq for Vec<u8> {
...
+    fn merge_field(
+        &mut self,
+        tag: u32,
+        wire_type: prost::encoding::WireType,
+        buf: &mut impl bytes::Buf,
+        ctx: prost::encoding::DecodeContext,
+    ) -> Result<(), prost::DecodeError> {
+        if tag == 1 {
+            self.peek_mut().merge_field(tag, wire_type, buf, ctx)?;
+            Ok(())
```
SanchithHegde: Same here.

```suggestion
            self.peek_mut().merge_field(tag, wire_type, buf, ctx)
```

crates/masking/src/secret.rs
Debarshi-Gupta: I have resolved it, thanks.

crates/masking/src/strong_secret.rs
Debarshi-Gupta: I have resolved it, thanks.

SanchithHegde (approved):

PR #8477, opened 2025-06-27
feat(core): add validation for key identifier 

Sakilmostak:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

- Add validation method for Conversion Trait
- The new method does validation of the key identifier for the given entity
- Add this method on all present implementation of Conversion Trait
- Add new Validation Error type i.e Decryption Error
- The new error type is used to propagate decryption related error
- Added error conversion from Validation Error to Storage Error for better context on the error

### Additional Changes

## How did you test it?

Tested through Postman:

Create an Merchant Account (M1) and Business Profile (P1) along with it:

- curl for merchant account
```curl --location '{{baseUrl}}/accounts' \
--header 'api-key: {{api_key}}' \
--data-raw '{
    "merchant_id": "merchant_1751535362",
    "locker_id": "m0010",
    "merchant_details": {
        "website": "www.example.com",
        "address": {
            "line1": "1467",
            "line2": "Harrison Street",
            "line3": "Harrison Street",
            "city": "San Fransico",
            "state": "California",
            "zip": "94122",
            "country": "US"
        }
    },
    "webhook_details": {
        "webhook_version": "1.0.1",
    },
    "metadata": {
        "city": "NY",
        "unit": "245"
    },
    "primary_business_details": [
        {
            "country": "US",
            "business": "default"
        }
    ]
}'
```
- curl for business profile create
```
curl --location '{{baseUrl}}/account/{{merchant_id}}/business_profile' \
--header 'api-key:{{api_key}}' \
--data '{
  "profile_name": "food"
}'
```

Create another Merchant Account (M2) and try to retrieve Business Profile P1 with it:
- curl for profile retrieve
```
curl --location '{{baseUrl}}/account/{{merchant_id}}/business_profile/{{profile_id}}' \
--header 'api-key:{{api_key}}'
```

Following should be the error response (4xx):

```
{
    "error": {
        "type": "invalid_request",
        "message": "Business profile with the given id pro_skzBddaDpo9jRD7fTR28 does not exist",
        "code": "HE_02"
    }
}
```

## Checklist

### DIFF
--- a/crates/common_utils/src/errors.rs
+++ b/crates/common_utils/src/errors.rs
@@ -69,6 +69,10 @@ pub enum ValidationError {
     /// An invalid input was provided.
     #[error("{message}")]
     InvalidValue { message: String },
+
+    /// An error occurred while decrypting the input.
+    #[error("Error while decrypting data: {message}")]
+    DecryptionError { message: String },
 }
 
 /// Integrity check errors.
--- a/crates/hyperswitch_domain_models/src/behaviour.rs
+++ b/crates/hyperswitch_domain_models/src/behaviour.rs
@@ -12,8 +12,8 @@ pub trait Conversion {
     async fn convert(self) -> CustomResult<Self::DstType, ValidationError>;
 
     fn validate(
-        item: Self::DstType,
-        key_manager_identifier: Identifier,
+        item: &Self::DstType,
+        key_manager_identifier: &Identifier,
     ) -> CustomResult<(), ValidationError>;
 
     async fn convert_back(
@@ -39,14 +39,14 @@ pub trait ReverseConversion<SrcType: Conversion> {
 }
 
 #[async_trait::async_trait]
-impl<T: Send + Clone, U: Conversion<DstType = T>> ReverseConversion<U> for T {
+impl<T: Send, U: Conversion<DstType = T>> ReverseConversion<U> for T {
     async fn convert(
         self,
         state: &KeyManagerState,
         key: &Secret<Vec<u8>>,
         key_manager_identifier: Identifier,
     ) -> CustomResult<U, ValidationError> {
-        U::validate(self.clone(), key_manager_identifier.clone())?;
+        U::validate(&self, &key_manager_identifier)?;
         U::convert_back(state, self, key, key_manager_identifier).await
     }
 }
--- a/crates/hyperswitch_domain_models/src/behaviour.rs
+++ b/crates/hyperswitch_domain_models/src/behaviour.rs
@@ -11,6 +11,11 @@ pub trait Conversion {
     type NewDstType;
     async fn convert(self) -> CustomResult<Self::DstType, ValidationError>;
 
+    fn validate(
+        item: Self::DstType,
+        key_manager_identifier: Identifier,
+    ) -> CustomResult<(), ValidationError>;
+
     async fn convert_back(
         state: &KeyManagerState,
         item: Self::DstType,
@@ -34,13 +39,14 @@ pub trait ReverseConversion<SrcType: Conversion> {
 }
 
 #[async_trait::async_trait]
-impl<T: Send, U: Conversion<DstType = T>> ReverseConversion<U> for T {
+impl<T: Send + Clone, U: Conversion<DstType = T>> ReverseConversion<U> for T {
     async fn convert(
         self,
         state: &KeyManagerState,
         key: &Secret<Vec<u8>>,
         key_manager_identifier: Identifier,
     ) -> CustomResult<U, ValidationError> {
+        U::validate(self.clone(), key_manager_identifier.clone())?;
         U::convert_back(state, self, key, key_manager_identifier).await
     }
 }
--- a/crates/hyperswitch_domain_models/src/business_profile.rs
+++ b/crates/hyperswitch_domain_models/src/business_profile.rs
@@ -821,6 +821,28 @@ impl super::behaviour::Conversion for Profile {
         })
     }
 
+    fn validate(
+        item: Self::DstType,
+        key_manager_identifier: keymanager::Identifier,
+    ) -> CustomResult<(), ValidationError>
+    where
+        Self: Sized,
+    {
+        match key_manager_identifier {
+            keymanager::Identifier::Merchant(merchant_id) => {
+                if item.merchant_id != merchant_id {
+                    return Err(ValidationError::IncorrectValueProvided {
+                        field_name: "Profile ID",
+                    }
+                    .into());
+                }
+
+                Ok(())
+            }
+            _ => Ok(()),
+        }
+    }
+
     async fn convert_back(
         state: &keymanager::KeyManagerState,
         item: Self::DstType,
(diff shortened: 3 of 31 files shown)


crates/hyperswitch_domain_models/src/behaviour.rs
```diff
@@ -11,6 +11,11 @@ pub trait Conversion {
     type NewDstType;
     async fn convert(self) -> CustomResult<Self::DstType, ValidationError>;
 
+    fn validate(
+        item: Self::DstType,
```
SanchithHegde: Can we accept a reference instead of the owned value?

```suggestion
        item: &Self::DstType,
```

crates/hyperswitch_domain_models/src/behaviour.rs
```diff
@@ -34,13 +39,14 @@ pub trait ReverseConversion<SrcType: Conversion> {
...
 #[async_trait::async_trait]
-impl<T: Send, U: Conversion<DstType = T>> ReverseConversion<U> for T {
+impl<T: Send + Clone, U: Conversion<DstType = T>> ReverseConversion<U> for T {
     async fn convert(
         self,
         state: &KeyManagerState,
         key: &Secret<Vec<u8>>,
         key_manager_identifier: Identifier,
     ) -> CustomResult<U, ValidationError> {
+        U::validate(self.clone(), key_manager_identifier.clone())?;
```
SanchithHegde: And then we can avoid the need to clone here.

Can we avoid the key manager identifier clone as well?

crates/common_utils/src/errors.rs
```diff
@@ -69,6 +69,10 @@ pub enum ValidationError {
     /// An invalid input was provided.
     #[error("{message}")]
     InvalidValue { message: String },
+
+    /// An error occurred while decrypting the input.
+    #[error("Error while decrypting data: {message}")]
+    DecryptionError { message: String },
```
SanchithHegde, line 75: Ummm, a decryption error doesn't seem like a validation error however.

We can consider returning a different error type altogether from the trait methods, if that's an option. Or can you think of anything else?

crates/hyperswitch_domain_models/src/business_profile.rs
```diff
@@ -821,6 +821,28 @@ impl super::behaviour::Conversion for Profile {
...
+        match key_manager_identifier {
+            keymanager::Identifier::Merchant(merchant_id) => {
+                if item.merchant_id != merchant_id {
+                    return Err(ValidationError::IncorrectValueProvided {
+                        field_name: "Profile ID",
+                    }
+                    .into());
+                }
+
+                Ok(())
```
SanchithHegde: 1. To be precise, this isn't the case that an incorrect value (profile ID) was provided, it's that the merchant cannot / should not access the profile, right?
2. Can probably use `when()` instead.

crates/hyperswitch_domain_models/src/business_profile.rs
```diff
@@ -821,6 +821,28 @@ impl super::behaviour::Conversion for Profile {
...
+            }
+            _ => Ok(()),
```
SanchithHegde: In my opinion, if we get any other variant of the keymanager identifier here, that's unexpected itself, we should probably treat it as an error. What do you say? 

Also, let's not use wildcard `match` arms (`_`) please.

crates/storage_impl/src/errors.rs
```diff
@@ -70,6 +71,21 @@ impl From<error_stack::Report<DatabaseError>> for StorageError {
...
+impl From<error_stack::Report<ValidationError>> for StorageError {
+    fn from(err: error_stack::Report<ValidationError>) -> Self {
+        match err.current_context() {
+            ValidationError::InvalidValue { message } => Self::ValueNotFound(message.to_string()),
+            ValidationError::MissingRequiredField { field_name } => {
+                Self::ValueNotFound(format!("Missing required field: {field_name}"))
+            }
+            ValidationError::IncorrectValueProvided { field_name } => {
+                Self::ValueNotFound(format!("Incorrect Value for field: {field_name}"))
+            }
```
SanchithHegde, line 83: The error mapping here doesn't seem right, the missing required field case stands out clearly, for example.

crates/storage_impl/src/errors.rs
```diff
@@ -70,6 +71,21 @@ impl From<error_stack::Report<DatabaseError>> for StorageError {
...
+            ValidationError::DecryptionError { .. } => Self::DecryptionError,
```
SanchithHegde, line 84: And in this case, we're losing the context about the original error (the data within the `ValidationError::DecryptionError` variant.

crates/storage_impl/src/kv_router_store.rs
```diff
@@ -269,7 +269,7 @@ impl<T: DatabaseStore> KVRouterStore<T> {
                 key_store.merchant_id.clone().into(),
             )
             .await
-            .change_context(errors::StorageError::DecryptionError)
+            .map_err(|error| error_stack::report!(errors::StorageError::from(error)))
```
SanchithHegde, line 272: And because there's a possibility of losing context about the original error when calling `map_err()`, I'd prefer `change_context()` being called instead.

SanchithHegde (commented):
One of the concerns I have is about having proper error propagation and reporting, which is necessary for debugging. We shouldn't compromise on that.

Also, @apoorvdixit88, how are we handling data decryption in the platform merchant use case, which merchant's key would be used when encrypting / decrypting data?

crates/common_utils/src/errors.rs
Sakilmostak: I asked gemini for some suggestions, some of them seems valid, listing them out here:

`InvalidCiphertext`: A variant for when decryption specifically fails. This clearly points to a decryption problem.

`BusinessRuleViolation`: For when the data, though correctly formatted, violates a specific business rule.

crates/hyperswitch_domain_models/src/behaviour.rs
```diff
@@ -11,6 +11,11 @@ pub trait Conversion {
     type NewDstType;
     async fn convert(self) -> CustomResult<Self::DstType, ValidationError>;
 
+    fn validate(
```
XyneSpaces, line 14: :bulb: Consider adding default implementation for `validate` that returns `Ok(())` to avoid breaking existing implementations that don't need validation. This would make the trait backward-compatible.

XyneSpaces (commented):
## PR Summary

This PR introduces a `validate` method to the `Conversion` trait to enable key identifier validation for entities during the conversion process. The implementation:

1. Adds a new `validate` method to the `Conversion` trait
2. Introduces a new `DecryptionError` variant to `ValidationError` for better error propagation
3. Updates all existing `Conversion` trait implementations to include the new validation logic
4. Adds proper error conversion from `ValidationError` to `StorageError`

**Note:** Since the `validate` method is added as a required trait method, this is a **breaking change** for any downstream implementations of `Conversion`. See my inline comment regarding potential backward compatibility considerations.

PR #8559, opened 2025-07-06
refactor(routing): add conditional check for invoking DE routing flows

Sarthak1799:
## Type of Change

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

This PR introduces a conditional check for Decision Engine (DE) Euclid API integration within the static routing flow. Previously, the system always attempted to perform Euclid-based routing regardless of configuration. With this change, the Euclid routing flow is invoked only if the OpenRouter URL is configured.  

Additionally, some unused or redundant code around logging and comments has been cleaned up in `utils.rs`.  

## Outcomes

- Prevents unnecessary calls to the Euclid API when the OpenRouter URL is not configured, reducing overhead and potential timeouts.
- Ensures cleaner and more consistent logs by removing redundant debug statements and placeholders.
- Improves maintainability by scoping the Euclid integration behind a clear configuration flag.  

### `crates/router/src/core/payments/routing.rs`
- Wrapped the `perform_decision_euclid_routing` call with a check for `state.conf.open_router.url`.
- Added a fallback to return an empty connector list if OpenRouter is not configured.

### `crates/router/src/core/payments/routing/utils.rs`
- Removed a redundant placeholder comment about logging Euclid response events.
- Streamlined debug logs for Euclid responses.

This PR addresses an integration improvement for Euclid static routing checks in the decision engine.  

### Additional Changes

## How did you test it?

Can be tested directly in prod as that is the only environment where this env in missing.

## Checklist

### DIFF
--- a/crates/router/src/core/payments/routing.rs
+++ b/crates/router/src/core/payments/routing.rs
@@ -490,7 +490,7 @@ pub async fn perform_static_routing_v1(
             .to_string(),
     };
 
-    let de_euclid_connectors = if !state.conf.open_router.url.is_empty() {
+    let de_euclid_connectors = if state.conf.open_router.static_routing_enabled {
         let routing_events_wrapper = utils::RoutingEventsWrapper::new(
             state.tenant.tenant_id.clone(),
             state.request_id,
--- a/config/config.example.toml
+++ b/config/config.example.toml
@@ -1094,7 +1094,8 @@ background_color = "#FFFFFF"                                                    
 billing_connectors_which_require_payment_sync = "stripebilling, recurly" # List of billing connectors which has payment sync api call
 
 [open_router]
-enabled = true                          # Enable or disable Open Router
+dynamic_routing_enabled = true                          # Enable or disable Open Router for dynamic routing
+static_routing_enabled = true           # Enable or disable Open Router for static routing
 url = "http://localhost:8080"           # Open Router URL
 
 [grpc_client.unified_connector_service]
--- a/config/development.toml
+++ b/config/development.toml
@@ -1189,7 +1189,8 @@ allow_connected_merchants = true
 click_to_pay = {connector_list = "adyen, cybersource"}
 
 [open_router]
-enabled = false
+dynamic_routing_enabled = false
+static_routing_enabled = false
 url = "http://localhost:8080"
 
 [grpc_client.unified_connector_service]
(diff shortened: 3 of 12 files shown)


prajjwalkumar17 (approved):

crates/router/src/core/payments/routing.rs
```diff
@@ -490,29 +490,34 @@ pub async fn perform_static_routing_v1(
...
-        state.request_id,
-        payment_id,
-        business_profile.get_id().to_owned(),
-        business_profile.merchant_id.to_owned(),
-        "DecisionEngine: Euclid Static Routing".to_string(),
-        None,
-        true,
-        false,
-    );
+    // check this condition
```
Chethan-rao: ??

crates/router/src/core/payments/routing.rs
```diff
@@ -490,29 +490,33 @@ pub async fn perform_static_routing_v1(
...
-        state.request_id,
-        payment_id,
-        business_profile.get_id().to_owned(),
-        business_profile.merchant_id.to_owned(),
-        "DecisionEngine: Euclid Static Routing".to_string(),
-        None,
-        true,
-        false,
-    );
+    let de_euclid_connectors = if !state.conf.open_router.url.is_empty() {
```
SanchithHegde: Can we use `open_router.enabled` instead?

Sarthak1799: We want euclid for diff check even when DE is disabled. So if the env is present (the url i.e.) we'll make the call for euclid but not other algorithms as they have a `open_router.enabled` check.

SanchithHegde: Why such a distinction?

In my opinion, this makes it harder when debugging, since someone self-hosting may have set open router enabled to `false` with a placeholder URL (say referring to our sample config), expecting it to not call open router, but it would still try to?

https://github.com/juspay/hyperswitch/blob/102f2785199a2f66d9cd3428678f54ba3f015ff7/config/config.example.toml#L1098-L1100

Sarthak1799: Diff check is a short term process. We'll remove this check entirely in some time and rely completely on the `open_router.enabled` @SanchithHegde

Chethan-rao (approved):

PR #8586, opened 2025-07-09
feat(connector): [DWOLLA] Connector integration

Vani-1107:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

### Additional Changes

## Checklist


crates/hyperswitch_connectors/src/connectors/dwolla.rs
```diff
@@ -311,10 +457,37 @@ impl ConnectorIntegration<PSync, PaymentsSyncData, PaymentsResponseData> for Dwo
...
+        }
+    }
+
+    parsed
+        .get("TokenizationResponse")
+        .and_then(|v| v.get("token"))
+        .and_then(|v| v.as_str())
+        .map(|s| s.to_string())
+        .ok_or_else(|| report!(errors::ConnectorError::ResponseDeserializationFailed))
+}
```
deepanshu-iiitu: Move this to transformers.rs

crates/hyperswitch_connectors/src/connectors/dwolla.rs
```diff
@@ -326,34 +499,45 @@ impl ConnectorIntegration<Capture, PaymentsCaptureData, PaymentsResponseData> fo
...
+        req: &PaymentsAuthorizeRouterData,
+        connectors: &Connectors,
     ) -> CustomResult<RequestContent, errors::ConnectorError> {
-        Err(errors::ConnectorError::NotImplemented("get_request_body method".to_string()).into())
+        let amount_in_minor_unit = MinorUnit::new(req.request.amount);
+        let amount = convert_amount(
+            self.amount_converter,
+            amount_in_minor_unit,
+            req.request.currency,
+        )?;
```
deepanshu-iiitu: You can directly pass `req.request.minor_amount`
The above parsing is not needed.

crates/hyperswitch_connectors/src/connectors/dwolla/transformers.rs
```diff
@@ -1,228 +1,506 @@
...
+}
+
+impl TryFrom<&types::ConnectorCustomerRouterData> for DwollaCustomerRequest {
+    type Error = error_stack::Report<errors::ConnectorError>;
+    fn try_from(item: &types::ConnectorCustomerRouterData) -> Result<Self, Self::Error> {
+        Ok(Self {
+            first_name: item.get_billing_first_name().map_err(|_| {
+                errors::ConnectorError::MissingRequiredField {
+                    field_name: "first_name",
+                }
```
deepanshu-iiitu: you don't need to specify the error message
it is already handled in get_billing_first_name function.
also the same for get_billing_last_name and get_billing_email

crates/hyperswitch_connectors/src/connectors/dwolla/transformers.rs
```diff
@@ -1,228 +1,506 @@
...
+            "{}/funding-sources/{}",
+            item.base_url,
+            source_funding.expose()
+        );
+
+        let destination_url = format!(
+            "{}/funding-sources/{}",
+            item.base_url,
+            metadata.merchant_funding_source.expose()
+        );
```
deepanshu-iiitu, line 288: We should get destination url as input from merchant from dashboard?
Are we doing that or is merchant going to send destination url in payment request metadata?

crates/hyperswitch_connectors/src/connectors/dwolla/transformers.rs
```diff
@@ -1,228 +1,506 @@
...
+    fn from(item: DwollaPaymentStatus) -> Self {
         match item {
-            RefundStatus::Succeeded => Self::Success,
-            RefundStatus::Failed => Self::Failure,
-            RefundStatus::Processing => Self::Pending,
-            //TODO: Review mapping
+            DwollaPaymentStatus::Succeeded => Self::Charged,
+            DwollaPaymentStatus::Processed => Self::Charged,
+            DwollaPaymentStatus::Failed => Self::Failure,
+            DwollaPaymentStatus::Processing => Self::Authorizing,
```
deepanshu-iiitu: Why is Processing marked as Authorizing

crates/hyperswitch_connectors/src/connectors/dwolla/transformers.rs
```diff
@@ -1,228 +1,506 @@
...
+            "customer_transfer_completed" => Self::CustomerTransferCompleted,
+            "customer_transfer_failed" => Self::CustomerTransferFailed,
+            "transfer_created" => Self::TransferCreated,
+            "transfer_pending" => Self::TransferPending,
+            "transfer_processed" => Self::TransferProcessed,
+            "transfer_failed" => Self::TransferFailed,
+            _ => Self::Unknown,
+        }
+    }
+}
```
deepanshu-iiitu: Why do we need this transformation?

deepanshu-iiitu (changes requested):

crates/hyperswitch_connectors/src/connectors/dwolla/transformers.rs
Vani-1107: Since the topic is extracted as a string from the headers of the incoming webhook request, we use this transformation to convert the string into a defined webhook event type.

crates/hyperswitch_connectors/src/connectors/dwolla/transformers.rs
Vani-1107: We are getting the merchant's funding ID from the metadata in the MCA call .

crates/hyperswitch_connectors/src/connectors/dwolla/transformers.rs
Vani-1107: Fixed

crates/hyperswitch_connectors/src/connectors/dwolla.rs
```diff
@@ -543,26 +828,144 @@ impl ConnectorIntegration<RSync, RefundsData, RefundsResponseData> for Dwolla {
...
+    dwolla_supported_payment_methods
+});
+
+static DWOLLA_CONNECTOR_INFO: ConnectorInfo = ConnectorInfo {
+    display_name: "Dwolla",
+    description: "Dwolla is a multinational financial technology company offering financial services and software as a service (SaaS)",
+    connector_type: enums::PaymentConnectorCategory::PaymentGateway,
+};
+
+impl ConnectorSpecifications for Dwolla {
```
deepanshu-iiitu, line 992: Please add get_supported_webhook_flows function as well

crates/hyperswitch_connectors/src/connectors/dwolla/transformers.rs
```diff
@@ -1,228 +1,611 @@
...
+
+#[derive(Debug, Serialize)]
+pub struct DwollaFundingSourceRequest {
+    #[serde(rename = "routingNumber")]
+    routing_number: Secret<String>,
+    #[serde(rename = "accountNumber")]
+    account_number: Secret<String>,
+    #[serde(rename = "type")]
+    account_type: common_enums::BankType,
+    name: Secret<String>,
```
deepanshu-iiitu: Use camel case here instead of renaming

crates/hyperswitch_connectors/src/connectors/dwolla/transformers.rs
```diff
@@ -1,228 +1,611 @@
...
-            //TODO: Review mapping
-        }
-    }
+#[derive(Debug, Serialize, PartialEq, Deserialize, Clone)]
+pub struct DwollaPaymentsRequest {
+    #[serde(rename = "_links")]
+    links: DwollaPaymentLinks,
+    amount: DwollaAmount,
+    #[serde(rename = "correlationId")]
+    correlation_id: String,
```
deepanshu-iiitu: Use camel case for the full struct except _links

crates/hyperswitch_connectors/src/connectors/dwolla/transformers.rs
```diff
@@ -1,228 +1,611 @@
...
 
-impl TryFrom<RefundsResponseRouterData<Execute, RefundResponse>> for RefundsRouterData<Execute> {
+#[derive(Debug, Serialize, PartialEq, Deserialize, Clone)]
+pub struct DwollaRefundsRequest {
+    #[serde(rename = "_links")]
+    links: DwollaPaymentLinks,
+    amount: DwollaAmount,
+    #[serde(rename = "correlationId")]
+    correlation_id: String,
+}
```
deepanshu-iiitu: Use camel case for the full struct except _links

crates/hyperswitch_connectors/src/connectors/dwolla/transformers.rs
```diff
@@ -1,228 +1,611 @@
...
+                source: DwollaRequestLink { href: source_url },
+                destination: DwollaRequestLink {
+                    href: destination_url,
+                },
+            },
+            amount: DwollaAmount {
+                currency: item.router_data.request.currency,
+                value: item.amount.to_owned(),
+            },
+            correlation_id: format!("payment_{}", uuid::Uuid::new_v4()),
```
deepanshu-iiitu: dont generate uuid
instead use item.router_data.connector_request_reference_id

crates/hyperswitch_connectors/src/connectors/dwolla/transformers.rs
```diff
@@ -1,228 +1,611 @@
...
+                source: DwollaRequestLink { href: source_url },
+                destination: DwollaRequestLink {
+                    href: destination_url,
+                },
+            },
+            amount: DwollaAmount {
+                currency: item.router_data.request.currency,
+                value: item.amount.to_owned(),
+            },
+            correlation_id: format!("refund_{}", uuid::Uuid::new_v4()),
```
deepanshu-iiitu: instead of generating uuid use the hyperswitch refund id

crates/hyperswitch_connectors/src/connectors/dwolla/transformers.rs
```diff
@@ -1,228 +1,611 @@
...
+pub struct DwollaErrorDetail {
+    pub code: Option<String>,
+    pub message: Option<String>,
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
+#[serde(rename_all = "camelCase")]
+pub struct DwollaWebhookDetails {
+    pub id: String,
+    #[serde(rename = "resourceId")]
```
deepanshu-iiitu: This is not needed if you use camelCase

crates/hyperswitch_connectors/src/connectors/dwolla/transformers.rs
```diff
@@ -1,228 +1,611 @@
...
+    pub resource_id: String,
+    pub topic: String,
+    #[serde(rename = "correlationId")]
```
deepanshu-iiitu: This is not needed if you use camelCase

crates/hyperswitch_connectors/src/connectors/dwolla/transformers.rs
```diff
@@ -1,228 +1,611 @@
...
+    CustomerTransferFailed,
+    TransferCreated,
+    TransferPending,
+    TransferCompleted,
+    TransferFailed,
+    #[serde(other)]
+    Unknown,
+}
+
+impl TryFrom<DwollaWebhookDetails> for api_models::webhooks::IncomingWebhookEvent {
```
deepanshu-iiitu, line 569: Please attach screenshots of source_verfied being true in pr description under test cases

crates/hyperswitch_connectors/src/connectors/dwolla/transformers.rs
```diff
@@ -1,228 +1,610 @@
...
+                source: DwollaRequestLink { href: source_url },
+                destination: DwollaRequestLink {
+                    href: destination_url,
+                },
+            },
+            amount: DwollaAmount {
+                currency: item.router_data.request.currency,
+                value: item.amount.to_owned(),
+            },
+            correlation_id: format!("refund_{}", item.router_data.connector_request_reference_id),
```
deepanshu-iiitu, line 411:
```suggestion
            correlation_id: format!("refund_{}", item.router_data.request.refund_id),
```

deepanshu-iiitu (dismissed):

crates/router/src/services/api.rs
```diff
@@ -350,6 +350,30 @@ where
...
+                                                        return connector_integration
+                                                            .handle_response(
+                                                                req,
+                                                                Some(&mut connector_event),
+                                                                response,
+                                                            );
+                                                    }
+                                                }
+                                            }
+
```
sahkal: Can you please explain this part, why are we handling 4xx response as 2xx, also put comments for such unintuitive approaches

Vani-1107: Dwolla creates a funding source ID for a customer with the help of their bank account details, including the account number, routing number, and account holder name.
So whenever the customer tries to make payment for the second time with the same details, it throws a 400 error with the funding source ID in the response.
So, we bypass the 4xx error and treat is as 2xx.

Vani-1107: Added comments.

jagan-jaya: Let us not handle the connector specific logic in the core layer, let us understand the error sent by connector during the second time payment and find a way to gracefully handle it. check if there is any field to mention the payment as returning user payment so that connector does not throw this error.

jagan-jaya: Approving it as we don't have any API level feature flag or merchant level flag to bypass this error in the connector level. Let's reach out to their customer support regarding the same.

jagan-jaya (dismissed):

crates/router/src/services/api.rs
```diff
@@ -350,6 +350,32 @@ where
...
+                                                        };
+                                                        return connector_integration
+                                                            .handle_response(
+                                                                req,
+                                                                Some(&mut connector_event),
+                                                                response,
+                                                            );
+                                                    }
+                                                }
+                                            }
```
ShankarSinghC: This can be handled in the connector layer using the `get_error_response` method by returning the appropriate http status code, attempt status in the `ErrorResponse`.

Vani-1107: In our case, we want to extract the funding source ID from the error response that Dwolla throws (400) and use it further in the flow. So instead of treating it as an actual error, we bypass it and treat it as a success case to retrieve and utilize the funding source ID.

crates/router/src/services/api.rs
```diff
@@ -350,6 +350,32 @@ where
                                             error_res
                                         }
                                         _ => {
+                                            // Dwolla creates a funding source ID using account details and if the same details are used again, it returns a 400 error with the funding source ID in the response.
+                                            // We handle this 400 as a success case and extract the funding source ID.
```
jarnura: Connector specific changes and all should not added in framework or common library function, this can be handled in handle_response in connector end itself, we are getting raw bytes and status code in that function, any kind of handling can happen there, we shouldn't do this here

crates/router/src/core/payments/tokenization.rs
```diff
@@ -1346,8 +1346,8 @@ pub fn update_router_data_with_payment_method_token_result<F: Clone, T>(
                 true
             }
             Err(err) => {
+                router_data.response = Err(err.clone());
                 if is_retry_payment {
-                    router_data.response = Err(err);
```
ShankarSinghC: Currently, when there is an error in the preprocessing step, we do not update the `router_data` with the error response. As a result, the flow proceeds to the connector layer, where it gets short-circuited because the payment method token is missing.

However, in the retry case, if the preprocessing step fails, we populate the error response in router_data, and `should_continue_further` is set to false, which prevents the connector from being called.

Should we populate the error response in all cases (both normal and retry payments) and consistently return `should_continue_further` as false? Would this impact any existing flows?

@deepanshu-iiitu  Is there any connector for which preprocessing step is optional and even then we perform ?
 
 cc: @jarnura @sai-harsha-vardhan @deepanshu-iiitu

deepanshu-iiitu: We should  return should_continue_further as false as preprocessing is not optional for any connector

ShankarSinghC: Ok then lets try to keep the behaviour consistent for all the flow

@Vani-1107 We can return false in case of any error, without needing to check whether the flow is a retry or not

ShankarSinghC: Also please test all necessary flows

Vani-1107: @ShankarSinghC  , 
If we return false and halt the flow when there's an error during tokenization, it will prevent second-time payments from going through in the case of Dwolla, where we rely on the error response to extract the funding source ID and use it during authroization .

jarnura (approved):

PR #8692, opened 2025-07-18
chore: updated open api spec documentation for create platform accoun…

gitanjli525:
…t api

## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [x] Documentation
- [ ] CI/CD

## Description
- generated open api spec for create platform api

### Additional Changes

## Checklist

### DIFF
--- a/crates/openapi/src/openapi.rs
+++ b/crates/openapi/src/openapi.rs
@@ -213,6 +213,9 @@ Never share your secret api keys. Keep them guarded and secure.
 
         // Routes for authentication
         routes::authentication::authentication_create,
+
+        // Routes for user
+        routes::user::create_platform_account,
     ),
     components(schemas(
         common_utils::types::MinorUnit,
--- a/api-reference/v1/openapi_spec_v1.json
+++ b/api-reference/v1/openapi_spec_v1.json
@@ -6390,6 +6390,53 @@
           }
         ]
       }
+    },
+    "/create_platform": {
+      "post": {
+        "tags": [
+          "User"
+        ],
+        "summary": "User - Create Platform Account",
+        "description": "Create a new platform account",
+        "operationId": "Create a Platform Account",
+        "requestBody": {
+          "content": {
+            "application/json": {
+              "schema": {
+                "$ref": "#/components/schemas/PlatformAccountCreateRequest"
+              },
+              "examples": {
+                "Create a platform account with organization_name": {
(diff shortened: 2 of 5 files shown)


crates/openapi/src/openapi.rs
```diff
@@ -362,6 +365,7 @@ Never share your secret api keys. Keep them guarded and secure.
         api_models::enums::MerchantProductType,
         api_models::enums::CtpServiceProvider,
         api_models::enums::PaymentLinkSdkLabelType,
+        api_models::enums::OrganizationType,
```
tsdk02, line 368: This has already been added, can be removed

crates/api_models/src/user.rs
```diff
@@ -150,17 +151,24 @@ pub struct UserOrgMerchantCreateRequest {
...
+    #[schema(max_length = 64, value_type = String, example = "organization_abc")]
     pub organization_name: Secret<String>,
 }
 
-#[derive(Debug, serde::Deserialize, serde::Serialize)]
+#[derive(Debug, serde::Deserialize, serde::Serialize, ToSchema)]
 pub struct PlatformAccountCreateResponse {
+    #[schema(value_type = String, max_length = 64, min_length = 1, example = "org_abc")]
     pub org_id: id_type::OrganizationId,
+    #[schema(value_type = String, example = "organization_abc")]
```
AnuthaDev:
```suggestion
    #[schema(value_type = Option<String>, example = "organization_abc")]
```

AnuthaDev (changes requested):

crates/api_models/src/user.rs
gitanjli525: @AnuthaDev We do not need to mention here Option<String> because the value_type is actually string, option just indicates it might or might not be present. If present the type will be string and that is how it will be implemented as string in open api spec json file

AnuthaDev: Since org_name is a nullable field. We need to mention that in our API ref.

Adding `Option<String>` here ensures the API docs mention that org_name can be null. Like this:

image

Otherwise the API docs will mention that it is a non-nullable field, misleading end users. Like this

image

tsdk02 (dismissed):

jagan-jaya (approved):

PR #8741, opened 2025-07-24
fix(connector): [Worldpay] handle multiple ddc submission for CompleteAuthorize

kashif-m:
## Type of Change
- [x] Bugfix

## Description

Fixes race condition in Worldpay 3DS Device Data Collection (DDC) flow causing `bodyDoesNotMatchSchema` errors.

**Changes made:**
- Extended DDC timeout from 8s to 10s in client-side JavaScript
- Added `ddcProcessed` flag to prevent duplicate submissions on client-side
- Reject `collectionReference` submissions when payment is in `AuthenticationPending` state
- Fixed CSS media query syntax (`@media (max-width: 600px)`)
- Added Cypress test case for both client and server-side race conditions for DDC submissions

### Additional Changes

## Motivation and Context

Race condition caused intermittent 3DS payment failures when DDC took > 8 seconds.

## How did you test it?

<details>
    <summary>1. DDC race case in isolation </summary>

Screenshot 2025-07-24 at 12 02 50 PM

</details>

<details>
    <summary>2. All test cases for Worldpay</summary>

Screenshot 2025-07-25 at 12 52 56 AM

Note
- PSync for WP fails due to 404 and WP sending back a HTML rather than a JSON response (https://github.com/juspay/hyperswitch/pull/8753)

</details>

## Checklist

### DIFF
--- a/crates/hyperswitch_connectors/src/connectors/worldpay.rs
+++ b/crates/hyperswitch_connectors/src/connectors/worldpay.rs
@@ -800,9 +800,13 @@ impl ConnectorIntegration<CompleteAuthorize, CompleteAuthorizeData, PaymentsResp
             enums::AttemptStatus::DeviceDataCollectionPending => "3dsDeviceData".to_string(),
             enums::AttemptStatus::AuthenticationPending => "3dsChallenges".to_string(),
             _ => {
-                return Err(errors::ConnectorError::RequestEncodingFailedWithReason(
-                    format!("Invalid payment status for complete authorize: {:?}", req.status)
-                ).into());
+                return Err(
+                    errors::ConnectorError::RequestEncodingFailedWithReason(format!(
+                        "Invalid payment status for complete authorize: {:?}",
+                        req.status
+                    ))
+                    .into(),
+                );
             }
         };
         Ok(format!(
--- a/crates/hyperswitch_connectors/src/connectors/worldpay.rs
+++ b/crates/hyperswitch_connectors/src/connectors/worldpay.rs
@@ -798,7 +798,12 @@ impl ConnectorIntegration<CompleteAuthorize, CompleteAuthorizeData, PaymentsResp
             .ok_or(errors::ConnectorError::MissingConnectorTransactionID)?;
         let stage = match req.status {
             enums::AttemptStatus::DeviceDataCollectionPending => "3dsDeviceData".to_string(),
-            _ => "3dsChallenges".to_string(),
+            enums::AttemptStatus::AuthenticationPending => "3dsChallenges".to_string(),
+            _ => {
+                return Err(errors::ConnectorError::RequestEncodingFailedWithReason(
+                    format!("Invalid payment status for complete authorize: {:?}", req.status)
+                ).into());
+            }
         };
         Ok(format!(
             "{}api/payments/{connector_payment_id}/{stage}",
--- a/cypress-tests/cypress/e2e/configs/Payment/Utils.js
+++ b/cypress-tests/cypress/e2e/configs/Payment/Utils.js
@@ -396,6 +396,8 @@ export const CONNECTOR_LISTS = {
       // "cybersource"    // issues with MULTIPLE_CONNECTORS handling
       "paypal",
     ],
+    DDC_RACE_CONDITION_SERVER_SIDE: ["worldpay"],
+    DDC_RACE_CONDITION_CLIENT_SIDE: ["worldpay"],
     // Add more inclusion lists
   },
 };
--- a/cypress-tests/cypress/e2e/configs/Payment/Utils.js
+++ b/cypress-tests/cypress/e2e/configs/Payment/Utils.js
@@ -396,8 +396,7 @@ export const CONNECTOR_LISTS = {
       // "cybersource"    // issues with MULTIPLE_CONNECTORS handling
       "paypal",
     ],
-    DDC_RACE_CONDITION_SERVER_SIDE: ["worldpay"],
-    DDC_RACE_CONDITION_CLIENT_SIDE: ["worldpay"],
+    DDC_RACE_CONDITION: ["worldpay"],
     // Add more inclusion lists
   },
 };
--- a/cypress-tests/cypress/e2e/configs/Payment/WorldPay.js
+++ b/cypress-tests/cypress/e2e/configs/Payment/WorldPay.js
@@ -45,13 +45,21 @@ const successfulThreeDsTestCardDetailsRequest = {
   card_cvc: "737",
 };
 
+const failedNoThreeDsCardDetails = {
+  card_number: "4242424242424242",
+  card_exp_month: "10",
+  card_exp_year: "30",
+  card_holder_name: "REFUSED13",
+  card_cvc: "737",
+};
+
 const paymentMethodDataNoThreeDsResponse = {
   card: {
     last4: "4242",
-    card_type: "CREDIT",
-    card_network: "Visa",
-    card_issuer: "STRIPE PAYMENTS UK LIMITED",
-    card_issuing_country: "UNITEDKINGDOM",
+    card_type: null,
+    card_network: null,
+    card_issuer: null,
+    card_issuing_country: null,
     card_isin: "424242",
     card_extended_bin: null,
     card_exp_month: "10",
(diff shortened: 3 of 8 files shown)


crates/hyperswitch_connectors/src/connectors/worldpay.rs
```diff
@@ -798,7 +798,16 @@ impl ConnectorIntegration<CompleteAuthorize, CompleteAuthorizeData, PaymentsResp
             .ok_or(errors::ConnectorError::MissingConnectorTransactionID)?;
         let stage = match req.status {
             enums::AttemptStatus::DeviceDataCollectionPending => "3dsDeviceData".to_string(),
-            _ => "3dsChallenges".to_string(),
+            enums::AttemptStatus::AuthenticationPending => "3dsChallenges".to_string(),
+            _ => {
+                return Err(
+                    errors::ConnectorError::RequestEncodingFailedWithReason(format!(
+                        "Invalid payment status for complete authorize: {:?}",
```
AkshayaFoiger, line 805: This returns a 5xx error, would a 4xx error be more appropriate?

sai-harsha-vardhan (dismissed):

crates/hyperswitch_connectors/src/connectors/worldpay.rs
kashif-m: Yes, this flow should never be triggered from core layer apart from other two states, so added as 5xx

AkshayaFoiger (approved):

jagan-jaya (dismissed):

cypress-tests/cypress/e2e/configs/Payment/WorldPay.js
```diff
@@ -710,6 +740,116 @@ export const connectorDetails = {
         },
       },
     },
+    DDCRaceConditionServerSide: {
```
pixincreate, line 743: can you add the same in commons too? asking because, you have added a new spec file, this will affect all connectors. other connector configs does not have this config. if you add it in commons (only the request), 501 will be expected by default for other connectors.

if you think race condition test is specific to a connector and will be applicable to no other connector, then you add it to inclusion list and make relevant changes to the spec file to read from it. that way, the test will run only against that specific connector.

cypress-tests/cypress/e2e/configs/Payment/WorldPay.js
```diff
@@ -710,6 +740,116 @@ export const connectorDetails = {
...
+              actualSubmissions: 0,
+              blockedSubmissions: 0
+            };
+            
+            // Override the submitCollectionReference function to test race conditions
+            var originalSubmit = window.submitCollectionReference;
+            
+            window.submitCollectionReference = function(collectionReference) {
+              window.testResults.submissionAttempts++;
+              console.log("SUBMISSION_ATTEMPT_" + window.testResults.submissionAttempts + ": " + collectionReference);
```
pixincreate, line 821: are the console logs expected?

quite surprising to see this not getting flagged by the linter...🤨

cypress-tests/cypress/e2e/spec/Payment/00030-DDCRaceCondition.cy.js
```diff
@@ -0,0 +1,264 @@
...
+            "";
+          cy.log(
+            `Single submission detected with ${ddcConfig.collectionReferenceParam}: "${collectionRef}"`
+          );
+        });
+
+        cy.log(
+          "✅ Client-side race condition protection verified - only one submission occurred"
+        );
+      });
```
pixincreate: i would suggest moving this to `support/commands.js` and call it from here.

pixincreate (commented):
other than that, lgtm.

cypress-tests/cypress/e2e/configs/Payment/WorldPay.js
```diff
@@ -45,6 +45,14 @@ const successfulThreeDsTestCardDetailsRequest = {
   card_cvc: "737",
 };
 
+const failedNoThreeDsCardDetails = {
```
pixincreate, line 48: also, can you please post a screenshot of tests running against worldpay in the pr description?

cypress-tests/cypress/e2e/configs/Payment/WorldPay.js
kashif-m: This script is injected in the incoming DDC response from the application, this not a part of cypress code directly

cypress-tests/cypress/e2e/configs/Payment/WorldPay.js
kashif-m: Done, it follows the same standards now

cypress-tests/cypress/e2e/configs/Payment/WorldPay.js
kashif-m: These were already present, updated again under "All test cases for Worldpay"

cypress-tests/cypress/e2e/configs/Payment/Utils.js
```diff
@@ -396,6 +396,8 @@ export const CONNECTOR_LISTS = {
       // "cybersource"    // issues with MULTIPLE_CONNECTORS handling
       "paypal",
     ],
+    DDC_RACE_CONDITION_SERVER_SIDE: ["worldpay"],
+    DDC_RACE_CONDITION_CLIENT_SIDE: ["worldpay"],
```
pixincreate: this can be a single config? `DDC_RACE_CONDITION`?

cypress-tests/cypress/e2e/spec/Payment/00030-DDCRaceCondition.cy.js
```diff
@@ -0,0 +1,159 @@
...
+
+        if (
+          shouldIncludeConnector(
+            connector,
+            CONNECTOR_LISTS.INCLUDE.DDC_RACE_CONDITION_SERVER_SIDE
+          ) &&
+          shouldIncludeConnector(
+            connector,
+            CONNECTOR_LISTS.INCLUDE.DDC_RACE_CONDITION_CLIENT_SIDE
+          )
```
pixincreate: with the above change, this would become 
```suggestion
        if (
          shouldIncludeConnector(
            connector,
            CONNECTOR_LISTS.INCLUDE.DDC_RACE_CONDITION
          )
```

cypress-tests/cypress/support/commands.js
```diff
@@ -4066,3 +4066,122 @@ Cypress.Commands.add("setConfigs", (globalState, key, value, requestType) => {
...
+      throw new Error(
+        "Missing merchantId - this indicates a critical state issue"
+      );
+    }
+
+    if (!paymentId) {
+      throw new Error(
+        "Failed to create payment intent - paymentID not found in globalState"
+      );
+    }
```
pixincreate: are these needed?

cypress-tests/cypress/support/commands.js
```diff
@@ -4066,3 +4066,122 @@ Cypress.Commands.add("setConfigs", (globalState, key, value, requestType) => {
...
```
pixincreate: same here

cypress-tests/cypress/support/commands.js
kashif-m: No these are not needed, as long as we pass the right confirmData in arguments

Removed these

cypress-tests/cypress/e2e/configs/Payment/WorldPay.js
```diff
@@ -710,6 +740,116 @@ export const connectorDetails = {
...
+          browser_info,
+        },
+        Response: {
+          status: 200,
+          body: {
+            status: "requires_customer_action",
+          },
+        },
+      }),
+      DDCConfig: {
```
pixincreate, line 801: nit: these can be imported from commons?

kashif-m: As per the discussions - the script for DDC is usually very different for different connectors.

Keeping them in `Worldpay` instead to ensure changing Commons does not affect this connector.

sai-harsha-vardhan (approved):

deepanshu-iiitu (approved):

PR #8742, opened 2025-07-24
feat: add ci support for mock server

pixincreate:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [x] CI/CD

## Description

This PR adds support for running tests against mock server for alpha connectors.

### Additional Changes

## Motivation and Context

The goal is to make sure that the implementation works.
closes https://github.com/juspay/hyperswitch/issues/8798

## How did you test it?

CI should pass / work.

## Checklist


pixincreate:
converting to draft as it still requires api fixes to the mock server and cypress configs.

iemyashasvi:
Fixed the cypress and mock server
Screenshot 2025-07-30 at 2 01 08 AM

.github/workflows/cypress-tests-runner.yml
```diff
@@ -149,30 +214,173 @@ jobs:
...
+        uses: actions/download-artifact@v4
+        with:
+          name: hyperswitch-server
+          path: target/debug
+
+      - name: Make server executable
+        if: ${{ env.RUN_TESTS == 'true' }}
+        run: chmod +x target/debug/router
+
+      - name: Download Encrypted TOML from S3 and Decrypt
```
srujanchikke:
```suggestion
      - name: Download Encrypted JSON from S3 and Decrypt
```

deepanshu-iiitu (approved):

.github/workflows/cypress-tests-runner.yml
```diff
@@ -80,8 +81,61 @@ jobs:
...
+      - name: Build project
+        if: ${{ env.RUN_TESTS == 'true' }}
+        run: cargo build --package router --bin router --jobs 3
+
+      - name: Upload binary artifact
+        if: ${{ env.RUN_TESTS == 'true' }}
+        uses: actions/upload-artifact@v4
+        with:
+          name: hyperswitch-server
+          overwrite: true
```
SanchithHegde, line 132: Why are we specifying `overwrite` here?

.github/workflows/cypress-tests-runner.yml
```diff
@@ -149,30 +214,173 @@ jobs:
...
+          path: |
+            cypress-tests/cypress/reports/
+          retention-days: 1
+
+  runner_alpha:
+    name: Run optional Cypress tests for alpha connectors
+    needs: build-hyperswitch
+    runs-on: ubuntu-latest
+    # This job is not required to pass for PR to be merged
+    continue-on-error: true
```
SanchithHegde: I don't think we need `continue-on-error: true` because we aren't using a matrix here. And this specific job won't be marked as required check for PRs, so knowing that this specific job failed would be fine.

.github/workflows/cypress-tests-runner.yml
pixincreate: we don't want to retain multiple versions of same artifact when we re-run the workflow, so we just overwrite. it is isolated from different workflow runs btw.

ShankarSinghC (approved):

likhinbopanna (approved):

PR #8763, opened 2025-07-26
[BUG] handle refused responses when only one of refusalCode or refusa…

### LINKED ISSUE #8749: [BUG] handle refused responses when only one of `refusalCode` or `refusalDescription` is present for Worldpay
`refusalCode` and `refusalDescription` are mandatory fields whenever Worldpay refuses the transactions. There is a case where only `refusalDescription` is present. Handle these cases accordingly for `refused` scenarios.

Use `NO_ERROR_CODE` and `NO_ERROR_MESSAGE` in connector integration when either one or both of them are missing in the response.

sarika-03:
fix(worldpay): handle refused responses when only refusalCode or refusalDescription is present

### Type of Change
- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

---

### Description
Some refused payment responses from Worldpay contain only `refusalCode` or only `refusalDescription`, but not both.  
Previously, the code assumed both fields would always be present, which led to `None` errors or blank error messages.

#### ✅ Fix Implemented:
- Generated a meaningful error message using whichever field is available.
- Added null-safe checks to avoid parsing failures when either field is missing.

---

### Additional Changes

---

### Motivation and Context
As per issue #8749, Worldpay refusal handling was failing when only one of the fields (`refusalCode` or `refusalDescription`) was present.  
This bug was affecting both error tracing and logging in production.

The fix ensures the response is consistent — regardless of which field is present, the user receives a proper error message.

---

### How did you test it?
Manually tested using local setup:

- Triggered mock refused responses with:
  - Only `refusalCode`
  - Only `refusalDescription`
  - Both fields present
- Verified that:
  - Error messages are correctly built
  - No crashes or missing error info
  - Logs and API response show expected output

---

### Checklist
- [x] I formatted the code using `cargo +nightly fmt --all`

Closes #8749

### DIFF
--- a/crates/hyperswitch_connectors/src/connectors/worldpay/response.rs
+++ b/crates/hyperswitch_connectors/src/connectors/worldpay/response.rs
@@ -652,3 +652,4 @@ pub enum WorldpayWebhookStatus {
 pub const WP_CORRELATION_ID: &str = "WP-CorrelationId";
 
 
+


crates/hyperswitch_connectors/src/connectors/worldpay/response.rs
```diff
@@ -64,16 +64,18 @@ pub struct FraudHighRiskResponse {
 #[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
 #[serde(rename_all = "camelCase")]
 pub struct RefusedResponse {
-    pub refusal_description: String,
-    // Access Worldpay returns a raw response code in the refusalCode field (if enabled) containing the unmodified response code received either directly from the card scheme for Worldpay-acquired transactions, or from third party acquirers.
```
kashif-m, line 68: Let's keep the comment

crates/hyperswitch_connectors/src/connectors/worldpay/response.rs
```diff
@@ -64,16 +64,18 @@ pub struct FraudHighRiskResponse {
...
-    pub refusal_code: String,
+    #[serde(default)]
+    pub refusal_description: Option<String>,
+    #[serde(default)]
+    pub refusal_code: Option<String>,
```
kashif-m: The problem with doing it this way is that the response will always most likely be deserialized into `RefusedResponse` if we make these fields optional.


```rust
pub enum WorldpayPaymentResponseFields {
    RefusedResponse(RefusedResponse),
    DDCResponse(DDCResponse),
    ThreeDsChallenged(ThreeDsChallengedResponse),
    FraudHighRisk(FraudHighRiskResponse),
    AuthorizedResponse(Box<AuthorizedResponse>),
}
```

A better way of doing this would be to deserialize the status first, and based on the status - deserialize rest of the fields. This can be done using a custom deserializer ([ref](https://serde.rs/deserialize-struct.html))

Custom deserializer can be written for

```rust
pub struct WorldpayPaymentsResponse {
    pub outcome: PaymentOutcome,
    pub transaction_reference: Option<String>,
    #[serde(flatten)]
    pub other_fields: Option<WorldpayPaymentResponseFields>,
}
```

sarika-03:
I've added a custom Deserialize implementation for WorldpayPaymentsResponse, which conditionally deserializes based on the outcome field. If outcome == "refused", it maps to RefusedResponse; otherwise, it handles the other variants accordingly. Let me know if you'd like me to refactor or tighten any part further.

sarika-03:
fix: Introduce custom deserializer for WorldpayPaymentsResponse

This change implements a custom `Deserialize` for `WorldpayPaymentsResponse`
to address the issue of ambiguous `WorldpayPaymentResponseFields` deserialization.
The previous untagged enum approach could lead to incorrect parsing, especially
for `RefusedResponse`, when fields were made optional.

Now, the deserialization process first reads the `outcome` field and then
branches to deserialize the correct variant of `WorldpayPaymentResponseFields`,
ensuring accurate mapping of payment outcomes.

Also fixed `pub(super)` to `pub` for `WP_CORRELATION_ID` to resolve compilation error.

PR #8788, opened 2025-07-29
feat(core): populate UCS status_code in response headers

Aishwariyaa-Anand:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

This PR populates UCS status_code (connector HTTP code) in response headers.
It also introduces proxy_connector_http_status_code which when enabled, configures Hyperswitch to mirror (proxy) the connector’s HTTP status code in its own response.

### Additional Changes

## How did you test it?

Enable UCS Config
```
--data '{
    "key": "ucs_rollout_config_{{merchant_id}}_razorpay_upi_Authorize",
    "value": "1.0"
}'
```
Make a payment
```
curl --location 'http://localhost:8080/v2/payments' \
--header 'api-key: dev_mwKUL6tWhsO8FbwWtKjNllSa7GnZutXqKsv9oLxJRv4c7l8UX9muRdsC4r9Wh2A5' \
--header 'x-profile-id: pro_cQ4q1A6oYpi89Zi7zyxW' \
--header 'Authorization: api-key=dev_mwKUL6tWhsO8FbwWtKjNllSa7GnZutXqKsv9oLxJRv4c7l8UX9muRdsC4r9Wh2A5' \
--data-raw '{
    "amount_details": {
        "order_amount": 100,
        "currency": "INR"
    },
    "capture_method":"automatic",
    "description": "Its my first payment request",
    "authentication_type": "no_three_ds",
    "return_url": "https://google.com",
    "payment_method_data": {
        "upi": {
            "upi_collect": {
                "vpa_id": "success"
            }
        },
        "billing": {
            "address": {
                "line1": "1467",
                "line2": "Harrison Street",
                "line3": "Harrison Street",
                "city": "San Fransico",
                "state": "California",
                "zip": "94122",
                "country": "IN",
                "first_name": "Swangi",
                "last_name": "Kumari"
            },
            "phone": {
                "number": "8056594427",
                "country_code": "+91"
            },
            "email": "swangi.kumari@juspay.in"
        }
    },
    "payment_method_subtype": "upi_collect",
    "payment_method_type": "upi",
    "return_raw_connector_response": true
}'
```
Response 
```
{
    "id": "12345_pay_019860443e3d7f939e80e9a98effba3d",
    "status": "requires_merchant_action",
    "amount": {
        "order_amount": 100,
        "currency": "INR",
        "shipping_cost": null,
        "order_tax_amount": null,
        "external_tax_calculation": "skip",
        "surcharge_calculation": "skip",
        "surcharge_amount": null,
        "tax_on_surcharge": null,
        "net_amount": 100,
        "amount_to_capture": null,
        "amount_capturable": 0,
        "amount_captured": null
    },
    "customer_id": null,
    "connector": "razorpay",
    "created": "2025-07-31T11:35:41.885Z",
    "payment_method_data": {
        "billing": {
            "address": {
                "city": "San Fransico",
                "country": "IN",
                "line1": "1467",
                "line2": "Harrison Street",
                "line3": "Harrison Street",
                "zip": "94122",
                "state": "California",
                "first_name": "Swangi",
                "last_name": "Kumari"
            },
            "phone": {
                "number": "8056594427",
                "country_code": "+91"
            },
            "email": "swangi.kumari@juspay.in"
        }
    },
    "payment_method_type": "upi",
    "payment_method_subtype": "upi_collect",
    "connector_transaction_id": "order_QzezHblnRQM45h",
    "connector_reference_id": null,
    "merchant_connector_id": "mca_BIFmH1dr6Vh5UoYu2srb",
    "browser_info": null,
    "error": {
        "code": "BAD_REQUEST_ERROR",
        "message": "Invalid VPA. Please enter a valid Virtual Payment Address",
        "unified_code": null,
        "unified_message": null,
        "network_advice_code": null,
        "network_decline_code": null,
        "network_error_message": null
    },
    "shipping": null,
    "billing": null,
    "attempts": null,
    "connector_token_details": null,
    "payment_method_id": null,
    "next_action": null,
    "return_url": "https://google.com/",
    "authentication_type": "no_three_ds",
    "authentication_type_applied": "no_three_ds",
    "merchant_reference_id": null,
    "raw_connector_response": "{\"error\":{\"code\":\"BAD_REQUEST_ERROR\",\"description\":\"Invalid VPA. Please enter a valid Virtual Payment Address\",\"source\":\"customer\",\"step\":\"payment_initiation\",\"reason\":\"invalid_vpa\",\"metadata\":{},\"field\":\"vpa\"}}",
    "feature_metadata": null
}
```
Verify response headers for connector_http_status_code
image

If proxy_connector_http_status_code is enabled
Hyperswitch resoponse status code will also be 400
image

## Checklist


sai-harsha-vardhan (commented):
We do already send `connector_http_status_code` in response headers, why do we need it in response payload as well? @Aishwariyaa-Anand

crates/api_models/src/payments.rs
```diff
@@ -5702,6 +5702,8 @@ pub struct ErrorDetails {
     pub network_decline_code: Option<String>,
     /// A string indicating how to proceed with an network error if payment gateway provide one. This is used to understand the network error code better.
     pub network_error_message: Option<String>,
+    /// HTTP status code returned by the connector
+    pub status_code: Option<i32>,
```
hrithikesh026: Are we not returning http status code in case of success ?

v2_migrations/2025-07-28-100937_add_status_code_in_payment_attempt/up.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE payment_attempt
+ADD COLUMN IF NOT EXISTS status_code INTEGER;
```
hrithikesh026: Please move this to v2_compatible_migrations

crates/router/src/core/unified_connector_service/transformers.rs
```diff
@@ -426,12 +432,18 @@ impl ForeignTryFrom<payments_grpc::PaymentServiceGetResponse>
                     })
             });
 
+        let status_code = u16::try_from(response.status_code.unwrap_or(500)).map_err(|_| {
```
hrithikesh026: can't we change payments_grpc::PaymentServiceGetResponse::status_code to u16?

Aishwariyaa-Anand: u16 is not supported in proto

crates/router/src/core/unified_connector_service/transformers.rs
```diff
@@ -964,3 +972,18 @@ impl ForeignTryFrom<common_types::payments::CustomerAcceptance>
...
     }
 }
+
+impl ForeignTryFrom<Option<u32>> for u16 {
+    type Error = error_stack::Report<UnifiedConnectorServiceError>;
+
+    fn foreign_try_from(status_code: Option<u32>) -> Result<Self, Self::Error> {
+        let code = status_code.unwrap_or(500);
+
+        Self::try_from(code).map_err(|_| {
```
sai-harsha-vardhan: Can we try logging the error here?

crates/router/src/core/payments/transformers.rs
```diff
@@ -1585,9 +1585,17 @@ where
             status: payment_intent.status,
         };
 
+        let headers = connector_http_status_code
+            .map(|status_code| {
+                vec![(
+                    "connector_http_status_code".to_string(),
```
SanchithHegde: 1. If this is meant to be a header name, consider using hyphens instead of underscores.
2. Typically, we include custom (non-standard) headers with an `x-` prefix. So, this header name would be `x-connector-http-status-code`.
3. Consider storing the header name in a constant. I believe we have a `headers` module with similar constants.

crates/router/src/core/unified_connector_service/transformers.rs
```diff
@@ -964,3 +972,19 @@ impl ForeignTryFrom<common_types::payments::CustomerAcceptance>
...
+
+        Self::try_from(code).map_err(|err| {
+            UnifiedConnectorServiceError::RequestEncodingFailedWithReason(format!(
+                "Failed to convert status code to u16: {}",
+                err
+            ))
+            .into()
+        })
+    }
+}
```
SanchithHegde: This implementation looks like a very generic `Option<u32>` -> `u16` conversion implementation at first glance. This does not convey that we are trying to transform a status code stored in a `u32` (I'm guessing this is coming from gRPC) to `u16`. Let's avoid writing such generic-looking implementations which only solve a specific use case of ours (because this trait implementation would be available throughout the `router` crate, even where it's not relevant).

Write this as a function instead. It could be called `convert_connector_service_status_code()` for example.

crates/router/src/core/payments.rs
```diff
@@ -265,6 +266,10 @@ where
 
             let payments_response_operation = Box::new(PaymentResponse);
 
+            connector_http_status_code = router_data.connector_http_status_code;
+            //add connector http status code metrics
```
SanchithHegde: We can remove this redundant comment, here and elsewhere.

crates/router/src/core/unified_connector_service.rs
```diff
@@ -233,55 +233,79 @@ pub fn build_unified_connector_service_auth_metadata(
...
     UnifiedConnectorServiceError,
 > {
     let status = AttemptStatus::foreign_try_from(response.status())?;
 
     let router_data_response =
-        Result::<PaymentsResponseData, ErrorResponse>::foreign_try_from(response)?;
+        Result::<PaymentsResponseData, ErrorResponse>::foreign_try_from(response.clone())?;
 
-    Ok((status, router_data_response))
+    let status_code = u16::foreign_try_from(response.status_code)?;
```
SanchithHegde: Can we avoid cloning the entire response struct if all we need is the response status code? We can also obtain the status code before assigning `router_data_response`?

Please make the same change elsewhere as well.

crates/router/src/services/api.rs
```diff
@@ -898,8 +898,25 @@ where
...
             });
+            let proxy_connector_http_status_code = if state
+                .conf
+                .proxy_status_mapping
+                .enable_proxy_connector_http_status_code
+            {
+                headers
+                    .iter()
+                    .find(|(key, _)| key == "connector_http_status_code")
+                    .and_then(|(_, value)| value.clone().into_inner().parse::<u16>().ok())
```
SanchithHegde: We are ignoring the parsing error here (when the value is not a valid `u16`). Should we consider logging it here instead?

crates/router/src/services/api.rs
```diff
@@ -991,8 +1008,13 @@ pub fn http_response_json_with_headers<T: body::MessageBody + 'static>(
     response: T,
     headers: Vec<(String, Maskable<String>)>,
     request_duration: Option<Duration>,
+    proxy_connector_http_status_code: Option<u16>,
 ) -> HttpResponse {
-    let mut response_builder = HttpResponse::Ok();
+    let mut response_builder = HttpResponse::build(
+        proxy_connector_http_status_code
+            .and_then(|status_code| http::StatusCode::from_u16(status_code).ok())
```
SanchithHegde: We are ignoring the error here as well, in case the `u16` value is not a valid status code. 

Now that I think about it, I feel we can have this function accept an `Option<http::StatusCode>` instead of `Option<u16>`, so that we are sure that the parameter will always be a valid status code, at this point of returning a response.

crates/router/src/services/api.rs
```diff
@@ -991,8 +1008,13 @@ pub fn http_response_json_with_headers<T: body::MessageBody + 'static>(
     response: T,
     headers: Vec<(String, Maskable<String>)>,
     request_duration: Option<Duration>,
+    proxy_connector_http_status_code: Option<u16>,
```
SanchithHegde: Then, we can call this parameter as just `status_code`, which we'll include in the response if one is provided, otherwise we use `200 OK`.

crates/router/src/configs/settings.rs
```diff
@@ -844,6 +845,12 @@ pub struct MerchantIdAuthSettings {
     pub merchant_id_auth_enabled: bool,
 }
 
+#[derive(Debug, Clone, Default, Deserialize)]
+#[serde(default)]
+pub struct ProxyStatusMapping {
+    pub enable_proxy_connector_http_status_code: bool,
```
SanchithHegde: Nit (naming): I don't think we need the `enable` in the name, just `proxy_connector_http_status_code` should suffice. `proxy_connector_http_status_code` being a `bool` indicates that this is a toggle, and I feel that conveys whether the connector status code should be forwarded or not.

By the way, is this applicable to any connector integration, or is it something very specific to the connector service (UCS) integration?

crates/router/src/core/payments/transformers.rs
Aishwariyaa-Anand: `connector_http_status_code` is already exposed in v1 response headers, so renaming it would break backward compatibility. We can introduce the new name in v2 only. However, doing so would require handling the difference where `proxy_connector_http_status_code` is set, since the header name will differ between v1 and v2.

crates/router/src/configs/settings.rs
Aishwariyaa-Anand: This is not specific to UCS integration.
This is something like if a service wants it enabled, Hyperswitch would proxy the connector’s HTTP status code in its own response.

crates/router/src/core/unified_connector_service/transformers.rs
```diff
@@ -964,3 +972,16 @@ impl ForeignTryFrom<common_types::payments::CustomerAcceptance>
         })
     }
 }
+
+pub fn convert_connector_service_status_code(
+    status_code: Option<u32>,
+) -> Result<u16, error_stack::Report<UnifiedConnectorServiceError>> {
+    let code = status_code.unwrap_or(500);
```
SanchithHegde: Should we include a log message in case we don't have the status code from connector service (it is `None`)?

crates/router/src/lib.rs
```diff
@@ -92,6 +92,10 @@ pub mod headers {
     pub const X_CLIENT_SECRET: &str = "X-Client-Secret";
     pub const X_CUSTOMER_ID: &str = "X-Customer-Id";
     pub const X_CONNECTED_MERCHANT_ID: &str = "x-connected-merchant-id";
+    #[cfg(feature = "v1")]
+    pub const X_CONNECTOR_HTTP_STATUS_CODE: &str = "connector_http_status_code";
```
SanchithHegde, line 98: We should either name the constant same as the header (to indicate what the constant actually holds), or we should consider adding a comment indicating why the header name is different from the constant name.


Moreover, where is this constant being read in v1 code?

crates/router/src/services/api.rs
```diff
@@ -898,8 +898,45 @@ where
...
+                                        err
+                                    );
+                                    None
+                                }
+                            },
+                            Err(err) => {
+                                logger::error!(
+                                    "Failed to parse connector_http_status_code from header: {}",
+                                    err
+                                );
```
SanchithHegde: Can we consider using the `Debug` impl when logging errors?

crates/router/src/lib.rs
Aishwariyaa-Anand: This constant in v1 is used during `generate_response` for payments

crates/router/src/core/unified_connector_service/transformers.rs
Aishwariyaa-Anand: UCS not sends status_code as mandatory

crates/router/src/core/unified_connector_service.rs
```diff
@@ -241,11 +241,17 @@ pub fn build_unified_connector_service_auth_metadata(
...
+        Result<PaymentsResponseData, ErrorResponse>,
+        u16,
+    ),
     UnifiedConnectorServiceError,
 > {
     let status = AttemptStatus::foreign_try_from(response.status())?;
 
+    let status_code = transformers::convert_connector_service_status_code(response.status_code)?;
+
     // <<<<<<< HEAD
```
SanchithHegde: What is this commented out merge conflict code 🤦, can we please remove it?

crates/router/src/services/api.rs
```diff
@@ -898,8 +898,45 @@ where
...
+                            },
+                            Err(err) => {
+                                logger::error!(
+                                    "Failed to parse connector_http_status_code from header: {:?}",
+                                    err
+                                );
+                                None
+                            }
+                        }
+                    })
```
hrithikesh026, line 929: nit:
```suggestion
                    .map(|(_, value)| value.clone().into_inner().parse::<u16>())
                    .transpose()
                    .inspect_err(|error| {
                        logger::error!(
                            "Failed to parse connector_http_status_code from header: {:?}",
                            err
                        );
                    })?.map(|code| http::StatusCode::from_u16(code))
                    .transpose()
                    .inspect_err(|error| {
                        logger::error!(
                            "Invalid HTTP status code parsed from connector_http_status_code: {:?}",
                            err
                        );
                    )?
```

hrithikesh026 (approved):

crates/router/src/services/api.rs
Aishwariyaa-Anand: We cant use ? operator as this function doesnot return a error

PR #8831, opened 2025-08-04
feat: add hyperswitch ai chats table

JeevaRamu0104:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

### Additional Changes

## Checklist


crates/api_models/src/chat.rs
```diff
@@ -16,3 +17,31 @@ pub struct ChatResponse {
...
+    pub merchant_id: Option<String>,
+    pub profile_id: Option<String>,
+    pub org_id: Option<String>,
+    pub role_id: Option<String>,
+    pub user_query: Secret<String>,
+    pub response: Secret<serde_json::Value>,
+    pub database_query: Option<String>,
+    pub interaction_status: Option<String>,
+    pub created_at: PrimitiveDateTime,
+}
```
su-shivanshmathur, line 43: Can we have it under a feature flag 
Cc: @jarnura

crates/diesel_models/src/hyperswitch_ai_interaction.rs
```diff
@@ -0,0 +1,49 @@
...
+    Identifiable,
+    Queryable,
+    Selectable,
+    Serialize,
+    router_derive::DebugAsDisplay,
+)]
+#[diesel(table_name = hyperswitch_ai_interaction, primary_key(id), check_for_backend(diesel::pg::Pg))]
+pub struct HyperswitchAiInteraction {
+    pub id: String,
+    pub session_id: Option<String>,
```
su-shivanshmathur, line 21: Difference between `id` and `session_id`

su-shivanshmathur (changes requested):

crates/diesel_models/src/hyperswitch_ai_interaction.rs
JeevaRamu0104: `id` is a request-id and `session_id` is something which generated for that particular chat window

apoorvdixit88: id is the message_id* for particular user request query and response, and the session_id is related to the whole conversation, i.e. current chat window

crates/api_models/src/chat.rs
apoorvdixit88: currently all the routes related to the chat are inside chat config, it is a runtime feature flag `chat_enabled` that says whether chat feature is enabled or not in the given env.

crates/diesel_models/src/hyperswitch_ai_interaction.rs
```diff
@@ -0,0 +1,49 @@
...
+    pub user_query: Option<Encryption>,
+    pub response: Option<Encryption>,
+    pub database_query: Option<String>,
+    pub interaction_status: Option<String>,
+    pub created_at: PrimitiveDateTime,
+}
+
+#[derive(router_derive::Setter, Clone, Debug, Insertable, router_derive::DebugAsDisplay)]
+#[diesel(table_name = hyperswitch_ai_interaction)]
+pub struct HyperswitchAiInteractionNew {
```
inventvenkat, line 36: Aren't we going to have this in a different database?

inventvenkat:
What is the max no. of reads and writes this feature uses, if we enable it?

crates/diesel_models/src/hyperswitch_ai_interaction.rs
JeevaRamu0104: No its going to be in the same database

JeevaRamu0104:
> What is the max no. of reads and writes this feature uses, if we enable it?

no of writes going to be more when compared to read.As we want to store every user query.And writes are specific to OLAP

inventvenkat:
> > What is the max no. of reads and writes this feature uses, if we enable it?
> 
> no of writes going to be more when compared to read.As we want to store every user query.And writes are specific to OLAP

Lets have some limits, such that these queries doesn't affect the main database.

JeevaRamu0104:
> > > What is the max no. of reads and writes this feature uses, if we enable it?
> > 
> > 
> > no of writes going to be more when compared to read.As we want to store every user query.And writes are specific to OLAP
> 
> Lets have some limits, such that these queries doesn't affect the main database.

we will have rate limiting for this api

su-shivanshmathur (approved):

crates/api_models/src/chat.rs
```diff
@@ -16,3 +17,31 @@ pub struct ChatResponse {
...
+    pub user_id: Option<String>,
+    pub merchant_id: Option<String>,
+    pub profile_id: Option<String>,
+    pub org_id: Option<String>,
+    pub role_id: Option<String>,
+    pub user_query: Secret<String>,
+    pub response: Secret<serde_json::Value>,
+    pub database_query: Option<String>,
+    pub interaction_status: Option<String>,
+    pub created_at: PrimitiveDateTime,
```
SanchithHegde, line 42: Consider adding a `#[serde(with = "common_utils::custom_serde::iso8601")]` annotation to ensure consistent serialization and deserialization of timestamp fields.

crates/diesel_models/src/hyperswitch_ai_interaction.rs
```diff
@@ -0,0 +1,49 @@
...
+    Queryable,
+    Selectable,
+    Serialize,
+    router_derive::DebugAsDisplay,
+)]
+#[diesel(table_name = hyperswitch_ai_interaction, primary_key(id), check_for_backend(diesel::pg::Pg))]
+pub struct HyperswitchAiInteraction {
+    pub id: String,
+    pub session_id: Option<String>,
+    pub user_id: Option<String>,
```
SanchithHegde, line 22: Are there any cases where `user_id` would be `NULL` / `None`, or can we make this field required? (Same applies to the `ChatConversation` type as well.)

crates/diesel_models/src/query/hyperswitch_ai_interaction.rs
```diff
@@ -0,0 +1,30 @@
+use diesel::{associations::HasTable, ExpressionMethods};
+
+use crate::{
+    hyperswitch_ai_interaction::*, query::generics, schema::hyperswitch_ai_interaction::dsl,
```
SanchithHegde: Nit: Try to avoid wildcard imports.

crates/diesel_models/src/hyperswitch_ai_interaction.rs
```diff
@@ -0,0 +1,49 @@
...
+    Debug,
+    Deserialize,
+    Identifiable,
+    Queryable,
+    Selectable,
+    Serialize,
+    router_derive::DebugAsDisplay,
+)]
+#[diesel(table_name = hyperswitch_ai_interaction, primary_key(id), check_for_backend(diesel::pg::Pg))]
+pub struct HyperswitchAiInteraction {
```
SanchithHegde: The SQL migration specifies that the primary key is `(id, created_at)`, but here we're only specifying `id` in the `diesel` annotation?

crates/diesel_models/src/hyperswitch_ai_interaction.rs
```diff
@@ -0,0 +1,49 @@
...
+    pub id: String,
+    pub session_id: Option<String>,
+    pub user_id: Option<String>,
+    pub merchant_id: Option<String>,
+    pub profile_id: Option<String>,
+    pub org_id: Option<String>,
+    pub role_id: Option<String>,
```
SanchithHegde, line 26: How is the `role_id` field expected to be used?

crates/router/src/configs/secrets_transformers.rs
```diff
@@ -298,6 +298,25 @@ impl SecretsHandler for settings::UserAuthMethodSettings {
...
+impl SecretsHandler for settings::ChatSettings {
+    async fn convert_to_raw_secret(
+        value: SecretStateContainer<Self, SecuredSecret>,
+        secret_management_client: &dyn SecretManagementInterface,
+    ) -> CustomResult<SecretStateContainer<Self, RawSecret>, SecretsManagementError> {
+        let chat_settings = value.get_inner();
+
+        let encryption_key = secret_management_client
+            .get_secret(chat_settings.encryption_key.clone())
+            .await?;
```
SanchithHegde: We'd have to perform the KMS decryption only if `chat_settings.enabled` is `true`. Otherwise, this would cause a failure at startup if AWS KMS is enabled, but chat is disabled.

crates/router/src/configs/settings.rs
```diff
@@ -1048,7 +1049,6 @@ impl Settings<SecuredSecret> {
         self.secrets.get_inner().validate()?;
         self.locker.validate()?;
         self.connectors.validate("connectors")?;
-        self.chat.validate()?;
```
SanchithHegde, line 1051: Why did we remove this?

crates/router/src/core/errors/chat.rs
```diff
@@ -6,6 +6,8 @@ pub enum ChatErrors {
     MissingConfigError,
     #[error("Chat response deserialization failed")]
     ChatResponseDeserializationFailed,
+    #[error("Unauthorized access, internal role required")]
+    UnauthorizedAccess,
```
SanchithHegde: Do we want to hardcode "internal role required" in the error message, or can we remove it? In other words, are there situations where we may raise this error variant, but it may not require an internal role necessarily?

crates/router/src/db/hyperswitch_ai_interaction.rs
```diff
@@ -0,0 +1,122 @@
...
+        hyperswitch_ai_interaction: storage::HyperswitchAiInteractionNew,
+    ) -> CustomResult<storage::HyperswitchAiInteraction, errors::StorageError> {
+        let conn = connection::pg_connection_write(self).await?;
+        hyperswitch_ai_interaction
+            .insert(&conn)
+            .await
+            .map_err(|error| report!(errors::StorageError::from(error)))
+    }
+
+    async fn list_hyperswitch_ai_interactions(
```
SanchithHegde, line 42: Add an `#[instrument(skip_all)]` annotation here.

migrations/2025-07-21-120614_add_table_for_hyperswitch_ai_interactions/up.sql
```diff
@@ -0,0 +1,21 @@
...
+    user_id VARCHAR(64),
+    merchant_id VARCHAR(64),
+    profile_id VARCHAR(64),
+    org_id VARCHAR(64),
+    role_id VARCHAR(64),
+    user_query BYTEA,
+    response BYTEA,
+    database_query TEXT,
+    interaction_status VARCHAR(64),
+    created_at TIMESTAMP NOT NULL DEFAULT now(),
```
SanchithHegde: Let's avoid the default being provided in the table, let's force the application to always provide one.

crates/router/src/core/chat.rs
```diff
@@ -57,10 +77,127 @@ pub async fn get_data_from_hyperswitch_ai_workflow(
...
     .attach_printable("Error when sending request to AI service")?
-    .json::<_>()
+    .json::<chat_api::ChatResponse>()
     .await
     .change_context(ChatErrors::InternalServerError)
     .attach_printable("Error when deserializing response from AI service")?;
 
-    Ok(ApplicationResponse::Json(response))
+    let response_to_return = response.clone();
+    tokio::spawn(async move {
```
SanchithHegde: Add `.in_current_span()` when calling `tokio::spawn()`, refer to other usages of the method for an example.

crates/router/src/core/chat.rs
```diff
@@ -57,10 +77,127 @@ pub async fn get_data_from_hyperswitch_ai_workflow(
...
+
+    let mut conversations = Vec::new();
+
+    for interaction in hyperswitch_ai_interactions {
+        let user_query_encrypted = interaction
+            .user_query
+            .ok_or(ChatErrors::InternalServerError)?;
+        let response_encrypted = interaction
+            .response
+            .ok_or(ChatErrors::InternalServerError)?;
```
SanchithHegde: Attach context messages when raising internal server errors.

crates/router/src/core/chat.rs
```diff
@@ -57,10 +77,127 @@ pub async fn get_data_from_hyperswitch_ai_workflow(
...
+                logger::error!("Failed to construct hyperswitch_ai_interaction: {:?}", e);
+            }
+        }
+    });
+
+    Ok(ApplicationResponse::Json(response_to_return))
+}
+
+#[instrument(skip_all)]
+pub async fn list_chat_conversations(
```
SanchithHegde, line 119: Is this supposed to list messages in a conversation / session, or is this supposed to list conversations / sessions created by a user?

crates/router/src/routes/chat.rs
```diff
@@ -45,3 +45,24 @@ pub async fn get_data_from_hyperswitch_ai_workflow(
...
+    let flow = Flow::ListAllChatInteractions;
+    Box::pin(api::server_wrap(
+        flow.clone(),
+        state,
+        &http_req,
+        payload.into_inner(),
+        |state, user: auth::UserFromToken, payload, _| {
+            chat_core::list_chat_conversations(state, user, payload)
+        },
+        &auth::DashboardNoPermissionAuth,
```
SanchithHegde, line 64: No permissions required? So someone without merchant / profile payment read access would be able to get payment aggregates, etc.? This shouldn't be the case, right?

crates/common_utils/src/consts.rs
```diff
@@ -197,3 +197,9 @@ pub const REQUEST_TIME_OUT: u64 = 30;
 
 /// API client request timeout for ai service (in seconds)
 pub const REQUEST_TIME_OUT_FOR_AI_SERVICE: u64 = 120;
+
+/// Default limit for list operations (can be used across different entities)
+pub const DEFAULT_LIST_LIMIT: i64 = 100;
+
+/// Default offset for list operations (can be used across different entities)
+pub const DEFAULT_LIST_OFFSET: i64 = 0;
```
SanchithHegde, line 205: Since this is not being used anywhere other than the chat related code, should we consider moving these constants to the chat module?

crates/diesel_models/src/hyperswitch_ai_interaction.rs
apoorvdixit88: Role ID will be leveraged to track actions based on the assigned role. For example, profile-level users will not be allowed to access org-level information. The AI service can also utilize roles to perform differently and provide results accordingly, therefore storing role_id to keep the track.

crates/diesel_models/src/hyperswitch_ai_interaction.rs
apoorvdixit88: This can be optional. If a user is interacting outside the dashboard (without signing in), for example by chatting with docs, then we won’t have user_id or token-related information in that case

crates/router/src/core/chat.rs
apoorvdixit88: as per current implementation it can list all the messages across conversations for different users, and since its mostly made for internal purpose, other filtering in this list can be added as per use cases we try to cover.

crates/router/src/routes/chat.rs
apoorvdixit88: In the core function, there is a check to ensure that this only works for internal roles only. We cannot separate org, merchant and profile permissions for an internal role. If the role is internal and the user has a valid JWT token (verified by DashboardNoPermissionAuth), access is granted.

crates/common_utils/src/consts.rs
apoorvdixit88: Since every time we create something new, we usually move it to the respective domain/module, I thought of keeping this as a default so that there’s a baseline value to consider. This way, it can be utilised in cases where the value isn’t supposed to be new feature-specific, rather than duplicating defaults across modules.

migrations/2025-07-21-120614_add_table_for_hyperswitch_ai_interactions/up.sql
```diff
@@ -0,0 +1,21 @@
...
+    merchant_id VARCHAR(64),
+    profile_id VARCHAR(64),
+    org_id VARCHAR(64),
+    role_id VARCHAR(64),
+    user_query BYTEA,
+    response BYTEA,
+    database_query TEXT,
+    interaction_status VARCHAR(64),
+    created_at TIMESTAMP NOT NULL,
+    PRIMARY KEY (id, created_at)
```
SanchithHegde, line 15: Why are we including `created_at` in the primary key, is it because of the table being partitioned on that column?

SanchithHegde (commented):

migrations/2025-07-21-120614_add_table_for_hyperswitch_ai_interactions/up.sql
JeevaRamu0104: yes

SanchithHegde (approved):

inventvenkat (approved):

PR #8959, opened 2025-08-14
feat(injector): add support for new crate -  injector for external vault proxy

su-shivanshmathur:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

This is the injector crate which would be used for External Vault proxy usage

Screenshot 2025-07-09 at 12 03 03 PM

Screenshot 2025-07-09 at 12 15 08 PM

### Additional Changes

## Motivation and Context

Needed a generic way of passing on text as card number which would be swaped on vaults's end for not having PCI compliant data in router.

## How did you test it?

Have added a test case in the crate itself

Screenshot 2025-08-05 at 5 48 40 PM

Screenshot 2025-08-05 at 5 49 22 PM

## Checklist

### DIFF
--- a/crates/injector/src/injector.rs
+++ b/crates/injector/src/injector.rs
@@ -21,11 +21,11 @@ pub mod core {
     impl From<injector_types::HttpMethod> for Method {
         fn from(method: injector_types::HttpMethod) -> Self {
             match method {
-                injector_types::HttpMethod::GET => Method::Get,
-                injector_types::HttpMethod::POST => Method::Post,
-                injector_types::HttpMethod::PUT => Method::Put,
-                injector_types::HttpMethod::PATCH => Method::Patch,
-                injector_types::HttpMethod::DELETE => Method::Delete,
+                injector_types::HttpMethod::GET => Self::Get,
+                injector_types::HttpMethod::POST => Self::Post,
+                injector_types::HttpMethod::PUT => Self::Put,
+                injector_types::HttpMethod::PATCH => Self::Patch,
+                injector_types::HttpMethod::DELETE => Self::Delete,
             }
         }
     }


crates/injector/src/lib.rs
```diff
@@ -0,0 +1,783 @@
...
+                            )
+                            .map(|processed| (key, processed))
+                        })
+                        .collect::<error_stack::Result<serde_json::Map<_, _>, InjectorError>>()?;
+                    Ok(Value::Object(new_obj))
+                }
+                Value::String(s) => {
+                    // Use regex to find all tokens and replace them
+                    use regex::Regex;
+                    let token_regex = Regex::new(r"\{\{\$([a-zA-Z_][a-zA-Z0-9_]*)\}\}")
```
prasunna09: can we please use nom parser instead of regex.

crates/injector/src/lib.rs
```diff
@@ -0,0 +1,783 @@
...
+                        .change_context(InjectorError::InvalidTemplate(
+                            "Invalid regex pattern".to_string(),
+                        ))?;
+                    let mut result = s.clone();
+
+                    for captures in token_regex.captures_iter(&s) {
```
prasunna09: we can use into_iter

crates/api_models/src/injector.rs
```diff
@@ -0,0 +1,151 @@
...
+        PartialEq,
+        Deserialize,
+        Serialize,
+        strum::Display,
+        strum::EnumString,
+        ToSchema,
+    )]
+    #[serde(rename_all = "UPPERCASE")]
+    #[strum(serialize_all = "UPPERCASE")]
+    pub enum VaultType {
```
prasunna09: Instead of VaultType, can we rename it to VaultConnectors?

crates/api_models/src/injector.rs
```diff
@@ -0,0 +1,151 @@
...
+        pub endpoint_path: String,
+        pub http_method: HttpMethod,
+        pub headers: HashMap<String, Secret<String>>,
+        pub proxy_url: Option<String>,
+        pub client_cert: Option<Secret<String>>,
+        pub client_key: Option<Secret<String>>,
+        pub ca_cert: Option<Secret<String>>,
+        pub insecure: Option<bool>,
+        pub cert_password: Option<Secret<String>>,
+        pub cert_format: Option<String>,
```
prasunna09: can we have doc comments for these fields?

crates/hyperswitch_domain_models/src/injector.rs
```diff
@@ -0,0 +1,334 @@
...
+    }
+
+    #[derive(Clone, Debug, PartialEq, Eq)]
+    pub enum HttpMethod {
+        GET,
+        POST,
+        PUT,
+        PATCH,
+        DELETE,
+    }
```
prasunna09: if we have same enums defined in api, and domain, shouldn't these be in common_enums.rs?

crates/hyperswitch_domain_models/src/injector.rs
```diff
@@ -0,0 +1,334 @@
...
+            match method {
+                api_models::injector::HttpMethod::GET => Self::GET,
+                api_models::injector::HttpMethod::POST => Self::POST,
+                api_models::injector::HttpMethod::PUT => Self::PUT,
+                api_models::injector::HttpMethod::PATCH => Self::PATCH,
+                api_models::injector::HttpMethod::DELETE => Self::DELETE,
+            }
+        }
+    }
+
```
prasunna09: if we move this enum to common_enums, then we dont require this transformations right?

crates/hyperswitch_domain_models/src/injector.rs
```diff
@@ -0,0 +1,334 @@
...
+    use std::collections::HashMap;
+
+    use crate::ApiModelToDieselModelConvertor;
+
+    #[derive(Clone, Debug, PartialEq, Eq)]
+    pub enum VaultType {
+        VGS,
+    }
+
+    impl ApiModelToDieselModelConvertor<api_models::injector::VaultType> for VaultType {
```
prasunna09: why are we using api model to diesel model conversions?
it has to be api to domain model right?

crates/hyperswitch_interfaces/src/api.rs
```diff
@@ -395,7 +395,7 @@ pub trait ConnectorSpecifications {
         false
     }
 
-    #[cfg(not(feature = "v2"))]
+    #[cfg(feature = "v1")]
```
prasunna09: is this required?

crates/injector/src/lib.rs
```diff
@@ -0,0 +1,783 @@
...
+
+            // Extract token data from SecretSerdeValue for vault data lookup
+            let vault_data = domain_request
+                .token_data
+                .specific_token_data
+                .expose()
+                .clone();
+
+            // Validate template length to prevent potential memory issues
+            if domain_request.connector_payload.template.len() > 1_000_000 {
```
prasunna09: why do we require this check?

crates/injector/src/lib.rs
```diff
@@ -0,0 +1,783 @@
...
+            }
+
+            logger::debug!(
+                template_length = domain_request.connector_payload.template.len(),
+                vault_type = ?domain_request.token_data.vault_type,
+                "Processing token injection request"
+            );
+
+            // Process template string directly with vault-specific logic
+            let template_value = Value::String(domain_request.connector_payload.template);
```
prasunna09: why are we wrapping the template with Value::String?

crates/injector/src/lib.rs
```diff
@@ -0,0 +1,783 @@
...
+                        .map(|(key, val)| {
+                            self.interpolate_token_references_with_vault_data(
+                                val, vault_data, vault_type,
+                            )
+                            .map(|processed| (key, processed))
+                        })
+                        .collect::<error_stack::Result<serde_json::Map<_, _>, InjectorError>>()?;
+                    Ok(Value::Object(new_obj))
+                }
+                Value::String(s) => {
```
prasunna09: since vault_data is being wrapped in value:: string earlier, other than this arm, nothing will be used?

prasunna09:
```suggestion
                Value::String(token_value) => {
```
please have a better naming convention

crates/injector/src/lib.rs
```diff
@@ -0,0 +1,783 @@
...
+
+            let processed_payload = match processed_value {
+                Value::String(s) => s,
+                _ => {
+                    // This shouldn't happen since we started with a string
+                    return Err(error_stack::Report::new(InjectorError::InvalidTemplate(
+                        "Template processing resulted in non-string value".to_string(),
+                    )));
+                }
+            };
```
prasunna09:
```suggestion
                   Err(error_stack::Report::new(InjectorError::InvalidTemplate(
                        "Template processing resulted in non-string value".to_string(),
                    )))?
                }
            };
```

crates/injector/src/lib.rs
```diff
@@ -0,0 +1,783 @@
...
+
+            logger::debug!(
+                processed_payload_length = processed_payload.len(),
```
prasunna09: why do we need length here?

crates/injector/src/lib.rs
```diff
@@ -0,0 +1,783 @@
...
+                    "application/xml" => Some(ContentType::ApplicationXml),
+                    "text/xml" => Some(ContentType::TextXml),
+                    "text/plain" => Some(ContentType::TextPlain),
+                    _ => None,
+                })
+                .unwrap_or(ContentType::ApplicationXWwwFormUrlencoded);
+
+            // Make HTTP request to connector and return raw response
+            let response_data = self
+                .make_http_request(
```
prasunna09: do we need this fn? can't we re-use existing `call_connector_api`?

crates/api_models/src/injector.rs
```diff
@@ -0,0 +1,72 @@
...
+    pub struct ConnectorPayload {
+        /// Template string containing token references in the format {{$field_name}}
+        pub template: String,
+    }
+
+    /// Configuration for HTTP connection to the external connector
+    #[derive(Clone, Debug, Deserialize, Serialize, ToSchema)]
+    pub struct ConnectionConfig {
+        /// Base URL of the connector endpoint
+        pub base_url: String,
```
prasunna09:
```suggestion
        pub base_url: url::Url,
```

crates/api_models/src/injector.rs
```diff
@@ -0,0 +1,72 @@
...
+        /// Path to append to the base URL for the specific endpoint
+        pub endpoint_path: String,
+        /// HTTP method to use for the request
+        pub http_method: HttpMethod,
+        /// HTTP headers to include in the request
+        pub headers: HashMap<String, Secret<String>>,
+        /// Optional proxy URL for routing the request
+        pub proxy_url: Option<String>,
```
prasunna09: What is this proxy url for? why does UCS need to send it injector?
or will this be shared by Payments service to UCS in request's metadata?

prasunna09:
```suggestion
        pub proxy_url: Option<Url>,
```

crates/hyperswitch_domain_models/src/injector.rs
```diff
@@ -0,0 +1,114 @@
...
+        InjectorHttpMethod as HttpMethod, InjectorVaultConnectors as VaultConnectors,
+    };
+
+    /// Domain model for token data containing vault-specific information
+    #[derive(Clone, Debug)]
+    pub struct TokenData {
+        /// The specific token data retrieved from the vault, containing sensitive PII
+        pub specific_token_data: common_utils::pii::SecretSerdeValue,
+        /// The type of vault connector being used for token retrieval
+        pub vault_type: VaultConnectors,
```
prasunna09:
```suggestion
        pub vault_connector: VaultConnectors,
```

crates/hyperswitch_connectors/src/connectors/redsys.rs
```diff
@@ -904,7 +904,6 @@ impl ConnectorSpecifications for Redsys {
         Some(&REDSYS_SUPPORTED_WEBHOOK_FLOWS)
     }
 
-    #[cfg(feature = "v1")]
```
prasunna09: please revert this change if not needed.

crates/api_models/src/injector.rs
jarnura, line 1: Injector models can be in injector crate itself, so that the making this as standalone library will don't have dependency

crates/common_enums/src/enums.rs
jarnura, line 1: Injector enums can be in injector crate itself, so that the making this as standalone library will don't have dependency

crates/hyperswitch_connectors/src/connectors/redsys.rs
```diff
@@ -919,4 +919,17 @@ impl ConnectorSpecifications for Redsys {
...
         }
     }
+
+    #[cfg(feature = "v2")]
+    fn generate_connector_request_reference_id(
+        &self,
+        payment_intent: &hyperswitch_domain_models::payments::PaymentIntent,
+        _payment_attempt: &hyperswitch_domain_models::payments::payment_attempt::PaymentAttempt,
+    ) -> String {
+        payment_intent
```
jarnura: Is this change required?

crates/hyperswitch_domain_models/src/injector.rs
jarnura, line 1: Injector models can be in injector crate itself, so that the making this as standalone library will don't have dependency

crates/injector/src/injector.rs
```diff
@@ -0,0 +1,791 @@
...
+                endpoint = %config.endpoint_path,
+                content_type = ?content_type,
+                payload_length = payload.len(),
+                headers_count = config.headers.len(),
+                "Making HTTP request to connector"
+            );
+            // Validate inputs first
+            if config.endpoint_path.is_empty() {
+                logger::error!("Endpoint path is empty");
+                return Err(error_stack::Report::new(InjectorError::InvalidTemplate(
```
jarnura: instead of return use ? pattern to throw errors

crates/injector/src/injector.rs
```diff
@@ -0,0 +1,791 @@
...
+                .change_context(InjectorError::HttpRequestFailed)?;
+
+            // Validate response text length to prevent potential memory issues
+            if response_text.len() > 10_000_000 {
+                // 10MB limit
+                logger::error!(
+                    response_length = response_text.len(),
+                    "Response from connector is too large, potential DoS or memory exhaustion"
+                );
+                return Err(error_stack::Report::new(InjectorError::HttpRequestFailed));
```
jarnura: instead of return use ? pattern to throw errors

crates/api_models/src/injector.rs
```diff
@@ -0,0 +1,78 @@
...
+    pub struct InjectorRequest {
+        /// Token data from the vault
+        pub token_data: TokenData,
+        /// Payload template to process
+        pub connector_payload: ConnectorPayload,
+        /// HTTP connection configuration
+        pub connection_config: ConnectionConfig,
+    }
+
+    pub type InjectorResponse = serde_json::Value;
```
jarnura: Should we make this as secret? Since this is a untyped value

crates/hyperswitch_domain_models/src/injector.rs
```diff
@@ -0,0 +1,115 @@
...
+        fn from(request: api_models::injector::InjectorRequest) -> Self {
+            Self {
+                token_data: request.token_data.into(),
+                connector_payload: request.connector_payload.into(),
+                connection_config: request.connection_config.into(),
+            }
+        }
+    }
+
+    pub type InjectorResponse = serde_json::Value;
```
jarnura: Make this as secret

crates/hyperswitch_connectors/src/connectors/redsys.rs
```diff
@@ -921,4 +921,17 @@ impl ConnectorSpecifications for Redsys {
...
+    //     &self,
+    //     payment_intent: &hyperswitch_domain_models::payments::PaymentIntent,
+    //     _payment_attempt: &hyperswitch_domain_models::payments::payment_attempt::PaymentAttempt,
+    // ) -> String {
+    //     payment_intent
+    //         .merchant_reference_id
+    //         .as_ref()
+    //         .map(|id| id.get_string_repr().to_owned())
+    //         .unwrap_or_else(|| connector_utils::generate_12_digit_number().to_string())
+    // }
```
prasunna09: please revert this change. seems like this is irrelevant change

crates/hyperswitch_interfaces/src/api.rs
prasunna09: if not required, please revert this

crates/injector/src/injector.rs
```diff
@@ -0,0 +1,792 @@
...
+            &self,
+            template: String,
+            vault_data: &Value,
+            vault_connector: &injector_types::VaultConnectors,
+        ) -> error_stack::Result<String, InjectorError> {
+            // Find all tokens using nom parser
+            let tokens = find_all_tokens(&template);
+            let mut result = template;
+
+            for token_ref in tokens {
```
prasunna09: nit: Can use into_iter

prasunna09 (changes requested):

crates/injector/src/types.rs
```diff
@@ -0,0 +1,220 @@
...
+    /// Content types supported by the injector for HTTP requests
+    #[derive(Clone, Copy, Debug, Eq, PartialEq, serde::Serialize, serde::Deserialize)]
+    #[serde(rename_all = "snake_case")]
+    pub enum ContentType {
+        ApplicationJson,
+        ApplicationXWwwFormUrlencoded,
+        ApplicationXml,
+        TextXml,
+        TextPlain,
+    }
```
prasunna09, line 20: good that we moved to this crate.
cargo hack has been failing due to this particular structure is in hyperswitch interfaces which indirectly uses domain models and others.

but what of maintenance part? for the similar structures as well?
ps. not a blocker currently
CC: @jarnura

crates/injector/src/injector.rs
```diff
@@ -0,0 +1,891 @@
...
+                .collect();
+
+            // Determine method and request content
+            let method = match config.http_method {
+                injector_types::HttpMethod::GET => Method::Get,
+                injector_types::HttpMethod::POST => Method::Post,
+                injector_types::HttpMethod::PUT => Method::Put,
+                injector_types::HttpMethod::PATCH => Method::Patch,
+                injector_types::HttpMethod::DELETE => Method::Delete,
+            };
```
prasunna09: we can write a from impl for this

crates/injector/src/injector.rs
```diff
@@ -0,0 +1,891 @@
...
+                .await
+                .change_context(InjectorError::HttpRequestFailed)?;
+
+            // Validate response text length to prevent potential memory issues
+            if response_text.len() > 10_000_000 {
+                // 10MB limit
+                logger::error!(
+                    response_length = response_text.len(),
+                    "Response from connector is too large, potential DoS or memory exhaustion"
+                );
```
prasunna09: why are we checking the response length? do we do this for normal connector call as well??

crates/injector/src/injector.rs
```diff
@@ -0,0 +1,887 @@
...
+        })?;
+
+        // Build the request
+        let method = match request.method {
+            Method::Get => reqwest::Method::GET,
+            Method::Post => reqwest::Method::POST,
+            Method::Put => reqwest::Method::PUT,
+            Method::Patch => reqwest::Method::PATCH,
+            Method::Delete => reqwest::Method::DELETE,
+        };
```
prasunna09, line 100: can be a from impl

crates/injector/src/injector.rs
```diff
@@ -0,0 +1,887 @@
...
+                    req_builder = req_builder.json(&payload);
+                }
+                RequestContent::FormUrlEncoded(payload) => {
+                    req_builder = req_builder.form(&payload);
+                }
+                RequestContent::RawBytes(payload) => {
+                    req_builder = req_builder.body(payload);
+                }
+                _ => {
+                    logger::warn!("Unsupported request content type, using raw bytes");
```
prasunna09, line 126: throw an error instead of warning

crates/injector/src/injector.rs
```diff
@@ -0,0 +1,887 @@
...
+        let response = req_builder.send().await.map_err(|e| {
+            logger::error!("HTTP request failed: {}", e);
+            error_stack::Report::new(InjectorError::HttpRequestFailed)
+        })?;
+
+        Ok(response)
+    }
+
+    #[derive(Error, Debug)]
+    pub enum InjectorError {
```
prasunna09, line 141: Ideally this `HttpRequestFailed` should be API error, and injector errors are different which would include `TokenReplacementFailed`.

crates/injector/src/injector.rs
```diff
@@ -0,0 +1,887 @@
...
+                            .into_owned()
+                            .collect();
+                    Some(RequestContent::FormUrlEncoded(Box::new(form_data)))
+                }
+                ContentType::ApplicationXml | ContentType::TextXml => {
+                    Some(RequestContent::RawBytes(payload.as_bytes().to_vec()))
+                }
+                ContentType::TextPlain => {
+                    Some(RequestContent::RawBytes(payload.as_bytes().to_vec()))
+                }
```
prasunna09, line 445:
```suggestion
                ContentType::ApplicationXml | ContentType::TextXml | 
                ContentType::TextPlain => {
                    Some(RequestContent::RawBytes(payload.as_bytes().to_vec()))
                }
```

jarnura (approved):

PR #8966, opened 2025-08-17
refactor(gRPC): refactor Revenue-recovery decider gRPC client side

AdityaKumaar21:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [x] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

### Additional Changes

## Checklist

### DIFF
--- a/crates/hyperswitch_connectors/src/connectors/cybersource/transformers.rs
+++ b/crates/hyperswitch_connectors/src/connectors/cybersource/transformers.rs
@@ -169,7 +169,6 @@ impl TryFrom<&SetupMandateRouterData> for CybersourceZeroMandateRequest {
             PaymentMethodData::Card(ccard) => {
                 let card_type = match ccard
                     .card_network
-                    .clone()
                     .and_then(get_cybersource_card_type)
                 {
                     Some(card_network) => Some(card_network.to_string()),
@@ -1365,9 +1364,9 @@ impl
             })
             .ok();
 
-        let raw_card_type = ccard.card_network.clone().or(additional_card_network);
+        let raw_card_type = ccard.card_network.or(additional_card_network);
 
-        let card_type = match raw_card_type.clone().and_then(get_cybersource_card_type) {
+        let card_type = match raw_card_type.and_then(get_cybersource_card_type) {
             Some(card_network) => Some(card_network.to_string()),
             None => ccard.get_card_issuer().ok().map(String::from),
         };
@@ -1854,7 +1853,6 @@ impl
 
         let card_type = match ccard
             .card_network
-            .clone()
             .and_then(get_cybersource_card_type)
         {
             Some(card_network) => Some(card_network.to_string()),
@@ -2133,10 +2131,11 @@ impl
                         })?,
                 },
             }));
+        let card_network = google_pay_data.info.card_network.clone();
         let processing_information = ProcessingInformation::try_from((
             item,
             Some(PaymentSolution::GooglePay),
-            Some(google_pay_data.info.card_network.clone()),
+            Some(card_network.clone()),
         ))?;
         let client_reference_information = ClientReferenceInformation::from(item);
         let merchant_defined_information = item
@@ -2147,7 +2146,7 @@ impl
             .map(convert_metadata_to_merchant_defined_info);
 
         let ucaf_collection_indicator =
-            match google_pay_data.info.card_network.to_lowercase().as_str() {
+            match card_network.to_lowercase().as_str() {
                 "mastercard" => Some("2".to_string()),
                 _ => None,
             };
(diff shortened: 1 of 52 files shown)


crates/hyperswitch_connectors/src/connectors/cybersource/transformers.rs
```diff
@@ -2133,10 +2125,11 @@ impl
                         })?,
                 },
             }));
+        let card_network = google_pay_data.info.card_network.clone();
         let processing_information = ProcessingInformation::try_from((
             item,
             Some(PaymentSolution::GooglePay),
-            Some(google_pay_data.info.card_network.clone()),
+            Some(card_network.clone()),
```
SanchithHegde, line 2174: Why was this code modified when the `card_network` field here seems to be a `String`?

crates/router/src/core/payment_methods/surcharge_decision_configs.rs
```diff
@@ -414,7 +414,7 @@ pub async fn perform_surcharge_decision_management_for_saved_cards(
 
 //         let card_network = match customer_payment_method.payment_method_data.as_ref() {
 //             Some(api_models::payment_methods::PaymentMethodListData::Card(card)) => {
-//                 card.card_network.clone()
```
SanchithHegde, line 417: Umm, I think we can revert this change, since we aren't aware if this would compile or not.

crates/router/src/core/revenue_recovery.rs
```diff
@@ -273,9 +273,9 @@ pub async fn perform_payments_sync(
         pcr_status.update_pt_status_based_on_attempt_status_for_payments_sync(
             state,
             payment_intent,
-            process.clone(),
+            &process.clone(),
```
SanchithHegde: If you're anyway accepting a reference to the `ProcessTracker` object, then why do you need the clone?

crates/router/src/core/revenue_recovery/types.rs
```diff
@@ -385,7 +389,7 @@ impl Action {
                     Self::decide_retry_failure_action(
                         state,
                         merchant_id,
-                        process.clone(),
+                        &process.clone(),
```
SanchithHegde: Same here, if you only need a reference, why the clone?

crates/router/src/core/revenue_recovery/types.rs
```diff
@@ -563,22 +567,22 @@ impl Action {
...
                 RevenueRecoveryPaymentsAttemptStatus::Succeeded => {
-                    Ok(Self::SuccessfulPayment(payment_attempt))
+                    Ok(Self::SuccessfulPayment(payment_attempt.clone()))
                 }
                 RevenueRecoveryPaymentsAttemptStatus::Failed => {
                     Self::decide_retry_failure_action(
                         state,
                         revenue_recovery_payment_data.merchant_account.get_id(),
-                        process.clone(),
+                        &process.clone(),
```
SanchithHegde: Same here.

crates/router/src/core/webhooks/recovery_incoming.rs
```diff
@@ -200,7 +200,11 @@ async fn handle_monitoring_threshold(
 ) -> CustomResult<webhooks::WebhookResponseTracker, errors::RevenueRecoveryError> {
     let db = &*state.store;
     let key_manager_state = &(state).into();
-    let monitoring_threshold_config = state.conf.revenue_recovery.monitoring_threshold_in_seconds;
+    let monitoring_threshold_config = state
+        .conf
+        .revenue_recovery
+        .monitoring_threshold_in_seconds
+        .into();
```
SanchithHegde, line 221: Prefer calling `T::from()` over `.into()`, makes it easier to understand what the resultant type is, when viewing code without an IDE.

crates/router/src/types/storage/revenue_recovery.rs
```diff
@@ -67,14 +67,14 @@ impl RevenueRecoveryPaymentData {
 
 #[derive(Debug, serde::Deserialize, Clone, Default)]
 pub struct RevenueRecoverySettings {
```
SanchithHegde, line 71: Same concern as I'd pointed out before (https://github.com/juspay/hyperswitch/pull/8178#discussion_r2253903974), because of the `Default` derive, the value for `monitoring_threshold_in_seconds` field would be `0` if it none of the fields in this struct are provided. Are you okay with that?

Also, why is the `revenue_recovery` field in the `Settings` struct required (when the `v2` feature flag is enabled)? Please make that field optional, there are cases where we're deploying v2 builds internally but don't need any revenue recovery features to be enabled...

And consider moving this to the `settings.rs` file maybe?

PR #8970, opened 2025-08-18
fix(common_utils): Prevent logging sensitive information on deserialization failure

### LINKED ISSUE #8969: [BUG] Logging sensitive information on deserialization failure
### Bug Description

In `common_utils/src/ext_traits.rs` we are logging serde_json Values in plaintext when deserialization fails. This is a problem in cases where the JSON objects contain sensitive data.

### Expected Behavior

Sensitive data is not logged. Values are logged after masking.

### Actual Behavior

Sensitive information is logged.

### Steps To Reproduce

Supply malformed data to the `parse_value` method in `ValueExt` trait.

### Context For The Bug

_No response_

### Environment

Are you using hyperswitch hosted version? Yes

AnuthaDev:
## Type of Change

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description
- Modified `parse_value` method in `ValueExt` trait to log the JSON object after masking on deserialization failure.
- Modified `parse_struct` method in `BytesExt`, `ByteSliceExt` and `StringExt` to log values after converting them into a `serde_json::Value` and then masking

### Additional Changes

## Motivation and Context

Closes #8969 

## How did you test it?

Tested by supplying malformed data to the method and verified it logs data after masking.

Object:
image

String:
image

Array:
image

## Checklist

### DIFF
--- a/crates/common_utils/src/ext_traits.rs
+++ b/crates/common_utils/src/ext_traits.rs
@@ -175,7 +175,7 @@ impl BytesExt for bytes::Bytes {
             .change_context(errors::ParsingError::StructParseFailure(type_name))
             .attach_printable_lazy(|| {
                 let variable_type = std::any::type_name::<T>();
-                let value = serde_json::from_slice::<serde_json::Value>(&self)
+                let value = serde_json::from_slice::<serde_json::Value>(self)
                     .unwrap_or_else(|_| serde_json::Value::String(String::new()));
 
                 format!(
@@ -209,7 +209,7 @@ impl ByteSliceExt for [u8] {
         serde_json::from_slice(self)
             .change_context(errors::ParsingError::StructParseFailure(type_name))
             .attach_printable_lazy(|| {
-                let value = serde_json::from_slice::<serde_json::Value>(&self)
+                let value = serde_json::from_slice::<serde_json::Value>(self)
                     .unwrap_or_else(|_| serde_json::Value::String(String::new()));
 
                 format!(


crates/common_utils/src/ext_traits.rs
```diff
@@ -219,10 +219,7 @@ impl ValueExt for serde_json::Value {
     where
         T: serde::de::DeserializeOwned,
     {
-        let debug = format!(
-            "Unable to parse {type_name} from serde_json::Value: {:?}",
-            &self
-        );
+        let debug = format!("Unable to parse {type_name} from serde_json::Value");
```
sai-harsha-vardhan: Can we atleast try to log the available keys here? It would be very hard to debug any PROD issue without any context of original data point 
cc: @jarnura @SanchithHegde

AnuthaDev: Yes, we will change this to log the object after masking values. Currently doing this to fix the alerts ASAP.

In the meantime, We will still have context in the error message about what went wrong:

image

See, it mentions which field was missing

sai-harsha-vardhan: In some cases, the error message won't be helpful (it would be something like did not match any enum variant). Should we fix this properly and push it to PROD? Can we bear some alerts for couple of days?

AnuthaDev: Changed to log keys with masked values:

image

crates/common_utils/src/ext_traits.rs
```diff
@@ -221,7 +221,8 @@ impl ValueExt for serde_json::Value {
     {
         let debug = format!(
             "Unable to parse {type_name} from serde_json::Value: {:?}",
-            &self
+            // Required to prevent logging sensitive data in case of deserialization failure
```
sai-harsha-vardhan: We should handle this in other places like `parse_struct` as well, right?

crates/common_utils/src/ext_traits.rs
```diff
@@ -219,13 +233,15 @@ impl ValueExt for serde_json::Value {
     where
         T: serde::de::DeserializeOwned,
     {
-        let debug = format!(
-            "Unable to parse {type_name} from serde_json::Value: {:?}",
-            &self
-        );
-        serde_json::from_value::<T>(self)
+        serde_json::from_value::<T>(self.clone())
```
SanchithHegde, line 236: Is there an option for us to avoid the clone here and do the clone within the `attach_printable_lazy()` closure?

AnuthaDev: We could do a `T::deserialize(&self)`, but reading through the serde documentation, passing an owned value allows serde to use a previously allocated buffer. If we pass a reference it cannot do so, ultimately defeating the purpose of removing the clone.

Apart from this I couldn't really find a way.

SanchithHegde (approved):

PR #9008, opened 2025-08-21
feat(connector): [SANTANDER] Added Boleto Payment Method

bsayak03:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

Added Authorize and IncomingWebhooks flow for Boleto Payment Method in Santander Connector

### Additional Changes

## Motivation and Context

There isn't any documentation but an excel sheet exists for request response schemas. 

## How did you test it?

This is an alpha integration and it cannot be tested. (Awaiting on creds)

## Checklist


crates/common_enums/src/enums.rs
```diff
@@ -571,6 +571,27 @@ pub enum CallConnectorAction {
     HandleResponse(Vec<u8>),
 }
 
+#[derive(
```
dgeee13, line 579: are all the derives needed?

crates/hyperswitch_connectors/src/connectors/santander.rs
```diff
@@ -279,26 +300,49 @@ impl ConnectorIntegration<Authorize, PaymentsAuthorizeData, PaymentsResponseData
...
-        Ok(format!(
-            "{}cob/{}",
-            self.base_url(connectors),
-            req.payment_id
-        ))
+        let santander_mca_metadata =
+            santander::SantanderMetadataObject::try_from(&req.connector_meta_data)?;
+        match req.request.payment_method_type {
+            Some(enums::PaymentMethodType::Pix) => Ok(format!(
+                "{}cob/{}",
```
dgeee13: can you please verify if the url is correct or should it be like the following instead?
```suggestion
                "{}/cob/{}",
```

crates/hyperswitch_connectors/src/connectors/santander.rs
```diff
@@ -563,12 +619,21 @@ impl ConnectorIntegration<Void, PaymentsCancelData, PaymentsResponseData> for Sa
...
-        Ok(format!(
-            "{}cob/{}",
-            self.base_url(connectors),
-            connector_payment_id
-        ))
+        match req.payment_method {
+            enums::PaymentMethod::BankTransfer => {
+                let connector_payment_id = req.request.connector_transaction_id.clone();
+                Ok(format!(
+                    "{}cob/{}",
```
dgeee13, line 709: here as well 
https://github.com/juspay/hyperswitch/pull/9008/files#r2293018500

crates/hyperswitch_connectors/src/connectors/santander.rs
```diff
@@ -644,25 +709,34 @@ impl ConnectorIntegration<Execute, RefundsData, RefundsResponseData> for Santand
...
+                    .and_then(|val| val.as_str().map(|id| id.to_string()))
+                    .ok_or_else(|| errors::ConnectorError::MissingRequiredField {
+                        field_name: "end_to_end_id",
+                    })?;
+
+                let refund_id = req.request.connector_refund_id.clone();
+                Ok(format!(
+                    "{}{}{}{}{:?}",
+                    self.base_url(connectors),
+                    "pix/",
```
dgeee13, line 811: here as well - similar to previous comment, please check if the url is correct
https://github.com/juspay/hyperswitch/pull/9008/files#r2293018500

crates/common_enums/src/enums.rs
bsayak03: needed because this goes in the api models where Schema and all is required since the openapispec file gets also generated

crates/hyperswitch_connectors/src/connectors/santander.rs
bsayak03: the first `/` is not needed because the base url itself has a `/` at the last

crates/hyperswitch_connectors/src/connectors/santander.rs
bsayak03: the first `/` is not needed because the base url itself has a `/` at the last

crates/hyperswitch_connectors/src/connectors/santander.rs
bsayak03: not needed, all the `/` are covered

crates/api_models/src/payments.rs
```diff
@@ -4153,11 +4153,43 @@ pub struct RewardData {
...
+    #[schema(value_type = Option<i64>)]
+    pub write_off_quantity_days: Option<MinorUnit>,
+
+    /// Custom messages or instructions to display on the boleto
+    #[schema(value_type = Option<Vec<String>>)]
+    pub messages: Option<Vec<String>>,
+
+    #[schema(value_type = Option<NaiveDate>)]
+    // The date upon which the boleto is due and is of format: "YYYY-MM-DD"
+    pub due_date: Option<chrono::NaiveDate>,
```
dgeee13: can we directly use `time::Date` instead of `chrono::NaiveDate`?
```suggestion
    #[schema(value_type = Option<Date>)]
    // The date upon which the boleto is due and is of format: "YYYY-MM-DD"
    pub due_date: Option<time::Date>,
```

crates/hyperswitch_connectors/src/connectors/santander/transformers.rs
```diff
@@ -211,6 +227,112 @@ impl TryFrom<&SantanderRouterData<&PaymentsAuthorizeRouterData>> for SantanderPa
...
+                    .join(" "),
+                ),
+                neighborhood: value.0.router_data.get_billing_line1()?,
+                city: value.0.router_data.get_billing_city()?,
+                state: value.0.router_data.get_billing_state()?,
+                zipcode: value.0.router_data.get_billing_zip()?,
+            },
+            beneficiary: None,
+            document_kind: BoletoDocumentKind::BoletoProposta, // to change
+<<<<<<< Updated upstream
```
dgeee13: please update this

crates/common_utils/src/types.rs
```diff
@@ -602,9 +602,16 @@ where
     }
 }
 
+impl PartialEq for FloatMajorUnit {
```
dgeee13: what is the use-case for this custom PartialEq?

crates/api_models/src/payments.rs
bsayak03: we can, but no such difference

crates/common_utils/src/types.rs
bsayak03: FloatMajorUnit was a data type of a field in one of the structs which derived PartialEq and Eq. Hence needed this custom one. However, i've changed its data type to just String now. So no need, removing

crates/api_models/src/payments.rs
dgeee13: If there isn’t a functional difference, we could stick with Date from the existing crate and avoid adding a dependency on Chrono.

crates/hyperswitch_connectors/src/connectors/santander.rs
```diff
@@ -644,25 +753,34 @@ impl ConnectorIntegration<Execute, RefundsData, RefundsResponseData> for Santand
...
-        Ok(format!(
-            "{}{}{}{}{:?}",
-            self.base_url(connectors),
-            "pix/",
-            end_to_end_id,
-            "/refund/",
-            refund_id
-        ))
+        match req.payment_method {
+            enums::PaymentMethod::BankTransfer => {
```
dgeee13, line 796: We could consider matching on req.request.payment_method first and then on req.request.payment_method_type. That could make the logic a bit more explicit - you could  make it the same for the other flows too.

crates/hyperswitch_connectors/src/connectors/santander.rs
```diff
@@ -830,27 +948,59 @@ impl ConnectorIntegration<RSync, RefundsData, RefundsResponseData> for Santander
...
-    fn get_webhook_object_reference_id(
+    async fn verify_webhook_source(
         &self,
         _request: &webhooks::IncomingWebhookRequestDetails<'_>,
+        _merchant_id: &common_utils::id_type::MerchantId,
+        _connector_webhook_details: Option<common_utils::pii::SecretSerdeValue>,
+        _connector_account_details: crypto::Encryptable<Secret<serde_json::Value>>,
+        _connector_name: &str,
+    ) -> CustomResult<bool, errors::ConnectorError> {
+        Ok(true)
```
dgeee13: Please add a comment here for hardcoding `Ok(true)` here mentioning the current integration limitations

crates/hyperswitch_connectors/src/connectors/santander.rs
```diff
@@ -830,27 +948,59 @@ impl ConnectorIntegration<RSync, RefundsData, RefundsResponseData> for Santander
...
     }
 
     fn get_webhook_resource_object(
         &self,
-        _request: &webhooks::IncomingWebhookRequestDetails<'_>,
+        request: &webhooks::IncomingWebhookRequestDetails<'_>,
     ) -> CustomResult<Box<dyn masking::ErasedMaskSerialize>, errors::ConnectorError> {
-        Err(report!(errors::ConnectorError::WebhooksNotImplemented))
+        let webhook_body = transformers::get_webhook_object_from_body(request.body)
+            .change_context(errors::ConnectorError::WebhookEventTypeNotFound)?;
```
dgeee13:
```suggestion
            .change_context(errors::ConnectorError::WebhookResourceObjectNotFound)?;
```

crates/hyperswitch_connectors/src/connectors/santander/transformers.rs
```diff
@@ -188,21 +191,32 @@ impl<F, T> TryFrom<ResponseRouterData<F, SantanderAuthUpdateResponse, T, AccessT
...
     type Error = Error;
     fn try_from(
-        item: &SantanderRouterData<&PaymentsAuthorizeRouterData>,
+        value: (
+            &SantanderRouterData<&PaymentsAuthorizeRouterData>,
+            &router_env::env::Env,
+        ),
     ) -> Result<Self, Self::Error> {
-        if item.router_data.request.capture_method != Some(enums::CaptureMethod::Automatic) {
+        if value.0.router_data.request.capture_method != Some(enums::CaptureMethod::Automatic) {
```
dgeee13: can you rename value.0 and value.1 for readability?

crates/hyperswitch_connectors/src/connectors/santander/transformers.rs
```diff
@@ -331,8 +662,103 @@ impl From<SantanderPaymentStatus> for common_enums::AttemptStatus {
     }
 }
 
-#[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
-pub struct SantanderPaymentsResponse {
+impl TryFrom<&str> for BrazilianState {
```
dgeee13: for getting the state codes for Brazil, you don't need this util fn we already have a fn that exactly does the same - you can use it directly 
you can find it here - `to_state_code`
https://github.com/juspay/hyperswitch/blob/main/crates/hyperswitch_connectors/src/utils.rs#L1429 (please note - you have to add Brazil's alpha2 country code in this utility fn, rest everything else is already taken care of)

crates/hyperswitch_connectors/src/connectors/santander/transformers.rs
```diff
@@ -435,25 +867,70 @@ impl<F, T> TryFrom<ResponseRouterData<F, SantanderPaymentsSyncResponse, T, Payme
...
+        match response {
+            SantanderPaymentsSyncResponse::PixQRCode(pix_data) => {
+                let connector_metadata = pix_data.pix.first().map(|pix| {
+                    serde_json::json!({
+                        "end_to_end_id": pix.end_to_end_id.clone().expose()
+                    })
+                });
+                Ok(Self {
+                    status: common_enums::AttemptStatus::from(pix_data.status),
+                    response: Ok(PaymentsResponseData::TransactionResponse {
```
dgeee13, line 880: is it anywhere mentioned in the docs - that the connector can throw 400 for failure? If so, it needs to be handled in all the responses in the transformers file, otherwise the current 
 ErrorHandling should be enough

dgeee13: is it anywhere mentioned in the docs - that the connector can throw 400 for failure? If so, it needs to be handled in all the responses in the transformers file, otherwise the current error handling should be fine 

EDIT  - this concern is for 2xx errors, and not 4xx

bsayak03: for 4xx status codes, the err res struct remains the same

crates/hyperswitch_connectors/src/connectors/santander/transformers.rs
bsayak03: Sure

crates/hyperswitch_connectors/src/connectors/santander/transformers.rs
bsayak03: i can but that would mean unecessary declaration of another variable.

crates/hyperswitch_connectors/src/connectors/santander.rs
bsayak03: Sure

crates/hyperswitch_connectors/src/connectors/santander.rs
bsayak03: Sure but only possible for Authorize because in Cancel and Refund flows, payment_method_type doesn't exist in `PaymentsCancelRouterData` and `RefundsData`

crates/hyperswitch_connectors/src/connectors/santander/transformers.rs
dgeee13: @deepanshu-iiitu For 2xx failures, if the connector doesn’t return any error response or code, should we just mark the status as failure?
We’re seeing the same behavior with Santander

dgeee13:
other than this https://github.com/juspay/hyperswitch/pull/9008#discussion_r2300921951, the rest of the PR LGTM 👍

crates/hyperswitch_connectors/src/connectors/santander/transformers.rs
deepanshu-iiitu: Yes. They should be marked as failure, but why are we not getting error message in this case?

bsayak03: The connector sends the same struct for 2xx success as well as failure. Only difference is `status`. For 4xx and 5xx, it returns a error struct

deepanshu-iiitu: then populate no error code and no error message

bsayak03: it should not go under `Ok` you're suggesting right?

deepanshu-iiitu: it can go under ok, because payment failed with connector

bsayak03: @dgeee13 then we are good to go i guess?

dgeee13: @bsayak03 We can send the overall paymentsResponse as `Ok`, but like @deepanshu-iiitu mentioned we need to add this change:
In the `Ok` response, change the field `response` at line 880 to `ErrorResponse` and populate it with no error code and no error message. We need this change for all the other flows too.

bsayak03: Did that for Authorize and PSync. For Void - since there seems to be only one status "REMOVED_BY_RECEIVING_USER", hence didnot do the failure mapping. For Refund, the status is unclear.

dgeee13 (approved):

deepanshu-iiitu (approved):

crates/api_models/src/payments.rs
```diff
@@ -4260,11 +4260,44 @@ pub struct RewardData {
...
+#[derive(Debug, Clone, PartialEq, Eq, serde::Serialize, serde::Deserialize, ToSchema)]
 pub struct BoletoVoucherData {
-    /// The shopper's social security number
+    /// The shopper's social security number (CPF or CNPJ)
     #[schema(value_type = Option<String>)]
     pub social_security_number: Option<Secret<String>>,
+
+    /// The shopper's bank account number associated with the boleto
+    #[schema(value_type = Option<String>)]
+    pub bank_number: Option<String>,
```
jagan-jaya:
```suggestion
    pub bank_number: Option< Secret <String>>,
```

crates/api_models/src/payments.rs
```diff
@@ -4983,6 +5018,8 @@ pub struct VoucherNextStepData {
     pub download_url: Option<Url>,
     /// Url to payment instruction page
     pub instructions_url: Option<Url>,
+    /// Human-readable numeric version of the barcode.
+    pub digitable_line: Option<String>,
```
jagan-jaya:
```suggestion
    pub digitable_line: Option< Secret <String>>,
```

crates/hyperswitch_domain_models/src/payment_method_data.rs
```diff
@@ -662,6 +662,22 @@ pub enum VoucherData {
 pub struct BoletoVoucherData {
     /// The shopper's social security number
     pub social_security_number: Option<Secret<String>>,
+    /// The bank number associated with the boleto
+    pub bank_number: Option<String>,
```
jagan-jaya:
```suggestion
    pub bank_number: Option<Secret<String>>,
```

deepanshu-iiitu (approved):

PR #8887, opened 2025-08-09
feat(core): add support for api locking with multiple keys for a single api

hrithikesh026:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

added support for api locking with multiple keys for a single api.
* introduced a couple of redis interface commands to support this feature.

### Additional Changes

- [ ] This PR modifies the API contract
- [ ] This PR modifies the database schema
- [ ] This PR modifies application configuration/environment variables

## How did you test it?

Redis commands unit test result:
image

Parallel payments-create script output:
before this change:
image

after this change:
image

<details>
This script will make n number of parallel payment create calls with same customer id.
<summary>Click to expand script</summary>

```shell
#!/bin/bash

# Parallel Payments Testing Script
# Usage: ./parallel_payments.sh <num_requests> <api_key> <profile_id> [endpoint_url]

# Colors for output
RED='\033[0;31m'
GREEN='\033[0;32m'
YELLOW='\033[1;33m'
BLUE='\033[0;34m'
NC='\033[0m' # No Color

# Function to print colored output
print_color() {
    local color=$1
    local message=$2
    echo -e "${color}${message}${NC}"
}

# Function to show usage
show_usage() {
    echo "Usage: $0 <num_requests> <api_key> <profile_id> [endpoint_url]"
    echo ""
    echo "Arguments:"
    echo "  num_requests  - Number of parallel requests to make (positive integer)"
    echo "  api_key       - API key for authentication"
    echo "  profile_id    - Profile ID for the requests"
    echo "  endpoint_url  - Optional endpoint URL (default: http://localhost:8080/payments)"
    echo ""
    echo "Example:"
    echo "  $0 10 dev_6AB2Fe34Y9nyTa2EaPJZZfUCnR2XcpqQr8ymti5WbUgLVmoQ2yPCjfnwbdVsLw14 profile123"
    echo "  $0 50 dev_key profile456 http://localhost:8080/payments"
}

# Validate arguments
if [ $# -lt 3 ] || [ $# -gt 4 ]; then
    print_color $RED "Error: Invalid number of arguments"
    show_usage
    exit 1
fi

NUM_REQUESTS=$1
API_KEY=$2
PROFILE_ID=$3
ENDPOINT_URL=${4:-"http://localhost:8080/payments"}

# Validate num_requests is a positive integer
[...]

PR #9019, opened 2025-08-21
fix(redis): add support for fetching multiple keys for redis cluster

hrithikesh026:
## Type of Change

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

added support for fetching multiple keys for redis cluster.

Earlier, we were using `mget` to fetch multiple keys. While this works fine for standalone redis, it would fail for redis cluster(it won't work if redis keys are stored in multiple hash slots).

So, introduced a new redis interface function to use `mget` for standalone redis and parallel `get` calls for cluster redis.

### Additional Changes

## How did you test it?

test cases same as this PR: https://github.com/juspay/hyperswitch/pull/8887

Tested in an internal environment with redis cluster setup enabled.

## Checklist

### DIFF
--- a/crates/redis_interface/src/commands.rs
+++ b/crates/redis_interface/src/commands.rs
@@ -198,8 +198,7 @@ impl super::RedisConnectionPool {
         }
     }
 
-    #[instrument(level = "DEBUG", skip(self))]
-    pub async fn get_multiple_keys<V>(
+    async fn get_multiple_keys_with_mget<V>(
         &self,
         keys: &[RedisKey],
     ) -> CustomResult<Vec<Option<V>>, errors::RedisError>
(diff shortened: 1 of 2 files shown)


SanchithHegde (approved):

crates/redis_interface/src/commands.rs
```diff
@@ -212,13 +212,69 @@ impl super::RedisConnectionPool {
...
+    where
+        V: FromRedis + Unpin + Send + 'static,
+    {
+        if keys.is_empty() {
+            return Ok(Vec::new());
+        }
+        let tenant_aware_keys: Vec<String> =
+            keys.iter().map(|key| key.tenant_aware_key(self)).collect();
+
+        let futures = tenant_aware_keys
```
akshay-97, line 235: are we expecting many keys in input, we could check fred pipeline so keys belonging to same shard can be pipelined

hrithikesh026: For now, We are fetching only 2 keys. Will explore fred pipeline option.

PR #9024, opened 2025-08-22
Celero cit mit

Nithin1506200:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

CIT MIT implementation for CELERO 

### Additional Changes

## Checklist

### DIFF
--- a/crates/hyperswitch_connectors/src/connectors/celero/transformers.rs
+++ b/crates/hyperswitch_connectors/src/connectors/celero/transformers.rs
@@ -87,6 +87,17 @@ pub struct CeleroPaymentsRequest {
     shipping_address: Option<CeleroAddress>,
     #[serde(skip_serializing_if = "Option::is_none")]
     create_vault_record: Option<bool>,
+    // CIT/MIT fields
+    #[serde(skip_serializing_if = "Option::is_none")]
+    card_on_file_indicator: Option<CardOnFileIndicator>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    initiated_by: Option<InitiatedBy>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    initial_transaction_id: Option<String>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    stored_credential_indicator: Option<StoredCredentialIndicator>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    billing_method: Option<BillingMethod>,
 }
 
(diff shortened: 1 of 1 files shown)


crates/hyperswitch_connectors/src/connectors/celero/transformers.rs
```diff
@@ -248,12 +268,68 @@ impl TryFrom<&CeleroRouterData<&PaymentsAuthorizeRouterData>> for CeleroPayments
...
+) {
+    // Default values
+    let mut card_on_file_indicator = None;
+    let mut initiated_by = None;
+    let mut initial_transaction_id = None;
+    let mut stored_credential_indicator = None;
+    let mut billing_method = None;
+
+    // Check if this is a mandate payment
+    if router_data.request.is_mandate_payment() {
```
djeada, line 300: neat: Long if/else chain is hard to read and may miss edge cases as the logic grows.

match statement might be more readable:

```
match (
    router_data.request.is_mandate_payment(),
    router_data.request.setup_mandate_details.is_some(),
    router_data.request.off_session.unwrap_or(false),
) {
    (true, _, _) => { /* ... */ }
    (false, true, _) => { /* ... */ }
    (false, false, true) => { /* ... */ }
    _ => { /* ... */ }
}
```

crates/hyperswitch_connectors/src/connectors/nuvei.rs
```diff
@@ -57,8 +58,17 @@ use crate::{
...
 };
 
-#[derive(Debug, Clone)]
-pub struct Nuvei;
+#[derive(Clone)]
+pub struct Nuvei {
+    pub amount_convertor: &'static (dyn AmountConvertor<Output = StringMajorUnit> + Sync),
+}
+impl Nuvei {
+    pub fn new() -> &'static Self {
```
djeada, line 66: The struct is constructed with a static method returning a reference to a new value every call—this is unidiomatic and can be problematic for thread safety. Would use `LazyLock` if singleton is preferred or just normal method otherwise.

crates/hyperswitch_connectors/src/connectors/celero/transformers.rs
Nithin1506200: Looks good djeada. This is just a draft pr and will follow this pattern matching in future.

PR #9104, opened 2025-08-29
feat(security):Add XSS validation for dashboard metadata fields

kanikabansal-juspay:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

Added comprehensive XSS (Cross-Site Scripting) protection to dashboard metadata fields. Implemented a new `SafeString` type wrapper that validates input against XSS attacks during JSON deserialization.

**Dependencies Added**
- `ammonia = "3.3"` - HTML sanitization library

### Additional Changes

## Motivation and Context

- XSS vulnerability in dashboard metadata fields (`legal_business_name`, `description`, etc.)
- Malicious JavaScript payloads were being persistently stored and executed

### XSS Test Cases Validated

**All malicious payloads return 400 Bad Request with validation error:**
```
{
    "error": {
        "error_type": "invalid_request",
        "message": "Json deserialize error: Input contains potentially malicious content at line 5 column 62",
        "code": "IR_06"
    }
}

```

#### 1. Basic Script Tags
```json
{
    "ProdIntent": {
        "legal_business_name": "<script>alert('xss')</script>"
    }
}
```

#### 2. Event Handlers
```json
{
    "ProdIntent": {
        "comments": "onload=alert('xss') onclick=alert('xss') onmouseover=alert('xss')"
    }
}
```

#### 3. JavaScript URLs
```json
{
    "ProdIntent": {
        "business_website": "javascript:alert('xss')"
    }
}
```

#### 4. Data URLs
```json
{
    "ProdIntent": {
        "business_website": "data:text/html,<script>alert('xss')</script>"
    }
}
```

#### 5. Dangerous Tags (iframe, svg, object, embed)
```json
{
    "ProdIntent": {
        "legal_business_name": "<iframe src=javascript:alert('xss')></iframe>"
    }
}
```

#### 6. SVG with Script
```json
{
    "ProdIntent": {
        "legal_business_name": "<svg><script>alert('xss')</script></svg>"
    }
}
```

#### 7. Object/Embed Tags
```json
{
    "ProdIntent": {
        "legal_business_name": "<object data=javascript:alert('xss')></object>"
    }
}
```

#### 8. Mixed Case and Encoded Payloads
```json
{
    "ProdIntent": {
        "legal_business_name": "<ScRiPt>alert('xss')</ScRiPt>"
    }
}
```

#### 9. Multiple Fields with XSS
```json
{
    "ProdIntent": {
        "legal_business_name": "<script>alert('xss1')</script>",
        "business_website": "javascript:alert('xss2')",
        "comments": "onload=alert('xss3')"
    }
}
```

#### 10. Edge Cases
```json
{
    "ProdIntent": {
        "legal_business_name": "Test<script>alert('xss')</script>Company"
    }
}
```

#### 11. Math/Embedded Objects
```json
{
    "ProdIntent": {
        "legal_business_name": "<math><script>alert('xss')</script></math>"
    }
}
```

#### 12. Percent Encoding
```json
{
    "ProdIntent": {
        "legal_business_name": "javascript%3Aalert('xss')"
    }
}
```

#### 13. Legitimate Business Data
```json
{
    "ProdIntent": {
        "poc_email": "kanika.bansal+1234@juspay.in",
        "is_completed": true,
        "legal_business_name": "My Test Company Ltd.",
        "business_location": "IN",
        "business_website": "https://example.com",
        "poc_name": "Kanika Bansal",
        "business_country_name": "India",
        "product_type": "orchestration",
        "comments": "This is a legitimate business comment with no malicious content."
    }
}
```
**200 OK**

## Checklist

### DIFF
--- a/crates/api_models/src/user/dashboard_metadata.rs
+++ b/crates/api_models/src/user/dashboard_metadata.rs
@@ -1,4 +1,5 @@
 use common_enums::{CountryAlpha2, MerchantProductType};
+use common_types::primitive_wrappers::SafeString;
 use common_utils::{id_type, pii};
 use masking::Secret;
 use strum::EnumString;
@@ -50,16 +51,16 @@ pub struct ProcessorConnected {
 
 #[derive(Debug, serde::Deserialize, serde::Serialize)]
 pub struct OnboardingSurvey {
-    pub designation: Option<String>,
-    pub about_business: Option<String>,
-    pub business_website: Option<String>,
-    pub hyperswitch_req: Option<String>,
+    pub designation: Option<SafeString>,
+    pub about_business: Option<SafeString>,
+    pub business_website: Option<SafeString>,
+    pub hyperswitch_req: Option<SafeString>,
     pub major_markets: Option<Vec<String>>,
-    pub business_size: Option<String>,
+    pub business_size: Option<SafeString>,
     pub required_features: Option<Vec<String>>,
     pub required_processors: Option<Vec<String>>,
-    pub planned_live_date: Option<String>,
-    pub miscellaneous: Option<String>,
+    pub planned_live_date: Option<SafeString>,
+    pub miscellaneous: Option<SafeString>,
 }
 
 #[derive(Debug, serde::Deserialize, serde::Serialize)]
@@ -85,27 +86,27 @@ pub enum ConfigurationType {
 #[derive(Debug, serde::Deserialize, serde::Serialize, Clone)]
 pub struct Feedback {
     pub email: pii::Email,
-    pub description: Option<String>,
+    pub description: Option<SafeString>,
     pub rating: Option<i32>,
-    pub category: Option<String>,
+    pub category: Option<SafeString>,
 }
 #[derive(Debug, serde::Deserialize, serde::Serialize, Clone)]
 pub struct ProdIntent {
-    pub legal_business_name: Option<String>,
-    pub business_label: Option<String>,
+    pub legal_business_name: Option<SafeString>,
+    pub business_label: Option<SafeString>,
     pub business_location: Option<CountryAlpha2>,
-    pub display_name: Option<String>,
+    pub display_name: Option<SafeString>,
     pub poc_email: Option<Secret<String>>,
-    pub business_type: Option<String>,
-    pub business_identifier: Option<String>,
-    pub business_website: Option<String>,
+    pub business_type: Option<SafeString>,
+    pub business_identifier: Option<SafeString>,
+    pub business_website: Option<SafeString>,
     pub poc_name: Option<Secret<String>>,
     pub poc_contact: Option<Secret<String>>,
-    pub comments: Option<String>,
+    pub comments: Option<SafeString>,
     pub is_completed: bool,
     #[serde(default)]
     pub product_type: MerchantProductType,
-    pub business_country_name: Option<String>,
+    pub business_country_name: Option<SafeString>,
 }
 
 #[derive(Debug, serde::Deserialize, serde::Serialize, Clone)]
--- a/crates/common_types/src/primitive_wrappers.rs
+++ b/crates/common_types/src/primitive_wrappers.rs
@@ -1,4 +1,5 @@
 pub use bool_wrappers::*;
+pub use safe_string::*;
 pub use u32_wrappers::*;
 mod bool_wrappers {
     use std::ops::Deref;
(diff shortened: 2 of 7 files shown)


crates/common_types/src/primitive_wrappers.rs
```diff
@@ -241,3 +242,113 @@ mod u32_wrappers {
...
+
+    impl SafeString {
+        /// Creates a new SafeString after XSS and SQL injection validation
+        pub fn new(value: String) -> Result<Self, String> {
+            if contains_potential_xss_or_sqli(&value) {
+                return Err("Input contains potentially malicious content".into());
+            }
+
+            // Basic length validation
+            const MAX_LENGTH: usize = 1000;
```
knutties: Would this be limiting in any case at all ?  People should ensure that SafeString is not used for fields like some description which might contain more than 1000 chars.

crates/common_utils/src/validation.rs
```diff
@@ -82,6 +83,27 @@ pub fn validate_domain_against_allowed_domains(
     })
 }
 
+/// checks whether the input string contains potential XSS or SQL injection attack vectors
+pub fn contains_potential_xss_or_sqli(input: &str) -> bool {
+    let decoded_input = urlencoding::decode(input).unwrap_or_else(|_| input.into());
```
knutties: Why do we need a `urlencoding::decode` here?  Isn't data url-decoded by this point in time?

crates/common_types/src/primitive_wrappers.rs
kanikabansal08: I added length validation for DoS protection, but it might be overkill since we already handle XSS/SQLi. It could limit fields like 'about business' and make this less reusable.

crates/common_utils/src/validation.rs
kanikabansal08: JSON string values aren't URL-decoded by Actix-web only query/path/form params are. 
Attackers can bypass XSS/SQLi detection by encoding malicious payloads in JSON like
`{"name": "%3Cscript%3Ealert()%3C/script%3E"}` 
Without decoding first, our validation would miss the actual `<script>` tag.

knutties: got it - but would these values be url-decoded by our application and echoed?  While I understand why we are doing it - I am wondering why we might in first place url-decode this JSON and add to HTML - what do we do if the application twice url-decodes and then such code will slip in right.

kanikabansal08: The updated implementation catches single, double, and triple-encoded attacks while avoiding false positives like 50% or valid %25 in URLs.

crates/api_models/src/user/dashboard_metadata.rs
```diff
@@ -85,27 +86,27 @@ pub enum ConfigurationType {
...
-    pub display_name: Option<String>,
-    pub poc_email: Option<Secret<String>>,
-    pub business_type: Option<String>,
-    pub business_identifier: Option<String>,
-    pub business_website: Option<String>,
-    pub poc_name: Option<Secret<String>>,
-    pub poc_contact: Option<Secret<String>>,
-    pub comments: Option<String>,
+    pub display_name: Option<SafeString>,
+    pub poc_email: Option<Secret<SafeString>>,
```
ThisIsMani: Can we use `pii::Email` here?

crates/common_utils/src/validation.rs
```diff
@@ -82,6 +83,42 @@ pub fn validate_domain_against_allowed_domains(
     })
 }
 
+/// checks whether the input string contains potential XSS or SQL injection attack vectors
+///
+/// # Panics
```
ThisIsMani: I don't think this panics.

crates/common_utils/src/validation.rs
```diff
@@ -82,6 +83,42 @@ pub fn validate_domain_against_allowed_domains(
...
+///
+/// This function may panic if the regex patterns fail to compile, though this should not happen
+/// with the hardcoded patterns used.
+pub fn contains_potential_xss_or_sqli(input: &str) -> bool {
```
ThisIsMani, line 88: Did we check if there is any crate that already does this?

kanikabansal08: There’s a crate called libinjectionrs, but it’s not well maintained. We already use ammonia for XSS, which parses and removes unwanted HTML tags but it only works on actual HTML. For raw strings in JSON (like onload=alert(1)), we still need some regex checks to catch those cases.

crates/common_utils/src/validation.rs
```diff
@@ -82,6 +83,37 @@ pub fn validate_domain_against_allowed_domains(
...
+        .tags(HashSet::new())
+        .clean(&decoded)
+        .to_string();
+    if cleaned != decoded {
+        return true;
+    }
+    static SUSPICIOUS: LazyLock<Result<Regex, regex::Error>> = LazyLock::new(|| {
+        Regex::new(
+            r"(?is)\bon[a-z]+\s*=|\bjavascript\s*:|\bdata\s*:\s*text/html|<\s*(script|iframe|svg|math|object|embed)\b|</\s*script\s*>|(?:')\s*or\s*'?\d+'?=?\d*|union\s+select|insert\s+into|drop\s+table|information_schema|sleep\s*\(|--|;",
+        )
```
ThisIsMani, line 130: Can we split it into multiple sets... One for HTML, one for SQL, etc...

crates/common_utils/src/validation.rs
```diff
@@ -82,6 +83,37 @@ pub fn validate_domain_against_allowed_domains(
...
+    let cleaned = AmmoniaBuilder::default()
+        .tags(HashSet::new())
+        .clean(&decoded)
+        .to_string();
+    if cleaned != decoded {
+        return true;
+    }
+    static SUSPICIOUS: LazyLock<Result<Regex, regex::Error>> = LazyLock::new(|| {
+        Regex::new(
+            r"(?is)\bon[a-z]+\s*=|\bjavascript\s*:|\bdata\s*:\s*text/html|<\s*(script|iframe|svg|math|object|embed)\b|</\s*script\s*>|(?:')\s*or\s*'?\d+'?=?\d*|union\s+select|insert\s+into|drop\s+table|information_schema|sleep\s*\(|--|;",
```
ThisIsMani: Aren't `script|iframe|svg|match...` are part of html tags? Doesn't ammonia remove them?

ThisIsMani: We are not not blocking any javascript specific stuff, is it? I think we should be able to block function calls and stuff...

kanikabansal08: Removed redundant regex check

kanikabansal08: With the updated implementation, Ammonia strips all HTML tags, and the regex now detects dangerous attributes, JS protocols, and function calls.

crates/common_utils/src/validation.rs
```diff
@@ -82,6 +83,46 @@ pub fn validate_domain_against_allowed_domains(
...
+        )
+    });
+
+    static SQLI: LazyLock<Result<Regex, regex::Error>> = LazyLock::new(|| {
+        Regex::new(
+            r"(?is)(?:')\s*or\s*'?\d+'?=?\d*|union\s+select|insert\s+into|drop\s+table|information_schema|sleep\s*\(|--|;",
+        )
+    });
+
+    XSS.as_ref().is_ok_and(|re| re.is_match(&decoded))
```
ThisIsMani: Can we log if the Regex is wrong.

ThisIsMani (approved):

crates/common_utils/src/validation.rs
```diff
@@ -82,6 +83,46 @@ pub fn validate_domain_against_allowed_domains(
...
+    }
+
+    let cleaned = AmmoniaBuilder::default()
+        .tags(HashSet::new())
+        .clean(&decoded)
+        .to_string();
+
+    if cleaned != decoded {
+        return true;
+    }
```
SanchithHegde: If we only need to figure out whether the input contains any HTML (or HTML-like data) and are not going to make use of the sanitized input, can't we use [`ammonia::is_html()`](https://docs.rs/ammonia/4.1.1/ammonia/fn.is_html.html)? Is there any case where that would fail?

crates/common_utils/src/validation.rs
```diff
@@ -82,6 +83,46 @@ pub fn validate_domain_against_allowed_domains(
...
+        Regex::new(
+            r"(?is)\bon[a-z]+\s*=|\bjavascript\s*:|\bdata\s*:\s*text/html|\b(alert|prompt|confirm|eval)\s*\(",
+        )
+    });
+
+    static SQLI: LazyLock<Result<Regex, regex::Error>> = LazyLock::new(|| {
+        Regex::new(
+            r"(?is)(?:')\s*or\s*'?\d+'?=?\d*|union\s+select|insert\s+into|drop\s+table|information_schema|sleep\s*\(|--|;",
+        )
+    });
```
SanchithHegde: Do we have any source for these regexes that we can include as links here?

crates/common_utils/src/validation.rs
```diff
@@ -2,6 +2,7 @@
 
 use std::{collections::HashSet, sync::LazyLock};
 
+use ammonia::Builder as AmmoniaBuilder;
```
SanchithHegde: Nit: If there's going to be only one usage, I think you can directly qualify it instead of importing it.

crates/common_utils/src/validation.rs
```diff
@@ -82,6 +83,46 @@ pub fn validate_domain_against_allowed_domains(
...
+    static PERCENT_ENCODED: LazyLock<Result<Regex, regex::Error>> =
+        LazyLock::new(|| Regex::new(r"%[0-9A-Fa-f]{2}"));
+
+    if decoded.contains('%')
+        && PERCENT_ENCODED
+            .as_ref()
+            .is_ok_and(|re| re.is_match(&decoded))
+    {
+        return true;
+    }
```
SanchithHegde, line 110: We are not allowing percent-encoded input, but are allowing URL-encoded input, because we just URL-decode at the beginning, but don't check whether the input is possibly URL-encoded. Is this the intended behavior?

kanikabansal08: We allow a single decode to avoid false positives from valid encoded inputs (like URLs or filters not necessarily for this flow but possible elsewhere). If we still see encoding after that, it’s likely double/triple encoding, which isn’t normal user behavior, so we flag it.

SanchithHegde (approved):

crates/common_utils/src/validation.rs
kanikabansal08: These regexes were built iteratively from our XSS/SQLi test cases There’s no single perfect regex, and even WAFs with CRS use multiple patterns to cover different attack types.

crates/common_utils/src/validation.rs
SanchithHegde: As of today, clippy will also error out in case of invalid regexes, because the [`clippy::invalid_regex`](https://rust-lang.github.io/rust-clippy/stable/index.html#invalid_regex) lint is deny by default (as of current stable version). So we should be able to catch invalid regexes in CI itself.

PR #9123, opened 2025-08-31
feat(subscriptions): add route for creating subscription intent

prajjwalkumar17:
## Summary

Add **Subscriptions API (v1/OLAP)** entrypoint to create a _subscription intent_ and optionally create/attach a customer, returning a `client_secret` and basic subscription payload.

- New route: `POST /subscription/create` (behind `olap` + `v1`)
- Creates a `subscription` row (using existing table) and generates a `client_secret`
- Optionally **gets or creates** a `Customer` from inline `customer` payload or `customer_id`
- Persists optional `mca_id` on the subscription row (used later for connector selection)
- Returns a typed `CreateSubscriptionResponse` with `subscription`, `client_secret`, `merchant_id`, `mca_id`, and optional `customer`, `invoice` (placeholder)

### Create Subscription Intent

`POST /subscription/create`

**Auth**
- API key (HeaderAuth) + JWT (`Permission::ProfileRoutingWrite`)

**Request Body**
```json
{
    "customer_id": "{{customer_id}}"
}
```

**Behavior**
- Generates `sub_<random>` for subscription row id
- If `customer_id` provided → fetch; else if `customer` has any PII fields → **create customer** (encrypting PII); else → `400 CustomerNotFound`
- Persists subscription with:
  - `id`
  - `subscription_id` = `null` (placeholder, to be set later)
  - `mca_id` (if provided)
  - `client_secret` = generated via `SubscriptionNew::generate_client_secret()`
  - `merchant_id`, `customer_id`
  - `billing_processor`, `payment_method_id`, `metadata` = `null` (placeholders)
- Returns `200` with response below

**Success Response**
```json
{
    "id": "subscription_f8Ng47rKaB296XHolKKG",
    "merchant_reference_id": null,
    "status": "Created",
    "plan_id": null,
    "profile_id": "pro_UjmcM35TVpYh9Zb3LaZV",
    "client_secret": "subscription_f8Ng47rKaB296XHolKKG_secret_0fz5A3aa4NTcdiQF5W7i",
    "merchant_id": "merchant_1758146582",
    "coupon_code": null,
    "customer_id": "cus_jtJ8WHTHnsguskpiorEt"
}
```

---

## Implementation Notes

- **Core**
  - `crates/router/src/core/subscription.rs` – service handler
  - `crates/router/src/core/subscription/utils.rs` – helpers, request/response types
- **Routing**
  - `crates/router/src/routes/subscription.rs` – web handler (`/subscription/create`)
  - `crates/router/src/routes/app.rs` – `Subscription::server()` scope under `/subscription`
  - `crates/router/src/lib.rs` – wire route
- **Domain / DTO**
  - `hyperswitch_domain_models::router_request_types::CustomerDetails` now `Serialize`/`Deserialize`
  - `hyperswitch_domain_models::customer::Customer` derives `Serialize` (v1)
- **Storage**
  - Uses existing `subscription` table and `SubscriptionNew::generate_client_secret()`
  - No new migrations introduced in this PR
- **Customer flow**
  - If not found, creates via encrypted batch operation; respects merchant key store and storage scheme

---

## Testing
```
curl --location 'http://localhost:8080/subscription/create' \
--header 'X-Profile-Id: pro_UjmcM35TVpYh9Zb3LaZV' \
--header 'api-key: dev_pPBMnOEL5SJnMfxxp2bh4du87ESZoMJNtUmn3oIS8kpCNfUJAZ2a8F832n1je5zQ' \
--data '{
    "customer_id": "cus_jtJ8WHTHnsguskpiorEt"
}'
```

Response:
```
{
    "id": "subscription_f8Ng47rKaB296XHolKKG",
    "merchant_reference_id": null,
    "status": "Created",
    "plan_id": null,
    "profile_id": "pro_UjmcM35TVpYh9Zb3LaZV",
    "client_secret": "subscription_f8Ng47rKaB296XHolKKG_secret_0fz5A3aa4NTcdiQF5W7i",
    "merchant_id": "merchant_1758146582",
    "coupon_code": null,
    "customer_id": "cus_jtJ8WHTHnsguskpiorEt"
}
```
---

### DIFF
--- a/crates/diesel_models/src/schema.rs
+++ b/crates/diesel_models/src/schema.rs
@@ -190,7 +190,9 @@ diesel::table! {
     use diesel::sql_types::*;
     use crate::enums::diesel_exports::*;
 
-    business_profile (id) {
+    business_profile (profile_id) {
+        #[max_length = 64]
+        profile_id -> Varchar,
         #[max_length = 64]
         merchant_id -> Varchar,
         #[max_length = 64]
@@ -204,6 +206,10 @@ diesel::table! {
         redirect_to_merchant_with_http_post -> Bool,
         webhook_details -> Nullable<Json>,
         metadata -> Nullable<Json>,
+        routing_algorithm -> Nullable<Json>,
+        intent_fulfillment_time -> Nullable<Int8>,
+        frm_routing_algorithm -> Nullable<Jsonb>,
+        payout_routing_algorithm -> Nullable<Jsonb>,
         is_recon_enabled -> Bool,
         applepay_verified_domains -> Nullable<Array<Nullable<Text>>>,
         payment_link_config -> Nullable<Jsonb>,
@@ -237,8 +243,9 @@ diesel::table! {
         is_debit_routing_enabled -> Bool,
         merchant_business_country -> Nullable<CountryAlpha2>,
         #[max_length = 64]
-        id -> Varchar,
+        id -> Nullable<Varchar>,
         is_iframe_redirection_enabled -> Nullable<Bool>,
+        is_pre_network_tokenization_enabled -> Nullable<Bool>,
         three_ds_decision_rule_algorithm -> Nullable<Jsonb>,
         acquirer_config_map -> Nullable<Jsonb>,
         #[max_length = 16]
@@ -299,6 +306,8 @@ diesel::table! {
         capture_sequence -> Int2,
         #[max_length = 128]
         connector_response_reference_id -> Nullable<Varchar>,
+        #[max_length = 512]
+        connector_capture_data -> Nullable<Varchar>,
         processor_capture_data -> Nullable<Text>,
     }
 }
@@ -342,7 +351,9 @@ diesel::table! {
     use diesel::sql_types::*;
     use crate::enums::diesel_exports::*;
 
-    customers (id) {
+    customers (customer_id, merchant_id) {
+        #[max_length = 64]
+        customer_id -> Varchar,
         #[max_length = 64]
         merchant_id -> Varchar,
         name -> Nullable<Bytea>,
@@ -357,18 +368,13 @@ diesel::table! {
         connector_customer -> Nullable<Jsonb>,
         modified_at -> Timestamp,
         #[max_length = 64]
+        address_id -> Nullable<Varchar>,
+        #[max_length = 64]
         default_payment_method_id -> Nullable<Varchar>,
         #[max_length = 64]
         updated_by -> Nullable<Varchar>,
         version -> ApiVersion,
         tax_registration_id -> Nullable<Bytea>,
-        #[max_length = 64]
-        merchant_reference_id -> Nullable<Varchar>,
-        default_billing_address -> Nullable<Bytea>,
-        default_shipping_address -> Nullable<Bytea>,
-        status -> Nullable<DeleteStatus>,
-        #[max_length = 64]
-        id -> Varchar,
     }
 }
 
@@ -794,22 +800,46 @@ diesel::table! {
     use diesel::sql_types::*;
     use crate::enums::diesel_exports::*;
 
-    merchant_account (id) {
+    merchant_account (merchant_id) {
+        #[max_length = 64]
+        merchant_id -> Varchar,
+        #[max_length = 255]
+        return_url -> Nullable<Varchar>,
+        enable_payment_response_hash -> Bool,
+        #[max_length = 255]
+        payment_response_hash_key -> Nullable<Varchar>,
+        redirect_to_merchant_with_http_post -> Bool,
         merchant_name -> Nullable<Bytea>,
         merchant_details -> Nullable<Bytea>,
+        webhook_details -> Nullable<Json>,
+        sub_merchants_enabled -> Nullable<Bool>,
+        #[max_length = 64]
+        parent_merchant_id -> Nullable<Varchar>,
         #[max_length = 128]
         publishable_key -> Nullable<Varchar>,
         storage_scheme -> MerchantStorageScheme,
+        #[max_length = 64]
+        locker_id -> Nullable<Varchar>,
         metadata -> Nullable<Jsonb>,
+        routing_algorithm -> Nullable<Json>,
+        primary_business_details -> Json,
+        intent_fulfillment_time -> Nullable<Int8>,
         created_at -> Timestamp,
         modified_at -> Timestamp,
+        frm_routing_algorithm -> Nullable<Jsonb>,
+        payout_routing_algorithm -> Nullable<Jsonb>,
         #[max_length = 32]
         organization_id -> Varchar,
+        is_recon_enabled -> Bool,
+        #[max_length = 64]
+        default_profile -> Nullable<Varchar>,
         recon_status -> ReconStatus,
+        payment_link_config -> Nullable<Jsonb>,
+        pm_collect_link_config -> Nullable<Jsonb>,
         version -> ApiVersion,
         is_platform_account -> Bool,
         #[max_length = 64]
-        id -> Varchar,
+        id -> Nullable<Varchar>,
         #[max_length = 64]
         product_type -> Nullable<Varchar>,
         #[max_length = 64]
@@ -821,18 +851,27 @@ diesel::table! {
     use diesel::sql_types::*;
     use crate::enums::diesel_exports::*;
 
-    merchant_connector_account (id) {
+    merchant_connector_account (merchant_connector_id) {
         #[max_length = 64]
         merchant_id -> Varchar,
         #[max_length = 64]
         connector_name -> Varchar,
         connector_account_details -> Bytea,
+        test_mode -> Nullable<Bool>,
         disabled -> Nullable<Bool>,
+        #[max_length = 128]
+        merchant_connector_id -> Varchar,
         payment_methods_enabled -> Nullable<Array<Nullable<Json>>>,
         connector_type -> ConnectorType,
         metadata -> Nullable<Jsonb>,
         #[max_length = 255]
         connector_label -> Nullable<Varchar>,
+        business_country -> Nullable<CountryAlpha2>,
+        #[max_length = 255]
+        business_label -> Nullable<Varchar>,
+        #[max_length = 64]
+        business_sub_label -> Nullable<Varchar>,
+        frm_configs -> Nullable<Jsonb>,
         created_at -> Timestamp,
         modified_at -> Timestamp,
         connector_webhook_details -> Nullable<Jsonb>,
@@ -846,8 +885,7 @@ diesel::table! {
         connector_wallets_details -> Nullable<Bytea>,
         version -> ApiVersion,
         #[max_length = 64]
-        id -> Varchar,
-        feature_metadata -> Nullable<Jsonb>,
+        id -> Nullable<Varchar>,
     }
 }
 
@@ -867,13 +905,16 @@ diesel::table! {
     use diesel::sql_types::*;
     use crate::enums::diesel_exports::*;
 
-    organization (id) {
+    organization (org_id) {
+        #[max_length = 32]
+        org_id -> Varchar,
+        org_name -> Nullable<Text>,
         organization_details -> Nullable<Jsonb>,
         metadata -> Nullable<Jsonb>,
         created_at -> Timestamp,
         modified_at -> Timestamp,
         #[max_length = 32]
-        id -> Varchar,
+        id -> Nullable<Varchar>,
         organization_name -> Nullable<Text>,
         version -> ApiVersion,
         #[max_length = 64]
@@ -887,18 +928,31 @@ diesel::table! {
     use diesel::sql_types::*;
     use crate::enums::diesel_exports::*;
 
-    payment_attempt (id) {
+    payment_attempt (attempt_id, merchant_id) {
         #[max_length = 64]
         payment_id -> Varchar,
         #[max_length = 64]
         merchant_id -> Varchar,
+        #[max_length = 64]
+        attempt_id -> Varchar,
         status -> AttemptStatus,
+        amount -> Int8,
+        currency -> Nullable<Currency>,
+        save_to_locker -> Nullable<Bool>,
         #[max_length = 64]
         connector -> Nullable<Varchar>,
         error_message -> Nullable<Text>,
+        offer_amount -> Nullable<Int8>,
         surcharge_amount -> Nullable<Int8>,
+        tax_amount -> Nullable<Int8>,
         #[max_length = 64]
         payment_method_id -> Nullable<Varchar>,
+        payment_method -> Nullable<Varchar>,
+        #[max_length = 128]
+        connector_transaction_id -> Nullable<Varchar>,
+        capture_method -> Nullable<CaptureMethod>,
+        capture_on -> Nullable<Timestamp>,
+        confirm -> Bool,
         authentication_type -> Nullable<AuthenticationType>,
         created_at -> Timestamp,
         modified_at -> Timestamp,
@@ -906,6 +960,8 @@ diesel::table! {
         #[max_length = 255]
         cancellation_reason -> Nullable<Varchar>,
         amount_to_capture -> Nullable<Int8>,
+        #[max_length = 64]
+        mandate_id -> Nullable<Varchar>,
         browser_info -> Nullable<Jsonb>,
         #[max_length = 255]
         error_code -> Nullable<Varchar>,
@@ -914,8 +970,14 @@ diesel::table! {
         connector_metadata -> Nullable<Jsonb>,
         #[max_length = 50]
         payment_experience -> Nullable<Varchar>,
+        #[max_length = 64]
+        payment_method_type -> Nullable<Varchar>,
         payment_method_data -> Nullable<Jsonb>,
+        #[max_length = 64]
+        business_sub_label -> Nullable<Varchar>,
+        straight_through_algorithm -> Nullable<Jsonb>,
         preprocessing_step_id -> Nullable<Varchar>,
+        mandate_details -> Nullable<Jsonb>,
         error_reason -> Nullable<Text>,
         multiple_capture_count -> Nullable<Int2>,
         #[max_length = 128]
@@ -925,6 +987,7 @@ diesel::table! {
         updated_by -> Varchar,
         #[max_length = 32]
         merchant_connector_id -> Nullable<Varchar>,
+        authentication_data -> Nullable<Json>,
         encoded_data -> Nullable<Text>,
         #[max_length = 255]
         unified_code -> Nullable<Varchar>,
@@ -936,8 +999,13 @@ diesel::table! {
         authentication_connector -> Nullable<Varchar>,
         #[max_length = 64]
         authentication_id -> Nullable<Varchar>,
+        mandate_data -> Nullable<Jsonb>,
         #[max_length = 64]
         fingerprint_id -> Nullable<Varchar>,
+        #[max_length = 64]
+        payment_method_billing_address_id -> Nullable<Varchar>,
+        #[max_length = 64]
+        charge_id -> Nullable<Varchar>,
         #[max_length = 64]
         client_source -> Nullable<Varchar>,
         #[max_length = 64]
@@ -951,15 +1019,24 @@ diesel::table! {
         card_network -> Nullable<Varchar>,
         shipping_cost -> Nullable<Int8>,
         order_tax_amount -> Nullable<Int8>,
+        #[max_length = 512]
+        connector_transaction_data -> Nullable<Varchar>,
+        connector_mandate_detail -> Nullable<Jsonb>,
         request_extended_authorization -> Nullable<Bool>,
         extended_authorization_applied -> Nullable<Bool>,
         capture_before -> Nullable<Timestamp>,
+        processor_transaction_data -> Nullable<Text>,
         card_discovery -> Nullable<CardDiscovery>,
         charges -> Nullable<Jsonb>,
+        #[max_length = 64]
+        issuer_error_code -> Nullable<Varchar>,
+        issuer_error_message -> Nullable<Text>,
         #[max_length = 64]
         processor_merchant_id -> Nullable<Varchar>,
         #[max_length = 255]
         created_by -> Nullable<Varchar>,
+        setup_future_usage_applied -> Nullable<FutureUsage>,
+        routing_approach -> Nullable<RoutingApproach>,
         #[max_length = 255]
         connector_request_reference_id -> Nullable<Varchar>,
         #[max_length = 255]
@@ -973,7 +1050,9 @@ diesel::table! {
     use diesel::sql_types::*;
     use crate::enums::diesel_exports::*;
 
-    payment_intent (id) {
+    payment_intent (payment_id, merchant_id) {
+        #[max_length = 64]
+        payment_id -> Varchar,
         #[max_length = 64]
         merchant_id -> Varchar,
         status -> IntentStatus,
@@ -987,12 +1066,28 @@ diesel::table! {
         #[max_length = 255]
         return_url -> Nullable<Varchar>,
         metadata -> Nullable<Jsonb>,
+        #[max_length = 64]
+        connector_id -> Nullable<Varchar>,
+        #[max_length = 64]
+        shipping_address_id -> Nullable<Varchar>,
+        #[max_length = 64]
+        billing_address_id -> Nullable<Varchar>,
+        #[max_length = 255]
+        statement_descriptor_name -> Nullable<Varchar>,
+        #[max_length = 255]
+        statement_descriptor_suffix -> Nullable<Varchar>,
         created_at -> Timestamp,
         modified_at -> Timestamp,
         last_synced -> Nullable<Timestamp>,
         setup_future_usage -> Nullable<FutureUsage>,
+        off_session -> Nullable<Bool>,
+        #[max_length = 128]
+        client_secret -> Nullable<Varchar>,
         #[max_length = 64]
-        active_attempt_id -> Nullable<Varchar>,
+        active_attempt_id -> Varchar,
+        business_country -> Nullable<CountryAlpha2>,
+        #[max_length = 64]
+        business_label -> Nullable<Varchar>,
         order_details -> Nullable<Array<Nullable<Jsonb>>>,
         allowed_payment_method_types -> Nullable<Json>,
         connector_metadata -> Nullable<Json>,
@@ -1000,17 +1095,29 @@ diesel::table! {
         attempt_count -> Int2,
         #[max_length = 64]
         profile_id -> Nullable<Varchar>,
+        #[max_length = 64]
+        merchant_decision -> Nullable<Varchar>,
         #[max_length = 255]
         payment_link_id -> Nullable<Varchar>,
+        payment_confirm_source -> Nullable<PaymentSource>,
         #[max_length = 32]
         updated_by -> Varchar,
         surcharge_applicable -> Nullable<Bool>,
         request_incremental_authorization -> Nullable<RequestIncrementalAuthorization>,
+        incremental_authorization_allowed -> Nullable<Bool>,
         authorization_count -> Nullable<Int4>,
         session_expiry -> Nullable<Timestamp>,
+        #[max_length = 64]
+        fingerprint_id -> Nullable<Varchar>,
         request_external_three_ds_authentication -> Nullable<Bool>,
+        charges -> Nullable<Jsonb>,
         frm_metadata -> Nullable<Jsonb>,
         customer_details -> Nullable<Bytea>,
+        billing_details -> Nullable<Bytea>,
+        #[max_length = 255]
+        merchant_order_reference_id -> Nullable<Varchar>,
+        shipping_details -> Nullable<Bytea>,
+        is_payment_processor_token_flow -> Nullable<Bool>,
         shipping_cost -> Nullable<Int8>,
         #[max_length = 32]
         organization_id -> Varchar,
@@ -1028,6 +1135,8 @@ diesel::table! {
         #[max_length = 255]
         created_by -> Nullable<Varchar>,
         is_iframe_redirection_enabled -> Nullable<Bool>,
+        #[max_length = 2048]
+        extended_return_url -> Nullable<Varchar>,
         is_payment_id_from_merchant -> Nullable<Bool>,
         #[max_length = 64]
         payment_channel -> Nullable<Varchar>,
@@ -1075,13 +1184,39 @@ diesel::table! {
     use diesel::sql_types::*;
     use crate::enums::diesel_exports::*;
 
-    payment_methods (id) {
+    payment_methods (payment_method_id) {
         #[max_length = 64]
         customer_id -> Varchar,
         #[max_length = 64]
         merchant_id -> Varchar,
+        #[max_length = 64]
+        payment_method_id -> Varchar,
+        accepted_currency -> Nullable<Array<Nullable<Currency>>>,
+        #[max_length = 32]
+        scheme -> Nullable<Varchar>,
+        #[max_length = 128]
+        token -> Nullable<Varchar>,
+        #[max_length = 255]
+        cardholder_name -> Nullable<Varchar>,
+        #[max_length = 64]
+        issuer_name -> Nullable<Varchar>,
+        #[max_length = 64]
+        issuer_country -> Nullable<Varchar>,
+        payer_country -> Nullable<Array<Nullable<Text>>>,
+        is_stored -> Nullable<Bool>,
+        #[max_length = 32]
+        swift_code -> Nullable<Varchar>,
+        #[max_length = 128]
+        direct_debit_token -> Nullable<Varchar>,
         created_at -> Timestamp,
         last_modified -> Timestamp,
+        payment_method -> Nullable<Varchar>,
+        #[max_length = 64]
+        payment_method_type -> Nullable<Varchar>,
+        #[max_length = 128]
+        payment_method_issuer -> Nullable<Varchar>,
+        payment_method_issuer_code -> Nullable<PaymentMethodIssuerCode>,
+        metadata -> Nullable<Json>,
         payment_method_data -> Nullable<Bytea>,
         #[max_length = 64]
         locker_id -> Nullable<Varchar>,
@@ -1103,17 +1238,6 @@ diesel::table! {
         #[max_length = 64]
         network_token_locker_id -> Nullable<Varchar>,
         network_token_payment_method_data -> Nullable<Bytea>,
-        #[max_length = 64]
-        locker_fingerprint_id -> Nullable<Varchar>,
-        #[max_length = 64]
-        payment_method_type_v2 -> Nullable<Varchar>,
-        #[max_length = 64]
-        payment_method_subtype -> Nullable<Varchar>,
-        #[max_length = 64]
-        id -> Varchar,
-        #[max_length = 64]
-        external_vault_source -> Nullable<Varchar>,
-        external_vault_token_data -> Nullable<Bytea>,
     }
 }
 
@@ -1238,7 +1362,11 @@ diesel::table! {
     use diesel::sql_types::*;
     use crate::enums::diesel_exports::*;
 
-    refund (id) {
+    refund (merchant_id, refund_id) {
+        #[max_length = 64]
+        internal_reference_id -> Varchar,
+        #[max_length = 64]
+        refund_id -> Varchar,
         #[max_length = 64]
         payment_id -> Varchar,
         #[max_length = 64]
@@ -1274,9 +1402,15 @@ diesel::table! {
         profile_id -> Nullable<Varchar>,
         #[max_length = 32]
         updated_by -> Varchar,
+        #[max_length = 32]
+        merchant_connector_id -> Nullable<Varchar>,
         charges -> Nullable<Jsonb>,
         #[max_length = 32]
         organization_id -> Varchar,
+        #[max_length = 512]
+        connector_refund_data -> Nullable<Varchar>,
+        #[max_length = 512]
+        connector_transaction_data -> Nullable<Varchar>,
         split_refunds -> Nullable<Jsonb>,
         #[max_length = 255]
         unified_code -> Nullable<Varchar>,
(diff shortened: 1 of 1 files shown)


crates/router/src/routes/subscription.rs
```diff
@@ -0,0 +1,52 @@
+//! Analysis for usage of Routing in Payment flows
+//!
+//! Functions that are used to perform the api level configuration, retrieval, updation
+//! of Routing configs.
```
jagan-jaya:
```suggestion

```

crates/router/src/routes/subscription.rs
```diff
@@ -0,0 +1,52 @@
...
+};
+
+#[cfg(all(feature = "olap", feature = "v1"))]
+#[instrument(skip_all)]
+pub async fn create_subscription(
+    state: web::Data<AppState>,
+    req: HttpRequest,
+    json_payload: web::Json<subscription_types::CreateSubscriptionRequest>,
+) -> impl Responder {
+    let flow = Flow::RoutingCreateConfig;
```
jagan-jaya:
```suggestion
    let flow = Flow::CreateSubscription;
```

crates/router/src/routes/subscription.rs
```diff
@@ -0,0 +1,52 @@
...
+            ));
+            subscription::create_subscription(state, merchant_context, payload.clone())
+        },
+        auth::auth_type(
+            &auth::HeaderAuth(auth::ApiKeyAuth {
+                is_connected_allowed: false,
+                is_platform_allowed: false,
+            }),
+            &auth::JWTAuth {
+                permission: Permission::ProfileRoutingWrite,
```
jagan-jaya:
```suggestion
                permission: Permission::ProfileSubscriptionCreate,
```

crates/router/src/core/subscription/utils.rs
```diff
@@ -0,0 +1,281 @@
...
+        .as_ref()
+        .and_then(|customer_details| customer_details.tax_registration_id.clone());
+
+    CustomerDetails {
+        customer_id,
+        name: customer_name,
+        email: customer_email,
+        phone: customer_phone,
+        phone_country_code: customer_phone_code,
+        tax_registration_id,
```
jagan-jaya:
```suggestion
        name: customer.and_then(|c| c.name.clone()),
        email: customer.and_then(|c| c.email.clone()),
        phone: customer.and_then(|c| c.phone.clone()),
        phone_country_code: customer.and_then(|c| c.phone_country_code.clone()),
        tax_registration_id: customer.and_then(|c| c.tax_registration_id.clone()),
```

crates/router/src/core/subscription/utils.rs
```diff
@@ -0,0 +1,281 @@
...
+
+    let customer_phone_code = request
+        .customer
+        .as_ref()
+        .and_then(|customer_details| customer_details.phone_country_code.clone());
+
+    let tax_registration_id = request
+        .customer
+        .as_ref()
+        .and_then(|customer_details| customer_details.tax_registration_id.clone());
```
jagan-jaya:
```suggestion
let customer = request.customer.as_ref();
```

crates/router/src/core/subscription/utils.rs
```diff
@@ -0,0 +1,281 @@
...
+                        format!(
+                            "Failed to insert customer [id - {customer_id:?}] for merchant [id - {merchant_id:?}]",
+                        )
+                    })?,
+                ))
+            } else {
+                Err(report!(errors::ApiErrorResponse::InvalidRequestData {
+                    message: format!("customer for id - {customer_id:?} not found"),
+                }))
+            }
```
jagan-jaya:
```suggestion
```suggestion
 match create_customer(
            state.clone(),
            merchant_context.clone(),
            customer_migration.customer,
            customer_migration.connector_customer_details,
        )
        .await?
```
we can directly call create_customer in core::customers if customer is not present already
```

**Review by jagan-jaya — CHANGES_REQUESTED** [review]

File: crates/router/src/core/subscription.rs [hunk 7]
```diff
@@ -0,0 +1,81 @@
...
+            .customer
+            .as_ref()
+            .and_then(|customer| customer.id.clone())
+    } else {
+        request.customer_id.clone()
+    }
+    .ok_or(errors::ApiErrorResponse::CustomerNotFound)
+    .attach_printable("subscriptions: unable to create a customer")?;
+
+    // If provided we can strore plan_id, coupon_code etc as metadata
```
jagan-jaya: is this comment needed?

jagan-jaya (approved):

crates/router/src/routes/app.rs
```diff
@@ -1155,6 +1157,22 @@ impl Routing {
...
 
+#[cfg(feature = "olap")]
+pub struct Subscription;
+
+#[cfg(all(feature = "olap", feature = "v1"))]
+impl Subscription {
+    pub fn server(state: AppState) -> Scope {
+        web::scope("/subscription")
+            .app_data(web::Data::new(state.clone()))
+            .service(web::resource("/create").route(
```
jagan-jaya: remove create and accept confirm:false in body

jagan-jaya (dismissed):

crates/api_models/src/subscription.rs
```diff
@@ -0,0 +1,108 @@
...
+};
+
+pub const SUBSCRIPTION_ID_PREFIX: &str = "sub";
+
+#[derive(Debug, Clone, serde::Serialize, serde::Deserialize)]
+pub struct CreateSubscriptionRequest {
+    pub subscription_id: Option<String>,
+    pub plan_id: Option<String>,
+    pub coupon_code: Option<String>,
+    pub mca_id: Option<String>,
```
sai-harsha-vardhan: Please use domain type for MCA ID

crates/api_models/src/subscription.rs
```diff
@@ -0,0 +1,108 @@
+use common_utils::{events::ApiEventMetric, pii};
+
+use crate::{
+    customers::{CustomerRequest, CustomerResponse},
+    payments::CustomerDetailsResponse,
+};
+
+pub const SUBSCRIPTION_ID_PREFIX: &str = "sub";
+
+#[derive(Debug, Clone, serde::Serialize, serde::Deserialize)]
```
sai-harsha-vardhan: Add ToSchema for all the api_models and provide description for each of the fields

crates/api_models/src/subscription.rs
```diff
@@ -0,0 +1,108 @@
...
+    pub fn get_customer_id(&self) -> Option<&common_utils::id_type::CustomerId> {
+        self.customer_id
+            .as_ref()
+            .or_else(|| self.customer.as_ref()?.customer_id.as_ref())
+    }
+}
+
+#[derive(Debug, Clone, serde::Serialize)]
+pub struct CreateSubscriptionResponse {
+    pub subscription: Subscription,
```
sai-harsha-vardhan: Why do we want this to be a nested object? Subscription response is expected to give back all the relevant details right? Subscription record is not a part of it

crates/api_models/src/subscription.rs
```diff
@@ -0,0 +1,108 @@
...
+    pub client_secret: Option<String>,
```
sai-harsha-vardhan: We can mark this as Secret

crates/api_models/src/subscription.rs
sai-harsha-vardhan: Do we need Serialize for this?

crates/api_models/src/subscription.rs
```diff
@@ -0,0 +1,108 @@
...
+    pub merchant_id: String,
+    pub mca_id: Option<String>,
```
sai-harsha-vardhan: nit: rename

crates/api_models/src/subscription.rs
```diff
@@ -0,0 +1,108 @@
...
+pub enum SubscriptionStatus {
+    Created,
+    Active,
+    InActive,
+}
+
+#[derive(Debug, Clone, serde::Serialize)]
+pub struct Invoice {
+    pub id: String,
+    pub total: u64,
```
sai-harsha-vardhan: Is this amount? If yes, please use MinorUnit

crates/api_models/src/subscription.rs
```diff
@@ -0,0 +1,108 @@
...
+
+#[derive(Debug, Clone, serde::Serialize)]
+pub struct CreateSubscriptionResponse {
+    pub subscription: Subscription,
+    pub client_secret: Option<String>,
+    pub merchant_id: String,
+    pub mca_id: Option<String>,
+    pub coupon_code: Option<String>,
+    pub customer: Option<CustomerDetailsResponse>,
+    pub invoice: Option<Invoice>,
```
sai-harsha-vardhan: Does a subscription has only one invoice?

crates/api_models/src/subscription.rs
```diff
@@ -0,0 +1,108 @@
...
+            coupon_code: None,
+            customer: None,
+            invoice: None,
+        }
+    }
+}
+
+pub fn map_customer_resp_to_details(r: &CustomerResponse) -> CustomerDetailsResponse {
+    CustomerDetailsResponse {
+        id: Some(r.customer_id.clone()),
```
sai-harsha-vardhan: Please use better names for variables under closures as well

crates/api_models/src/subscription.rs
```diff
@@ -0,0 +1,108 @@
...
+impl Invoice {
+    pub fn new(id: impl Into<String>, total: u64) -> Self {
+        Self {
+            id: id.into(),
+            total,
+        }
+    }
+}
+impl CreateSubscriptionResponse {
+    #[allow(clippy::too_many_arguments)]
```
sai-harsha-vardhan, line 83: Why do we need this?

crates/api_models/src/subscription.rs
```diff
@@ -0,0 +1,108 @@
...
+            .as_ref()
+            .or_else(|| self.customer.as_ref()?.customer_id.as_ref())
+    }
+}
+
+#[derive(Debug, Clone, serde::Serialize)]
+pub struct CreateSubscriptionResponse {
+    pub subscription: Subscription,
+    pub client_secret: Option<String>,
+    pub merchant_id: String,
```
sai-harsha-vardhan: Please use Domain ID

crates/api_models/src/subscription.rs
```diff
@@ -0,0 +1,108 @@
...
+            id: id.into(),
+            total,
+        }
+    }
+}
+impl CreateSubscriptionResponse {
+    #[allow(clippy::too_many_arguments)]
+    pub fn new(
+        subscription: Subscription,
+        merchant_id: impl Into<String>,
```
sai-harsha-vardhan: We can use Domain ID here

crates/api_models/src/subscription.rs
```diff
@@ -0,0 +1,108 @@
...
+    CustomerDetailsResponse {
+        id: Some(r.customer_id.clone()),
+        name: r.name.as_ref().map(|n| n.clone().into_inner()),
+        email: r.email.as_ref().map(|e| pii::Email::from(e.clone())),
+        phone: r.phone.as_ref().map(|p| p.clone().into_inner()),
+        phone_country_code: r.phone_country_code.clone(),
+    }
+}
+
+impl ApiEventMetric for CreateSubscriptionResponse {}
```
sai-harsha-vardhan, line 108: We can have API Events based on subscription id for subscription flows right?

crates/api_models/src/subscription.rs
```diff
@@ -0,0 +1,108 @@
...
+use crate::{
+    customers::{CustomerRequest, CustomerResponse},
+    payments::CustomerDetailsResponse,
+};
+
+pub const SUBSCRIPTION_ID_PREFIX: &str = "sub";
+
+#[derive(Debug, Clone, serde::Serialize, serde::Deserialize)]
+pub struct CreateSubscriptionRequest {
+    pub subscription_id: Option<String>,
```
sai-harsha-vardhan: Can we have domain type for subscription_id?

crates/diesel_models/src/subscription.rs
```diff
@@ -0,0 +1,103 @@
...
+use diesel::{AsChangeset, Identifiable, Insertable, Queryable, Selectable};
+use serde::{Deserialize, Serialize};
+
+use crate::schema::subscription;
+
+#[derive(Clone, Debug, Eq, Insertable, PartialEq, Serialize, Deserialize)]
+#[diesel(table_name = subscription)]
+pub struct SubscriptionNew {
+    subscription_id: String,
+    status: String,
```
sai-harsha-vardhan, line 12: Why's this a String?

crates/router/src/core/subscription/utils.rs
```diff
@@ -0,0 +1,73 @@
...
+    match db
+        .find_customer_optional_by_customer_id_merchant_id(
+            key_manager_state,
+            &customer_id,
+            merchant_id,
+            merchant_context.get_merchant_key_store(),
+            merchant_context.get_merchant_account().storage_scheme,
+        )
+        .await
+        .change_context(errors::CustomersErrorResponse::InternalServerError)
```
sai-harsha-vardhan: use `to_not_found_response` here

crates/api_models/src/subscription.rs
jagan-jaya:
```suggestion
    pub total: u64,
    pub currency: Currency,
```

crates/diesel_models/src/subscription.rs
```diff
@@ -1,4 +1,8 @@
+<<<<<<< HEAD
+use common_utils::{generate_id_with_default_len, pii::SecretSerdeValue};
+=======
 use common_utils::pii::SecretSerdeValue;
+>>>>>>> f10a713219ab13a4c584a1178e7ee2488f634f9e
```
jagan-jaya: resolve conflicts

jagan-jaya (changes requested):

crates/router/src/core/subscription.rs
```diff
@@ -0,0 +1,97 @@
...
+    let mut subscription = SubscriptionNew::new(
```
jagan-jaya, line 30:
```suggestion
    let subscription = SubscriptionNew::new(
```

prajjwalkumar17: We need it as mut because we are adding `client_secret` later in the flow.

jagan-jaya (approved):

jagan-jaya (approved):

crates/api_models/src/subscription.rs
```diff
@@ -0,0 +1,114 @@
...
+    pub profile_id: common_utils::id_type::ProfileId,
+    pub plan_id: Option<String>,
+    pub coupon_code: Option<String>,
+    pub mca_id: Option<String>,
```
Sarthak1799: `pub merchant_connector_account_id: Option<common_utils::id_type::MerchantConnectorAccountId>`

jagan-jaya (approved):

Sarthak1799 (approved):

apoorvdixit88 (dismissed):
Dashboard changes look fine.

crates/router/src/core/payouts/helpers.rs
```diff
@@ -756,7 +756,7 @@ pub(super) async fn get_or_create_customer_details(
 }
 
 #[cfg(feature = "v1")]
-pub(super) async fn get_or_create_customer_details(
+pub async fn get_or_create_customer_details(
```
sai-harsha-vardhan: Why's this change relevant?

crates/router/src/core/subscription/utils.rs
```diff
@@ -0,0 +1,73 @@
...
+            let api_customer: CustomerResponse =
+                (customer, None::<api_models::payments::AddressDetails>).foreign_into();
+            Ok(ApplicationResponse::Json(api_customer))
+        }
+
+        // Customer not found
+        None => Ok(create_customer(
+            state,
+            merchant_context,
+            customer_request.ok_or(errors::CustomersErrorResponse::CustomerNotFound)?,
```
sai-harsha-vardhan: We should throw BadRequest error and not 404 right?

crates/router/src/core/subscription/utils.rs
```diff
@@ -0,0 +1,73 @@
...
+            None,
+        )
+        .await?),
+    }
+}
+
+pub fn get_customer_details_from_request(request: CreateSubscriptionRequest) -> CustomerDetails {
```
sai-harsha-vardhan: This can be a From conversion or impl based fn

crates/api_models/src/subscription.rs
prajjwalkumar17: Why it is required to be marked as secret? As this is client facing response, how will they get hold of `client_secret` afterwards for further calls?

crates/api_models/src/subscription.rs
prajjwalkumar17: The struct is temporarily removed and this will be taken care, once this is added back.

crates/api_models/src/subscription.rs
prajjwalkumar17: This will be thought upon later, and this is temporarily removed.

crates/api_models/src/subscription.rs
prajjwalkumar17: We can have this, probably in some another PR, if required an issue can be created, but right now due to push from product this can be ignored.

crates/router/src/core/payouts/helpers.rs
prajjwalkumar17: This was already refactored.

crates/api_models/src/subscription.rs
prajjwalkumar17: Sure for logs, we will make this change.

crates/api_models/src/subscription.rs
sai-harsha-vardhan: Please create an issue for this

crates/api_models/src/subscription.rs
prajjwalkumar17: Sure we can have this in future, will create an issue for this.

crates/api_models/src/subscription.rs
```diff
@@ -0,0 +1,145 @@
...
+    pub plan_id: Option<String>,
+
+    /// Optional coupon code applied to the subscription.
+    pub coupon_code: Option<String>,
+
+    /// Optional merchant connector account ID for routing payments.
+    pub merchant_connector_account_id: Option<common_utils::id_type::MerchantConnectorAccountId>,
+
+    /// Whether to immediately confirm the subscription on creation.
+    pub confirm: bool,
```
sai-harsha-vardhan: Is this relevant? If it's required can we have a default as false?

crates/api_models/src/subscription.rs
```diff
@@ -0,0 +1,145 @@
...
+
+/// Possible states of a subscription lifecycle.
+///
+/// - `Created`: Subscription was created but not yet activated.
+/// - `Active`: Subscription is currently active.
+/// - `InActive`: Subscription is inactive (e.g., cancelled or expired).
+#[derive(Debug, Clone, serde::Serialize, strum::EnumString, strum::Display, ToSchema)]
+pub enum SubscriptionStatus {
+    /// Subscription is created but not yet active.
+    Created,
```
sai-harsha-vardhan: Don't we need `pending`?

crates/api_models/src/subscription.rs
```diff
@@ -0,0 +1,145 @@
...
+}
+
+/// Maps a [`CustomerResponse`] into a [`CustomerDetailsResponse`].
+///
+/// This is used to transform customer information returned by the customer API
+/// into the response format expected by the subscription API.
+pub fn map_customer_resp_to_details(response: &CustomerResponse) -> CustomerDetailsResponse {
+    CustomerDetailsResponse {
+        id: Some(response.customer_id.clone()),
+        name: response.name.as_ref().map(|n| n.clone().into_inner()),
```
sai-harsha-vardhan: please use meaningful names in closures as well

crates/router/src/core/subscription.rs
```diff
@@ -0,0 +1,89 @@
...
+        };
+        response.customer = customer_table_response;
+        response
+            .customer
+            .as_ref()
+            .and_then(|customer| customer.id.clone())
+    } else {
+        request.customer_id.clone()
+    }
+    .ok_or(errors::ApiErrorResponse::CustomerNotFound)
```
sai-harsha-vardhan: We can't say CustomerNotFound in case creation fails right?

crates/router/src/core/subscription.rs
```diff
@@ -0,0 +1,89 @@
...
+        None,
+        merchant_context.get_merchant_account().get_id().clone(),
+        customer_id,
+        None,
+        request.profile_id,
+    );
+    response.client_secret = Some(subscription.generate_and_set_client_secret());
+    db.insert_subscription_entry(subscription)
+        .await
+        .to_not_found_response(errors::ApiErrorResponse::ResourceIdNotFound)
```
sai-harsha-vardhan: Why do we need to_not_found_response during insertion?

crates/router/src/routes/subscription.rs
```diff
@@ -0,0 +1,52 @@
...
+};
+
+use crate::{
+    core::{api_locking, subscription},
+    routes::AppState,
+    services::{api as oss_api, authentication as auth, authorization::permissions::Permission},
+    types::domain,
+};
+
+#[cfg(all(feature = "olap", feature = "v1"))]
```
sai-harsha-vardhan: Why's this a olap feature? It should be OLTP

crates/api_models/src/subscription.rs
sai-harsha-vardhan: As this is relatively small change, it would be better if we make this changes here

crates/router/src/core/subscription.rs
prajjwalkumar17: Changed this to `InternalServerError`.

crates/router/src/core/subscription.rs
prajjwalkumar17: Changed this to `InternalServerError`.

crates/api_models/src/subscription.rs
prajjwalkumar17: This is made optional.

crates/api_models/src/subscription.rs
```diff
@@ -0,0 +1,151 @@
...
+    }
+}
+
+/// Response payload returned after successfully creating a subscription.
+///
+/// Includes details such as subscription ID, status, plan, merchant, and customer info.
+#[derive(Debug, Clone, serde::Serialize, ToSchema)]
+pub struct CreateSubscriptionResponse {
+    /// Unique identifier for the subscription.
+    pub id: String,
```
sai-harsha-vardhan: Could we make this consistent with the request and make it `subscription_id`?

crates/api_models/src/subscription.rs
```diff
@@ -0,0 +1,151 @@
...
+        profile_id: common_utils::id_type::ProfileId,
+        merchant_id: common_utils::id_type::MerchantId,
+        merchant_connector_account_id: Option<common_utils::id_type::MerchantConnectorAccountId>,
+    ) -> Self {
+        Self {
+            id,
+            status,
+            plan_id,
+            profile_id,
+            client_secret: None,
```
sai-harsha-vardhan: Why are these None in `new` implementation?

crates/router/src/routes/app.rs
```diff
@@ -1155,6 +1157,22 @@ impl Routing {
     }
 }
 
+#[cfg(feature = "olap")]
+pub struct Subscription;
+
+#[cfg(all(feature = "olap", feature = "v1"))]
```
sai-harsha-vardhan: This should be OLTP API

crates/api_models/src/subscription.rs
prajjwalkumar17: we are using impl function to create the `client_secret` afterwards.

crates/api_models/src/subscription.rs
```diff
@@ -0,0 +1,148 @@
...
+    customers::{CustomerRequest, CustomerResponse},
+    payments::CustomerDetailsResponse,
+};
+
+/// Request payload for creating a subscription.
+///
+/// This struct captures details required to create a subscription,
+/// including plan, profile, merchant connector, and optional customer info.
+#[derive(Debug, Clone, serde::Serialize, serde::Deserialize, ToSchema)]
+pub struct CreateSubscriptionRequest {
```
hrithikesh026, line 15: Please generate openapi spec for this new endpoint.

crates/api_models/src/subscription.rs
```diff
@@ -0,0 +1,148 @@
...
+    pub coupon_code: Option<String>,
+
+    /// Optional merchant connector account ID for routing payments.
+    pub merchant_connector_account_id: Option<common_utils::id_type::MerchantConnectorAccountId>,
+
+    /// Whether to immediately confirm the subscription on creation.
+    pub confirm: Option<bool>,
+
+    /// Optional customer ID associated with this subscription.
+    pub customer_id: Option<common_utils::id_type::CustomerId>,
```
hrithikesh026: Why to accept customer_id in two different places?

crates/router/src/core/subscription.rs
```diff
@@ -0,0 +1,84 @@
...
+            .customer
+            .as_ref()
+            .and_then(|customer| customer.id.clone())
+    } else {
+        request.customer_id.clone()
+    }
+    .ok_or(errors::ApiErrorResponse::InternalServerError)
+    .attach_printable("subscriptions: unable to create a customer")?;
+
+    // If provided we can strore plan_id, coupon_code etc as metadata
```
hrithikesh026:
```suggestion
    // If provided we can store plan_id, coupon_code etc as metadata
```

crates/router/src/core/subscription/utils.rs
```diff
@@ -0,0 +1,74 @@
...
+    },
+    routes::SessionState,
+    types::{api::CustomerResponse, transformers::ForeignInto},
+};
+
+pub async fn get_or_create_customer(
+    state: SessionState,
+    customer_request: Option<CustomerRequest>,
+    merchant_context: MerchantContext,
+) -> errors::CustomerResponse<CustomerResponse> {
```
hrithikesh026: There already a similar function that does this.
https://github.com/juspay/hyperswitch/blob/d4e717aff94abd99dd9038420fe83e10627946ae/crates/router/src/core/payouts/helpers.rs#L775

crates/router/src/core/subscription.rs
```diff
@@ -0,0 +1,84 @@
...
+
+pub async fn create_subscription(
+    state: SessionState,
+    merchant_context: MerchantContext,
+    request: subscription_types::CreateSubscriptionRequest,
+) -> RouterResponse<CreateSubscriptionResponse> {
+    let store = state.store.clone();
+    let db = store.as_ref();
+    let id = common_utils::id_type::SubscriptionId::generate();
+    let mut response = CreateSubscriptionResponse::new(
```
hrithikesh026: response must be constructed from domain model only.

crates/router/src/core/subscription.rs
```diff
@@ -0,0 +1,84 @@
...
+        request.merchant_connector_account_id,
+        None,
+        None,
+        merchant_context.get_merchant_account().get_id().clone(),
+        customer_id,
+        None,
+        request.profile_id,
+    );
+    response.client_secret = Some(subscription.generate_and_set_client_secret());
+    db.insert_subscription_entry(subscription)
```
hrithikesh026:
```suggestion
    let subscription = db.insert_subscription_entry(subscription)
```

ues this subscription to generate response.

crates/api_models/src/subscription.rs
prajjwalkumar17: [Issue](https://github.com/juspay/hyperswitch/issues/9415) for this is already created and @AnkitKmrGupta is working on this, we want this change to move in a separate PR as that will account for all the routes at once.

crates/api_models/src/subscription.rs
prajjwalkumar17: This is how we handle customer request for our payment's request. If a customer is already created, the outer id should be populated and if a customer isn't created and merchant wants to provide all the details regarding the customer the inner type should be used with/without customer_id.

crates/router/src/core/subscription/utils.rs
prajjwalkumar17: So there are mainly two reasons to not use that function (which is present in `payouts/helpers`):
--> We have created subscription's util function in such a way that it returns us with `CustomerResponse`, which we directly populate in our response, again this can be justified by writing a `try_from`, after which still codebit will increase instead of that I used `create_customer` functionality which is present in `core/customers` which isn't used by this function and the code is again duplicated there.

--> rather than using a `payouts/helper` function which might/might_not change in future(which is highly unlikely still) as we are not maintainers of that file, so it can be changed without a review from us, which I wouldn't want to.

If required this functionality can be refactored to a common place say `core/customer` or something in future then we can write try_froms if required.

crates/router/src/core/subscription.rs
prajjwalkumar17: Sure, can you explain how will this help as the fields being populated in response is from request and `merchant_account`, how will populating same fields in `diesel_models` struct and then re-utilizing it will help us ?

hrithikesh026: If for some reason, the fields are not populated in DB, we will never know about it if the response is constructed from request.

crates/api_models/src/subscription.rs
hrithikesh026: What if both customer ids are given and both are different customer id?

What if inner customer_id is given but that customer_id already present in DB?

What is outer customer_id is given along with some customer detail in the nested customer object.

prajjwalkumar17: What if both customer ids are given and both are different customer id?
--> The inner customer_id is always given priority and a new customer will be created using the inner id.

What if inner customer_id is given but that customer_id already present in DB?
--> This won't be a problem as we are doing fetch or create.

What is outer customer_id is given along with some customer detail in the nested customer object.
--> In this case we fetch the outer_customer id and update it with the provided inner fields.

crates/router/src/core/subscription.rs
prajjwalkumar17: Can this be possible if fields are not populated and we don't incur an error, we already have error_handling implemented on the code:
Correct me if I am wrong, as According to me either all of the fields will be populated or none.
```
    db.insert_subscription_entry(subscription)
        .await
        .attach_printable("subscriptions: unable to insert subscription entry to database")?;

```

prajjwalkumar17: Anyways for sake of review I have refactored as asked, can you once again review this.

crates/api_models/src/subscription.rs
hrithikesh026: What if inner customer_id is given but that customer_id already present in DB?
--> This won't be a problem as we are doing fetch or create.

You previously said, if the inner id is provided we will always create the customer. Now you are saying we will either create or get.

crates/router/src/core/subscription.rs
hrithikesh026: You may miss to populate fields from request in `SubscriptionNew` struct.

config/development.toml
```diff
@@ -1317,8 +1317,8 @@ click_to_pay = {connector_list = "adyen, cybersource, trustpay"}
 
 [open_router]
 dynamic_routing_enabled = false
-static_routing_enabled = false
-url = "http://localhost:8080"
+static_routing_enabled = true
+url = "http://localhost:8081"
```
hrithikesh026: is this intended?

crates/router/src/core/subscription.rs
```diff
@@ -0,0 +1,91 @@
...
+    let db = store.as_ref();
+    let id = common_utils::id_type::SubscriptionId::generate();
+    let mut customer_response = None;
+    let customer = CustomerDetails::from(request.clone());
+    let customer_id = if customer.customer_id.is_some()
+        || customer.name.is_some()
+        || customer.email.is_some()
+        || customer.phone.is_some()
+        || customer.phone_country_code.is_some()
+    {
```
hrithikesh026: why should we get_or_create customer only if any of these fields are provided?
What if a new field is introduced in CustomerRequest struct?

crates/router/src/core/subscription.rs
```diff
@@ -0,0 +1,91 @@
...
+    );
+
+    subscription.generate_and_set_client_secret();
+    let subscription_response = db
+        .insert_subscription_entry(subscription)
+        .await
+        .change_context(errors::ApiErrorResponse::InternalServerError)
+        .attach_printable("subscriptions: unable to insert subscription entry to database")?;
+
+    let response = CreateSubscriptionResponse::new(
```
hrithikesh026, line 52: better to have conversion from (diesel::Subcription, domain::Customer) -> SubcriptionRespone.

crates/api_models/src/subscription.rs
hrithikesh026: It will be better to accept customer_id in the inner object only. Eliminates a lot of confusion.

prajjwalkumar17: `You previously said, if the inner id is provided we will always create the customer`
I never mentioned we will always create, I mentioned:
`if a customer isn't created and merchant wants to provide all the details regarding the customer the inner type should be used with/without customer_id.`

in this case if a customer is already created then, it should fetch and not re-create.
Moreover, as I am calling the functionality of customer which is already coded and not re-creating it, the issues that you mentioned are previously solved. If required you can refer the code, moreover what I think is whenever an external entity provides an id for creation and storage we always do a `fetch_or_create`, if and all at any place if this isn't happening, it should be changed.

crates/router/src/core/subscription.rs
prajjwalkumar17: This is refactored anyways.

crates/router/src/core/subscription.rs
prajjwalkumar17: Currently we don't have a domain model for subscription as we are not internally converting anything but soon if requirements will occur to create one will do as asked.

config/development.toml
prajjwalkumar17: This is restored.

crates/router/src/core/subscription.rs
prajjwalkumar17: Sure, for solving this out I will move the `fetch_or_create` functionality to a common place, say `customers.rs` and respective codeowners can call the same function for their use-cases, will that be fine @hrithikesh026 ? , so that code remains in a common place and can be easily refactored, if required.

crates/api_models/src/subscription.rs
```diff
@@ -0,0 +1,151 @@
...
+};
+
+/// Request payload for creating a subscription.
+///
+/// This struct captures details required to create a subscription,
+/// including plan, profile, merchant connector, and optional customer info.
+#[derive(Debug, Clone, serde::Serialize, serde::Deserialize, ToSchema)]
+pub struct CreateSubscriptionRequest {
+    /// Unique identifier for the subscription (optional, generated if missing).
+    pub subscription_id: Option<common_utils::id_type::SubscriptionId>,
```
jarnura: We should not get the database reference id from merchants, id should be generated for us, introduce a merchant_reference_id where they can pass their ids but that is not the id we will be processing with our api's.

The id which we generate should be global unique without any composite id logic

crates/api_models/src/subscription.rs
```diff
@@ -0,0 +1,151 @@
...
+
+    /// Associated profile ID for this subscription.
+    pub profile_id: common_utils::id_type::ProfileId,
```
jarnura: Any lineage id should come from headers and not from the body

crates/api_models/src/subscription.rs
```diff
@@ -0,0 +1,151 @@
...
+
+    /// Identifier for the subscription plan.
+    pub plan_id: Option<String>,
+
+    /// Optional coupon code applied to the subscription.
+    pub coupon_code: Option<String>,
+
+    /// Optional merchant connector account ID for routing payments.
+    pub merchant_connector_account_id: Option<common_utils::id_type::MerchantConnectorAccountId>,
```
jarnura: This should be configured in profiles right? why we need to expose this to merchants? (these are you dashboard related stuffs)

crates/api_models/src/subscription.rs
```diff
@@ -0,0 +1,151 @@
...
+
+    /// Whether to immediately confirm the subscription on creation.
+    pub confirm: Option<bool>,
```
jarnura: We should have a separate api for create and confirm and create+confirm, we shouldn't not combine using this argument

crates/api_models/src/subscription.rs
```diff
@@ -0,0 +1,151 @@
...
+
+    /// Optional customer ID associated with this subscription.
+    pub customer_id: Option<common_utils::id_type::CustomerId>,
+
+    /// Optional full customer request data.
+    pub customer: Option<CustomerRequest>,
```
jarnura: Customer creation should happen outside this api and should not be included here

crates/api_models/src/subscription.rs
```diff
@@ -0,0 +1,151 @@
...
+    }
+}
+
+/// Response payload returned after successfully creating a subscription.
+///
+/// Includes details such as subscription ID, status, plan, merchant, and customer info.
+#[derive(Debug, Clone, serde::Serialize, ToSchema)]
+pub struct CreateSubscriptionResponse {
+    /// Unique identifier for the subscription.
+    pub subscription_id: common_utils::id_type::SubscriptionId,
```
jarnura:
```suggestion
    pub id: common_utils::id_type::SubscriptionId,
```

resource api name convention

crates/api_models/src/subscription.rs
```diff
@@ -0,0 +1,151 @@
...
+    pub merchant_id: common_utils::id_type::MerchantId,
+
+    /// Optional merchant connector account ID for routing payments.
+    pub merchant_connector_account_id: Option<common_utils::id_type::MerchantConnectorAccountId>,
+
+    /// Optional coupon code applied to this subscription.
+    pub coupon_code: Option<String>,
+
+    /// Optional customer details response.
+    pub customer: Option<CustomerDetailsResponse>,
```
jarnura: Customer creation should be outside the subscription create

jarnura (changes requested):

jagan-jaya (approved):

jagan-jaya (approved):

su-shivanshmathur (approved):

PR #9289, opened 2025-09-05
feat: introduce a framework to fetch configs from superposition

Shailesh-714:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

Adding Superposition as an independent module in External Services to abstract the complexity of integrating it with the core application.

### Additional Changes

## Motivation and Context

Updating the config fetch logic from db to superposition may introduce complexity and the need for developers to have context about superposition. This implementation abstracts the complexity and exposes generic methods to fetch configs from superposition.

## How did you test it?

1. Run `cargo run --features superposition --bin router -- -f ./config/docker_compose.toml`
2. Set `requires_cvv=true` in `superposition` dashboard:
Screenshot 2025-09-09 at 5 01 58 PM
3. Verify if the saved cards are listed in the `checkout page`:
Screenshot 2025-09-09 at 5 02 22 PM
4. Change the value in superposition dashboard to `requires_cvv=false`:
Screenshot 2025-09-09 at 5 02 49 PM
5. Verify no saved cards are listed in the `checkout page`:
Screenshot 2025-09-09 at 5 02 59 PM

## Checklist

### DIFF
--- a/crates/router/src/core/configs.rs
+++ b/crates/router/src/core/configs.rs
@@ -247,12 +247,10 @@ where
             .get_object_value(superposition_key, context.as_ref())
             .await
         {
-            Ok(json_value) => {
-                Some(
-                    serde_json::from_value::<T>(json_value)
-                        .change_context(errors::StorageError::DeserializationFailed)
-                )
-            }
+            Ok(json_value) => Some(
+                serde_json::from_value::<T>(json_value)
+                    .change_context(errors::StorageError::DeserializationFailed),
+            ),
             Err(err) => {
                 router_env::logger::warn!(
                     "Failed to retrieve config from superposition, falling back to database: {:?}",


.gitignore
```diff
@@ -5,6 +5,8 @@
 # Covers JetBrains IDEs: IntelliJ, RubyMine, PhpStorm, AppCode, PyCharm, CLion, Android Studio, WebStorm and Rider
 # Reference: https://intellij-support.jetbrains.com/hc/en-us/articles/206544839
 
+superposition-main
```
SanchithHegde: Why was this change made? Can you revert this?

config/dashboard.toml
SanchithHegde, line 1: Can you revert changes to this file, they are not related to the changes in this PR at all.

config/development.toml
```diff
@@ -1,3 +1,10 @@
+[superposition]
+enabled = true
+endpoint = "http://host.docker.internal:8080"
+token = "123456"
+org_id = "localorg"
+workspace_id = "dev"
```
SanchithHegde: Nits:

1. Move this later down in the file, this not be at the top of the file.
2. For the endpoint, you can use a URL like `localhost:<some port>`.
3. Maybe consider disabling by default, since this would fail unless everyone running hyperswitch also runs superposition beforehand.
   - If everyone is _always_ expected to run superposition beforehand, then you'd have to update the local setup guide in `docs/try_local_system.md`.

Also, add these configs in the following files:
- `config/config.example.toml`: Add comments explaining about the section, and each config option.
- `config/deployments/env_specific.toml`: Same, add comments here as well.

config/docker_compose.toml
```diff
@@ -3,6 +3,13 @@ port = 8080
 host = "0.0.0.0"
 request_body_limit = 16_384 # Post request body is limited to 16k.
 
+[superposition]
+enabled = true
+endpoint = "http://host.docker.internal:8080"
```
SanchithHegde: Ummm, but we'd be running the hyperswitch server on port 8080 and it would be exposed on the host?

Either consider disabling this by default, or include the superposistion service in the docker compose files (`docker-compose.yml` and `docker-compose-development.yml`), and then you can just use the service name for the hostname, need not use `host.docker.internal`.

crates/external_services/Cargo.toml
```diff
@@ -67,6 +68,8 @@ http = "0.2.12"
 url = { version = "2.5.4", features = ["serde"] }
 quick-xml = { version = "0.31.0", features = ["serialize"] }
 unified-connector-service-client = { git = "https://github.com/juspay/connector-service", rev = "a56de0a4ee393af5c401f02b6b4344cb263f4cac", package = "rust-grpc-client" }
+open-feature = { version = "0.2.5", optional = true }
+superposition_provider = { git = "https://github.com/juspay/superposition", branch="main", optional = true }
```
SanchithHegde: Nope, we shouldn't be using moving targets like branches when adding git dependencies. @knutties Can't we use the crate from `crates.io`, or is there a specific reason for us to use this as a git dependency?

crates/external_services/src/lib.rs
```diff
@@ -12,6 +12,9 @@ pub mod file_storage;
 #[cfg(feature = "hashicorp-vault")]
 pub mod hashicorp_vault;
 
+/// Dynamic configuration service using Superposition
+pub mod superposition;
```
SanchithHegde: Move this entire module behind the `superposition` feature flag.

crates/external_services/src/superposition/mod.rs
SanchithHegde, line 1: Avoid the `mod.rs` naming convention, clippy should have also caught this.

Rename this file to `crates/external_services/src/superposition.rs`.

crates/external_services/src/superposition/mod.rs
```diff
@@ -0,0 +1,21 @@
+//! Superposition integration for dynamic configuration management
+//!
+//! This module provides a simple interface for retrieving configuration values
+//! with automatic fallback logic:
+//! 1. Superposition (dynamic config) - if enabled and available
+//! 2. Default values - reliable fallback
+//!
+//! The interface follows the same pattern as existing config methods,
+//! requiring minimal changes to application code.
```
SanchithHegde: We can remove these two lines, I think they don't add any value in the doc comments.

crates/external_services/src/superposition/mod.rs
```diff
@@ -0,0 +1,21 @@
...
+
+/// Interface definitions for superposition service
+pub mod interface;
+/// Service implementation for superposition integration
+pub mod service;
+/// Superposition client wrapper and configuration
+pub mod superposition;
+
+// Re-export commonly used types
```
SanchithHegde: Remove this comment.

crates/external_services/src/superposition/service.rs
```diff
@@ -0,0 +1,96 @@
...
+use common_utils::errors::CustomResult;
+use error_stack::ResultExt;
+
+use super::{
+    interface::{ConfigContext, SuperpositionError, SuperpositionInterface},
+    superposition::{SuperpositionClient, SuperpositionClientConfig},
+};
+
+/// Configuration for the superposition service
+pub type SuperpositionConfig = SuperpositionClientConfig;
```
SanchithHegde: Do we necessarily need this type alias?

crates/external_services/src/superposition/service.rs
```diff
@@ -0,0 +1,96 @@
...
+
+/// Main superposition service implementation  
+#[derive(Debug)]
+pub struct SuperpositionService {
+    superposition_client: Option<Arc<SuperpositionClient>>,
```
SanchithHegde: Nit: We can call this just `client`, since the struct name already indicates that this is about superposition.

crates/external_services/src/superposition/service.rs
```diff
@@ -0,0 +1,96 @@
...
+        context: Option<ConfigContext>,
+        default_value: String,
+    ) -> CustomResult<String, SuperpositionError> {
+        if let Some(superposition_client) = &self.superposition_client {
+            match superposition_client
+                .get_string_value(key, context.as_ref())
+                .await
+            {
+                Ok(value) => return Ok(value),
+                Err(_) => {} // Continue to default
```
SanchithHegde: Let's not completely ignore errors, let's log them at least.
Same applies for the other methods implemented below.

crates/external_services/src/superposition/superposition.rs
```diff
@@ -0,0 +1,255 @@
...
+
+/// Configuration for Superposition integration
+#[derive(Debug, Clone, serde::Deserialize)]
+pub struct SuperpositionClientConfig {
+    /// Whether Superposition is enabled
+    pub enabled: bool,
+    /// Superposition API endpoint
+    pub endpoint: String,
+    /// Authentication token for Superposition
+    pub token: Secret<String>,
```
SanchithHegde: Shouldn't this token be stored as a KMS encrypted value preferably? Can you make the required changes in `secrets_transformers.rs` to handle KMS decryption?

crates/external_services/src/superposition/superposition.rs
```diff
@@ -0,0 +1,255 @@
...
+        }
+    }
+}
+
+/// Superposition client wrapper
+pub struct SuperpositionClient {
+    #[cfg(feature = "superposition")]
+    client: open_feature::Client,
+    #[cfg(not(feature = "superposition"))]
+    _phantom: std::marker::PhantomData<()>,
```
SanchithHegde: Instead of doing it this way, let's move the entire superposition module behind the feature flag. Ideally, none of the superposition code should be accessible if the feature flag is disabled.

crates/external_services/src/superposition/superposition.rs
```diff
@@ -0,0 +1,255 @@
...
+
+            let provider_options = SuperpositionProviderOptions {
+                endpoint: config.endpoint.clone(),
+                token: config.token.expose(),
+                org_id: config.org_id.clone(),
+                workspace_id: config.workspace_id.clone(),
+                fallback_config: None,
+                evaluation_cache: None,
+                refresh_strategy: RefreshStrategy::Polling(PollingStrategy {
+                    interval: 15,
```
SanchithHegde: Let's move this to config instead of hardcoding it here.

crates/external_services/src/superposition/superposition.rs
```diff
@@ -0,0 +1,255 @@
...
+                }),
+                experimentation_options: None,
+            };
+
+            // Create provider and set up OpenFeature
+            let provider = SuperpositionProvider::new(provider_options);
+
+            router_env::logger::info!(
+                "SUPERPOSITION_CLIENT: Created provider, initializing OpenFeature API..."
+            );
```
SanchithHegde: Nit: I think we can remove the `SUPERPOSITION_CLIENT:` prefix part, the logs will anyway indicate which file the logs came from. We can improve the message maybe (let's indicate what we've completed so far):

```suggestion
            router_env::logger::info!("Created superposition provider");
```

crates/external_services/src/superposition/superposition.rs
```diff
@@ -0,0 +1,255 @@
...
+
+            // Initialize OpenFeature API and set provider
+            let mut api = open_feature::OpenFeature::singleton_mut().await;
+            api.set_provider(provider).await;
+
+            router_env::logger::info!("SUPERPOSITION_CLIENT: Provider set, creating client and waiting for initialization...");
```
SanchithHegde: Along same lines:

```suggestion
            router_env::logger::info!("OpenFeature provider set");
```

crates/external_services/src/superposition/superposition.rs
```diff
@@ -0,0 +1,255 @@
...
+
+            // Create client and wait for initialization (as per Superposition examples)
+            let client = api.create_client();
+            tokio::time::sleep(tokio::time::Duration::from_secs(3)).await;
```
SanchithHegde: Why is this number hardcoded? Can we consider moving it to config?

crates/external_services/src/superposition/superposition.rs
```diff
@@ -0,0 +1,255 @@
...
+
+            router_env::logger::info!(
+                "SUPERPOSITION_CLIENT: Client created and initialized successfully!"
+            );
```
SanchithHegde:
```suggestion
            router_env::logger::info!("Superposition client created and initialized successfully!");
```

crates/external_services/src/superposition/superposition.rs
```diff
@@ -0,0 +1,255 @@
...
+
+            Ok(Self { client })
+        }
+
+        #[cfg(not(feature = "superposition"))]
+        {
+            let _ = config; // Suppress unused variable warning
+            Err(SuperpositionError::InvalidConfiguration(
+                "Superposition feature is not enabled".to_string(),
+            ))
```
SanchithHegde: Let's avoid such blocks entirely, let's move the module behind the feature flag instead.

crates/external_services/src/superposition/superposition.rs
```diff
@@ -0,0 +1,255 @@
...
+    pub async fn get_string_value(
+        &self,
+        key: &str,
+        context: Option<&ConfigContext>,
+    ) -> Result<String, SuperpositionError> {
+        #[cfg(feature = "superposition")]
+        {
+            use std::collections::HashMap;
+
+            let evaluation_context = EvaluationContext {
```
SanchithHegde: Where is `EvaluationContext` coming from? I don't see any imports, or the type being defined in this module...

crates/external_services/src/superposition/superposition.rs
```diff
@@ -0,0 +1,255 @@
...
+                custom_fields: if let Some(ctx) = context {
+                    ctx.values
+                        .iter()
+                        .map(|(k, v)| (k.clone(), EvaluationContextFieldValue::String(v.clone())))
```
SanchithHegde: This code seems repetitive, we can consider adding a `From` implementation to convert `&ConfigContext` to `EvaluationContext`, as required.

crates/external_services/src/superposition/superposition.rs
```diff
@@ -0,0 +1,255 @@
...
+                "Superposition feature is not enabled".to_string(),
+            ))
+        }
+    }
+}
+
+/// Validation for configuration
+impl SuperpositionClientConfig {
+    /// Validate the Superposition configuration
+    pub fn validate(&self) -> Result<(), SuperpositionError> {
```
SanchithHegde: What about the `token` field, shouldn't we be checking that as well?

crates/external_services/src/superposition/superposition.rs
SanchithHegde, line 1: Also, what's with this module inception? Why is both the folder named `superposition` and the file also named the same? 

Can't we move all this logic to `crates/external_services/src/superposition/mod.rs`, which would then be renamed as `crates/external_services/src/superposition.rs`?

SanchithHegde:
Also, please address the failing CI checks, and update the PR description with testing information and relevant screenshots (if applicable).

config/config.example.toml
```diff
@@ -1265,3 +1265,12 @@ encryption_key = ""                            # Key to encrypt and decrypt chat
...
 connector_list = "worldpayvantiv"
+
+[superposition]
+enabled = false                                  # Enable or disable Superposition integration (default: false)
+endpoint = "http://superposition:8080"          # Superposition service endpoint URL
+token = "your_superposition_token"              # Superposition token
+org_id = "your_org_id"                          # Organization ID in Superposition
+workspace_id = "your_workspace_id"              # Workspace ID in Superposition
+polling_interval = 15                           # Polling interval in seconds for configuration updates
+# request_timeout =                              # Request timeout in seconds for Superposition API calls (optional, default: none)
```
SanchithHegde: Nit: You can keep this uncommented in this file, it's okay.

config/development.toml
```diff
@@ -1376,3 +1376,12 @@ encryption_key = ""
 proxy_connector_http_status_code = false    # If enabled, the http status code of the connector will be proxied in the response
 [list_dispute_supported_connectors]
 connector_list = "worldpayvantiv"
+
+[superposition]
+enabled = true # Enable Superposition integration
```
SanchithHegde: https://github.com/juspay/hyperswitch/pull/9289#discussion_r2328432566

> Maybe consider disabling by default, since this would fail unless everyone running hyperswitch also runs superposition beforehand.
> 
> * If everyone is _always_ expected to run superposition beforehand, then you'd have to update the local setup guide in `docs/try_local_system.md`.

config/development.toml
```diff
@@ -1376,3 +1376,12 @@ encryption_key = ""
...
+endpoint = "http://superposition:8080" # Superposition service endpoint
```
SanchithHegde: https://github.com/juspay/hyperswitch/pull/9289#discussion_r2328432566

> For the endpoint, you can use a URL like `localhost:<some port>`.

config/docker_compose.toml
```diff
@@ -1271,3 +1271,12 @@ version = "HOSTNAME"        # value of HOSTNAME from deployment which tells its
 proxy_connector_http_status_code = false    # If enabled, the http status code of the connector will be proxied in the response
 [list_dispute_supported_connectors]
 connector_list = "worldpayvantiv"
+
+[superposition]
+enabled = false # Enable Superposition for dynamic configuration management
+endpoint = "http://superposition:8080" # Superposition service endpoint
```
SanchithHegde, line 1301: Are we not planning to include superposition service(s) in Docker Compose?

crates/external_services/src/superposition.rs
```diff
@@ -0,0 +1,365 @@
+//! Superposition client for dynamic configuration management
+
+use std::collections::HashMap;
+
+use common_utils::errors::CustomResult;
+use error_stack::ResultExt;
+use masking::{ExposeInterface, Secret};
+use open_feature::{EvaluationContext, EvaluationContextFieldValue, StructValue, Value as OFValue};
+use serde_json::{Map, Number, Value};
```
SanchithHegde: Nit: Prefer qualifying usages of these types instead of importing them.

crates/external_services/src/superposition.rs
```diff
@@ -0,0 +1,365 @@
...
+    /// Organization ID in Superposition
+    pub org_id: String,
+    /// Workspace ID in Superposition
+    pub workspace_id: String,
+    /// Polling interval in seconds for configuration updates
+    #[serde(default = "default_polling_interval")]
+    pub polling_interval: u64,
+    /// Request timeout in seconds for Superposition API calls (None = no timeout)
+    #[serde(default = "default_request_timeout")]
+    pub request_timeout: Option<u64>,
```
SanchithHegde: I think you can:
- Remove the `serde(default)` attributes on these two fields, and add the `#[serde(default)]` attribute on the struct instead.
- Remove the `default_polling_interval()` and `default_request_timeout()` functions, and move the default values to the `Default` trait implementation alone.

crates/external_services/src/superposition.rs
```diff
@@ -0,0 +1,365 @@
...
+    /// Create a new empty context
+    pub fn new() -> Self {
+        Self::default()
+    }
+
+    /// Add a key-value pair to the context
+    pub fn with(mut self, key: &str, value: &str) -> Self {
+        self.values.insert(key.to_string(), value.to_string());
+        self
+    }
```
SanchithHegde: We can also document that this method would replace the value if it exists, since that's the behavior of `HashMap::insert()` method.

crates/external_services/src/superposition.rs
```diff
@@ -0,0 +1,365 @@
...
+}
+
+impl std::fmt::Debug for SuperpositionClient {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("SuperpositionClient")
+            .field("org_id", &self.org_id)
+            .field("workspace_id", &self.workspace_id)
+            .finish_non_exhaustive()
+    }
+}
```
SanchithHegde: Do we necessarily need this `Debug` implementation for somewhere?

The downside with writing manually implementing this is that if new fields are added (on a later date), we may forget to include those fields in this implementation.

crates/external_services/src/superposition.rs
```diff
@@ -0,0 +1,365 @@
...
+        hyperswitch_interfaces::secrets_interface::secret_state::SecretStateContainer<
+            Self,
+            hyperswitch_interfaces::secrets_interface::secret_state::RawSecret,
+        >,
+        hyperswitch_interfaces::secrets_interface::SecretsManagementError,
+    > {
+        let superposition_config = value.get_inner();
+        let token = secret_management_client
+            .get_secret(superposition_config.token.clone())
+            .await?;
```
SanchithHegde: You would have to perform the KMS decryption only if superposition is enabled. Otherwise, this would try to decrypt an empty value, and AWS KMS decryption fails with an empty string.

crates/external_services/src/superposition.rs
```diff
@@ -0,0 +1,365 @@
...
+    }
+}
+
+impl SuperpositionClient {
+    /// Create a new Superposition client
+    pub async fn new(config: SuperpositionClientConfig) -> CustomResult<Self, SuperpositionError> {
+        if !config.enabled {
+            return Err(SuperpositionError::InvalidConfiguration(
+                "Superposition is not enabled".to_string(),
+            )
```
SanchithHegde: I think we should have a separate enum variant indicating that superposition is not enabled, so that users of this method can match on the enum variants and handle the specific enum variant suitably.

crates/external_services/src/superposition.rs
```diff
@@ -0,0 +1,365 @@
...
+    fn build_evaluation_context(&self, context: Option<&ConfigContext>) -> EvaluationContext {
+        EvaluationContext {
+            custom_fields: if let Some(ctx) = context {
+                ctx.values
+                    .iter()
+                    .map(|(k, v)| (k.clone(), EvaluationContextFieldValue::String(v.clone())))
+                    .collect()
+            } else {
+                HashMap::new()
+            },
```
SanchithHegde: I think we can use `map_or()` instead of `if let Some`.

crates/external_services/src/superposition.rs
```diff
@@ -0,0 +1,365 @@
...
+    #[error("Superposition client error: {0}")]
+    ClientError(String),
+    /// Invalid configuration provided
+    #[error("Invalid configuration: {0}")]
+    InvalidConfiguration(String),
+}
+
+/// Context for configuration requests
+#[derive(Debug, Clone, Default)]
+pub struct ConfigContext {
```
SanchithHegde: Do we have solid reasons to introduce the `ConfigContext` and `JsonValue` wrapper types where it's helping improve the interface? I don't see them being used much outside of this file...

crates/external_services/src/superposition.rs
```diff
@@ -0,0 +1,365 @@
...
+            .map_err(|e| {
+                SuperpositionError::ClientError(format!(
+                    "Failed to get bool value for key '{}': {:?}",
+                    key, e
+                ))
+            })
+            .change_context(SuperpositionError::ClientError(format!(
+                "Failed to retrieve bool config for key: {}",
+                key
+            )))
```
SanchithHegde: You need not use both `map_err()` and `change_context()`, only `change_context()` is sufficient here. Also, use `change_context_lazy()` instead, since you're having an allocation by calling `format!()`.

crates/external_services/src/superposition.rs
```diff
@@ -0,0 +1,365 @@
...
+            .map_err(|e| {
+                SuperpositionError::ClientError(format!(
+                    "Failed to get string value for key '{}': {:?}",
+                    key, e
+                ))
+            })
+            .change_context(SuperpositionError::ClientError(format!(
+                "Failed to retrieve string config for key: {}",
+                key
+            )))
```
SanchithHegde: Same here, and elsewhere: use `change_context_lazy()` alone.

crates/router/src/routes/app.rs
```diff
@@ -425,6 +431,17 @@ impl AppState {
             let grpc_client = conf.grpc_client.get_grpc_client_interface().await;
             let infra_component_values = Self::process_env_mappings(conf.infra_values.clone());
             let enhancement = conf.enhancement.clone();
+            #[cfg(feature = "superposition")]
+            let superposition_service = if conf.superposition.get_inner().enabled {
+                Some(Arc::new(
+                    #[allow(clippy::expect_used)]
+                    SuperpositionClient::new(conf.superposition.get_inner().clone())
+                        .await
+                        .expect("Failed to create superposition client"),
```
SanchithHegde: Do we necessarily need the `expect()` here?

crates/router/Cargo.toml
```diff
@@ -38,6 +38,7 @@ v1 = ["common_default", "api_models/v1", "diesel_models/v1", "hyperswitch_domain
 dynamic_routing = ["external_services/dynamic_routing", "storage_impl/dynamic_routing", "api_models/dynamic_routing"]
 revenue_recovery = ["api_models/revenue_recovery", "hyperswitch_interfaces/revenue_recovery", "hyperswitch_domain_models/revenue_recovery", "hyperswitch_connectors/revenue_recovery", "external_services/revenue_recovery", "dep:prost-types"]
 tokenization_v2 = ["api_models/tokenization_v2", "diesel_models/tokenization_v2", "hyperswitch_domain_models/tokenization_v2", "storage_impl/tokenization_v2"]
+superposition = ["external_services/superposition"]
```
SanchithHegde: This feature won't be enabled on images deployed in our environments unless you include this in the `release` feature set. But then this would be enabled in images we release as part of future stable releases as well.

crates/router/src/core/configs.rs
```diff
@@ -50,3 +51,163 @@ pub async fn config_delete(state: SessionState, key: String) -> RouterResponse<a
...
+/// Get a boolean configuration value with Superposition -> Database fallback
+pub async fn get_config_bool(
+    state: &SessionState,
+    #[cfg(feature = "superposition")] superposition_key: &str,
+    #[cfg(not(feature = "superposition"))] _superposition_key: &str,
+    db_key: &str,
+    #[cfg(feature = "superposition")] context: Option<std::collections::HashMap<String, String>>,
+    #[cfg(not(feature = "superposition"))] _context: Option<
+        std::collections::HashMap<String, String>,
+    >,
```
SanchithHegde: Should we consider using an enum instead, which could be named something like `ConfigRetrieveContext`, instead of having such parameters behind features flags?

```rust
enum ConfigRetrieveContext {
    DatabaseOnly {
        key: String
    },
    #[cfg(feature = "superposition")]
    SuperpositionWithDatabaseFallback {
        superposition_key: String,
        superposition_context: Option<HashMap<String, String>>,
        database_key: String,
    },
    #[cfg(feature = "superposition")]
    SuperpositionOnly {
        key: String,
        context: Option<HashMap<String, String>>,
    }
}
```

Something like this would allow us to eventually deprecate the database fallback logic, and also allowing new configs to be retrieved from superposition alone (if required).

@sai-harsha-vardhan What do you say?

crates/router/src/core/configs.rs
```diff
@@ -50,3 +51,163 @@ pub async fn config_delete(state: SessionState, key: String) -> RouterResponse<a
...
+    default_value: bool,
+) -> CustomResult<bool, errors::StorageError> {
+    #[cfg(feature = "superposition")]
+    if let Some(ref superposition_client) = state.superposition_service {
+        if let Ok(value) = superposition_client
+            .get_bool_value(superposition_key, context.as_ref())
+            .await
+        {
+            return Ok(value);
+        }
```
SanchithHegde: By using `if let Ok`, we're ignoring the error here, and won't even know it occurred.

Based on our requirements, we should consider either propagating it, or logging it and falling back to reading it from the database.

crates/router/src/core/configs.rs
```diff
@@ -50,3 +51,163 @@ pub async fn config_delete(state: SessionState, key: String) -> RouterResponse<a
...
+        .await?;
+
+    config
+        .config
+        .parse::<bool>()
+        .change_context(errors::StorageError::DeserializationFailed)
+}
+
+/// Get a string configuration value with Superposition -> Database fallback
+pub async fn get_config_string(
```
SanchithHegde, line 102: Same comments as `get_config_bool()` applies here and for the rest of the functions defined here.

crates/external_services/src/superposition.rs
```diff
@@ -0,0 +1,365 @@
...
+}
+
+/// Default request timeout (None means no timeout)
+const fn default_request_timeout() -> Option<u64> {
+    None
+}
+
+/// Configuration for Superposition integration
+#[derive(Debug, Clone, serde::Deserialize)]
+pub struct SuperpositionClientConfig {
```
SanchithHegde: Did you remove the `validate()` method that was previously implemented?

config/docker_compose.toml
Shailesh-714: Superposition is not completely implemented and tested. Until that its better if we dont include it as a service available for external use

crates/external_services/src/superposition.rs
Shailesh-714: ConfigContext will be  replaced inplace of hashmap. For JsonValue wrapper, it is mandatory due to Rust's orphan rule and OpenFeature's `get_struct_value<T>()` API automatically calls `T::try_from(struct_value)` where T must implement `TryFrom<StructValue>`

crates/router/src/routes/app.rs
Shailesh-714: I think expect is necessary unless we want the application not to panic even when the feature flag and superposition is enabled. This may be necessary when we make superposition the only config service without db fallback later

crates/router/Cargo.toml
Shailesh-714: I think we can add this to the release feature set once the application changes are done and tested.

crates/router/src/core/configs.rs
Shailesh-714: Is this necessary? as we will remove these once we remove the db fallback. Also trying to implement this only increase complexity of code which may have to be removed later too.

crates/router/src/routes/app.rs
SanchithHegde: Can't we choose to call `expect()` and have the application panic only when the superposition feature has been stabilized?

crates/router/src/core/configs.rs
```diff
@@ -50,3 +51,163 @@ pub async fn config_delete(state: SessionState, key: String) -> RouterResponse<a
...
+    state: &SessionState,
+    #[cfg(feature = "superposition")] superposition_key: &str,
+    #[cfg(not(feature = "superposition"))] _superposition_key: &str,
+    db_key: &str,
+    #[cfg(feature = "superposition")] context: Option<std::collections::HashMap<String, String>>,
+    #[cfg(not(feature = "superposition"))] _context: Option<
+        std::collections::HashMap<String, String>,
+    >,
+    default_value: serde_json::Value,
+) -> CustomResult<serde_json::Value, errors::StorageError> {
```
SanchithHegde: We can see if there's an option to return a generic type `T` after deserialization here, instead of `serde_json::Value`.

crates/external_services/src/superposition.rs
Shailesh-714: `Result<bool, EvaluationError>` doesn't implement trait `error_stack::Context` to use `change_context_lazy()`. So its neccessary to use `map_err` to convert it to superposition error to use change_context.
```rust
.map_err(|e| SuperpositionError::ClientError(format!("OpenFeature error: {e:?}")))
.change_context_lazy(|| SuperpositionError::ClientError(format!(
    "Failed to retrieve object config for key: {}",
    key
)))?;
```

crates/external_services/src/superposition.rs
Shailesh-714: The `open_feature::Client` doesn't implement `Debug`, and if we remove the manual debug implementation, we get warnings during compilation

crates/external_services/src/superposition.rs
```diff
@@ -0,0 +1,353 @@
...
+
+    /// Add a key-value pair to the context. Replaces existing value if key exists.
+    pub fn with(mut self, key: &str, value: &str) -> Self {
+        self.values.insert(key.to_string(), value.to_string());
+        self
+    }
+}
+
+/// Superposition client wrapper
+#[allow(missing_debug_implementations)]
```
SanchithHegde, line 34: Can we have a comment stating why we're allowing this lint?

crates/external_services/src/lib.rs
```diff
@@ -12,6 +12,8 @@ pub mod file_storage;
 #[cfg(feature = "hashicorp-vault")]
 pub mod hashicorp_vault;
 
+pub mod superposition;
```
SanchithHegde: Nit: Remove the newlines surrounding this module declaration, then run formatter.

crates/router/Cargo.toml
SanchithHegde: Remove this feature and directly enable `external_services/superposition` under `common_default`. This would help us reduce one `cargo hack` feature combination on CI.

SanchithHegde (approved):

docker-compose-development.yml
```diff
@@ -57,14 +57,23 @@ services:
       # format -> postgresql://DB_USER:DB_PASSWORD@HOST:PORT/DATABASE_NAME
       - DATABASE_URL=postgresql://db_user:db_pass@pg:5432/hyperswitch_db
 
+  superposition:
```
SanchithHegde, line 67: Nit: Can we move this behind a profile for now?

SanchithHegde (dismissed):

crates/external_services/src/superposition.rs
```diff
@@ -0,0 +1,347 @@
+//! Superposition client for dynamic configuration management
+
+use std::collections::HashMap;
+
+use common_utils::errors::CustomResult;
+use error_stack::report;
+use masking::{ExposeInterface, Secret};
+use superposition_provider::{
+    PollingStrategy, RefreshStrategy, SuperpositionProvider, SuperpositionProviderOptions,
```
sai-harsha-vardhan: try to avoid absolute imports at top level

config/config.example.toml
```diff
@@ -1270,3 +1270,12 @@ encryption_key = ""                            # Key to encrypt and decrypt chat
 proxy_connector_http_status_code = false    # If enabled, the http status code of the connector will be proxied in the response
 [list_dispute_supported_connectors]
 connector_list = "worldpayvantiv"
+
+[superposition]
+enabled = false                                  # Enable or disable Superposition integration (default: false)
+endpoint = "http://superposition:8080"          # Superposition service endpoint URL
```
sai-harsha-vardhan, line 1289: Should this be 8081?

crates/external_services/src/superposition.rs
```diff
@@ -0,0 +1,347 @@
...
+}
+
+/// Configuration for Superposition integration
+#[derive(Debug, Clone, serde::Deserialize)]
+#[serde(default)]
+pub struct SuperpositionClientConfig {
+    /// Whether Superposition is enabled
+    pub enabled: bool,
+    /// Superposition API endpoint
+    pub endpoint: String,
```
sai-harsha-vardhan: Can we use `URL` here?

crates/external_services/src/superposition.rs
```diff
@@ -0,0 +1,347 @@
...
+    /// Authentication token for Superposition
+    pub token: Secret<String>,
+    /// Organization ID in Superposition
+    pub org_id: String,
```
sai-harsha-vardhan: nit: rename `organization_id`

crates/external_services/src/superposition.rs
```diff
@@ -0,0 +1,347 @@
...
+        )),
+    }
+}
+
+/// Configuration for Superposition integration
+#[derive(Debug, Clone, serde::Deserialize)]
+#[serde(default)]
+pub struct SuperpositionClientConfig {
+    /// Whether Superposition is enabled
+    pub enabled: bool,
```
sai-harsha-vardhan: use domain type here

crates/external_services/src/superposition.rs
```diff
@@ -0,0 +1,347 @@
...
+                .map(convert_open_feature_value)
+                .collect::<Result<Vec<_>, _>>()?,
+        )),
+    }
+}
+
+/// Configuration for Superposition integration
+#[derive(Debug, Clone, serde::Deserialize)]
+#[serde(default)]
+pub struct SuperpositionClientConfig {
```
sai-harsha-vardhan: Can we move all these definitions to types mod for better seggregation?

crates/external_services/src/superposition.rs
```diff
@@ -0,0 +1,347 @@
...
+pub struct ConfigContext {
+    /// Key-value pairs for configuration context
+    values: HashMap<String, String>,
+}
+
+impl SuperpositionClientConfig {
+    /// Validate the Superposition configuration
+    pub fn validate(&self) -> Result<(), SuperpositionError> {
+        if self.enabled {
+            if self.endpoint.is_empty() {
```
sai-harsha-vardhan: we can use `when` from common_utils to avoid early returns?

crates/external_services/src/superposition.rs
```diff
@@ -0,0 +1,347 @@
...
+// Debug trait cannot be derived because open_feature::Client doesn't implement Debug
+#[allow(missing_debug_implementations)]
+pub struct SuperpositionClient {
+    client: open_feature::Client,
+}
+
+impl SuperpositionClient {
+    /// Create a new Superposition client
+    pub async fn new(config: SuperpositionClientConfig) -> CustomResult<Self, SuperpositionError> {
+        if !config.enabled {
```
sai-harsha-vardhan: Is this required because we are calling this only when the config is enabled right?

crates/external_services/src/superposition.rs
```diff
@@ -0,0 +1,347 @@
...
+            return Err(SuperpositionError::NotEnabled.into());
+        }
+
+        let provider_options = SuperpositionProviderOptions {
```
sai-harsha-vardhan: use `from` impl

crates/external_services/src/superposition.rs
```diff
@@ -0,0 +1,347 @@
...
+        // Create provider and set up OpenFeature
+        let provider = SuperpositionProvider::new(provider_options);
+
+        router_env::logger::info!("Created superposition provider");
+
+        // Initialize OpenFeature API and set provider
+        let mut api = open_feature::OpenFeature::singleton_mut().await;
+        api.set_provider(provider).await;
+
+        router_env::logger::info!("Set superposition provider, creating client");
```
sai-harsha-vardhan: Do we need the explicit loggers after each step?

crates/external_services/src/superposition.rs
```diff
@@ -0,0 +1,347 @@
...
+        &self,
+        key: &str,
+        context: Option<&ConfigContext>,
+    ) -> CustomResult<i64, SuperpositionError> {
+        let evaluation_context = self.build_evaluation_context(context);
+
+        self.client
+            .get_int_value(key, Some(&evaluation_context), None)
+            .await
+            .map_err(|e| {
```
sai-harsha-vardhan, line 141: Can't we change_context here?

crates/external_services/src/superposition.rs
```diff
@@ -0,0 +1,188 @@
...
+/// Type definitions for Superposition integration
+pub mod types;
+
+use std::collections::HashMap;
+
+use common_utils::errors::CustomResult;
+use error_stack::report;
+use masking::ExposeInterface;
+use superposition_provider;
+pub use types::*;
```
SanchithHegde: Avoid wildcard imports.

crates/external_services/src/superposition/types.rs
```diff
@@ -0,0 +1,162 @@
...
+    pub request_timeout: Option<u64>,
+}
+
+impl Default for SuperpositionClientConfig {
+    fn default() -> Self {
+        Self {
+            enabled: false,
+            #[allow(clippy::expect_used)]
+            endpoint: url::Url::parse("http://localhost:8080")
+                .expect("Failed to parse default Superposition URL"),
```
SanchithHegde: I'd prefer `expect`s to be not present in library code.

crates/external_services/src/superposition/types.rs
```diff
@@ -0,0 +1,162 @@
...
+    /// Invalid configuration provided
+    #[error("Invalid configuration: {0}")]
+    InvalidConfiguration(String),
+}
+
+/// Context for configuration requests
+#[derive(Debug, Clone, Default)]
+pub struct ConfigContext {
+    /// Key-value pairs for configuration context
+    pub(crate) values: HashMap<String, String>,
```
SanchithHegde: Can't we keep this field private? Or `pub(super)` at most?

crates/external_services/src/superposition.rs
```diff
@@ -0,0 +1,188 @@
...
+
+pub(crate) fn convert_open_feature_value(
```
SanchithHegde: Can't we have this function as private?

crates/external_services/src/superposition.rs
```diff
@@ -0,0 +1,187 @@
...
+
+use std::collections::HashMap;
+
+use common_utils::errors::CustomResult;
+use error_stack::report;
+use masking::ExposeInterface;
+use superposition_provider;
+
+fn convert_open_feature_value(
+    v: open_feature::Value,
```
sai-harsha-vardhan: nit: rename

crates/external_services/src/superposition/types.rs
```diff
@@ -0,0 +1,172 @@
+//! Type definitions for Superposition integration
+
+use std::collections::HashMap;
+
+use common_utils::{errors::CustomResult, fp_utils::when};
+use masking::{ExposeInterface, Secret};
+
+/// Wrapper type for JSON values from Superposition
+#[derive(Debug, Clone)]
+pub struct JsonValue(pub(crate) serde_json::Value);
```
sai-harsha-vardhan: nit: is `pub(crate)` relevant here?

crates/router/src/core/configs.rs
```diff
@@ -50,3 +52,180 @@ pub async fn config_delete(state: SessionState, key: String) -> RouterResponse<a
...
+    db_key: &str,
+    context: Option<ConfigContext>,
+    default_value: bool,
+) -> CustomResult<bool, errors::StorageError> {
+    if let Some(ref superposition_client) = state.superposition_service {
+        match superposition_client
+            .get_bool_value(superposition_key, context.as_ref())
+            .await
+        {
+            Ok(value) => return Ok(value),
```
sai-harsha-vardhan: Please avoid early returns

crates/router/src/core/configs.rs
```diff
@@ -126,30 +145,41 @@ pub async fn get_config_int(
...
             .await
         {
-            Ok(value) => return Ok(value),
+            Ok(value) => Ok(value),
             Err(err) => {
                 router_env::logger::warn!(
                     "Failed to retrieve config from superposition, falling back to database: {:?}",
                     err
                 );
+                let config = state
```
sai-harsha-vardhan: can we avoid this duplication?

crates/router/src/core/configs.rs
```diff
@@ -50,3 +52,227 @@ pub async fn config_delete(state: SessionState, key: String) -> RouterResponse<a
...
+/// Get a boolean configuration value with superposition and database fallback
+pub async fn get_config_bool(
+    state: &SessionState,
+    superposition_key: &str,
+    db_key: &str,
+    context: Option<ConfigContext>,
+    default_value: bool,
+) -> CustomResult<bool, errors::StorageError> {
+    // Try superposition first if available
+    let mut superposition_result = None;
```
sai-harsha-vardhan: we should try to avoid mutable values as much as possible

PR #9378, opened 2025-09-15
fix(connectors): [Nuvei] Webhooks

AkshayaFoiger:
## Type of Change

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

Fix Nuvei payments and chargeback webhooks

## Checklist

### DIFF
--- a/crates/hyperswitch_connectors/src/connectors/nuvei.rs
+++ b/crates/hyperswitch_connectors/src/connectors/nuvei.rs
@@ -546,13 +546,14 @@ impl ConnectorIntegration<PSync, PaymentsSyncData, PaymentsResponseData> for Nuv
         event_builder: Option<&mut ConnectorEvent>,
         res: Response,
     ) -> CustomResult<PaymentsSyncRouterData, errors::ConnectorError> {
-        let response: nuvei::NuveiPaymentsResponse = res
+        let response: nuvei::NuveiPaymentSyncResponse = res
             .response
             .parse_struct("NuveiPaymentsResponse")
             .switch()?;
 
         event_builder.map(|i| i.set_response_body(&response));
         router_env::logger::info!(connector_response=?response);
+        let nuvie_common_response = NuveiPaymentsResponse::try_from(response)?;
 
         RouterData::try_from(ResponseRouterData {
             response,
@@ -1005,7 +1006,18 @@ impl IncomingWebhook for Nuvei {
         request: &IncomingWebhookRequestDetails<'_>,
         _connector_webhook_secrets: &api_models::webhooks::ConnectorWebhookSecrets,
     ) -> CustomResult<Vec<u8>, errors::ConnectorError> {
-        let signature = utils::get_header_key_value("advanceResponseChecksum", request.headers)?;
+        let webhook = serde_urlencoded::from_bytes::<nuvei::NuveiWebhook>(&request.body)
+        .change_context(errors::ConnectorError::WebhookBodyDecodingFailed)?;
+
+
+        let signature = match webhook {
+            nuvei::NuveiWebhook::PaymentDmn(notification) => {
+                notification.advance_response_checksum.ok_or(errors::ConnectorError::WebhookSignatureNotFound)?
+            },
+            nuvei::NuveiWebhook::Chargeback(notification) => {
+                Err(errors::ConnectorError::WebhookResponseEncodingFailed)?
+            }
+        };
         hex::decode(signature).change_context(errors::ConnectorError::WebhookResponseEncodingFailed)
     }
 
@@ -1016,8 +1028,9 @@ impl IncomingWebhook for Nuvei {
         connector_webhook_secrets: &api_models::webhooks::ConnectorWebhookSecrets,
     ) -> CustomResult<Vec<u8>, errors::ConnectorError> {
         // Parse the webhook payload
-        let webhook = serde_urlencoded::from_str::<nuvei::NuveiWebhook>(&request.query_params)
-            .change_context(errors::ConnectorError::WebhookBodyDecodingFailed)?;
+        let webhook = serde_urlencoded::from_bytes::<nuvei::NuveiWebhook>(&request.body)
+        .change_context(errors::ConnectorError::WebhookBodyDecodingFailed)?;
+    router_env::logger::info!(sssssssssss=?webhook);
 
         let secret_str = std::str::from_utf8(&connector_webhook_secrets.secret)
             .change_context(errors::ConnectorError::WebhookBodyDecodingFailed)?;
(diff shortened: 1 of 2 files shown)


crates/hyperswitch_connectors/src/connectors/nuvei.rs
```diff
@@ -62,11 +68,17 @@ use crate::{
 #[derive(Clone)]
 pub struct Nuvei {
     pub amount_convertor: &'static (dyn AmountConvertor<Output = StringMajorUnit> + Sync),
+    amount_converter_string_minor_unit:
```
Nithin1506200, line 71: this `amount_converter_string_minor_unit` is same as `amount_convertor` .

crates/hyperswitch_connectors/src/connectors/nuvei.rs
```diff
@@ -546,13 +558,14 @@ impl ConnectorIntegration<PSync, PaymentsSyncData, PaymentsResponseData> for Nuv
         event_builder: Option<&mut ConnectorEvent>,
         res: Response,
     ) -> CustomResult<PaymentsSyncRouterData, errors::ConnectorError> {
-        let response: NuveiTransactionSyncResponse = res
+        let nuvie_psync_common_response: nuvei::NuveiPaymentSyncResponse = res
             .response
             .parse_struct("NuveiTransactionSyncResponse")
```
Nithin1506200: rename `NuveiTransactionSyncResponse` to `NuveiPaymentSyncResponse`

crates/hyperswitch_connectors/src/connectors/nuvei.rs
```diff
@@ -1025,38 +1070,29 @@ impl IncomingWebhook for Nuvei {
...
-                    notification.currency.unwrap_or_default(),
-                    notification.response_time_stamp.unwrap_or_default(),
-                    notification.ppp_transaction_id.unwrap_or_default(),
+                    notification.total_amount,
+                    notification.currency,
+                    notification.response_time_stamp,
+                    notification.ppp_transaction_id,
                     status,
-                    notification.product_id.unwrap_or_default(),
+                    notification.product_id.unwrap_or("NA".to_string()),
```
Nithin1506200, line 1085: are we sure we need to add `NA` if found null

crates/hyperswitch_connectors/src/connectors/nuvei/transformers.rs
```diff
@@ -1704,7 +1717,7 @@ where
 
             ..Default::default()
         })?;
-        let return_url = item.request.get_return_url_required()?;
+        let return_url = "https://baf141b098ed.ngrok-free.app".to_string(); // item.request.get_return_url_required()?;
```
Nithin1506200: revert

crates/hyperswitch_connectors/src/connectors/nuvei/transformers.rs
```diff
@@ -3449,3 +3716,195 @@ fn convert_to_additional_payment_method_connector_response(
...
+) -> Result<api_models::webhooks::IncomingWebhookEvent, error_stack::Report<errors::ConnectorError>>
+{
+    match (status, transaction_type) {
+        (
+            DmnStatus::Success | DmnStatus::Approved,
+            NuveiTransactionType::Auth | NuveiTransactionType::Auth3D,
+        ) => Ok(api_models::webhooks::IncomingWebhookEvent::PaymentIntentAuthorizationSuccess),
+        (
+            DmnStatus::Success | DmnStatus::Approved,
+            NuveiTransactionType::Sale | NuveiTransactionType::InitAuth3D,
```
Nithin1506200: InitAuth3d is preprocessing for 3ds. We cannot mark PaymentIntentSuccess here.

crates/hyperswitch_connectors/src/connectors/nuvei/transformers.rs
```diff
@@ -3449,3 +3716,195 @@ fn convert_to_additional_payment_method_connector_response(
...
+
+        DisputeUnifiedStatusCode::McCollaborationPreviouslyRefundedAuto
+        | DisputeUnifiedStatusCode::McCollaborationRejectedByMerchant
+        | DisputeUnifiedStatusCode::InquiryAutomaticallyRejected
+        | DisputeUnifiedStatusCode::InquiryPartialAcceptedPartialRefund
+        | DisputeUnifiedStatusCode::InquiryUpdated => {
+            Err(errors::ConnectorError::WebhookEventTypeNotFound.into())
+        }
+    }
+}
```
Nithin1506200, line 3868: Got gemini suggestion for some mappings , Ignore if correct

```md
Potential Errors and Inconsistencies

There are a few mappings that appear to be incorrect or questionable.

1. Clear Error in DisputeLost Mapping

There seems to be a direct contradiction in the DisputeLost block:

    DisputeUnifiedStatusCode::McCollaborationClosedMerchantFavour is mapped to DisputeLost.

This status explicitly states the outcome is in the merchant's favor, which means it should be mapped to DisputeWon, not DisputeLost.

2. Questionable Unhandled Events

Some status codes that are currently unhandled (returning an Err) seem like they should map to an existing event.

    DisputeUnifiedStatusCode::McCollaborationRejectedByMerchant

    DisputeUnifiedStatusCode::InquiryAutomaticallyRejected

Both of these statuses indicate a "rejection" from the merchant's side. This action is functionally equivalent to the statuses in the DisputeChallenged block (e.g., FirstChargebackRejectedByMerchant). It is inconsistent to treat these rejections as unhandled errors while others are mapped to DisputeChallenged.

3. Ambiguous DisputeLost Case

The following mapping is functionally plausible but could be debated depending on business logic:

    DisputeUnifiedStatusCode::FirstChargebackPartiallyAcceptedByMerchant is mapped to DisputeLost.

While the merchant lost some money (making it a partial loss), they also agreed to it (making it a partial acceptance). Mapping it to DisputeAccepted could also be considered correct. This choice depends on whether the business defines a "lost" dispute as any case where the merchant does not retain the full transaction amount.
```

crates/hyperswitch_connectors/src/connectors/nuvei/transformers.rs
```diff
@@ -3449,3 +3716,195 @@ fn convert_to_additional_payment_method_connector_response(
...
+            | DisputeUnifiedStatusCode::PreArbitrationRejectedByMerchantExpired
+            | DisputeUnifiedStatusCode::PreArbitrationAutomaticallyAcceptedByMerchant
+            | DisputeUnifiedStatusCode::PreArbitrationClosedRecall
+            | DisputeUnifiedStatusCode::RejectedPreArbAcceptedByMerchant
+            | DisputeUnifiedStatusCode::RejectedPreArbExpiredAutoAccepted => Self::PreArbitration,
+
+            // --- DisputeReversal ---
+            DisputeUnifiedStatusCode::CreditChargebackRecalledByIssuer => Self::DisputeReversal,
+        }
+    }
```
Nithin1506200, line 3909: Double-check if `FirstChargebackRecalledByIssuer` and `PreArbitrationClosedRecall` should also be mapped to the DisputeReversal stage for consistency.

crates/hyperswitch_connectors/src/connectors/nuvei.rs
AkshayaFoiger: Yes, we are sure about this

crates/hyperswitch_connectors/src/connectors/nuvei.rs
AkshayaFoiger: `amount_convertor` is used for converting minor unit to string major unit and vice versa. But we are using `amount_converter_string_minor_unit` to convert **minor unit for string minor unit**

crates/hyperswitch_connectors/src/connectors/nuvei/transformers.rs
AkshayaFoiger: Fixed the mapping for McCollaborationClosedMerchantFavour. However, McCollaborationRejectedByMerchant and InquiryAutomaticallyRejected should not be mapped to DisputeChallenged. Rejecting collaboration isn’t the same as a true merchant loss, since the merchant never contested it. It’s also not a win, because the case wasn’t adjudicated in the merchant’s favour — it was simply blocked

deepanshu-iiitu (approved):

PR #9388, opened 2025-09-15
feat(router): Add support for updating wallet pm_data to PM update API

### LINKED ISSUE #9387: Add support for updating wallet pm_data to PM update API
Currently PM update API only supports card. Need to add support for updating payment_method_data in case of wallet

AnuthaDev:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

- Added support for updating `payment_method_data` in case the `payment_method_type` is wallet.
- Added new response type: `CustomerPaymentMethodUpdateResponse` to include the wallet details being updated in response

For wallets the following fields can be updated:

- last4
- card_network
- type

### Additional Changes

## Motivation and Context

Closes #9387 

## How did you test it?

Request:

```
curl --location 'http://localhost:8080/payment_methods/pm_FiEUWFVVyrcdp6o2NW4T/update' \
--header 'api-key: dev_x85W6hQprLRMbeArLC7rFicIiZRBWelCJCNV8a4Lf2kq2tgCfoHw9LvMRrQ4tn8D' \
--data '{
  "wallet": {
    "last4": "1024",
    "card_network": "Visa",
    "type": "credit"
  }
}'
```

Response:
```json
{
    "merchant_id": "merchant_1759320450",
    "customer_id": "customer123",
    "payment_method_id": "pm_FiEUWFVVyrcdp6o2NW4T",
    "payment_method": "wallet",
    "payment_method_type": "apple_pay",
    "card": null,
    "wallet": {
      "last4": "1024",
      "card_network": "Visa",
      "type": "credit"
    }
    "recurring_enabled": false,
    "payment_experience": [
        "redirect_to_url"
    ],
    "metadata": null,
    "created": "2025-10-01T12:10:05.824Z",
    "last_used_at": "2025-10-06T08:37:30.787Z",
    "client_secret": "pm_FiEUWFVVyrcdp6o2NW4T_secret_9o1ejWTPIzRzn4WdfoGm"
}
```

## Checklist

### DIFF
--- a/crates/api_models/src/events/payment.rs
+++ b/crates/api_models/src/events/payment.rs
@@ -12,10 +12,10 @@ use crate::payment_methods::{
 };
 use crate::{
     payment_methods::{
-        self, CustomerPaymentMethodUpdateResponse, ListCountriesCurrenciesRequest,
-        ListCountriesCurrenciesResponse, PaymentMethodCollectLinkRenderRequest,
-        PaymentMethodCollectLinkRequest, PaymentMethodCollectLinkResponse,
-        PaymentMethodMigrateResponse, PaymentMethodResponse, PaymentMethodUpdate,
+        self, ListCountriesCurrenciesRequest, ListCountriesCurrenciesResponse,
+        PaymentMethodCollectLinkRenderRequest, PaymentMethodCollectLinkRequest,
+        PaymentMethodCollectLinkResponse, PaymentMethodMigrateResponse, PaymentMethodResponse,
+        PaymentMethodUpdate,
     },
     payments::{
         self, PaymentListConstraints, PaymentListFilters, PaymentListFiltersV2,
@@ -25,7 +25,9 @@ use crate::{
 };
 #[cfg(feature = "v1")]
 use crate::{
-    payment_methods::{PaymentMethodListRequest, PaymentMethodListResponse},
+    payment_methods::{
+        CustomerPaymentMethodUpdateResponse, PaymentMethodListRequest, PaymentMethodListResponse,
+    },
     payments::{
         ExtendedCardInfoResponse, PaymentIdType, PaymentListFilterConstraints,
         PaymentListResponseV2, PaymentsApproveRequest, PaymentsCancelPostCaptureRequest,
@@ -310,6 +312,7 @@ impl ApiEventMetric for PaymentMethodResponse {
     }
 }
 
+#[cfg(feature = "v1")]
 impl ApiEventMetric for CustomerPaymentMethodUpdateResponse {
     fn get_api_event_type(&self) -> Option<ApiEventsType> {
         Some(ApiEventsType::PaymentMethod {


crates/router/src/core/payment_methods/cards.rs
```diff
@@ -1844,6 +1844,79 @@ pub async fn update_customer_payment_method(
         };
 
         Ok(services::ApplicationResponse::Json(response))
+    } else if let Some(wallet_update) = req.wallet.clone() {
```
hrithikesh026, line 1873: else if will prevent wallet update is card update is present. Is that expected.

crates/router/src/core/payment_methods/cards.rs
```diff
@@ -1844,6 +1844,79 @@ pub async fn update_customer_payment_method(
...
+        )
+        .await
+        .change_context(errors::ApiErrorResponse::InternalServerError)
+        .attach_printable("Unable to encrypt payment method data")?;
+
+        let pm_update = storage::PaymentMethodUpdate::PaymentMethodDataUpdate {
+            payment_method_data: Some(pm_data_encrypted.into()),
+        };
+
+        db.update_payment_method(
```
hrithikesh026:
```suggestion
        let pm = db.update_payment_method(
```

crates/router/src/core/payment_methods/cards.rs
```diff
@@ -1844,6 +1844,79 @@ pub async fn update_customer_payment_method(
...
+            &((&state).into()),
+            merchant_context.get_merchant_key_store(),
+            pm.clone(),
```
hrithikesh026:
```suggestion
            pm,
```

crates/router/src/core/payment_methods/cards.rs
```diff
@@ -1844,6 +1844,80 @@ pub async fn update_customer_payment_method(
...
+            .await
+            .to_not_found_response(errors::ApiErrorResponse::PaymentMethodNotFound)?;
+
+        if let Some(cs) = &req.client_secret {
+            let is_client_secret_expired = authenticate_pm_client_secret_and_check_expiry(cs, &pm)?;
+
+            if is_client_secret_expired {
+                return Err((errors::ApiErrorResponse::ClientSecretExpired).into());
+            };
+        };
```
kashif-m: Can this step be common? I see it's done above for cards as well, so let's avoid duplicating the code

crates/router/src/core/payment_methods/cards.rs
```diff
@@ -1844,6 +1844,80 @@ pub async fn update_customer_payment_method(
...
+                &((&state).into()),
+                merchant_context.get_merchant_key_store(),
+                pm,
+                pm_update,
+                merchant_context.get_merchant_account().storage_scheme,
+            )
+            .await
+            .change_context(errors::ApiErrorResponse::InternalServerError)
+            .attach_printable("Failed to update payment method in db")?;
+
```
kashif-m, line 1907: Can all the changes be accommodated in a single update DB call?

kashif-m (changes requested):
Added minor comments around refactoring the flow

crates/router/src/core/payment_methods/cards.rs
AnuthaDev: Trying to have a single DB call for both card and wallet branches leads to convoluted code for response generation. It is cleaner to have 2 DB calls, contained within their branches.

crates/router/src/core/payment_methods/cards.rs
AnuthaDev: Done

crates/router/src/core/payment_methods/cards.rs
AnuthaDev: Yes. For one Payment Method, we can't have both card and wallet as payment method simultaneously.

crates/router/src/core/payment_methods/cards.rs
```diff
@@ -1629,28 +1629,28 @@ pub async fn update_customer_payment_method(
...
-            .to_not_found_response(errors::ApiErrorResponse::PaymentMethodNotFound)?;
+    let pm = db
+        .find_payment_method(
+            &((&state).into()),
+            merchant_context.get_merchant_key_store(),
+            payment_method_id,
+            merchant_context.get_merchant_account().storage_scheme,
+        )
+        .await
+        .to_not_found_response(errors::ApiErrorResponse::PaymentMethodNotFound)?;
```
kashif-m, line 1652: nitpick -

We can execute this DB query only when it's needed. Something like -

```
if req.card.is_some() || req.wallet.is_some() {
    // Perform DB query to fetch PM
    match (req.card, req.wallet) {
        (Some(card), None) => {
            // Update card details
        }
        (None, Some(wallet)) => {
            // Update wallet details
        }
        (Some(_), Some(_)) => {
            // Handle error: both card and wallet provided
        }
        (None, None) => {
            // will never happen due to the outer if condition
        }
    }
}
```

AnuthaDev: It can't be the case that both card and wallet are provided. Trying to do it this way increase response generation complexity.

su-shivanshmathur (approved):

crates/router/src/core/payment_methods/cards.rs
```diff
@@ -1637,29 +1637,29 @@ pub async fn update_customer_payment_method(
...
```
kashif-m, line 1651: nit: we can fetch this after validating the client secret?

crates/router/src/core/payment_methods/cards.rs
```diff
@@ -1852,6 +1870,61 @@ pub async fn update_customer_payment_method(
...
         Ok(services::ApplicationResponse::Json(response))
+    } else if let Some(wallet_update) = req.wallet.clone() {
+        if pm.payment_method != Some(common_enums::PaymentMethod::Wallet) {
+            return Err((errors::ApiErrorResponse::InvalidRequestData {
+                message: "The Payment Method is not wallet".to_string(),
+            })
+            .into());
+        }
+
+        let updated_pmd = PaymentMethodsData::WalletDetails(wallet_update);
```
kashif-m, line 1881: Suggestion:

Can we handle this similarly to how it's handled for cards [[ref](https://github.com/juspay/hyperswitch/pull/9388/files#diff-0cad9ec56cd7135aea31f7f8fb51ee0197f2125ebc87b716e9d023318361a827L1726)]

We should be able to update individual fields for wallet data (last4, card_type)

kashif-m (approved):

crates/api_models/src/payment_methods.rs
```diff
@@ -922,6 +925,70 @@ pub struct PaymentMethodResponse {
     pub client_secret: Option<String>,
 }
 
+#[cfg(feature = "v1")]
+#[derive(Debug, serde::Deserialize, serde::Serialize, ToSchema)]
+pub struct CustomerPaymentMethodUpdateResponse {
```
hrithikesh026, line 930: why can't er use the existing api response model?

crates/api_models/src/payment_methods.rs
```diff
@@ -481,6 +481,9 @@ pub struct PaymentMethodUpdate {
     "card_holder_name": "John Doe"}))]
     pub card: Option<CardDetailUpdate>,
```
hrithikesh026, line 482: PaymentMethodUpdate should be an enum. We shouldn't allow passing both card and wallet fields during update right?

crates/api_models/src/payment_methods.rs
AnuthaDev: Existing response struct is used in many other places as a vault response. Adding the wallet field does not make sense in those use cases, which is why I decided to have a different response struct for this API.

crates/router/src/core/payment_methods/cards.rs
```diff
@@ -1852,6 +1870,61 @@ pub async fn update_customer_payment_method(
         };
 
         Ok(services::ApplicationResponse::Json(response))
+    } else if let Some(wallet_update) = req.wallet.clone() {
```
hrithikesh026, line 1873: if both card and wallet is present? what is the expected behaviour? Should we update only card and ignore wallet? 
That is what will happen here.

crates/router/src/core/payment_methods/cards.rs
hrithikesh026: You can return storage::PaymentMethodUpdate from each of the branches and call update_payment_method once.

crates/router/src/core/payment_methods/cards.rs
AnuthaDev: Yeah, currently that is the logic. This also comes under the discussion of having payment_method_data as an enum. If we decide to make it an enum we won't have to add any error handling here. To be addressed in a future PR.

su-shivanshmathur (approved):

PR #9417, opened 2025-09-17
feat: dockerfile for building migration runner image

Shailesh-714:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

Adding a new `dockerfile` for building a `hyperswitch-migration-runner` image.
This dockerfile, `docker/migration-runner.Dockerfile`, contains the respective router version's:
1. Migration files
2. diesel.toml
along with diesel cli installed. This helps us run db migrations inside `air-gapped` clusters without the need to pull the migration files from github. 

### Additional Changes

## Motivation and Context

When deploying hyperswitch helm charts in Kubernetes cluster, we use a init pod to pull the migration files from the github and run db migrations. But in case of an air-gapped cluster, which doesn't have access to internet, this process is not feasible. 

Building and releasing the migration-runner image for each stable release along with the other standalone builds will solve this issue. 

The respective changes for the jenkins file has been made in the cloud repo:
PR: [https://github.com/juspay/hyperswitch-cloud/pull/10814](url)

## How did you test it?

1. Built image locally:
Screenshot 2025-09-17 at 3 20 20 PM

2. Run the image with local `pg` url, and migrations ran successfully:
Screenshot 2025-09-17 at 3 33 42 PM

## Checklist

### DIFF
--- a/docker/migration-runner.Dockerfile
+++ b/docker/migration-runner.Dockerfile
@@ -1,0 +1,45 @@
+# Migration Runner Dockerfile for Hyperswitch
+# This image contains migration files and diesel CLI for offline migration execution
+
+FROM debian:bookworm
+
+# Install necessary packages
+RUN apt-get update && apt-get install -y \
+    curl \
+    tar \
+    xz-utils \
+    ca-certificates \
+    libpq-dev \
+    && rm -rf /var/lib/apt/lists/*
+
+# Install diesel CLI (support multi-arch)
+ENV DIESEL_VERSION=2.3.0
+RUN ARCH=$(uname -m) && \
+    case "${ARCH}" in \
+      x86_64) DIESEL_ARCH="x86_64-unknown-linux-gnu" ;; \
+      aarch64) DIESEL_ARCH="aarch64-unknown-linux-gnu" ;; \
+      *) echo "Unsupported architecture: ${ARCH}" && exit 1 ;; \
+    esac && \
+    curl -L "https://github.com/diesel-rs/diesel/releases/download/v${DIESEL_VERSION}/diesel_cli-${DIESEL_ARCH}.tar.xz" | tar -xJ -C /tmp/ && \
+    mv "/tmp/diesel_cli-${DIESEL_ARCH}/diesel" /usr/local/bin/diesel && \
+    rm -rf "/tmp/diesel_cli-${DIESEL_ARCH}" && \
+    chmod +x /usr/local/bin/diesel
+
+# Set working directory
+WORKDIR /hyperswitch
+
+# Copy migration files and diesel config from the workspace
+COPY ./migrations/ ./migrations/
+COPY ./diesel.toml ./diesel.toml
+
+# Create the 'app' user and group (following hyperswitch pattern)
+RUN useradd --user-group --system --no-create-home --no-log-init app
+
+# Change ownership of working directory to app user
+RUN chown -R app:app /hyperswitch
+
+USER app:app
+
+# Default command to run migrations
+# Note: DATABASE_URL environment variable should be provided at runtime
+CMD ["diesel", "migration", "run"]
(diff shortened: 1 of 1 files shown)


docker/migration-runner.Dockerfile
```diff
@@ -0,0 +1,46 @@
+# Migration Runner Dockerfile for Hyperswitch
+# This image contains migration files and diesel CLI for offline migration execution
+
+FROM debian:bookworm
```
SanchithHegde: Can we use `bookworm-slim` or `trixie-slim` instead, or does something break with those image tags / variants?

docker/migration-runner.Dockerfile
```diff
@@ -0,0 +1,46 @@
...
+RUN ARCH=$(uname -m) && \
+    case "${ARCH}" in \
+      x86_64) DIESEL_ARCH="x86_64-unknown-linux-gnu" ;; \
+      aarch64) DIESEL_ARCH="aarch64-unknown-linux-gnu" ;; \
+      *) echo "Unsupported architecture: ${ARCH}" && exit 1 ;; \
+    esac && \
+    curl -L "https://github.com/diesel-rs/diesel/releases/download/v${DIESEL_VERSION}/diesel_cli-${DIESEL_ARCH}.tar.xz" | tar -xJ -C /tmp/ && \
+    mv "/tmp/diesel_cli-${DIESEL_ARCH}/diesel" /usr/local/bin/diesel && \
+    rm -rf "/tmp/diesel_cli-${DIESEL_ARCH}" && \
+    chmod +x /usr/local/bin/diesel
```
SanchithHegde: Any specific reasons we are downloading things manually instead of using the official installer script which handles multiple architectures?
Reference: https://diesel.rs/guides/getting-started#installing-diesel-cli

docker/migration-runner.Dockerfile
```diff
@@ -0,0 +1,46 @@
...
+
+# Set working directory
+WORKDIR /hyperswitch
+
+# Copy migration files and diesel config from the workspace
+COPY ./migrations/ ./migrations/
+COPY ./diesel.toml ./diesel.toml
+COPY ./crates/diesel_models/drop_id.patch ./crates/diesel_models/drop_id.patch
+
+# Create the 'app' user and group (following hyperswitch pattern)
```
SanchithHegde: You can remove the `following hyperswitch pattern` part of the comment.

```suggestion
# Create the 'app' user and group
```

docker/migration-runner.Dockerfile
```diff
@@ -0,0 +1,46 @@
...
+RUN useradd --user-group --system --no-create-home --no-log-init app
+
+# Change ownership of working directory to app user
+RUN chown -R app:app /hyperswitch
+
+USER app:app
+
+# Default command to run migrations
+# Note: DATABASE_URL environment variable should be provided at runtime
+CMD ["diesel", "migration", "run"]
```
SanchithHegde: Won't diesel CLI complain that the `schema.rs` file was not found?

Shailesh-714:
shouldn't we include the migration files in the router docker image itself and run the migration when the docker image is initialized?
cc: @SanchithHegde

SanchithHegde:
> shouldn't we include the migration files in the router docker image itself and run the migration when the docker image is initialized? cc: @SanchithHegde

Nope, I would not prefer doing this, to prevent destructive queries from being run automatically.

docker/migration-runner.Dockerfile
```diff
@@ -0,0 +1,40 @@
...
+
+FROM debian:trixie-slim
+
+# Install necessary packages
+RUN apt-get update && apt-get install -y \
+    curl \
+    tar \
+    xz-utils \
+    ca-certificates \
+    libpq-dev \
```
SanchithHegde: I think it should be sufficient to install `curl`, `tar` and `xz-utils` in most cases. Not sure if `ca-certificates` is needed, we can choose to keep it. 

I believe `libpq-dev` is not required, since the pre-built binaries are statically linked. We should be able to remove `libpq-dev`.

docker/migration-runner.Dockerfile
SanchithHegde, line 1: Have we tested this out with our Helm chart(s)?

pixincreate (approved):

PR #9453, opened 2025-09-19
feat(core): Update PaymentMethodData in PaymentsMethod Table when Connector sends new Card Info's in Webhooks

bsayak03:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

### Additional Changes

## Checklist

### DIFF
--- a/crates/api_models/src/payment_methods.rs
+++ b/crates/api_models/src/payment_methods.rs
@@ -781,6 +781,18 @@ pub struct CardDetailUpdate {
     /// Card Holder's Nick Name
     #[schema(value_type = Option<String>,example = "John Doe")]
     pub nick_name: Option<masking::Secret<String>>,
+
+    /// Card's Last 4 Digits
+    #[schema(value_type = Option<String>,example = "1111")]
+    pub last4_digits: Option<String>,
+
+    /// Issuing Bank of the Particular Card
+    #[schema(value_type = Option<String>,example = "Bank of America")]
+    pub card_issuer: Option<String>,
+
+    /// The country where that particular card was issued
+    #[schema(value_type = Option<String>,example = "US")]
+    pub issuer_country: Option<String>,
 }
 
 #[cfg(feature = "v1")]
--- a/crates/hyperswitch_connectors/src/connectors/adyen.rs
+++ b/crates/hyperswitch_connectors/src/connectors/adyen.rs
@@ -2056,8 +2056,64 @@ impl IncomingWebhook for Adyen {
                 });
         Ok(optional_network_txn_id)
     }
-}
 
+    fn get_additional_payment_method_data(
+        &self,
+        request: &IncomingWebhookRequestDetails<'_>,
+    ) -> CustomResult<
+        Option<api_models::payment_methods::PaymentMethodUpdate>,
+        errors::ConnectorError,
+    > {
+        let notif = get_webhook_object_from_body(request.body)
+            .change_context(errors::ConnectorError::WebhookBodyDecodingFailed)?;
+
+        let card_expiry_in_parts = notif
+            .additional_data
+            .expiry_date // expiry_date is of format: "M/YYYY"
+            .map(|date| {
+                date.expose()
+                    .split('/')
+                    .map(|s| s.to_string())
+                    .collect::<Vec<String>>()
+            })
+            .ok_or(errors::ConnectorError::ParsingFailed)?;
+
+        if card_expiry_in_parts.len() == 2 {
+            // Parse month as integer and format back as 2 digits
+            let month = card_expiry_in_parts
+                .get(1)
+                .ok_or(errors::ConnectorError::ParsingFailed)?
+                .parse::<u32>()
+                .change_context(errors::ConnectorError::ParsingFailed)?;
+            let month_str = Secret::new(format!("{:02}", month));
+
+            // Parse year as integer and take last 2 digits
+            let year = card_expiry_in_parts
+                .get(2)
+                .ok_or(errors::ConnectorError::ParsingFailed)?
+                .parse::<u32>()
+                .change_context(errors::ConnectorError::ParsingFailed)?;
+            let year_str = Secret::new(format!("{:02}", year % 100));
+            Ok(Some(api_models::payment_methods::PaymentMethodUpdate {
+                card: Some(api_models::payment_methods::CardDetailUpdate {
+                    card_exp_month: Some(month_str),
+                    card_exp_year: Some(year_str),
+                    card_holder_name: None,
+                    nick_name: None,
+                    issuer_country: notif.additional_data.card_issuing_country.clone(),
+                    card_issuer: notif.additional_data.card_issuing_bank.clone(),
+                    last4_digits: notif
+                        .additional_data
+                        .card_summary
+                        .map(|last4| last4.expose().to_string()),
+                }),
+                client_secret: None,
+            }))
+        } else {
+            Ok(None)
+        }
+    }
+}
 impl Dispute for Adyen {}
 impl DefendDispute for Adyen {}
 impl AcceptDispute for Adyen {}
--- a/crates/hyperswitch_connectors/src/connectors/adyen/transformers.rs
+++ b/crates/hyperswitch_connectors/src/connectors/adyen/transformers.rs
@@ -4981,6 +4981,10 @@ pub struct AdyenAdditionalDataWH {
     pub shopper_email: Option<String>,
     #[serde(rename = "shopperReference")]
     pub shopper_reference: Option<String>,
+    pub expiry_date: Option<Secret<String>>,
+    pub card_summary: Option<Secret<String>>,
+    pub card_issuing_country: Option<String>,
+    pub card_issuing_bank: Option<String>,
 }
 
 #[derive(Debug, Deserialize)]
--- a/crates/router/src/core/payment_methods/cards.rs
+++ b/crates/router/src/core/payment_methods/cards.rs
@@ -1765,8 +1765,12 @@ pub async fn update_customer_payment_method(
             // Construct new updated card object. Consider a field if passed in request or else populate it with the existing value from existing_card_data
             let updated_card = Some(api::CardDetailFromLocker {
                 scheme: existing_card_data.scheme,
-                last4_digits: Some(card_data_from_locker.card_number.get_last4()),
-                issuer_country: existing_card_data.issuer_country,
+                last4_digits: card_update
+                    .last4_digits
+                    .or(Some(card_data_from_locker.card_number.get_last4())),
+                issuer_country: card_update
+                    .issuer_country
+                    .or(existing_card_data.issuer_country),
                 card_number: existing_card_data.card_number,
                 expiry_month: card_update
                     .card_exp_month
@@ -1780,7 +1784,7 @@ pub async fn update_customer_payment_method(
                 nick_name: card_update.nick_name.or(existing_card_data.nick_name),
                 card_network: existing_card_data.card_network,
                 card_isin: existing_card_data.card_isin,
-                card_issuer: existing_card_data.card_issuer,
+                card_issuer: card_update.card_issuer.or(existing_card_data.card_issuer),
                 card_type: existing_card_data.card_type,
                 saved_to_locker: true,
             });
--- a/crates/router/src/core/webhooks/incoming.rs
+++ b/crates/router/src/core/webhooks/incoming.rs
@@ -29,6 +29,7 @@ use crate::{
         api_locking,
         errors::{self, ConnectorErrorExt, CustomResult, RouterResponse, StorageErrorExt},
         metrics, payment_methods,
+        payment_methods::cards,
         payments::{self, tokenization},
         refunds, relay, unified_connector_service, utils as core_utils,
         webhooks::{network_tokenization_incoming, utils::construct_webhook_router_data},
@@ -1019,6 +1020,18 @@ async fn payments_incoming_webhook_flow(
                 )
                 .await?
             };
+
+            if should_update_additional_payment_method_data(source_verified, event_type) {
+                update_additional_payment_method_data(
+                    &state,
+                    &merchant_context,
+                    webhook_details.object_reference_id.clone(),
+                    connector,
+                    request_details,
+                )
+                .await?
+            }
+
             lock_action
                 .free_lock_action(
                     &state,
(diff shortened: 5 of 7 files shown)


crates/api_models/src/payment_methods.rs
```diff
@@ -781,6 +781,18 @@ pub struct CardDetailUpdate {
...
+    /// Card's Last 4 Digits
+    #[schema(value_type = Option<String>,example = "1111")]
+    pub last4_digits: Option<String>,
+
+    /// Issuing Bank of the Particular Card
+    #[schema(value_type = Option<String>,example = "Bank of America")]
+    pub card_issuer: Option<String>,
+
+    /// The country where that particular card was issued
+    #[schema(value_type = Option<String>,example = "US")]
```
kashif-m: nit -
```suggestion
    #[schema(value_type = Option<String>, example = "1111")]
    pub last4_digits: Option<String>,

    /// Issuing Bank of the Particular Card
    #[schema(value_type = Option<String>, example = "Bank of America")]
    pub card_issuer: Option<String>,

    /// The country where that particular card was issued
    #[schema(value_type = Option<String>, example = "US")]
```

crates/hyperswitch_connectors/src/connectors/adyen/transformers.rs
```diff
@@ -4981,6 +4981,10 @@ pub struct AdyenAdditionalDataWH {
     pub shopper_email: Option<String>,
     #[serde(rename = "shopperReference")]
     pub shopper_reference: Option<String>,
+    pub expiry_date: Option<Secret<String>>,
+    pub card_summary: Option<Secret<String>>,
+    pub card_issuing_country: Option<String>,
+    pub card_issuing_bank: Option<String>,
```
kashif-m, line 5203: Do we get these fields back in the response?

crates/router/src/core/webhooks/incoming.rs
```diff
@@ -2321,6 +2334,45 @@ fn should_update_connector_mandate_details(
     source_verified && event_type == webhooks::IncomingWebhookEvent::PaymentIntentSuccess
 }
 
+fn should_update_additional_payment_method_data(
+    source_verified: bool,
+    event_type: webhooks::IncomingWebhookEvent,
+) -> bool {
+    source_verified && event_type == webhooks::IncomingWebhookEvent::PaymentIntentFailure
```
kashif-m: Why are updating only for failed payments?

We should update regardless of payment status. But we should check if existing details are different from what Adyen sent in the response

crates/hyperswitch_connectors/src/connectors/adyen.rs
```diff
@@ -2056,8 +2056,64 @@ impl IncomingWebhook for Adyen {
...
+    ) -> CustomResult<
+        Option<api_models::payment_methods::PaymentMethodUpdate>,
+        errors::ConnectorError,
+    > {
+        let notif = get_webhook_object_from_body(request.body)
+            .change_context(errors::ConnectorError::WebhookBodyDecodingFailed)?;
+
+        let card_expiry_in_parts = notif
+            .additional_data
+            .expiry_date // expiry_date is of format: "M/YYYY"
```
kashif-m: I have seen below formats in the response

`MM/YYYY`
`MM\/YYYY`

Let's handle the backward slash too (which is probably there to escape the forward slash)

crates/hyperswitch_connectors/src/connectors/adyen.rs
```diff
@@ -2056,8 +2056,64 @@ impl IncomingWebhook for Adyen {
...
+            .ok_or(errors::ConnectorError::ParsingFailed)?;
+
+        if card_expiry_in_parts.len() == 2 {
+            // Parse month as integer and format back as 2 digits
+            let month = card_expiry_in_parts
+                .get(1)
+                .ok_or(errors::ConnectorError::ParsingFailed)?
+                .parse::<u32>()
+                .change_context(errors::ConnectorError::ParsingFailed)?;
+            let month_str = Secret::new(format!("{:02}", month));
```
kashif-m: Is it always 2 characters? Let's make sure we handle all the cases

`M/YY`
`MM/YY`
`M/YYYY`
`MM/YYYY`

kashif-m: We can add an utility fn for this maybe?

crates/hyperswitch_connectors/src/connectors/adyen/transformers.rs
bsayak03: yes

crates/router/src/core/webhooks/incoming.rs
bsayak03: Cool, removed this check

crates/hyperswitch_connectors/src/connectors/adyen/transformers.rs
```diff
@@ -6259,3 +6263,34 @@ impl AdditionalData {
...
+
+pub fn parse_expiry_date(
+    raw: &str,
+) -> Result<(Secret<String>, Secret<String>), errors::ConnectorError> {
+    let cleaned = raw.replace('\\', "");
+    let parts: Vec<&str> = cleaned.split('/').collect();
+
+    let (month_str_raw, year_str_raw) = match (parts.first(), parts.get(1)) {
+        (Some(m), Some(y)) => (*m, *y),
+        _ => return Err(errors::ConnectorError::ParsingFailed),
```
kashif-m: nit: let's log the error here?

crates/router/src/core/payment_methods/cards.rs
```diff
@@ -1643,6 +1643,13 @@ pub async fn update_customer_payment_method(
             .await
             .to_not_found_response(errors::ApiErrorResponse::PaymentMethodNotFound)?;
 
+        if pm.locker_id.is_some() {
+        } else {
+            return Err(report!(errors::ApiErrorResponse::NotSupported {
+                message: "Payment method stored in locker, so it cannot be updated".into()
+            }));
+        }
```
kashif-m: Why is this checked added here? `update_customer_payment_method` is being used in other places too, right?

crates/router/src/core/webhooks/incoming.rs
```diff
@@ -1019,6 +1020,24 @@ async fn payments_incoming_webhook_flow(
...
+            .await
+            {
+                update_additional_payment_method_data(
+                    &state,
+                    &merchant_context,
+                    webhook_details.object_reference_id.clone(),
+                    connector,
+                    request_details,
+                )
+                .await?;
```
kashif-m: If this flow fails - we should silently log the error details, it should not affect the webhooks response status

crates/router/src/core/payment_methods/cards.rs
bsayak03: removing this check here since the locker id check is already there in `should_update_additional_payment_method_data` fn.

crates/router/src/core/webhooks/incoming.rs
```diff
@@ -1019,6 +1020,27 @@ async fn payments_incoming_webhook_flow(
...
+                if let Err(e) = update_additional_payment_method_data(
+                    &state,
+                    &merchant_context,
+                    webhook_details.object_reference_id.clone(),
+                    connector,
+                    request_details,
+                )
+                .await
+                {
+                    tracing::error!("Failed to update additional payment method data: {:?}", e);
```
kashif-m: We don't need to directly use tracing here, we can use router_env's logger - which is just renamed

```pub use tracing::{debug, error, event as log, info, warn};```

File: crates/router/src/core/webhooks/incoming.rs [hunk 10]
```diff
@@ -2321,6 +2346,71 @@ fn should_update_connector_mandate_details(
...
+    let db = state.store.as_ref();
+    let payment_attempt =
+        get_payment_attempt_from_object_reference_id(&state, object_ref_id, &merchant_context)
+            .await?;
+
+    let payment_method_id = payment_attempt
+        .payment_method_id
+        .clone()
+        .ok_or(errors::ApiErrorResponse::ResourceIdNotFound)
+        .attach_printable("No payment method id found in payment attempt")?;
```
kashif-m: Don't throw the error here

crates/router/src/core/webhooks/incoming.rs
```diff
@@ -2321,6 +2346,71 @@ fn should_update_connector_mandate_details(
...
+
+    let pm = db
+        .find_payment_method(
+            &((&state).into()),
+            merchant_context.get_merchant_key_store(),
+            &payment_method_id,
+            merchant_context.get_merchant_account().storage_scheme,
+        )
+        .await
+        .to_not_found_response(errors::ApiErrorResponse::PaymentMethodNotFound)?;
```
kashif-m: Same here, don't throw the error

crates/router/src/core/webhooks/incoming.rs
```diff
@@ -2375,6 +2397,81 @@ fn should_update_connector_mandate_details(
...
+        .await
+    {
+        Ok(pm) => pm,
+        Err(_) => return false,
+    };
+
+    if pm.locker_id.is_some() {
+        return false;
+    }
+
```
kashif-m: Can we add one more condition to check if update is actually needed?

Compare values received from connector's webhook and values present in DB

bsayak03: that check is already done in the `update_customer_payment_method` fn inside `crates/router/src/core/payment_methods/cards.rs`

crates/router/src/core/webhooks/incoming.rs
```diff
@@ -2375,6 +2397,81 @@ fn should_update_connector_mandate_details(
...
+            &payment_method_id,
+            merchant_context.get_merchant_account().storage_scheme,
+        )
+        .await
+    {
+        Ok(pm) => pm,
+        Err(_) => return false,
+    };
+
+    if pm.locker_id.is_some() {
```
hrithikesh026: can you add some comments explaining why these if checks are made to decide if payment method should be updated?

crates/router/src/core/webhooks/incoming.rs
```diff
@@ -2375,6 +2397,81 @@ fn should_update_connector_mandate_details(
...
+    object_ref_id: api::ObjectReferenceId,
+    connector: &ConnectorEnum,
+    request_details: &IncomingWebhookRequestDetails<'_>,
+) -> CustomResult<(), errors::ApiErrorResponse> {
+    let payment_method_update = connector
+        .get_additional_payment_method_data(request_details)
+        .change_context(errors::ApiErrorResponse::InternalServerError)?
+        .ok_or(errors::ApiErrorResponse::InternalServerError)?;
+
+    let payment_attempt =
```
hrithikesh026: payment_attempt is already fetched in the parent function. Pass it as argument here

crates/router/src/core/webhooks/incoming.rs
```diff
@@ -2375,6 +2397,81 @@ fn should_update_connector_mandate_details(
...
+        object_ref_id,
+        &merchant_context,
+    )
+    .await
+    {
+        Ok(pa) => pa,
+        Err(_) => return false,
+    };
+
+    let payment_method_id = match payment_attempt.payment_method_id.clone() {
```
hrithikesh026: If `payment_method_id` is the only value required, you can get it in payment_response from fn payments_core.
No need for this fetching attempt from DB.

crates/router/src/core/webhooks/incoming.rs
```diff
@@ -2375,6 +2397,81 @@ fn should_update_connector_mandate_details(
...
+        Some(id) => id,
+        None => return false,
+    };
+
+    let pm = match db
+        .find_payment_method(
```
hrithikesh026: Also fetching records from DB only for condition checking is not recommended. 
Try to have these checks in `fn update_customer_payment_method` handler only.

hrithikesh026 (dismissed):

crates/hyperswitch_connectors/src/connectors/adyen/transformers.rs
```diff
@@ -6303,3 +6308,134 @@ impl AdditionalData {
...
+    };
+
+    Ok((month_str, year_str))
+}
+
+pub fn from_payment_method_variant(value: Option<String>) -> Option<common_enums::CardNetwork> {
+    let val = value?.to_lowercase();
+    let cleaned = val.trim().split('_').next().unwrap_or("");
+
+    match cleaned {
```
jagan-jaya, line 6528: move this conversion to common_enums and use as `common_enums::CardNetwork.from(cleaned)`

crates/hyperswitch_connectors/src/connectors/adyen/transformers.rs
```diff
@@ -6303,3 +6308,134 @@ impl AdditionalData {
         })
     }
 }
+
+pub fn parse_expiry_date(
```
jagan-jaya: can we create a domain type for this type of expiry year? this code duplication can be avoided by having a type for it

jagan-jaya (changes requested):

crates/hyperswitch_connectors/src/connectors/adyen/transformers.rs
bsayak03: but this set of card network values are very specific to adyen, so should we move them to impl CardNetwork?

crates/router/src/core/webhooks/incoming.rs
```diff
@@ -2375,6 +2411,37 @@ fn should_update_connector_mandate_details(
     source_verified && event_type == webhooks::IncomingWebhookEvent::PaymentIntentSuccess
 }
 
+async fn should_update_additional_payment_method_data(
+    pm: hyperswitch_domain_models::payment_methods::PaymentMethod,
+) -> bool {
+    if pm.locker_id.is_some() {
+        return false;
+    }
+    true
```
hrithikesh026: nit: 
```suggestion
    }else {
        true
    }
```

bsayak03: no more needed because we removed this fn

crates/router/src/core/payment_methods/cards.rs
```diff
@@ -1636,6 +1636,7 @@ pub async fn update_customer_payment_method(
     merchant_context: domain::MerchantContext,
     req: api::PaymentMethodUpdate,
     payment_method_id: &str,
+    pm_fetch: Option<common_enums::enums::PaymentMethodFetch>,
```
hrithikesh026:
```suggestion
    pm_data: Option<domain::PaymentMethod>,
```

crates/router/src/core/payment_methods/cards.rs
```diff
@@ -1636,20 +1636,30 @@ pub async fn update_customer_payment_method(
...
+    pm_data: Option<domain::PaymentMethod>,
 ) -> errors::RouterResponse<api::PaymentMethodResponse> {
     // Currently update is supported only for cards
     if let Some(card_update) = req.card.clone() {
         let db = state.store.as_ref();
-
-        let pm = db
-            .find_payment_method(
+        let pm = if let Some(pm) = pm_data.clone() {
+            if pm.locker_id.is_some() {
```
hrithikesh026: this condition check should be done outside. Not here

crates/router/src/core/webhooks/incoming.rs
```diff
@@ -2375,6 +2397,41 @@ fn should_update_connector_mandate_details(
...
+        .find_payment_method(
+            &state.into(),
+            merchant_context.get_merchant_key_store(),
+            payment_method_id.as_str(),
+            merchant_context.get_merchant_account().storage_scheme,
+        )
+        .await
+        .to_not_found_response(errors::ApiErrorResponse::PaymentMethodNotFound)?;
+
+    Box::pin(cards::update_customer_payment_method(
```
hrithikesh026, line 2760: call this only if locker_id is none

crates/router/src/core/webhooks/incoming.rs
```diff
@@ -1136,6 +1139,25 @@ async fn payments_incoming_webhook_flow(
 
             let response = WebhookResponseTracker::Payment { payment_id, status };
 
+            match payment_method_id_opt {
```
hrithikesh026: I think this step should be moved before triggering webhooks.

crates/router/src/core/webhooks/incoming.rs
```diff
@@ -1121,8 +1124,8 @@ async fn payments_incoming_webhook_flow(
             if let Some(outgoing_event_type) = event_type {
                 let primary_object_created_at = payments_response.created;
                 Box::pin(super::create_event_and_trigger_outgoing_webhook(
-                    state,
-                    merchant_context,
+                    state.clone(),
+                    merchant_context.clone(),
```
hrithikesh026: try to avoid clone as much as possible

crates/router/src/core/webhooks/incoming.rs
hrithikesh026: This will eliminate cloning of state and merchant_context in `create_event_and_trigger_outgoing_webhook`.

crates/router/src/core/payment_methods/cards.rs
```diff
@@ -1636,20 +1636,23 @@ pub async fn update_customer_payment_method(
...
     payment_method_id: &str,
+    pm_data: Option<domain::PaymentMethod>,
 ) -> errors::RouterResponse<api::PaymentMethodResponse> {
     // Currently update is supported only for cards
     if let Some(card_update) = req.card.clone() {
         let db = state.store.as_ref();
-
-        let pm = db
-            .find_payment_method(
+        let pm = if let Some(pm) = pm_data.clone() {
```
hrithikesh026:
```suggestion
        let pm = if let Some(pm) = pm_data {
```

hrithikesh026 (approved):

PR #9502, opened 2025-09-23
feat(connector): [ACI] cypress added

swetasharma03:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

Added cypress for ACI.

Following bugs were detected while adding cypress, this PR includes the fix for those as well:

1. `card_network` is required field only for network_token and not for normal card_payment, so made the field optional and populated missing error only for network_token flow.
2. When clicked technical_error after redirection, on returning back server was throwing 500 because the code for technical_error was not included in `aci/aci_result_codes.rs`.
3. Some fields are optional in capture response, handled it by making them optional.
4. For Void flow, separate `try_from` was not there previously and one for `Capture` was getting used up. This was causing incorrect status mapping for `Void`.

### Additional Changes

## How did you test it?

Screenshot 2025-09-23 at 6 07 35 PM

## Checklist


crates/hyperswitch_connectors/src/connectors/aci/transformers.rs
```diff
@@ -1257,6 +1263,69 @@ impl<F, T> TryFrom<ResponseRouterData<F, AciCaptureResponse, T, PaymentsResponse
...
+    currency: String,
+    descriptor: String,
+    result: AciCaptureResult,
+    result_details: Option<AciCaptureResultDetails>,
+    build_number: String,
+    timestamp: String,
+    ndc: Secret<String>,
+}
+
+fn map_aci_void_status(item: AciCaptureStatus) -> enums::AttemptStatus {
```
deepanshu-iiitu: Rename AciCaptureStatus to AciStatus

PR #9575, opened 2025-09-26
fix(routing): add field length validation for the routing config name field

spritianeja03:
## Type of Change

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

This PR adds a length validation for the name field in the routing config to be less than 64 characters

### Additional Changes

## How did you test it?

Hitting the /routing endpoint with config name lengthgreater than 64 chars
Screenshot 2025-09-26 at 3 08 19 PM
Screenshot 2025-09-26 at 3 09 09 PM

Hitting /routing endpoint with config name length less than 64 chars
Screenshot 2025-09-26 at 3 10 41 PM

## Checklist

### DIFF
--- a/crates/api_models/src/routing.rs
+++ b/crates/api_models/src/routing.rs
@@ -73,7 +73,7 @@ impl RoutingConfigRequest {
         if let Some(name) = &self.name {
             if name.len() > 64 {
                 return Err(ValidationError::InvalidValue {
-                    message: "Name length must not exceed 64 characters".to_string()
+                    message: "Name length must not exceed 64 characters".to_string(),
                 });
             }
         }
--- a/crates/api_models/src/routing.rs
+++ b/crates/api_models/src/routing.rs
@@ -67,6 +67,20 @@ pub struct RoutingConfigRequest {
     pub transaction_type: Option<TransactionType>,
 }
 
+#[cfg(feature = "v1")]
+impl RoutingConfigRequest {
+    pub fn validate_name_length(&self) -> Result<(), ValidationError> {
+        if let Some(name) = &self.name {
+            if name.len() > 64 {
+                return Err(ValidationError::InvalidValue {
+                    message: "Name length must not exceed 64 characters".to_string()
+                });
+            }
+        }
+        Ok(())
+    }
+}
+
 #[derive(Debug, serde::Serialize, ToSchema)]
 pub struct ProfileDefaultRoutingConfig {
     #[schema(value_type = String)]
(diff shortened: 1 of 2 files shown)


crates/api_models/src/routing.rs
```diff
@@ -67,6 +67,20 @@ pub struct RoutingConfigRequest {
     pub transaction_type: Option<TransactionType>,
 }
 
+#[cfg(feature = "v1")]
+impl RoutingConfigRequest {
+    pub fn validate_name_length(&self) -> Result<(), ValidationError> {
+        if let Some(name) = &self.name {
+            if name.len() > 64 {
+                return Err(ValidationError::InvalidValue {
+                    message: "Name length must not exceed 64 characters".to_string(),
```
prajjwalkumar17:
```suggestion
                    message: "Length for name field must not exceed 64 characters".to_string(),
```

crates/api_models/src/routing.rs
```diff
@@ -67,6 +67,20 @@ pub struct RoutingConfigRequest {
     pub transaction_type: Option<TransactionType>,
 }
 
+#[cfg(feature = "v1")]
+impl RoutingConfigRequest {
+    pub fn validate_name_length(&self) -> Result<(), ValidationError> {
+        if let Some(name) = &self.name {
+            if name.len() > 64 {
```
prajjwalkumar17: Can we make this const.

prajjwalkumar17 (approved):

crates/api_models/src/routing.rs
```diff
@@ -67,6 +68,23 @@ pub struct RoutingConfigRequest {
     pub transaction_type: Option<TransactionType>,
 }
 
+#[cfg(feature = "v1")]
+impl RoutingConfigRequest {
```
ShankarSinghC: This validation can be handled by introducing a new type for `name` in the RoutingConfigRequest.

crates/api_models/src/routing.rs
```diff
@@ -67,6 +68,23 @@ pub struct RoutingConfigRequest {
...
+    pub fn validate_name_length(&self) -> Result<(), ValidationError> {
+        if let Some(name) = &self.name {
+            if name.len() > MAX_NAME_LENGTH {
+                return Err(ValidationError::InvalidValue {
+                    message: format!(
+                        "Length of name field must not exceed {} characters",
+                        MAX_NAME_LENGTH
+                    ),
+                });
```
ShankarSinghC: Also this is just a api level handling, returning this error message won't be enough from the merchant's view point.
But we need to also handle it in the fronted for the better experience. 

For example we can have a validation in the fronted where is the name is greater than 64 character the create run button will be displayed.

prajjwalkumar17 (approved):

PR #9604, opened 2025-09-29
refactor(users): remove deprecated permission groups

kanikabansal08:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

Deprecated groups include:

1. organizationManage
2. merchantDetailsView
3. merchantDetailsManage

These permissions are equivalent to
organizationManage == accountManage 
merchantDetailsManage == accountManage
merchantDetailsView == accountView

Also earlier validation blocks organizationManage when creating custom roles because this permission was only available for predefined roles. We will be removing that validation check as the permission itself doesn't exists now.

### Additional Changes

## Motivation and Context

Required to remove deprecated apis.

## How did you test it?

Local Verification Steps:

1. Created custom roles using deprecated permissions.
2. Verified the actual access those roles provided.
3. Ran the query to replace/update the deprecated permissions.
4. Rechecked the same roles after the update.
5. Confirmed that the access remained unchanged post-migration.

## Checklist


migrations/2025-10-06-111411_deprecated_roles_backfill/down.sql
```diff
@@ -0,0 +1,2 @@
+-- This file should undo anything in `up.sql`
+SELECT 1;
```
sai-harsha-vardhan, line 2: We can have a down.sql doing the backfill right?

kanikabansal08: The migrations are both backward and forward compatible so there’s no need to run them when rolling back the code.

PR #9661, opened 2025-10-03
fix(router): bump serde to latest version

sahkal:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

### Additional Changes

## Checklist

### DIFF
--- a/crates/masking/src/serde.rs
+++ b/crates/masking/src/serde.rs
@@ -149,12 +149,6 @@ mod pii_serializer {
         // or send empty map if the serializer is of type FlatMapSerializer over PiiSerializer
         if std::any::type_name::<S>() == std::any::type_name::<PIISerializer>() {
             format!("{value:?}").serialize(serializer)
-        } else if std::any::type_name::<S>()
-            == std::any::type_name::<
-                serde::__private::ser::FlatMapSerializer<'_, SerializeMap<PIISerializer>>,
-            >()
-        {
-            std::collections::HashMap::<String, String>::from([]).serialize(serializer)
         } else {
             value.peek().serialize(serializer)
         }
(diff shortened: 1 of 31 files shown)


crates/masking/src/serde.rs
```diff
@@ -149,12 +149,6 @@ mod pii_serializer {
         // or send empty map if the serializer is of type FlatMapSerializer over PiiSerializer
         if std::any::type_name::<S>() == std::any::type_name::<PIISerializer>() {
             format!("{value:?}").serialize(serializer)
-        } else if std::any::type_name::<S>()
-            == std::any::type_name::<
-                serde::__private::ser::FlatMapSerializer<'_, SerializeMap<PIISerializer>>,
-            >()
-        {
-            std::collections::HashMap::<String, String>::from([]).serialize(serializer)
```
SanchithHegde, line 157: I'm not completely sure if this affects masked serialization in any way, and / or whether this was included (in #2681) to handle any specific edge cases.

One thing that we're sure about is that nobody else can pass `FlatMapSerializer` because it is now private. However, I'm not sure if this changes any behavior where the derive macros are used (which is I think most places), considering that this is called in `serde_derive` when there is `#[serde(flatten)]`:
https://github.com/serde-rs/serde/blob/v1.0.228/serde_derive/src/ser.rs#L1135-L1140

sahkal: FlatMap issue will popup on structs where a param marked as secret is flattened

Something like

```
#[derive(Serialize)]
struct Outer {
    #[serde(flatten)]
    inner: Inner,
}

struct Inner {
    secret: Secret<String>,
}

```

pii_serialize will receive `S = FlatMapSerializer<PIISerializer>`. The type-name check fails, so it falls through to value.peek().serialize(serializer), meaning the secret is exposed rather than masked.

Issue #9641, opened 2025-10-01
[ENHANCE] Improve Error Reporting in change_password with Printable Context


apoorvdixit88:
This issue aims to improve the debuggability of the `change_password` function by adding a printable message to a potential database error. Currently, if the database update fails, a generic `InternalServerError` is returned. By attaching a specific, contextual message, we can make debugging and logging much more effective.

In the [change_password](https://github.com/juspay/hyperswitch/blob/cf30da2df05aac69fd82b676e639b2fbc85dcae5/crates/router/src/core/user.rs#L429) function located in `crates/router/src/core/user.rs`, we update the user's password in the database. The relevant code block is:

If the update_user_by_user_id operation fails (e.g., due to a database connection issue, a constraint violation, etc.), it's wrapped in a [UserErrors::InternalServerError](https://github.com/juspay/hyperswitch/blob/cf30da2df05aac69fd82b676e639b2fbc85dcae5/crates/router/src/core/user.rs#L461). While this correctly signals a server-side problem, the error itself lacks specific context. When debugging, an engineer seeing this error in the logs would have to trace back the execution flow to understand which operation and for which user it failed.

#### __Proposed Solution 💡__

We can enhance this by using the `.attach_printable()` method from the `error-stack` crate. This method allows us to attach a static or pre-formatted string message to the error context.

The goal is to modify the code to include the error message, making it immediately clear what failed during InternalServerError

#### __How to Test Your Changes 🧪__

1. __Make the code change__: Apply the suggested modification to the `change_password` function.

2. __Build the project__: Run `cargo build` or `cargo check` to ensure your changes compile correctly. __If it compiles, the `attach_printable` should work fine.__

3. __Simulate an error (Advanced)__:

   - A good way to test this end-to-end is to:

     1. Create a user via the signup flow.
     2. Log in and get an authentication token.
     3. Call the `change_password` endpoint.

   - To reliably trigger the error for observation, you could temporarily modify the `update_user_by_user_id` method in its respective implementation to always return an error.

   - A simpler way to verify the message is being constructed is to temporarily introduce a `panic!` or a different, guaranteed error inside the `attach_printable` argument to see if it gets hit (remember to remove this temporary change!).

4. __Observe the output__: If you can trigger the error, you should see your new, more descriptive message in the application logs or error response, making it clear that the failure happened during the password update for a specific user.

####

####

Prakhar0013:
[...]

PR #9667, opened 2025-10-04
Add attach_printable() for better error logging

Prakhar0013:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [✔️ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description
Just modified the error logging on change of password for getting better error details and description. This helps in knowing about error in more detail. 

### Additional Changes

I made changes to crates/router/src/core/user.rs

## Motivation and Context

[ENHANCE] Improve Error Reporting in change_password with Printable Context #9641 issue was there which was assigned to me. 

## How did you test it?
Tried to run cargo build.

### DIFF
--- a/crates/router/src/core/user.rs
+++ b/crates/router/src/core/user.rs
@@ -449,6 +449,7 @@ pub async fn change_password(
     let new_password_hash =
         utils::user::password::generate_password_hash(new_password.get_secret())?;
 
+    let user_id = user.get_user_id();
     let _ = state
         .global_store
         .update_user_by_user_id(
@@ -458,6 +459,7 @@ pub async fn change_password(
             },
         )
         .await
+        .attach_printable_lazy(|| format!("Failed to update password in database for user_id: {}", user_id))
         .change_context(UserErrors::InternalServerError)?;
 
     let _ = auth::blacklist::insert_user_in_blacklist(&state, user.get_user_id())


masashi123456:
Sorry, I'm new to this, but what is the main benefit of this error logging approach?

crates/router/src/core/user.rs
```diff
@@ -449,6 +449,7 @@ pub async fn change_password(
     let new_password_hash =
         utils::user::password::generate_password_hash(new_password.get_secret())?;
 
+    let user_id = user.get_user_id();
```
apoorvdixit88: please check if we can remove this variable assignment.

crates/router/src/core/user.rs
```diff
@@ -458,6 +459,7 @@ pub async fn change_password(
             },
         )
         .await
+        .attach_printable_lazy(|| format!("Failed to update password in database for user_id: {}", user_id))
```
apoorvdixit88: please check if we can use `attach_printable` instead of `attach_printable_lazy`.

apoorvdixit88:
> Sorry, I'm new to this, but what is the main benefit of this error logging approach?

Hi @masashi123456 
The main benefit of using error logging or using attach_printable() here is better context in debugging. Normally, when an internal error occurs (like a DB update failure), we only see a generic InternalServerError, which doesn’t tell us where or why it happened.

By attaching a printable message, we make the logs and error traces more descriptive, for example:

`Failed to update password for user_id: xyz due to database constraint violation`

This saves time during debugging since engineers can **Immediately** see what operation failed, without having to trace through multiple layers of code or let's say going to particular line number in file while debugging.

crates/router/src/core/user.rs
Prakhar0013: Yeah, sure.

crates/router/src/core/user.rs
Prakhar0013: Okay, will do.

Prakhar0013:
Is this right? Please suggest some improvements if any.

Prakhar0013:
Okay

Prakhar0013:
I just recorrected one last mistake of mine

Prakhar0013:
How to remove these commit requests? Is there any way? Please let me know

apoorvdixit88 (commented):
Can you please revert changes apart from this issue, spell check is not required check. If we cover it, then we have to generate regenerate openspec api json and all. Please try to cover required check as of now, for other changes we can create new issue.

Prakhar0013:
Sure

Prakhar0013:
I have reverted all the changes in comments and docs except for the ones in user.rs. Sorry , guess I was a bit overboard earlier.

apoorvdixit88 (approved):

Prakhar0013:
@apoorvdixit88, please get it reviewed by another reviewer for approval.

PR #9710, opened 2025-10-07
ci: add API and database Schema validation workflow

itsharshvb:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [x] CI/CD

## Description

Added automated validation workflows to detect breaking changes in API schemas and database migrations.
**New Workflows:**
- `api-compatibility.yml` - Validates API schema compatibility using oasdiff and checks migrations for breaking SQL operations

**Configuration:**
- Added oasdiff config files to `.github/oasdiff/` (`.oasdiff-err-ignore.yaml`, `.oasdiff-warn-ignore.yaml`, `.oasdiff-severity-levels.yaml`)

### Additional Changes

## Motivation and Context

As the API evolves, we need automated checks to:
- Detect breaking changes in API contracts before production deployment
- Prevent dangerous database operations (DROP COLUMN, TRUNCATE, DELETE FROM)
- Maintain backward compatibility for API consumers
- Provide clear, actionable feedback on compatibility issues

The workflow validates:
- **API Changes:** Removed endpoints, new required fields, type changes, response schema modifications
- **Migration Safety:** DROP operations, TRUNCATE commands, potentially dangerous DELETE statements

## How did you test it?

1. Created intentional breaking changes to test detection:
   - Added required field `organization_email` to organization API
   - Removed `organization_details` from response schema
   - Changed endpoint path (breaking route change)
   - Added DROP COLUMN migrations in both v1 and v2

2. Verified workflow triggers on:
   - Changes to `crates/api_models/**`, `crates/openapi/**`, `crates/router/**`
   - Migration file changes in `migrations/**` , `v2_migrations/**` and `v2_compatible_migrations/**`
 
 
Screenshot 2025-10-15 at 11 38 12 PM

Screenshot 2025-12-03 at 12 20 09 PM

## Checklist


.github/oasdiff/.oasdiff-severity-levels.yaml
```diff
@@ -0,0 +1,100 @@
+api-path-removed-without-deprecation: ERR
+api-path-removed-before-sunset: WARN
+api-removed-without-deprecation: ERR
+api-removed-before-sunset: WARN
```
pixincreate: this could be escalated to ERR for payment apis? context dependent, of course.

.github/api-migration-compatibility/.oasdiff-severity-levels.yaml
pixincreate, line 1: should we consider adding below ones too?

```yaml
# changing the type of the amount field (e.g., from integer to float) could break calculations or compliance
payment-amount-field-type-changed: ERR
```
```yaml
# changing the format (e.g., JSON to XML) is breaking
request-body-format-changed: ERR
```
```yaml
# changing the response content type (e.g., from JSON to XML) is breaking
response-content-type-changed: ERR
```
```yaml
# changing the possible values of a payment status enum (e.g., from "pending" to "in_progress") could break client workflows
payment-status-enum-changed: ERR
```
```yaml
# Deprecating a parameter should be a warning, as it signals future removal
request-parameter-deprecated: WARN
```

.github/oasdiff/.oasdiff-severity-levels.yaml
```diff
@@ -0,0 +1,100 @@
+api-path-removed-without-deprecation: ERR
+api-path-removed-before-sunset: WARN
```
pixincreate: shouldn't this is `ERR`?

i feel it is too lenient for payment apis

.github/oasdiff/.oasdiff-severity-levels.yaml
```diff
@@ -0,0 +1,100 @@
...
+api-schema-removed: ERR
+new-required-request-header: ERR
+new-optional-request-header: INFO
+endpoint-deprecated: WARN
+endpoint-reactivated: INFO
+sunset-date-too-small: WARN
+payment-amount-field-removed: ERR
+payment-currency-field-removed: ERR
+payment-method-enum-removed: ERR
+connector-enum-removed: WARN
```
pixincreate: i feel, this could be an `ERR` instead, since this would be a rare case where is a connector is removed, clients using that connector will break?

.github/workflows/api-compatibility.yml
```diff
@@ -0,0 +1,432 @@
...
+          repository: ${{ github.event.pull_request.head.repo.full_name }}
+
+      - name: Checkout PR with token
+        if: ${{ (github.event_name == 'pull_request') && (github.event.pull_request.head.repo.full_name == github.event.pull_request.base.repo.full_name) }}
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+          ref: ${{ github.event.pull_request.head.ref }}
+          repository: ${{ github.event.pull_request.head.repo.full_name }}
+          token: ${{ steps.generate_token.outputs.token }}
```
pixincreate: why is this needed? i do not see you doing a commit or anything as such. it can just be a simple checkout step, right?

the latest version is `v5`, btw

.github/workflows/api-compatibility.yml
```diff
@@ -0,0 +1,432 @@
...
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  api-compatibility-check:
+    name: API & Migration Compatibility Check
+    runs-on: ubuntu-latest
+
+    if: github.event_name == 'pull_request' || github.event_name == 'workflow_dispatch'
```
pixincreate: i do not think it is necessary here since it is already defined at the top.

.github/workflows/api-compatibility.yml
```diff
@@ -0,0 +1,432 @@
...
+
+      - name: Setup Rust cache
+        uses: Swatinem/rust-cache@v2
+        with:
+          # Cache key based on Cargo files
+          key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
+          # Cache the target directory
+          cache-targets: true
+          # Cache cargo registry
+          cache-all-crates: true
```
pixincreate: you're running only 2 cargo commands, and that to for generating the openapi spec, do we need this step?

if really needed, you can use `sccache` instead?

```yaml
- name: Install sccache
  uses: taiki-e/install-action@v2
  with:
    tool: sccache
    checksum: true
```

.github/workflows/api-compatibility.yml
```diff
@@ -0,0 +1,432 @@
...
+        run: sudo apt-get update && sudo apt-get install -y jq
+
+      - name: Setup Go for oasdiff
+        uses: actions/setup-go@v5
+        with:
+          go-version: '1.21'
+
+      - name: Install oasdiff
+        run: |
+          go install github.com/oasdiff/oasdiff@latest
```
pixincreate: i would prefer using a specific version over `@latest`.

.github/workflows/api-migrations-compatibility.yml
```diff
@@ -0,0 +1,432 @@
...
+
+      - name: Setup Go for oasdiff
+        uses: actions/setup-go@v5
+        with:
+          go-version: '1.21'
+
+      - name: Install oasdiff
+        run: |
+          go install github.com/oasdiff/oasdiff@latest
+          echo "$HOME/go/bin" >> $GITHUB_PATH
```
pixincreate, line 47: why?

.github/workflows/api-compatibility.yml
```diff
@@ -0,0 +1,432 @@
...
+          echo "$HOME/go/bin" >> $GITHUB_PATH
+
+      - name: Verify oasdiff installation
+        run: |
+          export PATH="$HOME/go/bin:$PATH"
```
pixincreate: again, why?

.github/workflows/api-compatibility.yml
```diff
@@ -0,0 +1,432 @@
...
+          repository: ${{ github.event.pull_request.head.repo.full_name }}
+
+      - name: Checkout PR with token
+        if: ${{ (github.event_name == 'pull_request') && (github.event.pull_request.head.repo.full_name == github.event.pull_request.base.repo.full_name) }}
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+          ref: ${{ github.event.pull_request.head.ref }}
+          repository: ${{ github.event.pull_request.head.repo.full_name }}
+          token: ${{ steps.generate_token.outputs.token }}
```
pixincreate: same questions applies to here as well

.github/workflows/api-compatibility.yml
```diff
@@ -0,0 +1,432 @@
...
+            v2-detailed-diff.txt
+            v1-breaking-status.txt
+            v2-breaking-status.txt
+          retention-days: 30
+
+  migration-compatibility-check:
+    name: Migration Breaking Change Detection
+    runs-on: ubuntu-latest
+
+    if: github.event_name == 'pull_request' || github.event_name == 'workflow_dispatch'
```
pixincreate: we do not need this i guess

.github/oasdiff/.oasdiff-severity-levels.yaml
itsharshvb: updated this to ERR

.github/api-migration-compatibility/.oasdiff-severity-levels.yaml
itsharshvb: the first four are not available oasdiff's rule set. 5th one was i available i have included it.
`payment-amount-field-type-changed` - Covered by: `request-property-type-changed: ERR` & `response-property-type-changed: ERR`
others - covered by `request-body-type-changed: ERR`,`response-media-type-removed: ERR (via oasdiff default)` & `request-property-enum-value-removed: ERR`

.github/oasdiff/.oasdiff-severity-levels.yaml
itsharshvb: updated this to ERR

.github/oasdiff/.oasdiff-severity-levels.yaml
itsharshvb: the `connector-enum-removed` rule wasn't valid/present in `https://github.com/oasdiff/oasdiff/blob/main/checker/localizations_src/en/messages.yaml` so removed it , this should be handle via `request-property-enum-value-removed: ERR`

.github/workflows/api-migrations-compatibility.yml
itsharshvb: Go installs binaries into $HOME/go/bin, which is user-specific . so explicitly add it

.github/workflows/api-compatibility.yml
itsharshvb: removed cargo build since its handle in required workflow `validate-openapi-spec.yaml`

.github/oasdiff/.oasdiff-severity-levels.yaml
```diff
@@ -0,0 +1,61 @@
...
+request-parameter-min-length-increased: ERR
+request-parameter-min-length-decreased: INFO
+request-parameter-pattern-changed: ERR
+request-parameter-pattern-removed: INFO
+request-parameter-type-changed: ERR
+request-parameter-x-extensible-enum-value-removed: ERR
+request-parameter-deprecated: WARN
+api-schema-removed: INFO
+endpoint-deprecated: WARN
+endpoint-reactivated: INFO
```
pixincreate: nit: add a new line at the end of both the files

pixincreate (dismissed):

.github/workflows/api-compatibility.yml
```diff
@@ -0,0 +1,314 @@
...
+# This workflow:
+# - Runs on ALL PRs that modify API-related files
+# - Always fails if breaking changes are detected
+
+on:
+  pull_request:
+    paths:
+      - 'crates/api_models/**'
+      - 'crates/openapi/**'
+      - 'crates/router/**'
```
SanchithHegde: Is there anything in the `router` crate that the OpenAPI spec diff or the migration diff checks depend on?

.github/workflows/api-compatibility.yml
```diff
@@ -0,0 +1,314 @@
...
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v5
+        with:
+          fetch-depth: 0
+
+      - name: Install system dependencies
+        run: sudo apt-get update && sudo apt-get install -y jq
```
SanchithHegde: Do we necessarily need this? I believe jq is available on the runners by default?

.github/workflows/api-compatibility.yml
```diff
@@ -0,0 +1,314 @@
...
+
+          while IFS= read -r file; do
+            [[ -z "$file" ]] && continue
+            CONTENT=$(cat "$file")
+
+            if echo "$CONTENT" | grep -iE "(DROP TABLE|DROP COLUMN|DELETE FROM|TRUNCATE)" >/dev/null; then
+              echo "BREAKING: $file"
+              echo "$CONTENT" | grep -inE "(DROP TABLE|DROP COLUMN|DELETE FROM|TRUNCATE)" | sed 's/^/  /'
+              BREAKING=$((BREAKING + 1))
+            fi
```
SanchithHegde: I believe we are okay with `DROP COLUMN` queries being present in the `v2_migrations` directory, @hrithikesh026 please confirm.

.github/workflows/api-compatibility.yml
```diff
@@ -0,0 +1,314 @@
...
+          else
+            echo "validation_status=ALL CHECKS PASSED" >> $GITHUB_OUTPUT
+          fi
+        env:
+          MIGRATION_BREAKING: ${{ steps.migration_check.outputs.breaking_changes }}
+          MIGRATION_WARNINGS: ${{ steps.migration_check.outputs.warnings }}
+          V2_MIGRATION_BREAKING: ${{ steps.v2_migration_check.outputs.breaking_changes }}
+          V2_MIGRATION_WARNINGS: ${{ steps.v2_migration_check.outputs.warnings }}
+          V2_COMPAT_MIGRATION_BREAKING: ${{ steps.v2_compat_migration_check.outputs.breaking_changes }}
+          V2_COMPAT_MIGRATION_WARNINGS: ${{ steps.v2_compat_migration_check.outputs.warnings }}
```
SanchithHegde: Nit: Can you please move the `env` block before the `run` block?

.github/workflows/api-migrations-compatibility.yml
SanchithHegde, line 1: Nit: We can consider naming this as `api-migrations-compatibility.yaml`, since we do checks for both.

.github/workflows/api-compatibility.yml
itsharshvb: relaxed `DROP COLUMN` in `v2_migrations` as warning, however `v2_compatiable migration` would enforce`DROP COLUMN` as breaking change.

.github/workflows/api-compatibility.yml
itsharshvb: no, removed `router` crate from path as `api_models` is used for api type definition.

SanchithHegde (dismissed):

.github/workflows/api-migrations-compatibility.yml
```diff
@@ -0,0 +1,309 @@
...
+          fi
+
+          BREAKING=0
+          WARNINGS=0
+
+          while IFS= read -r file; do
+            [[ -z "$file" ]] && continue
+            CONTENT=$(cat "$file")
+
+            if echo "$CONTENT" | grep -iE "(DROP TABLE|DROP COLUMN|DELETE FROM|TRUNCATE)" >/dev/null; then
```
hrithikesh026: Wouldn't this fail if someone drops column like this `DROP       COLUM` . multiple empty spaces between DROP and COLUMN. It would be better to use regex matching.

Also there are other queries that are backwards incompatible. Like Making a NULLABLE column NOT NULL. Renaming columns. changing type of a column renaming table, Droping indexes etc

hrithikesh026: We would need an implementation where we could easily introduce a new rule based on our learnings in future.

itsharshvb: Added regex pattern filter for multiple empty spaces. 
Added support for `migration-rules.yml` for easier rule addition and migration wise classification.

.github/api-migration-compatibility/migration-rules.yaml
```diff
@@ -0,0 +1,108 @@
...
+
+  rename_column:
+    pattern: 'RENAME\s+COLUMN\s+'
+    description: "Renaming a column breaks applications using the old name"
+
+  set_not_null:
+    pattern: 'ALTER\s+COLUMN\s+.*\s+SET\s+NOT\s+NULL'
+    description: "Making an existing nullable column NOT NULL can fail with existing NULL data"
+
+  add_column_not_null_no_default:
```
hrithikesh026, line 35: Not null columns should never be introduced in existing tables. But can be introduced in a new table.

.github/api-migration-compatibility/migration-rules.yaml
```diff
@@ -0,0 +1,108 @@
...
+      - drop_table
+      - drop_column
+      - delete_data
+      - truncate_table
+      - rename_table
+      - rename_column
+      - set_not_null
+      - add_column_not_null_no_default
+    warnings:
+      - alter_column_type
```
hrithikesh026, line 69: alter_column_type should be breaking

.github/workflows/api-migrations-compatibility.yml
```diff
@@ -0,0 +1,386 @@
...
+      - name: Fetch base branch
+        env:
+          BASE_REF: ${{ github.event.pull_request.base.ref }}
+        run: |
+          git fetch origin "$BASE_REF:$BASE_REF"
+
+      - name: Validate V1 migrations
+        id: migration_check
+        run: |
+          BASE_REF="origin/${{ github.event.pull_request.base.ref }}"
```
hrithikesh026: this shell script can be moved to .github/scripts ig.

.github/api-migration-compatibility/migration-rules.yaml
itsharshvb: added `ALTER\s+TABLE\s+.` to the query pattern

hrithikesh026 (dismissed):

.github/scripts/validate_migrations.sh
```diff
@@ -0,0 +1,81 @@
...
+
+while IFS= read -r file; do
+  [[ -z "$file" ]] && continue
+
+  # Check for breaking changes
+  if grep -iEq "$BREAKING_PATTERNS" "$file" 2>/dev/null; then
+    echo "BREAKING: $file"
+    MATCH_COUNT=$(grep -icE "$BREAKING_PATTERNS" "$file" 2>/dev/null || echo "0")
+    grep -inE "$BREAKING_PATTERNS" "$file" 2>/dev/null | sed 's/^/  /' || true
+    BREAKING=$((BREAKING + MATCH_COUNT))
```
SanchithHegde: Nit: Prefer to use the longer variant of the flags (for example, passed to the `grep` command) over the short ones.

.github/scripts/validate_migrations.sh
```diff
@@ -0,0 +1,81 @@
...
+  fi
+
+  # Check for warnings
+  if grep -iEq "$WARNING_PATTERNS" "$file" 2>/dev/null; then
+    echo "WARNING: $file"
+    MATCH_COUNT=$(grep -icE "$WARNING_PATTERNS" "$file" 2>/dev/null || echo "0")
+    grep -inE "$WARNING_PATTERNS" "$file" 2>/dev/null | sed 's/^/  /' || true
```
SanchithHegde: Same here.

SanchithHegde (commented):

PR #9816, opened 2025-10-13
feat(framework): Diesel and domain model changes to support multiple outgoing webhooks

swetasharma03:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

Added the `multiple_webhook_detail` field to the Profile DB and domain model. Also populated these fields to as `None`.
Also we need `webhook_endpoint_id`, added support for this idType as well in this PR.

### Additional Changes

## How did you test it?

Only domain and diesel model changes (addition of fields only and no values populated), hence no testing required.

## Checklist

### DIFF
--- a/crates/hyperswitch_domain_models/src/business_profile.rs
+++ b/crates/hyperswitch_domain_models/src/business_profile.rs
@@ -16,7 +16,7 @@ use diesel_models::business_profile::RevenueRecoveryAlgorithmData;
 use diesel_models::business_profile::{
     self as storage_types, AuthenticationConnectorDetails, BusinessPaymentLinkConfig,
     BusinessPayoutLinkConfig, CardTestingGuardConfig, ExternalVaultConnectorDetails,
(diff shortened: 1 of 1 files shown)


crates/common_enums/src/enums.rs
```diff
@@ -1640,6 +1640,30 @@ pub enum WebhookDeliveryAttempt {
...
+    Debug,
+    Eq,
+    PartialEq,
+    serde::Deserialize,
+    serde::Serialize,
+    strum::Display,
+    strum::EnumString,
+    ToSchema,
+)]
+#[router_derive::diesel_enum(storage_type = "db_enum")]
```
SanchithHegde: We can remove this attribute macro / annotation if we are not using this enum within a column of a database table.

PR #8848, opened 2025-08-06
feat(revenue_recovery):- Introducing new calculate job for card switching and invoice queueing

AdityaKumaar21:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

### Additional Changes

- [ ] This PR modifies the API contract
- [ ] This PR modifies the database schema
- [ ] This PR modifies application configuration/environment variables

## Checklist

- [ ] I formatted the code `cargo +nightly fmt --all`
- [ ] I addressed lints thrown by `cargo clippy`
- [ ] I reviewed the submitted code
- [ ] I added unit tests for my changes where possible

PR #9935, opened 2025-10-21
fix(revenue-recovery): Ensure payment attempt Insert and proxy call use the same payment processor token

aniketburman014:
## Type of Change

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

This PR fixes an issue where multiple tokens were marked as active for the same customer due to concurrent Redis updates and redundant fetches during payment processing.

---

### **Problem**
- The code was fetching tokens from Redis **twice**:
  1. Once while creating the payment attempt.  
  2. Again while calling the proxy.  
- Because of this, when multiple tokens were active (due to an earlier update failure), both calls could randomly pick different tokens, leading to inconsistent token usage within a single payment flow.  
- Additionally, the activation logic did not reliably ensure that only one token remained active at a time.

---

### **Fix Implemented**
1. **Single Redis Fetch:**  
   - Fetch token details **only once** during payment attempt creation.  
   - Pass the fetched token as an argument when calling the proxy instead of fetching again.

2. **Safe Activation Logic:**  
   - When setting a token as active, first mark **all existing tokens as inactive**.  
   - Then, mark only the selected token as active, ensuring a single active token per customer at any time.

---

### Additional Changes

## How did you test it?

For testing refer to this : https://github.com/juspay/hyperswitch/pull/8848

## Checklist

### DIFF
--- a/crates/router/src/core/revenue_recovery/types.rs
+++ b/crates/router/src/core/revenue_recovery/types.rs
@@ -454,7 +454,7 @@ impl Action {
                             e
                         );
                     };
-        
+
                     // update the status of token in redis
                     let _update_error_code = storage::revenue_recovery_redis_operation::RedisTokenManager::update_payment_processor_token_error_code_from_process_tracker(
                         state,


srujanchikke (dismissed):

crates/router/src/core/revenue_recovery.rs
```diff
@@ -210,25 +213,55 @@ pub async fn perform_execute_payment(
...
-                })?
-                .ok_or(errors::ApiErrorResponse::GenericNotFoundError {
-                    message: "Failed to fetch token details from redis".to_string(),
+            )
+            .await
+            .change_context(errors::ApiErrorResponse::GenericNotFoundError {
+                message: "Failed to fetch token details from redis".to_string(),
             })?;
+
+            let payment_processor_token = match processor_token {
```
hrithikesh026: this match is not needed.

```suggestion
            if processor_token.is_none() {
```

PR #10145, opened 2025-11-05
feat(oidc): Add OIDC infrastructure and discovery endpoints

kanikabansal08:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

This PR introduces the foundational infrastructure for OpenID Connect (OIDC) provider support in Hyperswitch. It implements:

1. A runtime config section under [oidc] for signing keys [oidc.key.k1] and [oidc.key.k2] and clients [oidc.client.c1].
2. Public endpoints: /.well-known/openid-configuration (Discovery) and /oauth2/jwks (JWKS).
3. Configuration parsing logic to load keys and clients from TOML.
4. Basic scaffolding for future authorization/token flows (signing keys publication etc.).

This change does not yet implement the authorization endpoint (/oauth2/authorize), token endpoint (/oauth2/token), or full user session/consent flows — those will be covered in follow-up work.

### Additional Changes

## Motivation and Context

We are building the OIDC provider capability so that Hyperswitch can act as an identity issuer for external apps (Relying Parties).
Enabling this now allows us to:
1. Publish JWKS so external RPs can verify tokens.
2. Support discovery for dynamic configuration (issuer, endpoints).

## How did you test it?

1. Manual testing of /.well-known/openid-configuration: verified correct issuer, token_endpoint (placeholder), jwks_uri.
2. Manual testing of /oauth2/jwks: verified RSA public keys are published with correct kid, n, e.

1. Use below commands to create private key on terminal.
```
openssl genrsa -out private_key2.pem 2048
```
we can have 2 set of [oidc.keys].

2. Use `openssl rand -hex 16` to generate kid and `openssl rand -hex 8` client id for testing purposes.
3. Cargo r
4.  Discovery Endpoint:
```
curl --location 'http://localhost:8080/.well-known/openid-configuration'
```

Response:
```{
    "issuer": "http://localhost:8080",
    "authorization_endpoint": "http://localhost:8080/oauth2/authorize",
    "token_endpoint": "http://localhost:8080/oauth2/token",
    "jwks_uri": "http://localhost:8080/oauth2/jwks",
    "response_types_supported": [
        "code"
    ],
    "response_modes_supported": [
        "query"
    ],
    "subject_types_supported": [
        "public"
    ],
    "id_token_signing_alg_values_supported": [
        "RS256"
    ],
    "grant_types_supported": [
        "authorization_code"
    ],
    "scopes_supported": [
        "openid",
        "email"
    ],
    "token_endpoint_auth_methods_supported": [
        "client_secret_basic"
    ],
    "claims_supported": [
        "aud",
        "email",
        "email_verified",
        "exp",
        "iat",
        "iss",
        "sub"
    ]
}
```

2. JWKS URI
```
curl --location 'http://localhost:8080/oauth2/jwks'
```
response:
```
{
    "keys": [
        {
            "kty": "RSA",
            "kid": "27e0c798ee168e5af39f3a0ba8ebde06",
            "use": "sig",
            "alg": "RS256",
            "n": "tDYApck1RWUosF632govl01qpn6YCxOf5SbxfeVKQi8JsOr6klpWP93NV-P4Pv5xwPLxaUq1qYvSmwVCKmrdWuDA0wy_VOqHSjXROc1XBnRvSZutdt6-ir5-zT4isxVXB6n9rGZIlGUT7pKp0RHWb1AEQ1BYI80E8IJQuNc-2i4385CnVk60GhGlpxRZRxeZz75zB-01Gh_y8C34NDp9a8PpZ0YelMs_0kBDhfjwZYQlzfiRNdJQzA3nrMYzHY-4zYPkgO_2iYecQ6Lq4k-uimL-_V6yhv2jJ8H2uj6MuPQym36E0s-Ah9sNqahGNGOjV5Nu4A7PNYfpYqrr4AWcdQ",
            "e": "AQAB"
        },
        {
            "kty": "RSA",
            "kid": "032786b1e0e2cda1785fdf98c07cc062",
            "use": "sig",
            "alg": "RS256",
            "n": "wiJMLNCqT4SQoqn_CsKrViMOeKmU7xirW-_eQ4f4a4Pa2t8EC1_v7PyG-ToTnpJtPpdoJWFMUZDjT_k4NV1wEm-rSwYVVENK5b_W6DkYGSCa8twqy3pBQ0gj77aBxiP5ADvPZXIbqw-Fb4Y1mRPuvNHcwHLgXH0Z0tA5TmZ9XarqQiYKgBzs9sG604wVokhWwlWOf9NCRb1ZbdpbeaO-O-L0qK9YyeLKBDDeMjMP6NXeHCzx0hrb0z8J3oo4TtUK9g0HjWtNNYZm7egEeoTCnWEtKL4OqC5e9tAmQMb9M37VeI1TBwrac51SnCRn8cbLAFgkv1xfm1_YmHUBA_9pZw",
            "e": "AQAB"
        }
    ]
}
```

## Checklist

### DIFF
--- a/crates/router/src/routes/app.rs
+++ b/crates/router/src/routes/app.rs
@@ -442,14 +442,14 @@ impl AppState {
             let cache_store = get_cache_store(&conf.clone(), shut_down_signal, testable)
                 .await
                 .expect("Failed to create store");
-            let global_store: Box<dyn GlobalStorageInterface> = Self::get_store_interface(
+            let global_store: Box<dyn GlobalStorageInterface> = Box::pin(Self::get_store_interface(
                 &storage_impl,
                 &event_handler,
                 &conf,
                 &conf.multitenancy.global_tenant,
                 Arc::clone(&cache_store),
                 testable,
-            )
+            ))
             .await
             .get_global_storage_interface();
             #[cfg(feature = "olap")]
--- a/crates/router/src/configs/settings.rs
+++ b/crates/router/src/configs/settings.rs
@@ -278,14 +278,14 @@ impl TenantConfig {
             .await
             .expect("Failed to create event handler");
         futures::future::join_all(self.0.iter().map(|(tenant_name, tenant)| async {
-            let store = AppState::get_store_interface(
+            let store = Box::pin(AppState::get_store_interface(
                 storage_impl,
                 &event_handler,
                 conf,
                 tenant,
                 cache_store.clone(),
                 testable,
-            )
+            ))
             .await
             .get_storage_interface();
             (tenant_name.clone(), store)
@@ -311,14 +311,14 @@ impl TenantConfig {
             .await
             .expect("Failed to create event handler");
         futures::future::join_all(self.0.iter().map(|(tenant_name, tenant)| async {
-            let store = AppState::get_store_interface(
+            let store = Box::pin(AppState::get_store_interface(
                 storage_impl,
                 &event_handler,
                 conf,
                 tenant,
                 cache_store.clone(),
                 testable,
-            )
+            ))
             .await
             .get_accounts_storage_interface();
             (tenant_name.clone(), store)


crates/api_models/src/oidc.rs
```diff
@@ -0,0 +1,339 @@
...
+    /// RSA public key modulus
+    pub n: String,
+
+    /// RSA public key exponent
+    #[schema(example = "AQAB")]
+    pub e: String,
+}
+
+/// Custom deserializer for space-separated scope strings
+fn deserialize_scope_vec<'de, D>(deserializer: D) -> Result<Vec<Scope>, D::Error>
```
kanikabansal08, line 251: In oidc/auth, the `scope` comes in as one space-separated string (`scope=openid email`) so this custom deserializer splits the string and maps each value into our `Scope` enum. This keeps the request format spec-compliant.

crates/router/src/services/authentication.rs
```diff
@@ -416,6 +471,36 @@ where
     }
 }
 
+#[async_trait]
```
kanikabansal08, line 522: This adds a reusable `Basic Authentication ( username:password)`  that providers can plug into for their own credential lookup. It centralizes parsing + verification. This is required for token flow of OIDC.

crates/router/src/services/oidc_provider.rs
```diff
@@ -0,0 +1,54 @@
...
+
+use crate::{
+    core::errors::{ApiErrorResponse, RouterResponse},
+    routes::app::SessionState,
+    services::api::ApplicationResponse,
+    utils::user::get_base_url,
+};
+
+/// Build OIDC discovery document
+pub async fn get_discovery_document(state: SessionState) -> RouterResponse<OidcDiscoveryResponse> {
```
kanikabansal08, line 15: We deliberately do not cache the OIDC discovery document using oncecell because the issuer and authorization_endpoint / token_endpoint / jwks_uri are derived from `state.tenant.base_url and get_base_url(&state)`, which vary per tenant in a multi-tenant deployment.

config/deployments/integration_test.toml
SanchithHegde, line 1: Please revert the changes to the integ, sandbox and production TOML files, and instead add it in the `config/deployments/env_specific.toml` file.

Also, would be great if you could include some instructions / commands to generate some of these values, in comments. Please include the documentation comments in the `config/deployments/env_specific.toml` and `config/config.example.toml` file.

crates/api_models/src/oidc.rs
```diff
@@ -0,0 +1,339 @@
...
+    Claim::Exp,
+    Claim::Iat,
+    Claim::Iss,
+    Claim::Sub,
+];
+
+/// OIDC Response Type
+#[derive(Clone, Debug, serde::Serialize, serde::Deserialize)]
+#[serde(rename_all = "snake_case")]
+pub enum ResponseType {
```
SanchithHegde: Nit: You can derive `Copy` on this.

Same applies to all the other enums in this file with unit variants (no data / fields).

crates/api_models/src/oidc.rs
```diff
@@ -0,0 +1,339 @@
...
+
+    /// Algorithm
+    pub alg: SigningAlgorithm,
+
+    /// RSA public key modulus
+    pub n: String,
+
+    /// RSA public key exponent
+    #[schema(example = "AQAB")]
+    pub e: String,
```
SanchithHegde, line 247: Would it be a concern if these two values get logged in plaintext? Or should we consider wrapping them within a `Secret`?

crates/api_models/src/oidc.rs
```diff
@@ -0,0 +1,339 @@
...
+    #[serde(deserialize_with = "deserialize_scope_vec")]
+    pub scope: Vec<Scope>,
+
+    /// Opaque value used to maintain state between request and callback
+    #[schema(example = "state_xyz789")]
+    pub state: String,
+
+    /// String value used to associate a Client session with an ID Token
+    #[schema(example = "nonce_abc123")]
+    pub nonce: Option<String>,
```
SanchithHegde, line 289: Is this field considered sensitive / should we wrap it within a `Secret`?

crates/api_models/src/oidc.rs
```diff
@@ -0,0 +1,339 @@
...
+
+/// OIDC Token Request
+#[derive(Debug, Clone, Serialize, Deserialize, ToSchema)]
+pub struct OidcTokenRequest {
+    /// OAuth 2.0 Grant Type value
+    pub grant_type: GrantType,
+
+    /// Authorization code received from the authorization server
+    #[schema(example = "auth_code_xyz789")]
+    pub code: String,
```
SanchithHegde, line 300: Is this field considered sensitive?

crates/api_models/src/oidc.rs
```diff
@@ -0,0 +1,339 @@
...
+    /// OAuth 2.0 Client Identifier
+    #[schema(example = "client_abc123")]
+    pub client_id: String,
+}
+
+/// OIDC Token Response
+#[derive(Debug, Clone, Serialize, Deserialize, ToSchema)]
+pub struct OidcTokenResponse {
+    /// ID Token value associated with the authenticated session
+    pub id_token: String,
```
SanchithHegde: Same here, is this field considered sensitive?

crates/api_models/src/oidc.rs
```diff
@@ -0,0 +1,339 @@
...
+    #[schema(example = 3600)]
+    pub expires_in: u64,
+}
+
+impl ApiEventMetric for OidcDiscoveryResponse {}
+impl ApiEventMetric for JwksResponse {}
+impl ApiEventMetric for OidcAuthorizeQuery {}
+impl ApiEventMetric for AuthCodeData {}
+impl ApiEventMetric for OidcTokenRequest {}
+impl ApiEventMetric for OidcTokenResponse {}
```
SanchithHegde: I think we can have an enum variant `ApiEventsType::Oidc` maybe, and have these implementations return that variant. The [default implementation returns `None`](https://github.com/juspay/hyperswitch/blob/a690a40bab3deae2f79ff9717b8eee81bec29462/crates/common_utils/src/events.rs#L6-L8), if we don't override it.

crates/hyperswitch_domain_models/src/errors/api_error_response.rs
```diff
@@ -224,6 +224,11 @@ pub enum ApiErrorResponse {
         message = "Access forbidden, invalid JWT token was used"
     )]
     InvalidJwtToken,
+    #[error(
+        error_type = ErrorType::InvalidRequestError, code = "IR_17_1",
```
SanchithHegde: Can we not do this please? Let the error code be 2 digit numbers, if possible.

crates/router/src/configs/secrets_transformers.rs
```diff
@@ -524,6 +524,7 @@ pub(crate) async fn fetch_raw_secrets(
         #[cfg(feature = "email")]
         email: conf.email,
         user: conf.user,
+        oidc: conf.oidc,
```
SanchithHegde: You would need to call `convert_to_raw_secret()` if you need any values in the struct to be KMS decrypted at app startup.

crates/router/src/routes/app.rs
```diff
@@ -1780,6 +1780,20 @@ impl Hypersense {
     }
 }
 
+pub struct Oidc;
+
+impl Oidc {
+    pub fn server(state: AppState) -> Scope {
+        web::scope("")
```
SanchithHegde, line 1795: Is it intentional that we aren't specifying any prefix here?

config/deployments/integration_test.toml
kanikabansal08: My understanding is that `env_specific.toml` is used for shared defaults, so let me know if that’s incorrect. Since OIDC credentials will differ across environments (different client IDs, redirect URIs, issuers, etc.), I added them to the environment-specific files.

I’ve already mentioned the commands to generate these values in the PR description, and I’ll add complete documentation comments as well in `config/config.example.toml`, as suggested.

crates/api_models/src/oidc.rs
kanikabansal08: These are public key components, so they aren’t sensitive.
They’re expected to be exposed anyway (through the `JWKS` endpoint), so logging them in plaintext shouldn’t be a concern.

crates/api_models/src/oidc.rs
kanikabansal08: This value isn’t sensitive.
The nonce is expected to be present in the browser URL as part of the OIDC flow, so wrapping it in a Secret isn’t necessary.

crates/api_models/src/oidc.rs
kanikabansal08: No

crates/router/src/routes/app.rs
kanikabansal08: yes

config/deployments/integration_test.toml
kanikabansal08: As discussed, I have updated the `env_specific.toml`.

crates/api_models/src/oidc.rs
```diff
@@ -0,0 +1,372 @@
...
+    pub state: String,
+
+    /// String value used to associate a Client session with an ID Token
+    #[schema(example = "nonce_abc123")]
+    pub nonce: Option<String>,
+}
+
+/// Authorization Code Data stored in Redis
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct AuthCodeData {
```
racnan: Does not belong in this crate.

crates/api_models/src/oidc.rs
```diff
@@ -0,0 +1,372 @@
...
+    #[serde(default)]
+    #[schema(example = false)]
+    pub request_uri_parameter_supported: bool,
+}
+
+impl OidcDiscoveryResponse {
+    pub fn new(issuer: String, control_center_url: String) -> Self {
+        let authorization_endpoint = format!("{}/oauth2/authorize", control_center_url);
+        let token_endpoint = format!("{}/oauth2/token", issuer);
+        let jwks_uri = format!("{}/oauth2/jwks", issuer);
```
SanchithHegde, line 196: Would we want to hardcode these paths always?

crates/router/src/services/oidc_provider.rs
```diff
@@ -0,0 +1,54 @@
...
+    let backend_base_url = state.tenant.base_url.clone();
+    let control_center_url = get_base_url(&state);
```
SanchithHegde: Nit: Would have preferred qualifying this with module name instead, such as `utils::get_base_url()` or `user_utils::get_base_url()`.

SanchithHegde (commented):

crates/api_models/src/oidc.rs
kanikabansal08: These are fixed endpoint paths for our OIDC surface (/oauth2/authorize, /oauth2/token, /oauth2/jwks). OIDC discovery expects these exact URLs

racnan (approved):

SanchithHegde (dismissed):

tsdk02 (dismissed):

SanchithHegde (approved):

racnan (approved):

PR #10147, opened 2025-11-06
feat(oidc): Implement OIDC authorize and token flows

kanikabansal08:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

Implemented auth and token endpoints for oidc.
- Internal auth endpoint oidc/authorize
- Token Endpoint /oauth2/token

### Additional Changes

## Motivation and Context
Requirement. In continuation of #10145 

## How did you test it?

1. Authorization request:
```sh
curl --location 'http://localhost:8080/oidc/authorize' \
--header 'Authorization: Bearer ******' \
--data '{
  "response_type": "code",
  "client_id": "****",
  "redirect_uri": "https://www.google.com",
  "scope": "openid email",
  "state": "UzaI****iIWTw-AfOcnf",
  "nonce": "KYut****k7CU5WuaNMA"
}'
```
2. Token request:
```sh
curl --location 'http://localhost:8080/oauth2/token' \
--header 'Content-Type: application/x-www-form-urlencoded' \
--header 'Authorization: Basic *******Q=' \
--data-urlencode 'client_id=a5*******5824f9' \
--data-urlencode 'grant_type=authorization_code' \
--data-urlencode 'code=CJejU8n**********U8VXyAelO' \
--data-urlencode 'redirect_uri=https://www.google.com'
```
Response:
1. 302 to REDIRECT URI with `"https://www.google.com/?code=CJen***********EU8VXyAelO&state=Uza*******w-AfOcnf"`

2. 
```json
{
    "id_token": ***,
    "token_type": "Bearer",
    "expires_in": 3600
}
```
## Checklist

### DIFF
--- a/crates/router/src/utils.rs
+++ b/crates/router/src/utils.rs
@@ -699,6 +699,18 @@ pub fn get_http_status_code_type(
     Ok(status_code_type.to_string())
 }
 
+// Trims whitespace from a Secret string and returns None if empty
+pub fn trim_secret_string(secret: masking::Secret<String>) -> Option<masking::Secret<String>> {
+    let trimmed = secret.expose().trim().to_string();
+    (!trimmed.is_empty()).then(|| masking::Secret::new(trimmed))
+}
+
+// Trims whitespace from a regular string and returns None if empty
+pub fn trim_string(value: String) -> Option<String> {
+    let trimmed = value.trim().to_string();
+    (!trimmed.is_empty()).then_some(trimmed)
+}
+
 pub fn add_connector_http_status_code_metrics(option_status_code: Option<u16>) {
     if let Some(status_code) = option_status_code {
         let status_code_type = get_http_status_code_type(status_code).ok();
@@ -1023,17 +1035,18 @@ pub fn add_apple_pay_flow_metrics(
 ) {
     if let Some(flow) = apple_pay_flow {
         match flow {
-            domain::ApplePayFlow::Simplified(_) => metrics::APPLE_PAY_SIMPLIFIED_FLOW.add(
-                1,
-                router_env::metric_attributes!(
-                    (
-                        "connector",
-                        connector.to_owned().unwrap_or("null".to_string()),
+            domain::ApplePayFlow::DecryptAtApplication(_) => metrics::APPLE_PAY_SIMPLIFIED_FLOW
+                .add(
+                    1,
+                    router_env::metric_attributes!(
+                        (
+                            "connector",
+                            connector.to_owned().unwrap_or("null".to_string()),
+                        ),
+                        ("merchant_id", merchant_id.clone()),
                     ),
-                    ("merchant_id", merchant_id.clone()),
                 ),
-            ),
-            domain::ApplePayFlow::Manual => metrics::APPLE_PAY_MANUAL_FLOW.add(
+            domain::ApplePayFlow::SkipDecryption => metrics::APPLE_PAY_MANUAL_FLOW.add(
                 1,
                 router_env::metric_attributes!(
                     (
@@ -1056,7 +1069,7 @@ pub fn add_apple_pay_payment_status_metrics(
     if payment_attempt_status == enums::AttemptStatus::Charged {
         if let Some(flow) = apple_pay_flow {
             match flow {
-                domain::ApplePayFlow::Simplified(_) => {
+                domain::ApplePayFlow::DecryptAtApplication(_) => {
                     metrics::APPLE_PAY_SIMPLIFIED_FLOW_SUCCESSFUL_PAYMENT.add(
                         1,
                         router_env::metric_attributes!(
@@ -1068,8 +1081,8 @@ pub fn add_apple_pay_payment_status_metrics(
                         ),
                     )
                 }
-                domain::ApplePayFlow::Manual => metrics::APPLE_PAY_MANUAL_FLOW_SUCCESSFUL_PAYMENT
-                    .add(
+                domain::ApplePayFlow::SkipDecryption => {
+                    metrics::APPLE_PAY_MANUAL_FLOW_SUCCESSFUL_PAYMENT.add(
                         1,
                         router_env::metric_attributes!(
                             (
@@ -1078,13 +1091,14 @@ pub fn add_apple_pay_payment_status_metrics(
                             ),
                             ("merchant_id", merchant_id.clone()),
                         ),
-                    ),
+                    )
+                }
             }
         }
     } else if payment_attempt_status == enums::AttemptStatus::Failure {
         if let Some(flow) = apple_pay_flow {
             match flow {
-                domain::ApplePayFlow::Simplified(_) => {
+                domain::ApplePayFlow::DecryptAtApplication(_) => {
                     metrics::APPLE_PAY_SIMPLIFIED_FLOW_FAILED_PAYMENT.add(
                         1,
                         router_env::metric_attributes!(
@@ -1096,16 +1110,18 @@ pub fn add_apple_pay_payment_status_metrics(
                         ),
                     )
                 }
-                domain::ApplePayFlow::Manual => metrics::APPLE_PAY_MANUAL_FLOW_FAILED_PAYMENT.add(
-                    1,
-                    router_env::metric_attributes!(
-                        (
-                            "connector",
-                            connector.to_owned().unwrap_or("null".to_string()),
+                domain::ApplePayFlow::SkipDecryption => {
+                    metrics::APPLE_PAY_MANUAL_FLOW_FAILED_PAYMENT.add(
+                        1,
+                        router_env::metric_attributes!(
+                            (
+                                "connector",
+                                connector.to_owned().unwrap_or("null".to_string()),
+                            ),
+                            ("merchant_id", merchant_id.clone()),
                         ),
-                        ("merchant_id", merchant_id.clone()),
-                    ),
-                ),
+                    )
+                }
             }
         }
     }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,20 @@ All notable changes to HyperSwitch will be documented here.
 
 - - -
 
+## 2026.01.14.1
+
+### Features
+
+- **Wallets:** APPLEPAY AND GOOGLE PAY DECRYPTED FLOW support ([#10329](https://github.com/juspay/hyperswitch/pull/10329)) ([`d36b783`](https://github.com/juspay/hyperswitch/commit/d36b7836cb107076e7e43810d1ae1b53e87aad4f))
+
+### Bug Fixes
+
+- **connector:** Handle Mollie failed payments with error details from details object ([#10824](https://github.com/juspay/hyperswitch/pull/10824)) ([`8429e48`](https://github.com/juspay/hyperswitch/commit/8429e486c5edcec3ecc03b0d4e87aa215fbbeab5))
+
+**Full Changelog:** [`2026.01.14.0...2026.01.14.1`](https://github.com/juspay/hyperswitch/compare/2026.01.14.0...2026.01.14.1)
+
+- - -
+
 ## 2026.01.14.0
 
 ### Features
--- a/crates/api_models/src/payments.rs
+++ b/crates/api_models/src/payments.rs
@@ -9658,9 +9658,20 @@ pub struct GpayMetaData {
 }
 
 #[derive(Debug, Clone, serde::Serialize, serde::Deserialize)]
+pub struct GooglePayDetailsWrapper {
+    #[serde(flatten)]
+    pub data: Option<GpayMetaData>,
+    support_predecrypted_token: Option<bool>,
+}
+impl GooglePayDetailsWrapper {
+    pub fn is_predecrypted_token_supported(&self) -> bool {
+        self.support_predecrypted_token.unwrap_or(false)
+    }
+}
+
+#[derive(Debug, Clone, serde::Serialize, serde::Deserialize)]
 pub struct GpaySessionTokenData {
-    #[serde(rename = "google_pay")]
-    pub data: GpayMetaData,
+    pub google_pay: GooglePayDetailsWrapper,
 }
 
 #[derive(Debug, Clone, serde::Serialize, serde::Deserialize)]
@@ -9826,14 +9837,35 @@ pub struct ApplepaySessionTokenData {
 }
 
 #[derive(Debug, Clone, serde::Serialize, serde::Deserialize)]
+pub struct ApplePayCombinedWrapper {
+    #[serde(flatten)]
+    pub data: Option<ApplePayCombinedMetadata>,
+    support_predecrypted_token: Option<bool>,
+}
+impl ApplePayCombinedWrapper {
+    pub fn is_predecrypted_token_supported(&self) -> bool {
+        self.support_predecrypted_token.unwrap_or(false)
+    }
+    pub fn get_combined_metadata_required(
+        &self,
+    ) -> Result<ApplePayCombinedMetadata, ValidationError> {
+        self.data
+            .clone()
+            .ok_or(ValidationError::IncorrectValueProvided {
+                field_name: "metadata.apple_pay_combined",
+            })
+    }
+}
+
+#[derive(Debug, Clone, serde::Serialize, serde::Deserialize)]
 pub struct ApplepayCombinedSessionTokenData {
-    pub apple_pay_combined: ApplePayCombinedMetadata,
+    pub apple_pay_combined: ApplePayCombinedWrapper,
 }
 
 #[derive(Debug, Clone, serde::Serialize, serde::Deserialize)]
 #[serde(rename_all = "snake_case")]
 pub enum ApplepaySessionTokenMetadata {
-    ApplePayCombined(ApplePayCombinedMetadata),
+    ApplePayCombined(ApplePayCombinedWrapper),
     ApplePay(ApplePayMetadata),
 }
 
--- a/crates/hyperswitch_connectors/src/connectors/bluesnap/transformers.rs
+++ b/crates/hyperswitch_connectors/src/connectors/bluesnap/transformers.rs
@@ -461,17 +461,20 @@ impl TryFrom<&types::PaymentsSessionRouterData> for BluesnapCreateWalletToken {
         let applepay_metadata = apple_pay_metadata
             .clone()
             .parse_value::<ApplepayCombinedSessionTokenData>("ApplepayCombinedSessionTokenData")
+            .change_context(errors::ConnectorError::ParsingFailed)
             .map(|combined_metadata| {
-                ApplepaySessionTokenMetadata::ApplePayCombined(combined_metadata.apple_pay_combined)
+                ApplepaySessionTokenMetadata::ApplePayCombined(
+                    combined_metadata.apple_pay_combined.clone(),
+                )
             })
             .or_else(|_| {
                 apple_pay_metadata
                     .parse_value::<ApplepaySessionTokenData>("ApplepaySessionTokenData")
+                    .change_context(errors::ConnectorError::ParsingFailed)
                     .map(|old_metadata| {
                         ApplepaySessionTokenMetadata::ApplePay(old_metadata.apple_pay)
                     })
-            })
-            .change_context(errors::ConnectorError::ParsingFailed)?;
+            })?;
         let session_token_data = match applepay_metadata {
             ApplepaySessionTokenMetadata::ApplePay(apple_pay_data) => {
                 Ok(apple_pay_data.session_token_data)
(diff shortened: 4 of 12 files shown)


crates/router/src/services/oidc_provider.rs
```diff
@@ -0,0 +1,414 @@
...
+            }));
+        }
+        Some(user) => user,
+    };
+
+    let auth_code = generate_and_store_authorization_code(&state, &user.user_id, &payload).await?;
+
+    let redirect_url = build_oidc_redirect_url(&payload.redirect_uri, &auth_code, &payload.state)?;
+
+    #[cfg(feature = "v1")]
```
kanikabansal08: Since `JsonForRedirection` is version specific response I have wrapped it under feature flag.

crates/hyperswitch_domain_models/src/errors/api_error_response.rs
```diff
@@ -312,6 +315,16 @@ pub enum ApiErrorResponse {
         message = "Access forbidden, invalid Basic authentication credentials"
     )]
     InvalidBasicAuth,
+    #[error(error_type = ErrorType::InvalidRequestError, code = "IR_52", message = "{error}: {description}")]
+    OidcAuthorizationError {
+        error: OidcAuthorizationError,
+        description: String,
```
racnan: This can be derived from above error type.

crates/router/src/utils.rs
```diff
@@ -4,9 +4,9 @@ pub mod connector_onboarding;
 pub mod currency;
 pub mod db_utils;
 pub mod ext_traits;
+pub mod oidc;
 #[cfg(feature = "kv_store")]
 pub mod storage_partitioning;
-#[cfg(feature = "olap")]
 pub mod user;
```
SanchithHegde, line 10: Why are we removing this feature flag?

We should probably consider moving the `oidc` module(s) under the `olap` feature flag as well.

crates/router/src/utils/oidc.rs
```diff
@@ -0,0 +1,160 @@
...
+    utils::user as user_utils,
+};
+
+fn get_auth_code_key(auth_code: &str) -> String {
+    format!("{REDIS_AUTH_CODE_PREFIX}{auth_code}")
+}
+
+pub fn validate_client_id_match(registered_client_id: &str, provided_client_id: &str) -> bool {
+    registered_client_id.trim() == provided_client_id.trim()
+}
```
SanchithHegde, line 23: Nit: Can add `#[inline]` annotation on these I feel.

crates/router/src/utils/oidc.rs
```diff
@@ -0,0 +1,160 @@
...
+
+    match (Url::parse(registered), Url::parse(provided)) {
+        (Ok(registered_url), Ok(provided_url)) => registered_url == provided_url,
+        _ => registered == provided,
+    }
+}
+
+pub fn build_oidc_redirect_url(
+    redirect_uri: &str,
+    auth_code: &str,
```
SanchithHegde, line 37: Should `auth_code` be wrapped in `Secret`?

crates/router/src/services/oidc_provider.rs
```diff
@@ -53,3 +64,250 @@ pub async fn get_jwks(state: SessionState) -> RouterResponse<&'static JwksRespon
...
+}
+
+pub async fn process_authorize_request(
+    state: SessionState,
+    payload: OidcAuthorizeQuery,
+    user: Option<UserFromToken>,
+) -> RouterResponse<()> {
+    // Validate all parameters
+    validate_authorize_params(&payload, &state)?;
+    // TODO: Handle users who are not already logged in
```
SanchithHegde, line 127: What do we plan to do about this `TODO`? Have we addressed it, or not yet?

crates/router/src/services/oidc_provider.rs
```diff
@@ -53,3 +64,250 @@ pub async fn get_jwks(state: SessionState) -> RouterResponse<&'static JwksRespon
...
+        oidc_utils::build_oidc_redirect_url(&payload.redirect_uri, &auth_code, &payload.state)?;
+
+    #[cfg(feature = "v1")]
+    {
+        Ok(ApplicationResponse::JsonForRedirection(
+            RedirectionResponse {
+                headers: Vec::with_capacity(0),
+                return_url: String::new(),
+                http_method: String::new(),
+                params: Vec::with_capacity(0),
```
SanchithHegde: Nit: You can also do `Vec::new()`.

crates/router/src/services/oidc_provider.rs
```diff
@@ -53,3 +64,250 @@ pub async fn get_jwks(state: SessionState) -> RouterResponse<&'static JwksRespon
         })
         .map(ApplicationResponse::Json)
 }
+
+pub fn validate_authorize_params(
```
SanchithHegde: Nit: Can we update the visibility of functions to keep them private by default, public (or `pub(crate)`) only if necessary?

crates/router/src/services/oidc_provider.rs
```diff
@@ -53,3 +64,250 @@ pub async fn get_jwks(state: SessionState) -> RouterResponse<&'static JwksRespon
...
+    }
+
+    if !oidc_utils::validate_redirect_uri_match(&auth_code_data.redirect_uri, redirect_uri) {
+        return Err(report!(ApiErrorResponse::OidcTokenError {
+            error: OidcTokenError::InvalidGrant,
+            description: "redirect_uri mismatch".into(),
+        }));
+    }
+
+    oidc_utils::delete_auth_code_from_redis(state, code).await?;
```
SanchithHegde: Would we delete the value from Redis only in case of successful validation / authentication? What if there are some N failed attempts, what is the expected behavior in such a case?

crates/router/src/services/oidc_provider.rs
kanikabansal08: This will be implemented in a later phase of OIDC. Currently, it only supports users already authenticated in HS. OIDC is used to let external services verify the user session and fetch identity details (e.g., email or user ID).
Redirecting unauthenticated users to Hyperswitch for login will be handled later in this `TODO`.

crates/router/src/utils/oidc.rs
kanikabansal08: Not required as it will be already visible in browser's url.

crates/router/src/services/oidc_provider.rs
kanikabansal08: The previous implementation only deleted the auth code after successful validation, which allowed multiple failed attempts and enumeration. The code is now deleted immediately after retrieval (before validation), implementing a consume-on-read pattern.

crates/router/src/configs/settings.rs
```diff
@@ -728,7 +729,13 @@ impl OidcSettings {
...
 
     pub fn get_signing_key(&self) -> Option<&OidcKey> {
-        self.key.values().next()
+        let len = self.key.len();
+        if len == 0 {
+            return None;
+        }
+        let mut rng = rand::thread_rng();
+        let random_index = rng.gen_range(0..len);
+        self.key.values().nth(random_index)
```
SanchithHegde: You can also consider using [`rand::seq::IteratorRandom::choose_stable()`](https://docs.rs/rand/0.8.5/rand/seq/trait.IteratorRandom.html#method.choose_stable) or [`rand::seq::IteratorRandom::choose()`](https://docs.rs/rand/0.8.5/rand/seq/trait.IteratorRandom.html#method.choose) if that helps. Optional change, by the way.

crates/router/src/core/errors/oidc.rs
```diff
@@ -0,0 +1,173 @@
...
+            Self::InvalidClient => "invalid_client",
+            Self::InvalidGrant => "invalid_grant",
+            Self::UnauthorizedClientForGrant => "unauthorized_client",
+            Self::UnsupportedGrantType => "unsupported_grant_type",
+            Self::InvalidTokenScope => "invalid_scope",
+        }
+    }
+
+    /// Returns the RFC 6749 compliant error description
+    pub fn get_error_message(&self) -> String {
```
SanchithHegde: Can't we have the return type as `&'static str`? I don't see any formatting being done here.

crates/router/src/services/oidc_provider.rs
```diff
@@ -49,7 +62,245 @@ pub async fn get_jwks(state: SessionState) -> RouterResponse<&'static JwksRespon
...
+    let user_from_db = state
+        .global_store
+        .find_user_by_id(user_id)
+        .await
+        .change_context(OidcErrors::ServerError)
+        .attach_printable("Failed to fetch user from database")?;
+
+    let user_email = pii::Email::try_from(user_from_db.email.peek().to_string())
+        .change_context(OidcErrors::ServerError)
+        .attach_printable("Failed to parse user email")?;
```
SanchithHegde: Isn't the `user_from_db.email` field already of type `pii::Email`?

crates/router/src/services/oidc_provider.rs
SanchithHegde: One thing to possibly think about is the case where retrieving the auth code fails due some reason, and the deletion would not happen because of the `?` present when calling `get_auth_code_from_redis()`. Could that be a possible concern?

kanikabansal08: If retrieval fails, the client receives an error and cannot proceed. The code may remain in Redis, but it'll expire via TTL. The client cannot use it because they got an error.
Even if it's later retrieved, the client already failed and would need to start over. I think it will not be a concern.

crates/api_models/src/oidc.rs
```diff
@@ -319,7 +295,7 @@ pub struct OidcTokenResponse {
     #[schema(example = "Bearer")]
     pub token_type: String,
 
-    /// Expiration time of the ID Token in seconds since the response was generated
+    /// Expiration time of the Access Token in seconds since the response was generated
```
racnan, line 298: expiration time only for access token ? what about id_token ? is that not part of the protocol ?

crates/router/src/core/errors/oidc.rs
```diff
@@ -0,0 +1,173 @@
+use common_utils::errors::CustomResult;
+
+use crate::services::ApplicationResponse;
+
+pub type OidcResult<T> = CustomResult<T, OidcErrors>;
+pub type OidcResponse<T> = CustomResult<ApplicationResponse<T>, OidcErrors>;
+
+#[derive(Debug, thiserror::Error)]
+pub enum OidcErrors {
```
racnan, line 9: Some of the errors here are not used ? what are the cases where we will use errors like `temporarilyUnavailable` ?

crates/router/src/utils/oidc.rs
```diff
@@ -0,0 +1,135 @@
...
+    state: &str,
+) -> OidcResult<String> {
+    Url::parse(redirect_uri)
+        .map_err(|_| {
+            report!(OidcErrors::ServerError)
```
racnan: this throws `500` right ? is that correct or should it be `400` ?

crates/api_models/src/oidc.rs
kanikabansal08: `expires_in` is only defined for the access token as per `OAuth2/OIDC.`
The ID Token carries its expiry in the exp claim inside the JWT, so a separate expires_in for id_token is not part of the spec.

crates/router/src/core/errors/oidc.rs
kanikabansal08: `temporarily_unavailable` is meant for transient server conditions (overload, maintenance, infra issues) where a retry may succeed.
These errors are defined by RFC 6749 and I kept for spec compliance and future extensibility. If we prefer we can remove unused ones for now and add them later if needed.

racnan (dismissed):

crates/router/src/core/errors/oidc.rs
racnan: if this is infra error, then why is it present in app.

kanikabansal08: makes sense. we can remove `temporarily_unavailable`

crates/router/src/core/errors/oidc.rs
```diff
@@ -0,0 +1,123 @@
...
+    InvalidTokenRequest,
+    #[error("Invalid client")]
+    InvalidClient,
+    #[error("Invalid grant")]
+    InvalidGrant,
+}
+
+impl OidcErrors {
+    /// Returns the RFC 6749 compliant error code
+    pub fn get_rfc_error_code(&self) -> &'static str {
```
SanchithHegde: Nit: `get_rfc_error_code()` and `get_error_message()` can be private if they aren't being used anywhere else.

crates/router/src/services/oidc_provider.rs
```diff
@@ -49,7 +61,237 @@ pub async fn get_jwks(state: SessionState) -> RouterResponse<&'static JwksRespon
...
+
+pub async fn process_authorize_request(
+    state: SessionState,
+    payload: OidcAuthorizeQuery,
+    user: Option<UserFromToken>,
+) -> OidcResponse<()> {
+    // Validate all parameters
+    validate_authorize_params(&payload, &state)?;
+    // TODO: Handle users who are not already logged in
+    let user = user.ok_or_else(|| report!(OidcErrors::AccessDenied))?;
```
sai-harsha-vardhan: Do we want to have more specific error indicating user is missing, or we can have an attach_printable indicating that

crates/router/src/services/oidc_provider.rs
```diff
@@ -49,7 +61,237 @@ pub async fn get_jwks(state: SessionState) -> RouterResponse<&'static JwksRespon
...
+
+    let auth_code = generate_and_store_authorization_code(&state, &user.user_id, &payload).await?;
+
+    let redirect_url =
+        oidc_utils::build_oidc_redirect_url(&payload.redirect_uri, &auth_code, &payload.state)
+            .attach_printable("Failed to build redirect URL")?;
+
+    #[cfg(feature = "v1")]
+    {
```
sai-harsha-vardhan: Instead of having feature flag based branching in code, we can have a fn for both the versions and implementation is abstracted in those fns

crates/router/src/services/oidc_provider.rs
```diff
@@ -49,7 +61,237 @@ pub async fn get_jwks(state: SessionState) -> RouterResponse<&'static JwksRespon
...
+    }
+    let client = state
+        .conf
+        .oidc
+        .get_inner()
+        .get_client(&payload.client_id)
+        .ok_or_else(|| report!(OidcErrors::UnauthorizedClient))?;
+
+    if !oidc_utils::validate_redirect_uri_match(&client.redirect_uri, &payload.redirect_uri) {
+        return Err(report!(OidcErrors::InvalidRequest));
```
sai-harsha-vardhan: Please try to avoid early returns

crates/router/src/services/oidc_provider.rs
```diff
@@ -49,7 +61,237 @@ pub async fn get_jwks(state: SessionState) -> RouterResponse<&'static JwksRespon
...
 
-            Ok::<_, error_stack::Report<ApiErrorResponse>>(JwksResponse { keys })
+            Ok(JwksResponse { keys })
         })
         .map(ApplicationResponse::Json)
 }
+
+fn validate_authorize_params(payload: &OidcAuthorizeQuery, state: &SessionState) -> OidcResult<()> {
+    if !payload.scope.contains(&Scope::Openid) {
+        return Err(report!(OidcErrors::InvalidScope));
```
sai-harsha-vardhan: Please try to avoid early returns

crates/router/src/services/oidc_provider.rs
```diff
@@ -49,7 +61,237 @@ pub async fn get_jwks(state: SessionState) -> RouterResponse<&'static JwksRespon
...
+    }
+}
+
+fn validate_token_request(
+    state: &SessionState,
+    client_id: &str,
+    redirect_uri: &str,
+) -> OidcResult<()> {
+    if redirect_uri.trim().is_empty() {
+        return Err(report!(OidcErrors::InvalidTokenRequest));
```
sai-harsha-vardhan: Do we need to have attach_printable here? Also, avoid early returns

crates/router/src/services/oidc_provider.rs
```diff
@@ -49,7 +61,237 @@ pub async fn get_jwks(state: SessionState) -> RouterResponse<&'static JwksRespon
...
+
+    if let Err(err) = oidc_utils::delete_auth_code_from_redis(state, code).await {
+        logger::warn!(
+            error = ?err,
+            "Failed to delete authorization code from Redis after consumption"
+        );
+    }
+
+    if !oidc_utils::validate_client_id_match(&auth_code_data.client_id, client_id) {
+        return Err(report!(OidcErrors::InvalidGrant));
```
sai-harsha-vardhan: Do we need a logger indicating that if validation failure is due to clied_id or redirect_uri?

crates/router/src/utils/oidc.rs
```diff
@@ -0,0 +1,135 @@
...
+#[inline]
+pub fn validate_client_id_match(registered_client_id: &str, provided_client_id: &str) -> bool {
+    registered_client_id.trim() == provided_client_id.trim()
+}
+
+pub fn validate_redirect_uri_match(registered_uri: &str, provided_uri: &str) -> bool {
+    let registered = registered_uri.trim();
+    let provided = provided_uri.trim();
+
+    match (Url::parse(registered), Url::parse(provided)) {
```
sai-harsha-vardhan, line 29: We can use `.zip` here

crates/router/src/utils/oidc.rs
```diff
@@ -0,0 +1,135 @@
...
+    }
+}
+
+pub fn build_oidc_redirect_url(
+    redirect_uri: &str,
+    auth_code: &str,
+    state: &str,
+) -> OidcResult<String> {
+    Url::parse(redirect_uri)
+        .map_err(|_| {
```
sai-harsha-vardhan, line 41: We may have to log this error

crates/router/src/utils/oidc.rs
```diff
@@ -0,0 +1,135 @@
...
+            report!(OidcErrors::InvalidRequest)
+                .attach_printable("Invalid redirect_uri in OIDC authorize request")
+        })
+        .and_then(|base_url| {
+            Url::parse_with_params(base_url.as_str(), &[("code", auth_code), ("state", state)])
+                .map_err(|_| {
```
sai-harsha-vardhan, line 47: Same here

crates/router/src/utils/oidc.rs
```diff
@@ -0,0 +1,135 @@
...
+    let connection = user_utils::get_redis_connection_for_global_tenant(state)
+        .change_context(OidcErrors::ServerError)
+        .attach_printable("Failed to get redis connection")?;
+    let key = get_auth_code_key(code);
+    connection
+        .delete_key(&key.into())
+        .await
+        .change_context(OidcErrors::ServerError)
+        .attach_printable("Failed to delete authorization code from redis")
+        .map(|_| ())
```
sai-harsha-vardhan, line 102: Same here

crates/router/src/services/oidc_provider.rs
kanikabansal08: These are guard-style validations, so I’ve intentionally used early returns to short-circuit on invalid requests. I think for these functions, forcing a single-exit or nested if/else structure would reduce readability.

crates/router/src/utils/oidc.rs
kanikabansal08: we don't need to log this as this would be a client error and not a server failure. Logging it would add noise without actionable value. Therefore I am collapsing it to one protocol level error.

crates/router/src/utils/oidc.rs
kanikabansal08: We’re already attaching printable context at the point of failure and rely on the top-level OIDC boundary to log all ServerError. Do we stil need extra logging?

crates/router/src/utils/oidc.rs
kanikabansal08: `build_oidc_redirect_url` is a utility function, so it only attaches contextual information.
Logging here would result in the same failure being logged twice for a single request.

crates/router/src/utils/oidc.rs
kanikabansal08: Since Url::parse() returns Result, matching on the tuple of Results is more direct and idiomatic. Using zip would require converting both Results to Option

crates/router/src/services/oidc_provider.rs
```diff
@@ -49,7 +58,222 @@ pub async fn get_jwks(state: SessionState) -> RouterResponse<&'static JwksRespon
...
+    let client = state
+        .conf
+        .oidc
+        .get_inner()
+        .get_client(&payload.client_id)
+        .ok_or_else(|| report!(OidcErrors::UnauthorizedClient))
+        .attach_printable("Invalid client ID")?;
+
+    if !oidc_utils::validate_redirect_uri_match(&client.redirect_uri, &payload.redirect_uri) {
+        return Err(report!(OidcErrors::InvalidRequest)).attach_printable("Redirect URI mismatch");
```
sai-harsha-vardhan: nit: Can we propogate the error using `?`

crates/router/src/services/oidc_provider.rs
```diff
@@ -49,7 +58,222 @@ pub async fn get_jwks(state: SessionState) -> RouterResponse<&'static JwksRespon
...
 
-            Ok::<_, error_stack::Report<ApiErrorResponse>>(JwksResponse { keys })
+            Ok(JwksResponse { keys })
         })
         .map(ApplicationResponse::Json)
 }
+
+fn validate_authorize_params(payload: &OidcAuthorizeQuery, state: &SessionState) -> OidcResult<()> {
+    if !payload.scope.contains(&Scope::Openid) {
+        return Err(report!(OidcErrors::InvalidScope))
```
sai-harsha-vardhan: nit: Can we propogate the error using `?`

crates/router/src/utils/oidc.rs
sai-harsha-vardhan: we would need to log the redis deletiong failure error from `delete_key` right?

kanikabansal08: The Redis error won’t be lost it’s propagated as a ServerError with attached context of redis the deletion failure. We also follow the same pattern across the codebase for redis deletion.

PR #10282, opened 2025-11-15
refactor(euclid): remove disabled mcas from eligibility

prajjwalkumar17:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

This pull request updates the connector filtering logic in the payment routing process to ensure that only active merchant connector accounts are considered. The main change is the addition of a check for active MCAs before adding a connector to the final selection.

Connector filtering improvements:

* Added a database query to retrieve merchant connector accounts that are not disabled, and updated the filtering condition to only include connectors with active MCAs. (`crates/router/src/core/payments/routing.rs`)

## The testing flows are as follows:
1. create three connectors (cybersource1, adyen, cybersource2)
2. create routing rule (pm = card -> {cybersource1, adyen})
```
curl --location 'http://localhost:8080/routing' \
--header 'api-key: dev_S9MVF6fZI8LA6rS6kb7vzQ6mWSu65rWkG4CeBc0mEA5a0AtKAULdJCNVpmjjaOPC' \
--data '
{
    "name": "Rule Based Routing-2025-11-05",
    "description": "This is a rule based routing created at Wed, 05 Nov 2025 17:36:15 GMT",
    "profile_id": "pro_xFxhvVY8kFlb1Yy3yhC0",
    "algorithm": {
        "type": "advanced",
        "data": {
            "defaultSelection": {
                "type": "priority",
                "data": []
            },
            "rules": [
                {
                    "name": "paypal rule",
                    "connectorSelection": {
                        "type": "priority",
                        "data": [
                            {
                                "connector": "cybersource",
                                "merchant_connector_id": "mca_uOOjUjdYzJgTWQY6zqTE"
                            },
                            {
                                "connector": "adyen",
                                "merchant_connector_id": "mca_EA6RApHUvETHhNoWljHd"
                            }
                        ]
                    },
                    "statements": [
                        {
                               "condition": [
                                {
                                    "lhs": "payment_method",
                                    "comparison": "equal",
                                    "value": {
                                        "type": "enum_variant",
                                        "value": "card"
                                    },
                                    "metadata": {}
                                }
                            ],
                            "nested": null
                        }
                    ]
                }
            ],
            "metadata": {}
        }
    }
}'
```
3. payment should go through cybersource1.
4. disable cybersource1
5. payment now should go through adyen (previously it was again going with cybersource1 and failing).

## Checklist

### DIFF
--- a/crates/router/src/core/payments/routing.rs
+++ b/crates/router/src/core/payments/routing.rs
@@ -988,7 +988,7 @@ pub async fn perform_cgraph_filtering(
             .merchant_connector_id
             .as_ref()
             .map(|id| active_mca_ids.contains(id))
-            .unwrap_or(true);
+            .unwrap_or(false);
 
         if cgraph_eligible && filter_eligible && mca_active {
             final_selection.push(choice);


crates/router/src/core/payments/routing.rs
```diff
@@ -962,7 +962,24 @@ pub async fn perform_cgraph_filtering(
         let filter_eligible =
             eligible_connectors.is_none_or(|list| list.contains(&routable_connector));
 
-        if cgraph_eligible && filter_eligible {
+        let db_mcas = state
+            .store
+            .find_merchant_connector_account_by_merchant_id_and_disabled_list(
```
jagan-jaya: call it once in the top level and convert mca_ids into hashset for O(1) searching instead of iterating and comparing every time

crates/router/src/core/payments/routing.rs
```diff
@@ -962,7 +984,13 @@ pub async fn perform_cgraph_filtering(
         let filter_eligible =
             eligible_connectors.is_none_or(|list| list.contains(&routable_connector));
 
-        if cgraph_eligible && filter_eligible {
+        let mca_active = choice
+            .merchant_connector_id
+            .as_ref()
+            .map(|id| active_mca_ids.contains(id))
+            .unwrap_or(true);
```
Aprabhat19: Should we not have unwrap or as false? Because if the mca is not in the active list why would we push it to the final selection?

jagan-jaya (approved):

PR #10325, opened 2025-11-18
feat(platform): introduce platform type support

apoorvdixit88:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [x] Enhancement
- [x] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description
This PR introduces the Platform concept into Hyperswitch by modeling every operation from the pair of Provider + Processor:

Provider → The business owner. In VSAAS this corresponds to our merchant, and this is typically the platform merchant. Internally, Provider wraps a MerchantAccount and a MerchantKeyStore.

```
pub struct Provider {
    account: MerchantAccount,
    key_store: MerchantKeyStore,
}
```

Processor → The merchant whose connector credentials are used to execute the operation (for example, making a payment with a specific PSP). Internally, Processor also wraps a MerchantAccount and a MerchantKeyStore.

```
pub struct Processor {
    account: MerchantAccount,
    key_store: MerchantKeyStore,
}
```

Both are grouped inside a Platform type:

```
pub struct Platform {
    provider: Box<Provider>,
    processor: Box<Processor>,
}
```

This allows us to clearly distinguish who is who is governing entity (provider) from whose credentials are used to talk to connectors (processor).

For Standard/Normal merchants, we construct Platform by passing the **same** MerchantAccount and MerchantKeyStore for both, so provider and processor point to the same logical merchant. That means there is no functional change for existing behavior: standard flows still behave exactly as before, just represented through the new types.

To keep the transition safe and incremental, all modules in this PR still operate at the processor level. Wherever we need to execute something against a connector, we continue to read from Processor (via get_account() / get_key_store()). For standard merchants, since provider and processor are identical, current behavior remains unaffected. Future PRs will gradually update individual modules and APIs to start using Provider and Processor separately, enforcing the distinction in a type-safe way.

- On the authentication side, this PR updates the API-key–based auth path so that AuthenticationData now produces a Platform value, instead of only a single merchant. That means downstream code can start consuming Platform and access get_provider() and get_processor() explicitly. Other auth mechanisms (JWT, Publishable Key, Internal Merchant/Profile based auth, etc.) are intentionally not changed in this PR and will be handled in follow-up work.

Overall, this PR is foundational. It introduces the Provider, Processor, and Platform types and wires them into API auth, while keeping runtime behavior unchanged. This unlocks the ability to safely implement true Platform → Connected merchant flows in subsequent PRs, without destabilizing existing  flows

### Additional Changes

## Motivation and Context
Closes [#12265](https://github.com/juspay/hyperswitch-cloud/issues/12265)

## How did you test it?
The code compiles successfully and all existing test cases pass for the current Standard merchant flow. Since this PR only introduces the Platform model without changing runtime behavior, there is no impact on existing execution paths. As follow-up PRs gradually enable Platform-specific flows, each API and module will be tested individually to ensure full functional coverage.

## Checklist

### DIFF
--- a/crates/router/tests/payments.rs
+++ b/crates/router/tests/payments.rs
@@ -608,7 +608,7 @@ async fn payments_create_core_adyen_no_redirect() {
         .await
         .unwrap();
 
-    let platform = hyperswitch_domain_models::Platform::new(
+    let platform = hyperswitch_domain_models::platform::Platform::new(
         merchant_account.clone(),
         key_store.clone(),
         merchant_account.clone(),


crates/router/src/core/admin.rs
```diff
@@ -2643,15 +2638,11 @@ async fn validate_pm_auth(
...
+            platform.get_processor().get_key_store(),
         )
         .await
         .change_context(errors::ApiErrorResponse::MerchantConnectorAccountNotFound {
-            id: merchant_context
-                .get_merchant_account()
-                .get_id()
-                .get_string_repr()
-                .to_owned(),
+            id: merchant_id.get_string_repr().to_owned(),
```
tsdk02: Can we use the merchant_id from platform here instead of `merchant_id` param. We can remove the `merchant_id` parameter from this function signature later.

crates/router/src/core/admin.rs
```diff
@@ -2856,13 +2847,15 @@ pub async fn update_connector(
         .await?;
     core_utils::validate_profile_id_from_auth_layer(profile_id, &mca)?;
 
-    let merchant_context = domain::MerchantContext::NormalMerchant(Box::new(domain::Context(
+    let platform = domain::Platform::new(
```
tsdk02, line 2855: Can we construct the platform in the handler and pass it to the core function instead of retrieving the keystore here, this will standardize things.

crates/router/src/routes/relay.rs
```diff
@@ -63,12 +61,12 @@ pub async fn relay_retrieve(
         &req,
         relay_retrieve_request,
         |state, auth: auth::AuthenticationData, req, _| {
-            let merchant_context = crate::types::domain::MerchantContext::NormalMerchant(Box::new(
-                crate::types::domain::Context(auth.merchant_account, auth.key_store),
-            ));
+            use hyperswitch_domain_models::platform::Platform;
```
tsdk02: Can we have imports at the top of the file

crates/router/src/routes/relay.rs
```diff
@@ -63,12 +61,12 @@ pub async fn relay_retrieve(
...
+
+            let platform: Platform = auth.clone().into();
```
tsdk02: Do we need type annotations here though?

crates/router/src/routes/profile_acquirer.rs
```diff
@@ -69,16 +63,14 @@ pub async fn profile_acquirer_update(
         state,
         &req,
         payload,
-        |state: super::SessionState, auth_data, req, _| {
-            let merchant_context = domain::MerchantContext::NormalMerchant(Box::new(
-                domain::Context(auth_data.merchant_account, auth_data.key_store),
-            ));
+        |state: super::SessionState, auth_data: auth::AuthenticationData, req, _| {
+            let platform: domain::Platform = auth_data.into(); // check for type annotation here
```
tsdk02: Type annotations may not be required, pls check and remove the comment.

crates/router/src/routes/profile_acquirer.rs
```diff
@@ -24,15 +24,9 @@ pub async fn create_profile_acquirer(
...
-            let merchant_context = domain::MerchantContext::NormalMerchant(Box::new(
-                domain::Context(auth_data.merchant_account, auth_data.key_store),
-            ));
-            crate::core::profile_acquirer::create_profile_acquirer(
-                state,
-                req,
-                merchant_context.clone(),
-            )
+        |state: super::SessionState, auth_data: auth::AuthenticationData, req, _| {
+            let platform: domain::Platform = auth_data.into();
```
tsdk02: Check if type annotations are required for auth_data, and also for platform

crates/hyperswitch_domain_models/src/platform.rs
```diff
@@ -0,0 +1,89 @@
+pub use crate::{merchant_account::MerchantAccount, merchant_key_store::MerchantKeyStore};
+
+/// Provider = The business owner or the governing entity in the hierarchy.
+/// In a platform connected setup this is related to platform merchant.
```
tsdk02:
```suggestion
/// In a platform-connected setup this is represented by the platform merchant.
```

crates/hyperswitch_domain_models/src/platform.rs
```diff
@@ -0,0 +1,89 @@
...
+/// For a standard merchant, provider and processor are the same entity.
```
tsdk02:
```suggestion
/// For a standard merchant, both provider and processor are represented by the same entity.
```

crates/hyperswitch_domain_models/src/platform.rs
```diff
@@ -0,0 +1,89 @@
...
+    /// Returns a reference to the key store associated with the processor.
+    pub fn get_key_store(&self) -> &MerchantKeyStore {
+        &self.key_store
+    }
+}
+
+/// Platform holds both Provider and Processor together.
+/// This struct makes it possible to distinguish the business owner for the org
+/// versus whose processor credentials are used for execution.
+/// For a standard merchant flow, provider == processor.
```
tsdk02, line 54:
```suggestion
/// Platform holds both Provider and Processor together.
/// This struct makes it possible to distinguish the business owner for the org versus whose processor credentials are used for execution.
/// For a standard merchant flow, provider == processor.
```

crates/hyperswitch_domain_models/src/platform.rs
```diff
@@ -0,0 +1,89 @@
...
+#[derive(Clone, Debug)]
+pub struct Platform {
+    provider: Box<Provider>,
+    processor: Box<Processor>,
+}
+
+impl Platform {
+    /// Creates a Platform pairing from two merchant identities:
+    /// one acting as provider and one as processor
+    /// Standard merchants can pass the same account/key_store for both
```
tsdk02:
```suggestion
    /// Standard merchants can pass the same account/key_store for both provider and processor
```

crates/router/src/core/admin.rs
tsdk02: Probably this is not possible currently as we do not get any merchant context from the authentication.
We should probably use `AdminApiAuthWithMerchantIdFromHeader` auth for this, and construct platform in the handler and pass it to core.

crates/router/src/services/authentication.rs
```diff
@@ -72,7 +72,38 @@ pub struct AuthenticationData {
     pub merchant_account: domain::MerchantAccount,
     pub key_store: domain::MerchantKeyStore,
     pub profile: domain::Profile,
-    pub platform_merchant_account: Option<domain::MerchantAccount>,
+    pub platform_account_with_key_store: Option<PlatformAccountWithKeyStore>,
+}
+
+#[derive(Clone, Debug)]
+pub struct PlatformAccountWithKeyStore {
```
jarnura, line 79: Why we need a extra types, this should be the `Provider` type

jarnura (dismissed):

tsdk02 (dismissed):

crates/hyperswitch_domain_models/src/platform.rs
```diff
@@ -0,0 +1,89 @@
...
+        processor_account: MerchantAccount,
+        processor_key_store: MerchantKeyStore,
+    ) -> Self {
+        let provider = Provider::new(provider_account, provider_key_store);
+        let processor = Processor::new(processor_account, processor_key_store);
+        Self {
+            provider: Box::new(provider),
+            processor: Box::new(processor),
+        }
+    }
```
ShankarSinghC, line 77: Since `provider_account` and `processor_account` are the same type, developers might accidentally interchange them.

How can we prevent this?

Can we add validation before constructing the object? For example, for the standard merchant, `provider_account should always be equal to processor_account`, and for other flows, `processor_account` should always be platform.

apoorvdixit88: Mostly the Platform which consist of provider and processor will come from Authentication there it needs to be constructed carefully. In future these two may not be the merchant account always, provider can just hold reference to id and some config, then all params will be different.
But type safety is very important,  in the upcoming PRs we will try to figure out ways for the type checks.

crates/router/src/services/authentication.rs
apoorvdixit88: This is temporary, will change this later, inside authentication data it will have platform field only.

tsdk02 (approved):

jarnura (approved):

PR #10334, opened 2025-11-18
feat(payment-methods): add support for session create api  and customer pml for guest checkout in v2

prasunna09:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

 add api domain changes for session create and customer pml changes for guest checkout .
 Made customer id optional in session create call.
 and in pml, customer pml would be empty vector since there customer id present in session data.
 
 in confirm flow, we will throw error if customer id is not present since we only have support for storing in persistent storage. this will be taken up in following pr
feature description - [10361](https://github.com/juspay/hyperswitch/issues/10361)

### Additional Changes

## How did you test it?

no testing required, since core flow behavior is not modified in this pr.
only api changes, when customer id made optional, 
list will give empty vector if customer id is not present in session data.
else usual flow.

to test payment method create - 
```
curl --location 'https://integ.hyperswitch.io/api/v2/payment-methods' \
--header 'X-Profile-Id: pro_ftn9Q2Wg26QqPMNZH2TD' \
--header 'Authorization: api-key=snd_lNHDRWu3h72cA7XM64cmH9u2r4WfZko6Clzm8OPUR9ki8JPoKiyIr5m3XVLWj1vd' \
--header 'api-key: snd_lNHDRWu3h72cA7XM64cmH9u2r4WfZko6Clzm8OPUR9ki8JPoKiyIr5m3XVLWj1vd' \
--data '{
    "customer_id": "12345_cus_019a39a506c6711182c41bb936404fc6", //remo
    "payment_method_type": "card",
    "payment_method_subtype": "credit",
    "payment_method_data": {
        "card": {
            "card_number": "<card number>",
            "card_exp_month": "12",
            "card_exp_year": "2026",
            "card_holder_name": "joseph Doe"
        }
    }

}'
```

## Checklist


crates/api_models/src/payment_methods.rs
```diff
@@ -3315,7 +3304,7 @@ pub struct PaymentMethodSessionResponse {
 
     /// The customer id for which the payment methods session is to be created
     #[schema(value_type = String, example = "12345_cus_01926c58bc6e77c09e809964e72af8c8")]
```
Sakilmostak: shouldn't the value_type be Option<String>

crates/router/src/core/payment_methods.rs
```diff
@@ -3531,7 +3511,10 @@ pub async fn payment_methods_session_confirm(
         })
         .or_else(|| payment_method_session_billing.clone());
 
-    let customer_id = payment_method_session.customer_id.clone();
+    let customer_id = payment_method_session
+        .customer_id
+        .clone()
+        .get_required_value("customer_id")?;
```
Sakilmostak, line 3583: Are we only allowing guest customer for payment method session create? If confirm still requires the customer_id, does that means confirm will always fail for guest customer checkout since payment_method_session don't have customer_id

Sakilmostak (changes requested):

crates/router/src/core/payment_methods.rs
prasunna09: no, this confirm will always store data in persistent db which requires customer id, hence the get_required_value method is used.

there will be different core flow, for guest checkout

Sakilmostak: cool 👍

Sakilmostak (approved):

PR #10423, opened 2025-11-25
fix(core): Merchant Category Code Enum to strict Struct

Nithin1506200:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

Previously Merchant category code was a fixed enum but it can be anywhere between 0000 - 9999.
- MCC should Strict type (string) ranging 0000 to 9999
- Having a Enum with 10k variants isn't feasible

### update MCC

```
curl --location 'localhost:8080/account/merchant_1764051682/business_profile/pro_YTdBjabwLVzft5YFBfNN' \
--header 'api-key: dev_xMNMfAyvkAw70hbFCWrFMWBAFXWFZeZTt74sHps9Y2Zapje3LFgt98CB5wYk1K7l' \
--data '{
    "merchant_category_code": "1000"
}'
```

```json
{
    "merchant_id": "merchant_1764051682",
    "profile_id": "pro_YTdBjabwLVzft5YFBfNN",
    "profile_name": "US_default",
    "enable_payment_response_hash": true,
    "payment_response_hash_key": "sI2DALHd7VgpuPGzAGf8e3GjWbCrjcETfEVnWSsn1z8X7RjG3P0c3my2fKCmPsJ4",
    "redirect_to_merchant_with_http_post": false,
    "webhook_details": {
        "webhook_version": "1.0.1",
        "webhook_url": null,
        "payment_failed_enabled": true,
        "payment_statuses_enabled": null,
        "refund_statuses_enabled": null,
        "payout_statuses_enabled": null
    },
    "metadata": null,
    "routing_algorithm": null,
    "intent_fulfillment_time": 900,
    "frm_routing_algorithm": {
        "data": "signifyd",
        "type": "single"
    },
    "payout_routing_algorithm": null,
    "applepay_verified_domains": null,
    "session_expiry": 900,
    "payment_link_config": null,
    "authentication_connector_details": null,
    "use_billing_as_payment_method_billing": true,
    "extended_card_info_config": null,
    "collect_shipping_details_from_wallet_connector": false,
    "collect_billing_details_from_wallet_connector": false,
    "always_collect_shipping_details_from_wallet_connector": false,
    "always_collect_billing_details_from_wallet_connector": false,
    "is_connector_agnostic_mit_enabled": false,
    "payout_link_config": null,
    "outgoing_webhook_custom_http_headers": null,
    "tax_connector_id": null,
    "is_tax_connector_enabled": false,
    "is_network_tokenization_enabled": false,
    "is_auto_retries_enabled": false,
    "max_auto_retries_enabled": null,
    "always_request_extended_authorization": null,
    "is_click_to_pay_enabled": false,
    "authentication_product_ids": null,
    "card_testing_guard_config": {
        "card_ip_blocking_status": "disabled",
        "card_ip_blocking_threshold": 3,
        "guest_user_card_blocking_status": "disabled",
        "guest_user_card_blocking_threshold": 10,
        "customer_id_blocking_status": "disabled",
        "customer_id_blocking_threshold": 5,
        "card_testing_guard_expiry": 3600
    },
    "is_clear_pan_retries_enabled": false,
    "force_3ds_challenge": false,
    "is_debit_routing_enabled": false,
    "merchant_business_country": null,
    "is_pre_network_tokenization_enabled": false,
    "acquirer_configs": null,
    "merchant_category_code": "1000",
    "merchant_country_code": null,
    "dispute_polling_interval": null,
    "is_manual_retry_enabled": null,
    "always_enable_overcapture": null,
    "is_external_vault_enabled": "skip",
    "external_vault_connector_details": null,
    "billing_processor_id": null,
    "is_l2_l3_enabled": false
}
```

### EDGE CASES

1. 
```json
{
    "merchant_category_code": "100"
}

{
    "error": {
        "error_type": "invalid_request",
        "message": "Json deserialize error: Invalid MCC: MCC must be exactly 4 characters long (e.g., '0001'), but got '100' at line 3 column 1",
        "code": "IR_06"
    }
}
```
2.

```json
{
    "merchant_category_code": "00004"
}

{
    "error": {
        "error_type": "invalid_request",
        "message": "Json deserialize error: Invalid MCC: MCC must be exactly 4 characters long (e.g., '0001'), but got '00004' at line 3 column 1",
        "code": "IR_06"
    }
}
```

```json
{
    "merchant_category_code": "xfdwerf"
}

{
    "error": {
        "error_type": "invalid_request",
        "message": "Json deserialize error: Invalid MCC: Code must be a number: invalid digit found in string at line 3 column 1",
        "code": "IR_06"
    }
}
```

```json
{
    "merchant_category_code": "1.00"
}

{
    "error": {
        "error_type": "invalid_request",
        "message": "Json deserialize error: Invalid MCC: Code must be a number: invalid digit found in string at line 3 column 1",
        "code": "IR_06"
    }
}

```

DB snapshot

Screenshot 2025-11-25 at 12 29 04 PM

### wasm changes
- Deliberately added wrong mcc to test logs

Screenshot 2025-12-04 at 11 48 05 AM

### Additional Changes

## Checklist

### DIFF
--- a/crates/common_enums/src/enums.rs
+++ b/crates/common_enums/src/enums.rs
@@ -3024,7 +3024,8 @@ impl MerchantCategoryCode {
     }
 
     pub fn get_category_name(&self) -> Result<&str, InvalidMccError> {
-        match self.get_code()? {
+        let code = self.get_code()?;
+        match code {
             // specific mapping needs to be depricated
             5411 => Ok("Grocery Stores, Supermarkets (5411)"),
             7011 => Ok("Lodging-Hotels, Motels, Resorts-not elsewhere classified (7011)"),
@@ -3034,20 +3035,8 @@ impl MerchantCategoryCode {
             4816 => Ok("Computer Network/Information Services (4816)"),
             5661 => Ok("Shoe Stores (5661)"),
 
-            // general mapping
-            1..=1499 => Ok("Agricultural Services"),
-            1500..=2999 => Ok("Contracted Services"),
-            3000..=3999 => Ok("Travel and Entertainment"),
-            4000..=4799 => Ok("Transportation Services"),
-            4800..=4999 => Ok("Utility Services"),
-            5000..=5599 => Ok("Retail Outlet Services"),
-            5600..=5699 => Ok("Clothing Stores"),
-            5700..=7299 => Ok("Miscellaneous Stores"),
-            7300..=7999 => Ok("Business Services"),
-            8000..=8999 => Ok("Professional Services and Membership Organizations"),
-            9000..=9999 => Ok("Government Services"),
             _ => Err(InvalidMccError {
-                message: "Invalid MCC range found . Required 0001-9999".to_string(),
+                message: format!("Category name not found for {}", code),
             }),
         }
     }
--- a/crates/euclid_wasm/src/lib.rs
+++ b/crates/euclid_wasm/src/lib.rs
@@ -50,6 +50,15 @@ struct SeedData {
 static SEED_DATA: OnceLock<SeedData> = OnceLock::new();
 static SEED_FOREX: OnceLock<currency_conversion_types::ExchangeRates> = OnceLock::new();
 
+#[wasm_bindgen]
+extern "C" {
+    #[wasm_bindgen(js_namespace = console, js_name = log)]
+    fn console_log(s: &str);
+
+    #[wasm_bindgen(js_namespace = console, js_name = error)]
+    fn console_error(s: &str);
+}
+
 /// This function can be used by the frontend to educate wasm about the forex rates data.
 /// The input argument is a struct fields base_currency and conversion where later is all the conversions associated with the base_currency
 /// to all different currencies present.
@@ -100,15 +109,29 @@ pub fn get_two_letter_country_code() -> JsResult {
 /// along with their names.
 #[wasm_bindgen(js_name=getMerchantCategoryCodeWithName)]
 pub fn get_merchant_category_code_with_name() -> JsResult {
-    let merchant_category_codes_with_name = vec![5413, 7011, 763, 8111, 5021, 4816, 5661]
+    let merchant_category_codes_with_name = vec![5411, 7011, 763, 8111, 5021, 4816, 5661]
         .into_iter()
-        .filter_map(|mcc_value| {
-            let mcc = MerchantCategoryCode::new(mcc_value).ok()?;
-            let mcc_name = mcc.get_category_name().ok()?;
-            Some(MerchantCategoryCodeWithName {
-                code: mcc.clone(),
-                name: mcc_name.to_string(),
-            })
+        .filter_map(|mcc_value| match MerchantCategoryCode::new(mcc_value) {
+            Ok(mcc) => match mcc.get_category_name() {
+                Ok(mcc_name) => Some(MerchantCategoryCodeWithName {
+                    code: mcc.clone(),
+                    name: mcc_name.to_string(),
+                }),
+                Err(err) => {
+                    console_error(&format!(
+                        "Failed to get category name for MCC {}: {:?}",
+                        mcc_value, err
+                    ));
+                    None
+                }
+            },
+            Err(err) => {
+                console_error(&format!(
+                    "Failed to create MCC for value {}: {:?}",
+                    mcc_value, err
+                ));
+                None
+            }
         })
         .collect::<Vec<_>>();
 


crates/common_enums/src/enums.rs
```diff
@@ -3043,96 +2995,153 @@ pub enum DisputeStatus {
...
+            // specific mapping needs to be depricated
+            5411 => "Grocery Stores, Supermarkets (5411)",
+            7011 => "Lodging-Hotels, Motels, Resorts-not elsewhere classified (7011)",
+            763 => "Agricultural Cooperatives (0763)",
+            8111 => "Attorneys, Legal Services (8111)",
+            5021 => "Office and Commercial Furniture (5021)",
+            4816 => "Computer Network/Information Services (4816)",
+            5661 => "Shoe Stores (5661)",
+
+            // general mapping
```
sai-harsha-vardhan: What is the source of the mapping?

crates/common_enums/src/enums.rs
```diff
@@ -3043,96 +2995,153 @@ pub enum DisputeStatus {
...
+            3000..=3999 => "Travel and Entertainment",
+            4000..=4799 => "Transportation Services",
+            4800..=4999 => "Utility Services",
+            5000..=5599 => "Retail Outlet Services",
+            5600..=5699 => "Clothing Stores",
+            5700..=7299 => "Miscellaneous Stores",
+            7300..=7999 => "Business Services",
+            8000..=8999 => "Professional Services and Membership Organizations",
+            9000..=9999 => "Government Services",
+            _ => "Unknown or Out-of-Range",
```
sai-harsha-vardhan: Should we throw error in this case?

crates/common_enums/src/enums.rs
```diff
@@ -3043,96 +2995,153 @@ pub enum DisputeStatus {
...
-    #[serde(rename = "Computer Network/Information Services (4816)")]
-    ComputerNetworkInformationServices,
-    #[serde(rename = "Shoe Stores (5661)")]
-    ShoeStores,
+#[diesel(sql_type = Text)]
+pub struct MerchantCategoryCode(String);
+
+impl MerchantCategoryCode {
+    pub fn get_code(&self) -> u16 {
+        self.0.parse::<u16>().unwrap_or(0)
```
sai-harsha-vardhan: In case of parse failure, we may need to throw error

crates/common_enums/src/enums.rs
Nithin1506200: https://stripe.com/in/guides/merchant-category-codes

crates/common_enums/src/enums.rs
Nithin1506200: MCC(String) the field is Privatized and we have check both for serializing and from SQL 
```rust
impl FromStr for MerchantCategoryCode {
    type Err = InvalidMccError;
    fn from_str(s: &str) -> Result<Self, Self::Err> {
        let _code: u16 = s.parse()?;

        if s.len() != 4 {
            return Err(InvalidMccError {
                message: format!(
                    "MCC must be exactly 4 characters long (e.g., '0001'), but got '{}'",
                    s
                ),
            });
        }
        Ok(Self(s.to_string()))
    }
}
```

in other words we cannot create a MCC object MCC("randomString") directly in any other crates.
since we are using get_code(&self)-> & get_category_name(&self) we can ensure self.0 is always btw 0000-9999 nothing other than this. So is wrapping Result really a compulsory at this point? Can't we make an exemption out of here?
- If we had some function like get_category_name(code:&u16) we would have added error as they can pass any sstring here.

crates/common_enums/src/enums.rs
Nithin1506200: same explanation here https://github.com/juspay/hyperswitch/pull/10423/files#r2575974543

crates/common_enums/src/enums.rs
Nithin1506200: currently we are not using this anywhere including in Euclid.

crates/common_enums/src/enums.rs
```diff
@@ -3043,96 +2995,158 @@ pub enum DisputeStatus {
...
+            1500..=2999 => Ok("Contracted Services"),
+            3000..=3999 => Ok("Travel and Entertainment"),
+            4000..=4799 => Ok("Transportation Services"),
+            4800..=4999 => Ok("Utility Services"),
+            5000..=5599 => Ok("Retail Outlet Services"),
+            5600..=5699 => Ok("Clothing Stores"),
+            5700..=7299 => Ok("Miscellaneous Stores"),
+            7300..=7999 => Ok("Business Services"),
+            8000..=8999 => Ok("Professional Services and Membership Organizations"),
+            9000..=9999 => Ok("Government Services"),
```
ShankarSinghC: Why do we need this mcc range based handling ?

I see that the `get_category_name` is just called over the hardcode mcc (i.e `vec![5413, 7011, 763, 8111, 5021, 4816, 5661]`), in this we would not have to handle the exception cases.

crates/euclid_wasm/src/lib.rs
```diff
@@ -100,10 +100,15 @@ pub fn get_two_letter_country_code() -> JsResult {
 /// along with their names.
 #[wasm_bindgen(js_name=getMerchantCategoryCodeWithName)]
 pub fn get_merchant_category_code_with_name() -> JsResult {
-    let merchant_category_codes_with_name = MerchantCategoryCode::iter()
-        .map(|mcc_value| MerchantCategoryCodeWithName {
-            code: mcc_value,
-            name: mcc_value.to_merchant_category_name(),
+    let merchant_category_codes_with_name = vec![5413, 7011, 763, 8111, 5021, 4816, 5661]
```
ShankarSinghC: Instead of having it as hardcoded this could be stored in db

crates/euclid_wasm/src/lib.rs
```diff
@@ -100,10 +100,15 @@ pub fn get_two_letter_country_code() -> JsResult {
...
+        .into_iter()
+        .filter_map(|mcc_value| {
+            let mcc = MerchantCategoryCode::new(mcc_value).ok()?;
+            let mcc_name = mcc.get_category_name().ok()?;
```
ShankarSinghC: Can we log the error in case of error ?

If we are not able to get the category name for an mcc it will be skipped and the remaining mcc will be displayed in the list. And it would be difficult to debug why an mcc is being skipped without any logs.

ShankarSinghC (approved):

PR #10439, opened 2025-11-25
refactor(euclid): fallback connectors for sessions and PML call, if routing rule validation fails

### LINKED ISSUE #10391: refactor(euclid): default fallback to hs routing result for 4xx from DE service

prajjwalkumar17:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

This pull request introduces improved error handling for session flow routing in the payment methods and payments modules. When routing fails, the system now logs the error and gracefully falls back to default connectors instead of returning an internal server error. This ensures more robust and reliable payment processing.

**Error handling and fallback logic improvements:**

* In `cards.rs` (`list_payment_methods`), added logic to catch errors from `perform_session_flow_routing`, log the failure, and construct a fallback routing map using the default connectors. The routing approach is set to `DefaultFallback` when this occurs.
* In `payments.rs`, updated the routing flow to handle errors from `perform_session_flow_routing` by logging the error and building a fallback connector mapping, also marking the routing approach as `DefaultFallback`.

## How did you test it?

We have to explicitly test for two different flows:
1. 4xx from decision engine should pick hyps result(can't be tested, as we won't be able to produce 4xx from DE).
2. Rule mis-configured shouldn't result in 5xx insted should default-fallback.

Testing 2nd scenario.
--> configure and activate a rule with card_bin of length 7 to mis-configure a rule(it should be 6).
```
curl --location 'http://localhost:8080/routing' \
--header 'api-key: dev_ssIoVnUDrSVryP63v305gq3mc0yYKuV8HZkW0gl9qiWSCkhOK33LqZaaJ8nNXWWj' \
--data '
{
    "name": "Rule Based Routing-2025-11-05",
    "description": "This is a rule based routing created at Wed, 05 Nov 2025 17:36:15 GMT",
    "profile_id": "pro_rWfa1iSEbPUl85aVYYh6",
    "algorithm": {
        "type": "advanced",
        "data": {
            "defaultSelection": {
                "type": "priority",
                "data": []
            },
            "rules": [
                {
                    "name": "paypal rule",
                    "connectorSelection": {
                        "type": "priority",
                        "data": [
                            {
                                "connector": "nmi",
                                "merchant_connector_id": "mca_uCRMeZFMDGOX27Gawd1Q"
                            },
                            {
                                "connector": "braintree",
                                "merchant_connector_id": "mca_g6HmsYY8YCE0Agc5Q70Z"
                            }
                     
                        ]
                    },
                    "statements": [
                        {
                            "condition": [
                                {
                                    "lhs": "card_bin",
                                    "comparison": "equal",
                                    "value": {
                                        "type": "str_value",
                                        "value": "4242424"
                                    },
                                    "metadata": {}
                                }
                            ],
                            "nested": null
                        }
                    ]
                }
            ],
            "metadata": {}
        }
    }
}'
```
--> session_token call: was giving 5xx earlier (should be fixed now)
```
curl --location 'http://localhost:8080/payments/session_tokens' \
--header 'api-key: pk_dev_903e13005b014af2ac07e63a1901d70e' \
--data '{
    "payment_id": "pay_OiwNGOSLY3ZXG1fucoMU",
    "wallets": [],
    "client_secret": "pay_OiwNGOSLY3ZXG1fucoMU_secret_PZOEHCfz4e2Z63RWwgrr"
}'
```
--> Merchant PML was giving 5xx earlier (should be fixed now)
```
curl --location 'http://localhost:8080/account/payment_methods?client_secret=pay_OiwNGOSLY3ZXG1fucoMU_secret_PZOEHCfz4e2Z63RWwgrr' \
--header 'api-key: pk_dev_903e13005b014af2ac07e63a1901d70e' \
--data ''
```
--> process a transaction, earlier it would have resulted in 5xx but now the error will be logged and it will be default_fallbacked.

## Checklist

### DIFF
--- a/crates/router/src/core/payment_methods/cards.rs
+++ b/crates/router/src/core/payment_methods/cards.rs
@@ -2930,16 +2930,51 @@ pub async fn list_payment_methods(
             merchant_account: platform.get_processor().get_account(),
             payment_attempt,
             payment_intent,
-            chosen,
+            chosen: chosen.clone(),
         };
-        let (result, routing_approach) = routing::perform_session_flow_routing(
+
+        let (result, routing_approach) = match routing::perform_session_flow_routing(
             sfr,
             &business_profile,
             &enums::TransactionType::Payment,
         )
         .await
-        .change_context(errors::ApiErrorResponse::InternalServerError)
-        .attach_printable("error performing session flow routing")?;
+        {
+            Ok(ok) => ok,
+
+            Err(err) => {
+                logger::error!(
+                    error=?err,
+                    "euclid_routing: list_payment_methods routing failed, falling back to default connectors"
+                );
+
+                let mut fallback: rustc_hash::FxHashMap<
+                    api_enums::PaymentMethodType,
+                    Vec<routing_types::SessionRoutingChoice>,
+                > = rustc_hash::FxHashMap::default();
+
+                for c in chosen.clone() {
+                    let connector_data = api::ConnectorData {
+                        connector: c.connector.connector,
+                        connector_name: c.connector.connector_name,
+                        get_token: c.connector.get_token,
+                        merchant_connector_id: c.connector.merchant_connector_id,
+                    };
+
+                    let choice = routing_types::SessionRoutingChoice {
+                        connector: connector_data,
+                        payment_method_type: c.payment_method_sub_type,
+                    };
+
+                    fallback
+                        .entry(c.payment_method_sub_type)
+                        .or_default()
+                        .push(choice);
+                }
+
+                (fallback, Some(api_enums::RoutingApproach::DefaultFallback))
+            }
+        };
 
         response.retain(|intermediate| {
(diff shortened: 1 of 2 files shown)


crates/router/src/core/payment_methods/cards.rs
```diff
@@ -2930,16 +2930,54 @@ pub async fn list_payment_methods(
...
+                        .or_default()
+                        .push(choice);
+                }
+
+                (
+                    fallback_connectors,
+                    Some(api_enums::RoutingApproach::DefaultFallback),
+                )
+            }
+        };
```
ShankarSinghC, line 2980: To avoid duplication of code we should move this logic inside some common function. Also are we suppose to run the fallback logic for any error that occurs in the `perform_session_flow_routing` or only when routing rule evaluation fails ?

PR #10112, opened 2025-11-04
feat(revenue_recovery): Add account updater support in revenue recovery flow for worldpayvantiv connector

NISHANTH1221:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description
This pr enables the account updater support for revenue recovery for worldpay vantiv connector. Worldpay vantiv when account updater is enabled by the merchant at the connectors end, it response with the new account updater details in the response of authorize flow. So to consume this in Revenue recovery flow we need to propogate these details up the ladder till the function `call_proxy_api` in revenue recovery.

When worldpay vantiv responds with the account updater we check for the payment method which should be `MANDATEPAYMENT`. If it is a mandate payment and it has account updater details we will respond with the updated token details. We will replace the source of the mandate reference from token response to account updater response and populate the same data in mandate_data in `PaymentConfirmData<>`

Token update can have two cases:
1. Expiry update: In this case token is same but its underlying details will change like expiry and token
2. Token Update: In this case token will get changed completely

Based on the Account updater it updates the datasource of revenue recovery.
1. If it is an expiry change, it modifies the underlying details for that token
2. If it is a new token, it set the old token is_active flag to false and insert the new token

### Additional Changes

- [ ] This PR modifies the API contract
- [ ] This PR modifies the database schema
- [ ] This PR modifies application configuration/environment variables

## How did you test it?

1. Create  worldpay-vantiv mca 
```
curl --location 'http://localhost:8080/v2/connector-accounts' \
--header 'Content-Type: application/json' \
--header 'Accept: application/jso' \
--header 'x-merchant-id:{{merchant_id}}' \
--header 'x-profile-id: {{profile_id}}' \
--header 'Authorization: admin-api-key=test_admin' \
--header 'api-key: admin-api-key=test_admin' \
--data '{
    "connector_type": "payment_processor",
    "connector_name": "worldpayvantiv",
    "connector_account_details": { 
        "auth_type": "SignatureKey",
        "api_key": "api_key", 
        "key1": "key1", 
        "api_secret": "api_secret" 
    },
    "payment_methods_enabled": [
        {
            "payment_method_type": "card",
            "payment_method_subtypes": [
                {
                    "payment_method_subtype": "credit",
                    "payment_experience": null,
                    "card_networks": [
                        "Visa",
                        "Mastercard"
                    ],
                    "accepted_currencies": null,
[...]

PR #10501, opened 2025-12-02
fix(revenue_recovery): Daily Retry History is being wrongly updated when its an expiry update

NISHANTH1221:
## Type of Change

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description
When account updater is done on a particular card token it may result in new token with new details(`New Token`) or old token with new details(`Expiry Update`). The bug is that when we were updating the  token in case of  Expiry update we were copying the daily retry history of the token and sending it to `upsert_payment_processor_token` function that function get the daily retry history with the new token and get the token from redis and update the redis with the same retry histry which results in count to be doubled

Before doing account updater details
Redis -> Token1 ( daily_retry_history : {"02-12-25": 1} )

after getting
updated old token -> upsert_payment_processor_token ( In this we fetch the existing Token1 daily retry history  which is this : ( daily_retry_history : {"02-12-25": 1} ) and we will get updated Token1 which will be same : ( daily_retry_history : {"02-12-25": 1} ) The upset function iterates through the daily_retry_history of the updated Old token and make the net daily retry history as ( daily_retry_history : {"02-12-25": 2} ) . 

And second bug is that we shouldn't update the `is_hard_decline` field just after the expiry update and give it chance to do one more retry. If the hard decline flag is updated to true just after the expiry update then the token wont be picked up again. So there is no point of account updater and we didn't do any retry with the updated expiry details.

### Additional Changes

## How did you test it?

Can follow the steps mentioned in #10112 to test this.

## Checklist

### DIFF
--- a/crates/router/src/core/revenue_recovery/types.rs
+++ b/crates/router/src/core/revenue_recovery/types.rs
@@ -713,9 +713,10 @@ impl Action {
                         .await
                         .ok();
 
-
                         // when account updater is an expiry update we dont want to update the hard decline flag to give it a chance to do a retry
-                        let should_we_update_hard_decline_flag = is_expiry_update.and_then(|is_expiry| is_expiry.then_some(false)).or(is_hard_decline);
+                        let should_we_update_hard_decline_flag = is_expiry_update
+                            .and_then(|is_expiry| is_expiry.then_some(false))
+                            .or(is_hard_decline);
 
                         let _update_connector_customer_id = storage::revenue_recovery_redis_operation::RedisTokenManager::update_payment_processor_token_error_code_from_process_tracker(
                             state,


crates/router/src/types/storage/revenue_recovery_redis_operation.rs
```diff
@@ -612,9 +613,11 @@ impl RedisTokenManager {
                         .and_modify(|v| *v += value)
                         .or_insert(value);
                 }
-                existing_token.account_update_history = token_data.account_update_history.clone();
                 existing_token.payment_processor_token_details =
                     token_data.payment_processor_token_details.clone();
+                if is_account_updater {
+                    existing_token.is_active = token_data.is_active;
+                }
```
srujanchikke:
```suggestion
                (is_account_updater).then(|| {
    existing_token.is_active = token_data.is_active;
});
```

crates/router/src/types/storage/revenue_recovery_redis_operation.rs
```diff
@@ -1326,16 +1324,26 @@ impl AccountUpdaterAction {
                     OffsetDateTime::now_utc().date(),
                     OffsetDateTime::now_utc().time(),
                 ));
+                updated_token.daily_retry_history = HashMap::new();
 
                 RedisTokenManager::upsert_payment_processor_token(
                     state,
                     customer_id,
                     updated_token,
+                    true,
```
srujanchikke, line 1333: can you add comment here

crates/router/src/types/storage/revenue_recovery_redis_operation.rs
```diff
@@ -1377,9 +1385,19 @@ impl AccountUpdaterAction {
...
+                } else {
+                    RedisTokenManager::upsert_payment_processor_token(
+                        state,
+                        customer_id,
+                        new_token,
+                        true,
+                    )
                     .await?;
-                logger::info!("Successfully updated token with new token information.")
+                    logger::info!("Successfully updated token with new token information.");
```
srujanchikke, line 1399: can we write match here

PR #10505, opened 2025-12-02
feat(scheduler): add task segregation based on application source

kashif-m:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description
This PR introduces environment isolation for scheduler services by implementing application source-based task segregation. The feature enables a dedicated Playground / CUG environment for testing new features without impacting production workflows.

### Key Changes:

- __Database Schema__: Added `application_source VARCHAR(64)` column to `process_tracker` table with migration
- __Configuration__: Added `application_source` setting (`main`/`cug`) and `cug_stream` for scheduler routing
- __Enum__: Introduced `ApplicationSource` enum with `Main` and `Cug` variants
- __Scheduler Logic__: Updated producer to partition tasks by application source and route to appropriate Redis streams
- __Task Creation__: Updated all process tracker task creation functions to accept and store application source

### Additional Changes

## Motivation and Context
This change enables complete isolation between production and Playground / CUG scheduler workflows, allowing independent testing and deployment of scheduler changes without affecting live operations. 

## How did you test it?
Tested locally by:

1. Running one of each instance
- Main router
- CUG router
- Main consumer
- CUG consumer 
- Single producer
2. Creating a payment in main application which schedules an outgoing webhook retry with `main` application source
3. Creating a payment in CUG application which schedules an outgoing webhook retry with `cug` application source
4. Wait ~2mins for the producer to pickup and schedule these tasks in consumer
5. Main consumer executes the main task
6. CUG consumer executes the CUG task

## Checklist

### DIFF
--- a/config/development.toml
+++ b/config/development.toml
@@ -61,7 +61,6 @@ id_reuse_strategy = "ignore_incoming"  # Options: "ignore_incoming" (default) or
 [server]
 # HTTP Request body limit. Defaults to 32kB
 request_body_limit = 32768
-port = 8080
 
 [secrets]
 admin_api_key = "test_admin"


config/config.example.toml
```diff
@@ -414,12 +417,14 @@ cards = [
...
 [scheduler]
 stream = "SCHEDULER_STREAM"
+cug_stream = "CUG_SCHEDULER_STREAM"
 graceful_shutdown_interval = 60000 # Specifies how much time to wait while re-attempting shutdown for a service (in milliseconds)
 loop_interval = 5000               # Specifies how much time to wait before starting the defined behaviour of producer or consumer (in milliseconds)
 
 [scheduler.consumer]
 consumer_group = "SCHEDULER_GROUP"
 disabled = false                   # This flag decides if the consumer should actively consume task
+application_source = "main"        # Application source to filter tasks for consumer
```
jagan-jaya: let's not have duplicates for application_source. use the root value itself

crates/router/src/core/refunds_v2.rs
```diff
@@ -1441,6 +1441,7 @@ pub async fn add_refund_execute_task(
         None,
         schedule_time,
         common_types::consts::API_VERSION,
+        enums::ApplicationSource::Main,
```
jagan-jaya: Why `main` here specifically?

crates/router/src/core/refunds_v2.rs
```diff
@@ -1478,6 +1479,7 @@ pub async fn add_refund_sync_task(
...
```
jagan-jaya: why `main` here? shouldn't we use application_source from conf?

crates/subscriptions/src/workflows/invoice_sync.rs
```diff
@@ -411,6 +411,7 @@ pub async fn create_invoice_sync_job(
         Some(0),
         common_utils::date_time::now(),
         common_types::consts::API_VERSION,
+        common_enums::ApplicationSource::Main,
```
jagan-jaya: why `Main` used here?

jagan-jaya (approved):

config/config.example.toml
```diff
@@ -2,6 +2,11 @@
 # all the available configuration options, and is intended to be used
 # solely as a reference. Please copy this file to create a config.
 
+# This is used to identify the running application instance
+# Check common_enums::ApplicationSource for available options
+# For more details - https://github.com/juspay/hyperswitch/issues/10489
+application_source = "main"
```
SanchithHegde, line 8: @kashif-m Adding this as an item at the top level makes it dependent on the order / position, when specifying in the TOML file, for it to be picked up correctly. It must be specified at the top of the TOML file itself, before any sections (such as `[server]`) are defined.

Not sure if this would be a concern when deploying via the various options we support.

Would recommend keeping this nested instead, to avoid such quirks...

config/config.example.toml
SanchithHegde, line 1: Please replicate these changes in the other config files as well:

- `config/docker_compose.toml`
- `config/deployments/env_specific.toml`: Include the same comments as included in the `config.example.toml` file.

config/config.example.toml
kashif-m: Keeping this at top level makes logical sense, since it's part of the router instance. I wanted to move it inside `server` but looking at other configs there did not make a lot of sense.

Will check if we can move it inside a different config 👍

config/config.example.toml
kashif-m: https://github.com/juspay/hyperswitch/pull/10596

Review?

config/config.example.toml
SanchithHegde: We can also have a different section, if we're able to come up with a suitable name.

PR #10544, opened 2025-12-05
fix: Invoke UCS in shadow mode even when the HS request fails HS↔UCS tunnel

AmitsinghTanwar007:
## Type of Change

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

 Enhanced shadow UCS (Unified Connector Service) testing to execute even when the primary HS (Hyperswitch) direct connector call fails. This ensures comprehensive comparison data is collected for all payment flows, improving observability and shadow testing accuracy.

### Additional Changes

## How did you test it?

1)You can try to run a req through HS<>UCS tunnel where UCS is in shadow mode.
2) make environment in such a way that hs req fails ( in my scenario i am changing the url of connector so that it fails to connect)
3) check does the req goes to ucs or not in shadow mode.
if it goes then the change is successful.
you may not be able to test this things in hosted enviroments.

## Checklist


crates/router/src/core/payments/helpers.rs
```diff
@@ -8659,10 +8659,15 @@ where
...
     )
-    .await?;
+    .await;
+
+    // Spawn shadow UCS call in background regardless of HS result
+    // Use the initial router_data as fallback when HS fails
+    let direct_router_data = match &result {
+        Ok(success_data) => Some(success_data.0.clone()),
+        Err(_) => None,
+    };
```
hrithikesh026, line 8669: should we not compare the router data if HS call fails???

AmitsinghTanwar007: in error case we will not get the updated router data so there is no need to compare them

PR #5164, opened 2024-06-28
feat: encryption service integration to support batch encryption and decryption

ArjunKarthik:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

### Additional Changes

- [ ] This PR modifies the API contract
- [ ] This PR modifies the database schema
- [ ] This PR modifies application configuration/environment variables

## Motivation and Context
PCI compliance requires application to follow some standards where encryption and decryption should not happen within the application as snapshot might have DEK and data together. So separate service should be used for encryption and decryption. In application wherever we are calling encrypt and decrypt it will be replaced by API call to encryption service.

## How did you test it?
Changes can be verified only through logs. Look for "Fall back to Application Encryption" or "Fall back to Application Decryption" in the logs post deployment after running below tests. If logs are present encryption/decryption failed with encryption service and fall back to the application encryption which has to be investigated.

**Test with merchant created before the changes deployed**
       1. Create Payment
       2. Retrieve Payment
       3. Create Refund
       4. Retrieve Payment
**Test with merchant created after the changes deployed**
       1. Create Merchant
       2. Create API Key
       3. Create MCA
       4. Create Payment
       6. Retrieve Payment
       7. Create Refund
       8. Retrieve Payment
**Requires sanity on all the flows**
       
## Checklist

- [x] I formatted the code `cargo +nightly fmt --all`
- [x] I addressed lints thrown by `cargo clippy`
- [x] I reviewed the submitted code
- [ ] I added unit tests for my changes where possible

PR #10572, opened 2025-12-08
fix: fix fallback decryption failures in case of unavailability of encryption service

SanchithHegde:
## Type of Change

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [x] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

This PR includes 2 fixes / improvements around the encryption service integration code:

1. Updates the application fallback decryption logic to correctly handle the differences in format of data encrypted by the encryption service and router (using the merchant / user key store).

   - This was done as the data encrypted by the encryption service is of the format `v{version_number}:<base64 encoded encrypted bytes>`, while that encrypted by the router itself just consists of the encrypted bytes (no base64 encoding involved, no version information included).
   - To address this, we now call the `obtain_data_to_decrypt_locally()` function whenever we are falling back to application decryption, in case of failures to decrypt data with the encryption service.

2. Updates the storage and retrieval logic of the merchant and user key store entries to be encrypted and decrypted locally instead of being encrypted / decrypted using the encryption service.

   - There is a bug (introduced in #5164 I believe) which causes the merchant and user key store entries to be encrypted using the encryption service instead of router's master encryption key, which causes decryption failures when trying to retrieve either of them. 
   - This fixes it by correctly encrypting and decrypting the merchant and user key store entries using the router's master encryption key.
   - I'm aware that there would be existing merchant and user key store entries in the router's database that would still not work. I'll take up the task of fixing the existing entries in a separate PR.

> [!NOTE]
> The changes in this PR do not accommodate key rotation being performed at the encryption service side: if the data encryption key for a merchant / user has been rotated in encryption service and the encryption service is unavailable, then the fallback decryption logic would not work for some of the encrypted data in router's database, since the router may not have the updated data encryption keys with it.

## Motivation and Context

This PR includes fixes to some bugs that caused a downtime a while ago. The changes in this PR should allow the router to function well even in case the encryption service is unavailable for a short duration, thus making the router slightly more reliable.

### Unit Tests

This PR includes some unit tests (tests were AI generated, I've reviewed the test cases however) for the code that handles the differences in format of encrypted data between the encryption service and router. They can be run using the command:

```shell
cargo test --package hyperswitch_domain_models --features v1 --lib type_encryption::tests
```

Attaching a screenshot of the unit tests run:

Screenshot of successful unit test run

### Chaos Testing

In addition to the unit tests, I've tested the changes in this PR by running both router and encryption service locally, creating some merchants, payments, etc. Then, I brought down the encryption service, and tried out various scenarios (merchant account create, payments create, merchant account retrieve, payments retrieve, etc.) using Postman.

Before this PR / on the `main` branch, this would be the outcome in case the encryption service goes down:

*Note that "ES" is an abbreviation for "encryption service", and that "data" could be any entity in router's database that contains encrypted information such as customer, payment method, business profile, etc. ✅ indicates a successful API response, while ❌ indicates a failed API response.*

|                                                                                 | **Merchant created before / without ES integration (before key transfer)** |                               **Merchant created before / without ES integration (after key transfer)**                                |   **Merchant created after / with ES integration**    |
| :-----------------------------------------------------------------------------: | :------------------------------------------------------------------------: | :------------------------------------------------------------------------------------------------------------------------------------: | :---------------------------------------------------: |
|                          **Merchant Account Retrieve**                          |                             ✅ (Fallback works)                             |                                                           ✅ (Fallback works)                                                           |  ❌ (401, because merchant account decryption fails)   |
|            **Retrieve (decrypt) data created before ES integration**            |                             ✅ (Fallback works)                             |                                                           ✅ (Fallback works)                                                           |                          N/A                          |
|                          **Create (encrypt) new data**                          |                             ✅ (Fallback works)                             |      ✅ (Fallback works, can create new customer, new payment (only with a new customer, not with customer encrypted by ES), etc.)      | ❌ (401, merchant account retrieve / decryption fails) |
| **Retrieve (decrypt) data created after ES integration (data encrypted by ES)** |     ✅ / N/A (Data would be encrypted by router itself, fallback works)     | ❌ (500, merchant account is decrypted with fallback, while data cannot be. Customer encrypted by ES cannot be decrypted, for example.) | ❌ (401, merchant account retrieve / decryption fails) |

With this PR's changes included, the various scenarios are as follows, if / when the encryption service is down:

|                                                                                 | **Merchant created before / without ES integration (before key transfer)** | **Merchant created before / without ES integration (after key transfer)** | **Merchant created after / with ES integration** |
| :-----------------------------------------------------------------------------: | :------------------------------------------------------------------------: | :-----------------------------------------------------------------------: | :----------------------------------------------: |
|                          **Merchant Account Retrieve**                          |                                     ✅                                      |                                     ✅                                     |                        ✅                         |
|            **Retrieve (decrypt) data created before ES integration**            |                                     ✅                                      |                                     ✅                                     |                       N/A                        |
|                          **Create (encrypt) new data**                          |                                     ✅                                      |                                     ✅                                     |                        ✅                         |
| **Retrieve (decrypt) data created after ES integration (data encrypted by ES)** |                                  ✅ / N/A                                   |                                     ✅                                     |                        ✅                         |

> [!NOTE]
> This PR does not address the incorrect entries in the merchant and user key store tables (which were encrypted by the encryption service) that may exist in the database. They would be addressed in a separate PR.

## Checklist

### DIFF
--- a/crates/router/src/types/domain/user.rs
+++ b/crates/router/src/types/domain/user.rs
@@ -1229,7 +1229,7 @@ impl UserFromStorage {
                 key: domain_types::crypto_operation(
                     key_manager_state,
                     type_name!(UserKeyStore),
-                    domain_types::CryptoOperation::Encrypt(key.to_vec().into()),
+                    domain_types::CryptoOperation::EncryptLocally(key.to_vec().into()),
                     Identifier::User(self.get_user_id().to_string()),
                     master_key,
                 )
--- a/crates/router/src/types/domain/user_key_store.rs
+++ b/crates/router/src/types/domain/user_key_store.rs
@@ -44,14 +44,14 @@ impl super::behaviour::Conversion for UserKeyStore {
             key: crypto_operation(
                 state,
                 type_name!(Self::DstType),
-                CryptoOperation::Decrypt(item.key),
+                CryptoOperation::DecryptLocally(item.key),
                 identifier,
                 key.peek(),
             )
             .await
             .and_then(|val| val.try_into_operation())
             .change_context(ValidationError::InvalidValue {
-                message: "Failed while decrypting customer data".to_string(),
+                message: "Failed while decrypting user key store".to_string(),
             })?,
             user_id: item.user_id,
             created_at: item.created_at,


crates/hyperswitch_domain_models/src/type_encryption.rs
```diff
@@ -1315,6 +1361,71 @@ where
...
+{
+    use base64::Engine;
+    use common_utils::consts::BASE64_ENGINE;
+    use error_stack::ResultExt;
+    use masking::PeekInterface;
+
+    if let Some((_version, base64_encoded_data)) = split_version_prefix(encrypted_data.peek()) {
+        // Data encrypted by encryption service.
+        // Split data at colon (to remove version prefix), base64 decode and then proceed with decryption.
+        router_env::logger::debug!("Attempting to decrypt data encrypted by encryption service");
```
SanchithHegde, line 1379: I can remove the log lines in this function, if you folks think this may become too noisy.

tsdk02 (approved):

crates/hyperswitch_domain_models/src/type_encryption.rs
```diff
@@ -1315,6 +1361,71 @@ where
...
+fn obtain_data_to_decrypt_locally<S>(
+    encrypted_data: Secret<Vec<u8>, S>,
+) -> CustomResult<Secret<Vec<u8>, S>, CryptoError>
+where
+    S: masking::Strategy<Vec<u8>>,
+{
+    use base64::Engine;
+    use common_utils::consts::BASE64_ENGINE;
+    use error_stack::ResultExt;
+    use masking::PeekInterface;
```
ShankarSinghC, line 1374: Why are we having these as function level imports ?

crates/hyperswitch_domain_models/src/merchant_key_store.rs
```diff
@@ -46,14 +46,14 @@ impl super::behaviour::Conversion for MerchantKeyStore {
             key: crypto_operation(
                 state,
                 type_name!(Self::DstType),
-                CryptoOperation::Decrypt(item.key),
+                CryptoOperation::DecryptLocally(item.key),
```
sai-harsha-vardhan, line 49: nit: Should we name this to ApplicationDecrypt or some other name which indicates that decryption will happen within application?

crates/hyperswitch_domain_models/src/type_encryption.rs
SanchithHegde: No specific reason to be honest.

There are some imports within `mod encrypt`, while some are top-level. Looks like there's nothing strictly being followed in this file.

It's also that 3 of the imports are traits, so I kept them within the function to provide an idea as to what traits are required by the code in this function.

PR #10593, opened 2025-12-10
feat(theme): implement theme versioning API and database schema

Muditbhatia12:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

This PR implements a theme versioning system that allows tracking theme configuration changes over time. 

Key changes:
- Added `theme_config_version` column to the themes table
- Created a new API endpoint `GET /themes/{theme_id}/version` to retrieve the current theme version
- Updated theme creation and update flows to generate and track version timestamps
- Added `ThemeVersionResponse` API model'
- Added redis caching for version
- Implemented proper versioning logic in both database and application layers

### Additional Changes

## Motivation and Context

The theme versioning system enables:
- Tracking when theme configurations are modified
- Supporting cache invalidation based on theme versions
- Enabling future features like theme rollback or change history

## How did you test it?

- Request
```
curl --location 'http://localhost:8080/user/theme/ad93d45c-9c5c-45af-84ef-977b9d8ad928/version'
```
- Response
```
{
    "theme_id": "ad93d45c-9c5c-45af-84ef-977b9d8ad928",
    "theme_config_version": "1765358954"
}
```

- Database migration tested successfully
- New API endpoint returns correct theme version information
- Redis flow checked
- Theme creation and update flows properly generate version timestamps

## Checklist

### DIFF
--- a/config/deployments/integration_test.toml
+++ b/config/deployments/integration_test.toml
@@ -216,7 +216,7 @@ payout_connector_list = "nomupay,stripe,wise"
 connectors_with_delayed_session_response = "trustpay,payme"     # List of connectors which have delayed session response
 
 [save_payment_method_on_session.unsupported_payment_methods]
-pay_later = ["klarna", "affirm"]
+pay_later = "klarna,affirm"
 
 [zero_mandates.supported_payment_methods]
 bank_debit.ach = { connector_list = "gocardless,adyen,payload" }
--- a/config/deployments/production.toml
+++ b/config/deployments/production.toml
@@ -185,7 +185,7 @@ netcetera.base_url = "https://{{merchant_endpoint_prefix}}.3ds-server.prod.netce
 connectors_with_delayed_session_response = "trustpay,payme"       # List of connectors which have delayed session response
 
 [save_payment_method_on_session.unsupported_payment_methods]
-pay_later = ["klarna", "affirm"]
+pay_later = "klarna,affirm"
 
 [dummy_connector]
 enabled = false                                                         # Whether dummy connector is enabled or not
--- a/config/deployments/sandbox.toml
+++ b/config/deployments/sandbox.toml
@@ -185,7 +185,7 @@ netcetera.base_url = "https://{{merchant_endpoint_prefix}}.3ds-server.prev.netce
 connectors_with_delayed_session_response = "trustpay,payme" # List of connectors which have delayed session response
 
 [save_payment_method_on_session.unsupported_payment_methods]
-pay_later = ["klarna", "affirm"]
+pay_later = "klarna,affirm"
 
 [dummy_connector]
 enabled = true                                                          # Whether dummy connector is enabled or not
--- a/config/development.toml
+++ b/config/development.toml
@@ -1113,7 +1113,7 @@ discord_invite_url = "https://discord.gg/wJZ7DVW8mm"
 connectors_with_delayed_session_response = "trustpay,payme"
 
 [save_payment_method_on_session.unsupported_payment_methods]
-pay_later = ["klarna", "affirm"]
+pay_later = "klarna,affirm"
 
 [webhook_source_verification_call]
 connectors_with_webhook_source_verification_call = "paypal"
--- a/config/docker_compose.toml
+++ b/config/docker_compose.toml
@@ -399,7 +399,7 @@ cards = [
 connectors_with_delayed_session_response = "trustpay,payme"
 
 [save_payment_method_on_session.unsupported_payment_methods]
-pay_later = ["klarna", "affirm"]
+pay_later = "klarna,affirm"
 
 [webhook_source_verification_call]
 connectors_with_webhook_source_verification_call = "paypal"
--- a/crates/router/src/configs/settings.rs
+++ b/crates/router/src/configs/settings.rs
@@ -187,6 +187,7 @@ pub struct Settings<S: SecretState> {
 
 #[derive(Debug, Deserialize, Clone, Default)]
 pub struct OnSessionConfig {
+    #[serde(default, deserialize_with = "deserialize_hashmap")]
     pub unsupported_payment_methods:
         HashMap<enums::PaymentMethod, HashSet<enums::PaymentMethodType>>,
 }


crates/router/src/core/user/theme.rs
```diff
@@ -572,3 +584,36 @@ pub async fn get_user_theme_using_lineage(
...
+        .store
+        .find_theme_by_theme_id(theme_id.clone())
+        .await
+        .to_not_found_response(UserErrors::ThemeNotFound)?;
+
+    theme_utils::set_theme_version_in_redis(
+        &state,
+        &theme_id,
+        theme.theme_config_version.clone(),
+        crate::consts::user::REDIS_THEME_CONFIG_VERSION_TTL_IN_SECS,
```
ThisIsMani: Import crate::consts.

crates/router/src/utils/user/theme.rs
```diff
@@ -12,6 +12,57 @@ use crate::{
...
+    theme_id: &str,
+) -> UserResult<()> {
+    let redis_key = get_theme_version_redis_key(theme_id);
+    let redis_conn = super::get_redis_connection_for_global_tenant(state)?;
+
+    redis_conn
+        .delete_key(&redis_key.into())
+        .await
+        .change_context(UserErrors::InternalServerError)
+        .map(|_| ())
```
ThisIsMani, line 64: Will this throw error if key is not present in redis?

migrations/2025-12-09-091856-0000_add_version_to_themes_table/down.sql
```diff
@@ -0,0 +1,3 @@
+-- This file should undo anything in `up.sql`
+ALTER TABLE themes
+DROP COLUMN  IF EXISTS theme_config_version;
```
ThisIsMani:
```suggestion
DROP COLUMN IF EXISTS theme_config_version;
```

migrations/2025-12-09-091856-0000_add_version_to_themes_table/up.sql
```diff
@@ -0,0 +1,3 @@
+-- Your SQL goes here
+ALTER TABLE themes
```
sai-harsha-vardhan, line 2: Why do we want to have a not null column added here? This may not be backward compatible right? It's recommended to have a nullable column in DB with application defaults

crates/router/src/utils/user/theme.rs
```diff
@@ -12,6 +12,57 @@ use crate::{
     services::authentication::UserFromToken,
 };
 
+// Redis key format: {theme_id}_version
+pub fn get_theme_version_redis_key(theme_id: &str) -> String {
+    format!("{}_version", theme_id)
+}
```
SanchithHegde: Nit: Do you need this to be public, or can we make this private?

SanchithHegde (commented):
I have a few questions about the PR, would be great if you could clarify them / address these in the issue or PR description:

1. What is the expectation when we say "theme versioning"?
   - When a theme is updated, should it be a new entry in the database, or do we just update a version field in the database?
   - What about the files in the S3 bucket, should new files be created with each update?
   - If we are only supposed to update the version field in the table without creating new files in S3 (but overwrite existing ones), then what's the point of having the `theme_config_version` field? Isn't it similar to the `last_modified_at` field, in a way?
2. What's the outcome of the PR? What are we trying to solve?
   - Is this supposed to enable rollbacks to previous theme versions?
   - Are we expecting to have an audit trail of theme updates, over time?
   - Anything else?

Muditbhatia12:
> I have a few questions about the PR, would be great if you could clarify them / address these in the issue or PR description:
> 
> 1. What is the expectation when we say "theme versioning"?
>    
>    * When a theme is updated, should it be a new entry in the database, or do we just update a version field in the database?

We just update a version field.
>    * What about the files in the S3 bucket, should new files be created with each update?

As long as the theme_data is changed the theme_config_version should be updated.
>    * If we are only supposed to update the version field in the table without creating new files in S3 (but overwrite existing ones), then what's the point of having the `theme_config_version` field? Isn't it similar to the `last_modified_at` field, in a way?

`last_modified_at` will change even when just the email_config is updated(this can be done using postman). This column is to track the changes in theme_data (which is sent to the s3 as themes.json).
> 2. What's the outcome of the PR? What are we trying to solve?
>    
>    * Is this supposed to enable rollbacks to previous theme versions?
>    * Are we expecting to have an audit trail of theme updates, over time?
>    * Anything else?

The purpose of this pr is to enable cache invalidation using query params when the FE makes calls to the cdn.The version will be sent as a query param.

migrations/2025-12-09-091856-0000_add_version_to_themes_table/up.sql
Muditbhatia12: We dont have domain models for any of these flows.

crates/diesel_models/src/user/theme.rs
```diff
@@ -95,22 +98,38 @@ pub struct ThemeUpdateInternal {
     pub email_background_color: Option<String>,
     pub email_entity_name: Option<String>,
     pub email_entity_logo_url: Option<String>,
+    pub last_modified_at: Option<PrimitiveDateTime>,
```
SanchithHegde: Nit: Make this a required field, that way you can ensure that all updates to the record in the table would enforce us to update the `last_modified_at` value.

crates/router/src/utils/user/theme.rs
```diff
@@ -12,6 +12,57 @@ use crate::{
     services::authentication::UserFromToken,
 };
 
+// Redis key format: {theme_id}_version
+fn get_theme_version_redis_key(theme_id: &str) -> String {
+    format!("{}_version", theme_id)
+}
```
SanchithHegde, line 19: Nit: Can probably add `#[inline]` annotation on this.

SanchithHegde (commented):
Could there be any issues in cases of staggered deployments or rollbacks to a previous version of the code, where the theme update (which happened with the newer version of the code) is not picked up by the previous version of the code?

Muditbhatia12:
> Could there be any issues in cases of staggered deployments or rollbacks to a previous version of the code, where the theme update (which happened with the newer version of the code) is not picked up by the previous version of the code?
> 
> Other than that, looks good to me!

Whenever a theme is created or updated we are putting few things in db and the files in s3, this pr doesn't change anything related to s3 and since the db has defaults there shouldn't be any problems in the rollback case.

ThisIsMani (approved):

PR #10620, opened 2025-12-10
fix(ucs): properly map `connector_request_reference_id` to `request_ref_id`

pixincreate:
## Type of Change

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

ucs transformers has a few bugs that are addressed in this pr:

- `connector_request_ref_id` should contain the `attempt_id` (or, in other words, the id should be generated at the merchant's end. in this case, hyperswitch). but, `connector_transaction_id` is being passed instead which is wrong. this pr properly maps it while introducing a new field. once the ucs changes are merged, this pr can be merged.

### Additional Changes

## Motivation and Context

jpmorgan connector in ucs, during diff check, should some diffs that were totally wrong -- void and capture flow straight up failed. and during capture and void, simply because state param was not being passed, it would try to re-generate the access token while it was unnecessary, but would fail.

## How did you test it?

tested via jpmorgan through ucs: https://github.com/juspay/connector-service/pull/358

this change only affects the psync flow. we previously used to get connector transaction id and this pr fixes that by replacing it with attempt id. below attached is the diff check showing no diffs after fixes:

image

logs:

hyperswitch:
image

ucs: 
image

## Checklist

### DIFF
--- a/crates/router/src/core/unified_connector_service/transformers.rs
+++ b/crates/router/src/core/unified_connector_service/transformers.rs
@@ -731,12 +731,19 @@ impl
             amount: router_data.request.amount.get_amount_as_i64(),
             currency: currency.into(),
             state,
-            connector_metadata: None,
-            merchant_account_metadata: HashMap::new(),
-            metadata: HashMap::new(),
+            connector_metadata: router_data
+                .request
+                .connector_meta
+                .as_ref()
+                .map(serde_json::to_string)
+                .transpose()
+                .change_context(UnifiedConnectorServiceError::RequestEncodingFailed)?
+                .map(|s| s.into()),
             sync_type: Some(
                 payments_grpc::SyncRequestType::foreign_from(&router_data.request.sync_type).into(),
             ),
+            merchant_account_metadata: HashMap::new(),
+            metadata: HashMap::new(),
         })
     }
 }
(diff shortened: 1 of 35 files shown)


Aishwariyaa-Anand (commented):
this can only be merged post UCS proto changes for connector_order_ref_id.

crates/router/src/core/unified_connector_service/transformers.rs
```diff
@@ -1787,7 +1787,7 @@ impl
                     router_data.connector_request_reference_id.clone(),
                 )),
             }),
-            mandate_reference,
+            mandate_reference_id: None,
```
Aishwariyaa-Anand: why is this None?

Aishwariyaa-Anand (dismissed):

crates/router/src/core/unified_connector_service/transformers.rs
pixincreate: https://github.com/juspay/connector-service/commit/a910df283e18aa625070dc9aea6cd31d83f7167e#diff-b458a9878e13fcb745a361945e619fbe041a764b248293f28ba33f7d9199c32eR2139

cc: @Saptak88

crates/router/src/core/unified_connector_service/transformers.rs
```diff
@@ -679,14 +679,13 @@ impl
...
+        let connector_order_reference_id = router_data.request.connector_reference_id.clone();
 
-        let connector_ref_id = router_data
-            .request
-            .connector_reference_id
-            .clone()
-            .map(|id| Identifier {
-                id_type: Some(payments_grpc::identifier::IdType::Id(id)),
-            });
+        let request_ref_id = Some(Identifier {
```
hrithikesh026, line 684: wouldn't this be backwards incompatible? UCS is currently expecting `connector_reference_id` and now we are sending `connector_request_reference_id`

pixincreate: yeah, this was a bug on the ucs side psync which has been fixed in [this](https://github.com/juspay/connector-service/commit/3dcfd30ef1405c81ba711c370397cabba277e854) commit.

so, now, request_ref_id will contain the connector request reference id (which is usually the attempt id) and connector_order_reference_id will contain the connector reference id (and this is usually the connector transaction id).

hrithikesh026 (dismissed):

hrithikesh026 (approved):

Aishwariyaa-Anand (approved):

PR #10645, opened 2025-12-12
feat(core): propagate metadata from locker response to payment method response

Sakilmostak:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description:
Currently, we support using the Hyperswitch application with any card vault that follows the same API contract as the Hyperswitch card vault. Some merchant store some additional details in their vault. The requirement is that when the Get Payment Methods API is called, we need to call the locker to fetch these additional details along with the card details, and then include these additional details in the metadata field of the Payment Methods response.

### Additional Changes

adds metadata to locker response

## How did you test it?

Tested through Postman in Local:
Note: Can't be tested on Integ or Sandbox

- Create local changes in locker repo to pass metadata in response
- Create a Payment Method Entry using Payment Method Create API
```
curl --location '{{baseUrl}}/payment_methods' \
--header 'api-key:{{api_key}}' \
--data '{
    "payment_method": "card",
    "payment_method_issuer": "Visa",
    "card": {
        "card_exp_month": "03",
        "card_exp_year": "2030",
        "card_holder_name": "Test Holder"
        
    },
    "billing": {
        "address": {
            "line1": "1467",
            "line2": "Harrison Street",
            "line3": "Harrison Street",
            "city": "San Fransico",
            "state": "California",
            "zip": "94122",
            "country": "US",
            "first_name": "John",
            "last_name": "Doe"
        }
    },
    "customer_id": "{{customer_id}}",
    "metadata": {
        "city": "NY",
        "unit": "245"
    }
}'
```
- Do a retrieve using payment_method_id, following should be the response:
```
{
    "merchant_id": "merchant_1766409748",
    "customer_id": "cus_LTfnNAXrLavew0AmlpXm",
    "payment_method_id": "pm_fkfB86WJKVswXiCZxU3F",
    "payment_method": "card",
    "card": {
        "scheme": null,
        "issuer_country": null,
        "issuer_country_code": null,
        "last4_digits": "4242",
        "expiry_month": "03",
        "expiry_year": "2030",
        "card_token": null,
        "card_holder_name": "Test Holder",
        "card_fingerprint": null,
        "nick_name": null,
        "card_network": null,
        "card_isin": null,
        "card_issuer": null,
        "card_type": null,
        "saved_to_locker": true
    },
    "recurring_enabled": false,
    "payment_experience": [
        "redirect_to_url"
    ],
    "metadata": {
        "city": "NY",
        "unit": "245",
        "par": "sakil",
        "upi": "shankar"
    },
    "created": "2025-12-22T13:26:32.375Z",
    "last_used_at": "2025-12-22T13:26:32.375Z",
    "client_secret": "pm_fkfB86WJKVswXiCZxU3F_secret_NbBCv8uZ3QB8uvPFbi8N"
}
```

## Checklist

### DIFF
--- a/crates/router/src/core/payment_methods/cards.rs
+++ b/crates/router/src/core/payment_methods/cards.rs
@@ -907,9 +907,9 @@ impl PaymentMethodsController for PmCards<'_> {
             .await
             .to_not_found_response(errors::ApiErrorResponse::PaymentMethodNotFound)?;
 
-        let (card, payload_metadata) =
+        let (card, locker_metadata) =
             if pm.get_payment_method_type() == Some(enums::PaymentMethod::Card) {
-                let (card_detail, payload_metadata) = if self.state.conf.locker.locker_enabled {
+                let (card_detail, locker_metadata) = if self.state.conf.locker.locker_enabled {
                     let locker_card_response = get_card_from_locker(
                         self.state,
                         &pm.customer_id,
@@ -932,7 +932,7 @@ impl PaymentMethodsController for PmCards<'_> {
                         None,
                     )
                 };
-                (Some(card_detail), payload_metadata)
+                (Some(card_detail), locker_metadata)
             } else {
                 (None, None)
             };
@@ -947,7 +947,7 @@ impl PaymentMethodsController for PmCards<'_> {
                 bank_transfer: None,
                 card,
                 // merge payload metadata with pm metadata
-                metadata: pm.get_payment_method_metadata(payload_metadata),
+                metadata: pm.get_payment_method_metadata(locker_metadata),
                 created: Some(pm.created_at),
                 recurring_enabled: Some(false),
                 installment_payment_enabled: Some(false),
@@ -1848,6 +1848,7 @@ pub async fn update_customer_payment_method(
                 bank_transfer: None,
                 card: Some(existing_card_data),
                 wallet: None,
+                // since locker is not called, no metadata is merged to payment method metadata
                 metadata: pm.get_payment_method_metadata(None),
                 created: Some(pm.created_at),
                 recurring_enabled: Some(false),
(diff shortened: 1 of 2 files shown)


crates/router/src/core/payment_methods/cards.rs
```diff
@@ -938,7 +948,8 @@ impl PaymentMethodsController for PmCards<'_> {
                 #[cfg(feature = "payouts")]
                 bank_transfer: None,
                 card,
-                metadata: pm.metadata,
+                // merge payload metadata with pm metadata
+                metadata: merge_json_values(payload_metadata, pm.metadata),
```
ShankarSinghC: Can we have it as a impl on `PaymentMethod`

pm.get_payment_method_metadata(locker_card_response.metadata)

crates/router/src/core/payment_methods/utils.rs
```diff
@@ -907,3 +907,35 @@ pub(super) async fn delete_payment_token_data(
...
+                    ))
+                }
+                (first_val, _) => Some(common_utils::pii::SecretSerdeValue::new(first_val)),
+            }
+        }
+        (Some(first_val), None) => Some(first_val),
+        (None, Some(second_val)) => Some(second_val),
+        (None, None) => None,
+    }
+}
```
ShankarSinghC: Please add comment lines explaining the code
For example with respect to the above logic if the `first` and `second` has the same keys then the value will be updated with that of the `second`.

Also consider using zip to avoid multiple match arms
```
match first.zip(second) { 
 Some(data) => {},
 None => (first.or(second))
}
```

kashif-m (approved):

PR #10662, opened 2025-12-15
ci(cypress): Increase the number of connectors that cypress tests are being run on while merging and run them in parallel

NISHANTH1221:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description
This PR increases the coverage of the tests from two connectors currently ( stripe, cybersource ) to more connectors when the pr gets merged. And this PR executes the connectors in parallel in sets of three. The current support is extended to 25 connector out of which two are mandatory and 23 are non mandatory connectors. This pr divides the optional connectors into batches of 3 and executes the tests form them.

### Additional Changes

## How did you test it?

A sample run with the above changes: https://github.com/juspay/hyperswitch/actions/runs/20748548238?pr=10662

## Checklist

### DIFF
--- a/.github/workflows/cypress-tests-runner.yml
+++ b/.github/workflows/cypress-tests-runner.yml
@@ -17,14 +17,14 @@ env:
   CARGO_INCREMENTAL: 0
   CARGO_NET_RETRY: 10
   PAYMENTS_CONNECTORS: "cybersource stripe"
-  OPTIONAL_PAYMENTS_CONNECTORS_BATCH_1: "braintree redsys nmi"
-  OPTIONAL_PAYMENTS_CONNECTORS_BATCH_2: "nexixpay iatapay cashtocode"
-  OPTIONAL_PAYMENTS_CONNECTORS_BATCH_3: "fiuu noon adyen"
-  OPTIONAL_PAYMENTS_CONNECTORS_BATCH_4: "novalnet airwallex checkout"
-  OPTIONAL_PAYMENTS_CONNECTORS_BATCH_5: "nuvei payload paypal"
-  OPTIONAL_PAYMENTS_CONNECTORS_BATCH_6: "trustpay worldpay zift"
-  OPTIONAL_PAYMENTS_CONNECTORS_BATCH_7: "worldpayxml finix cryptopay"
-  OPTIONAL_PAYMENTS_CONNECTORS_BATCH_8: "peachpayments mollie"
+  OPTIONAL_PAYMENTS_CONNECTORS_BATCH_1: "noon payload nmi"
+  OPTIONAL_PAYMENTS_CONNECTORS_BATCH_2: "trustpay braintree adyen"
+  OPTIONAL_PAYMENTS_CONNECTORS_BATCH_3: "nexixpay cryptopay finix"
+  OPTIONAL_PAYMENTS_CONNECTORS_BATCH_4: "iatapay cashtocode peachpayments"
+  OPTIONAL_PAYMENTS_CONNECTORS_BATCH_5: "worldpayxml checkout mollie"
+  OPTIONAL_PAYMENTS_CONNECTORS_BATCH_6: "fiuu worldpay zift"
+  OPTIONAL_PAYMENTS_CONNECTORS_BATCH_7: "aiwallex novalnet redsys"
+  OPTIONAL_PAYMENTS_CONNECTORS_BATCH_8: "paypal nuvei"
   ALPHA_PAYMENTS_CONNECTORS: "celero silverflow"
   RUST_BACKTRACE: short
   RUSTUP_MAX_RETRIES: 10


.github/workflows/cypress-tests-runner.yml
```diff
@@ -132,11 +141,14 @@ jobs:
           overwrite: true
           path: target/debug/router
           retention-days: 1
-
+
```
pixincreate: nit: 


```suggestion

```

.github/workflows/cypress-tests-runner.yml
```diff
@@ -164,14 +176,28 @@ jobs:
 
     steps:
       - name: Skip tests for PRs from forks
-        if: ${{ env.RUN_TESTS == 'false' }}
+        if: ${{ env.RUN_TESTS == 'false' }}
```
pixincreate: nit:


```suggestion
        if: ${{ env.RUN_TESTS == 'false' }}
```

.github/workflows/cypress-tests-runner.yml
```diff
@@ -164,14 +176,28 @@ jobs:
...
       - name: Checkout repository
         if: ${{ env.RUN_TESTS == 'true' }}
         uses: actions/checkout@v4
 
+      - name: Install Rust
+        if: ${{ env.RUN_TESTS == 'true' }}
+        uses: dtolnay/rust-toolchain@master
+        with:
+          toolchain: stable 2 weeks ago
+          components: llvm-tools-preview, rustfmt
```
pixincreate: why do we need `llvm-tools-preview` and `rustfmt`?

.github/workflows/cypress-tests-runner.yml
```diff
@@ -234,7 +260,7 @@ jobs:
         uses: actions/setup-node@v4
         with:
           node-version: "24"
-
+
```
pixincreate: nit:


```suggestion

```

can you remove trailing white spaces?

.github/workflows/cypress-tests-runner.yml
```diff
@@ -286,24 +312,395 @@ jobs:
...
+        options: >-
+          --health-cmd pg_isready
+          --health-interval 10s
+          --health-timeout 5s
+          --health-retries 5
+        ports:
+          - 5432:5432
+
+    steps:
+
```
pixincreate:
```suggestion
```

.github/workflows/cypress-tests-runner.yml
```diff
@@ -286,24 +312,395 @@ jobs:
...
+      - name: Skip tests for PRs
```
pixincreate:
```suggestion
      - name: Skip tests running on PRs raised by forks
```

.github/workflows/cypress-tests-runner.yml
```diff
@@ -286,24 +312,395 @@ jobs:
...
+        if: ${{ (env.RUN_OPTIONAL_CONNECTORS_TESTS == 'false')  }}
+        shell: bash
+        run: echo 'Skipping tests for PRs from forks'
+
+      - name: Checkout repository
+        if: ${{ env.RUN_OPTIONAL_CONNECTORS_TESTS == 'true' }}
+        uses: actions/checkout@v4
```
pixincreate:
```suggestion
        uses: actions/checkout@v5
```

.github/workflows/cypress-tests-runner.yml
```diff
@@ -286,24 +312,395 @@ jobs:
...
+        options: >-
+          --health-cmd pg_isready
+          --health-interval 10s
+          --health-timeout 5s
+          --health-retries 5
+        ports:
+          - 5432:5432
+
+    steps:
+      - name: Skip tests for PRs
```
pixincreate: same as previous, needs update

.github/workflows/cypress-tests-runner.yml
```diff
@@ -286,24 +312,395 @@ jobs:
...
+        if: ${{ env.RUN_OPTIONAL_CONNECTORS_TESTS == 'false' }}
+        shell: bash
+        run: echo 'Skipping tests for PRs from forks'
+
+      - name: Checkout repository
+        if: ${{ env.RUN_OPTIONAL_CONNECTORS_TESTS == 'true' }}
+        uses: actions/checkout@v4
```
pixincreate:
```suggestion
        uses: actions/checkout@v5
```

.github/workflows/cypress-tests-runner.yml
```diff
@@ -286,24 +312,395 @@ jobs:
...
+
+      - name: Install Rust
+        if: ${{ env.RUN_OPTIONAL_CONNECTORS_TESTS == 'true' }}
+        uses: dtolnay/rust-toolchain@master
+        with:
+          toolchain: stable 2 weeks ago
+          components: llvm-tools-preview, rustfmt
```
pixincreate: why?

.github/workflows/cypress-tests-runner.yml
```diff
@@ -286,24 +312,395 @@ jobs:
...
+      - name: Download hyperswitch server artifact
+        if: ${{ env.RUN_OPTIONAL_CONNECTORS_TESTS == 'true' }}
+        uses: actions/download-artifact@v4
+        with:
+          name: hyperswitch-server
+          path: target/debug
+
+      - name: Make server executable
+        if: ${{ env.RUN_OPTIONAL_CONNECTORS_TESTS == 'true' }}
+        run: chmod +x target/debug/router
```
pixincreate, line 465: now that i think about it, we can move this to `hyperswitch-server` job and make it an executable just before uploading it as an artifact.

.github/workflows/cypress-tests-runner.yml
```diff
@@ -286,24 +312,395 @@ jobs:
...
+      - name: Download Encrypted TOML from S3 and Decrypt
+        if: ${{ env.RUN_OPTIONAL_CONNECTORS_TESTS == 'true' }}
+        env:
+          AWS_ACCESS_KEY_ID: ${{ secrets.CONNECTOR_CREDS_AWS_ACCESS_KEY_ID }}
+          AWS_REGION: ${{ secrets.CONNECTOR_CREDS_AWS_REGION }}
+          AWS_SECRET_ACCESS_KEY: ${{ secrets.CONNECTOR_CREDS_AWS_SECRET_ACCESS_KEY }}
+          CONNECTOR_AUTH_PASSPHRASE: ${{ secrets.CONNECTOR_AUTH_PASSPHRASE }}
+          CONNECTOR_CREDS_S3_BUCKET_URI: ${{ secrets.CONNECTOR_CREDS_S3_BUCKET_URI}}
+          DESTINATION_FILE_NAME: "creds.json.gpg"
+          S3_SOURCE_FILE_NAME: "f2861b17-2965-4687-acfe-0cb6cd19e693.json.gpg"
```
pixincreate: cc: @likhinbopanna

.github/workflows/cypress-tests-runner.yml
```diff
@@ -286,24 +312,395 @@ jobs:
...
+
+      - name: Run Cypress tests
+        if: ${{ env.RUN_OPTIONAL_CONNECTORS_TESTS == 'true' }}
+        env:
+          CYPRESS_BASEURL: "http://localhost:8080"
+          ROUTER__SERVER__WORKERS: 4
+          PAYMENTS_CONNECTORS: ${{ env.OPTIONAL_PAYMENTS_CONNECTORS_BATCH_2 }}
+        shell: bash -leuo pipefail {0}
+        run: |
+          scripts/execute_cypress.sh --parallel 4
```
pixincreate: parallel 4?

scripts/execute_cypress.sh
```diff
@@ -1,189 +1,251 @@
...
-  print_color "yellow" "Cleaning up..."
-  if [[ -d "cypress-tests" ]]; then
-    cd -
+  # -----------------------------
+  # Final Summary
+  # -----------------------------
+  if [[ -s "$results_log" ]]; then
+
+    print_color green "\n========================================"
+    print_color green "🟢 SUCCESSFUL CONNECTORS"
```
pixincreate: do we really need these emojis and comments??

.github/workflows/cypress-tests-runner.yml
NISHANTH1221: Was testing things out

scripts/execute_cypress.sh
NISHANTH1221: This might help in scraping the logs in future. Removing emoji

.github/workflows/cypress-tests-runner.yml
NISHANTH1221: Optional Connectors wont be run at any PR level this will be done only when they are in merge queue

.github/workflows/cypress-tests-runner.yml
```diff
@@ -172,17 +187,27 @@ jobs:
         if: ${{ env.RUN_TESTS == 'true' }}
         uses: actions/checkout@v4
 
+      - name: Install Rust
+        if: ${{ env.RUN_TESTS == 'true' }}
+        uses: dtolnay/rust-toolchain@master
+        with:
+          toolchain: stable 2 weeks ago
+          components: llvm-tools-preview
```
SanchithHegde: Do we necessarily need this component?

.github/workflows/cypress-tests-runner.yml
```diff
@@ -286,24 +310,386 @@ jobs:
...
         if: env.RUN_TESTS == 'true' && failure()
         uses: actions/upload-artifact@v4
         with:
-          name: cypress-test-results
+          name: cypress-test-results-mandatory
+          path: |
+            cypress-tests/cypress/reports/
+          retention-days: 1
+
+  runner-batch-1:
```
SanchithHegde: Can we consider using a matrix strategy instead of having to define each batch individually (and repeating a lot of code)?

scripts/execute_cypress.sh
```diff
@@ -1,189 +1,251 @@
...
-      [[ $service == "payment_method_list" ]] && service="payment-method-list"
+  return 0
+}
 
-      echo "Running ${service} tests without connectors..."
-      export REPORT_NAME="${service}_report"
+export -f execute_test
+export -f command_exists
+export -f print_color
+export results_log
```
SanchithHegde, line 147: Do we necessarily need to export all of these?

.github/workflows/cypress-tests-runner.yml
NISHANTH1221: Sure

.github/workflows/cypress-tests-runner.yml
NISHANTH1221: Not sure why but when the components were removed the tests are failing in migration phase itself
Example run: https://github.com/juspay/hyperswitch/actions/runs/20727841487/job/59508411294?pr=10662.
Will look into this more..

.github/workflows/cypress-tests-runner.yml
NISHANTH1221: Tries it but throws an error

scripts/execute_cypress.sh
NISHANTH1221: I have read that while using parallel we need to explicitly export the function so that will be available for the child processes

.github/workflows/cypress-tests-runner.yml
NISHANTH1221: I also got to know while uploading the file the file metadata is not attached to it. So we have to anyway change the permissions while execution

.github/workflows/cypress-tests-runner.yml
NISHANTH1221: Not sure why but when I removed the components The migration is on db is failing

.github/workflows/cypress-tests-runner.yml
SanchithHegde: You might only need rustfmt for migrations, you mostly shouldn't need llvm-tools-preview.

.github/workflows/cypress-tests-runner.yml
```diff
@@ -170,15 +188,28 @@ jobs:
...
+          tool: sccache
+          checksum: true
 
       - name: Download hyperswitch server artifact
         if: ${{ env.RUN_TESTS == 'true' }}
         uses: actions/download-artifact@v4
         with:
           name: hyperswitch-server
           path: target/debug
-
```
SanchithHegde, line 181: Nit: Can you revert this change, let's keep the newlines between steps.

.github/workflows/cypress-tests-runner.yml
```diff
@@ -234,7 +265,6 @@ jobs:
         uses: actions/setup-node@v4
         with:
           node-version: "24"
-
```
SanchithHegde, line 237: Same here.

.github/workflows/cypress-tests-runner.yml
```diff
@@ -286,24 +316,287 @@ jobs:
...
+          } >> "$GITHUB_OUTPUT"
+
+
+  runner-optional-connectors:
+    name: Run Optional Connectors Cypress tests
+    needs: [ build-hyperswitch, prepare-matrix ]
+    runs-on: hyperswitch-runners
+    strategy:
+      fail-fast: false
+      matrix: ${{ fromJson(needs.prepare-matrix.outputs.matrix) }}
```
SanchithHegde, line 413: Would this work correctly for PRs from forks: as I understand, this output value won't be set in case the PR was opened from a fork...

Since we don't plan to run this check on PRs in merge queue, this check won't run if a PR is opened from a fork at all: it would neither run directly, nor when it's added to merge queue. Is that the intended behavior?

If this in indeed the intended behavior, then we can consider adding a job-level `if` condition for this job and the prepare-matrix job, to conditionally skip them entirely.

.github/workflows/cypress-tests-runner.yml
SanchithHegde: Can probably remove llvm-tools-preview, keeping rustfmt for migrations.

.github/workflows/cypress-tests-runner.yml
```diff
@@ -286,24 +316,287 @@ jobs:
...
+      - name: Upload Cypress test results
+        if: env.RUN_OPTIONAL_CONNECTORS_TESTS == 'true' && failure()
+        uses: actions/upload-artifact@v4
+        with:
+          name:  ${{ matrix.batch.report_artifact_name }}
           path: |
             cypress-tests/cypress/reports/
           retention-days: 1
 
   runner_alpha:
```
SanchithHegde, line 577: 1. Do we need alpha connectors running on these runners, or can they use the public ones?
2. Do we need them running on merge queue?

NISHANTH1221: 1. We can use the public ones but since the build is happening in hyperswitch-runners which is of different architecture and if we move the alpha connectors to public one we need to build server again in that architecture..
2. We will be running them in merge queue

.github/workflows/cypress-tests-runner.yml
NISHANTH1221: Added one more runner to decide on what to do. Based on its results we are skipping the tests at PR level

.github/workflows/cypress-tests-runner.yml
```diff
@@ -286,24 +318,298 @@ jobs:
...
+    outputs:
+      run_optional: ${{ steps.decide.outputs.run_optional }}
+    steps:
+      - id: decide
+        run: |
+          echo "run_optional=${RUN_OPTIONAL_CONNECTORS_TESTS}" >> "$GITHUB_OUTPUT"
+
+  prepare-matrix:
+    name: Prepare Optional Connectors matrix to run cypress tests
+    if: ${{ needs.decide.outputs.run_optional == 'true' }}
```
SanchithHegde: You don't need the additional `decide-whether-to-run-optional` job, just have this as:

```yaml
if: ${{ env.RUN_OPTIONAL_CONNECTORS_TESTS == 'true' }}
```

Same for the `runner-optional-connectors` job.

Or if that does not work, directly add the `event == merge group` condition here.

.github/workflows/cypress-tests-runner.yml
```diff
@@ -286,24 +318,298 @@ jobs:
...
+    runs-on: hyperswitch-runners
+    outputs:
+      matrix: ${{ steps.set-matrix.outputs.matrix }}
+    steps:
+      - name: Skipping matrix setup at PR level
+        if: ${{ env.RUN_OPTIONAL_CONNECTORS_TESTS == 'false' }}
```
SanchithHegde: Because you have the job-level `if` condition, you don't need the `if` specified on each step. Also, this specific step can be removed.

.github/workflows/cypress-tests-runner.yml
```diff
@@ -286,24 +318,298 @@ jobs:
...
+          --health-timeout 5s
+          --health-retries 5
+        ports:
+          - 5432:5432
+
+    steps:
+      - name: Skip tests on PR level
+        if: ${{ (env.RUN_OPTIONAL_CONNECTORS_TESTS == 'false')  }}
+        shell: bash
+        run: echo 'Skipping tests at PR level for optional connectors'
```
SanchithHegde: This specific step can be removed, and you can remove the `if` specified for each of the following steps because of the job-level `if` condition.

.github/workflows/cypress-tests-runner.yml
```diff
@@ -286,24 +318,298 @@ jobs:
...
+          npm ci --prefix ./cypress-tests
+
+      - name: Run database migrations
+        if: ${{ env.RUN_OPTIONAL_CONNECTORS_TESTS == 'true' }}
+        shell: bash
+        env:
+          DATABASE_URL: postgres://db_user:db_pass@localhost:5432/hyperswitch_db
+        run: just migrate run --locked-schema
+
+      - name: Insert card info into the database
```
SanchithHegde, line 523: If #10853 gets merged before your PR, you may have to insert GSM data as well into the database.

scripts/execute_cypress.sh
```diff
@@ -1,189 +1,251 @@
...
   local connector="$1"
   local service="$2"
-  local tmp_file="$3"
+  local failure_log="$3"
+  local job_slot="${4:-1}"
+
+  local start_ts
+  start_ts=$(date +%s)
+  local start_fmt
+  start_fmt=$(date '+%H:%M:%S')
```
SanchithHegde: Nit: Can't we combine these lines?


```suggestion
  local start_ts=$(date +%s)
  local start_fmt=$(date '+%H:%M:%S')
```

scripts/execute_cypress.sh
```diff
@@ -1,189 +1,251 @@
...
   fi
-}
-export -f execute_test
 
-# Function to run tests
-function run_tests() {
-  local jobs="${1:-1}"
-  tmp_file=$(mktemp)
+  local end_ts
+  end_ts=$(date +%s)
```
SanchithHegde: Same here:

```suggestion
  local end_ts=$(date +%s)
```

scripts/execute_cypress.sh
```diff
@@ -1,189 +1,251 @@
...
-}
+  for service in "${active_services[@]}"; do
+    declare -n connectors="$service"
 
-# Function to check and install dependencies
-function check_dependencies() {
-  # parallel and npm are mandatory dependencies. exit the script if not found.
-  local dependencies=("parallel" "npm")
+    if [[ ${#connectors[@]} -gt 0 ]]; then
+      print_color yellow "🚀 Starting Parallel Execution for '$service' (Jobs: $jobs)"
```
SanchithHegde: We can maybe get rid of the emoji considering that none of the other log lines have it.

.github/workflows/cypress-tests-runner.yml
NISHANTH1221: Yeah forgot to remove

.github/workflows/cypress-tests-runner.yml
NISHANTH1221: Already used the below before it didnt work.
```if: ${{ env.RUN_OPTIONAL_CONNECTORS_TESTS == 'true' }}```
updated with this block `event == merge group`

File: .github/workflows/cypress-tests-runner.yml [hunk 29]
```diff
@@ -291,24 +322,265 @@ jobs:
...
+        id: set-matrix
+        shell: bash
+        run: |
+          {
+            echo "matrix<<EOF"
+            echo '{
+              "batch": [
+                {
+                  "name": "batch-1",
+                  "connectors": "'"${OPTIONAL_PAYMENTS_CONNECTORS_BATCH_1}"'",
```
SanchithHegde, line 355: What's with the single quotes over here?

.github/workflows/cypress-tests-runner.yml
SanchithHegde, line 1: @pixincreate Should we consider having a [reusable workflow](https://docs.github.com/en/actions/how-tos/reuse-automations/reuse-workflows) for running the Cypress tests for required, optional, and alpha connectors? We can also consider moving the v2 checks to use it, if we split the build step to a separate job.

Of course, this would depend on the amount of similarity among these jobs. And we can take it up in a separate PR, not a problem.

SanchithHegde (commented):

.github/workflows/cypress-tests-runner.yml
NISHANTH1221: So when I tried this `"${OPTIONAL_PAYMENTS_CONNECTORS_BATCH_1}"` and the ran the cypress it started to consider the three connectors as a single on and it was not able to assign the env value to the connectors at the prepare matrix phase. So I moved to the current version where it is able to assign the connectors to connectors variable

.github/workflows/cypress-tests-runner.yml
pixincreate: Yeah, almost everything in Cypress is similar.

Building Hyperswitch is kept as a separate job. It uploads the binary as an artifact which is pulled by other jobs - mandatory, mock server, and now optional (2 batches). The only thing that differs is the connector names (except for mock server since we need to build / start the server).

So yeah, we can do that.

SanchithHegde: We can accept inputs for things like these which involve conditional behavior. We can have an input for say whether to start a mock server as a boolean, one input for the list of connectors (as a comma separated string), etc.

SanchithHegde (approved):

.github/workflows/cypress-tests-runner.yml
```diff
@@ -291,24 +322,265 @@ jobs:
...
         with:
-          name: cypress-test-results
+          name: cypress-test-results-mandatory
+          path: |
+            cypress-tests/cypress/reports/
+          retention-days: 1
+
+  prepare-matrix:
+    name: Prepare Optional Connectors matrix to run cypress tests
+    if: ${{ github.event_name == 'merge_group' }}
```
pixincreate, line 340: this will only run in merge group, so, if tests fail there and it is not because of cypress but rather genuine bugs, how are we planning to address it given that this is not a required check?

NISHANTH1221: Currently the cypress test are not stable for all the connectors. When they become stable like the stripe and cybersource we will be making mandatory

PR #10687, opened 2025-12-16
feat(permissions): Split Account parent group and backfill permissions

### LINKED ISSUE #10880: chore: Split Account parent group and backfill permissions

kanikabansal08:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [x] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

This PR refactors the Account-related permission model and introduces a dedicated dashboard bootstrapping API to eliminate implicit Account access while preserving current behavior for existing roles.

<b>Scope of changes:</b>

1. <b>Split the Account parent group</b>  
Previously, the Account parent group bundled multiple resources:  
- Account  
- WebhookEvents  
- ApiKeys  

Now, resources are split into distinct parent groups:  
- <b>Account:</b> Account  
- <b>Webhooks:</b> WebhookEvents  
- <b>ApiKeys:</b> ApiKeys  

Only the Account parent group now contains the Account resource.

2. <b>Introduce dashboardMerchantDetails API for dashboard bootstrapping</b>  
- New lightweight API: <code>dashboardMerchantDetails</code>  
- Returns: <code>recon_status</code>, <code>product_type</code>  
- Auth: <code>DashboardNoPermissionAuth</code> (no business/account permissions required)  
- Usage:  
  - Dashboard boot flow uses <code>dashboardMerchantDetails</code> instead of <code>merchantAccountAPI</code>  
  - <code>merchantAccountAPI</code> remains unchanged and is now used only from the homepage, requiring explicit Account permissions

3. <b>Remove Account resource from non-Account parent groups</b>  
- Account resource is removed from all other parent groups where it was implicitly present  
- After this change, Account exists only under the Account parent group  
- Prevents accidental/implicit Account-level permission grants originating from other parent groups

4. <b>Add explicit Account access via “accessible” permissions for dependent features</b>  
Some flows require limited Account data before hitting feature-specific APIs. We add “accessible” permissions to encode this dependency explicitly:  
- <code>WebhooksView</code> → also grants <code>AccountView</code>  
- <code>WebhooksManage</code> → also grants <code>AccountManage</code>  
- <code>ApiKeysView</code> → also grants <code>AccountView</code>  
- <code>ApiKeysManage</code> → also grants <code>AccountManage</code>  

This mirrors the “accessible permissions” pattern used for Connectors / Operations, ensuring flows continue to work without global Account leakage.

5. <b>Permission backfilling for backward compatibility</b>  
To avoid breaking existing roles which relied on implicit Account access, we backfill permissions.  

<b>Predefined roles:</b>  
- If role had <code>AccountManage</code>: add <code>WebhooksManage</code> + <code>ApiKeysManage</code>  
- If role had <code>AccountView</code>: add <code>WebhooksView</code> + <code>ApiKeysView</code>  
- If role had any *Manage (excluding AccountManage): add <code>AccountManage</code>  
- If role had any *View (excluding AccountView): add <code>AccountView</code>  

<b>Existing custom roles:</b> Same rules applied:  
- <code>AccountManage</code> → <code>WebhooksManage</code> + <code>ApiKeysManage</code>  
- <code>AccountView</code> → <code>WebhooksView</code> + <code>ApiKeysView</code>  
- Any other *Manage → <code>AccountManage</code>  
- Any other *View → <code>AccountView</code>  

Ensures existing predefined and custom roles maintain their effective access.

6. <b>Behavior for future custom roles</b>  
- Account permissions are not added implicitly anymore  
- Example: A role with only <code>OperationsManage</code>  
  - Does not get Account access  
  - Does not see Payment Settings  

This is the intended forward behavior: Account access must be granted explicitly.

<b>Resulting behavior:</b>  
- No more implicit Account leakage  
- Account resource is no longer indirectly granted via Webhooks / ApiKeys / other parent groups  
- Payment settings visibility is strictly tied to explicit Account permissions or to specific accessible-permission chains

<b>FE usage changes:</b>  
- Dashboard boot: Uses <code>dashboardMerchantDetails</code> (no Account permission requirement)  
- Homepage / areas that truly need Account: Continue using <code>merchantAccountAPI</code> with explicit <code>AccountView</code> / <code>AccountManage</code> permissions

### Additional Changes

## Motivation and Context
<b>Problem:</b>  

The Account resource was implicitly available through multiple parent groups (e.g. Webhooks, ApiKeys), so users gained Account-level permissions indirectly.  

Frontend uses Account to show Payment Settings and Webhooks. Because of implicit access, users could see Payment Settings even when:  
- They were not explicitly granted Account permissions, and  
- The Account resource did not originate from the Account parent group.  

This resulted in permission leakage and inconsistent behavior across roles.

<b>Goal:</b>  
- Make Account access explicit and predictable  
- Preserve behavior for all existing (predefined and custom) roles to avoid surprise permission loss  
- Provide a dedicated, low-privilege dashboard bootstrapping API so FE can bootstrap without needing full Account permissions

<b>How this PR addresses it:</b>  
- Splits the parent groups  
- Removes accidental Account exposure  
- Adds a dedicated dashboard boot API  
- Encodes feature dependencies via accessible permissions  
- Applies a backfilling strategy for existing roles

## How did you test it?

1. Dashboard details API:
Request:
```zsh
curl --location 'http://localhost:8080/user/merchant_details \
--header 'authorization: Bearer ****' \
--header 'Cookie: login_token=***'
```
Response:
```zsh
{
    "recon_status": "not_requested",
    "product_type": "orchestration"
}
```

2. Checked the current permissions granted for a role having different combinations of manage and view. Ran queries and checked if the role is granted the same permission as before. 

## Checklist


crates/router/src/core/admin.rs
```diff
@@ -853,6 +853,32 @@ pub async fn get_merchant_account(
...
+        .await
+        .to_not_found_response(errors::ApiErrorResponse::MerchantAccountNotFound)?;
+
+    Ok(service_api::ApplicationResponse::Json(
+        api::MerchantAccountDetailsResponse {
+            recon_status: merchant_account.recon_status,
+            product_type: merchant_account.product_type,
+        },
+    ))
+}
```
ThisIsMani: Shouldn't this be in user module?

migrations/2026-01-07-082238_merchant_account_decoupling/up.sql
```diff
@@ -0,0 +1,35 @@
...
+  AND NOT EXISTS (
+    SELECT 1
+    FROM unnest(groups) AS g
+    WHERE g LIKE '%_manage'
+  )
+  AND EXISTS (
+    SELECT 1
+    FROM unnest(groups) AS g
+    WHERE g LIKE '%_view'
+  );
```
ThisIsMani: This can be added to all the roles which doesn't have `account_view`? I don't think we need extra checks because we will have atleast one of these in the role... But not a problem with having those checks as well.

migrations/2026-01-07-082238_merchant_account_decoupling/up.sql
ThisIsMani, line 1: Add comments on the migrations.

migrations/2026-01-07-082238_merchant_account_decoupling/up.sql
kanikabansal08: Yes, agreed.

crates/api_models/src/events.rs
```diff
@@ -103,6 +103,7 @@ impl_api_event_type!(
         MerchantId,
         CardsInfoRequest,
         MerchantAccountResponse,
+        MerchantAccountDetailsResponse,
```
ThisIsMani: This should be in `events/user.rs`.

crates/api_models/src/user.rs
```diff
@@ -470,3 +470,27 @@ pub struct ListProfilesForUserInOrgAndMerchantAccountResponse {
     pub profile_id: id_type::ProfileId,
     pub profile_name: String,
 }
+
+#[cfg(feature = "v1")]
+#[derive(Clone, Debug, ToSchema, serde::Serialize)]
+pub struct MerchantAccountDetailsResponse {
+    /// Used to indicate the status of the recon module for a merchant account
+    #[schema(value_type = ReconStatus, example = "not_requested")]
```
ThisIsMani: Schema is not required for user APIs. Can be removed.

crates/api_models/src/user.rs
```diff
@@ -470,3 +470,27 @@ pub struct ListProfilesForUserInOrgAndMerchantAccountResponse {
...
+    pub recon_status: common_enums::ReconStatus,
+
+    /// Product Type of this merchant account
+    #[schema(value_type = Option<MerchantProductType>, example = "Orchestration")]
+    pub product_type: Option<common_enums::MerchantProductType>,
+}
```
ThisIsMani: There is no difference between v1 and v2... I don't think we need 2 different definitions for this.

crates/api_models/src/user.rs
```diff
@@ -470,3 +470,27 @@ pub struct ListProfilesForUserInOrgAndMerchantAccountResponse {
...
+#[cfg(feature = "v1")]
+#[derive(Clone, Debug, ToSchema, serde::Serialize)]
+pub struct MerchantAccountDetailsResponse {
+    /// Used to indicate the status of the recon module for a merchant account
+    #[schema(value_type = ReconStatus, example = "not_requested")]
+    pub recon_status: common_enums::ReconStatus,
+
+    /// Product Type of this merchant account
+    #[schema(value_type = Option<MerchantProductType>, example = "Orchestration")]
+    pub product_type: Option<common_enums::MerchantProductType>,
```
ThisIsMani: How will you handle this option?

crates/router/src/routes/app.rs
```diff
@@ -2685,6 +2685,10 @@ impl User {
             web::resource("/create-merchant")
                 .route(web::post().to(user::user_merchant_account_create)),
         );
+        route = route.service(
+            web::resource("/accounts/{id}/details")
+                .route(web::get().to(user::retrieve_merchant_account_details)),
+        );
```
ThisIsMani, line 2691: Can we name this something like `/merchant_details`. And also, `merchant_id` should be taken from token. Not from the path.

crates/router/src/routes/user.rs
```diff
@@ -1063,3 +1063,45 @@ pub async fn clone_connector(
     ))
     .await
 }
+
+/// Get merchant account details (recon_status and product_type).
+#[cfg(feature = "v1")]
```
ThisIsMani: Are you sure you need 2 different impls?

crates/router/src/routes/user.rs
```diff
@@ -1063,3 +1063,45 @@ pub async fn clone_connector(
...
+    merchant_id: web::Path<common_utils::id_type::MerchantId>,
+) -> HttpResponse {
+    let flow = Flow::MerchantsAccountDetailsRetrieve;
+    let mid = merchant_id.into_inner();
+    Box::pin(api::server_wrap(
+        flow,
+        state,
+        &req,
+        mid,
+        |state, _user: (), req, _| user_core::get_merchant_account_details(state, req),
```
ThisIsMani: Take token here.

ThisIsMani (changes requested):

crates/api_models/src/user.rs
kanikabansal08: Throughout the codebase product type is kept as option type. It would default to `None` if missing and send null as response.

crates/router/src/routes/app.rs
```diff
@@ -2744,6 +2748,10 @@ impl User {
                 web::resource("/create_merchant")
                     .route(web::post().to(user::user_merchant_account_create)),
             )
+            .service(
+                web::resource("/merchant-details")
```
ThisIsMani:
```suggestion
                web::resource("/merchant_details")
```

crates/router/src/routes/app.rs
```diff
@@ -2685,6 +2685,10 @@ impl User {
             web::resource("/create-merchant")
                 .route(web::post().to(user::user_merchant_account_create)),
         );
+        route = route.service(
+            web::resource("/merchant-details")
```
ThisIsMani, line 2689:
```suggestion
            web::resource("/merchant_details")
```

ThisIsMani (commented):
Minor concerns. Other than that LGTM.

crates/api_models/src/events.rs
```diff
@@ -43,7 +43,7 @@ use crate::{
     },
     payment_methods::*,
     payments::*,
-    user::{UserKeyTransferRequest, UserTransferKeyResponse},
+    user::{MerchantAccountDetailsResponse, UserKeyTransferRequest, UserTransferKeyResponse},
```
ThisIsMani: I think this import can be removed.

ThisIsMani (dismissed):

crates/router/src/services/authorization/permission_groups.rs
```diff
@@ -117,6 +131,8 @@ impl ParentGroupExt for ParentGroup {
             Self::Analytics => ANALYTICS.to_vec(),
             Self::Users => USERS.to_vec(),
             Self::Account => ACCOUNT.to_vec(),
+            Self::Webhook => WEBHOOK.to_vec(),
+            Self::ApiKeys => APIKEYS.to_vec(),
```
tsdk02:
```suggestion
            Self::ApiKeys => API_KEYS.to_vec(),
```

tsdk02 (commented):
Minor nit, other than that looks good to me!

tsdk02 (dismissed):

ThisIsMani (dismissed):

migrations/2026-01-07-082238_merchant_account_decoupling/down.sql
```diff
@@ -0,0 +1,48 @@
...
+-- should retain only account_manage.
+UPDATE roles
+SET groups = array_remove(
+               array_remove(groups, 'webhooks_manage'),
+               'api_keys_manage'
+             )
+WHERE
+  'account_manage' = ANY (groups)
+  AND 'webhooks_manage' = ANY (groups)
+  AND 'api_keys_manage' = ANY (groups);
```
hrithikesh026, line 36: Is there a use case of undoing these backfills?

migrations/2026-01-07-082238_merchant_account_decoupling/up.sql
```diff
@@ -0,0 +1,31 @@
...
+  AND EXISTS (
+    SELECT 1
+    FROM unnest(groups) AS g
+    WHERE g LIKE '%_manage'
+  );
+
+-- Roles without account_view must explicitly have account_view
+UPDATE roles
+SET groups = array_append(groups, 'account_view')
+WHERE 'account_view' <> ALL (groups);
```
hrithikesh026, line 31: are these run in a local setup and verified?

Also are these backwards compatible?

crates/router/src/routes/app.rs
```diff
@@ -2685,6 +2685,10 @@ impl User {
             web::resource("/create-merchant")
                 .route(web::post().to(user::user_merchant_account_create)),
         );
+        route = route.service(
+            web::resource("/merchant_details")
```
hrithikesh026: follow v2 conventions.
```suggestion
            web::resource("/merchant-details")
```

migrations/2026-01-07-082238_merchant_account_decoupling/down.sql
kanikabansal08: Yes, if we ever need to revert to the previous state of the code, these statements would need to be executed, since some new permissions have been added and these updates ensure roles are correctly reverted

migrations/2026-01-07-082238_merchant_account_decoupling/up.sql
kanikabansal08: yes these are verified and backwards compatible.

PR #10600, opened 2025-12-10
feat(gsm): add standardised error fields to GSM models and database schema

ayush22667:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

This PR adds new columns to the GSM (Gateway Status Mapping) table and updates the create/update APIs to include these new fields for better error categorization and user guidance.

### Additional Changes

- [x] This PR modifies the API contract
- [x] This PR modifies the database schema
- [ ] This PR modifies application configuration/environment variables

## Motivation and Context

closes #10599 

## How did you test it?

Create Request:
```
curl --location 'http://localhost:8080/gsm' \
--header 'Content-Type: application/json' \
--header 'api-key: test_admin' \
--data '{
  "connector": "stripe",
  "flow": "Payment",
  "sub_flow": "Authoriz",
  "code": "card_declined",
  "message": "Your card was declined.",
  "status": "Failure",
  "router_error": "CARD_DECLINED",
  "decision": "retry",
  "step_up_possible": true,
  "unified_code": "IR_01",
  "unified_message": "Card declined by issuer",
  "error_category": "processor_downtime",
  "clear_pan_possible": false,
  "feature": "retry",
  "feature_data": {
    "retry": {
      "step_up_possible": true,
      "clear_pan_possible": false,
      "alternate_network_possible": true,
      "decision": "retry"
    }
  },
  "standardised_code": "PM_INSUFFICIENT_FUNDS",
  "description": "The card was declined due to insufficient funds in the account.",
  "user_guidance_message": "Your payment was declined due to insufficient funds. Please use a different card or add funds to your account."
}'
```
Response:
```
{
    "connector": "stripe",
    "flow": "Payment",
    "sub_flow": "Authoriz",
    "code": "card_declined",
    "message": "Your card was declined.",
    "status": "Failure",
    "router_error": "CARD_DECLINED",
    "decision": "retry",
    "step_up_possible": true,
    "unified_code": "IR_01",
    "unified_message": "Card declined by issuer",
    "error_category": "processor_downtime",
    "clear_pan_possible": false,
    "feature": "retry",
    "feature_data": {
        "retry": {
            "step_up_possible": true,
            "clear_pan_possible": false,
            "alternate_network_possible": true,
            "decision": "retry"
        }
    },
    "standardised_code": "PM_INSUFFICIENT_FUNDS",
[...]

PR #10729, opened 2025-12-18
Revert "feat(gsm): add standardised error fields to GSM models and database schema"

ayush22667:
Reverts juspay/hyperswitch#10600


ayush22667:
This got solved by increasing the stack size.

PR #10820, opened 2026-01-05
ci(cypress): Implementation of the trigger mechanism for running of cypress-tests from commit-based to label-based on PR level

AdityaKumaar21:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [x] CI/CD

## Description

This PR changes the Cypress test trigger mechanism from __commit-based__ to __label-based__:

- Cypress tests now only run when the `S-test-ready` label is added to a PR
- All jobs (`formatter`, `build-hyperswitch`, `runner`, `runner_alpha`, `runner_v2`) are gated behind the `S-test-ready` label check
- If tests fail, the `S-test-ready` label is automatically removed, signaling that the PR needs fixes before re-testing

### Changes Made

1. __Added `labeled` event type__ to the `pull_request` trigger

2. __Added label condition__ to all jobs:

   - `formatter` - Only formats Cypress tests when label is present
   - `build-hyperswitch` - Only builds when tests will run
   - `runner` - Main Cypress test runner
   - `runner_alpha` - Alpha connector tests
   - `runner_v2` - V2 Cypress tests with coverage

3. __Added automatic label removal__ on failure for `runner` and `runner_v2` jobs

### How It Works

When a pull request is opened or updated, Cypress tests are skipped by default.
When the `S-test-ready `label is added, all Cypress tests run.
If the tests pass, the label stays and the pull request can be merged.
If the tests fail, the label is automatically removed.
When the pull request enters the merge queue, Cypress tests always run, regardless of labels.

### Additional Changes

## Motivation and Context

Previously, Cypress tests ran on every commit pushed to a pull request. This caused:

- __Long CI wait times__ for developers on every push.
- __Unnecessary resource consumption__ on non-important commits.

## How did you test it?

1) Cypress tests being skipped when `S-test-ready` label not present

Screenshot 2026-01-08 at 14 08 20
2) Cypress tests running when `S-test-ready` label is present

  - First add the label `S-test-ready`

Screenshot 2026-01-06 at 12 29 41

  - It will show:-

Screenshot 2026-01-06 at 01 40 37

  - Cypress-tests will be running after this

Screenshot 2026-01-05 at 14 58 09

3) Label being removed when cypress-tests fail

Screenshot 2026-01-06 at 01 40 25

## Checklist


.github/workflows/cypress-tests-runner.yml
```diff
@@ -24,7 +25,9 @@ env:
 jobs:
   formatter:
     name: Run formatter on Cypress tests and address lints
-    if: ${{ github.event_name == 'pull_request' }}
+    if: |
+      github.event_name == 'pull_request' && 
+      contains(github.event.pull_request.labels.*.name, 'S-test-ready')
```
SanchithHegde: You can consider moving this to an env var, if that helps.

.github/workflows/cypress-tests-runner.yml
```diff
@@ -299,9 +310,19 @@ jobs:
             cypress-tests/cypress/reports/
           retention-days: 1
 
+      - name: Remove S-test-ready label on failure
+        if: failure() && github.event_name == 'pull_request'
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          gh pr edit ${{ github.event.pull_request.number }} --remove-label "S-test-ready" --repo ${{ github.repository }}
```
SanchithHegde: We might have to handle this situation for PRs raised from forks. If I remember correctly, if a PR is raised from a fork, you can't use the `github.token` to modify labels on the PR, because the checks run in the context of the fork, not the upstream repo.

How should we handle this step for forks? Should we not run this step for forks (check the `RUN_TESTS` env var), or do you have any other suggestions?

AdityaKumaar21: Since PRs from forks, `github.token` is read-only, so label modification isn’t allowed.
We already detect fork PRs using `RUN_TESTS`, so I am thinking to gate the label-removal step to only run when `env.RUN_TESTS == 'true' && github.event_name != 'merge_group' `.

.github/workflows/cypress-tests-runner.yml
AdityaKumaar21: i am not able move it to an env var because GitHub Actions doesn't allow using the env context in job-level if conditions. The env context is only available within steps, not at the job level.

.github/workflows/cypress-tests-runner.yml
```diff
@@ -5,6 +5,7 @@ on:
     types:
       - checks_requested
   pull_request:
+    types: [opened, synchronize, reopened, labeled]
```
SanchithHegde: Doesn't the `labeled` trigger have the unfortunate quirk where all jobs in the workflow restart if any label is added or removed?

For example, you have a PR with labels A, B and your `S-test-ready` (let's say S for now). So the Cypress tests would be in progress. About 10 minutes in, you realize you forgot to add another label C, or someone else adds the label C to your PR. From what I remember, this should cause all the jobs to restart, just because the PR labels were updated, even though the S label was already present.

This may become a concern for checks like this one, which take significant time, as adding / removing any label can cause checks to be rerun unnecessarily.

.github/workflows/cypress-tests-runner.yml
```diff
@@ -574,9 +584,19 @@ jobs:
...
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          gh pr edit ${{ github.event.pull_request.number }} --remove-label "S-test-ready" --repo ${{ github.repository }}
+
   runner_alpha:
     name: Run optional Cypress tests for alpha connectors
     needs: build-hyperswitch
+    if: |
+      github.event_name == 'merge_group' ||
+      (github.event_name == 'pull_request' && contains(github.event.pull_request.labels.*.name, 'S-test-ready'))
```
SanchithHegde: @pixincreate Do we need to run these on merge queue?

pixincreate: no. it is alpha connectors anyway.

.github/workflows/cypress-tests-runner.yml
```diff
@@ -35,7 +35,9 @@ env:
 jobs:
   formatter:
     name: Run formatter on Cypress tests and address lints
-    if: ${{ github.event_name == 'pull_request' }}
+    if: |
+      github.event_name == 'pull_request' && 
+      contains(github.event.pull_request.labels.*.name, 'S-test-ready')
```
pixincreate: so the formatter and linter will only run when the label is added going forward? why? i feel, this change is not needed here.

pixincreate:
so, moving forward, developers will have to manually add the label to trigger the tests?

> Long CI wait times for developers on every push.

also, i do not think this is true. from what i'm aware, on every commit push, the previous run will get cancelled.

.github/workflows/cypress-tests-runner.yml
AdityaKumaar21: sure, i will remove it.

AdityaKumaar21:
> so, moving forward, developers will have to manually add the label to trigger the tests?
> 
> > Long CI wait times for developers on every push.
> 
> also, i do not think this is true. from what i'm aware, on every commit push, the previous run will get cancelled.

yes, from now onwards developers will have to add `S-ready-test` label before their commit or re-run the cypress-tests workflow to run cypress tests, so that we can reduce the over-consumption of runners on every commits.

.github/workflows/cypress-tests-runner.yml
```diff
@@ -1001,3 +1022,10 @@ jobs:
           token: ${{ secrets.CODECOV_TOKEN }}
           files: ${{ env.CODECOV_FILE }}
           disable_search: true
+
+      - name: Remove S-test-ready label on failure
+        if: failure() && env.RUN_TESTS == 'true' && github.event_name != 'merge_group'
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          gh pr edit ${{ github.event.pull_request.number }} --remove-label "S-test-ready" --repo ${{ github.repository }}
```
pixincreate: from what i'm aware, v2 tests are not mandatory. you'll be removing the label when that fail. intended?

AdityaKumaar21: I have fixed this, now we will only remove the label when we get a failed mandatory case (v1).
PS:- removed the step of removal of label on failure in `runner_v2` and `alpha_connectors`

.github/workflows/cypress-tests-runner.yml
SanchithHegde, line 423: @pixincreate @NISHANTH1221 What's the expected behavior for these, should these be running on PR or merge queue?

NISHANTH1221: They are supposed to run at Merge level only. In this pr #10917 I will be pausing the optional connector tests at merge level too currently.

PR #10825, opened 2026-01-05
feat(router): add an endpoint to list configured connector webhooks

AkshayaFoiger:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

Add an endpoint to list configured connector webhooks

### Additional Changes

## How did you test it?

Register a Standard webhook with Adyen

```
curl --location 'http://localhost:8080/account/postman_merchant_GHAction_ef2f9327-56ed-4e2f-8746-5a4d509fd1cc/connectors/webhooks/mca_knk2EbiG6GNcbvr5KPWX' \
--header 'api-key: dev_loZWT8fmby2JfUMjrpcabsNpTRwfPlONQrxVAhwlMvqDQv9XlzFRJJ0Jcq8tLcIu' \
--data '{
    "event_type": "all_events"
}'
```

Response 

```
{
    "event_type": "standard",
    "connector_webhook_id": "WBHK4295S223229S5NM4D9TDGF2WZJ",
    "webhook_registration_status": "Success",
    "error_code": null,
    "error_message": null
}
```

List webhooks registered for a merchant connector

```
curl --location 'http://localhost:8080/account/postman_merchant_GHAction_8bc459fc-f072-44b5-bf00-eca1554f3208/webhooks/mca_DZMdjJJM5LsDkEblLe93' \
--header 'api-key: dev_e26o6AOPYilYJFrW0dJTW5tuC7HrZUqvKQprgqgRNickmoiWTnHtPBRqkdqt9e8g'
``` 

Response

```
{
    "connector": "adyen",
    "webhooks": [
        {
            "event_type": "all_events",
            "connector_webhook_id": "WBHK42CNZ22322B25NX4WRJ52J6WDK"
        }
    ]
}
```

If the connector does not support registering webhook

```
    "error": {
        "type": "invalid_request",
        "message": "Webhook Registration flow not supported",
        "code": "IR_20",
        "connector": "stripe"
    }
}
```

## Checklist

### DIFF
--- a/.github/workflows/cypress-tests-runner.yml
+++ b/.github/workflows/cypress-tests-runner.yml
@@ -16,15 +16,14 @@ concurrency:
 env:
   CARGO_INCREMENTAL: 0
   CARGO_NET_RETRY: 10
-  PAYMENTS_CONNECTORS: "stripe cybersource"
-  OPTIONAL_PAYMENTS_CONNECTORS_BATCH_1: "cryptopay finix nexixpay"
-  OPTIONAL_PAYMENTS_CONNECTORS_BATCH_2: "adyen checkout zift"
-  OPTIONAL_PAYMENTS_CONNECTORS_BATCH_3: "airwallex finix noon"
-  OPTIONAL_PAYMENTS_CONNECTORS_BATCH_4: "nmi fiuu gigadat"
-  OPTIONAL_PAYMENTS_CONNECTORS_BATCH_5: "worldpayxml worldpay loonio"
-  OPTIONAL_PAYMENTS_CONNECTORS_BATCH_6: "authorizedotnet cashtocode volt"
-  OPTIONAL_PAYMENTS_CONNECTORS_BATCH_7: "paypal novalnet redsys"
-  OPTIONAL_PAYMENTS_CONNECTORS_BATCH_8: "paypal nuvei"
+  MANDATORY_PAYMENTS_CONNECTORS_BATCH_1: "cybersource stripe gigadat"
+  MANDATORY_PAYMENTS_CONNECTORS_BATCH_2: "cryptopay finix nexixpay"
+  MANDATORY_PAYMENTS_CONNECTORS_BATCH_3: "adyen checkout zift"
+  MANDATORY_PAYMENTS_CONNECTORS_BATCH_4: "paypal noon fiuu"
+  MANDATORY_PAYMENTS_CONNECTORS_BATCH_5: "authorizedotnet cashtocode loonio"
+  OPTIONAL_PAYMENTS_CONNECTORS_BATCH_1: "worldpayxml worldpay volt"
+  OPTIONAL_PAYMENTS_CONNECTORS_BATCH_2: "airwallex nmi redsys"
+  OPTIONAL_PAYMENTS_CONNECTORS_BATCH_3: "redsys nuvei"
   ALPHA_PAYMENTS_CONNECTORS: "celero silverflow"
   RUST_BACKTRACE: short
   RUSTUP_MAX_RETRIES: 10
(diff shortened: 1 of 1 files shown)


crates/diesel_models/src/merchant_connector_account.rs
```diff
@@ -106,6 +106,7 @@ pub struct MerchantConnectorAccount {
     pub version: common_enums::ApiVersion,
     pub id: id_type::MerchantConnectorAccountId,
     pub feature_metadata: Option<MerchantConnectorAccountFeatureMetadata>,
+    #[serde(default)]
```
hrithikesh026, line 109: why is this needed?

crates/router/src/core/merchant_connector_webhook_management/transformers.rs
```diff
@@ -212,3 +212,31 @@ pub fn construct_connector_webhook_registration_response(
         error_message: register_webhook_response.error_message.clone(),
     })
 }
+
+#[cfg(feature = "v1")]
+pub fn get_connector_webhook_list_response(
```
hrithikesh026, line 216: webhook details are anyway returned in the MCA retrieve call right? Why do we need a separate endpoint for it?

crates/router/src/core/merchant_connector_webhook_management.rs
```diff
@@ -134,3 +135,46 @@ pub async fn register_connector_webhook(
...
+            &key_store,
+        )
+        .await
+        .to_not_found_response(errors::ApiErrorResponse::MerchantConnectorAccountNotFound {
+            id: merchant_connector_id.get_string_repr().to_string(),
+        })?;
+
+    let connector_webook_data =
+        configure_connector_webhook_flow::get_connector_webhook_list_response(
+            &mca.connector_webhook_registration_details,
```
hrithikesh026, line 169: we will need a domain model for mca.connector_webhook_registration_details. 
Having it as serde value will not scale.

crates/diesel_models/src/merchant_connector_account.rs
AkshayaFoiger: The drainer reads records from a Kafka/Redis stream where data is serialized as JSON (via serde). When connector_webhook_registration_details is a newly added column, existing messages already in the stream were serialized before this field existed — so they won't contain this key in their JSON.

Without #[serde(default)], deserializing those older messages would fail

crates/router/src/core/merchant_connector_webhook_management/transformers.rs
AkshayaFoiger: No, webhook details are returned only through this API call. We follow this approach because the frontend would need to refresh the webhook list multiple times for the same MCA.

crates/router/src/core/merchant_connector_webhook_management.rs
AkshayaFoiger: We kept `connector_webhook_registration_details` in MerchantConnectorAccountUpdate as a value because the implementation From<MerchantConnectorAccountUpdate> for MerchantConnectorAccountUpdateInternal is infallible. The From trait does not allow returning errors. Changing this to TryFrom would require refactoring all existing call sites that currently rely on .into() for this conversion. Therefore, it is handled in the same way as the other fields.

hrithikesh026 (approved):

crates/router/src/core/merchant_connector_webhook_management/transformers.rs
```diff
@@ -212,3 +212,31 @@ pub fn construct_connector_webhook_registration_response(
...
+                api_models::merchant_connector_webhook_management::ConnectorWebhookResponse {
+                    event_type: webhook_data.event_type,
+                    connector_webhook_id: connector_webhook_id.clone(),
+                },
+            );
+        }
+    }
+
+    Ok(webhooks)
+}
```
hrithikesh026, line 240: Convert the serde json value to a strict type instead of this custom transform

hrithikesh026 (changes requested):

deepanshu-iiitu (approved):
LGTM

PR #4071, opened 2024-03-13
ci(postman): fix bluesnap and checkout collections for saved card flow

pixincreate:
## Type of Change

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [x] CI/CD

## Description

- NMI: Removed `Copy` from `.meta.json` to fix `Scenario 10` an `11` from running before `Scenario 1`
- Bluesnap: Make `customer_id` unique to avoid payment method from being saved

Closes #4070

> [!NOTE]
> `.json` files are auto-generated

### Additional Changes

- [ ] This PR modifies the API contract
- [ ] This PR modifies the database schema
- [ ] This PR modifies application configuration/environment variables

## How did you test it?

- NMI (mandates are failing. @Aprabhat19 is looking into it)
  image
  image

- Bluesnap
  image
- Checkout
  image

Wrote a script:

image

NMI failed as mentioned above:

image

## Checklist

- [ ] I formatted the code `cargo +nightly fmt --all`
- [ ] I addressed lints thrown by `cargo clippy`
- [x] I reviewed the submitted code
- [ ] I added unit tests for my changes where possible
- [ ] I added a [CHANGELOG](/CHANGELOG.md) entry if applicable

PR #10859, opened 2026-01-07
fix(users): Check key manager runtime feature flag when calling transfer key

ThisIsMani:
## Type of Change

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

We use `transfer_key_to_key_manager` to transfer key to the encryption service. When that function was introduced in the user module, there was only compile time feature flag. Afterwards a runtime feature flag is also introduced which when disabled, the expectation is the encryption service is not supposed to be used.

When this runtime flag is introduced, the user code was not updated to check if the runtime feature flag is enabled before calling the transfer function. Because of this, the code is trying to transfer the key to encryption service even when the runtime feature flag is set to false.

Ideally, this check should've been in transfer function itself since that is how all the other functions (`crypto_operation`) functions are written.

This PR adds check in `transfer_key_to_key_manager` function which checks if the runtime feature flag for key manager is enabled or not.

This was being done manually before, and was not added in user part of code, which caused some problems, so, I moved the check to the transfer function itself and callers doesn't have to check this manually from now on.

### Additional Changes

## Motivation and Context

Closes [#4071](https://github.com/juspay/hyperswitch-control-center/issues/4071)

## How did you test it?

Ran both Hyperswitch and Encryption Service locally. Turned on the `encryption_service` compile time feature flag in Hyperswitch.

Turned on the runtime feature flag, created a user, which creates merchant and setup TOTP, both merchant and user related operations are done in this flow along with encryption and decryption. Everything worked in this case.

Turned off the runtime feature flag and logged in with the old user, which also worked. - This was the main problem before.

Created a new user in this state (runtime flag turned off).

Turned on the runtime feature flag and logged in again with both users. Second user has fell back to local decryption since the key is not present in encryption service, everything worked as expected.

For QA: This is an internal change and doesn't affect our SaaS deployments.

## Checklist


ThisIsMani:
To check if the dependencies are being enabled correctly or not, I've disabled `encryption_service` feature flag in router and enabled the runtime feature flag.

I've added a piece of code in the `/health/ready` api, which logs the output of the `is_encryption_service_enabled` function.

With the old feature flag dependencies, with the above setup, the log outputs `true` even though the compile time feature flag in router is disabled.

With the new ones, the log shows false, which is expected.

ThisIsMani:
These dependencies are fragile though. There is a very high chance that somebody would later use wrong dependency graph, which will revert back to the old behaviour.

EDIT: Should not be a problem because in all the images we give out, `encryption_service` feature flag is enabled.

crates/common_utils/src/keymanager.rs
```diff
@@ -235,19 +235,37 @@ impl ConvertRaw for TransientBatchDecryptDataRequest {
...
     state: &KeyManagerState,
     request_body: EncryptionCreateRequest,
-) -> errors::CustomResult<DataKeyCreateResponse, errors::KeyManagerError> {
-    call_encryption_service(state, Method::POST, "key/create", request_body)
-        .await
-        .change_context(errors::KeyManagerError::KeyAddFailed)
+) -> errors::CustomResult<Option<DataKeyCreateResponse>, errors::KeyManagerError> {
+    if !state.is_encryption_service_enabled() {
+        logger::info!(
+            "Encryption service is disabled, skipping key transfer for identifier: {:?}",
```
tsdk02: We are not transferring the key here right? As far as I understand, we are creating the key in the encryption service here. The logging message can be modified.

crates/common_utils/src/types/keymanager.rs
```diff
@@ -91,6 +92,17 @@ impl KeyManagerState {
             infra_values
         })
     }
+
+    pub fn is_encryption_service_enabled(&self) -> bool {
```
sai-harsha-vardhan, line 93: Instead of having feature flag based branching in a fn, we can have two definitions for v1 and v2 right

ThisIsMani: Modified the signature
```rust
    pub fn is_encryption_service_enabled(&self) -> bool {
        cfg!(feature = "encryption_service") && self.enabled
    }
```

sai-harsha-vardhan (approved):

PR #10899, opened 2026-01-12
fix(encryption-service): re-introduce legacy key store decryption behavior for backward compatibility

SanchithHegde:
## Type of Change

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [x] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

This PR fixes a backward compatibility issue introduced in #10572, which attempted to fix the buggy behavior present previously with respect to encryption and decryption of `merchant_key_store` and `user_key_store` tables entries in the database. Prior to #10572, the `merchant_key_store` and `user_key_store` tables entries were being encrypted and decrypted by the encryption service at the time of insertion and retrieval respectively, while the correct behavior should have been to encrypt and decrypt them using the master encryption key configured in the application.

While the changes in #10572 fixed it (and that it works correctly for new entries being inserted), the use of the master encryption key during decryption caused backward compatibility issues for existing entries in these tables, that were encrypted using the encryption service (unless the correct decryption keys were migrated from the encryption service to the router). This PR attempts to address it by reintroducing the previous behavior of using the encryption service for decryption, which can be toggled by means of a runtime config, `use_legacy_key_store_decryption`.

Additionally, this PR also includes some minor changes around `KeyManagerState`:

- Renamed the `KeyManagerState::new()` constructor to `KeyManagerState::mock()`, since the function wasn't accepting any configs and used to populate all fields with default values. Moreover, I noticed that this was only being used in tests, so I renamed it to better reflect its purpose.
- Removed the `Default` implementation on `KeyManagerState`, since it wasn't being used anywhere.

### Additional Changes

This PR introduces a config under the `key_manager` section, called `use_legacy_key_store_decryption` which determines whether the legacy decryption behavior (using the encryption service) should be used for decrypting entries in `merchant_key_store` and `user_key_store` tables.

For existing users updating from a previous version of Hyperswitch, it is recommended that they set this to `true`, while new deployments can set this to `false`.

## Motivation and Context

This PR addresses backward compatibility issues introduced in #10572.

## How did you test it?

I've tested the changes in the PR by running both router and the encryption service locally, with the following scenarios:

1. Existing entries in `merchant_key_store` table encrypted using the encryption service:
   - With `use_legacy_key_store_decryption` set to `true`: Merchant account retrieve operation via Postman succeeds, indicating successful decryption using the encryption service. :white_check_mark:
   - With `use_legacy_key_store_decryption` set to `false`: Merchant account retrieve using the admin API key fails with a 401 status code, and logs indicate decryption failure. :x:
2. Newer entries in the `merchant_key_store` table encrypted using the master encryption key (created after the changes in #10572):
   - With `use_legacy_key_store_decryption` set to `true`: Merchant account retrieve succeeds, with logs indicating that the initial decryption attempt with encryption service failed, but decryption with master key succeeded. :white_check_mark:
   - With `use_legacy_key_store_decryption` set to `false`: Merchant account retrieve succeeds, with logs indicating successful decryption using the master key. :white_check_mark:

## Checklist

### DIFF
--- a/crates/common_utils/src/types/keymanager.rs
+++ b/crates/common_utils/src/types/keymanager.rs
@@ -53,20 +53,9 @@ pub struct KeyManagerState {
     pub use_legacy_key_store_decryption: bool,
 }
 
-impl Default for KeyManagerState {
-    fn default() -> Self {
-        Self::new()
-    }
-}
-
 impl KeyManagerState {
-    /// Creates a KeyManagerState for usage in tests.
-    ///
-    /// # Warning
-    /// This constructor should only be used in test code as this does not accept any configs
-    /// and instead uses default values.
-    /// Production code should use suitable methods or implementations which accept relevant configs.
-    pub fn new() -> Self {
+    /// Creates a mock KeyManagerState with default values for testing.
+    pub fn mock() -> Self {
         Self {
             tenant_id: id_type::TenantId::get_default_tenant_id(),
             global_tenant_id: id_type::TenantId::get_default_global_tenant_id(),
--- a/crates/router/src/db/api_keys.rs
+++ b/crates/router/src/db/api_keys.rs
@@ -394,7 +394,7 @@ mod tests {
         #[allow(clippy::expect_used)]
         let mockdb = MockDb::new(
             &redis_interface::RedisSettings::default(),
-            KeyManagerState::new(),
+            KeyManagerState::mock(),
         )
         .await
         .expect("Failed to create Mock store");
@@ -494,7 +494,7 @@ mod tests {
 
         let db = MockDb::new(
             &redis_interface::RedisSettings::default(),
-            KeyManagerState::new(),
+            KeyManagerState::mock(),
         )
         .await
         .expect("Failed to create Mock store");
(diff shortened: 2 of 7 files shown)


ThisIsMani:
I have a question, since we've already pushed out code with new encryption logic. Some of the open source people might have a mix of both legacy keys and new ones.

If I understand correctly, with the flag that this PR is introducing, users can use either only the legacy logic or the new logic, but not both. So, some of the data won't work in both cases. Correct me if I am wrong on this.

If my assumption is right, should we switch to the legacy code based on the type of data? Since we have different format for the new encryption logic, I think we can leverage that.

SanchithHegde:
> I have a question, since we've already pushed out code with new encryption logic. Some of the open source people might have a mix of both legacy keys and new ones.

We haven't done any monthly releases which include this change.

> If I understand correctly, with the flag that this PR is introducing, users can use either only the legacy logic or the new logic, but not both. So, some of the data won't work in both cases. Correct me if I am wrong on this.

The legacy behavior (`CryptoOperation::Decrypt`) was to try decrypting key stores with encryption service first and falling back to decryption using master encryption key in case of failures, while the new behavior (`CryptoOperation::DecryptLocally`) is to always use the master encryption key for decryption. Technically, the legacy behavior would work for all scenarios (because of the fallback logic), but would add an unnecessary network call and a few unnecessary database calls (on the encryption service side), which the toggle introduced in this PR would help prevent.

ShankarSinghC (approved):

PR #10917, opened 2026-01-13
ci(cypress): move the working connectors to mandatory

NISHANTH1221:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

This pr makes the cypress tests for these connectors: `cryptopay finix worldpay nexixpay fiuu zift` along with other connectors `cybersoucre stripe` as mandatory connectors.

### Additional Changes

## Checklist

### DIFF
--- a/.github/workflows/cypress-tests-runner.yml
+++ b/.github/workflows/cypress-tests-runner.yml
@@ -399,7 +399,7 @@ jobs:
  
       - name: Upload server logs on failure
         if: env.RUN_TESTS == 'true' && failure()
-        uses: actions/upload-artifact@v6
+        uses: actions/upload-artifact@v7
         with:
           name: ${{ matrix.batch.server_logs_artifact }}
           path: ${{ env.SERVER_LOG_PATH }}
@@ -655,7 +655,7 @@ jobs:
 
       - name: Upload server logs on failure
         if: env.RUN_TESTS == 'true' && failure()
-        uses: actions/upload-artifact@v6
+        uses: actions/upload-artifact@v7
         with:
           name: ${{ matrix.batch.server_logs_artifact }}
           path: ${{ env.SERVER_LOG_PATH }} 


.github/workflows/cypress-tests-runner.yml
```diff
@@ -147,10 +148,52 @@ jobs:
...
+      mandatory-matrix: ${{ steps.set-mandatory-matrix.outputs.mandatory-matrix }} # Key must match
+      run_tests: ${{ steps.set-mandatory-matrix.outputs.run_tests }}
+    steps:    
+      - name : Set matrix for mandatory connectors batches
+        id: set-mandatory-matrix
+        shell: bash
+        run: |
+          echo "run_tests=${RUN_TESTS}" >> "$GITHUB_OUTPUT"
+          {
+            echo "mandatory-matrix<<EOF" # Changed from "matrix<<EOF"
```
SanchithHegde: Nit: Can remove the `# Changed from "matrix<<EOF"` comment.

Also, I don't think there's a problem with this being called just `matrix` here, you would just change the `outputs.mandatory-matrix` line like so:

```yaml
mandatory-matrix: ${{ steps.set-mandatory-matrix.outputs.matrix }}
```

.github/workflows/cypress-tests-runner.yml
```diff
@@ -147,10 +148,52 @@ jobs:
           path: target/debug/router
           retention-days: 1
 
+  prepare-mandatory-matrix:
+    name: Prepare Mandatory Connectors matrix to run cypress tests
+    runs-on: hyperswitch-runners
+    outputs:
+      mandatory-matrix: ${{ steps.set-mandatory-matrix.outputs.mandatory-matrix }} # Key must match
+      run_tests: ${{ steps.set-mandatory-matrix.outputs.run_tests }}
```
SanchithHegde: Can't we point this at `RUN_TESTS` directly?

.github/workflows/cypress-tests-runner.yml
```diff
@@ -147,10 +148,52 @@ jobs:
           path: target/debug/router
           retention-days: 1
 
+  prepare-mandatory-matrix:
+    name: Prepare Mandatory Connectors matrix to run cypress tests
+    runs-on: hyperswitch-runners
+    outputs:
+      mandatory-matrix: ${{ steps.set-mandatory-matrix.outputs.mandatory-matrix }} # Key must match
```
SanchithHegde: Nit: Remove comment.

.github/workflows/cypress-tests-runner.yml
```diff
@@ -320,24 +360,54 @@ jobs:
...
         uses: actions/upload-artifact@v4
         with:
-          name: cypress-test-results-mandatory
+          name: ${{ matrix.batch.report_artifact_name }}
           path: |
             cypress-tests/cypress/reports/
           retention-days: 1
+  
+  mandatory-cypress-status-check:
+    name: Check on Mandatory connector tests
```
SanchithHegde: Nit: Can we update the name to indicate that these are Cypress tests?

Something along the lines of "Consolidated mandatory Cypress test result"? 

Can optionally also update the job ID...

.github/workflows/cypress-tests-runner.yml
```diff
@@ -150,15 +150,58 @@ jobs:
...
+                  "connectors": "'"${MANDATORY_PAYMENTS_CONNECTORS_BATCH_3}"'",
+                  "parallel": 3,
+                  "report_artifact_name": "cypress-mandatory-test-results-batch-3"
+                }
+              ]
+            }'
+            echo "EOF"
+          } >> "$GITHUB_OUTPUT"
+
+  runner:
```
SanchithHegde: Nit: Can we please rename this?
Also, remember to update references to this job ID elsewhere in the workflow file.

.github/workflows/cypress-tests-runner.yml
```diff
@@ -328,31 +371,64 @@ jobs:
...
-        env:
-          GH_TOKEN: ${{ github.token }}
+  mandatory-cypress-tests-consolidated-result:
+    name: Mandatory Cypress Tests — Consolidated Result
+    needs: [runner]
+    runs-on: ubuntu-latest
+    if: |
+      (github.event_name == 'merge_group' ||
+      (github.event_name == 'pull_request' &&
+        contains(github.event.pull_request.labels.*.name, 'S-test-ready'))) && always()
```
SanchithHegde: Why do we need `always()` even when we also have some other conditions in the `if` condition? Is there a reason behind it?

.github/workflows/cypress-tests-runner.yml
```diff
@@ -328,31 +371,64 @@ jobs:
...
+    env:
+      GH_TOKEN: ${{ github.token }}
```
SanchithHegde: Please move this to be read at the step level instead of setting it for the entire job.

.github/workflows/cypress-tests-runner.yml
```diff
@@ -328,31 +371,64 @@ jobs:
...
+
+          echo "Runner result: $RESULT"
+          echo "Run tests env: $RUN_TESTS_VAL"
 
-  prepare-matrix:
+          if [ "$RUN_TESTS_VAL" == "false" ]; then
+            gh pr edit ${{ github.event.pull_request.number }} --remove-label "S-test-ready" --repo ${{ github.repository }}
+            echo "Tests were skipped (Fork PR). Marking as success."
+            exit 0
+          fi
```
SanchithHegde, line 443: You won't even be able to edit the PR (to remove label) when the check is running from a fork. The token won't have permissions to edit the PR, because this check would be running in the context of a fork.

.github/workflows/cypress-tests-runner.yml
NISHANTH1221: Because the runner implicitly uses `success()` if `always()` is not mentioned so what happens is that in case of failure of a mandatory test even though the condition is true but it implicitly got false it will skip the tests. But I want the test not to be skipped but to fail.

.github/workflows/cypress-tests-runner.yml
NISHANTH1221: Yes. Removing it.

.github/workflows/cypress-tests-runner.yml
```diff
@@ -150,15 +150,58 @@ jobs:
...
+                  "connectors": "'"${MANDATORY_PAYMENTS_CONNECTORS_BATCH_3}"'",
+                  "parallel": 3,
+                  "report_artifact_name": "cypress-mandatory-test-results-batch-3"
+                }
+              ]
+            }'
+            echo "EOF"
+          } >> "$GITHUB_OUTPUT"
+
+  mandatory-cypress-tests-runners:
```
SanchithHegde: Nit: Sufficient to name this `mandatory-cypress-tests`

.github/workflows/cypress-tests-runner.yml
SanchithHegde: Then you probably need `(success() || failure())` instead of `always()`, as `always()` causes this to run even if the workflow is cancelled:

https://docs.github.com/en/actions/reference/workflows-and-actions/expressions#always

Or we can also use `!cancelled()`, as suggested in the GitHub docs.

.github/workflows/cypress-tests-runner.yml
```diff
@@ -328,31 +371,68 @@ jobs:
...
+    steps:
+      - name: Set GitHub token for gh
+        if: ${{ env.RUN_TESTS == 'true' }}
+        shell: bash
         run: |
-          gh pr edit ${{ github.event.pull_request.number }} --remove-label "S-test-ready" --repo ${{ github.repository }}
+          echo "GH_TOKEN=${{ github.token }}" >> "$GITHUB_ENV"
+    
+      - name: Evaluate Matrix Results
+        shell: bash
```
SanchithHegde, line 430: Add an `env` block for this step and remove the `Set GitHub token for gh` step. Something like:


```yaml
      - name: Evaluate Matrix Results
        shell: bash
        env:
          GH_TOKEN: ${{ github.token }}
        run: ...
```

.github/workflows/cypress-tests-runner.yml
```diff
@@ -328,31 +371,64 @@ jobs:
...
+          if [ "$RUN_TESTS_VAL" == "false" ]; then
+            echo "Tests were skipped (Fork PR). Marking as success."
+            exit 0
+          fi
+
+          if [ "$RESULT" == "success" ]; then
+            echo "All mandatory tests passed."
+            exit 0
+          else
+            echo "Mandatory tests failed or were cancelled."
```
SanchithHegde: Nit: We won't be running this if the workflow was cancelled though?

.github/workflows/cypress-tests-runner.yml
```diff
@@ -328,31 +371,64 @@ jobs:
...
+              echo "Removing 'S-test-ready' label from the PR."
+              gh pr edit ${{ github.event.pull_request.number }} --remove-label "S-test-ready" --repo ${{ github.repository }}
+            fi
+            exit 1
+          fi
+
+  prepare-optional-matrix:
     name: Prepare Optional Connectors matrix to run cypress tests
-    if: ${{ github.event_name == 'merge_group' }}
+    if: false
```
SanchithHegde, line 460: Nit: Can add a comment to this and the `runner-optional-connectors` job, stating why we've disabled them for now.

.github/workflows/cypress-tests-runner.yml
```diff
@@ -328,31 +371,64 @@ jobs:
...
         shell: bash -leuo pipefail {0}
         run: |
-          scripts/execute_cypress.sh --parallel 2
+          scripts/execute_cypress.sh --parallel ${{ matrix.batch.parallel }}
 
           kill "${{ env.PID }}"
 
       - name: Upload Cypress test results
-        if: env.RUN_TESTS == 'true' && failure()
+        if: env.RUN_TESTS == 'true'  && failure()
```
SanchithHegde: Nit: If pushing any other changes, please revert this whitespace change.

SanchithHegde (approved):

SanchithHegde (dismissed):

likhinbopanna (dismissed):

.github/workflows/cypress-tests-runner.yml
SanchithHegde: We can use `upload-artifact@v7` and `download-artifact@v8`, if that helps. I believe they support direct uploads and downloads, and don't necessarily require zipping artifacts before uploading.

SanchithHegde, line 465: What about optional connectors batch 3? Doesn't seem to be included in this list?

NISHANTH1221: We wont be running optional since there is a resource constraint.

PR #11036, opened 2026-01-23
ci(cypress): pause the optional connector runs, add delay to parallel jobs, run tests on both github and hyperswitch-runners, add retry config for cypress

NISHANTH1221:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

This pr pause the optional connector cypress tests which were running on merge level. This PR also pause the v2 tests at merge level.
 This Pr also runs cypress tests with both `hyperswitch-runners` and `github-runners`.  While running parallel it also adds delay between the jobs so they wont fight over the same cypress resources and crash while loading the cypress tests.

This pr also adds `retries:2` in cypress tests which is a config which enables cypress to retry the failed tests 3 times in total before declaring it as failed. This would help us tackle the failures.

### Additional Changes

## Checklist

### DIFF
--- a/.github/workflows/cypress-tests-runner.yml
+++ b/.github/workflows/cypress-tests-runner.yml
@@ -197,7 +197,7 @@ jobs:
 
       - name: Build project
         if: ${{ env.RUN_TESTS == 'true' }}
-        run: cargo build --package router --bin router --jobs 3
+        run: cargo build --package router --bin router
 
       - name: Upload binary artifact
         if: ${{ env.RUN_TESTS == 'true' }}


.github/workflows/cypress-tests-runner.yml
```diff
@@ -551,7 +822,7 @@ jobs:
           ROUTER__APPLEPAY_DECRYPT_KEYS__APPLE_PAY_MERCHANT_CERT_KEY: ${{ secrets.APPLE_PAY_MERCHANT_CERTIFICATE_KEY }}
         run: |
           # Start the server in the background
-          target/debug/router &
+          target/debug/router > "${SERVER_LOG_PATH}" 2>&1 &
```
SanchithHegde: Where is this env var being set for the `runner-optional-connectors` job?

.github/workflows/cypress-tests-runner.yml
```diff
@@ -594,7 +865,7 @@ jobs:
     needs: build-hyperswitch
     if: |
       github.event_name == 'pull_request' && contains(github.event.pull_request.labels.*.name, 'S-test-ready')
-    runs-on: hyperswitch-runners
+    runs-on: ubuntu-latest
```
SanchithHegde: I feel we can keep "alpha" connectors on hyperswitch-runners itself, to reduce the usage of GitHub runners. What do you think?

cypress-tests/cypress.config.js
```diff
@@ -61,6 +61,7 @@ export default defineConfig({
       return config;
     },
     experimentalRunAllSpecs: true,
+    retries: 2,
```
SanchithHegde, line 64: Can you add this in PR description?

We can also update PR title suitably, for this and the delay being passed to GNU parallel, if required.

.github/workflows/cypress-tests-runner.yml
NISHANTH1221: We dont need server logs here since we are pausing them. Had this pushed by mistake

.github/workflows/cypress-tests-runner.yml
NISHANTH1221: Since it is a non mandatory check and we have plenty of hyperswitch-runners ( since we are pausing optional runs) yes we can move it to hyperswitch-runners

cypress-tests/cypress.config.js
NISHANTH1221: Done

.github/workflows/cypress-tests-runner.yml
```diff
@@ -149,7 +149,65 @@ jobs:
...
+      - name: Install sccache
+        if: ${{ env.RUN_TESTS == 'true' }}
+        uses: taiki-e/install-action@v2
+        with:
+          tool: sccache
+          checksum: true
+
+      - name: Build project
+        if: ${{ env.RUN_TESTS == 'true' }}
+        run: cargo build --package router --bin router --jobs 3
```
SanchithHegde: Nit: Do we need the `--jobs 3` for `hyperswitch-runners`?

cypress-tests/cypress.config.js
SanchithHegde: Do we necessarily need 2 retries or is one sufficient? Considering the time involved in running them...

.github/workflows/cypress-tests-runner.yml
NISHANTH1221: removing it and let it choose the value on its own

cypress-tests/cypress.config.js
NISHANTH1221: The retry happens for that test case which failed only not for whole connector so this wont cause any significant increase in time

PR #11076, opened 2026-01-29
ci(cypress): refactor label-based trigger of cypress-tests

AdityaKumaar21:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [x] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

The PR refactors the label-based trigger from if block to separate step in the jobs so that we can re run the failed workflows and also to avoid adding additional commit just to run after cypress-tests.

### Additional Changes

## How did you test it?

When `S-test-ready` label is not present:-
Screenshot 2026-02-18 at 12 09 32
Resolve S-test-ready label step will fail
Screenshot 2026-02-18 at 12 09 44
Cypress tests will be skipped:-
Screenshot 2026-02-18 at 12 10 15

To run the cypress tests without adding additional commits
Step 1:-
Add the S-test-ready label
Screenshot 2026-02-18 at 12 15 28
Step 2:-
Click on the failed step of Resolve S-test-ready
Screenshot 2026-02-18 at 12 14 19
Step 3:-
Click on the re-run option of failed jobs present on top right side 
Screenshot 2026-02-18 at 12 14 57

Demo video:-

https://github.com/user-attachments/assets/f9991506-50ff-4235-9acd-d39df1b65001

## Checklist


.github/workflows/cypress-tests-runner.yml
```diff
@@ -95,9 +95,9 @@ jobs:
   build-hyperswitch:
     name: Build hyperswitch server
     if: |
-      github.event_name == 'merge_group' ||
-      (github.event_name == 'pull_request' && contains(github.event.pull_request.labels.*.name, 'S-test-ready'))
+      github.event_name == 'merge_group' || github.event_name == 'pull_request'
```
SanchithHegde: Do we need this anymore, since we have the `RUN_TESTS` env var to determine things?

This was added in #10820, so we should be able to remove these job-level `if` conditions?

.github/workflows/cypress-tests-runner.yml
```diff
@@ -106,6 +106,29 @@ jobs:
         shell: bash
         if: ${{ env.RUN_TESTS == 'false' }}
         run: echo 'Skipping tests for PRs from forks'
+        
+      - name: Check S-test-ready label
+        if: github.event_name == 'pull_request'
```
SanchithHegde: This would be:

```suggestion
        if: ${{ env.RUN_TESTS == 'true' }}
```

We can't run tests in case of forks.

.github/workflows/cypress-tests-runner.yml
```diff
@@ -106,6 +106,29 @@ jobs:
...
+        env:
+          GH_TOKEN: ${{ github.token }}
+        shell: bash
+        run: |
+          if gh pr view ${{ github.event.pull_request.number }} \
+              --repo ${{ github.repository }} \
+              --json labels \
+              --jq '.labels[].name' | grep -q '^S-test-ready$'; then
```
SanchithHegde: Nit: Prefer to use long forms of the CLI flags, since they tend to be more readable. In this case, the `-q` flag passed to `grep`.

.github/workflows/cypress-tests-runner.yml
```diff
@@ -106,6 +106,29 @@ jobs:
...
+          else
+            echo "RUN_CYPRESS=false" >> $GITHUB_ENV
+            echo "S-test-ready label not found"
+          fi
+
+      - name: Fail if S-test-ready label is missing
+        if: github.event_name == 'pull_request' && env.RUN_CYPRESS == 'false'
+        run: |
+          echo "::error::S-test-ready label is missing. Add the label and re-run jobs to execute Cypress tests."
+          exit 1
```
SanchithHegde: Can't we move this erroring out to the `else` block in the previous step? 

Also keep in mind that we'd have to skip tests for forks regardless of label, and they would run only in merge queue in case of PRs opened from forks.

.github/workflows/cypress-tests-runner.yml
```diff
@@ -211,9 +234,7 @@ jobs:
   runner:
     name: Run mandatory Cypress tests
     needs: build-hyperswitch
-    if: |
-      github.event_name == 'merge_group' ||
-      (github.event_name == 'pull_request' && contains(github.event.pull_request.labels.*.name, 'S-test-ready'))
+    if: github.event_name == 'merge_group' || github.event_name == 'pull_request'
```
SanchithHegde: Same comments as before applicable to the rest of the changes.

.github/workflows/cypress-tests-runner.yml
```diff
@@ -863,8 +907,7 @@ jobs:
   runner_alpha:
     name: Run optional Cypress tests for alpha connectors
     needs: build-hyperswitch
```
pixincreate: this should be `build-hyperswitch-runners` instead, imo.

.github/workflows/cypress-tests-runner.yml
```diff
@@ -106,6 +104,22 @@ jobs:
...
+        run: |
+          if gh pr view ${{ github.event.pull_request.number }} \
+              --repo ${{ github.repository }} \
+              --json labels \
+              --jq '.labels[].name' | grep --quiet '^S-test-ready$'; then
+            echo "S-test-ready label found"
+          else
+            echo "::error::S-test-ready label is missing. Add the label and re-run jobs to execute Cypress tests."
+            exit 1
+          fi
```
pixincreate: instead of having the same piece of code across the file, how about setting the value once, export / output the value. validate that in subsequent steps?

may be, @SanchithHegde might not agree with this, but yeah, just a thought.

we can instead, have a re-usable workflow, which can provide us a bit of leverage in reducing the redundant duplicate code, imo.

.github/workflows/cypress-tests-runner.yml
```diff
@@ -1070,7 +1112,7 @@ jobs:
 
   runner_v2:
     name: Run Cypress tests on v2 and generate coverage report
-    if: github.event_name == 'pull_request' && contains(github.event.pull_request.labels.*.name, 'S-test-ready')
+    if: github.event_name == 'merge_group' || github.event_name == 'pull_request'
```
pixincreate: we will be running this in merge queue too?

.github/workflows/cypress-tests-runner.yml
```diff
@@ -1106,9 +1148,25 @@ jobs:
 
     steps:
       - name: Skip tests for PRs from forks
-        shell: bash
         if: ${{ env.RUN_TESTS == 'false' }}
+        shell: bash
```
pixincreate, line 1127: these wont make any difference

.github/workflows/cypress-tests-runner.yml
```diff
@@ -33,6 +33,36 @@ env:
...
+          else
+            echo "::error::S-test-ready label is missing. Add the label and re-run jobs to execute Cypress tests."
+            exit 1
+          fi
+
+      - name: Set S-test-ready output for skipped tests
+        id: set-s-test-ready-for-skipped-tests
+        if: ${{ env.RUN_TESTS == 'false' }}
+        shell: bash
+        run: echo "s_test_ready=false" >> "$GITHUB_OUTPUT"
```
SanchithHegde, line 64: What would happen in case of PRs opened from forks?

.github/workflows/cypress-tests-runner.yml
```diff
@@ -1067,7 +1086,7 @@ jobs:
 
   runner_v2:
     name: Run Cypress tests on v2 and generate coverage report
-    if: github.event_name == 'pull_request' && contains(github.event.pull_request.labels.*.name, 'S-test-ready')
+    needs: resolve-s-test-ready
```
SanchithHegde, line 1091: Because we've removed the `if` here, does this mean that the v2 Cypress test would run in the merge queue as well?

.github/workflows/cypress-tests-runner.yml
AdityaKumaar21: If PRs are opened from forks, `s_test_ready=false` and jobs running cypress tests will be skipped like how it was used to be.

.github/workflows/cypress-tests-runner.yml
```diff
@@ -859,9 +880,7 @@ jobs:
 
   runner_alpha:
     name: Run optional Cypress tests for alpha connectors
-    needs: build-hyperswitch-runners
-    if: |
-      github.event_name == 'pull_request' && contains(github.event.pull_request.labels.*.name, 'S-test-ready')
```
SanchithHegde, line 864: Since we've removed the `if` here, would we be running this against merge queue as well?

PR #11084, opened 2026-01-29
feat(payments): add retry_type to PaymentAttempt and related schema updates

ayush22667:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

Currently, all payment retries are treated the same in the PaymentAttempt model. There is no way to distinguish whether a retry was triggered manually or automatically.

Introduce a new field retry_type in PaymentAttempt to explicitly capture the origin of a retry.
```
 RetryType {
    /// Manual retry initiated by user/merchant
    ManualRetry,
    /// Automatic retry triggered by the system (GSM-based)
    AutoRetry,
}
```

### Additional Changes

## Motivation and Context

To differentiate between auto retries and manual retries payment.

## How did you test it?

image

## Checklist

### DIFF
--- a/crates/router/src/core/payments/retry.rs
+++ b/crates/router/src/core/payments/retry.rs
@@ -506,7 +506,7 @@ where
     D: payments::OperationSessionGetters<F> + payments::OperationSessionSetters<F> + Send + Sync,
 {
     let new_attempt_count = payment_data.get_payment_intent().attempt_count + 1;
-    let new_payment_attempt = make_new_payment_attempt(
+    let new_payment_attempt = make_new_auto_retry_payment_attempt(
         connector,
         payment_data.get_payment_attempt().clone(),
         new_attempt_count,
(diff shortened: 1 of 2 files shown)


crates/router/src/core/payments/retry.rs
```diff
@@ -847,6 +847,7 @@ pub fn make_new_payment_attempt(
         debit_routing_savings: Default::default(),
         is_overcapture_enabled: Default::default(),
         error_details: Default::default(),
+        retry_type: Some(storage_enums::RetryType::AutoRetry),
```
hrithikesh026, line 850: this function is named `make_new_payment_attempt`. Why are we hardcoding this as AutoRetry

hrithikesh026:
how are we going to backfill for existing retried attempts?

crates/router/src/core/payments/retry.rs
ayush22667: because this will called when new attempt will be created by auto retry

hrithikesh026 (approved):

PR #11091, opened 2026-01-30
feat(payment_methods): add and integrate modular pm update service into payments flow

kashif-m:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description
Added modular PM update support by adding typed request/response models in the `payment_methods` client and integrating the post‑auth update call into the v1 payments flow for eligible merchants.

### Additional Changes

## Motivation and Context
Enable routing to the modular v2 PM service after authorization so connector token details and NTID can be updated via the modular service for eligible merchants, while keeping legacy behavior for others.

## How did you test it?
Manual local testing with `/payments` (v1) and `/v2/payment-methods/{id}/update-saved-payment-method` flows; validated headers and payloads in logs.

<details>
  <summary>Cypress tests for backward compatibility</summary>

image

[adyen_report.html](https://github.com/user-attachments/files/25063375/adyen_report.html)

</details>

## Checklist

### DIFF
--- a/crates/router/src/core/payment_methods/transformers.rs
+++ b/crates/router/src/core/payment_methods/transformers.rs
@@ -15,9 +15,9 @@ use error_stack::ResultExt;
 #[cfg(feature = "v2")]
 use hyperswitch_domain_models::payment_method_data;
 use josekit::jwe;
-use masking::{ExposeInterface, PeekInterface};
 #[cfg(feature = "v1")]
 use masking::Mask;
+use masking::{ExposeInterface, PeekInterface};
 #[cfg(feature = "v1")]
 use payment_methods::client::{
     PaymentMethodClient, UpdatePaymentMethod, UpdatePaymentMethodV1Payload,
--- a/crates/router/src/core/payments/operations/payment_response.rs
+++ b/crates/router/src/core/payments/operations/payment_response.rs
@@ -100,7 +100,7 @@ where
         .response
         .as_ref()
         .map_err(|err| {
-            logger::error!(error=?err, "Failed to obtain the network_transaction_id from payment response in modular payment method update call");
+            logger::debug!(error=?err, "Failed to obtain the network_transaction_id from payment response in modular payment method update call");
         })
         .ok()
         .and_then(types::PaymentsResponseData::get_network_transaction_id);


crates/payment_methods/src/client/update.rs
```diff
@@ -1,80 +1,289 @@
...
+    /// Network transaction ID for off-session updates.
     pub network_transaction_id: Option<Secret<String>>,
 }
 
-#[derive(Debug, Clone, Serialize)]
+/// Payment method update data.
+#[derive(Debug, Serialize, Clone)]
+#[serde(deny_unknown_fields)]
 #[serde(rename_all = "snake_case")]
+#[serde(rename = "payment_method_data")]
```
ShankarSinghC, line 58: Why is this required ?

crates/router/src/core/payments.rs
```diff
@@ -1196,15 +1239,17 @@ where
...
-                            state,
-                            platform.get_provider(),
-                            &payment_data,
-                            &router_data_for_pm_mandate,
-                        )
-                        .await?;
+                    if !is_modular_merchant {
+                        operation
+                            .to_post_update_tracker()?
+                            .update_pm_and_mandate(
```
ShankarSinghC: I see that `update_pm_and_mandate` was introduced for only platform use-case, but still don't we have to make this call in case of modular pm service ?

crates/router/src/core/payments.rs
ShankarSinghC, line 1: save_pm_and_mandate is implemented for different flow like `PaymentsAuthorizeData`, `PaymentsSyncData`, `SetupMandateRequestData`, `CompleteAuthorizeData` and does the update accordingly. Are we handling the update for all of these flows ?

crates/router/src/core/payments.rs
```diff
@@ -967,16 +970,35 @@ where
...
+                            .to_post_update_tracker()?
+                            .save_pm_and_mandate(
+                                state,
+                                &router_data,
+                                platform,
+                                &mut payment_data,
+                                &business_profile,
+                            )
+                            .await?;
+                    }
```
ShankarSinghC: Can we please put this into a function?
Also, add some debug logs to indicate whether it’s a modular or inline function call.

crates/router/src/core/payments/operations/payment_response.rs
```diff
@@ -58,6 +75,180 @@ use crate::{
...
+) -> CustomResult<(), errors::ApiErrorResponse>
+where
+    F: Clone,
+{
+    if payload.payment_method_data.is_none()
+        && payload.connector_token_details.is_none()
+        && payload.network_transaction_id.is_none()
+    {
+        return Ok(());
+    }
```
ShankarSinghC: if it is not a error then don't do early return

crates/router/src/core/payments/operations/payment_response.rs
```diff
@@ -58,6 +75,180 @@ use crate::{
...
+where
+    F: Clone + Send + Sync,
+{
+    // #0 - Skip unsupported payment methods.
+    if !matches!(
+        payment_data.payment_attempt.payment_method,
+        Some(enums::PaymentMethod::Card)
+    ) {
+        return Ok(());
+    }
```
ShankarSinghC: call_modular_payment_method_update

crates/router/src/core/payments/operations/payment_response.rs
```diff
@@ -58,6 +75,180 @@ use crate::{
...
+                field_name: "payment_method_id",
+            }
+        })?;
+
+    // #2 - Derive network transaction ID from the connector response.
+    let network_transaction_id = resp
+        .response
+        .as_ref()
+        .ok()
+        .and_then(types::PaymentsResponseData::get_network_transaction_id);
```
ShankarSinghC:
```suggestion
    let network_transaction_id = resp
        .response
        .as_ref()
        .map_err(|err| {
                                    logger::error!(error = ?err, "Failed to obtain the network_transaction_id from payment response in modular payment method update call");
                                })
        .ok()
        .and_then(types::PaymentsResponseData::get_network_transaction_id);
```

crates/router/src/core/payments/operations/payment_response.rs
```diff
@@ -58,6 +75,180 @@ use crate::{
...
+                        metadata: mandate_reference.mandate_metadata,
+                        token: masking::Secret::new(connector_mandate_id),
+                    },
+                )
+        }
+        None => None,
+    };
+
+    let payload = UpdatePaymentMethodV1Payload {
+        payment_method_data: None,
```
ShankarSinghC: Why is hardcoded to `None` ?
Isn't this to update card holder name or nickname ?

crates/payment_methods/src/client/update.rs
kashif-m: These types are duplicated as is from `api_models/payment_method.rs`

It's better to duplicate them as is, since we will remove these redundant types in future and move to using a single source of truth.

crates/router/src/core/payments.rs
kashif-m: `update_pm_and_mandate` calls `update_pm_connector_mandate_details` internally which basically updates `connector_mandate_details` column in `payment_method` table

This is already done as part of PM modular service

crates/router/src/core/payments/operations/payment_response.rs
kashif-m: If there's nothing to be updated, we can skip the API call

crates/router/src/core/payments/operations/payment_response.rs
kashif-m: Card holder name and nickname are sent as part of payment confirmation - which is part of PM creation API call

This data will not change in between confirmation and payment authorisation - as both these are in the same session call.

crates/router/src/core/payments/operations/payment_response.rs
kashif-m: If it's a payment method other than card, then we will skip it since PM modular service does not handle wallet flows yet, right?

crates/router/src/core/payments.rs
kashif-m: Yes, this is handled for all the flows accordingly

crates/router/src/core/payments/operations/payment_response.rs
ShankarSinghC: @prasunna09 can you please verify this ?

crates/router/src/core/payments/operations/payment_response.rs
ShankarSinghC: Sure, but let's not have it as early return

crates/router/src/core/payments/operations/payment_response.rs
prasunna09: @kashif-m it might change, when in repeat CIT, payment token along with CardTokenData {card holder name, cvc} is passed in confirm request, then, fetch payment method and update payment method happens. in this update we might want to update card holder name

crates/payment_methods/src/client/update.rs
```diff
@@ -1,80 +1,289 @@
...
     pub card_holder_name: Option<Secret<String>>,
+    /// Card holder nickname.
     pub nick_name: Option<Secret<String>>,
+    /// Card CVC (optional).
     pub card_cvc: Option<Secret<String>>,
 }
 
-#[derive(Clone, Debug, Deserialize)]
+/// Connector token details update payload.
+#[derive(Debug, Deserialize, Serialize, Clone)]
```
prasunna09: we have ConnectorTokenDetails in crates/payment_methods/src/types.rs, can we use that, instead of creating new struct

crates/payment_methods/src/client/update.rs
```diff
@@ -1,80 +1,289 @@
...
+#[serde(deny_unknown_fields)]
+#[serde(rename_all = "snake_case")]
+#[serde(rename = "payment_method_data")]
+pub enum PaymentMethodResponseData {
+    Card(CardDetailFromLocker),
+}
+
+/// Card details as returned by the modular service.
+#[derive(Debug, Deserialize)]
+pub struct CardDetailFromLocker {
```
prasunna09: same here, response structs are defined in types.rs. please re-use

prasunna09: and we have domain types for this as well

crates/router/src/core/payments/operations/payment_response.rs
```diff
@@ -58,6 +75,180 @@ use crate::{
...
 
+#[cfg(feature = "v1")]
+async fn call_modular_payment_method_update<F>(
+    state: &SessionState,
+    payment_data: &PaymentData<F>,
+    payment_method_id: &str,
+    payload: UpdatePaymentMethodV1Payload,
+) -> CustomResult<(), errors::ApiErrorResponse>
+where
+    F: Clone,
```
prasunna09: lets move this to payment_methods/transformers.rs file, where we can have all the transformation to Modular api calls

crates/router/src/core/payments/operations/payment_response.rs
```diff
@@ -58,6 +75,180 @@ use crate::{
...
+{
+    if payload.payment_method_data.is_none()
+        && payload.connector_token_details.is_none()
+        && payload.network_transaction_id.is_none()
+    {
```
prasunna09: also we have to call update api only if payments is successful transaction right?

prasunna09 (changes requested):

crates/router/src/core/payments/operations/payment_response.rs
kashif-m: Handled it as an if condition

crates/payment_methods/src/client/update.rs
```diff
@@ -1,73 +1,163 @@
...
+    /// Connector token details to update.
     pub connector_token_details: Option<ConnectorTokenDetails>,
+    /// Network transaction ID for off-session updates.
     pub network_transaction_id: Option<Secret<String>>,
 }
 
-#[derive(Debug, Clone, Serialize)]
+/// Payment method update data.
+#[derive(Debug, Serialize, Clone)]
+#[serde(deny_unknown_fields)]
```
prasunna09: why is deny_unknown on request?

crates/payment_methods/src/client/update.rs
```diff
@@ -1,73 +1,163 @@
...
     }
 }
 
 impl UpdatePaymentMethod {
+    fn validate_request(
+        &self,
+        request: &UpdatePaymentMethodV1Request,
+    ) -> Result<(), MicroserviceClientError> {
+        if request.payment_method_id.trim().is_empty() {
+            return Err(MicroserviceClientError {
```
prasunna09: please avoid using early returns.
Use `when` instead

crates/router/src/core/payment_methods/transformers.rs
```diff
@@ -1114,3 +1129,70 @@ impl transformers::ForeignFrom<&payment_method_data::SingleUsePaymentMethodToken
         }
     }
 }
+
+#[cfg(feature = "v1")]
+pub async fn call_modular_payment_method_update<F>(
+    state: &routes::SessionState,
+    payment_data: &PaymentData<F>,
```
prasunna09: why do we need to pass entire payment data? just send profile id as argument.

prasunna09: also we ll have domain PaymentMethod Object created during create/fetch, use domain PaymentMethodObject to fetch profile id, and others

crates/router/src/core/payments.rs
```diff
@@ -648,6 +648,7 @@ where
         mut payment_data,
         business_profile,
         mandate_type,
+        feature_set,
```
prasunna09: why in tracker response? fetch call that happens before get_trackers will use this.
returning in trackerResponse wont help fetch payment_method

prasunna09: and other flows like doesnt need this config fetch. there will be an unnecessary db call

crates/router/src/core/payments.rs
```diff
@@ -1196,15 +1201,17 @@ where
...
-                        .to_post_update_tracker()?
-                        .update_pm_and_mandate(
-                            state,
-                            platform.get_provider(),
-                            &payment_data,
-                            &router_data_for_pm_mandate,
-                        )
-                        .await?;
+                    if !feature_set.is_modular_merchant {
+                        operation
```
prasunna09: pass feature_set as argument to avoid future branching in payments operation core

crates/router/src/core/payments/operations/payment_response.rs
```diff
@@ -58,6 +67,134 @@ use crate::{
...
+
+        // #3 - Fill payment method data for cards (update card holder name, nick_name & cvc).
+        // Use request payment method data for card_holder_name and nick_name
+        let payment_method_data =
+            request_payment_method_data.and_then(|method_data| match method_data {
+                domain::PaymentMethodData::Card(card) => {
+                    Some(PaymentMethodUpdateData::Card(CardDetailUpdate {
+                        card_holder_name: card.card_holder_name.clone(),
+                        nick_name: card.nick_name.clone(),
+                        card_cvc: Some(card.card_cvc.clone()),
```
prasunna09: CVC update is not required in post transaction ryt?

crates/router/src/core/payments/operations/payment_response.rs
```diff
@@ -58,6 +67,134 @@ use crate::{
...
 
+/// This implementation executes the flow only when
+/// 1. Payment was created with supported payment methods
+/// 2. Payment attempt's status was not a terminal failure
+#[cfg(feature = "v1")]
+async fn update_modular_pm_and_mandate_impl<F, T>(
+    state: &SessionState,
+    resp: &types::RouterData<F, T, types::PaymentsResponseData>,
+    request_payment_method_data: Option<&domain::PaymentMethodData>,
+    payment_data: &mut PaymentData<F>,
```
prasunna09, line 78: why do we require mutable payment data?

crates/router/src/core/payments/operations/payment_response.rs
```diff
@@ -58,6 +67,134 @@ use crate::{
...
+        };
+
+        // #5 - Execute the modular payment-method update call if there is something to be updated
+        if payload.payment_method_data.is_some()
+            || payload.connector_token_details.is_some()
+            || payload.network_transaction_id.is_some()
+        {
+            call_modular_payment_method_update(state, payment_data, &payment_method_id, payload)
+                .await?;
+            payment_data.payment_attempt.payment_method_id = Some(payment_method_id);
```
prasunna09, line 191: we do we have to set payment method id in payment data, when its actually fetched from payment_data in the first line of this fn?

crates/router/src/core/payments/operations/payment_response.rs
kashif-m: To update `payment_attempt.payment_method_id` which is later updated in DB

crates/router/src/core/payments/operations/payment_response.rs
kashif-m: Will refactor it to fetch from domain payment method instead - is there any point in updating the `payment_method_id` at this point?

kashif-m: `payment_method_id` will be retrieved from domain PaymentMethod now.

crates/payment_methods/src/client/update.rs
kashif-m: This was taken as is from `payment_methods` api_models. Removed these changes

crates/router/src/core/payment_methods/transformers.rs
kashif-m: PaymentMethod object doesn't have `profile_id` since it's tied to a merchant, taking and passing it from payment_attempt instead

crates/router/src/core/payment_methods/transformers.rs
```diff
@@ -1114,3 +1128,56 @@ impl transformers::ForeignFrom<&payment_method_data::SingleUsePaymentMethodToken
...
+    ));
+    parent_headers.insert((
+        headers::X_INTERNAL_API_KEY.to_string(),
+        Maskable::Normal(
+            state
+                .conf
+                .internal_merchant_id_profile_id_auth
+                .internal_api_key
+                .peek()
+                .to_string(),
```
prasunna09: this value should be masked

crates/router/src/core/payment_methods/transformers.rs
```diff
@@ -1114,3 +1128,56 @@ impl transformers::ForeignFrom<&payment_method_data::SingleUsePaymentMethodToken
...
+        &client,
+        UpdatePaymentMethodV1Request {
+            payment_method_id: payment_method_id.to_string(),
+            payload,
+            modular_service_prefix: state.conf.micro_services.payment_methods_prefix.0.clone(),
+        },
+    )
+    .await
+    .map_err(|err| {
+        logger::error!(error=?err, "modular payment method update failed");
```
prasunna09, line 1177: please use logger debug

crates/router/src/core/payments/operations/payment_response.rs
```diff
@@ -58,6 +67,134 @@ use crate::{
...
+                logger::error!("Missing required Param payment_method_id");
+                ::payment_methods::errors::ModularPaymentMethodError::RetrieveFailed
+            })?;
+
+        // #2 - Derive network transaction ID from the connector response.
+        let network_transaction_id = resp
+        .response
+        .as_ref()
+        .map_err(|err| {
+            logger::error!(error=?err, "Failed to obtain the network_transaction_id from payment response in modular payment method update call");
```
prasunna09: use logger debug

crates/router/src/core/payment_methods/transformers.rs
kashif-m: This one can be logger error, since API call failed

ShankarSinghC (approved):

PR #11109, opened 2026-02-02
fix: webhook refactor HS<>UCS

gopikrishna000:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [x] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description
1. Refactored webhook in HS to separate the transformation and extraction logic which are done by hyperswitch and ucs into separate function for each. (Previously everything was there inside process_business_logic funciton and extraction happens interchangebly with fallback without clear dinstinction of which execution_path we are in.)

2. updated execution_mode to send ucs instead of hardcoded always primary mode.

### Additional Changes

## Motivation and Context
**problems**
1. HS always call ucs in primary mode even during shadow execution path.
2. Source verification happens in hs inside business logic., should be moved up to gateway code. Cause due to this, ucs shadow mode will not receive response from mitm as it hs source verification happens post receiving response from ucs.. (ucs call is not parallel)
3. Resource Object extraction also not part of business logic, so to move inside gateway code.
5. there was no response comparison currently through validation service for webhooks.
6. multiple logic uses hs extraction and fallback to ucs extraction interchangeably without clear distinction of which path we are in.
7. If UCS returns source_verified = false, currently we falls back to re-verification using hs source verification instead of trusting ucs
8. ucs should_call_unified_connector_service_for_webhooks dont return updated session state with updated proxy config., so mitm wasnt called from hyperswitch.

## How did you test it?
https://github.com/juspay/hyperswitch-cloud/issues/14178

paypal
https://github.com/juspay/hyperswitch-cloud/issues/14048#issuecomment-3853496877
nuvei
https://github.com/juspay/hyperswitch-cloud/issues/14047#issuecomment-3854245184

## Checklist

### DIFF
--- a/crates/api_models/src/webhooks.rs
+++ b/crates/api_models/src/webhooks.rs
@@ -421,7 +421,6 @@ impl ObjectReferenceId {
     }
 }
 
-#[derive(Clone, Debug)]
 pub struct IncomingWebhookDetails {
     pub object_reference_id: ObjectReferenceId,
     pub resource_object: Vec<u8>,
--- a/crates/api_models/src/webhooks.rs
+++ b/crates/api_models/src/webhooks.rs
@@ -421,21 +421,9 @@ impl ObjectReferenceId {
     }
 }
 
-fn serialize_resource_object<S>(resource_object: &Vec<u8>, serializer: S) -> Result<S::Ok, S::Error>
-where
-    S: serde::Serializer,
-{
-    // Parse Vec<u8> as JSON and serialize it
-    // resource_object is always valid JSON (created from serde_json::to_vec)
-    serde_json::from_slice::<serde_json::Value>(resource_object)
-        .map_err(serde::ser::Error::custom)?
-        .serialize(serializer)
-}
-
-#[derive(Clone, Debug, Serialize, Deserialize)]
+#[derive(Clone, Debug, Deserialize)]
 pub struct IncomingWebhookDetails {
     pub object_reference_id: ObjectReferenceId,
-    #[serde(serialize_with = "serialize_resource_object")]
     pub resource_object: Vec<u8>,
 }
 
--- a/crates/api_models/src/webhooks.rs
+++ b/crates/api_models/src/webhooks.rs
@@ -421,7 +421,7 @@ impl ObjectReferenceId {
     }
 }
 
-#[derive(Clone, Debug, Deserialize)]
+#[derive(Clone, Debug)]
 pub struct IncomingWebhookDetails {
     pub object_reference_id: ObjectReferenceId,
     pub resource_object: Vec<u8>,
--- a/crates/api_models/src/webhooks.rs
+++ b/crates/api_models/src/webhooks.rs
@@ -421,8 +421,21 @@ impl ObjectReferenceId {
     }
 }
 
+fn serialize_resource_object<S>(resource_object: &Vec<u8>, serializer: S) -> Result<S::Ok, S::Error>
+where
+    S: serde::Serializer,
+{
+    // Parse Vec<u8> as JSON and serialize it
+    // resource_object is always valid JSON (created from serde_json::to_vec)
+    serde_json::from_slice::<serde_json::Value>(resource_object)
+        .map_err(serde::ser::Error::custom)?
+        .serialize(serializer)
+}
+
+#[derive(Clone, Debug, Serialize, Deserialize)]
 pub struct IncomingWebhookDetails {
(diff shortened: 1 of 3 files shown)


crates/router/src/core/webhooks/incoming.rs
```diff
@@ -583,129 +701,169 @@ pub enum ThreeDsProcessingMode {
...
+    merchant_connector_account: Option<domain::MerchantConnectorAccount>,
+    execution_mode: common_enums::ExecutionMode,
 ) -> errors::RouterResult<WebhookProcessingResult<'a>> {
-    // Execute UCS path
-    let ucs_result = unified_connector_service::call_unified_connector_service_for_webhook(
+    // Extract object_ref_id early (needed for both UCS and Direct paths)
+    let object_ref_id = connector
+        .get_webhook_object_reference_id(request_details)
+        .change_context(errors::ApiErrorResponse::InternalServerError)
+        .ok();
```
Saptak88: Cant we pass object_ref_id similar to how its passed in process_ucs_webhook_transform

crates/router/src/core/webhooks/incoming.rs
```diff
@@ -268,28 +269,41 @@ async fn incoming_webhooks_core<W: types::OutgoingWebhookType>(
...
 
+            // Extract object_ref_id early (until UCS returns structured identifier info)
+            let object_ref_id = connector
+                .get_webhook_object_reference_id(&request_details)
+                .change_context(errors::ApiErrorResponse::InternalServerError)
+                .ok(); // Don't fail if extraction fails - gateway functions will fetch MCA when needed
+
+            // Pass MCA from URL directly - gateway functions will fetch if needed
+            let merchant_connector_account_owned = mca_data.merchant_connector_account.clone();
+
```
Saptak88: why do we need to clone it here?

crates/router/src/core/webhooks/incoming.rs
```diff
@@ -583,129 +701,169 @@ pub enum ThreeDsProcessingMode {
...
+            merchant_connector_account
+                .as_ref()
+                .and_then(|mca| mca.connector_webhook_details.clone()),
+            connector_name,
+        )
+        .await
+        .switch()
+        .attach_printable(
+            "There was an error in incoming webhook body decoding for shadow processing",
+        )?;
```
Saptak88: Can we add this inside fn process_direct_webhook_transform as decoding will be done for direct only?

crates/api_models/src/webhooks.rs
```diff
@@ -421,8 +421,21 @@ impl ObjectReferenceId {
...
+    // resource_object is always valid JSON (created from serde_json::to_vec)
+    serde_json::from_slice::<serde_json::Value>(resource_object)
+        .map_err(serde::ser::Error::custom)?
+        .serialize(serializer)
+}
+
+#[derive(Clone, Debug, Serialize, Deserialize)]
 pub struct IncomingWebhookDetails {
     pub object_reference_id: ObjectReferenceId,
+    #[serde(serialize_with = "serialize_resource_object")]
```
Saptak88: Why Serialize is needed? wont it be skipped for comparison

crates/router/src/core/webhooks/incoming.rs
```diff
@@ -517,32 +549,106 @@ async fn fetch_three_ds_execution_path(
...
             body,
             request_details,
-            merchant_connector_account,
+            Some(&merchant_connector_account), // Always Some - MCA is non-optional for UCS
+            execution_mode,
         )
         .await?;
+
+    let (webhook_details, event_object) = if let Some(ref object_ref_id) = object_ref_id {
+        let webhook_response_content =
```
Saptak88, line 596: Ideally, object_ref_id should come from UCS. Are we planning to handle this in a separate PR once support for it is added in UCS?

gopikrishna000: yes
https://github.com/juspay/connector-service/issues/485#issue-3901879377

crates/api_models/src/webhooks.rs
gopikrishna000: yes, will remove it

crates/router/src/core/webhooks/incoming.rs
Saptak88, line 1172: wont we have shadow_webhook_details from shadow_ucs_data

PR #11176, opened 2026-02-06
fix(webhook): add missing recovery events in webhooks as event not supported gets returned

pixincreate:
## Type of Change

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

when i was working on https://github.com/juspay/connector-service/pull/473, i observed that the recovery events were non-existent in webhook event list in hyperswitch side. so, if recovery webhook was triggered in the ucs side, this would result in `event not supported` error to be returned.

these fields already exist in ucs. it is just that it's missing in hyperswitch. added the mapping.

### Additional Changes

## Motivation and Context

i observed these fields to be missing in hyperswitch side. so, added.

## How did you test it?

no tests needed.

## Checklist

### DIFF
--- a/crates/api_models/src/webhooks.rs
+++ b/crates/api_models/src/webhooks.rs
@@ -128,6 +128,13 @@ impl IncomingWebhookEvent {
             35 => Self::PayoutExpired,
             #[cfg(feature = "payouts")]
             36 => Self::PayoutReversed,
+
+            // Recovery events
+            37 => Self::RecoveryPaymentFailure,
+            38 => Self::RecoveryPaymentSuccess,
+            39 => Self::RecoveryPaymentPending,
+            40 => Self::RecoveryInvoiceCancel,
+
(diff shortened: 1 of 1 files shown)


crates/api_models/src/webhooks.rs
```diff
@@ -128,6 +128,13 @@ impl IncomingWebhookEvent {
             35 => Self::PayoutExpired,
             #[cfg(feature = "payouts")]
             36 => Self::PayoutReversed,
+
+            // Recovery events
+            37 => Self::RecoveryPaymentFailure,
```
hrithikesh026, line 133: this kind of conversion is not recommended. We should transform from grpc enum to our enum. Not from integers

pixincreate:
closing this pr. not that of a priority right now.

PR #11274, opened 2026-02-16
feat(migration): Add support to migrate ACH tokens

deepanshu-iiitu:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

Extends the migrate-batch API to support migrating ACH bank account details (account number + routing number) alongside existing card migrations. This allows Hyperswitch to migrate saved bank debit payment methods from other PSPs into Hyperswitch.

### Additional Changes

## Motivation and Context

https://github.com/juspay/hyperswitch/issues/11275

## How did you test it?

- Card-only CSV (regression): Upload a CSV with only card records (no account_number/routing_number columns). Verify cards migrate as before.
- ACH-only CSV: Upload a CSV with bank records
- Mixed CSV: Upload a CSV containing both card and ACH rows. Verify card rows return card_migrated: true and ACH rows return bank_account_migrated: true.
- Backward compatibility: Confirm existing CSVs without account_number/routing_number columns still parse correctly.

## Checklist

### DIFF
--- a/crates/api_models/src/payment_methods.rs
+++ b/crates/api_models/src/payment_methods.rs
@@ -3025,6 +3025,8 @@ pub struct PaymentMethodRecord {
     pub nick_name: Option<masking::Secret<String>>,
     pub payment_instrument_id: Option<masking::Secret<String>>,
     pub connector_customer_id: Option<String>,
+    // Card fields are optional to support non-card CSV rows (e.g., ACH bank debit).
+    // For card rows these will still be populated normally from the CSV columns.
     #[serde(default)]
     pub card_number_masked: Option<masking::Secret<String>>,
     #[serde(default)]


crates/api_models/src/payment_methods.rs
```diff
@@ -247,6 +247,9 @@ pub struct PaymentMethodMigrate {
     /// Card Details
     pub card: Option<MigrateCardDetail>,
 
+    /// Bank Debit Details (e.g. ACH account_number + routing_number)
+    pub bank_debit: Option<MigrateBankDebitDetail>,
```
jagan-jaya: instead of adding new fileds for each payment methods, can we add payment_method_data here and handle other pms in one single field

crates/api_models/src/payment_methods.rs
```diff
@@ -290,6 +293,9 @@ pub struct PaymentMethodMigrateResponse {
     //card data migration status
     pub card_migrated: Option<bool>,
 
+    //bank account data migration status
+    pub bank_account_migrated: Option<bool>,
```
jagan-jaya:
```suggestion
    pub payment_method_migrated: Option<bool>,
```

crates/api_models/src/payment_methods.rs
deepanshu-iiitu: Addressed

crates/api_models/src/payment_methods.rs
deepanshu-iiitu: Addressed

crates/api_models/src/payment_methods.rs
```diff
@@ -3287,8 +3339,22 @@ impl
                 card_issuing_country: None,
                 card_issuing_country_code: None,
                 nick_name: record.nick_name.clone(),
-            }),
-            network_token: Some(MigrateNetworkTokenDetail {
+            })
+        };
+
+        let payment_method_data = if is_bank_debit {
+            Some(PaymentMethodCreateData::BankDebit(BankDebitDetail::Ach {
```
jagan-jaya: let's not hardcode bankdebit, try to make it generic

deepanshu-iiitu: Done

jagan-jaya (approved):

PR #11300, opened 2026-02-17
feat(users): Introduce Soft delete for Users

Grovy-3170:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

Implemented soft delete functionality for users in the HyperSwitch system. When a user loses their last role, instead of hard-deleting the user record, the system now:

1. **Marks the user as inactive** by setting `is_active = false`
2. **Purges all sensitive user data** (password, 2FA secrets, recovery codes, lineage context)
3. **Deletes associated dashboard metadata**

Users marked as inactive are effectively "deleted" from the system's perspective:
- Cannot sign in with existing credentials
- Do not appear in active user queries
- JWT tokens are invalidated

**Reactivation Flow**: When an inactive user is re-invited to the system, they are automatically reactivated with a fresh state (new credentials, no 2FA, unverified status). **The `user_id` will be the same as before.**

### Key Changes

- **Database Migration**: Added [is_active](cci:1://file:///Users/aman.kumar.002/JuspayProjects/hyperswitch.git/main/crates/router/src/types/domain/user.rs:1299:4-1301:5) boolean column to `users` table (defaults to `TRUE`)
- **User Model**: Updated [User](cci:2://file:///Users/aman.kumar.002/JuspayProjects/hyperswitch.git/main/crates/diesel_models/src/user.rs:13:0-28:1), [UserNew](cci:2://file:///Users/aman.kumar.002/JuspayProjects/hyperswitch.git/main/crates/diesel_models/src/user.rs:34:0-48:1), and [UserUpdateInternal](cci:2://file:///Users/aman.kumar.002/JuspayProjects/hyperswitch.git/main/crates/diesel_models/src/user.rs:52:0-63:1) structs with [is_active](cci:1://file:///Users/aman.kumar.002/JuspayProjects/hyperswitch.git/main/crates/router/src/types/domain/user.rs:1299:4-1301:5) field
- **Query Methods**: Added new query methods to distinguish between active and all users:
  - `find_active_user_by_email()` / [find_user_by_user_email()](cci:1://file:///Users/aman.kumar.002/JuspayProjects/hyperswitch.git/main/crates/router/src/db/user.rs:91:4-100:5)
  - [find_active_user_by_user_id()](cci:1://file:///Users/aman.kumar.002/JuspayProjects/hyperswitch.git/main/crates/router/src/db/user.rs:102:4-111:5)
  - [find_active_users_by_user_ids()](cci:1://file:///Users/aman.kumar.002/JuspayProjects/hyperswitch.git/main/crates/diesel_models/src/query/user.rs:92:4-111:5)
- **Deactivation**: [deactivate_user_by_user_id()](cci:1://file:///Users/aman.kumar.002/JuspayProjects/hyperswitch.git/main/crates/router/src/db/user.rs:139:4-147:5) replaces `delete_user_by_user_id()`
- **Reactivation**: Added [reactivate_user_by_user_id()](cci:1://file:///Users/aman.kumar.002/JuspayProjects/hyperswitch.git/main/crates/router/src/db/user.rs:159:4-170:5) for re-inviting inactive users
- **Update Enums**: Added `InActiveUpdate` and `ActiveUpdate` variants to `UserUpdate`
- **Dashboard Cleanup**: Added [delete_all_metadata_by_user_id()](cci:1://file:///Users/aman.kumar.002/JuspayProjects/hyperswitch.git/main/crates/router/src/db/dashboard_metadata.rs:12:4-20:5) to purge user's dashboard metadata on deactivation

### Additional Changes

**Database Schema Changes:**
- Migration: [migrations/2026-02-11-120200_add_soft_delete_to_users/up.sql](cci:7://file:///Users/aman.kumar.002/JuspayProjects/hyperswitch.git/main/migrations/2026-02-11-120200_add_soft_delete_to_users/up.sql:0:0-0:0)
- Schema: [crates/diesel_models/src/schema.rs](cci:7://file:///Users/aman.kumar.002/JuspayProjects/hyperswitch.git/main/crates/diesel_models/src/schema.rs:0:0-0:0) - Added [is_active](cci:1://file:///Users/aman.kumar.002/JuspayProjects/hyperswitch.git/main/crates/router/src/types/domain/user.rs:1299:4-1301:5) field to `users` table

## Motivation and Context

**Problem**: The current implementation performs hard deletes on users when they lose their last role. This:
- Causes foreign key constraint violations if user data is referenced elsewhere
- Makes it impossible to track historical user activity
- Prevents re-inviting previously deleted users

**Solution**: Soft delete preserves the user record while making the user functionally "deleted" by:
- Preventing authentication (inactive users cannot sign in)
- Filtering out inactive users from all normal queries
- Purging sensitive data to comply with the business requirement that "to the world, the user is gone"

This approach provides:
- Better data integrity (no orphaned references)
- Audit trail capability
- Ability to reactivate users on re-invitation
- Cleaner deletion flow without cascading deletes

### Manual Testing
1. ✅ Created test user with roles
2. ✅ Deleted all roles → verified user marked as `is_active = false`
3. ✅ Verified password, 2FA, and other sensitive fields were nulled
4. ✅ Attempted sign-in with old credentials → correctly rejected
5. ✅ Re-invited user → verified reactivation with fresh state
6. ✅ Verified new credentials work after reactivation

### Build Verification
- ✅ `cargo build` - successful compilation
- ✅ All modified crates compiled without errors
- ✅ MockDb implementation updated and verified

### Files Modified
- [crates/diesel_models/src/user.rs](cci:7://file:///Users/aman.kumar.002/JuspayProjects/hyperswitch.git/main/crates/diesel_models/src/user.rs:0:0-0:0) - Model updates
- [crates/diesel_models/src/query/user.rs](cci:7://file:///Users/aman.kumar.002/JuspayProjects/hyperswitch.git/main/crates/diesel_models/src/query/user.rs:0:0-0:0) - Query logic
- [crates/diesel_models/src/schema.rs](cci:7://file:///Users/aman.kumar.002/JuspayProjects/hyperswitch.git/main/crates/diesel_models/src/schema.rs:0:0-0:0) - Schema definition
- [crates/router/src/core/user.rs](cci:7://file:///Users/aman.kumar.002/JuspayProjects/hyperswitch.git/main/crates/router/src/core/user.rs:0:0-0:0) - Core user logic
- [crates/router/src/core/user_role.rs](cci:7://file:///Users/aman.kumar.002/JuspayProjects/hyperswitch.git/main/crates/router/src/core/user_role.rs:0:0-0:0) - Role deletion trigger
- [crates/router/src/db/user.rs](cci:7://file:///Users/aman.kumar.002/JuspayProjects/hyperswitch.git/main/crates/router/src/db/user.rs:0:0-0:0) - Storage interface (Store + MockDb)
- [crates/router/src/db/dashboard_metadata.rs](cci:7://file:///Users/aman.kumar.002/JuspayProjects/hyperswitch.git/main/crates/router/src/db/dashboard_metadata.rs:0:0-0:0) - Metadata cleanup
- [crates/router/src/types/domain/user.rs](cci:7://file:///Users/aman.kumar.002/JuspayProjects/hyperswitch.git/main/crates/router/src/types/domain/user.rs:0:0-0:0) - Domain logic
- `crates/router/src/services/authentication.rs` - JWT validation
- `migrations/2026-02-11-120200_add_soft_delete_to_users/` - Database migration

## Checklist

# Testing Flow:
### curl:
```
curl --location 'http://localhost:8080/user/user/invite_multiple' \
--header 'Authorization: Bearer {{token}}' \
--data-raw '[
    {
        "email": "test1@juspay.in",
        "name": "25",
        "role_id": "merchant_view_only"
    }
]'
```
### Response:
Screenshot 2026-02-20 at 7 52 39 PM

### Db State:

Screenshot 2026-02-20 at 7 53 52 PM

`user_id = 0d071c9e-8f3f-4721-87ba-5d10957d36c7`

### curl:

```
curl --location --request DELETE 'http://localhost:8080/user/user/delete' \
--header 'Authorization: Bearer {{token}}' \
--data-raw '{
    "email": "test1@juspay.in"
}'
```

### Response:

`Delete query so no response`

### Db State:

Screenshot 2026-02-20 at 7 57 04 PM
(The user is now `inactive`)

### Db State:

Screenshot 2026-02-20 at 7 58 54 PM

`user_id = 0d071c9e-8f3f-4721-87ba-5d10957d36c7` (same as before)

### curl:

```
curl --location 'http://localhost:8080/user/signup' \
--data-raw '{
    "email": "test1@juspay.in",
    "password": "Testing@123"
}'
```

### Db State:

Screenshot 2026-02-20 at 8 02 37 PM


crates/diesel_models/src/user.rs
```diff
@@ -78,6 +81,11 @@ pub enum UserUpdate {
     LineageContextUpdate {
         lineage_context: LineageContext,
     },
+    InActiveUpdate,
```
ThisIsMani: Name this better.

crates/diesel_models/src/user.rs
```diff
@@ -78,6 +81,11 @@ pub enum UserUpdate {
...
+    ActiveUpdate {
+        new_name: Option<String>,
+        new_password: Option<Secret<String>>,
+    },
```
ThisIsMani: It's better to separate this out.

crates/router/src/core/user.rs
```diff
@@ -203,6 +204,7 @@ pub async fn get_user_details(
             entity_type: role_info.get_entity_type(),
             theme_id: theme.map(|theme| theme.theme_id),
             version,
+            is_active: user.is_active(),
```
ThisIsMani: Not needed. This will never be false.

crates/api_models/src/user.rs
```diff
@@ -215,6 +215,7 @@ pub struct GetUserDetailsResponse {
     pub entity_type: EntityType,
     pub theme_id: Option<String>,
     pub version: common_enums::ApiVersion,
+    pub is_active: bool,
```
ThisIsMani: This is not needed

migrations/2026-02-11-120200_add_soft_delete_to_users/up.sql
```diff
@@ -0,0 +1,3 @@
+-- Your SQL goes here
+-- Add is_active column to users table with default value TRUE
```
ThisIsMani: We can remove this.

crates/router/src/db/user.rs
ThisIsMani, line 1: https://github.com/juspay/hyperswitch/blob/a345267863c932a2810b3191fb903a22e7756a57/crates/router/src/db/user.rs#L139-L146

Remove this.

ThisIsMani, line 1: https://github.com/juspay/hyperswitch/blob/a345267863c932a2810b3191fb903a22e7756a57/crates/router/src/db/user.rs#L434-L450

Can't we use into() here?

crates/router/src/routes/user.rs
ThisIsMani, line 1: https://github.com/juspay/hyperswitch/blob/a345267863c932a2810b3191fb903a22e7756a57/crates/router/src/routes/user.rs#L626-L626

This might not be a good name...

crates/router/src/core/user.rs
ThisIsMani, line 1: https://github.com/juspay/hyperswitch/blob/a345267863c932a2810b3191fb903a22e7756a57/crates/router/src/core/user.rs#L999-L999

We can rename this to `handle_new_or_inactive_user_invitation`

crates/router/src/types/domain/user.rs
ThisIsMani, line 1: https://github.com/juspay/hyperswitch/blob/a345267863c932a2810b3191fb903a22e7756a57/crates/router/src/types/domain/user.rs#L858-L858

Cleanup import

ThisIsMani, line 1: https://github.com/juspay/hyperswitch/blob/a345267863c932a2810b3191fb903a22e7756a57/crates/router/src/types/domain/user.rs#L882-L882

check_if_user_is_active_in_db

ThisIsMani, line 1: https://github.com/juspay/hyperswitch/blob/a345267863c932a2810b3191fb903a22e7756a57/crates/router/src/types/domain/user.rs#L872-L872

Shouldn't this be 500. We are already checking this above.

crates/router/src/types/domain/user.rs
```diff
@@ -837,27 +832,60 @@ impl NewUser {
...
+
+        match existing_user {
+            Some(user_from_db) => db
+                .reactivate_user_by_user_id(
+                    &user_from_db.user_id,
+                    diesel_models::user::ReactivateUserUpdate {
+                        new_name: Some(self.get_name().expose()),
+                        new_password: self
+                            .get_password()
+                            .map(|user_password| user_password.get_secret()),
```
ThisIsMani: Should use hashed password.

crates/router/src/db/user.rs
AmanKr-3170: No, There is no conversion happening here.

ThisIsMani (approved):

VenuMadhav2541 (approved):
Analytics LGTM

crates/diesel_models/src/user.rs
```diff
@@ -44,20 +45,22 @@ pub struct UserNew {
...
     last_modified_at: PrimitiveDateTime,
     totp_status: Option<TotpStatus>,
-    totp_secret: Option<Encryption>,
-    totp_recovery_codes: Option<Vec<Secret<String>>>,
-    last_password_modified_at: Option<PrimitiveDateTime>,
-    lineage_context: Option<LineageContext>,
+    totp_secret: Option<Option<Encryption>>,
+    totp_recovery_codes: Option<Option<Vec<Secret<String>>>>,
+    last_password_modified_at: Option<Option<PrimitiveDateTime>>,
+    lineage_context: Option<Option<LineageContext>>,
```
tsdk02, line 62: Why making these `Option<Option<..>>`?

crates/diesel_models/src/user.rs
```diff
@@ -78,6 +81,13 @@ pub enum UserUpdate {
...
         lineage_context: LineageContext,
     },
+    DeactivateUpdate,
+}
+
+#[derive(Debug)]
+pub struct ReactivateUserUpdate {
+    pub new_name: Option<String>,
+    pub new_password: Option<Secret<String>>,
 }
```
tsdk02, line 91: Why cant this be another variant in `UserUpdate` called `ReactivateUpdate`?

crates/diesel_models/src/user.rs
```diff
@@ -141,8 +155,39 @@ impl From<UserUpdate> for UserUpdateInternal {
                 totp_status: None,
                 totp_secret: None,
                 totp_recovery_codes: None,
-                lineage_context: Some(lineage_context),
+                lineage_context: Some(Some(lineage_context)),
+                is_active: None,
             },
+            UserUpdate::DeactivateUpdate => Self {
+                name: None,
+                password: Some(None),
```
tsdk02, line 163: Why is this `Some(None)`? Can it be `None`?

crates/router/src/db/dashboard_metadata.rs
```diff
@@ -344,4 +360,30 @@ impl DashboardMetadataInterface for MockDb {
...
+
+        let initial_len = dashboard_metadata.len();
+
+        dashboard_metadata.retain(|metadata_inner| {
+            !(metadata_inner
+                .user_id
+                .clone()
+                .map(|user_id_inner| user_id_inner == user_id)
+                .unwrap_or(false))
+        });
```
tsdk02:
```suggestion
        dashboard_metadata.retain(|metadata_inner| metadata_inner.user_id.as_deref() != Some(user_id));
```

Check if this can be done

crates/router/src/db/dashboard_metadata.rs
```diff
@@ -344,4 +360,30 @@ impl DashboardMetadataInterface for MockDb {
...
+
+        if dashboard_metadata.len() == initial_len {
+            return Err(errors::StorageError::ValueNotFound(format!(
+                "No user available for user_id = {user_id}"
```
tsdk02:
```suggestion
                "No dashboard metadata found for user_id = {user_id}"
```

Shouldn't the error message be like this?

crates/diesel_models/src/user.rs
AmanKr-3170: To set the values to null, this is required.
The first option tells which fields to update. The second option tells that the value can be nullable. And the value in 2nd option can be None or Some(x). Hence, to set a field to `null` we need Option< None >, which tells it to update the field with None.

**Why was it working earlier?**
Earlier there were no use cases where we would need to set the nullable fields to null. It was always null -> some_value or some_value -> another_value. But now since we have soft delete, we'll need to reset the information and hence will need to set few fields as `null`.

crates/diesel_models/src/user.rs
AmanKr-3170: We have two query functions: 
1. `update_active_by_user_id` (or by_user_email)
2. `reactivate_active_by_user_id`
The first query should never do a `Reactivate` update, since that query targets already `active` users. While the second query should only do `Reactivate` update, ,since that query targets in_active users.
Doing it this way helps us prevent wrong type of update sent to wrong type of user.

crates/diesel_models/src/user.rs
AmanKr-3170: None will tell the `UpdateInternal` to not update the password.
Hence, if password was `abc`, then after update, the password will still remain `abc`. We need Some(None), where Some tells that password needs to be updated, and None tells it to be updated with `null`.

crates/diesel_models/src/user.rs
```diff
@@ -78,6 +81,13 @@ pub enum UserUpdate {
     LineageContextUpdate {
         lineage_context: LineageContext,
     },
+    DeactivateUpdate,
```
hrithikesh026, line 84: this enum variant is stale and doesn't update any field in User table. Why is it here?

AmanKr-3170: This is a single-purpose update which only sets `is_active` to false, hence doesn't need a parameter as well.
Basically deactivating a user.

migrations/2026-02-11-120200_add_soft_delete_to_users/up.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE users ADD COLUMN IF NOT EXISTS is_active BOOLEAN NOT NULL DEFAULT TRUE;
```
hrithikesh026: new columns cannot be introduced as NOT NULL.

migrations/2026-02-11-120200_add_soft_delete_to_users/up.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE users ADD COLUMN IF NOT EXISTS is_active BOOLEAN;
```
ThisIsMani, line 1: Are we not making this enum?

crates/router/src/core/user_role.rs
```diff
@@ -721,9 +724,14 @@ pub async fn delete_user_role(
 
     // If user has no more role associated with him then deleting user
     if remaining_roles.is_empty() {
+        state
+            .store
+            .delete_all_metadata_by_user_id(user_from_db.get_user_id())
+            .await
+            .change_context(UserErrors::InternalServerError)?;
```
ThisIsMani: Should we throw 5xx? Isn't it better to log it and continue.

crates/router/src/core/user_role.rs
```diff
@@ -721,9 +725,18 @@ pub async fn delete_user_role(
...
     // If user has no more role associated with him then deleting user
     if remaining_roles.is_empty() {
+        let delete_metadatas_result = state
+            .store
+            .delete_all_metadata_by_user_id(user_from_db.get_user_id())
+            .await;
+
+        if let Err(e) = delete_metadatas_result {
+            logger::error!("Error while deleting user metadata: {}", e);
+        }
```
ThisIsMani: `inspect_err`

crates/router/src/db/user.rs
```diff
@@ -67,68 +73,82 @@ impl UserInterface for Store {
...
+        user_email: &domain::UserEmail,
+        user_update: storage::UserUpdate,
+    ) -> CustomResult<storage::User, errors::StorageError> {
         let conn = connection::pg_connection_write(self).await?;
-        storage::User::delete_by_user_id(&conn, user_id)
+
+        storage::User::update_active_by_user_email(&conn, user_email.get_inner(), user_update)
             .await
             .map_err(|error| report!(errors::StorageError::from(error)))
     }
```
ThisIsMani, line 133: Add #[instrument(skip_all)]

muditbhutani:
@product-manager what do you think about this implementation. We are now in backend remember

hs-agent:
> **Product Manager Analysis** *(powered by Claude)*

## Analysis: Soft Delete for Users (#11300)

**Core problem:** Hard-deleting users when they lose their last role causes FK constraint violations, destroys audit history, and prevents re-invitation — soft delete fixes all three.

The design is sound. Preserving `user_id` across deactivation/reactivation is the right call for a payment platform where user IDs may be referenced in audit logs, payment events, and external integrations. Here's my detailed assessment of the backend implementation:

### What's Good

1. **Data purge on deactivation** — nulling password, 2FA secrets, recovery codes, and lineage context means "soft delete" doesn't become a compliance liability. The user record is a tombstone, not a data retention risk.
2. **Active-only query methods** — `find_active_user_by_*` as the default path, with the old `find_user_by_*` methods still available when you genuinely need to find any user (e.g., during re-invitation). This is the correct layering.
3. **JWT invalidation via active check** — `authentication.rs` now calls `find_active_user_by_user_id`, so existing tokens for deactivated users immediately fail. No token revocation list needed.
4. **Reactivation on re-invite and signup** — both paths are covered, which is important for UX.

### Concerns and Risks

**1. Migration: `is_active` is `NULLABLE BOOLEAN` — should be `NOT NULL DEFAULT TRUE`**

The `up.sql` adds `is_active BOOLEAN` (nullable, no default). This means:
- All existing users get `NULL`, not `TRUE`
- Every query filtering on `is_active = TRUE` will **exclude all existing users** until backfilled
- This is a **production outage risk** if deployed without a backfill migration

This needs a two-step migration: add column with `DEFAULT TRUE`, then backfill. Or a single `ADD COLUMN is_active BOOLEAN NOT NULL DEFAULT TRUE`. The schema files (`schema.rs`, `schema_v2.rs`) show `Nullable<Bool>`, confirming this gap.

**2. Race condition: role deletion → deactivation is not atomic**

In `user_role.rs`, the flow is: delete metadata → deactivate user. If the process crashes between these steps, you get a user with no metadata but still `is_active = TRUE` and no roles — a zombie state. Consider wrapping this in a transaction, or at minimum making the deactivation idempotent so a retry can clean up.

**3. MockDb divergence on `delete_all_metadata_by_user_id`**

The real implementation returns `Ok(true)` when no metadata exists (catches `NotFound`), but MockDb returns `Err(ValueNotFound(...))`. This inconsistency will cause test failures if anyone tests the "deactivate user with no dashboard metadata" path.

**4. Signup reactivation path needs guardrails**

The alternate step 3 shows a deactivated user can sign up again freely. Consider:
- Should a previously-deactivated user be able to self-signup, or only be re-invited? If an admin removed someone for cause, self-signup is a security gap.
- At minimum, this should be a configurable policy, not a default-open behavior.

**5. No unit tests**

The checklist shows "I added unit tests" is unchecked. For a change touching auth, user lifecycle, and DB schema across 23 files, this is a significant gap. Key test cases needed:
- Deactivate user → verify auth fails
- Deactivate → re-invite → verify fresh state
- Deactivate user with no remaining metadata (edge case)
- Concurrent role deletion race

**6. `analytics.rs` has 15 identical call-site changes**

Every analytics route independently fetches the user with `find_active_user_by_user_id`. This is a maintenance smell — there should be a shared middleware or helper. Not a blocker for this PR, but worth noting for follow-up.

### Scope Boundaries

**v1 (this PR):** Soft delete + reactivation on re-invite — correctly scoped.

**Should wait for v2:**
- Admin audit log of deactivation/reactivation events
- Configurable policy for self-signup of deactivated users
- Scheduled cleanup/hard-delete of tombstone records after retention period (Scheduler job)
- Control Center UI changes to show deactivated users to org admins

### Next Steps

1. **Fix the migration** — `NOT NULL DEFAULT TRUE` is critical before merge. This is a blocker.
2. **Align MockDb behavior** with real DB on the `delete_all_metadata_by_user_id` NotFound case.
3. **Add unit tests** for the deactivation and reactivation flows at minimum.
4. **Decide on self-signup policy** for deactivated users — flag it as an open question for the team.
5. **Consider transactional deactivation** in `user_role.rs` to avoid the metadata-deleted-but-user-still-active zombie state.

AmanKr-3170:
@likhinbopanna @Gnanasundari24 
We have two bug fix PRs for this:
1. https://github.com/juspay/hyperswitch/pull/11469
2. https://github.com/juspay/hyperswitch/pull/11596

If in any production deployment, this PR is being deployed, we'll have to either have the above PRs in same deployment or raise hotfix

PR #7639, opened 2025-03-26
feat(router): add payment method display name to feature matrix response

AkshayaFoiger:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

This PR
1. Adds payment method display name to feature matrix response
2. Replace the 2 digit country codes into 3 digit country codes

## How did you test it?

List the feature matrix and ensure it contains the payment_method_type_dsplay_name field and the country code is 3 digits long. 

```
curl --location 'http://localhost:8080/feature_matrix' \
--header 'Content-Type: application/json' \
--header 'Accept: application/json'
```
Response

```
{
    "connector_count": 28,
    "connectors": [
        {
            "name": "BAMBORA",
            "display_name": "Bambora",
            "description": "Bambora is a leading online payment provider in Canada and United States.",
            "category": "payment_gateway",
            "supported_payment_methods": [
                {
                    "payment_method": "card",
                    "payment_method_type": "credit",
                    "payment_method_type_display_name": "Credit Card",
                    "mandates": "not_supported",
                    "refunds": "supported",
                    "supported_capture_methods": [
                        "automatic",
                        "manual",
                        "sequential_automatic"
                    ],
                    "three_ds": "supported",
                    "no_three_ds": "supported",
                    "supported_card_networks": [
                        "Visa",
                        "Mastercard",
                        "AmericanExpress",
                        "Discover",
                        "JCB",
                        "DinersClub"
                    ],
                    "supported_countries": [
                        "USA",
                        "CAN"
                    ],
                    "supported_currencies": [
                        "USD"
                    ]
                }
            ],
            "supported_webhook_flows": []
        },
        {
            "name": "BAMBORAAPAC",
            "display_name": "Bambora Asia-Pacific",
            "description": "Bambora Asia-Pacific, provides comprehensive payment solutions, offering merchants smart and smooth payment processing capabilities.",
            "category": "payment_gateway",
            "supported_payment_methods": [
                {
                    "payment_method": "card",
                    "payment_method_type": "debit",
                    "payment_method_type_display_name": "Debit Card",
                    "mandates": "supported",
                    "refunds": "supported",
                    "supported_capture_methods": [
                        "automatic",
                        "manual",
                        "sequential_automatic"
                    ],
                    "three_ds": "supported",
                    "no_three_ds": "supported",
                    "supported_card_networks": [
                        "Visa",
[...]

PR #11315, opened 2026-02-18
ci: introduce required spell check for certain critical directories

hrithikesh026:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

Added a new spell check in github CI workflow. 

this enables spell check only on certain directories.

### Additional Changes

## Motivation and Context

The current spell check job scans the entire codebase for typos. But it is not a required check. Hence these typos overflow to sensitive api models. Such typos in api models cannot be corrected later since it would break our api contract. 

So introduced a new spell check job to check for spelling errors in a few critical directories.

## How did you test it?

The CI check passed

## Checklist

### DIFF
--- a/crates/diesel_models/src/payment_methods_session.rs
+++ b/crates/diesel_models/src/payment_methods_session.rs
@@ -45,7 +45,7 @@ impl PaymentMethodSession {
             billing: update_session.billing.or(billing),
             psp_tokenization: update_session.psp_tokenization.or(psp_tokenization),
             network_tokenization: update_session.network_tokenization.or(network_tokenization),
-            tokenization_data: update_session.tokenzation_data.or(tokenization_data),
+            tokenization_data: update_session.tokenization_data.or(tokenization_data),
             expires_at,
             return_url,
             associated_payment_methods,
@@ -62,5 +62,5 @@ pub struct PaymentMethodsSessionUpdateInternal {
     pub billing: Option<common_utils::encryption::Encryption>,
     pub psp_tokenization: Option<common_types::payment_methods::PspTokenization>,
     pub network_tokenization: Option<common_types::payment_methods::NetworkTokenization>,
-    pub tokenzation_data: Option<masking::Secret<serde_json::Value>>,
+    pub tokenization_data: Option<masking::Secret<serde_json::Value>>,
 }


.github/workflows/CI-pr.yml
```diff
@@ -225,6 +225,24 @@ jobs:
       - name: Spell check
         uses: crate-ci/typos@master
 
+  typos-required:
+    name: Spell check specific critical directories
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
```
pixincreate: v6 is the latest one

.github/workflows/CI-pr.yml
```diff
@@ -225,6 +225,24 @@ jobs:
...
+        with:
+          files: |
+            api-reference/
+            migrations/
+            v2_migrations/
+            v2_compatible_migrations/
+            config/
+            docs/
+
   check-v2:
```
pixincreate, line 253: there's a check warning from github. make it readonly, it should fix that

.typos.toml
```diff
@@ -5,6 +5,9 @@ extend-ignore-identifiers-re = [
 ]
 
 [default.extend-identifiers]
+sepa_guarenteed_debit = "sepa_guarenteed_debit" # Would have to break the api contract for fixing this. Not worth the fix.
+sepa_guarenteed_bank_debit = "sepa_guarenteed_bank_debit" # Would have to break the api contract for fixing this. Not worth the fix.
```
pixincreate: may be, we can introduce a new field and then, mark the existing one (that has the typo) has deprecated?

.github/workflows/CI-pr.yml
```diff
@@ -225,6 +225,24 @@ jobs:
       - name: Spell check
         uses: crate-ci/typos@master
```
pixincreate, line 229: i feel, we should mandate the existing typos job. but yeah, the existing long list of typos must be resolved before doing that.

we can actually, pin the spell check version and set up a dependabot to update once a month or so?

what say, @SanchithHegde?

.github/workflows/CI-pr.yml
hrithikesh026: All other check are using v4. Also updating the version of a checkout step is not within the context of this pr. Can be a separate change.

.github/workflows/CI-pr.yml
hrithikesh026: How do I do that?

.typos.toml
hrithikesh026: Having two different strings indicating the same object can be confusing to people.

Anyway we can discuss what needs to be done for this particular case. But that is a different problem to solve.

.github/workflows/CI-pr.yml
pixincreate: add this to the job?

```
permissions:
  contents: read
```

.github/workflows/CI-pr.yml
pixincreate: since this step is being introduced in this pr, choosing the latest stable major version seems appropriate. otherwise we're intentionally adding a new workflow with an older version

if there's a specific reason to stick to v4, happy to understand it

.typos.toml
pixincreate: i agree that keeping two permanent strings would get confusing.
but what i'm suggesting is just temporary dual support, with a clear deprecation plan and removal in the next major version. since the api is versioned, that gives us a clean way to fix the typo without carrying the duplication long term.

crates/common_enums/src/enums.rs
SanchithHegde, line 2407: Where is this being called? I just want to confirm that this won't affect the API contract or database data in any manner.

hrithikesh026: this was introduced in this pr https://github.com/juspay/hyperswitch/pull/7639.

this is a feature matrix endpoint to list all features

crates/diesel_models/src/payment_methods_session.rs
```diff
@@ -62,5 +62,5 @@ pub struct PaymentMethodsSessionUpdateInternal {
     pub billing: Option<common_utils::encryption::Encryption>,
     pub psp_tokenization: Option<common_types::payment_methods::PspTokenization>,
     pub network_tokenization: Option<common_types::payment_methods::NetworkTokenization>,
-    pub tokenzation_data: Option<masking::Secret<serde_json::Value>>,
```
SanchithHegde, line 65: Was this not working correctly so far? Or was this struct not being used actively?

.typos.toml
```diff
@@ -5,6 +5,29 @@ extend-ignore-identifiers-re = [
...
+SiSaKet = "SiSaKet" # thailand state abbreviation
+Tak = "Tak" # thailand state abbreviation
+AGS = "AGS" # philippines state abbreviation
+AgusanDelSur = "AgusanDelSur" # philippines state abbreviation
+CamarinesSur = "CamarinesSur" # philippines state abbreviation
+DavaoDelSur = "DavaoDelSur" # philippines state abbreviation
+IlocosSur = "IlocosSur" # philippines state abbreviation
+LanaoDelSur = "LanaoDelSur" # philippines state abbreviation
+MaguindanaoDelSur = "MaguindanaoDelSur" # philippines state abbreviation
+SurigaoDelSur = "SurigaoDelSur" # philippines state abbreviation
```
SanchithHegde, line 25: Nit: These are not _abbreviations_, can remove the "abbreviations" word. Same for "SiSaKet".

hrithikesh026: it is mentioned in code that these are abbreviations.

crates/diesel_models/src/payment_methods_session.rs
hrithikesh026: I don't think this struct is used anywhere

Sakilmostak (approved):

Sakilmostak:
Payout and payment-method changes looks good

crates/common_enums/src/enums.rs
SanchithHegde: Could this change possibly cause backward compatibility issues or no? Can we please check this?

hrithikesh026: @AkshayaFoiger Can you please let us know if this would break any api contract?

AkshayaFoiger: This change will not break any API contracts. It is only used in the feature matrix to display the payment method type names

PR #11454, opened 2026-03-09
feat(superposition): Add support for create override macro to superposition

srujanchikke:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

Add support for create override in superposition.

- Current open feature SDK doesn't support write operations to superposition
- Use superposition_sdk to create override during merchant create(Account updater flows).
- Implement macro functions for writable configs.

### Additional Changes

## How did you test it?

Testcase 1: Create merchant account, which should create fingerprint secret override for merchant in superposition. 
```
--header 'Accept-Language: en' \
--data-raw '{
  "merchant_id": "merchant_1773037813",
  "locker_id": "m0010",
  "merchant_details": {
    "website": "www.example.com",
    "address": {
      "line1": "1467",
      "line2": "Harrison Street",
      "line3": "Harrison Street",
      "city": "San Fransico",
      "state": "California",
      "zip": "94122",
      "country": "US"
    }
  },
  "webhook_details": {
    "webhook_url" : "https://webhook.site/85f781d5-1ce3-4d6b-b1d3-bddd9ceb2e06",
    
    "webhook_version": "1.0.1",
  },
  "metadata": {
    "account_name" : "transaction_processing",
    "city": "NY",
    "unit": "245"
  }
}'
```
created superposition override for fingerprint secret 

Screenshot 2026-03-09 at 12 08 09 PM

Migrations : 
this default configs  and data has been migrated in lower environment.

## Checklist


crates/external_services/src/superposition.rs
```diff
@@ -300,6 +330,73 @@ impl SuperpositionClient {
...
+    /// # Type Parameters
+    /// * `T` - A type implementing `WritableConfig` that defines the config key and input type
+    ///
+    /// # Arguments
+    /// * `value` - The value to write
+    /// * `context` - The context (dimensions) for this config
+    ///
+    /// # Returns
+    /// * `CustomResult<(), SuperpositionError>` - Success or error
+    pub async fn set_config_value<T: WritableConfig>(
```
jarnura, line 351: what is the write use for superposition, from otlp this should be always read

jarnura (changes requested):

crates/external_services/src/superposition.rs
srujanchikke: merchant create is under olap

jarnura (dismissed):

crates/external_services/src/superposition/types.rs
```diff
@@ -2,12 +2,43 @@
...
+
+impl ToDocument for i64 {
+    fn to_document(&self) -> Document {
+        // u64::try_from(i64) has exactly one failure reason: the value is negative.
+        // Any non-negative i64 always fits in u64 (i64::MAX < u64::MAX), so Err(_)
+        // here means "was negative" — use NegInt. No overflow or precision loss possible.
+        match u64::try_from(*self) {
+            Ok(n) => Document::Number(aws_smithy_types::Number::PosInt(n)),
+            Err(_) => Document::Number(aws_smithy_types::Number::NegInt(*self)),
+        }
```
SanchithHegde: Do we necessarily need the `try_from()` to figure out if a value is positive or negative? Can't we use [`unsigned_abs()`](https://doc.rust-lang.org/std/primitive.i64.html#method.unsigned_abs) with something like this?


```suggestion
        match *self {
            n if n >= 0 => Document::Number(aws_smithy_types::Number::PosInt(n.unsigned_abs())),
            n => Document::Number(aws_smithy_types::Number::NegInt(n)),
        }
```

crates/external_services/src/superposition.rs
```diff
@@ -300,6 +330,73 @@ impl SuperpositionClient {
...
+        for (key, val) in &context.values {
+            builder = builder.context(key, Document::String(val.clone()));
+        }
+
+        let change_reason = generate_change_reason(T::SUPERPOSITION_KEY);
+
+        builder = builder
+            .r#override(T::SUPERPOSITION_KEY, value.to_document())
+            .change_reason(change_reason)
+            .description(format!("Config update for {}", T::SUPERPOSITION_KEY));
```
SanchithHegde: Nit: Do we want to also indicate which service has updated the config (for audit purposes), or would it be available by some other means on the superposition dashboard?

crates/router/src/core/configs/dimension_config.rs
```diff
@@ -1,9 +1,50 @@
...
+                type Input = $input;
+                const SUPERPOSITION_KEY: &'static str = superposition_consts::$key;
+            }
+
+            impl $requirement {
+                pub async fn [<set_ $key:lower>](
+                    &self,
+                    superposition_client: &superposition::SuperpositionClient,
+                    value: &$input,
+                ) -> CustomResult<(), superposition::SuperpositionError> {
```
SanchithHegde, line 33: Any specific reason this isn't a method on the client instead, which would accept the dimensions / context and the value as parameters?

Something like:

```rust
impl SuperpositionClient {
    fn set_fingerprint_secret(&self, dimensions: ..., value: ...) {
        // ...
    }
}
```

That way, you would be calling a method _on_ the client (the entity doing some action), and you pass it the data required to perform that action (the dimensions and value here).

scripts/seed_superposition.sh
```diff
@@ -14,21 +14,60 @@ echo "Workspace: $WORKSPACE_ID, Org: $ORG_ID"
...
-    if curl -s "$SUPERPOSITION_URL/health" > /dev/null 2>&1; then
+READY=0
+for i in {1..60}; do
+    if curl -sS -o /dev/null "$SUPERPOSITION_URL/health"; then
         echo "Superposition is ready!"
+        READY=1
         break
     fi
-    echo "Waiting for Superposition... ($i/30)"
+    echo "Waiting for Superposition... ($i/60)"
```
SanchithHegde: Nit: Can have number of retries defined as a variable as well. Optional change.

crates/router/src/core/configs/dimension_config.rs
srujanchikke: The current design isn't trying to avoid methods on the client entirely — it has both. The generic action lives on the client; the per-key wrappers sit alongside their per-key config metadata, and the dimension requirements get enforced through the receiver type.

Take `set_fingerprint_secret`: it's defined on `impl DimensionsWithProcessorAndProviderMerchantId`. The compiler simply won't let you call it on a Dimensions value that doesn't carry processor + provider merchant id. If you moved it to the client as client.set_fingerprint_secret(dimensions, value), you'd need to re-encode that requirement as a parameter type bound — same enforcement in theory, but the call site now reads as "any dimensions will do" until you dig into the signature. The illegal state isn't unrepresentable anymore; it's just documented somewhere the caller might not look.

tsdk02 (dismissed):
Dashboard changes look fine

crates/external_services/src/superposition.rs
```diff
@@ -115,14 +121,30 @@ impl GetValue<serde_json::Value> for open_feature::Client {
...
 }
 
 impl SuperpositionClient {
-    /// Create a new Superposition client
-    pub async fn new(config: SuperpositionClientConfig) -> CustomResult<Self, SuperpositionError> {
+    /// Create a new Superposition client.
+    ///
+    /// `service_name` is recorded on every write so the Superposition audit trail
+    /// can attribute config changes to the originating service. Callers should pass
+    /// `env!("CARGO_PKG_NAME")` from the binary/service so it's locked at compile time.
```
SanchithHegde: Nit, optional change:

We can suggest to preferably use `CARGO_BIN_NAME` if available, or else use `CARGO_PKG_NAME`. If the `CARGO_PKG_NAME` environment variable is read within a library crate, it'll include the value of the library crate, which may not help with audit logs.

crates/router/src/bin/router.rs
```diff
@@ -27,7 +27,13 @@ async fn main() -> ApplicationResult<()> {
...
         &conf.log,
         router_env::service_name!(),
-        [router_env::service_name!(), "actix_server"],
+        [
+            router_env::service_name!(),
+            "actix_server",
+            "open_feature",
+            "superposition_provider",
+            "superposition_sdk",
+        ],
```
SanchithHegde, line 36: Please replicate this change in `crates/router/src/bin/scheduler.rs` and `crates/drainer/src/main.rs` as well.

crates/router/build.rs
```diff
@@ -2,7 +2,7 @@ fn main() {
     // Set thread stack size to 11 MiB for debug builds
     // Reference: https://doc.rust-lang.org/std/thread/#stack-size
     #[cfg(debug_assertions)]
-    println!("cargo:rustc-env=RUST_MIN_STACK=11534336"); // 11 * 1024 * 1024 = 11 MiB
+    println!("cargo:rustc-env=RUST_MIN_STACK=19534336"); // 11 * 1024 * 1024 = 11 MiB
```
SanchithHegde: Nit: Can you please update the comment to match the updated size? Also, did you try with a value like 12 MiB, 16 MiB, etc. or did you blindly update the second digit to a higher value?

crates/router/src/routes/app.rs
```diff
@@ -512,7 +512,7 @@ impl AppState {
             let superposition_service = conf
                 .superposition
                 .get_inner()
-                .get_superposition_client()
+                .get_superposition_client(env!("CARGO_PKG_NAME"))
```
SanchithHegde: Nit: The concern with reading `CARGO_PKG_NAME` here is that the value of service name for the `scheduler` binary would be `router` (the name of this crate) and not `scheduler`.

SanchithHegde (dismissed):

tsdk02 (approved):

jarnura (approved):

SanchithHegde (approved):

srujanchikke:
@prajjwalkumar17 Previously fingerprint secret was stored in configs table and if it is not available we were creating new secret for merchant. I refactored this logic to move FS to superposition, Now if blocklist is enabled and for some reason FS is not available or empty we throw 5xx for all payments(since we cannot allow this blocked cards if block list is enabled). We will ensure data migrations before deploying the current change, Can you please confirm the behaviour of blocklist ? 
cc : @bernard-eugine

PR #11469, opened 2026-03-10
fix(users): prevent password requirement bypass on reactivated user signin

Grovy-3170:
## Description

Due to Issues with [Soft Delete User PR](https://github.com/juspay/hyperswitch/pull/11300) when a deleted user was re-invited and attempted to use the invitation link to sign in, they were not prompted to set a password and could directly log in. However, after logging out, they were unable to log back in due to missing password credentials.

## Root Cause

The reactivation logic unconditionally updated the `last_password_modified_at` field even when no password was being set. This caused the `decision_manager` to incorrectly assume the password was recently modified and skip the password rotation requirement.

## Changes

This PR ensures that:
- `last_password_modified_at` is only set when a password is actually provided during reactivation
- `is_password_rotate_required()` performs an early check to return `true` if no password exists
- The `ReactivateUserUpdate` struct now accepts an optional `last_password_modified_at` field for explicit control

## Type of Change
- [x] Bug fix (non-breaking change which fixes an issue)

### When a user is invited and has accepted:

Screenshot 2026-03-10 at 10 09 24 PM

### When the user is deleted with no other role (soft deleted)

Screenshot 2026-03-10 at 10 09 45 PM

### When user is re-invited, but hasn't accepted yet

Screenshot 2026-03-10 at 10 10 32 PM

### When the user tries to accept the invitation

Screenshot 2026-03-10 at 10 11 52 PM

DB:

Screenshot 2026-03-10 at 10 12 43 PM

### DIFF
--- a/context7.json
+++ b/context7.json
@@ -1,0 +1,4 @@
+{
+  "url": "https://context7.com/juspay/hyperswitch",
+  "public_key": "pk_ah3JPcqcUryuNnIrmC83D"
+}
--- a/crates/hyperswitch_connectors/src/connectors/adyen/transformers.rs
+++ b/crates/hyperswitch_connectors/src/connectors/adyen/transformers.rs
@@ -5288,10 +5288,10 @@ impl<F, Req>
         Ok(Self {
             status: adyen_payments_response_data.status,
             amount_captured: minor_amount_captured.map(|amount| amount.get_amount_as_i64()),
-            response: adyen_payments_response_data.error.map_or_else(
-                || Ok(adyen_payments_response_data.payments_response_data),
-                Err,
-            ),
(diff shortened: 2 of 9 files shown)


crates/router/src/types/domain/user.rs
```diff
@@ -861,11 +861,15 @@ impl NewUser {
                         password::generate_password_hash(user_password.get_secret())
                     })
                     .transpose()?;
+                let last_password_modified_at = hashed_password
+                    .as_ref()
+                    .map(|_| common_utils::date_time::now());
```
ThisIsMani: Use `is_some().then(|| {})`.

Issue #4408, opened 2024-04-22
update refunds filters and list


apoorvdixit88:
Include filter for amount and merchant connector id in refunds list.

PR #11548, opened 2026-03-18
feat(analytics): add split payment dimension for payment intent analytics

ThisIsMani:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [x] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

This PR adds `is_split_payment` as a boolean analytics dimension across the payment intent analytics pipeline so merchants can filter and group existing payment-intent metrics by split vs non-split flows.

Implemented changes:
- ClickHouse schema wiring in `payment_intents_queue`, `payment_intents`, and `payment_intents_mv`
- Kafka payload derivation for v1/v2 payment intent events via `split_payments.is_some()`
- API model updates for `PaymentIntentDimensions`, `PaymentIntentFilters`, and `PaymentIntentMetricsBucketIdentifier`
- Analytics query/core wiring (filter clause, filter/metric rows, filter mapping, dimension listing)
- Metric/sessionized metric bucket constructor propagation to include `is_split_payment`
- SQLx `FromRow` plumbing updates for the new row field

### Additional Changes

## Motivation and Context

Merchants using platform/split payment flows need visibility into split-payment performance in analytics. The source field (`split_payments`) existed in domain models but was not propagated end-to-end into analytics dimensions.

Reference: Closes [#4408](https://github.com/juspay/hyperswitch-control-center/issues/4408).

## How did you test it?

todo!()

## Checklist

### DIFF
--- a/crates/router/src/services/kafka/payment_intent.rs
+++ b/crates/router/src/services/kafka/payment_intent.rs
@@ -1,10 +1,10 @@
 #[cfg(feature = "v2")]
-use common_enums::RequestIncrementalAuthorization;
-#[cfg(feature = "v2")]
 use ::common_types::{
     payments,
     primitive_wrappers::{EnablePartialAuthorizationBool, RequestExtendedAuthorizationBool},
 };
+#[cfg(feature = "v2")]
+use common_enums::RequestIncrementalAuthorization;
 use common_utils::{
     crypto::Encryptable, hashing::HashedString, id_type, pii, types as common_types,
 };
--- a/crates/router/src/services/kafka/payment_intent_event.rs
+++ b/crates/router/src/services/kafka/payment_intent_event.rs
@@ -1,10 +1,10 @@
 #[cfg(feature = "v2")]
-use common_enums::{self, RequestIncrementalAuthorization};
-#[cfg(feature = "v2")]
 use ::common_types::{
     payments,
     primitive_wrappers::{EnablePartialAuthorizationBool, RequestExtendedAuthorizationBool},
 };
+#[cfg(feature = "v2")]
+use common_enums::{self, RequestIncrementalAuthorization};
 use common_utils::{
     crypto::Encryptable, hashing::HashedString, id_type, pii, types as common_types,
 };


hs-agent:
## Code Review: `feat(analytics): add split payment dimension for payment intent analytics`

Good work on the end-to-end wiring — the PR correctly propagates the `is_split_payment` dimension through all the layers (Kafka, ClickHouse, API models, query builders, SQLx). However, there are several issues that should be addressed before merging.

---

### CRITICAL — Bool-to-String Coercion in SQL Filters

**Files:** `crates/analytics/src/payment_intents/core.rs`, `crates/analytics/src/payment_intents/types.rs`

In `core.rs`:
```rust
PaymentIntentDimensions::IsSplitPayment => fil.is_split_payment.map(|i| i.to_string()),
```

And in `types.rs`:
```rust
builder
    .add_filter_in_range_clause(
        PaymentIntentDimensions::IsSplitPayment,
        &self.is_split_payment,   // Vec<bool>
    )
```

`bool::to_string()` produces `"true"`/`"false"`, but ClickHouse `Bool` columns store `1`/`0` internally. If `add_filter_in_range_clause` generates quoted string literals like `IN ('true', 'false')`, this will **silently return zero rows** against a `Bool` column unless there is implicit coercion. All other callers pass `&Vec<String>` for string-typed dimensions — passing `&Vec<bool>` is a type mismatch at the SQL semantics level even if it compiles.

**Action:** Verify that `add_filter_in_range_clause` generates SQL compatible with ClickHouse `Bool` columns, or map `true -> 1` / `false -> 0` explicitly.

---

### CRITICAL — No Tests

The PR checklist explicitly states `todo!()` for testing. At minimum, please add:
- A unit test for `PaymentIntentMetricsBucketIdentifier::new()` with `is_split_payment`
- Integration test coverage for the filter SQL generation path to verify the generated query is valid against a `Bool` column

---

### HIGH — ClickHouse MV Column Ordering Must Be Verified

**File:** `crates/analytics/docs/clickhouse/scripts/payment_intents.sql`

ClickHouse materialized views match columns by **position** in the `INSERT ... SELECT`, not by name. The new `is_split_payment` column is added between `organization_id` and `processor_merchant_id` in all three objects (queue, table, MV). Please double-check that the positional ordering is identical across all three DDL statements after this change — any mismatch will cause **silent data misalignment** in production.

---

### HIGH — Verify `PartialEq`/`Hash` Contract for `PaymentIntentMetricsBucketIdentifier`

**File:** `crates/api_models/src/analytics/payment_intents.rs`

The `Hash` impl is manually written and correctly includes `is_split_payment`. However, if `PartialEq` is also manually implemented (not derived), it **must** also include `is_split_payment`. Objects that compare equal must hash equal — violating this contract causes duplicate or dropped metric rows in `HashMap`-based bucketing.

---

### HIGH — `Vec<bool>` API Filter Type

**File:** `crates/api_models/src/analytics/payment_intents.rs`

```rust
pub is_split_payment: Vec<bool>,
```

For a boolean dimension, `Vec<bool>` is semantically odd — the valid set is `{true, false}`, so a `Vec<bool>` of length 2 is equivalent to no filter at all. While the existing pattern for multi-value dimensions uses `Vec<T>`, consider whether `Option<bool>` better represents the intent. If you want to keep `Vec<bool>` for consistency with the codebase pattern, that's acceptable but worth a comment explaining the rationale.

---

### HIGH — Ensure `IsSplitPayment` Has Display/Column Name Mapping

The diff adds the `IsSplitPayment` variant to `PaymentIntentDimensions` but doesn't show updates to any `Display` impl, `strum` attribute, or exhaustive match that maps dimension variants to SQL column name strings. If there's a `#[strum(serialize = "is_split_payment")]` derive or similar — great. If there's a manual `match` arm that converts dimensions to column names for query building, please confirm it covers `IsSplitPayment -> "is_split_payment"`.

---

### MEDIUM — Formatting & Linting

Author confirmed `cargo +nightly fmt --all` and `cargo clippy` were not run. These will likely fail CI. Please run both before requesting re-review.

---

### Positive Notes
- The Kafka derivation via `split_payments.is_some()` is clean and correct
- The `DEFAULT false` on ClickHouse columns handles backfill gracefully for historical data
- The SQLx `FromRow` impl follows the established `ColumnNotFound` fallback pattern
- Trailing newline fix in the SQL file is a nice cleanup

PR #11553, opened 2026-03-19
feat(core/connector): [SANTANDER] Add & Implement PushNotification & GenerateQr Connector Integration Flow 

bsayak03:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

- Added a new flow type `PushNotification` & `GenerateQr`
- The above two flows serves the purpose of a post-processing step wherein after the Authorize/SetupMandate Flow, either of `PushNotification` or `GenerateQr`is called. 
- The `PushNotification` flow makes a connector call to send a Push notification to the customer's banking app (requesting them to approve or reject the mandate) [Journey 1 of Pix Automatico]
- The `GenerateQr` flow serves the purpose of making a GET call and receiving the QR Code Data from the connector [Journey 2/3/4 of Pix Automatico]
- Integrated the `solicRec` api for `Journey 1 of Pix Automatico` in the `PushNotification` Flow in Santander
- Integrated the `Consult & Activate Journey` api for `Journey 2/3/4 of Pix Automatico` in the `GenerateQr` Flow in Santander

### Additional Changes

- [ ] This PR modifies the API contract
- [ ] This PR modifies the database schema
- [ ] This PR modifies application configuration/environment variables

## How did you test it?

For test cases refer to this [PR](https://github.com/juspay/hyperswitch/pull/11589)

ran `just clippy` & `just clippy_v2`

Screenshot 2026-04-02 at 11 38 54

## Checklist

- [x] I formatted the code `cargo +nightly fmt --all`
- [x] I addressed lints thrown by `cargo clippy`
- [x] I reviewed the submitted code
- [x] I added unit tests for my changes where possible

PR #11569, opened 2026-03-21
feat(core): Add `CurrentFlowInfo` in `AccessTokenRequestData` and moved `get_access_token_key` to `ConnectorSpecifications`

bsayak03:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

In Journey 3/4 of Pix Automatico in Santander, we are calling `payments_operation_core` twice. The first time with `authorize_flow.rs` [Leg 1] and the second time with `setup_mandate_flow.rs`[Leg 2]. This is because the access token required in Leg 1 is different from the one required in Leg 2. This is because both the access token endpoints are different even though the api credentials are different. 

Hence,`AccessTokenAuth` connector flow should know when Leg 1 is being executed and when Leg 2 is being executed so that it can route them to different AccessToken endpoints. Therefore, we had to add `CurrentFlowInfo` in `AccessTokenRequestData`. Also the access token key should be different for Leg 1 and Leg 2, hence a match case based on `CurrentFlowInfo` was also added in the `get_access_token_key` fn inside the `ConnectorSpecifications` trait implementation. 

Also refractored the `get_access_token_key` from the `ConnectorAccessTokenSuffix` to `ConnectorSpecifications` since the trait fn's of `ConnectorSpecifications` already had `CurrentFlowInfo` being passed to its fn parameters.

### Additional Changes

## How did you test it?

Refer to [PR](https://github.com/juspay/hyperswitch/pull/11589) for Test cases

## Checklist


crates/router/src/connector/utils.rs
```diff
@@ -998,6 +1003,12 @@ impl PaymentMethodTokenizationRequestData for types::PaymentMethodTokenizationDa
     }
 }
 
+impl PaymentsAccessTokenRequestData for types::AccessTokenRequestData {
```
hrithikesh026: what is the need for this trait?
This can be a impl function right?

crates/hyperswitch_domain_models/src/router_request_types.rs
```diff
@@ -31,6 +32,36 @@ use crate::{
     router_flow_types as flows, router_response_types as response_types,
     vault::PaymentMethodCustomVaultingData,
 };
+
+/// Current flow information passed to the connector specifications trait
+/// In order to make some decision about the preprocessing or alternate flow
+#[derive(Clone, Debug, Serialize)]
+pub enum CurrentFlowInfo {
```
hrithikesh026, line 39: this enum is already defined in `hyperswitch_interfaces` crate

crates/hyperswitch_domain_models/src/router_request_types.rs
```diff
@@ -1370,6 +1401,8 @@ pub struct AccessTokenRequestData {
     pub app_id: Secret<String>,
     pub id: Option<Secret<String>>,
     pub authentication_token: Option<AccessTokenAuthenticationResponse>,
+    pub current_flow: Option<CurrentFlowInfo>,
```
hrithikesh026, line 1404: why is current_flow needed in PaymentsAuthorizeData

crates/hyperswitch_connectors/src/connectors/santander.rs
```diff
@@ -494,24 +498,60 @@ impl ConnectorIntegration<AccessTokenAuth, AccessTokenRequestData, AccessToken>
...
+                            // Journey 3 & 4 CIT - Leg 2 & Journey 2
+                            Some(CurrentFlowInfo::SetupMandate { .. }) => Ok(format!(
+                                "{}auth/oauth/v2/token",
+                                connectors.santander.base_url
+                            )),
+                            _ => Err(errors::ConnectorError::NotSupported {
+                                message: req.payment_method.to_string(),
+                                connector: "Santander",
+                            }
+                            .into()),
```
hrithikesh026: CurrentFlowInfo should not overflow into connector request building logic. 
You can either create a brand new flow for the new AccessToken related flow. Or decide the endpoint based on some other parameter.

bsayak03: why create another access token flow just to differentiate two enpoints when we can take help of CurrentFlowInfo?
There is no other parameter we can decide the endpoint upon

crates/router/src/connector/utils.rs
bsayak03: to maintain consistency with codebase pattern, you would see similar thing done for `PaymentsAuthorizeData`, `PaymentMethodTokenizationData`

crates/hyperswitch_domain_models/src/router_request_types.rs
bsayak03: migrated it to domain models crate because connectors crate would be importing it, keeping it in hyperswitch interfaces crate would create a cyclic dependency

crates/hyperswitch_connectors/src/connectors/santander.rs
hrithikesh026: CurrentFlowInfo is meant for deciding what subflow needs to be called for a given connector, flow and other attributes, Its purpose is limited to ConnectorSpecification Trait. 
In this case, it seems that you need a flag to indicate what connector endpoint to hit

crates/hyperswitch_domain_models/src/router_request_types.rs
hrithikesh026: No. `hyperswitch_connector` crate has  `hyperswitch_interfaces` as dependency. CurrentFlowInfo is accessible in `hyperswitch_connectors` crate.

Also if you moved it to `hyperswitch_domain_models`, it is still available in `hyperswitch_interfaces`

crates/router/src/connector/utils.rs
hrithikesh026: That is not a pattern, it is done for some other reason, If you just need a behaviour for a fixed Type, Just introduce the function in impl block.

If you need a behaviour for many types, have a trait implementation.

crates/hyperswitch_domain_models/src/router_request_types.rs
bsayak03: my bad, i would be importing `CurrentFlowInfo` in `crates/hyperswitch_domain_models/src/router_request_types.rs` because i added it in `AccessTokenRequestData`. But domain models crate doesn't have hyperswitch interfaces crate as a dependency because interfaces crate depends on domain models, hence would create cyclic dependency

crates/hyperswitch_connectors/src/connectors/santander.rs
bsayak03: should i be passing down a flag in AccessTokenRequestData then?
while creation of AccessTokenRequestData in `add_access_token` fn, i can pass a flag depending upon a check based on `CurrentFlowInfo`

crates/router/src/connector/utils.rs
bsayak03: but then i would have to write the impl in the crate where the struct is declared (which is domain models) due to rust's orphan rule. However, all the connector util fn's are written in the utils of hyperswitch_connectors crate.

crates/hyperswitch_domain_models/src/router_request_types.rs
bsayak03: Its in AccessTokenRequest Data

crates/hyperswitch_connectors/src/connectors/santander.rs
deepanshu-iiitu: Add meaningful comments

bsayak03: there will be a lot of references of `Journey` term in the santander connector code, hence if i go on to add comments explaining journey, it will be a big comment. Hence described everything in the PR des @deepanshu-iiitu

crates/router/src/services/email/types.rs
```diff
@@ -760,8 +760,12 @@ impl EmailData for RoleDeleted {
...
         }
         .into_iter()
-        .map(|item| format!("<li>{}</li>", item))
-        .collect::<String>();
+        .fold(String::new(), |mut acc, item| {
+            acc.push_str("<li>");
+            acc.push_str(&item);
+            acc.push_str("</li>");
+            acc
+        });
```
ThisIsMani, line 768: This should not have been changed. Approving this for now since it doesn't change any functionality. Please raise a follow up PR to revert this change.

bsayak03: picked in [PR](https://github.com/juspay/hyperswitch/pull/11553)

PR #11638, opened 2026-03-27
ci(cypress): Add test for external vault

Gnanasundari24:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description
Adds a new Cypress test spec 42-ExternalVault.cy.js to cover the External Vault (VGS) connector feature.

### Additional Changes

## How did you test it?
Screenshot 2026-03-27 at 5 46 00 PM

## Checklist


cypress-tests/cypress/e2e/spec/Payment/40-ExternalVault.cy.js
```diff
@@ -0,0 +1,417 @@
+import * as fixtures from "../../../fixtures/imports";
+import State from "../../../utils/State";
+import getConnectorDetails, * as utils from "../../configs/Payment/Utils";
+
+let globalState;
+
+const VGS_CONNECTOR_NAME = "vgs";
```
likhinbopanna, line 7: Don't we have any other option than hardcoding , asking because in future we might use other vault connector , please verify this change before merging

cypress-tests/cypress/e2e/spec/Payment/42-ExternalVault.cy.js
```diff
@@ -0,0 +1,417 @@
...
+    }
+  );
+
+  // ── 5. Teardown: Remove VGS from business profile ────────────────────────
+
+  context(
+    "5. Teardown - remove VGS external vault from business profile",
+    () => {
+      it("Remove external_vault_connector_id from business profile", () => {
+        cy.request({
```
likhinbopanna: Can we use this from commands.js as this is spec file

cypress-tests/cypress/support/commands.js
```diff
@@ -1128,6 +1128,90 @@ Cypress.Commands.add(
   }
 );
 
+Cypress.Commands.add(
```
likhinbopanna, line 1132: Can we use re-use from payment_processor by overriding the processor_type ? any specific reason to use separate function ?

cypress-tests/cypress/e2e/spec/Payment/40-ExternalVault.cy.js
Gnanasundari24: This way, if we need to create another connector, we can simply change the name and reuse it. Otherwise, we would have to rely on a global variable.

What do you suggest?

cypress-tests/cypress/support/commands.js
Gnanasundari24: For the payment processor, we’ve implemented it to use the connector_id from the environment variable. However, for the external vault, the connector_id and connector_name are different. That’s why I created a new function.

cypress-tests/cypress/e2e/spec/Payment/40-ExternalVault.cy.js
likhinbopanna: lets go with this for now

cypress-tests/cypress/e2e/spec/Payment/40-ExternalVault.cy.js
```diff
@@ -0,0 +1,398 @@
...
+
+      cy.UpdateBusinessProfileTest(
+        updateBusinessProfileBody,
+        false,
+        false,
+        false,
+        false,
+        false,
+        globalState
+      );
```
pixincreate, line 50: add a comment or something explaining which field is which

cypress-tests/cypress/e2e/spec/Payment/40-ExternalVault.cy.js
```diff
@@ -0,0 +1,398 @@
...
+      };
+
+      cy.UpdateBusinessProfileTest(
+        updateBusinessProfileBody,
+        false,
+        false,
+        false,
+        false,
+        false,
+        globalState
```
pixincreate, line 394: same here

PR #11703, opened 2026-04-05
docs(router): add descriptive comment for initialization

crazy-develop:
Summary: This PR adds a descriptive comment at the beginning of the main() function in crates/router/src/bin/router.rs.

Why is this change necessary? The main entry point of the router is a critical part of the codebase. Adding a clear initialization comment makes it easier for new contributors to understand that the application bootup process is starting from this point. It enhances code readability with zero impact on runtime logic.

Changes:

Added // Initialize the Hyperswitch application comment in crates/router/src/bin/router.rs.


riyasaini-svg:
good sir

PR #11751, opened 2026-04-08
fix(cypress): Cypress for WorldpayWPG

Vani-1107:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

Cypress for WorldpayWPG
Note:The 3DS flow in Cypress has been skipped.
With the introduction of Device Data Collection (DDC) in Worldpay WPG, the flow now runs inside a hidden iframe. As part of this, backend logic was added to detect whether the page is running inside an iframe (window.self !== window.parent). If true, it sends a postMessage to the parent window with the redirect URL instead of performing a direct browser redirect.
Since Cypress runs the application inside an iframe, this condition is always true in the test environment. As a result, instead of performing a normal redirect to the return URL, the flow sends a postMessage, which Cypress does not automatically act upon. This prevents the application from navigating to the expected return URL during tests.
Hence the 3DS flow in cypress has been skipped.

### Additional Changes

## How did you test it?

Screenshot 2026-04-08 at 10 30 06 AM

## Checklist


cypress-tests/cypress/e2e/configs/Payment/Worldpayxml.js
```diff
@@ -66,6 +66,9 @@ export const connectorDetails = {
       },
     },
     "3DSManualCapture": getCustomExchange({
+      Configs: {
+        TRIGGER_SKIP: true,
```
Gnanasundari24, line 70: Why are we skipping the test?

likhinbopanna (approved):

Gnanasundari24 (approved):

PR #11780, opened 2026-04-12
fix(connector): fix UCS webhook config and update connector configs for Truelayer and Trustly

Vani-1107:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

This PR includes the following changes:

- Removed Truelayer and Trustly from the UCS-only connector list, as both connectors now have payouts implemented in Hyperswitch while payins are handled via UCS.

- Added Trustly to ucs_psync_disabled_connectors since it does not support the PSync flow.

- Updated the webhook configuration for UCS:
Previously: `ucs_rollout_config_{merchant_id}_{connector_name}_Webhooks`
Now: `ucs_rollout_config_{merchant_id}_{mca_id}_Webhooks`
This change addresses an issue where, for connectors with payins in UCS and payouts in Hyperswitch, enabling webhooks at the connector level caused payout webhooks to be incorrectly routed to UCS. With the new configuration, webhook routing is scoped more precisely, ensuring payout webhooks are handled by Hyperswitch as intended.

### Additional Changes

## How did you test it?

1. Enabled both payins and payouts for the same merchant
2. Enabled UCS config specifically for payins
3. Performed a payment → verified that the webhook is correctly handled by UCS
4. Performed a payout → verified that the webhook is correctly routed to Hyperswitch

## Checklist

### DIFF
--- a/config/deployments/env_specific.toml
+++ b/config/deployments/env_specific.toml
@@ -406,8 +406,8 @@ connector_names = "connector_names"     # Comma-separated list of allowed connec
 [grpc_client.unified_connector_service]
 base_url = "http://localhost:8000"      # Unified Connector Service Base URL
 connection_timeout = 10                 # Connection Timeout Duration in Seconds
-ucs_only_connectors = "paytm, phonepe, hyperpg, revolv3, truelayer, fiservcommercehub, trustly"    # Comma-separated list of connectors that use UCS only
-ucs_psync_disabled_connectors = "cashtocode"    # Comma-separated list of connectors to disable UCS PSync call
+ucs_only_connectors = "paytm, phonepe, hyperpg, revolv3, fiservcommercehub"    # Comma-separated list of connectors that use UCS only
+ucs_psync_disabled_connectors = "cashtocode, trustly"    # Comma-separated list of connectors to disable UCS PSync call
 
 [revenue_recovery]
 # monitoring threshold -  120 days
--- a/crates/router/src/core/unified_connector_service.rs
+++ b/crates/router/src/core/unified_connector_service.rs
@@ -689,6 +689,7 @@ pub async fn should_call_unified_connector_service_for_webhooks(
     state: &SessionState,
     processor: &Processor,
     connector_name: &str,
+    merchant_connector_id: Option<&id_type::MerchantConnectorAccountId>,
 ) -> RouterResult<ExecutionPath> {
     // Extract context information
     let merchant_id = processor.get_account().get_id().get_string_repr();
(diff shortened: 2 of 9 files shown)


crates/router/src/core/unified_connector_service.rs
```diff
@@ -702,12 +703,16 @@ pub async fn should_call_unified_connector_service_for_webhooks(
     // Check UCS availability using idiomatic helper
     let ucs_availability = check_ucs_availability(state).await;
 
+    let connector_key = merchant_connector_id
+        .map(|id| id.get_string_repr().to_owned())
+        .unwrap_or_else(|| connector_name.to_string());
+
```
AmitsinghTanwar007, line 709: can we log like if mca_id is not there then just log using connector name for the key

AmitsinghTanwar007 (approved):

deepanshu-iiitu (approved):

crates/router/src/core/unified_connector_service.rs
Vani-1107: sure will take this in next pr

config/deployments/env_specific.toml
```diff
@@ -406,8 +406,8 @@ connector_names = "connector_names"     # Comma-separated list of allowed connec
 [grpc_client.unified_connector_service]
 base_url = "http://localhost:8000"      # Unified Connector Service Base URL
 connection_timeout = 10                 # Connection Timeout Duration in Seconds
-ucs_only_connectors = "paytm, phonepe, hyperpg, revolv3, truelayer, fiservcommercehub, trustly"    # Comma-separated list of connectors that use UCS only
-ucs_psync_disabled_connectors = "cashtocode"    # Comma-separated list of connectors to disable UCS PSync call
+ucs_only_connectors = "paytm, phonepe, hyperpg, revolv3, fiservcommercehub"    # Comma-separated list of connectors that use UCS only
```
hrithikesh026, line 409: how will payments go through truelayer and trustly?

crates/router/src/core/unified_connector_service.rs
```diff
@@ -702,12 +703,16 @@ pub async fn should_call_unified_connector_service_for_webhooks(
...
     // Build rollout keys - webhooks don't use payment method, so use a simplified key format
     let rollout_key = format!(
         "{}_{}_{}_{}",
         consts::UCS_ROLLOUT_PERCENT_CONFIG_PREFIX,
         merchant_id,
-        connector_name,
+        connector_key,
```
hrithikesh026, line 715: this will make all existing configs invalid. Better to have a fallback implementation.

config/deployments/env_specific.toml
Vani-1107: we'll be enabling the config explicitly for these two connectors for payins as the payout implementation for these is in hyperswitch.

crates/router/src/core/unified_connector_service.rs
Vani-1107: We do not have any existing configs for Webhooks in sandbox and production.

hrithikesh026 (approved):

PR #11834, opened 2026-04-16
feat(observability): instrument key manager with external service call event emission

### LINKED ISSUE #12184: [FEATURE] Wide-event observability pipeline for per-request context
## Summary

Today, getting visibility into Hyperswitch behavior requires adding a
metric or log line in the relevant code path, deploying, and waiting for
data. This works for the questions we already know to ask — but it
leaves us blind to **unknown unknowns**: the questions we didn't think
to instrument for ahead of time, and only realize we need to answer
when something breaks.

This issue proposes a **wide-event observability pipeline** that
captures as much context as possible per API request and ships it to a
queryable analytics store, so most investigations become SQL queries
instead of code changes and redeploys.

## Motivation

- **Answer unknown unknowns** — new questions shouldn't require new
  instrumentation; the data should already be there.
- **Reduce time-to-answer** — investigations move from days (PR →
  deploy → wait → query) to minutes (SQL).
- **Avoid metric sprawl** — every dimension lives on one row; the
  analyst slices at query time, not at instrumentation time.
- **Enable per-request correlation** across every downstream
  interaction (KeyManager, processors, DB, cache, …) in one place.
- **Support high-cardinality dimensions** like `request_id`, `endpoint`,
  `merchant_id` that are expensive in traditional metrics systems.

## Success criteria

- Common observability questions can be answered without code changes
  or deploys.
- Any request can be traced end-to-end by `request_id` in a single
  query.

spritianeja03:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

This PR adds the infrastructure to capture and publish key manager (encryption service) HTTP call metadata as Kafka events, and provides the ClickHouse schema to store them for observability queries.

### Key Changes

1. **`ExternalServiceCall` event type** (`crates/common_utils/src/external_service.rs`)

   - New `ExternalServiceCall` struct capturing `service_name`, `endpoint`, `method`, `request_id`, `status_code`, `success`, `latency_ms`, and `created_at_timestamp` for every external service HTTP call
   - `ExternalServiceEventEmitter` trait with a `NoOpEventEmitter` default (used when emission is disabled or in tests)

2. **Key Manager instrumentation** (`crates/common_utils/src/keymanager.rs`)

   - `send_encryption_request` now records wall-clock latency and emits an `ExternalServiceCall` event after every HTTP call to the key manager
   - Event is only emitted when `request_id` is present on the `KeyManagerState` (i.e. within a request context); a warning is logged otherwise

3. **`KeyManagerState` extended** (`crates/common_utils/src/types/keymanager.rs`)

   - Added `request_id: Option<String>` (previously gated behind `km_forward_x_request_id` feature flag, now always present)
   - Added `event_emitter: Arc<dyn ExternalServiceEventEmitter>` so the key manager client can emit events without depending on router internals

4. **Event routing to Kafka** (`crates/router/src/events.rs`, `crates/router/src/events/external_service_call.rs`, `crates/router/src/services/kafka.rs`)

   - New `EventType::ExternalServiceCall` variant
   - `KafkaExternalServiceCall` wrapper that implements `KafkaMessage`, routing events to `external_service_call_topic`
   - `EventsHandler` implements `ExternalServiceEventEmitter`, delegating to `log_event` (Kafka or logger depending on config)
   - `EventsConfig` refactored: `source` field separated out into `EventsSource` enum; new `emit_external_service_call_events: bool` flag gates event emission

5. **App state wiring** (`crates/router/src/routes/app.rs`, `crates/router/src/types/domain/types.rs`)

   - `KeyManagerState` constructed with the live `event_handler` when `emit_external_service_call_events = true`, `NoOpEventEmitter` otherwise — both in `AppState` (storage init path) and `SessionState` (per-request path)

6. **Config** (`config/development.toml`, `config/docker_compose.toml`, `config/config.example.toml`, `config/deployments/env_specific.toml`)

   - New Kafka topic key: `external_service_call_topic`
   - New flag: `emit_external_service_call_events` under `[events]`

7. **ClickHouse schema** (`crates/analytics/docs/clickhouse/scripts/api_observability.sql`)

   - `api_observability_events_queue` — Kafka Engine table consuming `hyperswitch-api-observability-events` with `kafka_handle_error_mode = 'stream'`
   - `api_events_flat` — MergeTree table, one row per external service call span; ORDER BY `(service_name, merchant_id, created_at, request_status_code)` for latency and error-rate queries
   - `api_observability_events` — MergeTree table, one row per API request with external service calls in a Nested column; used for per-request correlation
   - `api_events_flat_mv` and `api_observability_events_mv` — materialized views that fan out rows from the Kafka queue into the two storage tables at insert time
   - Optional `service_latency_agg` pre-aggregation table (commented out; deploy when row counts warrant it)

### Additional Changes

Config paths changed:
- `config/development.toml`
- `config/docker_compose.toml`
- `config/config.example.toml`
- `config/deployments/env_specific.toml`

## Motivation and Context

End-to-end latency visibility for API requests requires knowing not just total request latency but how much time was spent in downstream services. The key manager is called on virtually every encrypted field read/write. This PR makes those calls observable so latency breakdowns, error rates, and per-merchant usage patterns can be queried in ClickHouse.

## How did you test it?

External Service calls land correctly in the newly created kafka topic
image

Wide event being pushed to observability_events clickhouse table (stores one row per API request)
image

Flat events being stored in the observability_events_flat clickhouse table (stores one row per external service call)
image

## Checklist


config/development.toml
```diff
@@ -1317,7 +1299,8 @@ soft_kill = false
 enabled = true
 
 [events]
-source = "logs"
+source = "kafka"
```
rohanm13: Keep the source as "logs".

config/docker_compose.toml
```diff
@@ -1203,7 +1186,8 @@ partner_id = ""
 enabled = true
 
 [events]
-source = "logs"
+source = "kafka"
```
rohanm13: keep the source as "logs"

crates/common_utils/src/types/keymanager.rs
```diff
@@ -43,8 +43,8 @@ pub struct KeyManagerState {
     pub enabled: bool,
     pub url: String,
     pub client_idle_timeout: Option<u64>,
-    #[cfg(feature = "km_forward_x_request_id")]
-    pub request_id: Option<RequestId>,
+    pub request_id: Option<String>,
```
rohanm13, line 46: Can we check if changing the type from Option<RequestId> to Option<String> will break any existing code?

crates/router/src/services/kafka.rs
```diff
@@ -256,6 +257,15 @@ impl KafkaSettings {
             ))
         })?;
 
+        common_utils::fp_utils::when(
+            self.external_service_call_topic.is_default_or_empty(),
```
hrithikesh026, line 271: we only require this when `emit_external_service_call_events` is true right?

spritianeja03: yes correct, should this check be moved to the `EventsConfig::validate()` function (parent function for the `KafkaSettings::validate()` function) then ?

hrithikesh026: yes

hrithikesh026 (dismissed):

jarnura (dismissed):

hrithikesh026 (approved):

rohanm13 (approved):

jarnura (approved):

PR #11848, opened 2026-04-18
feat(ucs): inline connector config header into shadow-mode ComparisonData

10X-GRACE:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

Adds an optional `connector_config_header` field to the `ComparisonData` envelope serialized by the shadow-mode comparison path. This gives the parity-trace RPC immediate access to the connector-config snapshot without reconstructing it from `connector_auth_type`.

### Changes

1. **`ComparisonData` struct** — new `connector_config_header: Option<serde_json::Value>` field with `#[serde(skip_serializing_if = "Option::is_none")]` to preserve backward compatibility of the serialized envelope.

2. **Backward-compatible delegate** — existing `serialize_router_data_and_send_to_comparison_service` keeps its signature and delegates to the new `serialize_router_data_with_config_and_send_to_comparison_service`, passing `None`. Existing callers in `refunds.rs` require no changes.

3. **Unit tests** — two tests verify the field serializes correctly when `Some` and is omitted when `None`.

### Follow-up required

The callers in `refunds.rs` (L659, L1227) and the payment path in `gateway.rs` still pass `None` via the delegate. A follow-up change should thread the connector config from `build_unified_connector_service_auth_metadata` through to the comparison call sites.

## Additional Changes

- No schema or protocol changes
- Shadow-mode path only, no live-traffic impact

## Motivation and Context

Part of the UCS parity workstream. The parity-trace comparator needs the connector-config snapshot to reproduce UCS requests faithfully for shadow-diff analysis.

## How did you test it?

```
cargo check -p router --lib  # ✅ compiles
cargo test -p router --lib core::unified_connector_service::tests  # ✅ 2/2 pass
```

## Checklist

### DIFF
--- a/crates/router/src/core/unified_connector_service.rs
+++ b/crates/router/src/core/unified_connector_service.rs
@@ -2837,6 +2837,8 @@ where
 pub struct ComparisonData {
     pub hyperswitch_data: Secret<serde_json::Value>,
     pub unified_connector_service_data: Secret<serde_json::Value>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub connector_config_header: Option<serde_json::Value>,
 }
 
 /// Generic function to serialize router data and send comparison to external service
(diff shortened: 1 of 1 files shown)


10X-GRACE:
Superseded by #11849 which includes these changes plus the caller threading from PRI-49.

PR #12102, opened 2026-05-04
fix(authorizedotnet): use actual error text in CreateConnectorCustomer error message

10X-GRACE:
## Summary

Fix two parity diffs in `CreateConnectorCustomer` error path for AuthorizeDotNet: use actual connector error text instead of literal `"Error"`, and return `PaymentMethodAwaited` instead of `Failure` since customer creation is a pre-payment step.

## Linked PRD

Closes juspay/hyperswitch-cloud#15708

## Understanding Summary

- **Connector**: authorizedotnet
- **Flow**: CreateConnectorCustomer
- **Root cause**: `ResultCode` enum `.to_string()` produces `"Error"` literal instead of the actual error message text; `AttemptStatus::Failure` is semantically wrong for a pre-payment step failure.

### Oracle Behavior (UCS)
- `message` = actual connector error text (e.g. "User authentication failed due to invalid authentication values.")
- `status` = `payment_method_awaited`

### Divergence in hyperswitch
- `message` = `"Error"` (from `ResultCode::Error.to_string()`)
- `status` = `Failure`

### Change 1: Line 662
```rust
// Before:
message: item.response.messages.result_code.to_string(),
// After:
message: error_reason.clone(),
```

### Change 2: Line 675
```rust
// Before:
status: enums::AttemptStatus::Failure,
// After:
status: enums::AttemptStatus::PaymentMethodAwaited,
```

## Build Tail

Pre-existing build failures in `diesel_models` and `api_models` crates (confirmed identical on `main`). No new errors introduced by this change.

## Acceptance Criteria

- [x] Change is scoped to exactly 2 lines in 1 file
- [x] `error_reason.clone()` provides actual error text before ownership moves to `reason`
- [x] `PaymentMethodAwaited` is a valid `AttemptStatus` variant (confirmed via grep)
- [x] No proto/domain-type changes needed
- [ ] Shadow-replay produces zero diff (CTO gate)

### DIFF
--- a/crates/hyperswitch_connectors/src/connectors/authorizedotnet/transformers.rs
+++ b/crates/hyperswitch_connectors/src/connectors/authorizedotnet/transformers.rs
@@ -672,7 +672,7 @@ impl<F, T> TryFrom<ResponseRouterData<F, AuthorizedotnetCustomerResponse, T, Pay
                     });
                     Ok(Self {
                         response,
-                        status: enums::AttemptStatus::PaymentMethodAwaited,
+                        status: enums::AttemptStatus::Failure,
                         ..item.data
                     })
                 }
--- a/crates/hyperswitch_connectors/src/connectors/authorizedotnet/transformers.rs
+++ b/crates/hyperswitch_connectors/src/connectors/authorizedotnet/transformers.rs
@@ -659,7 +659,7 @@ impl<F, T> TryFrom<ResponseRouterData<F, AuthorizedotnetCustomerResponse, T, Pay
                         .join(" ");
                     let response = Err(ErrorResponse {
                         code: error_code,
-                        message: item.response.messages.result_code.to_string(),
+                        message: error_reason.clone(),
                         reason: Some(error_reason),
                         status_code: item.http_code,
                         attempt_status: None,
@@ -672,7 +672,7 @@ impl<F, T> TryFrom<ResponseRouterData<F, AuthorizedotnetCustomerResponse, T, Pay
                     });
                     Ok(Self {
                         response,
-                        status: enums::AttemptStatus::Failure,
+                        status: enums::AttemptStatus::PaymentMethodAwaited,
                         ..item.data
                     })
                 }


gopikrishna000 (commented):
## Review

### L662 — `message: error_reason.clone()` ✅ Correct

`result_code.to_string()` produces the literal string `"Error"` (just the enum label), not the actual error text. Using `error_reason.clone()` is the right fix and aligns with UCS behavior.

---

### L675 — `status: PaymentMethodAwaited` ❌ Wrong direction

Looking at the actual shadow diff that triggered this PR:

- The error is **E00007 (authentication failed)** — bad credentials, connector rejected the request outright
- HS correctly returns `status: failure`
- UCS returns `payment_method_awaited` because the `CreateConnectorCustomer` transformer never sets `new_router_data.resource_common_data.status` on error paths — the pre-call status leaks through. Every other flow in the same UCS file (Authorize, PSync, Capture etc.) explicitly sets `resource_common_data.status` before returning, but `CreateConnectorCustomer` is missing this.
- UCS own `CreateConnectorCustomer` transformer sets `attempt_status: Some(AttemptStatus::Failure)` in every error path — confirming `Failure` is the intended status.

Changing HS to `PaymentMethodAwaited` makes HS match UCS incorrect leaked state — both would be wrong.

**Right fix:**
- Revert L675, keep only the L662 change.
- The `status` diff needs a UCS-side fix: add `new_router_data.resource_common_data.status = AttemptStatus::Failure;` in the error paths of `CreateConnectorCustomer` transformer in `hyperswitch-prism/authorizedotnet/transformers.rs`. One line, same pattern already used by every other flow in that file.

10X-GRACE:
Thanks for the thorough review @gopikrishna000.

Addressed the feedback:
- **L662 (`message: error_reason.clone()`)** — kept as-is ✅
- **L675 (`status: PaymentMethodAwaited`)** — reverted to `Failure` ✅ per your analysis that UCS's `PaymentMethodAwaited` is a leaked pre-call status, not intended behavior

The UCS-side fix (setting `resource_common_data.status = AttemptStatus::Failure` in `CreateConnectorCustomer` error paths) is a separate concern for the prism repo.

10X-GRACE:
## CTO Gate Review — Approved ✅

**Code change is correct**: `error_reason.clone()` passes the actual connector error text instead of the literal enum name `"Error"`. Status correctly stays `Failure` on error path.

**CI**: Test/compilation failures on `hyperswitch-runners` and spell check are pre-existing (same failures on recently merged PRs on main). Not related to this change.

**One action item before merge**: Update PR title — it still says "fix error message **and status mapping**" but the status change was reverted. Suggest:
```
fix(authorizedotnet): use actual error text in CreateConnectorCustomer error message
```
The `parity` prefix isn't a recognized conventional commit type, which is why the title check is failing.

gopikrishna000 (approved):

gopikrishna000 (approved):

PR #1328, opened 2023-05-31
fix(connector): [DummyConnector] change dummy connector names

ThisIsMani:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [x] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

- Changed dummy connector names.

### Additional Changes

- [ ] This PR modifies the API contract
- [ ] This PR modifies the database schema
- [ ] This PR modifies application configuration/environment variables

## Motivation and Context

It seems "dummy" means stupid, so I have decided to change that.

## How did you test it?

Postman

## Checklist

- [x] I formatted the code `cargo +nightly fmt --all`
- [x] I addressed lints thrown by `cargo clippy`
- [x] I reviewed submitted code
- [ ] I added unit tests for my changes where possible
- [ ] I added a [CHANGELOG](/CHANGELOG.md) entry if applicable

PR #12110, opened 2026-05-04
parity(ucs): pass connector_metadata through from gRPC response

10X-GRACE:
## Summary
The UCS response handler hardcoded `connector_metadata: None` when converting gRPC `PaymentServiceGetResponse` to `PaymentsResponseData`, discarding the `metadata` field that connectors populate. This caused all UCS-routed flows to return `connector_metadata: null` in the response, diverging from the non-UCS path.

Refs #15798

## Understanding Summary
- **Connector**: forte (but fix is cross-cutting for all connectors)
- **Flow**: PSync
- **Field**: `response.Ok.TransactionResponse.connector_metadata`
- **Root cause**: `crates/hyperswitch_interfaces/src/unified_connector_service/transformers.rs` line 302 set `connector_metadata: None` instead of extracting the gRPC `metadata` field (protobuf tag 21)
- Both the hyperswitch oracle and the prism UCS forte transformers already correctly populate `connector_metadata` — the issue was solely in the gRPC-to-domain conversion layer

## Implementation Plan
- **File**: `crates/hyperswitch_interfaces/src/unified_connector_service/transformers.rs`
- **Change**: Replace `connector_metadata: None` with extraction from `response.metadata` — deserialize `Secret<String>` to `serde_json::Value` via `serde_json::from_str`
- **Fallback**: If `metadata` is `None` or fails to parse, result is `None` (no regression)

## Acceptance Criteria
- [x] Formatter passes
- [x] Only PRD-named file modified (`git diff --stat` shows 1 file, 3 insertions, 1 deletion)
- [ ] Build passes (pre-existing `diesel_models` workspace errors on `main` prevent local full build — CI will validate)
- [ ] Shadow-replay produces zero diff (CTO gate)

### DIFF
--- a/crates/hyperswitch_interfaces/src/unified_connector_service/transformers.rs
+++ b/crates/hyperswitch_interfaces/src/unified_connector_service/transformers.rs
@@ -299,7 +299,9 @@ impl ForeignTryFrom<(payments_grpc::PaymentServiceGetResponse, AttemptStatus)>
                             .transpose()?,
                     ),
                     mandate_reference: Box::new(response.mandate_reference.map(hyperswitch_domain_models::router_response_types::MandateReference::foreign_try_from).transpose()?),
-                    connector_metadata: None,
+                    connector_metadata: response
+                        .metadata
+                        .and_then(|m| serde_json::from_str(&m.expose()).ok()),
                     network_txn_id: response.network_transaction_id.clone(),
                     connector_response_reference_id: response.merchant_transaction_id,
                     incremental_authorization_allowed: response.incremental_authorization_allowed,


gopikrishna000 (approved):

gopikrishna000:
## Review

The HS-side fix here is correct — reading `response.metadata` back from the gRPC `PaymentServiceGetResponse` field 21 is the right approach.

However, this PR alone has no effect. The chain is:

```
UCS forte transformer → sets connector_metadata: Some(ForteMeta{...})  ✅
         ↓
UCS types.rs generate_payment_sync_response L6129 → connector_metadata: _   ← DROPPED ❌
         ↓
proto PaymentServiceGetResponse → metadata field 21 = None (never written)
         ↓
HS (this PR) → reads response.metadata = None → still None
```

Root cause is in UCS `crates/types-traits/domain_types/src/types.rs` in `generate_payment_sync_response` — the `PaymentsResponseData::TransactionResponse` match arm has `connector_metadata: _` (discarded). It needs to map `connector_metadata` → proto `metadata` field 21 via `convert_connector_metadata_to_secret_string`.

**A paired UCS (prism) PR is required** to write `connector_metadata` into the proto response. Until that ships, this PR is dead code — `response.metadata` will always be `None` and the shadow diff remains.

This PR can merge safely (no regression — `None.and_then(...)` = `None`), but the diff only closes once both sides are live.

@10X-GRACE

10X-GRACE (commented):
**CTO Review — LGTM** ✓

Reviewed the gRPC-to-domain conversion change:

- Type chain is correct: `Option<Secret<String>>` → `expose()` → `serde_json::from_str` → `Option<serde_json::Value>`
- Graceful fallback via `.ok()` — no regression risk on malformed metadata
- Follows the established `.expose()` pattern used throughout the transformers file
- Minimal change (3 ins / 1 del) in a single file — clean scope

Note: Line 286 (`ErrorResponse` path) still hardcodes `connector_metadata: None`. That's fine — error responses typically don't carry connector metadata, and it's out of scope for this parity diff. Can be addressed separately if a future diff surfaces it.

Paired with hyperswitch-prism PR #1328. Both PRs together close the forte PSync `connector_metadata` parity diff.

Ready to merge once CI passes.

PR #12119, opened 2026-05-04
parity(authipay/authorize): fix request and response parity diffs

10X-GRACE:
## Summary

Fix authipay connector parity diffs between HS and UCS for the Authorize flow (request + response).

## Linked Issues

- Paperclip: PRI-105
- Parent: https://github.com/juspay/hyperswitch-cloud/issues/15750
- Request diffs: https://github.com/juspay/hyperswitch-cloud/issues/15723
- Response diffs: https://github.com/juspay/hyperswitch-cloud/issues/15722

### Request fixes (authipay.rs + transformers.rs)
- Removed `Auth-Token-Type` header (sent by HS but not by UCS)
- Added `holder` (cardholder name) to `paymentMethod.paymentCard` object
- Added `merchantTransactionId` field (from `connector_request_reference_id`)
- Added `order` object with `orderId` (from `merchant_order_reference_id`)

### Response fixes (transformers.rs)
- Set `network_txn_id` from `scheme_transaction_id` (was null in HS, string in UCS)
- Added `token_reusable` to `connector_metadata` from `payment_token.reusable`

### Remaining investigative items
- `connector_metadata.order_id` — HS sets it, UCS doesn't. Kept because HS void operations depend on it. Needs UCS code comparison.
- `connector_response_reference_id` — HS uses `order_id`, UCS uses a different field. Kept as `order_id` (matches Fiserv pattern). Needs UCS code comparison.

## Test Plan

- [ ] `cargo check -p hyperswitch_connectors` passes (note: pre-existing `diesel_models` errors on main unrelated to this change)
- [ ] Verify request body includes `holder`, `merchantTransactionId`, `order` fields
- [ ] Verify `Auth-Token-Type` header is no longer sent
- [ ] Verify response includes `network_txn_id` and `connector_metadata.token_reusable`
- [ ] Shadow-replay produces zero diff for fixed items (CTO gate)

Co-Authored-By: Paperclip <noreply@paperclip.ing>

### DIFF
--- a/crates/hyperswitch_connectors/src/connectors/authipay.rs
+++ b/crates/hyperswitch_connectors/src/connectors/authipay.rs
@@ -133,7 +133,6 @@ where
                     .into(),
             ),
             ("Client-Request-Id".to_string(), client_request_id.into()),
-            ("Auth-Token-Type".to_string(), "HMAC".to_string().into()),
             (headers::TIMESTAMP.to_string(), timestamp.to_string().into()),
             ("Message-Signature".to_string(), hmac.into_masked()),
         ];
--- a/crates/hyperswitch_connectors/src/connectors/authipay.rs
+++ b/crates/hyperswitch_connectors/src/connectors/authipay.rs
@@ -133,6 +133,7 @@ where
                     .into(),
             ),
             ("Client-Request-Id".to_string(), client_request_id.into()),
+            ("Auth-Token-Type".to_string(), "HMAC".to_string().into()),
             (headers::TIMESTAMP.to_string(), timestamp.to_string().into()),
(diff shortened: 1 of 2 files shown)


crates/hyperswitch_connectors/src/connectors/authipay.rs
```diff
@@ -133,7 +133,6 @@ where
                     .into(),
             ),
             ("Client-Request-Id".to_string(), client_request_id.into()),
-            ("Auth-Token-Type".to_string(), "HMAC".to_string().into()),
```
gopikrishna000, line 136: @10X-GRACE  lets not remove this

PR #12411, opened 2026-05-21
fix(split_payments): remove deny_unknown_fields from StripeSplitPayments

### LINKED ISSUE #12412: fix(split_payments): Remove deny_unknown_fields to fix backward compatibility during staggered deployment
## Bug Description

During staggered deployment, newer code writes records with the new `on_behalf_of` field to the database. However, older running instances fail to deserialize these records because `StripeSplitPaymentRequest` has `#[serde(deny_unknown_fields)]` enabled.

This caused SEV1 5xx errors in production with the error:
```
Error deserializing field 'split_payments': unknown field `on_behalf_of`, expected one of `charge_type`, `application_fees`, `transfer_account_id`
```

## Impact

- SEV1 5xx errors during staggered deployment
- Payment session token creation fails
- Affects merchants using Stripe split payments

## Root Cause

`#[serde(deny_unknown_fields)]` on `StripeSplitPaymentRequest` prevents forward compatibility. When the database contains records with fields added by newer code, older code cannot deserialize them.

## Proposed Fix

Remove `#[serde(deny_unknown_fields)]` from `StripeSplitPaymentRequest` to allow graceful handling of unknown fields during staggered deployments.

## Long-term Solution

Have a separate equivalent Diesel Type for SplitPayment instead of reusing the API model.

## Testing Note

This issue can only be reproduced during staggered deployment scenarios and cannot be tested locally. It requires a multi-version deployment environment where:
1. Newer code writes records with new fields to the database
2. Older code attempts to read those records

## Related

- PR: #12411

awasthi21:
## Summary

Removes #[serde(deny_unknown_fields)] from StripeSplitPaymentRequest to allow backward compatibility when new fields like  are present in the database but the older code doesn't recognize them.

Fixes #12412

## Context

A new field  was added to the  struct. However, the struct had  enabled, which caused deserialization failures when older application instances tried to read records from the database that contained this new field.

This caused SEV1 5xx errors in production during staggered deployment because:
- Newer code wrote records with the new  field to the database
- Older running instances couldn't deserialize these records due to 

## Why This Fix

Removing  allows forward compatibility - older code will ignore unknown fields instead of failing deserialization. This is essential for zero-downtime deployments where database records may contain fields from newer code versions.

## Testing

This issue can only be reproduced during staggered deployment scenarios where:
1. Newer code writes records with new fields to the database
2. Older code attempts to read those records

This cannot be tested locally as it requires a multi-version deployment environment. The fix will be validated during the next staggered deployment.

## Type of Change
- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

### DIFF
--- a/api-reference/v1/openapi_spec_v1.json
+++ b/api-reference/v1/openapi_spec_v1.json
@@ -41464,8 +41464,7 @@
             "example": "acct_1234567890",
             "nullable": true
           }
-        },
-        "additionalProperties": false
+        }
       },
       "StripeSplitRefundRequest": {
         "type": "object",
--- a/api-reference/v2/openapi_spec_v2.json
+++ b/api-reference/v2/openapi_spec_v2.json
@@ -30823,8 +30823,7 @@
             "example": "acct_1234567890",
             "nullable": true
           }
-        },
-        "additionalProperties": false
+        }
       },
       "StripeSplitRefundRequest": {
         "type": "object",
--- a/crates/common_types/src/payments.rs
+++ b/crates/common_types/src/payments.rs
@@ -75,7 +75,6 @@ impl_to_sql_from_sql_json!(SplitPaymentsRequest);
     SmithyModel,
 )]
 #[diesel(sql_type = Jsonb)]
-#[serde(deny_unknown_fields)]
 #[smithy(namespace = "com.hyperswitch.smithy.types")]
 /// Fee information for Split Payments to be charged on the payment being collected for Stripe
 pub struct StripeSplitPaymentRequest {


hrithikesh026 (dismissed):

hrithikesh026:
Could you please test this once? Maybe add a unit test where you try deserialize to `pub enum SplitPaymentsRequest {`

hrithikesh026 (approved):

deepanshu-iiitu (approved):

deepanshu-iiitu (approved):

PR #12449, opened 2026-05-24
refactor(drainer): store SQL queries in drainer stream instead of JSON-ified entities

SanchithHegde:
## Type of Change

- [x] Bugfix
- [ ] New feature
- [x] Enhancement
- [x] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

This PR updates code around the KV flow(s) and drainer to store the SQL queries for `INSERT` / `UPDATE` operations in the drainer stream, instead of the previously stored JSON-ified objects. The changes in the PR are :warning: **breaking changes** :warning: since the format of drainer stream entries has changed significantly, and the changes are made with the assumption that there are no active uses of drainer, considering its beta nature.

In addition to the aforementioned change, this PR also includes the following changes, which can be traced to the respective commits:

- Adds a `just` recipe to run the drainer, and addresses a couple of bugs that prevented running drainer locally
- Fixes a bug in the payment attempt update code where the domain model was being serialized and stored in Redis instead of the diesel model, which caused deserialization failures when retrieving the payment. 
  - For context, we want to have the diesel model stored in Redis and not the domain model, since the diesel model would have the sensitive data encrypted while the domain model would not. This would allow us to treat data in Redis similar to data in PostgreSQL, and holds the same guarantees in terms of compliance.
- Updates the OpenTelemetry Collector config file to address config deserialization failures with the latest version of the collector.
- Removes the `From<diesel::result::Error>` implementation on `DatabaseError` in favor of the `ErrorSwitchFrom` implementation, since the `From` implementation lost context of the original error.

### For Reviewers

For reviewers, I would suggest going through the changes / files in the following order:

1. `crates/diesel_models/src/kv.rs`, `crates/diesel_models/Cargo.toml`: Includes the primary changes for the PR, and types and methods related to generating the SQL queries and executing them.
2. `crates/diesel_models/src/query/*.rs`: Includes methods to generate `INSERT` / `UPDATE` queries for KV-supported entities.
3. `crates/storage_impl/src/redis/kv_store.rs`
4. `crates/storage_impl/src/kv_router_store.rs`
5. Remaining files in `storage_impl` crate and `router` crate
6. `crates/drainer/src/utils.rs`, `crates/drainer/src/types.rs`, `crates/drainer/src/query.rs`, `crates/drainer/src/handler.rs`

## Motivation and Context

This change is being done to reduce the possibility of deserialization errors occurring in the drainer execution flow. This also helps decouple the drainer's deployment cycle from the rest of the application to a significant extent, since the data in the drainer stream is not as coupled to the version of router running as it is currently.

## How did you test it?

Running router and drainer locally, as well as OpenTelemetry Collector (with Docker Compose) for testing metrics.

### Verifying data format in drainer stream

1. Enable KV for a merchant account.
2. Create a payment via Postman for the merchant account.
3. Check data in Redis. During local testing, I ran the `MONITOR` command.

Screenshot of XADD command from Redis MONITOR output

- Query would be serialized in plaintext.
- Binds would be base64-encoded, as they're serialized in the PostgreSQL wire format, and they may be human-readable only in some cases.
- Type metadata is a list of `(oid, array_oid)` entries, which are serialized in plaintext.
- `safe_to_cache_prepared`, `entity_type` and `operation` should also be included in the JSON value.

The query generated would also be logged by the router:

Screenshot of generated database query

### Verifying query execution by drainer

1. Configure drainer's loop interval during deployment to be around 500 millisecond or 1 second.
2. Enable KV for a merchant account.
3. Create a payment via Postman for the merchant account.
4. Obtain the payment ID from the response and query for the payment ID in the database immediately. The database record won't be available immediately, but should be available after a few seconds, once the drainer has executed the queries.

Screenshot of queries run against PostgreSQL console

The drainer logs should also include the SQL query being executed (but we won't be able to see the bindings), the request ID, the global ID, the entity type, the operation and number of rows affected after the successful execution of the query:

Screenshot of drainer logs

### Drainer Metrics

- The router should be exporting the `KV_PUSHED_TO_DRAINER` metric indicating the count of entries pushed to drainer stream. The metric should include the entity type and operation as attributes.

  Screenshot of Grafana querying for KV_PUSHED_TO_DRAINER metric

  Query used:

  ```text
  sum by(entity_type, operation) (increase(KV_PUSHED_TO_DRAINER_total[$__rate_interval]))
  ```

- Similarly, drainer should be exporting `QUERY_EXECUTION_TIME`, `DRAINER_DELAY_SECONDS`, `CYCLES_COMPLETED_SUCCESSFULLY` and a few other metrics.

## Checklist


crates/diesel_models/src/errors.rs
```diff
@@ -8,16 +8,15 @@ pub enum DatabaseError {
...
     QueryGenerationFailed,
-    // InsertFailed,
     #[error("An unknown error occurred")]
     Others,
 }
 
-impl From<diesel::result::Error> for DatabaseError {
-    fn from(error: diesel::result::Error) -> Self {
-        match error {
+impl common_utils::errors::ErrorSwitchFrom<diesel::result::Error> for DatabaseError {
```
SanchithHegde, line 17: I added the `ErrorSwitchFrom` impl as the previous `From` impl would lose the context of the underlying `diesel` error.

I've also replaced the usages of this `From` impl with `StorageError::from(...)`, which makes use of [this `ErrorSwitchFrom` implementation](https://github.com/juspay/hyperswitch/blob/3647f3bebbe022545242d419a32e0c042dc8f7f9/crates/storage_impl/src/errors.rs#L53-L74).

crates/diesel_models/src/kv.rs
```diff
@@ -1,189 +1,222 @@
...
-                Updateable::AddressUpdate(_) => "address",
-                Updateable::PayoutsUpdate(_) => "payouts",
-                Updateable::PayoutAttemptUpdate(_) => "payout_attempt",
-                Updateable::PaymentMethodUpdate(_) => "payment_method",
-                Updateable::MandateUpdate(_) => " mandate",
-            },
-        }
+/// Masking strategy for binary data or raw bytes
+#[derive(Debug)]
+pub enum BinaryDataStrategy {}
```
SanchithHegde: Let me know if I should move this to `common_utils` or another crate, or if it can remain here.

crates/diesel_models/src/kv.rs
```diff
@@ -1,189 +1,222 @@
...
+
+/// The SQL query and its bind parameters, in a (de)serialization-friendly representation
+#[derive(Debug, serde::Serialize, serde::Deserialize)]
+pub struct SerializableQuery {
+    /// The SQL query
+    sql: String,
+
+    /// The serialized bytes for each bind parameter
+    #[serde(with = "bind_params")]
+    binds: Vec<Option<SecretBinaryData>>,
```
SanchithHegde, line 37: This implementation (along with the `serde(with)`) serializes individual bind parameters (`Vec<u8>`) as base64-encoded values rather than as JSON arrays (for example, `[75, 105, 193, 73, 8, 154]`). These params would anyway not be human readable, I felt that we could encode bytes rather than using JSON arrays. 

This may also help slightly reduce the data transmitted over the network to the Redis server.

crates/storage_impl/src/customers.rs
```diff
@@ -166,13 +166,28 @@ impl<T: DatabaseStore> domain::CustomerInterface for kv_router_store::KVRouterSt
...
+        let updated_customer = customer_update_internal
+            .clone()
             .apply_changeset(customer.clone());
         let key = PartitionKey::MerchantIdCustomerId {
             merchant_id: &merchant_id,
             customer_id: &customer_id,
         };
         let field = format!("cust_{}", customer_id.get_string_repr());
+
+        let mut query_gen_conn = pg_connection_read(self).await?;
```
SanchithHegde: The Postgres type metadata lookup (OID and array OID) requires a database connection. I'm assuming that using a connection to the read replica should be sufficient for this purpose. 

One of the rare cases where this may break is when a new user defined type is added, hasn't been propagated to the replica yet, and router is trying to insert a value using that type.

crates/diesel_models/src/query/customers.rs
```diff
@@ -326,3 +333,29 @@ impl Customer {
         .await
     }
 }
+
+impl CustomerUpdateInternal {
```
SanchithHegde, line 337: For the `UPDATE` queries, I've tried to mirror the queries used in the [previous implementation](https://github.com/juspay/hyperswitch/blob/289685fa12210ac49e86bcc7fe9e021d5aa52213/crates/diesel_models/src/kv.rs#L114-L181).

crates/diesel_models/src/query/payment_attempt.rs
```diff
@@ -551,3 +559,41 @@ impl PaymentAttempt {
...
+        kv::generate_update_query_with_predicate::<<PaymentAttempt as HasTable>::Table, _, _>(
+            conn,
+            dsl::attempt_id
+                .eq(source_payment_attempt.attempt_id.clone())
+                .and(
+                    dsl::processor_merchant_id
+                        .eq(source_payment_attempt.processor_merchant_id.clone()),
+                ),
+            PaymentAttemptUpdateInternal::from(self)
+                .populate_derived_fields(source_payment_attempt),
```
SanchithHegde, line 579: Please confirm if calling `populate_derived_fields()` here is the expected behavior or not.

crates/drainer/src/query.rs
```diff
@@ -15,34 +15,37 @@ pub trait ExecuteQuery {
...
 
-        let tags = router_env::metric_attributes!(("operation", operation), ("table", table));
+        let metric_attributes = router_env::metric_attributes!(
+            ("operation", operation.clone()),
+            ("entity_type", entity_type.clone())
+        );
 
         let (result, execution_time) =
-            Box::pin(common_utils::date_time::time_it(|| self.execute(&conn))).await;
+            common_utils::date_time::time_it(|| self.execute(&mut conn)).await;
```
SanchithHegde, line 34: The database query execution now only returns `usize` (previously, we'd get the row after the `INSERT` / `UPDATE` query execution), indicating the number of rows affected. This is because `CollectedQuery` (from `diesel`) does not implement `RunQueryDsl` or `LoadQuery`, and the drainer won't know the return type with the full set of expected columns either.

crates/storage_impl/src/payments/payment_attempt.rs
```diff
@@ -1053,38 +1051,32 @@ impl<T: DatabaseStore> PaymentAttemptInterface for KVRouterStore<T> {
...
-                            },
-                        ))),
-                    },
-                };
+                let redis_value = serde_json::to_string(
+                    &updated_attempt
+                        .clone()
+                        .convert()
+                        .await
+                        .change_context(errors::StorageError::EncryptionError)?,
```
SanchithHegde: We were previously serializing the domain model here during payment attempt update and storing it in Redis, while we were supposed to store the diesel model instead. 

This was a bug, and I have fixed it when testing out the payment create + payment retrieve flows.

crates/diesel_models/src/kv.rs
```diff
@@ -1,189 +1,222 @@
...
-    PaymentMethodUpdateInternal, PgPooledConn,
+use async_bb8_diesel::AsyncConnection;
+use diesel::{
+    associations::HasTable,
+    debug_query,
+    dsl::{Filter, Find},
+    pg::Pg,
+    query_builder::{
+        bind_collector::RawBytesBindCollector, AsChangeset, AsQuery, CollectedQuery,
+        InsertStatement, IntoUpdateTarget, MoveableBindCollector, QueryBuilder, QueryFragment,
```
XyneSpaces, line 14: :rotating_light: **Breaking change - No backward compatibility handling**

The complete replacement of `TypedSql`/`DBOperation` with `SerializableQuery` removes support for deserializing old-format drainer stream entries. During a rolling deployment, old entries in the stream will fail deserialization and break the drainer loop.

Add a fallback deserializer or provide a migration procedure to drain old-format entries before deploying this change.

crates/diesel_models/src/kv.rs
```diff
@@ -1,189 +1,222 @@
...
-        match self {
-            Self::Insert { .. } => "insert",
-            Self::Update { .. } => "update",
-        }
-    }
-    pub fn table<'a>(&self) -> &'a str {
-        match self {
-            Self::Insert { insertable } => match **insertable {
-                Insertable::PaymentIntent(_) => "payment_intent",
-                Insertable::PaymentAttempt(_) => "payment_attempt",
```
XyneSpaces, line 42: :warning: **Error handling: Silent truncation in metadata serialization**

```rust
.map(|m| (m.oid().unwrap_or(0), m.array_oid().unwrap_or(0)))
```

Using `unwrap_or(0)` for OID values masks potential type metadata issues. OID 0 is invalid in PostgreSQL - this could cause query execution failures in the drainer that are hard to debug.

Propagate the error instead of defaulting to 0, or add explicit logging when OID is missing.

crates/diesel_models/src/kv.rs
```diff
@@ -1,189 +1,222 @@
...
-                Updateable::PayoutsUpdate(_) => "payouts",
-                Updateable::PayoutAttemptUpdate(_) => "payout_attempt",
-                Updateable::PaymentMethodUpdate(_) => "payment_method",
-                Updateable::MandateUpdate(_) => " mandate",
-            },
-        }
+/// Masking strategy for binary data or raw bytes
+#[derive(Debug)]
+pub enum BinaryDataStrategy {}
+
```
XyneSpaces: :warning: **Missing instrumentation on async function**

`generate_insert_query` and `generate_update_query_by_id` are async functions that interact with the database connection but lack `#[instrument]` annotations. This makes it difficult to trace query generation in production logs.

Add `#[instrument(skip_all, fields(entity_type))]` to these functions.

crates/storage_impl/src/kv_router_store.rs
```diff
@@ -202,20 +202,24 @@ impl<T: DatabaseStore> KVRouterStore<T> {
...
             .cache_store
             .redis_conn
             .stream_append_entry(
                 &stream_name.into(),
                 &redis_interface::RedisEntryId::AutoGeneratedID,
-                redis_entry
+                serializable_query
                     .to_field_value_pairs(request_id, global_id)
                     .change_context(RedisError::JsonSerializationFailed)?,
             )
```
XyneSpaces, line 218: :bulb: **Metric attribute creation in hot path**

Creating metric attributes on every drainer stream push adds allocation overhead:
```rust
let metric_attributes = router_env::metric_attributes!(
    ("operation", serializable_query.operation().to_string()),
    ("entity_type", serializable_query.entity_type())
);
```

The `operation()` call clones the string unnecessarily. Consider storing these as `&'static str` in the enum or caching the attributes.

crates/router/src/db/refund.rs
```diff
@@ -890,22 +891,25 @@ mod storage {
...
 
-                    let redis_entry = kv::TypedSql {
-                        op: kv::DBOperation::Update {
-                            updatable: Box::new(kv::Updateable::RefundUpdate(Box::new(
-                                kv::RefundUpdateMems {
-                                    orig: this,
-                                    update_data: refund,
-                                },
-                            ))),
-                        },
```
XyneSpaces, line 901: :warning: **Clone in query generation path**

```rust
let drainer_query = refund
    .clone()  // Unnecessary clone
    .generate_drainer_update_query(...)
```

The `clone()` here is unnecessary since `generate_drainer_update_query` takes `self` by value. Check if the `refund` value is used after this call; if not, remove the clone.

XyneSpaces (dismissed):
:rotating_light: **Critical: Breaking change lacks migration path documentation**

The PR description acknowledges this is a :warning: **breaking change** :warning: but there's no explicit migration guide for operators upgrading from the JSON-ified entity format to the new SQL query format. The drainer stream entries format change will cause deserialization failures during rolling deployments.

Add a `MIGRATION.md` section or update the PR description with:
1. Minimum drainer version required before router upgrade
2. Procedure for clearing existing stream entries during transition
3. Verification steps for operators

crates/diesel_models/src/kv.rs
```diff
@@ -1,189 +1,222 @@
...
+    /// The serialized bytes for each bind parameter
+    #[serde(with = "bind_params")]
+    binds: Vec<Option<SecretBinaryData>>,
+
+    /// The metadata associated with each bind parameter
+    #[serde(with = "pg_type_metadata")]
+    metadata: Vec<diesel::pg::PgTypeMetadata>,
+
+    /// Whether this query is safe to store in the prepared statement cache
+    safe_to_cache_prepared: bool,
```
hrithikesh026, line 44: why is this needed?

crates/diesel_models/src/kv.rs
```diff
@@ -1,189 +1,222 @@
...
+    pub fn entity_type(&self) -> String {
+        self.entity_type.clone()
+    }
+
+    pub fn operation(&self) -> DatabaseOperation {
+        self.operation
+    }
+
+    async fn from_query<Q>(
+        conn: &mut crate::PgPooledConn,
```
hrithikesh026, line 71: why does this have to be mutable?

crates/diesel_models/src/kv.rs
SanchithHegde: [`CollectedQuery::new()`](https://docs.rs/diesel/2.3.9/diesel/query_builder/struct.CollectedQuery.html#method.new) requires that to be passed.

crates/diesel_models/src/kv.rs
SanchithHegde: [`collect_binds()`](https://docs.rs/diesel/2.3.9/src/diesel/query_builder/mod.rs.html#253-260) needs a mutable value.

ShankarSinghC (dismissed):

Sakilmostak (dismissed):

hrithikesh026 (dismissed):

ShankarSinghC (dismissed):

crates/diesel_models/src/kv.rs
jarnura: this can be moved to common_utils

crates/diesel_models/src/kv.rs
```diff
@@ -1,189 +1,222 @@
...
-                    )
-                    .await?,
-                )),
-                #[cfg(feature = "v2")]
-                Updateable::CustomerUpdate(cust) => DBResult::Customer(Box::new(
-                    Customer::update_by_id(conn, cust.orig.id, cust.update_data).await?,
-                )),
-            },
-        })
+mod bind_params {
```
jarnura: seperate mod can have their own files instead of keeping everything in same file.

crates/diesel_models/src/kv.rs
```diff
@@ -204,100 +237,125 @@ impl TypedSql {
...
+            crate::payment_attempt::PaymentAttemptNew,
+            crate::payment_attempt::PaymentAttemptUpdateInternal,
+        }
+        "customer" => {
+            crate::customers::CustomerNew,
+            crate::customers::CustomerUpdateInternal,
+        }
+        "refund" => {
+            crate::refund::RefundNew,
+            crate::refund::RefundUpdate,
```
jarnura: Remove RefundUpdate or RefundUpdateInternal if possible

crates/storage_impl/src/customers.rs
jarnura: Make this master itself.

jarnura (changes requested):

jarnura (approved):

Sakilmostak (approved):

ShankarSinghC (approved):

PR #12480, opened 2026-05-26
feat(cache): add in-memory cache support for organization

SanchithHegde:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

This PR adds in-memory cache support for the organization table. As with the other tables with cache support, the cache would be populated on retrieval, and invalidated on updates. And no action would be performed on the cache during organization creation (inserts).

## Motivation and Context

We are in the process of adding in-memory cache support to missing entities in the organization-merchant-profile hierarchy as well as other accounts related entities (organization, merchant account, merchant key store, API key, profile, merchant connector account).

As of opening this PR, organization and profile are the only 2 entities without cache support. This PR adds support for organization, support for profiles will be added in a different PR.

## How did you test it?

Running the router service locally, and testing out the organization create, retrieve, update APIs with Postman.

1. Create organization using the API.
2. Retrieve organization using the API. This should populate Redis and in-memory cache.
   - Running `MONITOR` command against Redis before calling this API shows that a key was set with key being organization ID (if multi-tenancy is enabled, this key would also include the tenant's Redis key prefix) and value being the JSON serialized organization object.
     Screenshot from Redis CLI indicating GET and SET commands
     In this screenshot, a `GET` call was performed first. Since the value was not available in Redis, a `SET` command was called later.
   - Since this is the first time the value is being cached, we should also see a log line displaying the `SELECT` query being run against the database.
     Screenshot of SELECT SQL query
3. Further attempts to retrieve the organization would NOT display the database query, since the value would already be cached in Redis. Additionally, if there is only a single instance of the application running, retrieving the organization again won't reach Redis either, it would be directly returned from in-memory cache, this can be verified using the `MONITOR` command.
4. Update the organization using the API. This should invalidate the cache. Logs would indicate that the in-memory cache was invalidated. Note the "Received message on channel" and the "Done invalidating" logs in the screenshot.
   Screenshot of logs indicating invalidation of cache
   The Redis `MONITOR` command also indicates that the key would be deleted.
   Screenshot from Redis indicating the DEL and PUBLISH commands
5. The next attempt to retrieve the organization again would populate Redis and in-memory cache, by fetching the value from the database. This would be evident from the database query being printed in logs, as described in 2.

The same behavior would also be visible with v2 organization endpoints.

## Checklist

### DIFF
--- a/crates/diesel_models/src/organization.rs
+++ b/crates/diesel_models/src/organization.rs
@@ -1,4 +1,4 @@
-use common_utils::{id_type, pii};
+use common_utils::{custom_serde, id_type, pii};
 use diesel::{AsChangeset, Identifiable, Insertable, Queryable, Selectable};
 
 #[cfg(feature = "v1")]
@@ -11,7 +11,9 @@ pub trait OrganizationBridge {
     fn set_organization_name(&mut self, organization_name: String);
 }
 #[cfg(feature = "v1")]
-#[derive(Clone, Debug, Identifiable, Queryable, Selectable)]
+#[derive(
+    Clone, Debug, Identifiable, Queryable, Selectable, serde::Serialize, serde::Deserialize,
+)]
 #[diesel(
     table_name = organization,
     primary_key(org_id),
@@ -22,7 +24,9 @@ pub struct Organization {
     org_name: Option<String>,
     pub organization_details: Option<pii::SecretSerdeValue>,
     pub metadata: Option<pii::SecretSerdeValue>,
+    #[serde(with = "custom_serde::iso8601")]
     pub created_at: time::PrimitiveDateTime,
+    #[serde(with = "custom_serde::iso8601")]
     pub modified_at: time::PrimitiveDateTime,
     #[allow(dead_code)]
     id: Option<id_type::OrganizationId>,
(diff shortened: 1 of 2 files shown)


XyneSpaces (commented):
## Automated Code Review

### Summary
This PR adds in-memory cache support for the organization table, following the established pattern used by other entities. The implementation looks correct and follows the existing cache architecture.

### Findings

**:bulb: Consider adding `#[instrument]` to cache-wrapped methods**

The `find_by_org_id` and `update_by_org_id` methods now have cache logic. Consider adding `#[instrument(skip_all, fields(org_id = %org_id))]` to the trait implementations for observability in traces.

**File:** `crates/router/src/db/organization.rs`

---

**:mag: Minor: Feature gate consistency**

The cache feature gating is correct (`#[cfg(feature = "accounts_cache")]`). However, verify that the `accounts_cache` feature is enabled in all relevant environment configurations (development.toml, sandbox.toml, production.toml) if this is intended to be active.

---

**No critical issues found.** The implementation correctly:
- Uses `get_or_populate_in_memory` for read caching
- Uses `publish_and_redact` for cache invalidation on updates  
- Properly serializes/deserializes the Organization struct with ISO8601 datetime handling
- Follows the same pattern as merchant account and profile caching

### Verdict
`approve` with minor suggestions.

apoorvdixit88 (approved):

ShankarSinghC (approved):

hrithikesh026 (approved):

su-shivanshmathur (approved):

PR #12630, opened 2026-06-08
feat(kv): add Redis KV support for captures table

hrithikesh026:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD
## Description

This PR adds Redis KV (key-value store) support for the `captures` table, following the same pattern used by `payment_attempt` and `refund`. Previously, the captures table only supported direct PostgreSQL operations even when `kv_store` feature was enabled. With this change, captures can now be stored, updated, and queried from Redis with drainer-based async persistence to Postgres.
### Changes made:
1. **Drainer query generation (`crates/diesel_models/src/query/capture.rs`)**
   - Added `generate_drainer_insert_query()` to `CaptureNew`
   - Added `generate_drainer_update_query()` to `CaptureUpdate`
2. **Entity type registration (`crates/diesel_models/src/kv/entity_type.rs`)**
   - Registered `"capture"` entity type for `CaptureNew` and `CaptureUpdateInternal`
3. **Module exports (`crates/diesel_models/src/lib.rs`)**
   - Exported `capture::*` at the crate root so `Capture` is accessible as `diesel_models::Capture`
4. **KV storage traits (`crates/storage_impl/src/lib.rs`)**
   - Added `pub mod capture;`
   - Introduced `KvSupportedEntity` trait for entities that support KV, defining:
     - `get_partition_key()` — returns the Redis hash key
     - `get_hash_field_key()` — returns the Redis hash field
   - Implemented `KvSupportedEntity` for `Capture`
   - Implemented `UniqueConstraints` for `Capture`
5. **KV partition support (`crates/storage_impl/src/capture.rs`)**
   - Added `KvStorePartition` impl for `Capture`
6. **KV-aware interface (`crates/router/src/db/capture.rs`)**
   - Replaced the no-op `kv_store` implementation with full KV support:
     - `insert_capture` — uses `HSetNx` with drainer insert query, creates reverse lookups
     - `update_capture_with_capture_id` — uses `Hset` with drainer update query
     - `find_all_captures_by_merchant_id_payment_id_authorized_attempt_id` — uses `Scan` with fallback to DB
   - Falls back to PostgresOnly when `soft_kill_mode` is active or storage scheme is configured accordingly
### Additional Changes

## Motivation and Context

The `captures` table was the only major payments-related table that lacked KV store support. All other tables (`payment_intent`, `payment_attempt`, `refund`) already had Redis KV implementations. This inconsistency meant that capture operations always hit Postgres directly, bypassing the caching and performance benefits of the KV layer. This PR closes that gap by implementing KV support for captures in the same style as the existing tables.
## How did you test it?

- Verified compilation succeeds with `cargo check` for the `router` crate with `kv_store` and `v1` features enabled.
- The KV path is exercised via the same `decide_storage_scheme` / `kv_wrapper` infrastructure shared by `payment_attempt` and `refund`, which is already battle-tested in production.

capture insert and update queries are being run successfully by the drainer.

Screenshot 2026-06-08 at 2 55 29 PM

Screenshot 2026-06-08 at 2 55 49 PM

capture records are being written and updated in DB by the drainer.
image

For testing merging logic of redis and DB captures, Created 1st and 2nd capture together(read happened from redis alone) and then later created a 3rd capture after TTL, 3rd capture was in Redis Cache but first 2 were in DB. And the response contained all three captures(merge logic worked).

image
log line:

image

## Checklist

---


crates/storage_impl/src/lib.rs
```diff
@@ -363,6 +365,27 @@ pub trait UniqueConstraints {
     }
 }
 
+/// This trait defines behaviour that must be followed by any table that has support for KV
```
XyneSpaces, line 368: :mag: **Nit: Trait documentation could be more specific**

The `KvSupportedEntity` trait comment says "any table that has support for KV" but this is specifically for entities stored in Redis hash structures with partition keys and field keys. Consider clarifying:

```rust
/// Entities that support Redis KV storage via hash-based partitioning.
/// Implementors define how the entity maps to Redis hash keys/fields.
```

crates/router/src/db/capture.rs
```diff
@@ -49,32 +55,128 @@ mod storage {
...
+                        key,
+                    ))
                     .await
-                    .map_err(|error| report!(errors::StorageError::from(error)))
-            };
-            db_call().await
+                    .map_err(|err| err.to_redis_failed_response(&key_str))?
+                    .try_into_hsetnx()
+                    {
+                        Ok(HsetnxReply::KeyNotSet) => Err(errors::StorageError::DuplicateValue {
```
XyneSpaces, line 119: :mag: **Nit: Missing `#[instrument]` on async function**

The `insert_capture` async function in the `storage` module should have `#[instrument(skip_all)]` for observability consistency with other storage operations. All async functions in storage modules should carry instrumentation.

```rust
#[instrument(skip_all)]
async fn insert_capture(...)
```

crates/router/src/db/capture.rs
```diff
@@ -49,32 +55,128 @@ mod storage {
...
+                    let key_str = key.to_string();
+                    let updated_capture = capture.clone().apply_changeset(this.clone());
+
+                    let redis_value = serde_json::to_string(&updated_capture)
+                        .change_context(errors::StorageError::SerializationFailed)?;
+
+                    let mut query_gen_conn = connection::pg_connection_write(self).await?;
+                    let drainer_query = capture
+                        .generate_drainer_update_query(&mut query_gen_conn, this.capture_id.clone())
+                        .await
```
XyneSpaces, line 163: :mag: **Nit: Missing `#[instrument]` on async function**

The `update_capture_with_capture_id` async function should also have `#[instrument(skip_all)]` for observability consistency.

XyneSpaces (commented):
## Automated Review Summary

**Classification:** Enhancement — Redis KV Support for Captures Table  
**Risk Level:** Medium (KV infrastructure change affecting payment data persistence)

This PR adds Redis KV support for the captures table following the pattern established by `payment_attempt` and `refund`. The implementation properly integrates with the existing `kv_wrapper` infrastructure and drainer-based persistence.

### Overall Assessment
The implementation follows established patterns and properly handles the KV storage scheme with fallback to Postgres. No blocking issues found.

crates/router/src/db/capture.rs
```diff
@@ -96,7 +198,36 @@ mod storage {
...
+            ))
+            .await;
+            match storage_scheme {
+                enums::MerchantStorageScheme::PostgresOnly => database_call().await,
+                enums::MerchantStorageScheme::RedisKv => {
+                    let key = PartitionKey::MerchantIdPaymentId {
+                        merchant_id,
+                        payment_id,
+                    };
+                    let pattern = format!("pa_{authorized_attempt_id}_capture_*");
```
SanchithHegde, line 214: Nit: We've extracted out the partition key and field generation logic to a trait implementation, should we consider extracting this (pattern) as well?

crates/router/src/db/capture.rs
SanchithHegde, line 1: Should we consider using [`insert_resource()`](https://github.com/juspay/hyperswitch/blob/ec2dbd6addf3593019278b1ecaee69dcff2906f0/crates/storage_impl/src/kv_router_store.rs#L371) / [`update_resource()`](https://github.com/juspay/hyperswitch/blob/ec2dbd6addf3593019278b1ecaee69dcff2906f0/crates/storage_impl/src/kv_router_store.rs#L445) / [`filter_resources()`](https://github.com/juspay/hyperswitch/blob/ec2dbd6addf3593019278b1ecaee69dcff2906f0/crates/storage_impl/src/kv_router_store.rs#L505) instead?

I'm aware that may require us to implement `Conversion` trait for encryption / decryption support, but do you think we should go ahead with that? Currently, for capture entity, we would not have to actually do any encryptions / decryptions, as they're not required.

crates/router/src/db/capture.rs
```diff
@@ -96,7 +198,36 @@ mod storage {
...
-            db_call().await
+            let storage_scheme = Box::pin(decide_storage_scheme::<_, Capture>(
+                self,
+                storage_scheme,
+                Op::Find,
+            ))
+            .await;
+            match storage_scheme {
+                enums::MerchantStorageScheme::PostgresOnly => database_call().await,
+                enums::MerchantStorageScheme::RedisKv => {
```
SanchithHegde, line 209: Are we handling the case where a subset of the captures may possibly be present in the database?

SanchithHegde:
Also, please test out the flows locally and update PR description.

crates/router/src/db/capture.rs
hrithikesh026: Yeah better to do as well. Thanks.

hrithikesh026: I would rather pick that up in a separate pr though. Favouring ease of review.

crates/router/src/db/capture.rs
hrithikesh026: During Read operation, We don't really have the entity object(capture object in this case). So we are gonna need a different technique to solve for reads. Will see what can be done for that.

su-shivanshmathur (approved):

SanchithHegde (approved):

PR #12648, opened 2026-06-09
feat(payment_methods): add per-network surcharge to client payment methods list endpoint

### LINKED ISSUE #12649: add per-network surcharge to client payment methods list endpoint
Adds per-card-network surcharge details to the `/payments/{payment_id}/client` endpoint response, so the SDK can display accurate surcharge amounts per card network without a separate API call.

ShankarSinghC:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

Adds per-card-network surcharge details to the `/payments/{payment_id}/client` endpoint response, so the SDK can display accurate surcharge amounts per card network without a separate API call.

### Additional Changes

## Checklist


crates/api_models/src/payment_methods.rs
```diff
@@ -2912,16 +2912,26 @@ pub struct PaymentMethodListIntentData {
...
 #[serde(untagged)]
 pub enum PaymentMethodSubtypeSpecificDataForClient {
-    /// Card payment method — contains card network info
+    /// Card payment method — contains the list of enabled card networks
     Card {
         #[schema(value_type = Vec<CardNetwork>)]
         card_networks: Vec<api_enums::CardNetwork>,
+        /// Per-network surcharge details; empty when no surcharge rules are configured.
+        #[serde(default, skip_serializing_if = "HashMap::is_empty")]
+        card_network_surcharge_details: HashMap<api_enums::CardNetwork, SurchargeDetailsResponse>,
```
XyneSpaces, line 2921: :rotating_light: **Breaking API change**: The `#[serde(untagged)]` enum `PaymentMethodSubtypeSpecificDataForClient` adds a new `Wallet` variant and removes fields from `ResponsePaymentMethodsEnabledForClient`.

This is a backwards-incompatible change for existing SDK integrations. Clients expecting `collect_shipping_details_from_wallets` at the top level will receive it nested in `data` instead.

Verify that this change has been coordinated with SDK teams and is documented in the PR body. If SDK compatibility is required, consider keeping the deprecated fields with `#[serde(skip_deserializing)]` during a migration period.

crates/router/src/core/payment_methods/client.rs
```diff
@@ -65,6 +65,7 @@ fn to_client_pm(pm: CustomerPaymentMethod) -> CustomerPaymentMethodForClient {
         created: pm.created,
         last_used_at: pm.last_used_at,
         payment_method_data,
+        surcharge_details: None,
```
XyneSpaces, line 68: :warning: **Placeholder surcharge value**: `surcharge_details: None` is hardcoded here without a TODO or explanation.

Per the PR description, this endpoint should return surcharge details for saved cards. If this is intentionally deferred, add a `// TODO(pr#<number>): compute surcharge for saved PMs` comment. If this is in scope, the saved PM surcharge calculation should be implemented before merge.

crates/router/src/core/payment_methods/cards.rs
```diff
@@ -4742,6 +4741,120 @@ pub async fn call_surcharge_decision_management(
     Ok(merchant_sucharge_configs)
 }
 
+/// Converts flat client PMs into the legacy nested `ResponsePaymentMethodsEnabled` type
+/// so that `call_surcharge_decision_management` can run unchanged.
+///
+/// Only PM types whose surcharge is surfaced in the client API response are included here;
+/// skipped types are not passed to the surcharge engine so Redis stays consistent with
+/// what the response actually contains.
+///
```
XyneSpaces, line 4750: :bulb: **Documentation improvement**: The doc comment says `Only PM types whose surcharge is surfaced in the client API response are included here` but the match arm skips Wallet and Bank with `continue`.

Consider explicitly documenting *why* these are skipped (e.g., "Wallet/Bank surcharge not yet exposed in client API v1") so future maintainers understand when to extend this function.

crates/router/src/core/payment_methods/cards.rs
```diff
@@ -4742,6 +4741,120 @@ pub async fn call_surcharge_decision_management(
...
+/// Copies surcharge results from the legacy nested structure back into the matching
+/// flat client PM entries.
+///
+/// The `match` arms mirror those in `flat_pms_to_legacy_for_surcharge` — only PM types
+/// that were included in the legacy slice will find a match in `legacy` and get data.
+/// Skipped types (Wallet, Bank, None) leave the flat PM unchanged.
+///
+/// **Extending to a new PM type**: add an arm here that reads `pmt.surcharge_details`
+/// (for PM-level surcharge) or the relevant nested field, and writes it back.
+#[cfg(feature = "v1")]
```
XyneSpaces, line 4816: :warning: **Potential panic in production**: `card_network_surcharge_details.insert(...)` on a `HashMap` could panic if the map is mutated concurrently, though unlikely here.

More importantly, this code assumes `SurchargeDetailsResponse` is `Clone`. While the derive macro likely provides this, the dependency isn't visible in this file. Consider adding an explicit `Clone` bound check or ensure the type derives Clone in its definition.

XyneSpaces (commented):
## Review Findings

This PR introduces per-network surcharge details to the client payment methods list endpoint. Below are the findings organized by severity.

crates/api_models/src/payment_methods.rs
```diff
@@ -2912,16 +2912,26 @@ pub struct PaymentMethodListIntentData {
...
     },
     /// Bank payment method (bank_redirect) — contains supported bank names
     Bank {
         #[schema(value_type = Vec<BankNames>)]
         bank_names: Vec<common_enums::BankNames>,
     },
```
XyneSpaces, line 2927: :warning: **Breaking API change — removed fields from ResponsePaymentMethodsEnabledForClient**

The `collect_shipping_details_from_wallets` and `collect_billing_details_from_wallets` fields were removed from the struct and moved into a `Wallet` variant of the enum. This is a breaking API change for any clients that:
1. Deserialize responses expecting these fields at the top level
2. Have pattern matching that doesn't account for the new `Wallet` variant

Verify this endpoint is only consumed by the Hyperswitch SDK (which is updated in lockstep) and not by external API consumers.

crates/router/src/core/payment_methods/cards.rs
```diff
@@ -3420,10 +3424,11 @@ impl MerchantEnabledPmsContext {
...
                     payment_method_type: *payment_method_type,
-                    data: Some(PaymentMethodSubtypeSpecificDataForClient::Card { card_networks }),
+                    data: Some(PaymentMethodSubtypeSpecificDataForClient::Card {
+                        card_networks,
+                        card_network_surcharge_details: HashMap::new(),
+                    }),
                     payment_experience: None,
-                    collect_shipping_details_from_wallets: None,
-                    collect_billing_details_from_wallets: None,
                 });
```
XyneSpaces, line 3432: :bulb: **Documentation comment references non-existent Redis concern**

The doc comment states "so Redis stays consistent with what the response actually contains" — but there's no evidence in this PR that Redis is being used to cache these surcharge results. Either:
1. Remove the Redis reference if it's inaccurate
2. Add context about where/how Redis caching happens if it's a real concern

The comment creates confusion about the architectural constraints this code operates under.

crates/router/src/core/payment_methods/cards.rs
```diff
@@ -3471,8 +3474,6 @@ impl MerchantEnabledPmsContext {
                     payment_method_type: *payment_method_type,
                     data: None,
                     payment_experience: None,
```
XyneSpaces, line 3476: :warning: **Silent skip of surcharge for non-card payment methods**

The function silently continues for Wallet, Bank, and None payment method types without any logging. If a merchant configures surcharge rules for wallets or bank redirects, this will silently fail to calculate surcharges with no indication in logs.

Add an `info!` or `debug!` log when skipping surcharge calculation for these payment method types, or implement the missing surcharge support for completeness.

```rust
Some(PaymentMethodSubtypeSpecificDataForClient::Wallet { .. }) => {
    logger::debug!("Surcharge calculation skipped for Wallet payment method");
    continue;
}
```

crates/router/src/core/payment_methods/client.rs
```diff
@@ -65,6 +65,7 @@ fn to_client_pm(pm: CustomerPaymentMethod) -> CustomerPaymentMethodForClient {
...
     }
 }
```
XyneSpaces, line 71: :warning: **Incomplete surcharge implementation for saved payment methods**

The `to_client_pm` function sets `surcharge_details: None` for all customer payment methods. This means saved cards won't show surcharge information even when surcharge rules are configured.

If the intent is to support surcharges for saved cards, implement the surcharge lookup. If not, remove the field from the struct to avoid misleading consumers.

crates/router/src/core/payment_methods/cards.rs
```diff
@@ -4742,6 +4741,120 @@ pub async fn call_surcharge_decision_management(
...
+            Some(PaymentMethodSubtypeSpecificDataForClient::Card { card_networks, .. }) => {
+                card_networks
+                    .iter()
+                    .map(|cn| CardNetworkTypes {
+                        card_network: cn.clone(),
+                        eligible_connectors: vec![],
+                        surcharge_details: None,
+                    })
+                    .collect()
+            }
```
XyneSpaces, line 4772: :mag: **Nit: Inefficient O(n*m) lookup pattern**

The nested loop searches through `legacy` for each flat PM entry, resulting in O(n*m) complexity where n = flat PMs and m = legacy PMs. Since both vectors are typically small (< 100 items), this is acceptable, but consider using a HashMap keyed by `(payment_method, payment_method_type)` for O(n+m) complexity if surcharge lookup becomes a bottleneck.

XyneSpaces (commented):
## Review Summary

**Classification:** Feature — Per-network surcharge for client payment methods list  
**Risk Level:** Medium (changes API contract, touches surcharge calculation logic)  
**Files Changed:** 7 files (API models, OpenAPI spec, router core)

---

### Findings Overview

| Severity | Count | Issue |
|----------|-------|-------|
| :warning: | 1 | Removed fields from API response without backward compatibility concern |
| :bulb: | 1 | Empty HashMap default for surcharge details may confuse SDK clients |
| :mag: | 1 | Missing tests for new surcharge conversion functions |

**Verdict:** Please address the breaking API change concern and consider adding test coverage.

PR #12729, opened 2026-06-11
fix(connectors): graceful response deserialization for paybox

sai-harsha-vardhan:
## Summary
This PR applies graceful response deserialization fixes to the Paybox connector to prevent production SEV incidents caused by unexpected API changes.

### Changes Made
- **Fix 2 (Catch-all enum variant)**: Added `#[serde(other)] Unknown` variant to `PayboxStatus` enum, which is used in business logic for both payment status and refund status mapping.
- Added warning logs and safe defaults for the Unknown variant:
  - `From<PayboxStatus> for AttemptStatus`: Maps Unknown → `AttemptStatus::Pending` with warning log
  - `From<PayboxStatus> for RefundStatus`: Maps Unknown → `RefundStatus::Failure` with warning log

### Why
Previously, if Paybox added a new status value (e.g., a new transaction state), our serde deserialization would fail with `HE_00` (`server_not_available`). With the catch-all variant, unknown statuses are gracefully handled without hard failures.

### Notes
- Fix 1 (deny_unknown_fields): Not applicable — no response structs in the Paybox connector had `#[serde(deny_unknown_fields)]`.
- Fix 3 (opaque optional types): Not applicable — all fields on response structs are actively used in business logic (`TryFrom` implementations).

### DIFF
--- a/crates/hyperswitch_connectors/src/connectors/authorizedotnet/transformers.rs
+++ b/crates/hyperswitch_connectors/src/connectors/authorizedotnet/transformers.rs
@@ -2248,6 +2248,7 @@ impl From<AuthorizedotnetWebhookEvent> for SyncStatus {
             AuthorizedotnetWebhookEvent::PriorAuthCapture => Self::SettledSuccessfully,
             AuthorizedotnetWebhookEvent::VoidCreated => Self::Voided,
             AuthorizedotnetWebhookEvent::RefundCreated => Self::RefundSettledSuccessfully,
+            AuthorizedotnetWebhookEvent::Unknown => Self::GeneralError,
         }
     }
 }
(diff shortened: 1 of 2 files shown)


XyneSpaces (commented):
## Review: Graceful Response Deserialization for Paybox

### :white_check_mark: Approved

This PR applies the standard graceful deserialization fix to the Paybox connector:

- Added `#[serde(other)] Unknown` variant to `PayboxStatus` enum
- Added warning logs and appropriate fallback mappings:
  - `Unknown` → `AttemptStatus::Pending` for payments (preserves existing state)
  - `Unknown` → `RefundStatus::Failure` for refunds (safe default)

### :bulb: Findings

**1. Appropriate fallback choices**
- Payment status `Unknown` → `Pending` is correct because we don't know the actual state and shouldn't mark it as failed/charged incorrectly
- Refund status `Unknown` → `Failure` is a safe default — refunds are idempotent and can be retried if marked as failed

**2. Warning logs**
Both mappings include appropriate warning logs so operators can identify when unknown statuses are received.

**3. Authorize.Net webhook fix**
The PR also includes a small fix to add `Unknown => Self::GeneralError` mapping in the Authorize.Net webhook event conversion, which was likely missed in PR #12731.

### :mag: Notes

- Paybox's status values include accented characters ("Refusé") — the existing `#[serde(rename)]` handles this correctly
- The `Unknown` variant follows the same pattern as other connectors for consistency

**Verdict**: Clean implementation. Ready to merge.

PR #12844, opened 2026-06-18
feat(common_enums): add domain status framework with Unknown catch-all

### LINKED ISSUE #12843: feat(common_enums): domain status framework with internal-only `Unknown` for connector-response deserialization resilience
## Problem

Connectors occasionally return status values we do not yet model (a new processor state, a backward-incompatible API change, a typo). Mapping such a value directly onto a storage status enum (`AttemptStatus` / `RefundStatus` / `DisputeStatus`) hard-fails deserialization and drops the payment — the root cause behind the connector response deserialization incident class.

The storage enums are Postgres `db_enum`s that are also serialized to merchants, so they must **not** gain an `Unknown` variant (it would become persistable + merchant-visible, and forces a 359-file exhaustive-match fan-out + DB migration).

## Approach (agreed in design thread)

Introduce a parallel **domain** status type for each storage status enum. The domain type mirrors every storage variant and adds a single internal-only `#[serde(other)] Unknown` catch-all, so unrecognised values deserialize gracefully instead of erroring.

Contract:
- `Unknown` is **internal-only** — never persisted, never sent to merchants. Storage enums are untouched.
- `From<storage> for domain` is infallible (every storage state is known).
- `TryFrom<domain> for storage` **errors** on `Unknown`, so a missing previous-state resolution surfaces loudly instead of silently corrupting state.
- `resolve_or_keep(previous)` merges the connector response with the prior state: an `Unknown` is replaced by the previously known storage state before persistence.

A `domain_status_enum!` macro generates the enum + both conversions + helpers from a single variant list, so the two conversions can never drift, and adding a storage variant breaks the macro invocation at compile time until the mirror is updated.

## Affected files

- `crates/common_enums/src/domain_status.rs` *(new)* — `domain_status_enum!` macro, `UnknownStatusError`, and the three domain mirrors (`AttemptStatusDomain`, `RefundStatusDomain`, `DisputeStatusDomain`) + TDD tests.
- `crates/common_enums/src/lib.rs` — register `pub mod domain_status;`.

## Acceptance criteria

- [x] Unrecognised connector status deserializes to `Unknown` instead of hard-failing.
- [x] Known statuses deserialize / convert to the matching storage variant.
- [x] `Unknown -> storage` conversion returns an error (does not silently default).
- [x] `resolve_or_keep(previous)` replaces `Unknown` with the prior state and preserves known statuses.
- [x] Storage enums unchanged → no DB migration, no fan-out.

## Out of scope

- Wiring the domain types into individual connector transformers / the UCS response path (follow-up PRs, per-connector).
- The per-connector `deny_unknown_fields` removal campaign already in flight.

## Test plan

`cargo test -p common_enums domain_status` — 9 unit tests covering graceful deserialization, the storage-conversion guard, and previous-state resolution. All passing.

XyneSpaces:
Closes #12843

## What

Adds `#[derive(router_derive::DomainStatus)]`. Applied to a storage status enum, it generates a parallel *domain* type (`<Name>Domain`) that mirrors every storage variant and adds a single internal-only `#[serde(other)] Unknown` catch-all. Unrecognised connector statuses deserialize into `Unknown` instead of hard-failing and dropping the payment.

This is the **core framework** PR; wiring it into connector transformers / the UCS path is per-connector follow-ups.

## Design

- **Variants come from the storage enum — single source of truth.** The derive reads the variant list straight off the enum it is applied to; developers never re-list variants, so the storage type and its domain mirror can never drift. Add a variant to `AttemptStatus` and `AttemptStatusDomain` gains it automatically on the next build.
- **Storage enums stay untouched.** `AttemptStatus` / `RefundStatus` / `DisputeStatus` are Postgres `db_enum`s also serialized to merchants — adding `Unknown` there would make it persistable + merchant-visible and force a DB migration plus a large exhaustive-match fan-out. A derive only *adds* a new type; it cannot modify its input, so the storage enum is left exactly as-is. `Unknown` lives only on the generated mirror.
- The derive mirrors the storage enum's serde container attributes (e.g. `rename_all`) and carries over per-variant `serde` aliases/renames, so the domain type's wire representation matches the storage type's. It generates, for storage `S`:
  - `From<S> for SDomain` — infallible (every storage state is known).
  - `TryFrom<SDomain> for S` — returns `UnknownStatusError` on `Unknown`, so a missing previous-state resolution surfaces loudly instead of corrupting state.
  - `resolve_or_keep(previous)` — replaces an `Unknown` with the previously known storage state before persistence (the connector-response ⊕ previous-state merge); known statuses pass through.
  - `is_unknown()`, `to_storage()` helpers.

## Files

- `crates/router_derive/src/macros/domain_status.rs` *(new)* — `#[derive(DomainStatus)]` codegen.
- `crates/router_derive/src/macros.rs` — register module.
- `crates/router_derive/src/lib.rs` — `#[proc_macro_derive(DomainStatus)]` entry point.
- `crates/common_enums/src/enums.rs` — add `router_derive::DomainStatus` to the derive list of `AttemptStatus`, `RefundStatus`, `DisputeStatus` (one line each; enums otherwise unchanged).
- `crates/common_enums/src/domain_status.rs` — hand-written `UnknownStatusError` + module docs + tests.
- `crates/common_enums/src/lib.rs` — `extern crate self as common_enums;` (so the macro-emitted `common_enums::domain_status::UnknownStatusError` path resolves in-crate) and `pub mod domain_status;`.

## Tests (TDD)

`cargo test -p common_enums domain_status` → **9 passed; 0 failed**. Covers:
- unrecognised status → `Unknown` (no hard-fail)
- known status → matching variant
- `Unknown -> storage` errors loudly (the bug guard)
- `resolve_or_keep(previous)` replaces `Unknown` and preserves known statuses
- storage→domain→storage round-trip for known statuses
- same contract for refund + dispute mirrors

`cargo build -p router_derive`, `cargo check -p common_enums`, and `cargo clippy -p router_derive -p common_enums --all-targets` — clean, no warnings from this change.

## Risk

Additive: a new derive in `router_derive`, one extra derive on three existing enums, a new module + one `extern crate self` line in `common_enums`. No existing variant, match, schema, or migration is touched.


XyneSpaces (commented):
## PR Review: Domain Status Framework

**Classification:** core-flow-framework · baseline  
**Risk Level:** Low (purely additive, new module only)

### Findings Overview
- :rotating_light: 0 Critical
- :warning: 1 Warning
- :bulb: 1 Suggestion

---

### :warning: Warning: Consider exposing `Unknown` variant for explicit matching
**File:** `crates/common_enums/src/domain_status.rs` (macro-generated code)

The `Unknown` variant is marked with `#[serde(other)]` which is correct for deserialization resilience. However, downstream code may need to explicitly match on `Unknown` to trigger the `resolve_or_keep` logic. Consider if the macro should also generate an `is_resolved()` helper that returns `true` when the status is not `Unknown` — this makes the state check more explicit at call sites.

**Current:**
```rust
pub fn is_unknown(self) -> bool { matches!(self, Self::Unknown) }
```

**Consider adding:**
```rust
pub fn is_resolved(self) -> bool { !self.is_unknown() }
```



---

### :bulb: Suggestion: Document usage pattern for connector transformers
**File:** `crates/common_enums/src/domain_status.rs` (module-level docs)

The module docs explain the *what* and *why* well. Consider adding a brief *how* example showing the typical transformer usage pattern:

```rust
// Deserialize connector response
let domain: AttemptStatusDomain = serde_json::from_str(connector_status)?;

// Merge with previous state if unknown
let resolved = domain.resolve_or_keep(payment.attempt_status);

// Convert back to storage (errors if still Unknown)
let new_status = resolved.to_storage()?;
```

This helps connector developers adopt the pattern correctly without reading the full module source.



---

### Baseline Verification

| Check | Status |
|-------|--------|
| No unwrap/expect in non-test code | ✅ Pass |
| No unsafe blocks | ✅ Pass |
| error_stack chain preserved | ✅ Pass (uses thiserror properly) |
| Exhaustive match arms | ✅ Pass (macro-generated) |
| TryFrom errors on Unknown | ✅ Pass (correct safety guard) |
| Unit test coverage | ✅ Pass (9 tests, good scenarios) |

### Overall Assessment

This is a well-designed framework PR that:
- ✅ Provides graceful degradation for unknown connector statuses
- ✅ Prevents silent data corruption via `UnknownStatusError`
- ✅ Keeps storage enums clean (no `Unknown` variant)
- ✅ Follows macro patterns used elsewhere in the codebase
- ✅ Includes comprehensive tests for all three status types

**No blocking issues. Ready for merge.**

PR #12850, opened 2026-06-19
fix(grpc_client): apply per-RPC timeout to unified connector service calls

gopikrishna000:
## Type of Change

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

`build_grpc_client!` only bounded the **initial channel connect** via `connection_timeout`. Once the channel was established, every RPC to the unified connector service (authorize, sync, capture, etc.) was awaited with **no per-request deadline** — so a slow or stuck call could block the payment flow indefinitely.

- Add a `request_timeout` field to `[grpc_client.unified_connector_service]`, applied via `Endpoint::timeout` so it bounds **every** RPC on the channel (not just connect).
- `#[serde(default = "default_ucs_request_timeout")]` → defaults to `35s` when unset, so the bound applies even without explicit config; overridable per environment.
- `build_grpc_client!` now takes the connect timeout and the per-request timeout separately and builds the channel through an `Endpoint`.

`35s` is kept just above the connector-side request timeout (`30s`) so the connector timeout surfaces first with its real reason, rather than this layer masking it with a bare gRPC deadline.

### Additional Changes

Wired into:
- `config/config.example.toml`
- `config/development.toml`
- `config/deployments/env_specific.toml`

## Motivation and Context

A connector RPC that hangs after the channel is connected has no upper bound today, so it can stall the payment flow indefinitely. This adds a per-RPC deadline that is backward-compatible (existing configs without `request_timeout` fall back to `35s`).

## How did you test it?

- `cargo check -p external_services --features v1` passes.
- Backward-compatible: existing configs without `request_timeout` fall back to `35s`.

## Checklist


XyneSpaces (commented):
## Code Review: gRPC Per-RPC Timeout Fix

This PR addresses a critical issue where gRPC calls to the Unified Connector Service lacked per-request timeouts, potentially causing indefinite blocking of payment flows.

### Critical Observation

**Macro signature change**: The `build_grpc_client!` macro now requires 5 parameters instead of 4. This is a breaking change for any out-of-tree code using this macro. Consider:
1. Whether external contributors or forks might be affected
2. Adding a migration note to the PR description or CHANGELOG

### Questions

1. **Timeout relationship**: The 35s timeout is intentionally > 30s connector timeout. Is there telemetry/metrics to verify this ordering works as intended in production (i.e., we see connector timeouts, not gRPC timeouts)?

2. **Error handling**: When a gRPC request times out at 35s, how does this propagate to the caller? Is it distinguishable from a connector-side failure for debugging purposes?

crates/external_services/src/grpc_client/unified_connector_service.rs
```diff
@@ -206,15 +211,17 @@ impl UnifiedConnectorServiceClient {
...
 
                 let payment_service_client = build_grpc_client!(
                     payments_grpc::payment_service_client::PaymentServiceClient<
                         tonic::transport::Channel,
                     >,
                     "payment_service_client",
                     uri,
-                    timeout
+                    connection_timeout,
+                    request_timeout
```
XyneSpaces, line 248: **Naming consideration**: The field name `request_timeout` is clear, but consider if `rpc_timeout` or `per_request_timeout` would be more explicit about what this bounds (individual RPCs vs the entire operation which might involve retries). Not blocking, just a suggestion for clarity.

crates/external_services/src/grpc_client/unified_connector_service.rs
```diff
@@ -300,7 +316,8 @@ impl UnifiedConnectorServiceClient {
                     >,
                     "surcharge_service_client",
                     uri,
-                    timeout
+                    connection_timeout,
+                    request_timeout
```
XyneSpaces, line 344: **Macro expansion hygiene**: The macro captures `$request_timeout` and uses it in `endpoint.timeout()`. Verify this doesn't cause issues if the caller passes a complex expression (e.g., `compute_timeout()`) that gets evaluated multiple times due to macro expansion. Consider adding `let` bindings inside the macro to eagerly evaluate arguments.

config/config.example.toml
```diff
@@ -1340,6 +1340,7 @@ url = "http://localhost:8080"           # Open Router URL
 [grpc_client.unified_connector_service]
```
XyneSpaces, line 1342: **Documentation**: Consider adding a comment explaining *why* 35s was chosen (connector timeout is 30s) to help operators tune this appropriately.

XyneSpaces:
## Code Review: gRPC Per-RPC Timeout Fix

This PR correctly addresses a critical issue where UCS calls had no per-request deadline after initial connection establishment. The fix properly separates connection timeout from RPC timeout using tonic's `Endpoint::timeout`.

### 🟢 Positive Highlights

1. **Correct use of tonic API**: Using `Channel::builder().timeout()` properly bounds every RPC on the channel, not just the connect handshake.

2. **Sensible default value**: 35s is well-chosen - slightly above the 30s connector timeout allows underlying connector errors to surface first.

3. **Backward compatible**: The `#[serde(default)]` attribute ensures existing deployments without explicit `request_timeout` continue to work.

4. **Consistent config wiring**: All three config files (example, development, env_specific) are updated consistently.

### ⚠️ Suggestions

1. **Documentation** (`crates/external_services/src/lib.rs` line ~65): Consider adding a comment explaining the 35s choice directly in the const definition for future maintainers:
```rust
/// Default per-RPC timeout (seconds) for unified connector service calls.
/// Set to 35s to stay above typical 30s connector timeouts, allowing 
/// underlying connector errors to surface before this deadline.
```

2. **Documentation** (`crates/external_services/src/grpc_client/unified_connector_service.rs` line ~62): Consider clarifying the relationship with `connection_timeout`:
```rust
/// Per-RPC timeout (seconds) for calls to the unified connector service.
/// Unlike `connection_timeout` which only bounds the initial handshake,
/// this bounds every individual RPC after the channel is established.
```

3. **Metrics/Observability**: Consider adding a metric or log when requests hit the timeout, distinct from other connection errors, to help identify UCS slowness issues in production.

4. **Retry behavior**: Verify that tonic's timeout produces errors that the existing retry logic (if any) handles appropriately, or document that these are terminal failures.

Overall, this is a well-implemented fix for an important reliability issue. 👍

XyneSpaces:
## Review Summary

This PR adds configurable request timeouts for gRPC calls to the Unified Connector Service (UCS). Overall, this is a solid improvement that separates connection timeouts from per-RPC timeouts.

### Findings

1. **Good practice**: The `#[serde(default = "default_ucs_request_timeout")]` ensures backward compatibility for existing deployments that don't have this config field.

2. **Macro improvement**: The updated `build_grpc_client!` macro now properly distinguishes between `connection_timeout` and `request_timeout`, applying them to the correct operations.

3. **Config consistency**: All three config files (example.toml, env_specific.toml, development.toml) have been updated with the new `request_timeout = 35` value.

### Suggestions

- Consider documenting why `35` seconds was chosen as the default - is this based on UCS service SLAs or empirical data?
- The addition of `payconex` to the `ucs_only_connectors` list in development.toml is unrelated to the timeout change - consider separating this into a different PR for cleaner history.

No blocking issues found.

Shubhodip900 (dismissed):

crates/external_services/src/grpc_client/unified_connector_service.rs
```diff
@@ -59,6 +59,10 @@ pub struct UnifiedConnectorServiceClientConfig {
     /// Contains the connection timeout duration in seconds
     pub connection_timeout: u64,
 
+    /// Per-RPC timeout (seconds) for calls to the unified connector service.
+    #[serde(default = "default_ucs_request_timeout")]
+    pub request_timeout: u64,
```
hrithikesh026: Nit: 
A more robust approach would be to have a wrapper over u64. The default for that type can be defined as a Default trait implementation. Also you can specify any limits if needed on that type itself.

```suggestion
    /// Per-RPC timeout (seconds) for calls to the unified connector service.
    pub request_timeout: UcsRequestTimeoutInSeconds,
```

hrithikesh026 (dismissed):

Shubhodip900 (approved):

hrithikesh026 (approved):

PR #12853, opened 2026-06-20
fix(common_utils): compute surcharge percentage in Decimal, not f64

### LINKED ISSUE #12854: [BUG] Surcharge percentage is computed in f64 and can be off by a minor unit
### Bug Description

`Percentage::apply_and_ceil_result` in `crates/common_utils/src/types.rs` applies a percentage to a `MinorUnit` amount using `f64` arithmetic:

```rust
let percentage_f64 = f64::from(self.percentage);
let result = (amount as f64 * (percentage_f64 / 100.0)).ceil() as i64;
```

`percentage / 100.0` is not exactly representable in binary floating point (e.g. `2.22 / 100`), so the product can sit just above an integer and `.ceil()` rounds the wrong way — producing a result one (or more) minor units larger than the exact value. This routine backs surcharge calculation (`crates/router/src/core/payment_methods/surcharge_decision_configs.rs`), so the drift is money charged to customers.

### Expected Behavior

A percentage applied to an integer minor-unit amount should equal the exact value:
- 2.22% of `100_000_000` minor units = `2_220_000`
- 33.33% of `100_000_000` minor units = `33_330_000`

### Actual Behavior

The `f64` path returns:
- 2.22% of `100_000_000` → `2_220_001` (off by 1)
- 2.22% of `10_000_000` → `222_001` (off by 1)
- 33.33% of `100_000_000` → `33_330_002` (off by 2)

Reproduced with the same IEEE-754 `f64` arithmetic the code uses (including the `f32` storage of the percentage).

### Steps To Reproduce

Call `Percentage::<2>::from_string("2.22".to_string()).unwrap().apply_and_ceil_result(MinorUnit::new(100_000_000))` and observe `2_220_001` instead of `2_220_000`.

### Possible Solution

Compute the percentage in `Decimal` (`rust_decimal` is already a dependency and used elsewhere in the file), rounding the percentage to its validated `PRECISION` before multiplying, then `.ceil()`. A fix is proposed in #12853.

### Environment

Observed on `main` at commit `a828fb2`.

ManojLingala:
`Percentage::apply_and_ceil_result` converted the amount to `f64` and multiplied by `percentage / 100.0` before `.ceil()`. `percentage / 100.0` is not exactly representable in binary floating point (e.g. `2.22 / 100`), so the result could be off by a minor unit on real surcharge amounts:

```
2.22%  of 100_000_000 -> 2_220_001   (exact: 2_220_000)
33.33% of 100_000_000 -> 33_330_002  (exact: 33_330_000)
```

This routine backs surcharge calculation (`crates/router/src/core/payment_methods/surcharge_decision_configs.rs`), so the drift is charged to customers.

Closes #12854.

## Type of Change

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

`Percentage::apply_and_ceil_result` (`crates/common_utils/src/types.rs`) now computes the surcharge in `Decimal` rather than `f64`. The percentage (stored as `f32`, validated to at most `PRECISION` decimal places) is converted to `Decimal` and `round_dp(PRECISION)` recovers its exact intended value; the amount is converted with `Decimal::from_i64`; the result is `(amount * pct / 100).ceil().to_i64()`. `.ceil()` semantics are preserved. The now-redundant `#[allow(clippy::as_conversions)]` is removed since the `as` casts are gone.

### Additional Changes

## Motivation and Context

Surcharge amounts are charged to customers, so a one-minor-unit rounding error from binary floating point is a real money-correctness issue. Fixes #12854. `rust_decimal` is already a dependency and is used for other money conversions in the same file, so the fix adds no new dependency.

## How did you test it?

- Added a unit test `percentage_apply_and_ceil_is_decimal_exact` asserting `2.22%` of `100_000_000` = `2_220_000`, `33.33%` of `100_000_000` = `33_330_000`, and that a genuine fraction (`10%` of `105` minor units) still ceils to `11`.
- The off-by-one/-two was reproduced independently in IEEE-754 `f64` (the arithmetic the old code used).
- Disclosure: I do not have a local Rust toolchain, so I have **not** run `cargo fmt` / `clippy` / `test` locally — please rely on CI. The change uses only `rust_decimal` APIs already imported (`FromPrimitive` / `ToPrimitive`) and already used in this file.

## Checklist


XyneSpaces:
[should-fix] Potential duplicate error reporting in `apply_and_ceil_result`

On the error paths for `Decimal::from_f64` and `from_i64` conversions, the code uses `report!(PercentageError::UnableToApplyPercentage { ... })` which creates a new error report. However, these operations return `None` only when the value cannot be represented as a Decimal (e.g., NaN, Infinity), not for typical surcharge calculations.

Consider whether these error variants accurately describe the failure mode, or if a more specific error like `InvalidPercentageValue` or `InvalidAmount` would help operators debug faster when these edge cases occur.

ManojLingala:
Thanks, good catch — addressed in 914e4ab:

- `i64 -> Decimal` is infallible (every `i64` fits in a `Decimal`, the same as the `Decimal::from(100)` used just below), so I dropped that error path entirely and use the infallible `Decimal::from(amount)`.
- `Decimal::from_f64(percentage)` only returns `None` for a non-finite percentage, which `from_string` validation already rules out — it now reports `InvalidPercentageValue`, which accurately names the failure.
- `UnableToApplyPercentage { percentage, amount }` is kept only on the final `to_i64()` of the computed result, the one place it actually describes the failure.

So the duplicated `UnableToApplyPercentage` is gone and each remaining error path names its real failure mode.

XyneSpaces:
## Code Review: Surcharge Percentage Decimal Fix

This PR addresses a critical financial precision bug by replacing f64 arithmetic with Decimal for surcharge calculations. The fix is well-reasoned and the tests comprehensively verify the correction.

### Overall Assessment
✅ **Correctness**: The Decimal-based approach eliminates binary floating-point representation errors.  
✅ **Testing**: Unit tests verify exactness for problematic percentages (2.22%, 33.33%) and ceiling behavior.  
✅ **Safety**: Proper error handling for Decimal conversion failures.

### Inline Comments

**`crates/common_utils/src/types.rs:107-109`**
```rust
let percentage_decimal = Decimal::from_f64(f64::from(self.percentage))
    .map(|percentage| percentage.round_dp(u32::from(PRECISION)))
    .ok_or_else(|| report!(PercentageError::InvalidPercentageValue))?;
```

**[nit]** The conversion path `f64 -> Decimal` via `from_f64` is acceptable here because the subsequent `round_dp(PRECISION)` sanitizes any binary representation artifacts. Since the percentage is already validated to have at most `PRECISION` decimal places, this recovers the exact intended value. Consider adding a brief comment explaining this reasoning for future maintainers.

---

**`crates/common_utils/src/types.rs:113-117`**
```rust
let result = (amount_decimal * percentage_decimal / Decimal::from(100))
    .ceil()
    .to_i64()
    .ok_or_else(|| {
```

**[should-fix]** The `to_i64()` failure case returns `UnableToApplyPercentage`, but the actual failure here is that the computed result overflowed `i64`. Consider a more specific error variant like `PercentageError::ComputationOverflow` or include the computed value in the error context for debugging.

---

**`crates/common_utils/src/types.rs:127-129` (tests)**
```rust
// Surcharge percentages must be applied exactly. The previous f64 implementation
// returned 2_220_001 / 33_330_002 for these cases because `pct / 100.0` is not
// representable in binary floating point.
```

**[positive]** Excellent test documentation explaining the bug and expected behavior.

---

### Suggestions (Non-blocking)

1. **Edge case consideration**: The test coverage is good but could include:
   - Very small amounts (1 minor unit) with fractional percentages
   - Maximum `i64` amount to verify overflow handling
   - Boundary percentage values (0%, 100%, 999.99%)

2. **Performance note**: Decimal operations are slower than f64. Given this is in hot path for payment surcharge calculations, consider documenting the trade-off in a code comment.

### Summary
This is a solid fix for a precision-critical financial calculation. The Decimal approach is appropriate for monetary calculations where exactness trumps performance. Recommend merge after addressing the error variant suggestion or documenting the rationale.

XyneSpaces:
[should-fix] Verify rounding strategy in `apply_and_ceil_result`

The change uses `.round_dp(u32::from(PRECISION))` on the percentage value. Please verify this produces the expected ceiling behavior. The `round_dp` method uses banker's rounding (midpoint to nearest even), not always-away-from-zero.

For true ceiling of surcharge calculations, consider using:
```rust
.round_dp_with_strategy(u32::from(PRECISION), RoundingStrategy::AwayFromZero)
```

Or confirm that the default rounding behavior meets your requirements for financial calculations. The added tests cover the exact cases well — just ensure the rounding strategy aligns with the intended ceiling semantics.

ManojLingala:
Good question — I dug into it. There are two separate roundings here, and the one `round_dp` does is **not** the ceiling:

1. **The surcharge ceiling** is the `.ceil()` on `(amount_decimal * percentage_decimal / 100)`. `Decimal::ceil` rounds toward +∞ (always up), so the always-round-up surcharge semantics are correct, independent of how the percentage is rounded.

2. **`round_dp(PRECISION)` is applied to the percentage, not the result** — it only snaps off the `f32 -> f64` conversion noise to recover the percentage's exact validated value. `self.percentage` is an `f32`, so `2.22` widens to `2.2200000286…`; without recovering it, `100_000_000 * 2.2200000286 / 100 = 2_220_000.000…286`, whose `.ceil()` is the spurious `2_220_001` this PR fixes.

For that recovery I need **round-to-nearest**, which is why I'd keep the default rather than switch to `AwayFromZero`: `RoundingStrategy::AwayFromZero` rounds any non-zero fraction up, so it would turn `2.2200000286` into `2.23` and corrupt the percentage (and the result). `MidpointAwayFromZero` would be behaviorally identical to the current `MidpointNearestEven` here — the percentage is validated to ≤ `PRECISION` decimal places (`is_valid_precision_length`), so the value is an exact ≤PRECISION-digit number plus ~1e-7 of float noise and is never *at* a midpoint, so the tie-break never triggers. The committed tests (`2.22%` of `100_000_000 -> 2_220_000`, `33.33% -> 33_330_000`) would fail if the percentage rounding were off.

I pushed a comment (0ab1285) making the split explicit: `round_dp` recovers the percentage's exact value; the surcharge ceiling is the `.ceil()` on the result. Thanks for the careful look!

ManojLingala:
Thanks — added in b046876.

**Edge-case tests** (`percentage_apply_and_ceil_edge_cases`):
- `0%` → `0` and `100%` → the full amount (boundaries).
- A tiny surcharge on `1` minor unit (`2.22%`) still ceils up to `1`.
- Overflow guard: the boundary amount `i64::MAX / 10000` is accepted (`100%` of it round-trips), and `i64::MAX` is rejected with `UnableToApplyPercentage`.

On the `999.99%` case — that one can't actually occur: `Percentage` validates the range to `0..=100` in `is_valid_range`, so `from_string("999.99")` returns an error. I added an assertion documenting that the `>100%` percentage is rejected at construction, so a surcharge can never exceed 100%.

**Performance:** good call — I added a comment noting the trade-off. For context, a surcharge is computed once per payment (not in a tight loop), so the few extra `Decimal` ops are negligible next to the network/DB cost of the payment, and the precision (no minor-unit drift) is worth it. Happy to revisit if you'd want this micro-optimized.

XyneSpaces:
## Code Review: Decimal Precision Fix for Surcharge Calculation

### Overall Assessment
This is a high-quality fix for a financial precision bug. The migration from `f64` to `Decimal` eliminates off-by-one minor unit errors in surcharge calculations. The code is well-documented with extensive comments explaining the rationale, and comprehensive unit tests have been added.

### Findings

#### [nit] Documentation: `round_dp` clarification
**Location:** Line ~113-114

The comment states "The percentage is validated to at most PRECISION decimal places, so `round_dp` recovers its exact value". While accurate, it might help future maintainers to note that `round_dp` is still necessary even after validation because `Decimal::from_f64` can introduce representation artifacts for values like `2.22`.

```rust
// Consider adding:
// Note: Even validated percentages need round_dp because from_f64 
// may introduce binary representation artifacts.
```

---

#### [should-fix] Test coverage: Missing negative amount guard test
**Location:** `percentage_apply_and_ceil_edge_cases` test

The implementation validates that `amount <= max_amount` (where `max_amount = i64::MAX / 10000`), but there's no test asserting the error case at the boundary. Consider adding:

```rust
// Test boundary rejection
let over_max = MinorUnit::new(i64::MAX / 10000 + 1);
assert!(full.apply_and_ceil_result(over_max).is_err());
```

This ensures the guard clause is tested and prevents regression if the limit calculation changes.

---

#### [nit] Error variant selection for `from_f64` failure
**Location:** Lines ~116-118

The code maps `from_f64` failure to `InvalidPercentageValue`. This is semantically correct for non-finite values, but consider: should we ever expect non-finite values given the `Percentage` type already validates its range? If this error path is truly unreachable due to prior validation, a `panic!` with an explanatory message might be more appropriate to catch programming errors early.

However, using `InvalidPercentageValue` is defensively correct and matches the existing error handling style.

---

#### ✅ Positive observations

1. **Excellent commit history** - The iterative refinement shown in commits (error variant refinement, documentation improvements) demonstrates careful attention to review feedback.

2. **Performance acknowledgment** - The inline comment acknowledging Decimal's slower performance vs f64 shows engineering maturity. Given this runs once per payment, the correctness trade-off is justified.

3. **Test quality** - The test cases directly reference the bug being fixed (2.22% of 100M, 33.33% of 100M) making regression detection clear.

4. **Removed clippy suppression** - The `#[allow(clippy::as_conversions)]` removal is a nice cleanup since the `as` casts are eliminated.

---

### Verdict
Approve with minor nits. The core fix is correct and well-tested. Consider adding the boundary error test for completeness.

XyneSpaces:
## PR Review Summary

This is a **well-crafted fix** for a legitimate financial calculation precision bug. The change from `f64` to `Decimal` for surcharge calculations is the correct approach for monetary arithmetic.

---

### ✅ Good Practices Observed

1. **Comprehensive test coverage** — Edge cases include 0%, 100%, fractional amounts, single minor unit, and overflow boundaries
2. **Clear code comments** — Explains *why* the change was made (IEEE-754 representation limits)
3. **Removed clippy suppression** — `#[allow(clippy::as_conversions)]` was appropriately removed
4. **Proper error handling** — Both `from_f64` and `to_i64` failures are handled explicitly

---

### ⚠️ [should-fix] Verify `to_i64()` behavior on overflow

The code uses `to_i64()` after `ceil()`, but Decimal values exceeding `i64::MAX` will return `None` from `to_i64()`. While the `amount > max_amount` check prevents *input* overflow, the calculation:

```rust
(amount_decimal * percentage_decimal / Decimal::from(100)).ceil()
```

Could theoretically produce a value > `i64::MAX` for large inputs even within the checked bounds (e.g., 100% of `i64::MAX/10000` multiplied by a percentage > 10000%).

The current `max_amount` of `i64::MAX/10000` seems designed to prevent this, but **add a test case** verifying `to_i64()` returns `None` appropriately when the mathematical result exceeds `i64::MAX`:

```rust
#[test]
fn percentage_apply_ceil_overflow_to_i64() {
    // Verify graceful error when result exceeds i64 bounds
    let large_pct = Percentage::<2>::from_string("9999.99".to_string()).unwrap();
    let result = large_pct.apply_and_ceil_result(MinorUnit::new(i64::MAX / 100));
    assert!(result.is_err());
}
```

---

### 📋 [nit] Potential optimization opportunity (pre-existing)

Not introduced by this PR, but consider for future: the `percentage` field is stored as `f64` but always converted to `Decimal` or bounded by `PRECISION`. Storing as `Decimal` directly could eliminate repeated conversions. Out of scope for this fix.

XyneSpaces:
## Code Review: Surcharge Decimal Precision Fix

### Summary
This PR fixes a legitimate floating-point precision bug where `f64` arithmetic caused off-by-one errors in surcharge calculations. The migration to `rust_decimal::Decimal` is the correct approach for financial calculations.

### Findings

#### [should-fix] Error handling in `from_f64` conversion
**Location:** Lines 107-110

The error handling when `Decimal::from_f64()` returns `None` converts it to `PercentageError::InvalidPercentageValue`, which may not accurately describe the failure condition:

```rust
let percentage_decimal = Decimal::from_f64(f64::from(self.percentage))
    .map(|percentage| percentage.round_dp(u32::from(PRECISION)))
    .ok_or_else(|| report!(PercentageError::InvalidPercentageValue))?;
```

**Issue:** `from_f64()` only returns `None` for non-finite values (NaN, ±Infinity), which shouldn't occur given validation. However, if this somehow happens, `InvalidPercentageValue` is misleading—it implies the percentage format is wrong, not that an internal conversion failed.

**Suggestion:** Consider a more descriptive error variant like `InvalidPercentageValue` with a printable message explaining the conversion failure, or add context:

```rust
.ok_or_else(|| {
    report!(PercentageError::InvalidPercentageValue)
        .attach_printable("Percentage value could not be converted to decimal")
})?;
```

---

#### [nit] Test organization
**Location:** Lines 813-868

The new tests are well-written and cover the key bug scenarios. Consider adding a test case for the boundary where the `f64` bug would have manifested differently (e.g., verifying that specific percentages like `2.22`, `33.33` now produce exact results).

---

#### [nit] Documentation completeness
**Location:** Lines 97-105

The inline comment explaining the fix is excellent—clearly documents the rationale. Consider adding a brief note about why `round_dp(PRECISION)` is safe (the percentage is validated to at most PRECISION decimal places at construction).

---

### Overall Assessment
✅ **Correctness:** The Decimal-based implementation is mathematically correct for financial calculations.
✅ **Tests:** Comprehensive test coverage including edge cases.
✅ **Documentation:** Clear explanation of the bug and fix.

**Recommendation:** Address the error handling nit, otherwise LGTM.

XyneSpaces:
[nit] The `ok_or_else` for `Decimal::from_f64` returns `InvalidPercentageValue` which may be misleading since the percentage was already validated during construction. Consider a more descriptive error like `PercentageCalculationError` or add a comment explaining this path is technically unreachable but handled for completeness.

```rust
// Current:
.ok_or_else(|| report!(PercentageError::InvalidPercentageValue))?;

// Suggested - add comment:
// `from_f64` returns None only for non-finite values (NaN/Inf), which
// validation already prevents. This error path is defensive.
.ok_or_else(|| report!(PercentageError::InvalidPercentageValue))?;
```

XyneSpaces:
[should-fix] Error variant mismatch for `from_f64` failure

The error handling for `Decimal::from_f64` returning `None` uses `InvalidPercentageValue`, but the comment says this path is unreachable due to prior validation. If that's true, consider using `expect()` with a clear message, or if this is truly an error path, it should have a more specific error variant.

**Current:**
```rust
.ok_or_else(|| report!(PercentageError::InvalidPercentageValue))?;
```

The `InvalidPercentageValue` variant typically indicates the percentage string is malformed or out of range. But `from_f64` only fails for non-finite values (NaN, ±inf), which suggests an internal invariant violation rather than user input error.

**Consider:** Either:
1. Use `expect("percentage validated to be finite, but from_f64 returned None")` if truly unreachable
2. Or create a new error variant like `InternalInvariantViolation` for clarity

ManojLingala:
Thanks for the thorough pass — pushed 14c9b22. Grouped responses:

**`to_i64()` overflow (should-fix):** this branch is provably unreachable, so I documented it rather than add a test that can't trigger it:
- `Percentage` validates the range to `0..=100` (`is_valid_range`), so `result = ceil(amount * pct / 100) <= ceil(amount) = amount`, and `amount` is bounded by the guard to `i64::MAX / 10000`. The result therefore always fits in `i64` and `to_i64()` never returns `None` here.
- The suggested test wouldn't run as written — `Percentage::<2>::from_string("9999.99").unwrap()` panics, because `9999.99 > 100` is rejected at construction (a surcharge can't exceed 100%). The input-side guard is already covered by the new `percentage_apply_and_ceil_edge_cases` test (`i64::MAX` -> `Err`).
- Added a comment noting the `to_i64()` error is a defensive fallback the bounds make unreachable.

**`from_f64` error variant (should-fix):** fair — `InvalidPercentageValue` alone is thin for an internal conversion failure. I kept the graceful path (I'd avoid `expect()`/panic in a payment flow even on an "unreachable" branch) and attached a clear message + marked it defensive:
```rust
.ok_or_else(|| {
    report!(PercentageError::InvalidPercentageValue)
        .attach_printable("percentage could not be represented as a decimal")
})?;
```
A dedicated variant (e.g. `InternalInvariantViolation`) would be cleaner but touches the shared `PercentageError` enum — happy to add one if you'd prefer it over the printable.

**Storing `percentage` as `Decimal` (nit, pre-existing):** agreed, it'd remove the repeated `f32 -> Decimal` conversion. It changes the `Percentage` struct shape and predates this PR, so I'd keep it out of scope here — good follow-up.

**Exact-result tests for 2.22 / 33.33 (nit):** already covered — `percentage_apply_and_ceil_is_decimal_exact` asserts `2.22%` of `100_000_000 -> 2_220_000` and `33.33% -> 33_330_000` (the cases the old f64 path got wrong).

**`round_dp(PRECISION)` safety doc (nit):** the inline comment already notes it — "the percentage is validated to at most PRECISION decimal places, so `round_dp` recovers its exact value."

ManojLingala:
Folded the reasoning into the code comment itself (da33198) so the bound is self-documenting at the call site:

```rust
// `to_i64` returns None only above i64::MAX. Here the percentage is capped at 100%
// (range-validated) and `amount` is guarded to i64::MAX/10000, so the result =
// ceil(amount * pct/100) <= amount <= i64::MAX/10000, at most ~1/10000 of i64::MAX.
// The error is a defensive fallback the bounds make unreachable.
```

XyneSpaces:
[should-fix] Comment clarification on Decimal conversion

The comment at line 118 states "`i64 -> Decimal` is infallible" which is accurate, but could be more precise. Consider updating to:

```rust
// `Decimal::from(i64)` is infallible for all valid i64 values
```

This makes it clear the conversion is via `From<i64>` rather than the less common `From<Decimal>` path.

---

**Review Summary:**
- ✅ Correctly fixes floating-point precision bug by switching from `f64` to `Decimal` arithmetic
- ✅ Removes unnecessary `#[allow(clippy::as_conversions)]` attribute
- ✅ Comprehensive test coverage including edge cases
- ✅ Proper defensive error handling for theoretically unreachable paths

This is a well-constructed fix for a financial calculation precision issue.

ManojLingala:
Done in the latest push — updated to `// `Decimal::from(i64)` is infallible for any i64, so there is no error path here.`, naming the `From<i64>` impl so it's clear why that conversion (unlike `from_f64`/`from_i64`) needs no error handling. Thanks!

XyneSpaces:
## PR Review Summary

### Findings

**[nit]** Redundant `.round_dp()` call — Since `Percentage<PRECISION>` is already validated at construction to have at most PRECISION decimal places, the additional `round_dp(u32::from(PRECISION))` in the Decimal conversion is unnecessary defensive code. The percentage value has already been range-validated.

**[should-fix]** Comment inaccuracy — The comment states "`from_f64` only returns `None` for a non-finite percentage (NaN/Inf), which construction already rules out", but the construction validation uses `from_str` on a string representation, not `from_f64`. The finite-check happens implicitly in string parsing, so this claim about `from_f64` ruling out non-finite values is not accurate to the code path.

**[should-fix]** Missing error propagation test — The new error path for `Decimal::from_f64` returning `None` and the `to_i64` returning `None` are both marked as "defensive fallbacks" that are "unreachable" due to bounds checks, but there are no tests verifying these error paths actually work correctly.

### Positive Notes
- Excellent test coverage for edge cases (0%, 100%, overflow boundaries)
- Clear documentation explaining the motivation for Decimal over f64
- Proper use of `MinorUnit` newtype for amount handling

### Suggested Changes
1. Remove or clarify the unnecessary `.round_dp()` call
2. Fix the comment about `from_f64` and construction validation
3. Add tests for the defensive error paths to ensure they behave correctly

crates/common_utils/src/types.rs
```diff
@@ -105,8 +104,37 @@ impl<const PRECISION: u8> Percentage<PRECISION> {
...
+            // is not exactly representable as an f64 (e.g. `2.22 / 100`), so the previous
+            // `(amount as f64 * ..).ceil() as i64` could be off by a minor unit — 2.22% of
+            // 100_000_000 produced 2_220_001 instead of 2_220_000. The percentage is validated to at
+            // most PRECISION decimal places, so `round_dp` recovers its exact value; the surcharge
+            // ceiling itself is the `.ceil()` on the result below.
+            // Decimal is a bit slower than f64, but a surcharge runs once per payment, not in a
+            // tight loop, so correctness wins over the cost.
+            // `from_f64` only returns `None` for a non-finite percentage (NaN/Inf), which
+            // construction already rules out, so this is a defensive path.
+            let percentage_decimal = Decimal::from_f64(f64::from(self.percentage))
```
XyneSpaces, line 119: [should-fix] The `from_f64` + `round_dp` conversion is defensive but creates a theoretical precision edge case. Consider documenting why `round_dp(PRECISION)` is guaranteed to recover the original value, or alternatively, consider storing Percentage as Decimal internally to avoid the f64 round-trip entirely.

crates/common_utils/src/types.rs
```diff
@@ -105,8 +104,37 @@ impl<const PRECISION: u8> Percentage<PRECISION> {
...
+                .ok_or_else(|| {
+                    report!(PercentageError::InvalidPercentageValue)
+                        .attach_printable("percentage could not be represented as a decimal")
+                })?;
+            // `Decimal::from(i64)` is infallible for any i64, so there is no error path here.
+            let amount_decimal = Decimal::from(amount);
+            // `to_i64` returns None only above i64::MAX. Here the percentage is capped at 100%
+            // (range-validated) and `amount` is guarded to i64::MAX/10000, so the result =
+            // ceil(amount * pct/100) <= amount <= i64::MAX/10000, at most ~1/10000 of i64::MAX.
+            // The error is a defensive fallback the bounds make unreachable.
```
XyneSpaces, line 130: [nit] Consider extracting `Decimal::from(100)` as a const to avoid allocating a new Decimal on each call. Also, document the derivation of the `max_amount = i64::MAX / 10000` constant above - the 10000 factor appears magical without explanation.

XyneSpaces (commented):
The new tests cover the reported bug well, but miss edge cases for different PRECISION values (e.g., Percentage<4>). Consider adding a test with PRECISION != 2 to ensure the generic implementation works correctly across all precision configurations.

ManojLingala:
Thanks — working through these. One I'd push back on, one I've tightened, and one that's unreachable by construction.

**1. `round_dp()` is not redundant — it's the load-bearing line of the fix.** `Percentage` stores the value as an **`f32`** (`percentage: f32`), and construction validates the *string* has ≤ PRECISION decimals — it does **not** make the stored binary float exact. An `f32` can't represent `2.22`; it holds `~2.22000003`. `Decimal::from_f64` faithfully converts that, so **without** `round_dp` the result is `100_000_000 * 2.22000003 / 100 = 2_220_000.00…3`, and `.ceil()` → `2_220_001` — the exact bug this PR fixes.

This is falsifiable: remove `round_dp` and `percentage_apply_and_ceil_is_decimal_exact` fails (it asserts `2.22%` of `100_000_000 == 2_220_000`, but you'd get `2_220_001`). I reworded the comment so it's clear `round_dp` is required, not defensive — the old "validated to ≤ PRECISION places" wording is what made it look redundant.

**2. Comment accuracy — fair, tightened.** You're right the validation is on the string, not via `from_f64`. The accurate statement (now in the code): `self.percentage` is range-validated to `0..=100` at construction (NaN/Inf fail `is_valid_range`), so the stored `f32` is always finite, and `from_f64` (which returns `None` only for non-finite input) therefore can't fail here.

**3. "Test the defensive error paths" — they're unreachable, so no input can exercise them** (and this is in direct tension with #1, which asks to *remove* defensive code). A `from_f64`/`to_i64` `None` requires a non-finite percentage or a result `> i64::MAX`; the type invariants (`0..=100` percentage, `amount ≤ i64::MAX/10000` guard) make both impossible to construct through the public API. Testing them would mean refactoring the internals to inject values the type system forbids — i.e. testing a path production can never reach. The reachable boundary *is* covered: `percentage_apply_and_ceil_edge_cases` runs `100%` of `i64::MAX/10000` (the extreme) and confirms success without hitting the fallback. I kept graceful errors rather than `expect()` so a future invariant change degrades to an error instead of panicking in a payment path.

ManojLingala:
Thanks — good batch.

**Different PRECISION (`Percentage<4>`):** added `percentage_apply_and_ceil_other_precision`. The surcharge type uses `SURCHARGE_PERCENTAGE_PRECISION_LENGTH = 2`, but the impl is generic, so the new test runs `PRECISION = 4` — exact cases (`2.2225%` of `100_000_000 -> 2_222_500`, `0.0125%` of `80_000_000 -> 10_000`), a ceiling case (`0.3333%` of `1000 -> 4`), and rejection of a 5-decimal value. This also doubles as the executable proof that `round_dp` recovers higher-precision values.

**Why `round_dp(PRECISION)` is guaranteed to recover the value:** `f32` has a 24-bit mantissa (~7 significant decimal digits). A percentage in `0..=100` with ≤ PRECISION fractional digits has at most `3 + PRECISION` significant digits, so for the precisions actually in use (surcharge = 2; the test also covers 4) the `f32` representation error is far below half a ULP at PRECISION decimals — `round_dp(PRECISION)` snaps back to the exact validated value. Storing `Percentage` as `Decimal` would remove the round-trip entirely and is the cleaner long-term fix, but it changes the (pre-existing) struct shape, so I'd keep it out of scope here.

**`Decimal::from(100)` as a const:** there's actually nothing to optimize — `Decimal` is a 16-byte stack value (no heap allocation), so `Decimal::from(100)` just constructs an integer-valued `Decimal` inline; it doesn't allocate per call. (`Decimal::ONE_HUNDRED` exists in newer rust_decimal, but the repo currently only uses `Decimal::ONE`/`ZERO`, so I left `from(100)` to avoid depending on a const that may not be in this version.)

**`max_amount = i64::MAX / 10000`:** documented it — it caps the input so the *original* f64 math stayed precise (`~9.2e14`, well within f64's exact-integer range `2^53`). It's a pre-existing bound I kept as-is; with the Decimal change the result also can't overflow `i64` (see the `to_i64` comment).

crates/common_utils/src/types.rs
ManojLingala: Addressed in 22d8397 and fb8a97d.

- The comment now states the `round_dp` recovery is **required, not defensive** — `self.percentage` is an `f32`, which can't store e.g. `2.22` exactly (`~2.22000003`), so `round_dp(PRECISION)` removes that representation error.
- **Why recovery is guaranteed:** `f32` carries ~7 significant digits; a `0..=100` percentage with ≤ PRECISION decimals has ≤ `3 + PRECISION` significant digits, so the representation error stays far below half a ULP at PRECISION — `round_dp` snaps back exactly. The new `percentage_apply_and_ceil_other_precision` test (fb8a97d) confirms this at `PRECISION = 4`.
- Storing `Percentage` as `Decimal` would remove the round-trip but changes the (pre-existing) struct shape, so I'd keep that out of scope for this fix.

crates/common_utils/src/types.rs
ManojLingala: Addressed in fb8a97d.

- **`Decimal::from(100)`**: kept as-is — `Decimal` is a 16-byte stack value, so `from(100)` constructs an integer-valued `Decimal` inline with no heap allocation; there's nothing to optimize per call. (`Decimal::ONE_HUNDRED` exists in newer rust_decimal, but the repo currently only uses `Decimal::ONE`/`ZERO`, so I avoided depending on a const that may not be in this version.)
- **`max_amount = i64::MAX / 10000`**: documented — it caps the input so the *original* f64 math stayed precise (`~9.2e14`, well within f64's exact-integer range `2^53`). It's a pre-existing bound, kept as-is; with the Decimal change the result also can't overflow `i64` (see the `to_i64` comment).

ManojLingala:
@XyneSpaces any human reviews possible here ?

ManojLingala:
@XyneSpaces its been a while any human reviews possible ?

PR #12875, opened 2026-06-22
poc(connector/stripe): demonstrate AttemptStatusDomain framework for Unknown status handling

XyneSpaces:
---
### What this demonstrates
A framework (`#[derive(DomainStatus)]` on `router_derive`) that generates a **Domain mirror** of `AttemptStatus` with an extra `Unknown` variant plus a `Default` impl. Application/orchestration logic now carries the **Domain** type; the **storage** type (`common_enums::AttemptStatus`) is used only at the DB boundary. Verified end-to-end on `AttemptStatus` with **Stripe** as the pilot connector.

### Design
- **Domain vs storage split.** `RouterData.status` and the `router_data_v2` common-flow-types now hold `AttemptStatusDomain`. Storage types are untouched, so the DB schema and wire format are unchanged.
- **Conversion happens at the right layer.** Domain→storage resolution is wired into `get_attempt_status_for_db_update` (v1) — the single point where a status is committed to the DB. No conversion leaks into connector code.
- **Framework-generated `Default`.** `router_derive` emits the `Default` impl for the domain mirror, so `Unknown` collapses to a safe non-terminal `Pending` when a lossy read is unavoidable (`.to_storage().unwrap_or_default()`).
- **Unknown is preserved through Stripe.** Stripe payment responses build `AttemptStatusDomain` directly (`From<StripePaymentStatus> for AttemptStatusDomain`, `Unknown → AttemptStatusDomain::Unknown`) instead of pre-collapsing via the storage `From`. `Unknown` now propagates into the router where `get_attempt_status_for_db_update` resolves it against prior state.

### Scope
- Pilot on one status (`AttemptStatus`) + one connector (Stripe). The other ~120 connectors keep working via a backward-compat `From<StripePaymentStatus> for AttemptStatus` and lossy `.to_storage().unwrap_or_default()` reads for prior-state classification — prior state is never `Unknown`, so this is safe.
- v1 flow only. v2 `get_attempt_status_for_db_update` and other status enums are follow-up work.

### Build verification
- `cargo check -p router` (default features, v1) — **GREEN, 0 errors**. This is the canonical POC build target and compiles `hyperswitch_connectors` (including all Stripe changes) in the correct feature configuration.

### Out of scope
- No DB migration, no API-contract change, no routing/retry behavior change.
- Full rollout to all connectors and all status enums is deliberately deferred.
---


XyneSpaces:
## Review: AttemptStatusDomain Framework POC

### [should-fix] Unknown Variant Not Persisted Warning

The documentation correctly states that `Unknown` is "never persisted to the database". However, in the backward compatibility shim for `StripePaymentStatus`:

```rust
impl From<StripePaymentStatus> for AttemptStatus {
    fn from(item: StripePaymentStatus) -> Self {
        AttemptStatusDomain::from(item)
            .to_storage()
            .unwrap_or(AttemptStatus::Pending)
    }
}
```

This silently defaults to `Pending` when the domain status is `Unknown`. This could mask issues where a new Stripe status appears but isn't handled. Consider:
1. Logging a warning when this fallback occurs
2. Or adding metrics to track occurrences of Unknown→Pending fallbacks

### [should-fix] Derive Macro Error Message

In `domain_status.rs`, the derive macro checks:

```rust
if variant.ident == "Unknown" {
    return Err(syn::Error::new_spanned(
        variant,
        "storage enum must not declare `Unknown`; the derive appends it",
    ));
}
```

This prevents storage enums from having an Unknown variant, which is good. However, the error message could be more helpful by suggesting the fix (removing the Unknown variant from the storage enum).

### [nit] Test Coverage Gap

While the tests cover graceful deserialization and storage conversion guards, consider adding a test for the actual integration scenario:

```rust
#[test]
fn stripe_response_with_unknown_status_resolves_correctly() {
    // Simulate: Stripe returns "requires_action_v2" (new status)
    // Previous state was "requires_customer_action"
    // Result should preserve the previous state
}
```

### [nit] Documentation Consistency

The module-level documentation in `domain_status.rs` mentions `resolve_or_keep` but the example shows manual conversion. Consider adding a complete example showing the intended usage pattern:

```rust
// In connector transformer:
let domain_status: AttemptStatusDomain = stripe_status.into();
let resolved = domain_status.resolve_or_keep(previous_attempt_status);
let new_status = resolved.to_storage()?;
```

### Overall Assessment

Well-designed POC that provides a clean migration path. The derive macro reduces boilerplate significantly. The contract that Unknown never reaches storage is properly enforced through the type system.

XyneSpaces:
## Code Review: Domain Status Resilience Feature

### Overview
This PR introduces a robust pattern for handling unknown connector status values through a `DomainStatus` derive macro that creates domain-specific mirror enums with an `Unknown` catch-all variant.

### Findings

#### ✅ Positive Observations
1. **Excellent documentation** - The module-level docs in `domain_status.rs` clearly explain the contract and rationale
2. **Comprehensive test coverage** - Unit tests cover deserialization, round-trips, and error cases
3. **Safe defaults** - The `Unknown → storage` conversion fails loudly rather than silently defaulting

#### ⚠️ Suggestions

**1. Error Identifier Consistency (api_error_response.rs)**
The error code `IR_52` for `PaymentSessionExpired` uses a gap in numbering (skips from IR_48 to IR_52). While not blocking, consider whether intermediate codes are reserved for future use or if sequential numbering would be clearer.

**2. Stripe Error Mapping**
In `stripe/errors.rs`, the `PaymentSessionExpired` maps to `PaymentNotFound`. Consider if a more specific Stripe error code would better inform merchants about the actual issue (session expiry vs. missing payment).

**3. DomainStatus Macro Validation**
The derive macro prevents `Unknown` in storage enums at compile time. Consider adding a runtime debug assertion in the `TryFrom` implementation to double-check during testing.

### Security Considerations
- The `Unknown` variant is correctly marked as internal-only and never persisted
- The error type properly distinguishes between serialization failures and unknown status resolution failures

### Overall Assessment
Well-architected change that improves resilience against connector API changes. The macro-generated approach ensures the domain enum stays synchronized with the storage enum.

---
*Reviewed by XyneSpaces automated PR reviewer*

crates/router_derive/src/macros/domain_status.rs
```diff
@@ -0,0 +1,153 @@
...
+            ));
+        }
+        // Carry over doc comments and serde attrs (aliases/renames); drop the
+        // rest (`#[default]`, strum/diesel/schema helpers) which do not apply to
+        // the domain mirror.
+        let kept_attrs: Vec<_> = variant
+            .attrs
+            .iter()
+            .filter(|attr| attr.path().is_ident("doc") || attr.path().is_ident("serde"))
+            .collect();
```
XyneSpaces, line 72: [nit] The derive macro rejects storage enums that already have an Unknown variant, which is good defensive coding. However, consider adding a compile-time warning or note in the documentation that explains WHY Unknown is reserved - to prevent confusion for developers who might try to add it manually to storage enums.

XyneSpaces (commented):
Review comment on specific line

XyneSpaces (commented):
## Review Summary

DomainStatus framework implementation for resilient connector status handling.

**Strengths:**
- ✅ Comprehensive test coverage (8 tests) for AttemptStatusDomain, RefundStatusDomain, DisputeStatusDomain
- ✅ Procedural macro correctly prevents Unknown variant in storage enums
- ✅ Backward compatibility shim for StripePaymentStatus → AttemptStatus
- ✅ resolve_or_keep() pattern properly merges unknown statuses with previous state

**Inline Comment:**
- [nit] Documentation could explain why Unknown is reserved for domain enums

Clean implementation with good safety guarantees.

PR #12884, opened 2026-06-22
chore(metrics): adjust f64 histogram buckets size

su-shivanshmathur:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

Adjusting the initialization and bucket count for f64 histograms in the metrics helper.

### Additional Changes

## Motivation and Context

The current f64 histogram buckets start at 0.01 and only provide 15 buckets, resulting in coarse granularity for small values. Lowering the initial bucket boundary to 0.000_001 and expanding the array to 30 buckets improves metric resolution.

## How did you test it?

Screenshot 2026-06-22 at 11 49 55 PM

## Checklist


XyneSpaces:
## Review Findings

### [should-fix] Breaking Change - Histogram Bucket Count Increase

This PR increases f64 histogram buckets from 15 to 60 (4x increase). This is a **breaking change** for monitoring infrastructure:

**Impacts:**
1. **Dashboard Queries:** Existing Grafana/Prometheus queries referencing specific bucket boundaries (e.g., `le="0.01"`) will break or return different results
2. **Alert Thresholds:** P99/P95 calculations will shift due to different bucket distributions  
3. **Storage:** 4x increase in histogram time-series data points

**Recommendations:**
1. Document this as a breaking change in release notes
2. Consider making bucket configuration runtime-configurable rather than hardcoded
3. Validate that existing dashboards are updated post-deployment

### [nit] Starting Value Granularity

The new starting value `0.000001` (1 microsecond) may be unnecessarily granular for most payment operations:

- HTTP/database operations rarely complete in < 1ms
- This effectively "wastes" the first ~10-20 buckets on an unrealistic range
- Consider starting at `0.0001` (100 microseconds) or `0.001` (1 millisecond)

### [nit] Justification Comment

Please add a code comment explaining the bucket strategy:

```rust
// 60 buckets with 2x growth factor covers ~1µs to ~60s range
// Suitable for cache hits through slow PSP responses
```

---

*Reviewed via automated scan. Please document the breaking change prominently.*

SanchithHegde (dismissed):
Looks good to me!
Why were Cypress files affected though?

su-shivanshmathur:
> Looks good to me! Why were Cypress files affected though?

Not sure, got when pulled main branch

crates/router_env/src/metrics.rs
```diff
@@ -139,8 +139,8 @@ mod helpers {
     /// Returns the buckets to be used for a f64 histogram
     #[inline(always)]
     pub fn f64_histogram_buckets() -> Vec<f64> {
-        let mut init = 0.01;
-        let mut buckets: [f64; 15] = [0.0; 15];
+        let mut init = 0.000_001;
```
XyneSpaces, line 142: [nit] The bucket count increase from 15 to 30 will increase metric cardinality significantly. Consider documenting the memory impact in comments and ensuring the metrics backend can handle the additional series. Also, verify that the 0.000_001 starting bucket is necessary - this is 10 microseconds which may be finer granularity than needed for most operations.

XyneSpaces (commented):
Reviewing the histogram bucket changes.

crates/router_env/src/metrics.rs
SanchithHegde: Can you also reorder these lines, so that buckets start at 1 * 10^-6 rather than 2 * 10^-6?



```suggestion
*bucket = init;
init *= 2.0;
```

su-shivanshmathur: Done

SanchithHegde (approved):

likhinbopanna (approved):

PR #12885, opened 2026-06-23
refactor(configs): allow config retrieve API to be called with internal API key or admin API key

SanchithHegde:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [x] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

This PR updates the config retrieve endpoint (both v1 and v2) to also allow internal API key authentication alongside the existing admin API key authentication. This is being done so as to allow other internal systems to make use of Hyperswitch's config setup to retrieve configs. In this case, the internal system that requires this is our card vault (see juspay/hyperswitch-card-vault#171).

### Additional Changes

This PR updates the config retrieve endpoint to allow internal API key authentication alongside the existing admin API key authentication.

## Motivation and Context

This allows other internal systems such as our card vault to make use of Hyperswitch's config setup to retrieve runtime configs. See juspay/hyperswitch-card-vault#171.

## How did you test it?

- Enable internal API key authentication by setting `internal_merchant_id_profile_id_auth.enabled` to `true` in the config file.
- Create a config (`POST /configs/` endpoint) using admin API key.
  - If the internal API key is used here instead of the admin API key, the request would fail with an unauthenticated error.
- Retrieve the created config using either the admin API key, or the internal API key (from `internal_merchant_id_profile_id_auth.internal_api_key` field in the config file), both should succeed. Providing neither of them causes the API to fail with an unauthenticated error.
  - The internal API key can be used with the config retrieve endpoint like so:
    ```shell
    curl --location 'http://localhost:8080/configs/locker.use_read_replica' \
      --header 'X-Internal-Api-Key: ********'
    ```

## Checklist

### DIFF
--- a/crates/router/src/routes/configs.rs
+++ b/crates/router/src/routes/configs.rs
@@ -48,7 +48,7 @@ pub async fn config_key_retrieve(
         &req,
         &key,
         |state, _, key, _| configs::read_config(state, key),
-        &ADMIN_API_AUTH,
+        &auth::InternalApiKeyAuth(ADMIN_API_AUTH),
         api_locking::LockAction::NotApplicable,
     )
     .await
--- a/crates/router/src/services/authentication.rs
+++ b/crates/router/src/services/authentication.rs
@@ -201,6 +201,7 @@ pub enum AuthenticationType {
         merchant_id: id_type::MerchantId,
         profile_id: id_type::ProfileId,
     },
+    InternalApiKey,
     NoAuth,
 }
 
@@ -235,6 +236,7 @@ impl AuthenticationType {
             | Self::EmbeddedJwt { merchant_id, .. }
             | Self::SdkAuthorization { merchant_id, .. } => Some(merchant_id),
             Self::AdminApiKey
+            | Self::InternalApiKey
             | Self::OrganizationJwt { .. }
             | Self::BasicAuth { .. }
             | Self::UserJwt { .. }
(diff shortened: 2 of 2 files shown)


XyneSpaces:
## Review: Internal API Key Authentication

**File: `crates/router/src/services/authentication.rs`** (lines ~2790-2810)

### [should-fix] Timing Attack Vulnerability

The internal API key comparison uses direct string equality:

```rust
if key == *state.conf().internal_merchant_id_profile_id_auth.internal_api_key.peek()
```

**Issue**: String comparison with `==` is vulnerable to timing attacks. An attacker could measure response times to brute-force the internal API key byte-by-byte.

**Suggested fix**: Use a constant-time comparison:

```rust
use ring::constant_time::verify_slices_are_equal;
// or subtle crate

let expected = state.conf().internal_merchant_id_profile_id_auth.internal_api_key.peek();
if subtle::ConstantTimeEq::eq(key.as_bytes(), expected.as_bytes()).into() {
    Ok(((), AuthenticationType::InternalApiKey))
} else {
    Err(errors::ApiErrorResponse::Unauthorized)
}
```

---

### [nit] Fallback Auth Type Should Be Preserved

In the fallback branch (`None => self.0.authenticate_and_fetch(...)`), the authentication type returned is from the inner authenticator (`ADMIN_API_AUTH`). This is correct behavior, but consider documenting this behavior explicitly so future maintainers understand that:
- Internal key → `AuthenticationType::InternalApiKey`
- Admin key → `AuthenticationType::AdminApiKey`

su-shivanshmathur (approved):

XyneSpaces:
[should-fix] The `InternalApiKeyAuth` wrapper introduces a new authentication path that bypasses the standard admin API key validation when the internal API key feature is enabled. This creates a potential security concern if the internal API key is compromised, as it grants the same access level as admin API keys.

Consider adding rate limiting specifically for internal API key authentication attempts to mitigate brute force attacks against the internal API key. Also, ensure the internal API key is rotated regularly and has appropriate logging/alerting for its usage.

Additionally, the fallback to `ADMIN_API_AUTH` when the internal key feature is disabled should be explicitly documented in the API documentation to avoid confusion for API consumers.

ThisIsMani (approved):

XyneSpaces:
[should-fix] The `InternalApiKeyAuth` wrapper implementation introduces conditional authentication bypass. When `internal_merchant_id_profile_id_auth.enabled` is true and a valid `X_INTERNAL_API_KEY` header is present, the request bypasses the wrapped authenticator entirely.

Ensure that the internal API key is rotated regularly and stored securely. Consider adding audit logging for all requests authenticated via `InternalApiKey` to trace administrative actions. Additionally, verify that endpoints protected by `ADMIN_API_AUTH` don't have unintended exposure when this feature flag is enabled.

ShankarSinghC (approved):

jarnura (approved):

XyneSpaces:
[should-fix] **Security:** The new `InternalApiKeyAuth` wrapper introduces elevated privileges for config retrieval. Please verify the following before merge:

1. **Fail-closed behavior:** If `X-Internal-Api-Key` header is present but invalid, the request should return 401 immediately—never fall through to `AdminApiKey` validation

2. **Constant-time comparison:** Use `subtle::constant_time_eq()` for key comparison to prevent timing attacks:
   ```rust
   use subtle::ConstantTimeEq;
   if !provided_key.ct_eq(stored_key).into() {
       return Err(errors::ApiErrorResponse::Unauthorized);
   }
   ```

3. **Audit logging:** Ensure all `InternalApiKey` authentications are logged distinctly from regular admin auth for security monitoring

4. **Key storage:** Verify the internal API key is loaded from a secure secret store (KMS/Vault), not from plaintext config files

5. **Scope enforcement:** Confirm `InternalApiKey` is restricted to config endpoints only—it should not grant access to payment data or PII

PR #12941, opened 2026-06-24
fix(router): Prevent redundant api_event entries in case of webhooks

AnuthaDev:
## Type of Change

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

### Problem
Incoming webhook requests to Hyperswitch produce **duplicate rows** in ClickHouse `api_events` table — one from `server_wrap_util` (the generic request wrapper) and one from `incoming_webhooks_wrapper` (the webhook handler).
The outer `server_wrap_util` logs an event with `flow_type = miscellaneous, request = null,` and `response = null` because `ApplicationResponse::StatusOk` carries no body to serialize. The inner `incoming_webhooks_wrapper` then logs a second event with `flow_type = webhooks`, the real webhook body as request, and the `WebhookResponseTracker` business effect as response.

This results in two rows per webhook request, polluting analytics and inflating event counts.

### Solution
Introduced a new `IncomingWebhookEvent` variant on ApplicationResponse that wraps the ack response alongside `IncomingWebhookEventMetadata` (containing `serialized_request`, `webhook_tracker_data`, and `event_type`). The webhook handlers now return this variant instead of logging independently. `server_wrap_util` detects this variant, extracts the metadata to populate a single correct `ApiEvent`, and flattens it to the inner response for HTTP mapping.

**Changes**
- **crates/hyperswitch_domain_models/src/api.rs**: Added IncomingWebhookEvent variant and IncomingWebhookEventMetadata struct on ApplicationResponse. Updated ApiEventMetric and get_json_body impls.
- **crates/router/src/services/api.rs**: server_wrap_util overrides serialized_request, serialized_response, and event_type from metadata. server_wrap flattens IncomingWebhookEvent to inner response before HTTP mapping.
- **crates/router/src/core/webhooks/incoming.rs**: v1 incoming_webhooks_wrapper + network_token_incoming_webhooks_wrapper — removed inner log_event, return IncomingWebhookEvent.
- **crates/router/src/core/webhooks/incoming_v2.rs**: v2 incoming_webhooks_wrapper — same refactor.
- **crates/router/src/routes/metrics/request.rs**: Added IncomingWebhookEvent arm to track_response_status_code.
- **crates/router/src/compatibility/wrap.rs**: Flattens IncomingWebhookEvent in .map() before HTTP match.
- **crates/router/src/workflows/outgoing_webhook_retry.rs**: Added IncomingWebhookEvent arm to 3 exhaustive matches.

Verification

### Additional Changes

## Motivation and Context
Duplicate api_events rows for incoming webhooks corrupt analytics dashboards, inflate event counts, and make it ambiguous which row represents the authoritative record. The miscellaneous row from the outer wrapper carries no useful data (null request, null response) and exists solely as a side effect of the generic logging path. This refactor ensures a single authoritative event per webhook request, with all metadata populated by the handler that actually processed the webhook.

## How did you test it?

Tested end-to-end with Kafka + ClickHouse in Docker. A single incoming webhook (Adyen CAPTURE event) produced exactly 1 row in api_events:
- flow_type = webhooks (not miscellaneous)
- request = full webhook body (396 bytes)
- response = {"Payment":{"payment_id":"pay_...","status":"succeeded"}} (WebhookResponseTracker)
- status_code = 200
Before the fix, the same request produced 2 rows (one miscellaneous with nulls, one webhooks with data).

Before:
image

webhook flow type:
image

miscellaneous flow type:
image

After:
image

## Checklist


crates/hyperswitch_domain_models/src/api.rs
```diff
@@ -18,6 +18,21 @@ pub enum ApplicationResponse<R> {
     FileData((Vec<u8>, mime::Mime)),
     JsonWithHeaders((R, Vec<(String, hyperswitch_masking::Maskable<String>)>)),
     GenericLinkForm(Box<GenericLinks>),
+    /// `response` controls HTTP behavior; `metadata` is consumed by `server_wrap_util`
+    /// for the single ApiEvent log, replacing the former inner `log_event` that caused
+    /// duplicate rows.
```
XyneSpaces: [blocking] The new `ApplicationResponse::WebhookEvent` variant introduces a significant architectural change. Ensure backward compatibility is maintained for any external integrations that might be pattern-matching on `ApplicationResponse`. Consider if this variant should be `#[doc(hidden)]` or if migration documentation is needed.

XyneSpaces (commented):
Reviewing the new `ApplicationResponse::WebhookEvent` variant addition.

crates/router/src/services/api.rs
```diff
@@ -316,6 +316,24 @@ where
                     .conf
```
XyneSpaces, line 316: [should-fix] The unpacking logic in `server_wrap_util` for the `WebhookEvent` variant should include comprehensive error handling. If metadata extraction fails, the original webhook response should still be returned to the caller to prevent webhook delivery failures due to logging issues.

XyneSpaces (commented):
Review comments added for `server_wrap_util` changes.

crates/router/src/core/webhooks/incoming.rs
```diff
@@ -97,57 +89,32 @@ pub async fn incoming_webhooks_wrapper<W: types::OutgoingWebhookType>(
...
-        connector: connector_name_or_mca_id.to_string(),
-        payment_id: webhooks_response_tracker.get_payment_id(),
-        refund_id: webhooks_response_tracker.get_refund_id(),
+    let metadata = IncomingWebhookEventMetadata {
+        event_type: ApiEventsType::Webhooks {
+            connector: connector_name_or_mca_id.to_string(),
+            payment_id: webhooks_response_tracker.get_payment_id(),
+            refund_id: webhooks_response_tracker.get_refund_id(),
+        },
+        serialized_request: serialized_req.peek().clone(),
```
hrithikesh026: this should remain a secret.

crates/router/src/core/webhooks/incoming.rs
```diff
@@ -97,57 +89,32 @@ pub async fn incoming_webhooks_wrapper<W: types::OutgoingWebhookType>(
...
+        webhook_tracker_data: serde_json::to_value(&webhooks_response_tracker)
+            .change_context(errors::ApiErrorResponse::InternalServerError)
+            .attach_printable("Could not convert webhook effect to string")
```
hrithikesh026: No need to call these two function if you are calling .ok()

hrithikesh026: you could log the error using .inspect_err()

crates/hyperswitch_domain_models/src/api.rs
```diff
@@ -18,6 +18,19 @@ pub enum ApplicationResponse<R> {
     FileData((Vec<u8>, mime::Mime)),
     JsonWithHeaders((R, Vec<(String, hyperswitch_masking::Maskable<String>)>)),
     GenericLinkForm(Box<GenericLinks>),
+    /// `response` contains the HTTP response; `metadata` is consumed by `server_wrap_util`
+    /// for the ApiEvent log
+    IncomingWebhookEvent {
+        response: Box<Self>,
```
hrithikesh026: Is there a valid reason for this recursive reference?
Application response of any webhook will always be a pre defined format right?
I don't see a reason why webhook response would return GenericLinkForm, FileData, Form etc.

crates/router/src/core/webhooks/incoming_v2.rs
```diff
@@ -66,7 +60,6 @@ pub async fn incoming_webhooks_wrapper<W: types::OutgoingWebhookType>(
         .with_provider_merchant_id(platform.get_provider().get_provider_merchant_id())
         .with_processor_merchant_id(platform.get_processor().get_processor_merchant_id());
 
-    let start_instant = Instant::now();
     let (application_response, webhooks_response_tracker, serialized_req) =
         Box::pin(incoming_webhooks_core::<W>(
```
hrithikesh026, line 63: `incoming_webhooks_core` should always return `WebhookResponse` object instead of `ApplicationResponse`.

crates/router/src/core/webhooks/incoming_v2.rs
```diff
@@ -82,48 +75,28 @@ pub async fn incoming_webhooks_wrapper<W: types::OutgoingWebhookType>(
...
-        api_event,
-        req,
-        req.method(),
-        infra,
-    );
-    state.event_handler().log_event(&api_event);
-    Ok(application_response)
+    let webhook_response = WebhookResponse::try_from(application_response)
+        .map_err(|_| errors::ApiErrorResponse::InternalServerError)
+        .attach_printable("Invalid response type for incoming webhook")?;
```
hrithikesh026: if you do this https://github.com/juspay/hyperswitch/pull/12941/changes#r3503501216, this try_from conversion is unnecessary.

crates/hyperswitch_domain_models/src/api.rs
```diff
@@ -18,6 +19,59 @@ pub enum ApplicationResponse<R> {
...
+            ApplicationResponse::JsonWithHeaders((r, h)) => Ok(Self::JsonWithHeaders((r, h))),
+            ApplicationResponse::JsonForRedirection(_)
+            | ApplicationResponse::Form(_)
+            | ApplicationResponse::PaymentLinkForm(_)
+            | ApplicationResponse::FileData(_)
+            | ApplicationResponse::GenericLinkForm(_)
+            | ApplicationResponse::IncomingWebhookEvent { .. } => Err(resp),
+        }
+    }
+}
```
hrithikesh026: this conversion can be avoided

hrithikesh026 (changes requested):

hrithikesh026 (approved):

AkshayaFoiger (approved):

crates/router/src/compatibility/wrap.rs
```diff
@@ -175,6 +181,9 @@ where
                 }
             }
         }
+        // This match arm should never be reached as we map the inner field of IncomingWebhookEvent
+        // to ApplicationResponse above
+        Ok(api::ApplicationResponse::IncomingWebhookEvent { .. }) => api::http_response_ok(),
```
Sakilmostak, line 186: nit: shouldn't this throw error to represent it is unreachable

Sakilmostak (approved):

rohanm13 (approved):

PR #12948, opened 2026-06-24
fix(router): apply UCS error attempt_status on redsys authenticate failure (#17016)

10X-GRACE:
## What

Hyperswitch half of the redsys `authenticate` (post-3DS) error-path parity fix under parent #17016. Pairs with hyperswitch-prism#1660.

On a declined authenticate, the UCS readback decoded the error's `attempt_status`, but the complete-authorize **authenticate** caller (`complete_authorize_flow.rs`) only applied status on the `Ok` branch:

```rust
let router_data_response = router_data_response.map(|(response, status)| {
    router_data.status = status;   // Ok only
    response
});
```

So on a failed authenticate, `router_data.status` stayed at the prior `authentication_pending` while the Direct gateway moved it to `failure` (#17021).

## Fix

Mirror the authorize gateway (`gateway/authorize_gateway.rs`): on the `Err` branch, apply `err.attempt_status` to `router_data.status` when present.

```rust
Err(err) => {
    if let Some(attempt_status) = err.attempt_status {
        router_data.status = attempt_status;
    }
    Err(err)
}
```

Once prism#1660 emits `status=FAILURE` on the authenticate error response, the readback yields `attempt_status=Some(Failure)` and this change moves `router_data.status` to `failure`, matching Direct.

No proto / rev-pin change.

## Verification

- Builds clean (`cargo build -p router`, exit 0).
- The prism half is A/B-verified via direct gRPC (see prism#1660): the authenticate error response now carries `status=FAILURE` (was `UNSPECIFIED`), which this caller change applies to `router_data.status`.
- The change is a verbatim mirror of the already-shipped authorize-gateway error-status handling.

Fixes #17021.

PR #1664, opened 2023-07-10
feat(router): add connector_response_reference_id in payments response

sai-harsha-vardhan:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description
Get the connector_response_reference_id from connector flows, store it in payment_attempt table and send it back in payments response which would be used by a merchant as a reference to a payment at connector side

### Additional Changes

- [x] This PR modifies the API contract
- [x] This PR modifies the database schema
- [ ] This PR modifies application configuration/environment variables

## How did you test it?
**Manually tested**
Payments response
image

Payment attempt record
image

## Checklist

- [x] I formatted the code `cargo +nightly fmt --all`
- [x] I addressed lints thrown by `cargo clippy`
- [x] I reviewed submitted code
- [ ] I added unit tests for my changes where possible
- [ ] I added a [CHANGELOG](/CHANGELOG.md) entry if applicable

PR #12955, opened 2026-06-24
fix(router): apply UCS error attempt_status on redsys pre_authenticate failure (#17027)

### LINKED ISSUE #12994: [shadow-validation] redsys pre_authenticate: status valueDiff on HTTP-200 error envelope
### Shadow-validation parity: redsys `pre_authenticate` status on HTTP-200 error envelope

**Flow:** `pre_authenticate` · **Connector:** `redsys` · **Result:** `router_diff`

On a redsys SIS REST **HTTP-200** error envelope (e.g. `{"errorCode":"SIS0093"}`), the
UCS shadow path keeps the prior attempt status (`payment_method_awaited`) while the Direct
gateway resolves `failure`:

```
router.valueDiff:status { hyperswitch: "failure", ucs: "payment_method_awaited" }
```

**Root cause (two-repo, UCS-only, no proto):**
- `call_unified_connector_service_pre_authenticate` applied the mapped status **only on the
  `Ok` path** (`.map`); on an error envelope `router_data.status` was never updated.
- UCS `generate_payment_pre_authenticate_response` Err branch emitted
  `PAYMENT_STATUS_UNSPECIFIED` (`unwrap_or_default()`), which HS maps to `None`.

**Fix:** hyperswitch #12955 (apply `err.attempt_status` on the Err branch) +
hyperswitch-prism #1664 (emit `FAILURE` on the error envelope). Mirrors the already-merged
authenticate sub-flow fix (#12948 / #1660).

Tracking issue for the linked-issue convention check on #12955.

10X-GRACE:
## What
The `pre_authenticate` UCS caller (`call_unified_connector_service_pre_authenticate`) applied the mapped status **only on the `Ok` path** (via `.map`). On an error envelope the router therefore kept its prior status (e.g. `payment_method_awaited`) instead of `failure`, diverging from the Direct gateway.

Fix: rewrite as a `match` and apply `err.attempt_status` on the `Err` branch too — a 1:1 mirror of the authenticate caller in `complete_authorize_flow.rs` (hyperswitch #12948).

UCS-only, no proto change.

## Co-dependency
Pairs with **hyperswitch-prism #1664**, which makes UCS emit `Failure` (not `UNSPECIFIED`) on the redsys pre_authenticate error envelope. Without the prism change `err.attempt_status` is `None` and nothing is applied; without this change the emitted `Failure` is never written to `router_data.status`.

## RCA
redsys HTTP-200 `{"errorCode":"SIS0093",...}` → shadow-validation diff: `status` mismatch, hyperswitch reports `"failure"`, UCS reports `"payment_method_awaited"`.

## Verification
- prism #1664 A/B (direct gRPC, real redsys sandbox `PreAuthenticate`): UCS `status` `UNSPECIFIED` → **`FAILURE`** on the HTTP-200 error envelope.
- This change: router builds clean; readback maps `FAILURE → failure`, matching Direct. (Textually identical to the already-reviewed #12948 authenticate-caller fix.)

Closes #12994


XyneSpaces:
## Code Review: PR #12955

### Summary
Apply UCS error `attempt_status` on Redsys pre_authenticate failure.

### Findings

**File:** `crates/router/src/core/payments/flows/authorize_flow.rs`

✅ **[should-fix]** Good fix addressing the status propagation gap in the error branch of UCS pre-authenticate flow.

### Key Changes Reviewed:
1. **Lines 1642-1660**: Refactored the error handling to propagate `attempt_status` from UCS error response
2. The fix mirrors the pattern already established in `complete_authorize_flow.rs` (authenticate caller)

### Critical Observation
The comment added at line 1647-1651 clearly explains:
> "Apply the error's attempt_status on the failure path too (the Ok-only `.map` previously left router_data.status at the prior value, e.g. payment_method_awaited, while the Direct gateway sets Failure on an error envelope)"

### Recommendations:
1. ✅ Error status is now properly propagated from UCS error response
2. ✅ Consistent with the fix pattern in `complete_authorize_flow.rs`
3. Consider adding integration test coverage for Redsys 3DS pre-authenticate failure scenarios

LGTM 👍

XyneSpaces:
## Review: fix(router): apply UCS error attempt_status on redsys pre_authenticate failure

### Summary
Fixes an inconsistency where `router_data.status` was not updated on error paths in UCS pre-authenticate flow, causing status to remain at prior value instead of reflecting the failure.

### Findings

**[blocking]** **Logic correctness** - The fix properly applies `err.attempt_status` to `router_data.status` when available. However, verify that:
1. All error types passed to this code path have the `attempt_status` field populated
2. There's no case where applying the error's attempt_status would incorrectly overwrite a more specific status

**[should-fix]** **Consider adding a debug log** when the error's attempt_status is applied for better observability:
```rust
if let Some(attempt_status) = err.attempt_status {
    logger::debug!("Applying error attempt_status {:?} to router_data.status", attempt_status);
    router_data.status = attempt_status;
}
```

### Verified
- ✅ Mirrors the pattern used in `complete_authorize_flow.rs` authenticate caller
- ✅ Maintains consistency with Direct gateway behavior
- ✅ The `match` refactor improves readability over the previous `.map` approach

XyneSpaces:
## Review Summary

PR #12955 fixes a state management bug in the UCS pre-authenticate flow where `router_data.status` wasn't being updated on error responses. This is a critical fix for proper payment attempt status tracking.

### Analysis

**The bug:** The original code used `.map()` which only transforms the `Ok` variant, leaving `router_data.status` unchanged when the UCS call returns an error. This meant the status could remain as `payment_method_awaited` (or whatever the prior state was) even when authentication actually failed.

**The fix:** Replaces `.map()` with a `match` expression that handles both success and error paths:
- Success: Sets status and transforms response (same as before)
- Error: Sets `router_data.status = err.attempt_status` when available

### Code quality observations

**✅ Good:**
- Clear comment explaining the rationale and referencing `complete_authorize_flow.rs`
- Minimal, focused change
- Pattern matches the existing `authenticate` caller implementation

### [should-fix] Verify error.attempt_status is always appropriate

The code applies `err.attempt_status` unconditionally when present:
```rust
if let Some(attempt_status) = err.attempt_status {
    router_data.status = attempt_status;
}
```

Please confirm that:
1. The `attempt_status` from UCS errors is always the correct status to apply
2. There are no cases where we should preserve the original status instead
3. This aligns with how `complete_authorize_flow.rs` handles the same scenario

### [nit] Consider exhaustive comment

The comment could mention what status value is typically expected (e.g., `AttemptStatus::Failure`) to make the expected behavior more explicit.

---

**Overall:** This is a targeted bug fix that addresses a real issue in payment state management. The pattern is consistent with similar code in the codebase.

iemyashasvi (dismissed):

hrithikesh026 (approved):

gopikrishna000 (approved):

PR #12964, opened 2026-06-24
fix(external_vault): forward x-tenant-id on hyperswitch_vault calls

### LINKED ISSUE #12965: bug(external_vault): hyperswitch_vault calls miss x-tenant-id → HE_05 under multitenancy
## Bug

When `HyperswitchVault` is used as the external vault (hyperswitch acting as its own vault), the vault endpoint is a **self-loop** back into the multi-tenant Hyperswitch service. These in-cluster calls do **not** traverse the edge/ingress that injects `x-tenant-id`, so under multitenancy the vault rejects them with **`HE_05` "Missing Tenant ID in the request"** (surfaced to the API caller as `HE_00` / 500).

This breaks both the external-vault session setup and the external-vault-proxy payment confirm when the vault `base_url` points at the service in-cluster (which is itself required to avoid the egress forward-proxy / CloudFront returning a non-JSON 403).

## Affected self-loop calls

Three distinct components call the vault, each needing `x-tenant-id`:

| Caller | Call |
|---|---|
| Modular / PM service | `POST /v1/customers`, `POST /v1/payment-method-sessions` (connector customer-create / session create) |
| UCS injector | `POST /v1/proxy` (detokenization + forward) |
| Payment server | `GET /payment-methods/token/{token}/details` (`get_permanent_pm_id_from_temporary_token`) |

## Symptoms

- `session_tokens` → "Failed to get or create vault connector customer" (vault_session.rs) → `vault_details: null`.
- `POST /payments/{id}/confirm` (external vault proxy) → `transformers.rs` "Failed to parse hyperswitch vault token details response" → `error decoding response body: missing field id` (the unparsed `HE_05` error body).

## Root cause

The connector `build_headers`, the injector, and the router's direct token-details `http_client` call all omitted `x-tenant-id`. Externally these requests get the tenant injected by the ingress; in-cluster they don't, so the multi-tenant vault rejects them.

## Fix

Forward `x-tenant-id` (from the request/session tenant) on all three self-loop calls. Addressed in #12964.

ShankarSinghC:
## Type of Change

- [x] Bugfix

## Description

`HyperswitchVault` (hyperswitch acting as its own external vault) is a **self-loop** back into the multi-tenant Hyperswitch service. Unlike calls to external connectors, these requests never traverse the edge/ingress that injects `x-tenant-id`, so under multitenancy the vault rejected them with **`HE_05` "Missing Tenant ID in the request"** (which then surfaced to the caller as `HE_00`).

There are **three** distinct self-loop calls into the vault, each made by a different component, and all three now forward the tenant explicitly:

| Caller | Call | Where |
|---|---|---|
| Modular/PM service | `POST /v1/customers`, `POST /v1/payment-method-sessions` (connector customer-create / session create) | `hyperswitch_vault` connector `build_headers` — from `RouterData.tenant_id` |
| UCS injector | `POST /v1/proxy` (detokenization + forward) | `injector` `ConnectionConfig.tenant_id`, populated by UCS, emitted as `x-tenant-id` |
| Payment server | `GET /payment-methods/token/{token}/details` (`get_permanent_pm_id_from_temporary_token`) | router `transformers.rs` — from `state.tenant.tenant_id` |

The third call was the one missing in the initial fix: it's a direct `http_client::send_request` from the payment server, bypassing both the connector `build_headers` and the injector, so it needed its own `x-tenant-id` header. Without it, the external-vault-proxy **confirm** (detokenize step) failed.

### Additional Changes

## Motivation and Context

When `HyperswitchVault` is the external vault and the vault `base_url` points at the service **in-cluster** (required to avoid the egress forward-proxy / CloudFront returning non-JSON), the call no longer passes through the ingress that injects the tenant. The multi-tenant service then rejects the untenanted request with `HE_05`, breaking both the session-setup and the payment-confirm detokenization for the external-vault-proxy flow.

## How did you test it?

Verified end-to-end on integ with `HyperswitchVault` as the external vault:
- Before: `POST /payments/{id}/confirm` (with `vault_data_card`) failed at the detokenize step — `transformers.rs` "Failed to parse hyperswitch vault token details response" / `HE_05`.
- After deploying the payment server with this change: the detokenize call returns the permanent payment method id and the confirm proceeds through UCS authorize to the connector successfully.
- Reproduced the root cause directly: the same in-cluster `GET …/payment-methods/token/{token}/details` returns `{"error":{…"Missing Tenant ID…","code":"HE_05"}}` without `x-tenant-id`, and `{"id":"0a_pm_…", …}` with it.

## Checklist


XyneSpaces:
## Review Summary

PR #12964 forwards `x-tenant-id` header on HyperswitchVault self-loop calls to fix multi-tenant routing issues. The implementation looks correct but there's one potential issue to address:

### [should-fix] Inconsistent header value type in transformers.rs

In `crates/router/src/core/payment_methods/transformers.rs`, the `x-tenant-id` header uses `get_string_repr().to_string().into()` which creates a `Maskable<String>` via the default `From` implementation.

However, in the connector code (`hyperswitch_vault.rs`), the same header uses `get_string_repr().to_string().into()` pattern. Please verify this produces a `Maskable::new_normal(...)` variant, not `Maskable::Masked(...)`, since tenant IDs are identifiers (not secrets) and don't need masking.

If the `.into()` doesn't produce the expected variant, consider using `Maskable::new_normal(...)` explicitly for consistency with how other ID headers (like `X-PROFILE-ID`) are handled.

---

**Files reviewed:**
- `crates/hyperswitch_connectors/src/connectors/hyperswitch_vault.rs` ✅
- `crates/injector/src/injector.rs` ✅
- `crates/injector/src/types.rs` ✅
- `crates/router/src/core/payment_methods/transformers.rs` (see note above)

The PR correctly identifies the root cause (self-loop calls bypassing edge ingress) and applies the fix at all necessary call sites.

crates/router/src/core/payment_methods/transformers.rs
```diff
@@ -2183,6 +2183,14 @@ pub async fn get_permanent_pm_id_from_temporary_token(
                 headers::X_PROFILE_ID.to_string(),
                 vault_profile_id.expose().into_masked(),
             ),
+            // The hyperswitch vault is the multi-tenant hyperswitch service itself; when called
+            // in-cluster (bypassing the ingress that would otherwise inject the tenant) the tenant
+            // must be forwarded explicitly, else the vault rejects the call with HE_05
+            // ("Missing Tenant ID in the request").
+            (
+                headers::X_TENANT_ID.to_string(),
```
XyneSpaces, line 2191: [blocking] Ensure the `x-tenant-id` header forwarding doesn't create a header injection vulnerability. Validate that `merchant_account.tenant_id` is properly sanitized and doesn't contain CRLF characters before setting it as a header value. The tenant_id should ideally come from a trusted source, but defensive validation here prevents potential HTTP header injection if tenant configuration is ever compromised.

XyneSpaces (commented):
[blocking] Ensure the `x-tenant-id` header forwarding doesn't create a header injection vulnerability. Validate that `merchant_account.tenant_id` is properly sanitized and doesn't contain CRLF characters before setting it as a header value. The tenant_id should ideally come from a trusted source, but defensive validation here prevents potential HTTP header injection if tenant configuration is ever compromised.

PR #12963, opened 2026-06-24
fix(core): populate connector_http_status_code on UCS PSync connector error

AmitsinghTanwar007:
Closes #12986

PR #12971, opened 2026-06-24
fix(ucs): set connector_http_status_code on UCS authorize connector errors

### LINKED ISSUE #12970: UCS authorize connector-error path drops connector_http_status_code (shadow diff)
### Problem
On a UCS shadow authorize, when the connector returns a 4xx/5xx, UCS surfaces the connector error as a gRPC `Status` (decoded by HS into `UnifiedConnectorServiceError::ConnectorError`). HS's authorize gateway converts this into `router_data.response = Err(ErrorResponse { .. })` and early-returns `Ok`, but it never sets `router_data.connector_http_status_code`. It therefore stays `None`, whereas the Direct gateway reports the real connector HTTP status.

This produces a shadow-validation router diff `router.typeDiff:connector_http_status_code` (`hyperswitch: number`, `ucs: null`) on every failed authorize routed through UCS.

Reference (prod shadow validation): juspay/hyperswitch-cloud#17180 (adyen / authorize).

### Fix
In the UCS authorize connector-error path (`crates/router/src/core/payments/gateway/authorize_gateway.rs`), set `router_data.connector_http_status_code = Some(status_code)` before the early `Ok` return, mirroring the success path and the Direct gateway.

AmitsinghTanwar007:
## Shadow-validation diff — adyen / authorize — `router.typeDiff:connector_http_status_code`

Closes #12970

### Diff signature
`router.typeDiff:connector_http_status_code` — `{ "hyperswitch": "number", "ucs": "null" }`
(merchant `merchant_1770760984113`, adyen, authorize). Reproduces on every **failed** authorize routed through UCS.

### Root cause
On a UCS shadow authorize, when the connector returns a 4xx/5xx, UCS surfaces the connector error as a gRPC `Status` (with a proto-encoded `ConnectorError` in the status details). HS decodes it into `UnifiedConnectorServiceError::ConnectorError` and, in the authorize gateway, builds `router_data.response = Err(ErrorResponse { .. })` and early-returns `Ok` — but it **never sets `router_data.connector_http_status_code`**. The later success-path assignment (`router_data.connector_http_status_code = Some(ucs_data.status_code)`) is skipped by the early return, so the field stays `None`.

The Direct gateway, by contrast, reports the real connector HTTP status, so the shadow comparison sees `hyperswitch: <number>` vs `ucs: null` — a `typeDiff`.

Confirmed against prod logs for sample request `019efa7d-914f-7ca1-9e25-4d3712de84d3`:
```
external_services::grpc_client::unified_connector_service: UCS payment authorize gRPC call failed
gateway::authorize_gateway: Connector error via UCS (connector adyen, status 422): CONNECTOR_ERROR_RESPONSE - CVC is not the right length
```
HS-native (Direct) reported `connector_http_status_code = 422`; the UCS path left it `null`.

### Fix (HS → UCS integration, single line)
`crates/router/src/core/payments/gateway/authorize_gateway.rs` — set `router_data.connector_http_status_code = Some(status_code)` in the connector-error branch before the early `Ok` return, mirroring the success path and the Direct gateway. `status_code` is the connector's real HTTP status (e.g. `422`), already decoded into `ConnectorErrorInner`.

This is an HS-side UCS-integration mapping fix only — no connector transformer, proto, or UCS change. (The connector HTTP status is already carried over the wire in `ConnectorError.http_status_code`.)

### Verification
- `cargo build --bin router`: **passes** (clean compile).
- `cargo fmt -- --check` on the changed file: **clean**.
- Behavior (source parity): the connector-error branch now assigns `connector_http_status_code` identically to the success path and the Direct gateway, so the UCS path reports the same numeric status as HS-native → the `typeDiff:connector_http_status_code` is eliminated.
- A live local shadow `no_diff` could not be captured because adyen sandbox is TLS-intercepted by the local MITM and the shared local stack's HS/UCS were not wired through it; verified via prod-log + source parity (the established approach for TLS-blocked adyen shadow diffs).

### Note on sibling diffs
The same early-return branch also hardcodes `connector_transaction_id: None` and uses the generic `code`, which produce the sibling diffs internal tracking issue (`response.Err.connector_transaction_id`) and #17182 (`response.Err.code`). Those are tracked separately; this PR is scoped to `connector_http_status_code`.


XyneSpaces:
## Code Review

**PR:** fix(ucs): set connector_http_status_code on UCS authorize connector errors

### Finding: Potential Missing Error Case Coverage

The change adds `router_data.connector_http_status_code = Some(status_code)` in the UCS error handling path, which mirrors the Direct gateway behavior. However, I notice that the `status_code` variable should be verified to be within valid HTTP status code ranges (100-599) before assignment to avoid potential issues with malformed connector responses.

Consider adding a validation check:
```rust
if (100..=599).contains(&status_code) {
    router_data.connector_http_status_code = Some(status_code);
}
```

This prevents invalid status codes from propagating into metrics and logs.

### Finding: Comment Clarity

The comment is helpful, but consider making it more concise:
```rust
// Mirror Direct gateway: surface HTTP status code on UCS connector errors
// to prevent shadow diffs where Direct reports status but UCS returns None
router_data.connector_http_status_code = Some(status_code);
```

Otherwise, this fix addresses the parity issue correctly.

XyneSpaces:
[should-fix] This fix correctly surfaces `connector_http_status_code` on the UCS error path, which resolves the shadow-validation diff. 

However, consider extracting this into a helper method that both success and error paths can share, rather than duplicating the assignment logic inline. This would make the mirroring more explicit and prevent future drift.

```suggestion
// Consider refactoring to share status code assignment between success and error paths
```

The change itself looks correct and minimal.

XyneSpaces (commented):
Review comment on the new code.

crates/router/src/core/payments/gateway/authorize_gateway.rs
```diff
@@ -309,6 +309,11 @@ where
...
                                         connector_metadata: None,
                                     });
+                                // Surface the connector's HTTP status code on the router data,
+                                // mirroring the success path (and the Direct gateway). Without
+                                // this the field stays `None` on UCS connector errors while the
+                                // Direct path reports the real status, producing a shadow diff.
+                                router_data.connector_http_status_code = Some(status_code);
                                 // Return Ok with router_data containing the error response
                                 // This ensures the connector error flows through the normal
                                 // response handling path (same as direct connector errors)
```
XyneSpaces, line 322: [nit] The comment explains the 'why' well, but consider if this status_code propagation could be extracted to a shared helper to avoid repeating this logic across authorize_gateway.rs and psync_gateway.rs (seen in #12963). This would reduce shadow-diff risk between the two gateways.

AmitsinghTanwar007:
closed by #13003

PR #12988, opened 2026-06-25
fix(router): render Barclaycard 3DS DDC/step-up forms correctly via UCS

shuklatushar226:
## Type of Change
- [x] Bugfix

## Description

Barclaycard 3DS works HS-Direct (native connector) but fails through the **Unified Connector Service (UCS)** — the path Barclaycard actually uses. The device-data-collection (DDC) step hangs on "Please wait while we process your payment…", the CardinalCommerce `Cruise/Collect` response is downloaded instead of running in the hidden DDC iframe, and the step-up challenge never renders, so the payment never reaches `charged`.

### Root cause

3DS DDC/step-up forms are carried by the typed `RedirectForm` enum. HS renders the typed `BarclaycardAuthSetup` / `CybersourceAuthSetup` variants as a **hidden `<iframe>`** that POSTs `Collect` inside the iframe and waits for a `postMessage` from `cardinaltrusted.com` (`services/api.rs`). The generic `RedirectForm::Form` variant instead renders a **top-level auto-submitting form**, which makes the `Collect` response load/download at the top level — breaking the DDC handshake.

The gRPC `RedirectForm` proto has no typed DDC variant, so UCS flattens `CybersourceAuthSetup` / `CybersourceConsumerAuth` into a generic `Form`. HS reconstructs the typed variant from that generic `Form`, but the reconstruction was gated on `Connector::Cybersource` only. For `Connector::Barclaycard` it fell through to `_ => Ok(response_data)`, leaving the generic `Form` → top-level render → DDC download → 3DS stall.

### Fix

Extend the PreAuthenticate (DDC) and Authenticate (step-up) reconstruction to also cover `Connector::Barclaycard`, emitting the existing `BarclaycardAuthSetup` / `BarclaycardConsumerAuth` variants (which render identically to the already-working HS-Direct path). UCS already emits the exact field names the existing logic expects (`access_token`, `ddc_url`, `reference_id`; `step_up_url`), so the extraction is reused verbatim.

- `crates/router/src/core/payments/flows/authorize_flow.rs` — `transform_redirection_response_for_pre_authenticate_flow` + its `TransactionResponse` gate
- `crates/router/src/core/payments/flows/complete_authorize_flow.rs` — `transform_redirection_response_for_authenticate_flow` + its `TransactionResponse` gate

No proto, connector, or UCS changes required.

## How did you test it?

- `cargo check -p router` passes.
- Cypress `05-ThreeDSAutoCapture` for Barclaycard: previously failed at the DDC step (`Host changed or an iframe with foreign host exist`, `Collect` downloaded). With this fix the DDC renders in the hidden iframe, the Cardinal challenge loads, and the payment completes via PreAuthenticate → Authenticate → PostAuthenticate → Authorize.

## Checklist

### DIFF
--- a/config/config.example.toml
+++ b/config/config.example.toml
@@ -1340,7 +1340,7 @@ url = "http://localhost:8080"           # Open Router URL
 [grpc_client.unified_connector_service]
 base_url = "http://localhost:8000"      # Unified Connector Service Base URL
 connection_timeout = 10                 # Connection Timeout Duration in Seconds
-ucs_only_connectors = "imerchantsolutions, paytm, phonepe, hyperpg, revolv3, fiservcommercehub, absa_sanlam, interpayments, dlocal"    # Comma-separated list of connectors that use UCS only
+ucs_only_connectors = "imerchantsolutions, paytm, phonepe, hyperpg, revolv3, fiservcommercehub, absa_sanlam, interpayments, dlocal, barclaycard"    # Comma-separated list of connectors that use UCS only
 ucs_psync_disabled_connectors = "cashtocode, trustly"    # Comma-separated list of connectors to disable UCS PSync call
 
 [grpc_client.recovery_decider_client] # Revenue recovery client base url
--- a/config/deployments/env_specific.toml
+++ b/config/deployments/env_specific.toml
@@ -412,7 +412,7 @@ payment_method = "payment_method_types" # <payment_method> = comma-separated clo
 [grpc_client.unified_connector_service]
 base_url = "http://localhost:8000"      # Unified Connector Service Base URL
 connection_timeout = 10                 # Connection Timeout Duration in Seconds
-ucs_only_connectors = "imerchantsolutions, paytm, phonepe, hyperpg, revolv3, fiservcommercehub, absa_sanlam, interpayments, payconex, dlocal"    # Comma-separated list of connectors that use UCS only
+ucs_only_connectors = "imerchantsolutions, paytm, phonepe, hyperpg, revolv3, fiservcommercehub, absa_sanlam, interpayments, payconex, dlocal, barclaycard"    # Comma-separated list of connectors that use UCS only
 ucs_psync_disabled_connectors = "cashtocode, trustly"    # Comma-separated list of connectors to disable UCS PSync call
 
 [revenue_recovery]
--- a/config/deployments/integration_test.toml
+++ b/config/deployments/integration_test.toml
@@ -1020,7 +1020,7 @@ connector_list = "juspaythreedsserver, ctp_mastercard, ctp_visa"
 connector_list = "worldpayvantiv"
 
 [grpc_client.unified_connector_service]
-ucs_only_connectors = "imerchantsolutions, paytm, phonepe, hyperpg, revolv3, fiservcommercehub, absa_sanlam, interpayments, payconex, dlocal"    # Comma-separated list of connectors that use UCS only
+ucs_only_connectors = "imerchantsolutions, paytm, phonepe, hyperpg, revolv3, fiservcommercehub, absa_sanlam, interpayments, payconex, dlocal, barclaycard"    # Comma-separated list of connectors that use UCS only
 ucs_psync_disabled_connectors = "cashtocode, trustly"    # Comma-separated list of connectors to disable UCS PSync call
 
 # Merchant Advice Code Configuration
--- a/config/deployments/production.toml
+++ b/config/deployments/production.toml
@@ -1028,7 +1028,7 @@ click_to_pay = {connector_list = "adyen, cybersource, trustpay"}
 connector_list = "juspaythreedsserver, ctp_mastercard, ctp_visa"
 
 [grpc_client.unified_connector_service]
-ucs_only_connectors = "imerchantsolutions, paytm, phonepe, hyperpg, revolv3, fiservcommercehub, absa_sanlam, interpayments, payconex, dlocal"    # Comma-separated list of connectors that use UCS only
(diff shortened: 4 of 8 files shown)


XyneSpaces:
## Review Summary

This PR extends 3DS DDC and step-up form handling to support Barclaycard alongside Cybersource. Overall the change follows established patterns, but there are areas worth addressing:

### [should-fix] Code Duplication in Form Construction

The pattern of duplicating form construction logic for each connector increases maintenance burden:

```rust
// Current pattern (duplicated)
match connector {
    Connector::Barclaycard => Ok(RedirectForm::BarclaycardAuthSetup { ... }),
    _ => Ok(RedirectForm::CybersourceAuthSetup { ... }),
}
```

**Suggestion**: Consider extracting common fields into a shared helper that accepts the variant type as a parameter. This reduces the risk of inconsistencies when updating form fields.

### [nit] Verify Exhaustive Connector Matching

The change extends the match arm to include Barclaycard:

```rust
Connector::Cybersource | Connector::Barclaycard
```

Please confirm that:
1. No wildcard `_` patterns are used downstream that might bypass this explicit handling
2. Any new connectors requiring 3DS form reconstruction will be explicitly added (not caught by wildcards)

### [nit] SDK Compatibility Check

The PR creates new `RedirectForm` variants (`BarclaycardAuthSetup`, `BarclaycardConsumerAuth`). Please verify that:
- The Web SDK has corresponding handling for these new variants
- Older SDK versions have appropriate fallback behavior for unrecognized variants

### Security Verification

For the 3DS hidden iframe forms:
- Confirm iframe `sandbox` attributes are correctly applied
- Verify form actions are HTTPS-validated before rendering
- Ensure no PII leaks into 3DS method URL query parameters

---

Overall approach is sound. The main concern is the long-term maintainability with duplicated logic patterns.

XyneSpaces:
## Review Summary: Barclaycard 3DS DDC/Step-up Form Handling

### 🔴 [blocking] Verify `RedirectForm::Barclaycard` variant exists

The PR adds handling for `RedirectForm::BarclaycardAuthSetup` and `RedirectForm::BarclaycardConsumerAuth` variants. Please confirm these enum variants exist in the `RedirectForm` definition (in `api_models` or `hyperswitch_domain_models`). Without the variant definitions, this will cause compilation errors.

### 🟡 [should-fix] Verify field extraction error handling

When extracting form fields in both flows, ensure missing fields produce clear errors:
```rust
let access_token = form_fields.get("access_token")
    .ok_or(errors::ApiErrorResponse::MissingRequiredField { 
        field_name: "access_token" 
    })?;
```

The current code uses `ok_or(...)?` patterns - verify these map to appropriate error responses.

### 🟡 [should-fix] Field name mapping verification

UCS may use different field names for Barclaycard vs Cybersource. Verify field keys like `"JWT"`, `"accessToken"`, or `"jwt"` match what UCS actually sends for Barclaycard 3DS responses.

### 🟢 [nit] Pattern completeness

Consider adding a brief comment explaining why Barclaycard shares Cardinal infrastructure with Cybersource for maintainers unfamiliar with this connector relationship.

---

**Overall:** Pattern follows existing Cybersource implementation correctly. Verify enum variants exist and field mappings are correct before merge.

XyneSpaces:
## PR Review: fix(router): render Barclaycard 3DS DDC/step-up forms correctly via UCS

### Summary
Extends existing Cybersource-specific 3DS handling to also cover Barclaycard for both DDC (PreAuthenticate) and step-up (Authenticate) flows.

### Findings

**[should-fix] Pattern matching on Connector enum could use exhaustive approach**

The diff shows `Connector::Cybersource | Connector::Barclaycard` pattern additions. Consider:

1. Are there other connectors that use the same 3DS flow? If so, a trait-based approach or connector capability check would be more scalable than extending the match arm repeatedly.

2. The connector-specific branching (`BarclaycardAuthSetup` vs `CybersourceAuthSetup`) suggests the abstraction could be cleaner. Consider if these could share a common `ThreeDSAuthSetup` variant with connector-specific metadata.

**[nit] Code duplication between authorize_flow and complete_authorize_flow**

Similar Barclaycard/Cybersource branching appears in both files. Consider extracting a helper function or trait method for connector-specific 3DS type selection.

---
*Review conducted by XyneSpaces automated PR reviewer*

XyneSpaces:
## Review: PR #12988 - feat(connector): add mandate support for BamboraApac

### Summary
Extends 3DS redirect form reconstruction logic to support Barclaycard alongside Cybersource, ensuring UCS-flattened forms are correctly typed for hidden-iframe rendering during device data collection and consumer authentication flows.

### Findings

✅ **Architecture alignment**: Following the established pattern of connector-specific redirect form variants is correct for maintaining type safety across the authentication flow.

✅ **Documentation**: Inline comments clearly explain the UCS flattening concern and why the reconstruction is necessary.

⚠️ **[blocking] Exhaustiveness consideration**: The inner `match connector` statement uses `_ =>` wildcard fallback for Cybersource. Since the outer match arm explicitly matches on both `Cybersource | Barclaycard`, the wildcard is technically unreachable for other variants. However, if a third connector is added to the outer match in the future, the wildcard would incorrectly route it to Cybersource handling.

**Recommendation**: Consider making the inner match exhaustive for the connectors handled in the outer match:
```rust
match connector {
    connector_enums::Connector::Barclaycard => /* ... */,
    connector_enums::Connector::Cybersource => /* ... */,
}
```

This ensures compilation errors if the outer match is extended without updating the inner logic.

💡 **[nit] Code duplication**: Both `transform_redirection_response_for_pre_authenticate_flow` and `transform_redirection_response_for_authenticate_flow` have nearly identical inner match structures. Consider extracting a helper function if more connectors follow this pattern.

**Overall**: Important fix for Barclaycard 3DS support. Recommend addressing the exhaustiveness concern for maintainability.

hrithikesh026 (approved):

deepanshu-iiitu (approved):

gopikrishna000 (approved):

PR #13049, opened 2026-06-26
feat(router): [affirm] UCS support — config header, CompleteAuthorize payment_method, auto-capture

shuklatushar226:
## What
Adds Hyperswitch UCS (Unified Connector Service) support for the **Affirm** (BNPL redirect) connector, so payments complete end-to-end through UCS and reach `charged`. Pairs with the prism/connector-service affirm Authorize work.

## Changes
- **`unified_connector_service/connector_config.rs`** — add `ConnectorSpecificConfig::Affirm { public_key, private_key }` variant + the `Connector::Affirm` match arm (BodyKey `api_key`→`public_key`, `key1`→`private_key`). Without it `build_connector_config_header` hit the catch-all `Err` for affirm → `HE_00` (`FailedToObtainAuthType`) on the INITIATE leg.
- **`operations/payment_complete_authorize.rs`** — on the redirect-return CompleteAuthorize, `payment_method_data` is `None` (no card/token to rebuild via `make_pm_data`) while `payment_method_type` is `Affirm`. Reconstruct the fieldless `PayLater(AffirmRedirect)` from the type so the UCS Authorize request carries `payment_method` and the connector can run the transaction-create (complete) leg with the `checkout_token`. Without it UCS rejected the complete leg with `InvalidArgument`.
- **`flows.rs` (`should_initiate_capture_flow`)** — Affirm authorizes the loan on the complete leg and needs a follow-up capture (`POST /transactions/{id}/capture`) to settle; chain the capture flow for `Affirm` + `CaptureMethod::Automatic` so automatic-capture payments reach `Charged` (previously stuck at `requires_capture`).

## Scope
Affirm-only. Other redirect pay-later methods (Klarna/Afterpay) would want the same two arms (`payment_complete_authorize` reconstruct + `should_initiate_capture_flow`) — cheap to extend.

## Testing
Verified end-to-end against the Affirm sandbox via UCS (`ucs_only`), all implemented flows driven through the HS router (`:8085`) → UCS → Affirm.

> UCS-only `test-prism` / `grpcurl` can't validate the full flow — the Affirm hosted-checkout redirect can't be completed headlessly (INITIATE succeeds, then there's no `checkout_token`). Use the HTTP flow below and complete the redirect in a browser.

**Complete the redirect (browser):** open `next_action.redirect_to_url` → phone → OTP `123456` → pick a plan → ACH `021000021` / `000123456789` → agree → **Confirm**.

Set `KEY=<MERCHANT_API_KEY>` first.

**Create (automatic capture):**
```bash
curl -s -X POST http://localhost:8085/payments -H 'Content-Type: application/json' -H "api-key: $KEY" -d '{
  "amount":6000,"currency":"USD","confirm":true,"capture_method":"automatic","authentication_type":"no_three_ds",
  "customer_id":"cus_affirm","email":"buyer@example.com","name":"John Doe","return_url":"https://google.com",
  "routing":{"type":"single","data":"affirm"},
  "payment_method":"pay_later","payment_method_type":"affirm","payment_method_data":{"pay_later":{"affirm_redirect":{}}},
  "billing":{"address":{"first_name":"John","last_name":"Doe","line1":"123 Main St","city":"San Francisco","state":"CA","zip":"94105","country":"US"},"phone":{"number":"9085551234","country_code":"+1"}},
  "shipping":{"address":{"first_name":"John","last_name":"Doe","line1":"123 Main St","city":"San Francisco","state":"CA","zip":"94105","country":"US"}},
  "order_details":[{"product_name":"Test","quantity":1,"amount":6000,"account_name":"transaction_processing"}]
}'
# → requires_customer_action + redirect_to_url; complete redirect → auto-capture → status: succeeded, amount_received: 6000
```
Manual capture: same body with `"capture_method":"manual"` → `requires_capture` after the redirect.

**Capture / PSync / Refund / RSync / Void:**
```bash
# Full capture (Affirm allows full amount only)
curl -s -X POST http://localhost:8085/payments/<PAY_ID>/capture -H 'Content-Type: application/json' -H "api-key: $KEY" -d '{"amount_to_capture":6000}'
# PSync
curl -s "http://localhost:8085/payments/<PAY_ID>?force_sync=true" -H "api-key: $KEY"
# Refund (payment must be charged)
curl -s -X POST http://localhost:8085/refunds -H 'Content-Type: application/json' -H "api-key: $KEY" -d '{"payment_id":"<PAY_ID>","amount":6000,"reason":"customer_request"}'
# Refund sync (RSync)
curl -s "http://localhost:8085/refunds/<REFUND_ID>?force_sync=true" -H "api-key: $KEY"
# Void / cancel (payment must be requires_capture)
curl -s -X POST http://localhost:8085/payments/<PAY_ID>/cancel -H 'Content-Type: application/json' -H "api-key: $KEY" -d '{"cancellation_reason":"requested_by_customer"}'
```

**Coverage:**

| Flow | Result |
|---|---|
| Authorize (INITIATE + COMPLETE) | ✅ loan authorized |
| Capture — automatic, full | ✅ `succeeded`, amount_received 6000 |
| Capture — manual, full | ✅ `succeeded` |
| Capture — partial | ⛔ Affirm rejects (`409`, full-amount only) — integration forwards `amount` and surfaces the error |
| PSync | ✅ |
| Refund + RSync | ✅ `REFUND_SUCCESS` |
| Void / cancel | ✅ `cancelled` |

Payment method: only `PayLater(AffirmRedirect)` (BNPL) is implemented — fully covered.

> Note: Affirm supports **full-amount capture only** — a partial `amount_to_capture` returns `409 "cannot be captured for an amount unequal to authorization hold amount"`, which the integration surfaces (payment → `failed`).

### DIFF
--- a/crates/router/src/core/payments/flows.rs
+++ b/crates/router/src/core/payments/flows.rs
@@ -363,20 +363,23 @@ pub fn should_initiate_capture_flow(
     status: common_enums::AttemptStatus,
 ) -> bool {
     match status {
-        common_enums::AttemptStatus::Authorized => {
-            if let Some(api_enums::CaptureMethod::SequentialAutomatic) = capture_method {
-                match connector_name {
-                    router_types::Connector::Paybox => {
-                        // Check CIT conditions for Paybox
-                        setup_future_usage == Some(api_enums::FutureUsage::OffSession)
-                            && customer_acceptance.is_some()
-                    }
-                    _ => false,
+        common_enums::AttemptStatus::Authorized => match capture_method {
+            Some(api_enums::CaptureMethod::SequentialAutomatic) => match connector_name {
+                router_types::Connector::Paybox => {
+                    // Check CIT conditions for Paybox
+                    setup_future_usage == Some(api_enums::FutureUsage::OffSession)
+                        && customer_acceptance.is_some()
                 }
-            } else {
-                false
+                _ => false,
+            },
+            // Affirm (BNPL) authorizes the loan on the transaction-create (CompleteAuthorize)
+            // leg and needs a follow-up capture (POST /transactions/{id}/capture) to settle for
+            // automatic capture, so chain the capture flow when it comes back Authorized.
+            Some(api_enums::CaptureMethod::Automatic) => {
+                matches!(connector_name, router_types::Connector::Affirm)
             }
-        }
+            _ => false,
+        },
         _ => false,
     }
 }
--- a/crates/router/src/core/payments/operations/payment_complete_authorize.rs
+++ b/crates/router/src/core/payments/operations/payment_complete_authorize.rs
@@ -491,19 +491,6 @@ impl<F: Clone + Send + Sync> Domain<F, api::PaymentsRequest, PaymentData<F>> for
             should_retry_with_pan,
         ))
         .await?;
-        // For redirect pay-later PMs (e.g. Affirm BNPL), the redirect-completion request
-        // carries no payment_method_data and there is no stored card/token, so make_pm_data
-        // returns None. Reconstruct the stateless PayLater(AffirmRedirect) from the persisted
-        // payment_method_type so the UCS CompleteAuthorize request carries payment_method and
-        // the connector can run the COMPLETE (transaction-create) leg with the checkout_token.
-        let payment_method_data = payment_method_data.or_else(|| {
-            match payment_data.payment_attempt.payment_method_type {
-                Some(storage_enums::PaymentMethodType::Affirm) => Some(
-                    domain::PaymentMethodData::PayLater(domain::PayLaterData::AffirmRedirect {}),
-                ),
-                _ => None,
-            }
-        });
         Ok((op, payment_method_data, pm_id))
     }
 
--- a/crates/router/src/core/payments/operations/payment_complete_authorize.rs
+++ b/crates/router/src/core/payments/operations/payment_complete_authorize.rs
@@ -491,6 +491,19 @@ impl<F: Clone + Send + Sync> Domain<F, api::PaymentsRequest, PaymentData<F>> for
             should_retry_with_pan,
         ))
         .await?;
+        // For redirect pay-later PMs (e.g. Affirm BNPL), the redirect-completion request
+        // carries no payment_method_data and there is no stored card/token, so make_pm_data
+        // returns None. Reconstruct the stateless PayLater(AffirmRedirect) from the persisted
+        // payment_method_type so the UCS CompleteAuthorize request carries payment_method and
+        // the connector can run the COMPLETE (transaction-create) leg with the checkout_token.
+        let payment_method_data = payment_method_data.or_else(|| {
+            match payment_data.payment_attempt.payment_method_type {
+                Some(storage_enums::PaymentMethodType::Affirm) => Some(
+                    domain::PaymentMethodData::PayLater(domain::PayLaterData::AffirmRedirect {}),
+                ),
+                _ => None,
+            }
+        });
         Ok((op, payment_method_data, pm_id))
     }
 
--- a/config/config.example.toml
+++ b/config/config.example.toml
@@ -1340,7 +1340,7 @@ url = "http://localhost:8080"           # Open Router URL
 [grpc_client.unified_connector_service]
 base_url = "http://localhost:8000"      # Unified Connector Service Base URL
 connection_timeout = 10                 # Connection Timeout Duration in Seconds
-ucs_only_connectors = "imerchantsolutions, paytm, phonepe, hyperpg, revolv3, fiservcommercehub, absa_sanlam, interpayments, dlocal"    # Comma-separated list of connectors that use UCS only
+ucs_only_connectors = "imerchantsolutions, paytm, phonepe, hyperpg, revolv3, fiservcommercehub, absa_sanlam, interpayments, dlocal, affirm"    # Comma-separated list of connectors that use UCS only
 ucs_psync_disabled_connectors = "cashtocode, trustly"    # Comma-separated list of connectors to disable UCS PSync call
 
 [grpc_client.recovery_decider_client] # Revenue recovery client base url
--- a/config/config.example.toml
+++ b/config/config.example.toml
@@ -1340,7 +1340,7 @@ url = "http://localhost:8080"           # Open Router URL
 [grpc_client.unified_connector_service]
 base_url = "http://localhost:8000"      # Unified Connector Service Base URL
 connection_timeout = 10                 # Connection Timeout Duration in Seconds
-ucs_only_connectors = "imerchantsolutions, paytm, phonepe, hyperpg, revolv3, fiservcommercehub, absa_sanlam, interpayments, dlocal, affirm"    # Comma-separated list of connectors that use UCS only
+ucs_only_connectors = "imerchantsolutions, paytm, phonepe, hyperpg, revolv3, fiservcommercehub, absa_sanlam, interpayments, dlocal"    # Comma-separated list of connectors that use UCS only
 ucs_psync_disabled_connectors = "cashtocode, trustly"    # Comma-separated list of connectors to disable UCS PSync call
 
 [grpc_client.recovery_decider_client] # Revenue recovery client base url
--- a/config/deployments/env_specific.toml
+++ b/config/deployments/env_specific.toml
@@ -412,7 +412,7 @@ payment_method = "payment_method_types" # <payment_method> = comma-separated clo
 [grpc_client.unified_connector_service]
 base_url = "http://localhost:8000"      # Unified Connector Service Base URL
 connection_timeout = 10                 # Connection Timeout Duration in Seconds
-ucs_only_connectors = "imerchantsolutions, paytm, phonepe, hyperpg, revolv3, fiservcommercehub, absa_sanlam, interpayments, payconex, dlocal"    # Comma-separated list of connectors that use UCS only
+ucs_only_connectors = "imerchantsolutions, paytm, phonepe, hyperpg, revolv3, fiservcommercehub, absa_sanlam, interpayments, payconex, dlocal, affirm"    # Comma-separated list of connectors that use UCS only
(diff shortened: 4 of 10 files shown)


XyneSpaces:
[should-fix] Verify BodyKey auth type for Affirm UCS config

The `ConnectorSpecificConfig::Affirm { public_key, private_key }` variant maps to `BodyKey` authentication. Please verify:

1. **Auth type correctness**: Confirm Affirm uses `BodyKey` (not `HeaderKey` or `SignatureKey`) for UCS API calls in production.

2. **Field mapping**: Verify `public_key`/`private_key` naming matches `MerchantConnectorAccount`'s `api_key`/`key1` fields exactly.

3. **Complete match coverage**: Ensure all `match connector` expressions in `connector_config.rs` have the `Affirm` arm — missing arms cause `HE_00 (FailedToObtainAuthType)` at runtime.

4. **Secret masking**: Confirm `UcsConnectorConfig::new()` masks secrets in logs and error traces when constructing the `X-Connector-Config` header.

XyneSpaces:
[should-fix] Validate payment_method_type before PayLater reconstruction

In `payment_complete_authorize.rs`, the reconstruction of `PayLater(AffirmRedirect)` when `payment_method_data` is None should validate the stored `payment_method_type` is actually `AffirmRedirect`:

```rust
// Current code (add validation):
if payment_attempt.payment_method_type == Some(api_models::enums::PaymentMethodType::AffirmRedirect) {
    Some(api_models::payments::PaymentMethodData::PayLater(
        api_models::payments::PayLaterData::AffirmRedirect,
    ))
}
```

Without this check, corrupted or mismatched `payment_method_type` values could cause incorrect payment method injection. Confirm this validation happens before the reconstruction path is taken.

XyneSpaces:
[should-fix] Auto-capture logic verification for Affirm

The `should_initiate_capture_flow` function now chains capture for Affirm with `SequentialAutomatic` capture method. Please verify:

1. **Capture method selection**: Confirm Affirm requires `SequentialAutomatic` (not standard `Automatic`). Paybox uses this for CIT-specific flows with `setup_future_usage` checks — Affirm's implementation lacks these conditions. If Affirm needs unconditional capture, consider if `CaptureMethod::Automatic` is more appropriate.

2. **Timing of authorize**: Affirm's authorize happens on the complete leg (post-redirect). Ensure the `AttemptStatus::Authorized` check properly accounts for this timing.

```rust
// Current code captures for Affirm unconditionally when SequentialAutomatic
// Compare with Paybox which checks: setup_future_usage == Some(OffSession) && customer_acceptance.is_some()
```

3. **Klarna comparison**: As another PayLater/BNPL connector, compare Klarna's UCS implementation to ensure Affirm follows the same auto-capture patterns.

XyneSpaces:
[should-fix] The Affirm capture flow uses `CaptureMethod::Automatic`, but the comment mentions "automatic capture" which could mean `SequentialAutomatic`. Please verify that merchants configure Affirm with `capture_method: automatic` and not `sequential_automatic`, otherwise the capture flow won't trigger.

```suggestion
            // Affirm (BNPL) authorizes the loan on the transaction-create (CompleteAuthorize)
            // leg and needs a follow-up capture (POST /transactions/{id}/capture) to settle for
            // automatic capture, so chain the capture flow when it comes back Authorized.
            // Note: This specifically handles CaptureMethod::Automatic, not SequentialAutomatic.
```

XyneSpaces:
[should-fix] **Affirm auto-capture logic scope**: The `matches!` for Affirm in `should_initiate_capture_flow` is clean, but this function is growing complex with connector-specific logic. Consider extracting connector-specific capture behavior into a trait method or configuration-driven approach to maintain maintainability as more connectors are added.

---

[nit] **Config file duplication**: The `affirm` addition to `ucs_only_connectors` is repeated across 6 config files. Consider if these could be unified or inherited from a base configuration to reduce maintenance overhead when adding new UCS connectors.

---

[nit] **Comment clarity**: The comment explaining Affirm's capture behavior ("authorizes the loan on the transaction-create...") is helpful but could reference external documentation or API docs for future maintainers who need to understand why this specific connector needs automatic capture chaining.

XyneSpaces:
## PR Review Summary

### [blocking] `should_initiate_capture_flow` logic may have precedence issue

The new match arm for `CaptureMethod::Automatic` with Affirm check is nested inside the `AttemptStatus::Authorized` match. This is correct, but consider if there are any other connectors that might also need similar special handling. The pattern is:

```rust
Some(api_enums::CaptureMethod::Automatic) => {
    matches!(connector_name, router_types::Connector::Affirm)
}
```

This introduces a precedent where connector-specific logic is being added to generic flow code. Consider:
1. Documenting this pattern in code comments explaining why Affirm needs special handling
2. Checking if this should be data-driven via connector metadata rather than hardcoded

### [should-fix] Missing error handling in `payment_complete_authorize.rs`

The `or_else` block that reconstructs `PayLaterData::AffirmRedirect` doesn't account for cases where the `payment_method_type` might be `Affirm` but the connector expects additional fields. Verify that `AffirmRedirect {}` (empty struct) is sufficient for the UCS CompleteAuthorize flow.

### [nit] Connector config field naming consistency

In `connector_config.rs`, the Affirm config uses `public_key` and `private_key` which map from `api_key` and `key1`. Consider adding a comment explaining this mapping, as "public_key/private_key" terminology differs from the standard `api_key/key1` naming used in other connectors.

XyneSpaces:
[should-fix] The `make_pm_data` implementation for Affirm uses a broad pattern match on `payment_method_type` that could mask other redirect-based payment methods needing similar treatment.

The current fallback logic:
```rust
let payment_method_data = payment_method_data.or_else(|| {
    match payment_data.payment_attempt.payment_method_type {
        Some(storage_enums::PaymentMethodType::Affirm) => Some(...),
        _ => None,
    }
});
```

Consider whether this pattern should be extensible for other BNPL providers (Klarna, Afterpay, etc.) that may also need stateless PayLater data reconstruction. Documenting the rationale for Affirm specifically would help future maintainers understand when to extend this pattern.

Location: `crates/router/src/core/payments/operations/payment_complete_authorize.rs` around line 495

XyneSpaces (commented):
Review completed for Affirm UCS support PR. Key findings: 1) Config consistency verified across all 6 deployment files, 2) Credentials properly wrapped in Secret<String>, 3) One potential issue identified in auto-capture flow logic - see inline comment. No blocking security issues found. Recommend testing auto-capture flows for all connectors using CaptureMethod::Automatic before merge to prevent regression.

crates/router/src/core/payments/operations/payment_complete_authorize.rs
```diff
@@ -491,6 +491,19 @@ impl<F: Clone + Send + Sync> Domain<F, api::PaymentsRequest, PaymentData<F>> for
...
         ))
         .await?;
+        // For redirect pay-later PMs (e.g. Affirm BNPL), the redirect-completion request
+        // carries no payment_method_data and there is no stored card/token, so make_pm_data
+        // returns None. Reconstruct the stateless PayLater(AffirmRedirect) from the persisted
+        // payment_method_type so the UCS CompleteAuthorize request carries payment_method and
+        // the connector can run the COMPLETE (transaction-create) leg with the checkout_token.
+        let payment_method_data = payment_method_data.or_else(|| {
+            match payment_data.payment_attempt.payment_method_type {
+                Some(storage_enums::PaymentMethodType::Affirm) => Some(
```
XyneSpaces: [nit] Consider adding a comment explaining why Affirm payment method reconstruction is safe here. The code assumes that if payment_method_data is None AND payment_method_type is Affirm, we should inject AffirmRedirect. This could mask issues if:
1) A different connector also uses PaymentMethodType::Affirm but requires different data
2) Future PM types need similar treatment

A brief comment about UCS flow context would help future maintainers.

Suggestion: `// UCS redirect flow: Affirm returns no payment data on callback, reconstruct from persisted type`

crates/router/src/core/payments/flows.rs
```diff
@@ -363,20 +363,23 @@ pub fn should_initiate_capture_flow(
...
+                    // Check CIT conditions for Paybox
+                    setup_future_usage == Some(api_enums::FutureUsage::OffSession)
+                        && customer_acceptance.is_some()
                 }
-            } else {
-                false
+                _ => false,
+            },
+            // Affirm (BNPL) authorizes the loan on the transaction-create (CompleteAuthorize)
+            // leg and needs a follow-up capture (POST /transactions/{id}/capture) to settle for
```
XyneSpaces, line 376: [should-fix] The auto-capture flow refactor changes behavior for CaptureMethod::Automatic. Previously, connectors with Automatic capture would fall through to false (no auto-capture). Now, ONLY Affirm with Automatic triggers capture. This could break auto-capture for other connectors if they relied on the implicit non-capture behavior. Please confirm: 1) Are there other connectors using CaptureMethod::Automatic that should NOT auto-capture? 2) Should this use a config-driven approach instead of hardcoding Affirm?

```suggestion
            Some(api_enums::CaptureMethod::Automatic) => {
                // Explicit list of connectors requiring explicit capture even with Automatic
                // This makes the behavior intentional rather than accidental
                const CONNECTORS_REQUIRING_EXPLICIT_CAPTURE: &[router_types::Connector] = &[
                    router_types::Connector::Affirm,
                    // Add other connectors here if needed
                ];
                CONNECTORS_REQUIRING_EXPLICIT_CAPTURE.contains(connector_name)
            }
```

XyneSpaces (commented):
Review comment on the auto-capture flow logic for Affirm connector.

crates/router/src/core/payments/operations/payment_complete_authorize.rs
```diff
@@ -491,6 +491,19 @@ impl<F: Clone + Send + Sync> Domain<F, api::PaymentsRequest, PaymentData<F>> for
             should_retry_with_pan,
         ))
         .await?;
+        // For redirect pay-later PMs (e.g. Affirm BNPL), the redirect-completion request
```
hrithikesh026: this should be solved in make_pm_data itself.

hrithikesh026 (changes requested):

hrithikesh026 (approved):

AmitsinghTanwar007 (approved):

PR #13073, opened 2026-06-29
fix(docker): pin postgres image to 18 to avoid floating-tag breakage

### LINKED ISSUE #13072: Local `docker compose up` fails: `pg` container exits (1) due to unpinned `postgres:latest` (now PG18) + stale data volume
## Summary
A fresh / repeated `docker compose up` on a local machine fails to start. The visible errors are downstream:

```
Container hyperswitch-migration_runner-1  service "migration_runner" didn't complete successfully: exit 1
Container hyperswitch-superposition-init-1  Exited
dependency failed to start: container hyperswitch-pg-1 exited (1)
[INFO] Cleaning up any started containers...
```

The **actual** failure is the `pg` container exiting with code 1; `migration_runner` and `superposition-init` only cascade from it.

## Root cause
`docker-compose.yml` pins the database to a floating tag:

```yaml
pg:
  image: docker.io/postgres:latest
  volumes:
    - pg_data:/var/lib/postgresql
```

`postgres:latest` now resolves to **PostgreSQL 18**, which shipped a breaking change to the image's data-directory layout: `PGDATA` is now version-specific (`/var/lib/postgresql/18/docker`) and the declared `VOLUME` moved to `/var/lib/postgresql` (see [docker-library/postgres#1259](https://github.com/docker-library/postgres/pull/1259) and the [official image notes](https://hub.docker.com/_/postgres)).

A `pg_data` volume created by an earlier run (PG17 or older, or an older compose that mounted `/var/lib/postgresql/data`) is incompatible with PG18, so Postgres aborts at startup:

```
FATAL: database files are incompatible with server
DETAIL: The data directory was initialized by PostgreSQL version 17, which is not compatible with this version 18.x
```

Because the tag is unpinned, this hits new contributors with no action on their part whenever Docker Hub rolls the major version. Reported by more than one user.

## Repro
1. Have a `pg_data` volume from a previous run (or simply pull a newer `postgres:latest`).
2. `docker compose up`
3. `pg` exits (1) → `migration_runner` exit 1 → setup cleans up.

## Workaround (confirmed working)
```bash
docker compose down -v
docker volume ls | grep pg      # find the real volume name
docker volume rm <name>         # if down -v didn't already remove it
docker compose up
```

## Suggested durable fix
Make the local DB reproducible instead of tracking a floating major version. Options (pick one):
1. **Pin the image** — e.g. `image: docker.io/postgres:17` in `docker-compose.yml` (and `docker-compose-development.yml` if it duplicates the service).
2. **Pin + future-proof** — keep `postgres:latest` but set `PGDATA` and align the volume mount so major bumps don't break existing volumes.

Pinning (option 1) is the lowest-risk fix and matches what most projects did in response to the PG18 change.

## Environment
- macOS, Docker Desktop
- `latest` tag per README quickstart

XyneSpaces:
## Summary

`docker-compose.yml` and `docker-compose-development.yml` pin the `pg` service to `docker.io/postgres:latest`. `latest` now resolves to **PostgreSQL 18**, whose internal data-directory layout changed. A fresh `docker compose up` from a clean clone works, but any environment that pulls a newer `latest` than its existing `pg_data` volume gets an incompatible data dir — `hyperswitch-pg-1` exits `1`, which cascades into `migration_runner` (exit 1) and `superposition-init` failing.

This pins both compose files to a fixed major (`postgres:18`, matching the current `latest` and the existing `/var/lib/postgresql` mount) so future `latest` bumps no longer silently break local setup.

Closes #13072

## Changes
- `docker-compose.yml` — `pg.image`: `postgres:latest` → `postgres:18`
- `docker-compose-development.yml` — `pg.image`: `postgres:latest` → `postgres:18`

## Test evidence
- `docker compose -f docker-compose.yml config -q` → OK
- `docker compose -f docker-compose-development.yml config -q` → OK
- Rendered config confirms `image: docker.io/postgres:18` with mount target `/var/lib/postgresql`.

No application code touched; config-only change.

## Note for existing local setups
Users with a stale `pg_data` volume from an older Postgres major must reset it once: `docker compose down -v` then `docker compose up`.

### DIFF
--- a/docker-compose-development.yml
+++ b/docker-compose-development.yml
@@ -14,7 +14,7 @@ networks:
 services:
   ### Dependencies
   pg:
-    image: docker.io/postgres:latest
+    image: docker.io/postgres:18
     ports:
       - "5432:5432"
     networks:
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -25,7 +25,7 @@ services:
       - router_net
 
   pg:
-    image: docker.io/postgres:latest
+    image: docker.io/postgres:18
     ports:
       - "5432:5432"
     networks:


XyneSpaces:
## PR Review Summary

### Changes Overview
Pins PostgreSQL Docker image from `latest` to `18` in both `docker-compose.yml` and `docker-compose-development.yml`.

### Findings

**No blocking issues.**

This is a best-practice infrastructure fix that prevents unexpected breakages from floating `latest` tag updates.

**Considerations:**
- Ensure version 18 is the intended target (PostgreSQL 18 is a future version, currently latest stable is 16/17). Verify this aligns with the production database version.
- Document the chosen version in setup docs if not already present

XyneSpaces:
PR #13073 - **Reviewed by XyneSpaces Bot**

This PR pins the postgres Docker image from `latest` to version `18` in both docker-compose files. 

:bulb: **Finding**: This is a safe change that prevents floating-tag breakage. The `postgres:18` tag is a valid, stable version.

**Verdict**: No blocking issues. The change improves reproducibility of the development environment.

XyneSpaces:
:bulb: PostgreSQL 18 data directory change may break existing volumes

Pinning to `postgres:18` is correct (better than `latest`), but note that PostgreSQL 18 changed the data directory structure from `/var/lib/postgresql/data` to `/var/lib/postgresql/18/docker`.

**Impact:** Users with existing volumes from PG 17 will encounter `database files are incompatible with server` errors.

**Suggestion:** Add a changelog note about this breaking change for users upgrading from previous `postgres:latest` (which may have been PG 17).

Consider using `postgres:18.4` for reproducible builds rather than the floating `18` tag.

XyneSpaces:
:mag: **Postgres version pin should consider compatibility**

Pinning postgres to `:18` is good practice to avoid `latest` tag drift. However:

**Action**: Verify:
1. Existing development databases are compatible with Postgres 18 or document migration steps
2. The version 18 is the current stable release (not a beta/RC)
3. Consider using a more specific tag like `18.0` or `18-alpine` for reproducibility

XyneSpaces:
:white_check_mark: Docker image pinning looks correct.

Pinning postgres to major version 18 (instead of latest) is a best practice that prevents unexpected breaking changes when PostgreSQL releases new major versions. This ensures reproducible deployments and avoids potential data directory incompatibility issues on version upgrades.

XyneSpaces:
:rotating_light: **[blocking]** Invalid PostgreSQL Docker image tag

The tag `postgres:18` does not exist on Docker Hub. PostgreSQL versions use numeric tags like `15`, `16`, etc. The current latest stable is PostgreSQL 16.

```suggestion
-    image: docker.io/postgres:18
+    image: docker.io/postgres:16
```

Alternatively, if you want to pin to a specific patch version for reproducibility:
```suggestion
-    image: docker.io/postgres:18
+    image: docker.io/postgres:16.3
```

See available tags at: https://hub.docker.com/_/postgres/tags

SanchithHegde (approved):

PR #13081, opened 2026-06-30
feat(user): POST /user/launch_sage — federated sage session mint

### LINKED ISSUE #13082: [FEATURE] POST /user/launch_trace — federated Trace session mint from Control Center
## Feature description

Add a new authenticated route **`POST /user/launch_trace`** on Hyperswitch BE that mints a single-use federated session token for the downstream AI observability service (Trace), scoped to the calling Control Center user's currently-active merchant.

Today, when a CC operator wants to investigate a payment incident in Trace, they have to leave CC, separately log into the downstream service (internal VPN only), and re-enter the merchant context manually — there's no way to launch Trace bound to the merchant they were viewing. This change closes the gap: a "Launch Trace" button in CC can call this endpoint, get a one-time handoff URL, and redirect the browser to Trace already scoped to the right merchant.

## Why this matters

- **Operator workflow**: removes a multi-step manual context switch when investigating live merchant incidents.
- **Strict per-merchant isolation**: the user can only mint a session for a merchant they're already authorized for in CC. The token can't be retargeted by hand.
- **No oracle for tenant existence**: the downstream side of the handshake returns the same opaque 404 for unknown user / no merchant access / disabled user — an attacker holding the infra key can't enumerate which CC users exist.
- **Operational killswitch**: the federated session lives in a Redis on the downstream side with a 30-min sliding TTL, and the SOC team has `DELETE /api/sessions/by-user` to terminate any live session immediately.

## Cross-service architecture (design context)

Detailed design + threat model lives in a separate internal tracker. The downstream side is feature-complete — all phases merged, deployed to sandbox, and end-to-end tested locally.

The Hyperswitch BE side is just one new route:

```
CC browser ──POST /user/launch_trace──▶ HS BE
                                          │
                                          │ Authorization: Bearer TRACE_INFRA_KEY
                                          ▼
                                        Downstream POST /api/sage/session
                                          │
                                          │ mints handoff_url (60s TTL)
                                          ▼
HS BE returns { handoff_url, expires_at } ──▶ CC ──▶ browser navigates ──▶ Downstream /handoff?t=…
```

## Acceptance criteria

- [ ] New route `POST /user/launch_trace` registered under existing `/user` scope.
- [ ] Auth: `DashboardNoPermissionAuth` (any valid CC AuthToken is accepted; merchant binding comes from the JWT, never the body).
- [ ] New config block `[trace_integration]` with `enabled`, `sage_base_url`, `infra_key` (the last as `Secret<String>`, KMS-decrypted on boot via `SecretsHandler`).
- [ ] Boot validator that refuses to start when `enabled = true` AND `infra_key` is empty — fail-closed, named `TRACE_INFRA_KEY` in the error message.
- [ ] New error enum `LaunchTraceErrors` with deliberate status mapping:
  - `FeatureDisabled` → 404 (no oracle — matches the downstream service's behavior when the flag is off on either side).
  - `Forbidden` → 401 (HS BE rejects internal-role users locally, before any downstream call).
  - `UpstreamUnavailable` → 500 (downstream unreachable; Redis down on their side, etc.).
  - `UpstreamCredentialsRejected` → 502/500 (downstream 401 — almost always shared-key drift; pages SRE).
- [ ] Outbound role mapping: HS BE `EntityType` → `viewer` (Merchant/Profile) or `operator` (Tenant/Organization). The downstream service caps further but HS BE should pass the right starting value.
- [ ] Plaintext infra key is never logged. The `Secret<String>` wrapper redacts on `Debug`; the only `.peek()` is inside the `format!("Bearer …")` in the outbound request.
- [ ] Handoff URL logging strips the `?t=<token>` query (it's a 60s bearer; must not survive log aggregation).

## Out of scope

- CC frontend changes (separate ticket in CC repo).
- Downstream-side changes (already shipped).
- Production env config / KMS provisioning (sandbox first; prod env is a separate deploy).

## Related

- Implementation PR: #13081

### LINKED ISSUE #13125: [FEATURE] Send full user scope (merchants + profiles) on /user/launch_trace mint request
## Feature description

Extend the outbound mint request from `POST /user/launch_trace` to send the caller's **entire authorised scope** (merchants + profiles) instead of just the JWT's currently-active merchant.

Today, `launch_trace.rs` sends `{user_id, merchant_id, profile_id, ...}` — one merchant only. This ties the resulting federated downstream session to that one merchant, so a CC user who is `merchant_admin` on M1/M2/M3 (and `profile_admin` on selected profiles under M4) has to go back to CC → merchant switcher → relaunch for each merchant they want to investigate.

This change is the HS BE half of a wider design owned by the downstream AI observability service, which extends the downstream session to carry a structured scope.

## Reference architecture

The exact scope enumeration this needs already exists at [`crates/router/src/analytics.rs:3982-4020`](https://github.com/juspay/hyperswitch/blob/main/crates/router/src/analytics.rs#L3982) — `get_global_search_results`. It walks the user's `user_roles` rows and emits a `Vec<AuthInfo>` where each entry is either `MerchantLevel { merchant_ids }` (wildcard-over-profiles, i.e. merchant_admin) or `ProfileLevel { merchant_id, profile_ids }` (specific list, i.e. profile_admin). Reuse that transformation.

## Proposed shape

Extend the mint request body to include the scope Vec, plus keep the current `merchant_id` / `profile_id` as the initial-focus fields:

```rust
pub struct MintSessionRequest {
    pub user_id: String,
    pub launch_merchant_id: String,      // was: merchant_id — the JWT's current merchant (initial focus)
    pub launch_profile_id: String,       // was: profile_id — the JWT's current profile
    pub org_id: String,
    pub role: String,
    pub source: String,
    pub subject_jti: Option<String>,
    pub scope: Vec<ScopeEntry>,          // NEW — full authorised scope
}

pub struct ScopeEntry {
    pub merchant_id: String,
    pub profile_ids: ScopeProfiles,      // "*" or a Vec<String>
}

pub enum ScopeProfiles {
    All,                    // serialises to "*"
    Some(Vec<String>),      // serialises to ["P1","P2",...]
}
```

## Handler logic

Inside `launch_trace` at `crates/router/src/core/user/launch_trace.rs`, after auth:

1. Call `state.global_store.list_user_roles_by_user_id` with `(user_id, tenant_id, org_id=Some(...), merchant_id=None, profile_id=None, status=Some(Active))`. Same call site as `analytics.rs:3919-3955`.
2. For each row, use `user_role.get_entity_id_and_type()` (`crates/diesel_models/src/user_role.rs:30-47`) to branch on entity_type:
   - `EntityType::Tenant | Organization` — emit one `ScopeEntry { merchant_id: Mn, profile_ids: All }` for every merchant reachable from the org.
   - `EntityType::Merchant` — emit `ScopeEntry { merchant_id, profile_ids: All }`.
   - `EntityType::Profile` — emit `ScopeEntry { merchant_id, profile_ids: Some(vec![profile_id]) }`.
3. Collapse duplicates: multiple profile-level rows for the same merchant merge into a single `ScopeEntry` with a combined `Vec<String>`.
4. Send as the `scope` field on the outbound mint request.

## Related

- Parent design: tracked in the downstream AI observability service (private tracker).
- Current `launch_trace` PR: #13081

prajjwalkumar17:
Closes #13082.
Closes #13125.

Implements the HS BE side of the multi-merchant + multi-profile federation design for a downstream AI observability service ("sage"). This PR covers the scope-enumeration + mint-request-body changes on the HS side.

## Type of Change

- [x] New feature
- [ ] Bug fix
- [ ] Refactoring
- [ ] Documentation
- [ ] CI/Build

## Description

Adds `POST /user/launch_sage` — a new dashboard-authenticated endpoint that proxies a federated session mint request on behalf of a Control Center user who clicked the sage launcher.

**Feature-flagged**, off by default everywhere. The boot validator refuses to start the process if the flag is on with no shared infra key configured (fail-closed for sandbox typos).

This is the Hyperswitch BE half of the federation work. The downstream service has its own SessionStore + mint endpoint + federated cookie resolver + operator runbook already shipped.

**Marked Draft** so the dashboard + framework teams can sign off on four policy decisions before flag-flip — see "Open policy decisions" below. The scaffolding is mechanical and ready; the four risk points are tagged with `// TODO(launch_sage)` in the handler.

### 1. Auth strategy — `DashboardNoPermissionAuth` vs `JWTAuth { permission: ... }`

Currently: `DashboardNoPermissionAuth { allow_connected: true, allow_platform: true }` — same shape as `/list/org` and `/signout`. Federation is an identity-attestation step, not a payment/account permission, and the downstream service performs the authoritative merchant-access gate against its own allow-list (defense-in-depth re-check against the user's federated allow-list).

**Alternative:** force a specific `Permission` (e.g., `Permission::MerchantPaymentRead`). Pro: explicit gate at HS layer. Con: gates legitimate viewer-role users out, since "view your own merchant" isn't naturally any existing permission.

**Recommend:** keep `DashboardNoPermissionAuth`; trust the downstream allowlist as the authoritative barrier.

### 2. Upstream 401 → HS 502 (NOT 401 to client)

Currently: a `401` from the downstream service means the shared infra key is wrong (rotation drift). The CC user's JWT was fine. Bubbling 401 to CC would log the dashboard user out for an infra bug on the downstream side.

The handler maps upstream 401 → `LaunchSageErrors::UpstreamCredentialsRejected` → 502-equivalent + paging-grade error log so SRE sees rotation drift immediately.

**Confirm:** is this the right inversion?

### 3. `EntityType` → sage role mapping

Currently: `Tenant | Organization → "operator"`, `Merchant | Profile → "viewer"`. Internal roles are rejected outright. Federation caps at `operator` per the federated role-cap design.

**Confirm:** is this the right floor for tenant-/org-scoped users? A tenant admin federating in is still bound to a single merchant inside the downstream UI — broader-scoped users still operate one merchant at a time.

### 4. 404 on flag-off (no oracle)

Currently: when `sage_integration.enabled=false`, the handler returns `LaunchSageErrors::FeatureDisabled` → 404 (not 503 "service unavailable"). Mirrors the downstream service's own no-oracle policy — neither side reveals whether federation is enabled in a given env without proof of identity.

**Confirm:** is this consistent with HS's broader feature-flag conventions, or should it be 503?

## Verification

- [x] `cargo check --package router --features olap,v1` — clean
- [x] `cargo fmt --all -- --check` — clean
- [x] `cargo clippy --package router --features olap,v1 -- -D warnings` — clean
- [ ] Integration test against a real downstream instance — pending end-to-end after the handoff redeem + cookie issuance ship on the downstream side

## Reviewer guidance

- **@juspay/hyperswitch-dashboard** — please review the auth strategy choice (#1) and the role mapping (#3).
- **@juspay/hyperswitch-framework** — please review the settings + secrets-transformer wiring and the error mapping (#2, #4).

## Out of scope

- Integration tests against a live downstream instance (depends on the handoff redeem endpoint which sets the cookie the browser carries on next requests). Will land in a follow-up PR.
- Per-call audit row in HS BE (the downstream service owns its own audit table; HS BE's audit lives in the existing API event logs).
- Network policy restricting downstream ingress to HS BE pods — owned by the infra repo.
- CC frontend launcher button — owned by the CC team.

---

### Wire shape of a scope Vec

```
"scope": [
  {
    "entity_type": "merchant",
    "entity_id": "M1",
    "path": {"tenant_id": "T1", "org_id": "org_abc"}
  },
  {
    "entity_type": "profile",
    "entity_id": "P1",
    "path": {"tenant_id": "T1", "org_id": "org_abc", "merchant_id": "M4"}
  },
  {
    "entity_type": "profile",
    "entity_id": "P2",
    "path": {"tenant_id": "T1", "org_id": "org_abc", "merchant_id": "M4"}
  },
  {
    "entity_type": "profile",
    "entity_id": "P3",
    "path": {"tenant_id": "T1", "org_id": "org_abc", "merchant_id": "M4"}
  }
]
```

---

### Extensibility examples (no schema change needed)

Per-connector authorization — user allowed to see only Adyen data under M1:
`{ "entity_type": "connector", "entity_id": "adyen", "path": {"merchant_id": "M1"} }`

Currency-restricted grant — user can only query INR payments on M2:
`{ "entity_type": "merchant", "entity_id": "M2", "path": {...}, "constraints": {"currency_allowed": ["INR"]} }`

Time-bounded temporary access — vendor gets M3 access for 24h:
`{ "entity_type": "merchant", "entity_id": "M3", "path": {...}, "constraints": {"valid_until": "2026-07-02T18:00:00Z"} }`

Region-level grant — user manages EU merchants:
`{ "entity_type": "region", "entity_id": "EU", "path": {"tenant_id": "T1"} }`


jarnura (dismissed):

crates/router/src/core/user/launch_sage.rs
```diff
@@ -0,0 +1,214 @@
...
+    entity_id: String,
+    entity_type: EntityType,
+) -> launch_sage_api::ScopeEntry {
+    let mut path: HashMap<String, String> = HashMap::new();
+    path.insert(
+        "tenant_id".to_owned(),
+        user_role.tenant_id.get_string_repr().to_owned(),
+    );
+    let (type_name, include_org, include_merchant) = match entity_type {
+        EntityType::Tenant => ("tenant", false, false),
```
ThisIsMani: I think you can use to_string() on entity_type.

crates/router/src/core/user/launch_sage.rs
```diff
@@ -0,0 +1,214 @@
...
+        entity_id,
+        path,
+        constraints: HashMap::new(),
+    }
+}
+
+fn map_entity_type_to_role(entity: EntityType) -> &'static str {
+    match entity {
+        EntityType::Tenant | EntityType::Organization => "operator",
+        EntityType::Merchant | EntityType::Profile => "viewer",
```
ThisIsMani: Should we create an enum for sage roles?

crates/router/src/routes/user.rs
```diff
@@ -39,6 +39,27 @@ pub async fn get_active_user_details(state: web::Data<AppState>, req: HttpReques
...
+    Box::pin(api::server_wrap(
+        flow,
+        state,
+        &http_req,
+        (),
+        |state, user: auth::UserFromToken, _, _| user_core::launch_sage::launch_sage(state, user),
+        &auth::DashboardNoPermissionAuth {
+            allow_connected: true,
+            allow_platform: true,
+        },
```
ThisIsMani, line 57: Are we sure that we want to enable access to everyone? You should probably speak with @muditbhutani about this...

ThisIsMani (commented):
Mostly looks good to me. Minor concerns.

crates/router/src/core/user/launch_sage.rs
prajjwalkumar17: Done in [2ba2181f52](https://github.com/juspay/hyperswitch/pull/13081/commits/2ba2181f52) — `EntityType` derives `strum::Display` with `serialize_all = "snake_case"`, so the four variants already serialize byte-for-byte to `"tenant"/"organization"/"merchant"/"profile"`. Kept the outer match itself (it also produces the `include_org` / `include_merchant` booleans and its exhaustiveness catches new `EntityType` variants), dropped just the string arm and used `entity_type.to_string()` at the `ScopeEntry` construction.

crates/router/src/core/user/launch_sage.rs
prajjwalkumar17: Done in [2ba2181f52](https://github.com/juspay/hyperswitch/pull/13081/commits/2ba2181f52) — added a private `SageRole { Operator, Viewer }` enum colocated in `launch_sage.rs` with `strum::Display` + `serialize_all = "snake_case"` so the wire strings stay lockstep with hypersage's [`Role`](https://github.com/juspay/hypersage/blob/main/src/web/auth.py) enum. `map_entity_type_to_role` now returns `SageRole` and the string is emitted via `.to_string()` at the mint boundary. Kept it module-private (not in `common_enums`) since this is a hyperswitch→sage wire vocabulary rather than a hyperswitch domain concept, and the `role` field on `SageSessionRequest` stays `String` so the API-models crate isn't coupled to sage-side vocabulary.

ThisIsMani (approved):

jarnura (approved):

PR #13091, opened 2026-06-30
feat(redis_interface): instrument Redis commands with external service call event emission

### LINKED ISSUE #13090: feat(observability): capture per-command Redis call latency and errors as observable events
## Problem

Redis is the most-called backend in hyperswitch, yet its contribution to request latency is invisible. Every `GET`, `SET`, `HSET`, `XADD`, and related command appears as undifferentiated time in request spans — there is no way to distinguish a slow `XREADGROUP` from a fast `GET` or to see per-merchant Redis error rates.

## Requirement

Instrument all Redis commands in both the `fred` and `redis_rs` backends so that each command execution emits an `ExternalServiceCall` event carrying:
- command name (e.g. `SET`, `HGET`)
- latency in milliseconds
- success/failure
- the current request ID (for joining to API-level events)

Events should be emitted only within a request context (request ID is present); background workers produce no events since they cannot be correlated to a specific API request.

## Acceptance Criteria

- Every public async command method in `crates/redis_interface/src/module/fred/commands.rs` and `crates/redis_interface/src/module/redis_rs/commands.rs` wraps its body in an `observe()` call.
- A CI script enforces that new commands cannot be added without `observe()` coverage.
- Existing Redis unit tests continue to pass.
- No event is emitted when the emitter is a no-op (fast-path guard prevents extra allocations on the hot path).

spritianeja03:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

Instruments Redis with `ExternalServiceCall` event emission so Redis latency and errors flow through the same observability pipeline as key-manager calls, and can be correlated with the originating API request in ClickHouse.

Events are emitted **once per network roundtrip**, not once per logical command method.

### Key changes

1. **`crates/common_utils/src/request_context.rs`** (new)

   `RequestContext` — a one-method trait (`request_id() -> Option<&str>`) that lets infrastructure clients read the current request's ID without depending on router types. Named to avoid collision with `error_stack::Context`. Implemented by `RouterStore`, `KVRouterStore`, and drainer's `Store`.

2. **`RedisConnectionPool` carries the emitter** (both `fred` and `redis_rs` backends)

   New `event_emitter: Arc<dyn ExternalServiceEventEmitter>` field, supplied once at construction and propagated by `clone()`. `new_without_event_emitter()` is the `NoOpEventEmitter` shortcut for background workers and tests.

3. **`RedisConnectionWithContext`** (new) — a cheap, cloneable request-scoped handle

   ```rust
   pub struct RedisConnectionWithContext {
       pub redis_conn: Arc<RedisConnectionPool>,  // shared pool
       pub request_id: Option<String>,            // this request's ID
   }
   ```

   **All Redis command methods moved from `RedisConnectionPool` onto this type** (55 `track_redis_call` sites in `fred`, 49 in `redis_rs`). A Redis command can no longer be issued without first stating whether a request context exists, which makes instrumentation coverage complete by construction rather than by review discipline. The pool retains construction, `clone`/`add_prefix`, pub/sub and health.

4. **Emission folded into `track_redis_call`** (`crates/redis_interface/src/metrics.rs`)

   It already wrapped every raw client future for the OTel latency histogram. It now also takes `request_id` + emitter and emits one `ExternalServiceCall` per roundtrip. A new `RedisCallStatus` trait derives `success`/`status_code` — blanket impl over `Result`, plus a `Vec<String>` impl for `fred`'s streaming `SCAN`/`HSCAN`, which logs and drops page-level errors before this layer sees them.

5. **`RedisConnInterface::get_redis_conn()` returns the handle** instead of `Arc<RedisConnectionPool>`

   `RouterStore` builds it from its own `RequestContext`. Every caller (`kafka_store`, auth blacklist, authorization, card testing guard, OIDC, user utils, cache, scheduler) is a type change only — no call-site logic changed. `SessionState::add_request_id` now also propagates into `global_store`, which backs those auth/user Redis calls.

6. **Production wiring** — `app.rs` selects `EventsHandler` or `NoOpEventEmitter` from the existing `conf.events.emit_external_service_call_events` flag (already present on main, defaults to `false`), then threads it through `get_cache_store` → `RouterStore::cache_store` → `RedisConnectionPool::new`. **No config files are changed by this PR**; the feature is off by default.

### Emitted event

| Field | Value |
|---|---|
| `service_name` | `"redis"` |
| `endpoint` | The `RedisOperation` variant, e.g. `GetKey` — **not** the Redis key, which would carry PII and unbounded cardinality |
| `method` | `"Redis"` (protocol indicator, parity with `"gRPC"`) |
| `status_code` | `200` / `500` — not `0`, so error-rate dashboards filtering `>= 500` see Redis failures |
| `latency_ms` | This roundtrip only |
| `request_id` | From the request-scoped handle |

### Why per-roundtrip

Emitting once per public method under-counts and mis-attributes latency whenever a method performs more than one roundtrip:

| Method | Roundtrips | One event per method | Per roundtrip (this PR) |
|---|---|---|---|
| `get_key` (multitenancy fallback miss) | 2 | 1 event, latency = sum | 2 events, each real |
| `set_hash_field` | 2 (HSET, EXPIRE) | 1 event, latency = sum | 2 events, independently timed |
| `get_multiple_keys` (cluster) | N parallel GETs | 1 event labelled `MGET` — no MGET was sent | N events labelled `GetKey` |
| `scan` / `hscan` | 1–N cursor pages | 1 event, latency = full walk | 1 event per page |

It also makes call-count anomalies visible ("this request made 40 Redis roundtrips, not 4") instead of hiding them inside a single event.

### Background work emits nothing

When `request_id` is absent — drainer, scheduler, spawned tasks — no event is emitted. The downstream correlator joins on `request_id`; request-less rows would pile up under one un-joinable key. Drainer's pool is constructed with `NoOpEventEmitter` for the same reason. These paths are still plumbed through `RedisConnectionWithContext` so they pick up emission for free if a bypass rule is added later.

## Additional Changes

## Motivation and Context

Redis is called on virtually every payment and storage operation, yet its latency contribution is invisible — it appears as undifferentiated time in request spans. This PR makes each Redis roundtrip emit a discrete event carrying the command, latency, success status and request ID. Events flow through the same `ExternalServiceCall` Kafka topic introduced for key-manager instrumentation (PR #11834), landing in ClickHouse where per-command latency histograms, error rates and per-merchant Redis usage can be queried and joined against API events.

Closes #13090

## How did you test it

Redis Events landing in the correct kafka topic
Screenshot 2026-07-28 at 7 16 07 PM

Redis data in the observability_events_flat clickhouse table
Screenshot 2026-07-28 at 7 18 01 PM

Redis data in the external_service_calls array in the observability_events clickhouse table
Screenshot 2026-07-28 at 7 20 09 PM

Compile verification across the feature combinations that gate the touched code — a default build exercises only a subset, since much of the Redis command layer sits behind `deja` / `multitenancy_fallback` / backend cfgs:

- `cargo check -p redis_interface` (default, `redis-rs`)
- `cargo check -p redis_interface --no-default-features --features fred`
- `cargo check -p redis_interface --features deja`
- `cargo check -p redis_interface --no-default-features --features fred,deja`
- `cargo check -p redis_interface --features multitenancy_fallback`
- `cargo check -p storage_impl --features v1,payouts`
- `cargo check -p router --features v1`
- `cargo check -p drainer --features v1`

Manual verification of emission (flag defaults to `false`, so this is opt-in):

1. Set `events.emit_external_service_call_events = true`.
2. Issue any payment or admin API call.
3. Confirm `ExternalServiceCall` records with `service_name = "redis"` on the events topic, one per roundtrip, sharing the API call's `request_id`.
4. Confirm a drainer/scheduler cycle produces no Redis events.

## Checklist


crates/redis_interface/src/metrics.rs
```diff
@@ -75,5 +115,68 @@ where
...
+            pool.event_emitter
+                .emit_external_service_call(ExternalServiceCall {
+                    service_name: "redis".to_string(),
+                    endpoint: format!("{operation:?}"),
+                    method: "Redis".to_string(),
+                    request_id,
+                    status_code: if success { 200 } else { 500 },
+                    success,
+                    latency_ms: time_elapsed.as_millis(),
+                    created_at_timestamp: OffsetDateTime::now_utc().unix_timestamp_nanos(),
```
ThisIsMani: Can't we use common_utils::date_time::now_unix_timestamp()?

crates/redis_interface/src/types.rs
```diff
@@ -351,10 +351,12 @@ impl std::error::Error for StreamTrimThresholdError {}
 pub struct RedisKey(String);
 
 impl RedisKey {
+    #[cfg(any(feature = "fred", feature = "redis-rs"))]
```
ThisIsMani: Might not be required.

crates/redis_interface/src/module/fred/commands.rs
```diff
@@ -793,6 +832,7 @@ impl super::RedisConnectionPool {
         V: FromRedis + Unpin + Send + 'static,
     {
         match track_redis_call(
+            self,
```
ThisIsMani: Pass event emitter alone.

crates/drainer/src/services.rs
```diff
@@ -13,11 +14,23 @@ use crate::{
...
-    pub redis_conn: Arc<RedisConnectionPool>,
+    redis_conn: Arc<RedisConnectionPool>,
     pub config: StoreConfig,
     pub request_id: Option<String>,
 }
 
+impl Store {
+    pub fn get_redis_conn(&self) -> redis_interface::RedisConnectionWithContext {
+        redis_interface::RedisConnectionWithContext::new(self.redis_conn.clone(), self)
+    }
```
ThisIsMani: Check if there any already existing impl where this can be moved to.

crates/redis_interface/src/metrics.rs
```diff
@@ -42,6 +55,7 @@ pub(crate) enum RedisOperation {
     StreamGetLength,
     StreamReadEntries,
     StreamReadWithOptions,
+    StreamReadGroup,
```
ThisIsMani: We can remove this.

crates/storage_impl/src/connection.rs
```diff
@@ -15,7 +15,7 @@ pub type PgPooledConn = async_bb8_diesel::Connection<PgConnection>;
 pub async fn redis_connection(
     redis: &redis_interface::RedisSettings,
 ) -> redis_interface::RedisConnectionPool {
-    redis_interface::RedisConnectionPool::new(redis)
+    redis_interface::RedisConnectionPool::new_without_event_emitter(redis) //check why this has to be without event emitter
```
ThisIsMani: Remove this if not being used.

crates/router/src/connection.rs
```diff
@@ -9,15 +9,14 @@ pub type PgPool = bb8::Pool<async_bb8_diesel::ConnectionManager<PgConnection>>;
...
 /// Panics if failed to create a redis pool
 #[allow(clippy::expect_used)]
 pub async fn redis_connection(
     conf: &crate::configs::Settings,
 ) -> redis_interface::RedisConnectionPool {
-    redis_interface::RedisConnectionPool::new(&conf.redis)
+    redis_interface::RedisConnectionPool::new_without_event_emitter(&conf.redis)
         .await
         .expect("Failed to create Redis Connection Pool")
 }
```
ThisIsMani: This also can be removed if not being used.

ThisIsMani (dismissed):

jarnura (dismissed):

crates/redis_interface/src/module/fred.rs
```diff
@@ -309,9 +356,25 @@ impl RedisConnectionPool {
...
+            request_id: context.request_id().map(str::to_owned),
+        }
+    }
+
+    pub fn new_without_context(pool: Arc<RedisConnectionPool>) -> Self {
+        Self {
+            redis_conn: pool,
+            request_id: None,
+        }
+    }
```
ThisIsMani: NIT: Can we move this up near the struct definition.

crates/redis_interface/src/metrics.rs
```diff
@@ -55,11 +68,47 @@ pub(crate) enum RedisOperation {
...
+}
+
+/// Fred's streaming `SCAN`/`HSCAN` helpers collect successful pages into a
+/// `Vec<String>` after logging and dropping page-level errors. At this layer no
+/// error signal remains, so completion is treated as success for event emission.
+impl RedisCallStatus for Vec<String> {
+    fn is_success(&self) -> bool {
+        true
+    }
+}
```
SanchithHegde: Nit: Would be preferable if we used a newtype instead of implementing this on a general type such as `Vec<String>>`.

crates/redis_interface/src/metrics.rs
```diff
@@ -77,5 +126,19 @@ where
         REDIS_CALL_TIME.record(time_elapsed.as_secs_f64(), attributes);
     }
 
+    if let Some(request_id) = request_id {
+        let success = output.is_success();
+        event_emitter.emit_external_service_call(ExternalServiceCall {
+            service_name: "redis".to_string(),
+            endpoint: format!("{operation:?}"),
```
SanchithHegde, line 141: Nit, optional change: You can choose to derive `strum::Display` on the enum as well, if required.

crates/redis_interface/src/metrics.rs
```diff
@@ -77,5 +126,19 @@ where
...
+            method: "Redis".to_string(),
+            request_id: request_id.to_string(),
+            status_code: if success { 200 } else { 500 },
```
SanchithHegde, line 144: Nit: We seem to be using some HTTP-specific naming?

crates/storage_impl/src/lib.rs
```diff
@@ -117,7 +130,8 @@ where
                 db_conf,
                 tenant_config,
                 encryption_key,
-                Self::cache_store(&cache_conf, cache_error_signal).await?,
+                Self::cache_store(&cache_conf, cache_error_signal, Arc::new(NoOpEventEmitter))
```
SanchithHegde, line 133: Any specific reason to be hardcoding `NoOpEventEmitter` here?

SanchithHegde (commented):

crates/storage_impl/src/lib.rs
spritianeja03: `RouterStore::new()` isn't used anywhere today, so this is dead code in practice. The `NoOpEventEmitter` here is just a placeholder — cache_store() requires an event_emitter argument (it's not Option), and we don't have a real one available at this call site yet. We'll wire
  in the actual emitter in a follow-up PR (#13472), where we can pass a clone of the event_emitter already in scope instead of NoOpEventEmitter.

ThisIsMani (approved):

jarnura (approved):

SanchithHegde (approved):

PR #13124, opened 2026-07-01
fix(connector): [Cybersource] allow connector creation without metadata

deepanshu-iiitu:
## Type of Change

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

Creating a Cybersource connector via the API without a `connector_meta_data` field fails with `IR_06 - "The metadata is invalid"`, even though Cybersource metadata is entirely optional.

`CybersourceConnectorMetadataObject` has only optional fields and derives `Default`:

```rust
#[derive(Debug, Default, Serialize, Deserialize)]
pub struct CybersourceConnectorMetadataObject {
    pub disable_avs: Option<bool>,
    pub disable_cvn: Option<bool>,
}
```

However, its `TryFrom<&Option<SecretSerdeValue>>` impl routed through `to_connector_meta_from_secret`, which treats an absent (`None`) metadata as a `MissingRequiredField`. That error is then `change_context`'d to `InvalidConnectorConfig { config: "metadata" }` and surfaced to the caller as `IR_06 - "The metadata is invalid"`.

This PR makes the Cybersource metadata `TryFrom` return `Self::default()` (both fields `None`) when metadata is absent, while still validating strictly when a metadata object is supplied (a malformed object still errors). This makes "no metadata sent" behave identically to sending an empty `{}` object.

The change is scoped to Cybersource only — it is not pushed into the shared `to_connector_meta_from_secret` helper — because other connectors (e.g. Coingate, Calida) have genuinely mandatory metadata fields that must keep failing on `None`.

### Additional Changes

## Motivation and Context

Merchants integrating Cybersource without configuring AVS/CVN overrides were forced to send a placeholder `metadata: {}` in the create-connector request to work around `IR_06`. Since neither `disable_avs` nor `disable_cvn` is required, Hyperswitch should accept connector creation with no metadata at all.

## How did you test it?

Manually via the create Merchant Connector Account API against Cybersource:

1. `POST /account/{merchant_id}/connectors` with a valid Cybersource `connector_account_details` and no `metadata` field → previously returned `IR_06`, now succeeds (200) with the connector created.
2. Same request with `"metadata": {}` → succeeds (unchanged).
3. Same request with valid `"metadata": {"disable_avs": true}` → succeeds and the value is persisted (unchanged).
4. Same request with malformed `"metadata": {"disable_avs": "yes"}` → still rejected with `IR_06` (strict validation preserved).

## Checklist

### DIFF
--- a/crates/hyperswitch_connectors/src/connectors/cybersource/transformers.rs
+++ b/crates/hyperswitch_connectors/src/connectors/cybersource/transformers.rs
@@ -117,11 +117,15 @@ pub struct CybersourceConnectorMetadataObject {
 impl TryFrom<&Option<pii::SecretSerdeValue>> for CybersourceConnectorMetadataObject {
     type Error = error_stack::Report<errors::ConnectorError>;
     fn try_from(meta_data: &Option<pii::SecretSerdeValue>) -> Result<Self, Self::Error> {
-        let metadata = utils::to_connector_meta_from_secret::<Self>(meta_data.clone())
-            .change_context(errors::ConnectorError::InvalidConnectorConfig {
-                config: "metadata",
-            })?;
-        Ok(metadata)
+        // Cybersource metadata is optional (all fields are optional), so treat an absent
+        // metadata object as the default rather than a missing required field.
+        match meta_data {
+            Some(_) => utils::to_connector_meta_from_secret::<Self>(meta_data.clone())
+                .change_context(errors::ConnectorError::InvalidConnectorConfig {
+                    config: "metadata",
+                }),
+            None => Ok(Self::default()),
+        }
     }
 }
 


AkshayaFoiger (approved):

Nithin1506200 (approved):

XyneSpaces:
:bulb: **Pattern consistency suggestion**

The implementation correctly handles the `None` case by returning `Self::default()`. This is a good pattern for optional connector metadata. Consider adding a brief doc comment explaining that Cybersource metadata is intentionally optional:

```rust
/// Cybersource connector metadata is optional - all fields have sensible defaults.
/// When metadata is absent, we use default values rather than failing.
```

XyneSpaces:
## 📝 Review Summary

This PR makes Cybersource metadata optional - a sensible change.

### ✅ Positive observations:

1. **Correct approach**: Treats absent metadata as default instead of error
2. **Backwards compatible**: Existing integrations with metadata continue to work
3. **Clean implementation**: Simple match on `Option` instead of unwrap

### 💡 Suggestion:

Consider adding a debug log when metadata is absent to help with troubleshooting integration issues:

```rust
None => {
    router_env::logger::debug!("Cybersource metadata not provided, using defaults");
    Ok(Self::default())
}
```

No blocking issues - this is a safe change that improves integration flexibility.

---
*This is an automated review.*

XyneSpaces:
## Review Findings for PR #13124

### Summary
PR allows Cybersource connector creation without metadata by treating absent metadata as default instead of error.

### Analysis
**File: `crates/hyperswitch_connectors/src/connectors/cybersource/transformers.rs`**

The change modifies `CybersourceConnectorMetadataObject`'s `TryFrom` impl to:
- Return `Self::default()` when `meta_data` is `None`
- Still validate with error when metadata is present but invalid

```rust
match meta_data {
    Some(_) => utils::to_connector_meta_from_secret::<Self>(meta_data.clone())
        .change_context(...),
    None => Ok(Self::default()),
}
```

**Potential Concern:**
⚠️ Verify that `CybersourceConnectorMetadataObject::default()` produces sensible defaults for all fields. If any field lacks a meaningful default, connector operations may fail downstream with obscure errors.

**Suggestion:** Consider adding a comment documenting what `Default` returns and which flows rely on these defaults.

PR #13150, opened 2026-07-03
fix(payouts): route merchant-authentication (access-token) calls through the payout connector

gopikrishna000:
## Summary

During the merchant-authentication (access-token) flow, the access-token gateway unconditionally used the payment gRPC header builder (`x-connector`) and a payment reference — even when the caller was a **payout**. For payout access-token calls this selected the payment connector variant and carried a placeholder payment reference instead of the real payout reference, even though `router_data.payout_id` is available on the access-token `RouterData` (preserved through `router_data_type_conversion`).

## What this does

Makes the access-token path payout-aware:

- **`external_services`** — adds `create_access_token_for_payouts`, a sibling of `create_access_token` that uses `build_unified_connector_service_grpc_headers_for_payouts` (`x-payout-connector`).
- **gateway** (`access_token_gateway.rs`) — when `router_data.payout_id` is set, builds the payout reference (`UcsReferenceId::Payout`) and resource id (`UcsResourceId::PayoutAttempt`) and dispatches to the payout sibling, following the existing `payouts/gateway/*` pattern. The payment path and the existing `payout_transfer` / `payout_get` paths are untouched.

## Result

Payout access-token calls now select the payout connector variant, run its auth implementation, and carry the real payout reference.

## Companion change

Pairs with the connector-side implementation of the Itaú payout merchant-auth flow: juspay/hyperswitch-prism#1806

### DIFF
--- a/crates/external_services/src/grpc_client/unified_connector_service.rs
+++ b/crates/external_services/src/grpc_client/unified_connector_service.rs
@@ -1047,11 +1047,16 @@ impl UnifiedConnectorServiceClient {
     }
 
     /// Performs Create Access Token Granular
+    ///
+    /// When `is_payout` is set, the payout connector header (`x-payout-connector`) is
+    /// attached so the merchant-authentication flow resolves the payout connector
+    /// variant; otherwise the payment connector header (`x-connector`) is used.
     pub async fn create_access_token(
         &self,
         create_access_token_request: payments_grpc::MerchantAuthenticationServiceCreateServerAuthenticationTokenRequest,
         connector_auth_metadata: ConnectorAuthMetadata,
         grpc_headers: GrpcHeadersUcs,
+        is_payout: bool,
     ) -> UnifiedConnectorServiceResult<
         tonic::Response<
             payments_grpc::MerchantAuthenticationServiceCreateServerAuthenticationTokenResponse,
@@ -1059,8 +1064,14 @@ impl UnifiedConnectorServiceClient {
     > {
         let mut request = tonic::Request::new(create_access_token_request);
         let connector_name = connector_auth_metadata.connector_name.clone();
-        let metadata =
-            build_unified_connector_service_grpc_headers(connector_auth_metadata, grpc_headers)?;
+        let metadata = if is_payout {
+            build_unified_connector_service_grpc_headers_for_payouts(
+                connector_auth_metadata,
+                grpc_headers,
+            )?
+        } else {
+            build_unified_connector_service_grpc_headers(connector_auth_metadata, grpc_headers)?
+        };
         *request.metadata_mut() = metadata;
 
         self.merchant_authentication_service_client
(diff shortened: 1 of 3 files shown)


crates/external_services/src/grpc_client/unified_connector_service.rs
```diff
@@ -1124,6 +1083,49 @@ impl UnifiedConnectorServiceClient {
...
+            })
+            .inspect_err(|error| {
+                logger::error!(
+                    grpc_error=?error,
+                    method="create_server_authentication_token_for_payouts",
+                    connector_name=?connector_name,
+                    "UCS create server authentication token gRPC call failed"
+                )
+            })
+    }
```
Shubhodip900: the only difference between `create_access_token_for_payouts` and `create_access_token` is the metadata creation. so instead of duplicating the whole logic in `create_access_token` we can pass a field indicating payments or payouts, based on that we can create the metadata

XyneSpaces:
:warning: **Missing `#[instrument]` on async hot-path functions**

Both `create_access_token_for_payouts` in `unified_connector_service.rs` (~1086-1128) and the async closure in `access_token_gateway.rs` (~131-175) lack `#[instrument(skip_all, fields(...))]`. These are on the payment/payout gateway path and require tracing instrumentation.

Add `#[instrument(skip_all)]` with relevant fields like `connector_name`, `payout_id`/`payment_id` to maintain observability parity with sibling methods.

XyneSpaces:
## Review Findings

### [should-fix] Code duplication in access token methods
The new `create_access_token_for_payouts` method in `UnifiedConnectorServiceClient` duplicates most of the logic from `create_access_token`. The only difference is the metadata creation via `build_unified_connector_service_grpc_headers_for_payouts`. Consider refactoring to share the common logic.

### [should-fix] `is_payout` detection logic
The detection of payout vs payment flows relies on `router_data.payout_id.is_some()`. Verify this is the canonical way to distinguish these flows and document any edge cases where a payout might not have a `payout_id` at this stage.

crates/external_services/src/grpc_client/unified_connector_service.rs
gopikrishna000: Done in d417197 — collapsed `create_access_token_for_payouts` into `create_access_token` with an `is_payout` flag that selects the payout vs payment header builder; both call sites updated (gateway passes `is_payout`, the payments/relay path passes `false`).

XyneSpaces:
:warning: Payout detection via `payout_id.is_some()` may be fragile

The logic at lines 102-170 detects payout flow using `router_data.payout_id.is_some()`. If `payout_id` is ever set for non-payout flows (e.g., in shared response types), this will incorrectly route through payout connector paths.

Consider using a more explicit flow discriminator or validating that `is_payout` is only set when the operation context explicitly indicates a payout flow, rather than relying on optional field presence.

XyneSpaces:
:warning: **Boolean flag overload in UCS client interface**

In `crates/external_services/src/grpc_client/unified_connector_service.rs` (~line 45):

The addition of `is_payout: bool` parameter to `create_access_token` is heading toward a boolean trap:

```rust
pub async fn create_access_token(
    &self,
    ...
    is_payout: bool,  // What does true mean again?
)
```

If more connector types are added (payouts, fraud_check, identity, etc.), this becomes an unmaintainable series of booleans.

**Fix**: Use an enum to make call sites self-documenting:

```rust
pub enum ConnectorType {
    Payment,
    Payout,
}

// Call site becomes:
create_access_token(..., ConnectorType::Payout)
```

XyneSpaces:
:bulb: **Consider consolidating payout detection logic**

The `is_payout` flag is derived from `router_data.payout_id.is_some()` in the gateway but passed separately through the client. Ensure these stay synchronized. A mismatch between `payout_id` presence and `is_payout` flag could route payout access-token calls through payment headers or vice versa.

XyneSpaces:
:warning: **Silent degradation on invalid attempt_id parsing**

The `inspect_err` + `.ok()` pattern silently converts parse failures to `None`:

```rust
.inspect_err(|err| logger::warn!(...))
.ok()
.map(...)
```

Invalid `attempt_id` values are logged but don't fail the request. The gRPC call proceeds with incomplete resource identification headers. This could cause subtle downstream issues where UCS receives malformed payout/payment references.

Consider failing fast with a proper error instead of continuing with `None`.

Shubhodip900 (dismissed):

crates/external_services/src/grpc_client/unified_connector_service.rs
```diff
@@ -1088,20 +1047,31 @@ impl UnifiedConnectorServiceClient {
...
+    ///
+    /// When `is_payout` is set, the payout connector header (`x-payout-connector`) is
+    /// attached so the merchant-authentication flow resolves the payout connector
+    /// variant; otherwise the payment connector header (`x-connector`) is used.
     pub async fn create_access_token(
         &self,
         create_access_token_request: payments_grpc::MerchantAuthenticationServiceCreateServerAuthenticationTokenRequest,
         connector_auth_metadata: ConnectorAuthMetadata,
         grpc_headers: GrpcHeadersUcs,
+        is_payout: bool,
```
AmitsinghTanwar007: you can use enum instead of bool

JeevaRamu0104 (approved):

hrithikesh026 (approved):

PR #13156, opened 2026-07-03
feat(connector): [Authorizedotnet] add network transaction id support

likhinbopanna:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description
- Add authorizedotnet to network_transaction_id_supported_connectors across all config deployments
- Handle CardDetailsForNetworkTransactionId and MandatePayment in AuthorizeDotNet transformer payment request
- Fix debug-build stack overflow by setting RUST_MIN_STACK at runtime before Tokio runtime creation
- Update cypress test configs for authorizedotnet NTID flow

### Additional Changes

## How did you test it?

image

## Checklist


XyneSpaces:
PLACEHOLDER_COMMENT_FOR_RETRIEVAL

XyneSpaces:
## Code Review Findings

### Overall Assessment
This PR adds Network Transaction ID support for Authorizedotnet and includes a debug build stack overflow fix. The changes are generally well-structured, but there are a few items to address:

### Findings

**1. [should-fix] Missing SAFETY comment for runtime environment variable manipulation**

In `crates/router/src/bin/router.rs`, the code sets `RUST_MIN_STACK` at runtime before spawning the Tokio runtime. While the extensive inline comment explains *why*, there should be a `// SAFETY:` comment explicitly noting that this is safe because:
- It's done before any threads are spawned
- The environment variable is only read by `std::thread` spawning, which hasn't happened yet

```rust
// SAFETY: Setting RUST_MIN_STACK before any thread creation is safe as std::thread 
// reads this env var only at spawn time. We set this before building the Tokio runtime.
#[cfg(debug_assertions)]
if std::env::var_os("RUST_MIN_STACK").is_none() {
    std::env::set_var("RUST_MIN_STACK", "67108864"); // 64 MiB
}
```

**2. [nit] `MandatePayment` returning `None` for payment details**

In `transformers.rs`, the `MandatePayment` variant now returns `None` for the payment field. Please confirm this is intentional - for mandate payments, Authorizedotnet may still require some payment reference. If this is correct, consider adding a comment explaining why `None` is appropriate for mandates.

**3. [nit] Import reordering in Utils.js**

The alphabetization of imports (`bitpay`/`billwerk`, `mifinity` placement) is good housekeeping, but ideally should be a separate commit or PR to keep the functional changes isolated.

### Approved with minor suggestions
The functional changes look correct. The NTID support implementation follows the existing patterns, and the stack overflow fix addresses a real debug-build issue. Please consider adding the SAFETY comment before merge.

XyneSpaces:
:warning: Unreachable arm uses generic error instead of explicit handling

In the `payment` match arm at line 949, `PaymentMethodData::MandatePayment` maps to `None`. Since this is an explicit match arm for a variant moved from the wildcard, this branch should either:
1. Return a specific `NotImplemented` error explaining that MandatePayment is not supported for this flow
2. Include a comment explaining why `None` is the correct handling

The current silent `None` makes debugging difficult when this variant is unexpectedly encountered.

```rust
PaymentMethodData::MandatePayment => {
    // Explicitly not supported for this Authorizedotnet flow
    return Err(IntegrationError::NotImplemented(
        "MandatePayment not supported for Authorizedotnet".to_string(),
        Default::default(),
    ))?;
}
```

XyneSpaces:
:warning: Stack size configuration moved to runtime but lacks upper bound check

The `RUST_MIN_STACK` is now set at runtime to 64 MiB when not present. However, there's no validation that an explicitly set value is reasonable. A maliciously high value could cause resource exhaustion.

Consider adding a maximum bound check:
```rust
if let Some(val) = std::env::var_os("RUST_MIN_STACK") {
    if let Ok(bytes) = val.to_string_lossy().parse::<usize>() {
        if bytes > 512 * 1024 * 1024 { // 512 MiB max
            eprintln!("Warning: RUST_MIN_STACK exceeds maximum allowed, using default");
            std::env::set_var("RUST_MIN_STACK", "67108864");
        }
    }
}
```

XyneSpaces:
## Automated Code Review Results

**Team:** Connector

### Findings Summary

**✅ Well-Rounded Feature Implementation**
This PR adds Network Transaction ID support for Authorizedotnet across config, transformer logic, and tests. Also includes a valuable stack overflow fix.

### Detailed Review

**1. Connector Configuration (6 config files)**
Adding `authorizedotnet` to the NTID connector list is straightforward. The consistency across environments (example, integration_test, production, sandbox, development, docker_compose) is good.

**2. Transformer Changes (`transformers.rs`)**
The new match arm for `CardDetailsForNetworkTransactionId` properly extracts card details:
```rust
PaymentMethodData::CardDetailsForNetworkTransactionId(ref card_details) => {
    Some(PaymentDetails::CreditCard(CreditCardDetails {
        card_number: (*card_details.card_number).clone(),
        expiration_date: card_details.get_expiry_date_as_yyyymm("-"),
        card_code: None,
    }))
}
```

**⚠️ Pattern Consistency Note:**
The change from single-expression match to statement-style match (with explicit `Some(...)`) increases verbosity slightly. Consider if a helper function could reduce duplication between the `Card` and `CardDetailsForNetworkTransactionId` arms since they share identical field extraction logic.

**3. Stack Overflow Fix (`router/build.rs` → `router.rs`)**
This is an excellent fix! Moving `RUST_MIN_STACK` from compile-time to runtime ensures:
- All worker threads get the increased stack size
- Debug builds won't mysteriously abort on deeply-nested async flows
- The 64 MiB size (vs previous 11 MiB) provides ample headroom

**🔍 Suggestion:**
Consider adding a startup log line confirming the stack size:
```rust
#[cfg(debug_assertions)] {
    let stack_size = std::env::var("RUST_MIN_STACK").unwrap_or_default();
    logger::debug!("Worker thread stack size: {}", stack_size);
}
```

**4. Test Updates**
Removing `authorizedotnet` from `EXCLUDE.CONNECTOR_AGNOSTIC_NTID` is correct now that NTID is supported.

**Overall Assessment:** Clean implementation with a valuable infrastructure fix. Approved.

XyneSpaces:
:warning: **Missing error context in MandatePayment branch**

In `crates/hyperswitch_connectors/src/connectors/authorizedotnet/transformers.rs` (~line 938):

```rust
PaymentMethodData::MandatePayment => None,
```

Returning `None` for `MandatePayment` without any context or error could silently create an invalid request. Authorize.Net requires transaction details for all payment types.

**Questions**:
1. Is returning `None` here actually valid for NTID-based subsequent transactions?
2. If this represents an MIT (Merchant Initiated Transaction), shouldn't there be a different request structure entirely?
3. Should this return an error indicating MandatePayment isn't directly supported?

**Fix**: Add a comment explaining when `None` is valid, or return an appropriate error if this path shouldn't be reachable.

```rust
PaymentMethodData::MandatePayment => {
    // MandatePayment data comes from a stored mandate;
    // NTID flows don't need card details in the payment field
    None
}
```

XyneSpaces:
:bulb: **Runtime stack size modification approach**

In `crates/router/src/bin/router.rs` (~line 42):

```rust
std::env::set_var("RUST_MIN_STACK", "67108864"); // 64 MiB
```

Setting stack size at runtime has limitations:
1. It only affects threads spawned AFTER this call
2. The main thread's stack is already allocated
3. This affects ALL subsequently spawned threads, not just deep recursion cases

**Fix**: Move this configuration to:
1. A wrapper script that sets the env var before process start, OR
2. Documentation for operators to set in their environment, OR  
3. Compile-time configuration via `.cargo/config.toml`

```toml
# .cargo/config.toml
[env]
RUST_MIN_STACK = "67108864"
```

XyneSpaces:
:bulb: **Stack size increase lacks consistency with build.rs**

The runtime stack size (64 MiB) is 6× larger than the previous build.rs setting (11 MiB). Additionally, `std::env::set_var()` is deprecated in Rust 1.86+. Consider using `TOKIO_WORKER_THREAD_STACK_SIZE` or Tokio's runtime builder with `thread_stack_size()` for cleaner configuration.

XyneSpaces:
:warning: **CVV never sent for CardDetailsForNetworkTransactionId**

The mapping explicitly sets `card_code: None` even if CVV is available in `card_details`:

```rust
Some(PaymentDetails::CreditCard(CreditCardDetails {
    card_number: ...,
    expiration_date: ...,
    card_code: None,  // CVC explicitly suppressed
}))
```

For NTID flows, some issuers require CVV on the first transaction. Please verify this suppression is intentional and document if NTID transactions may fail without CVV.

XyneSpaces:
:warning: **MandatePayment silently returns None without credential-on-file data**

The `MandatePayment` arm maps to `None` for the payment field:

```rust
PaymentMethodData::MandatePayment => None,
```

This sends empty payment data to Authorize.Net for MIT transactions. The processor will likely reject this with a generic "missing payment data" error. Either extract the mandate reference for `subsequent_auth_information` or return a clear `NotImplemented` error instead of silent failure.

crates/router/src/bin/router.rs
```diff
@@ -6,8 +6,31 @@ use router::{
     routes::metrics,
 };
 
-#[tokio::main]
-async fn main() -> ApplicationResult<()> {
+fn main() -> ApplicationResult<()> {
```
hrithikesh026: this change can be reverted. Not needed for this PR.

crates/router/build.rs
```diff
@@ -1,8 +1,10 @@
 fn main() {
-    // Set thread stack size to 11 MiB for debug builds
+    // NOTE: the worker-thread stack size for debug builds is set at *runtime*
```
hrithikesh026: this change can be reverted. Not needed for this PR.

Nithin1506200 (approved):

swangi-kumari (approved):

hrithikesh026 (approved):

Gnanasundari24 (approved):

PR #13172, opened 2026-07-06
fix(storage): align kv helpers with intent insert path and build drainer queries lazily

hrithikesh026:
## Type of Change

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [x] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

This PR refactors the v1 PaymentIntent Redis KV paths to use the shared `KVRouterStore` helpers and defers drainer query generation until the Redis KV write path needs it.

Changes include:

- uses `KVRouterStore::insert_resource`, `update_resource`, and `find_resource_by_id` for v1 PaymentIntent KV paths
- changes `InsertResourceParams` and `UpdateResourceParams` to carry a future returning `StorageResult<kv::SerializableQuery>`
- avoids eagerly generating drainer queries for Postgres-only storage paths
- moves entity-specific drainer query error context into every `generate_drainer_insert_query` / `generate_drainer_update_query` definition
- keeps shared KV helper call sites aligned across payment intent, payment attempt, payment method, and customer paths

### Additional Changes

## Motivation and Context

The v1 PaymentIntent Redis KV implementation had custom insert/update/find logic instead of using the common KV resource helpers. This made the path inconsistent with other resources and increased the chance of behavior drift between Postgres and Redis KV storage paths.

Deferring drainer query generation also avoids building SerializableQuery values when the resolved storage path is Postgres-only, while preserving entity-specific failure context at the query-generation boundary.

## How did you test it?

Verified with:

- `cargo fmt --all`
- `cargo check -p diesel_models --features v1`
- `cargo check -p diesel_models --features v2`
- `cargo check -p storage_impl --features v1`
- `cargo check -p storage_impl --features v2,hyperswitch_domain_models/revenue_recovery`
- `just clippy`
- `just clippy_v2`

Notes:

- `cargo fmt --all` emits existing rustfmt warnings for nightly-only formatting options on the stable toolchain.
- Clippy commands completed with existing dependency future-incompatibility notes.

## Checklist


crates/storage_impl/src/payments/payment_intent.rs
```diff
@@ -92,47 +95,35 @@ impl<T: DatabaseStore> PaymentIntentInterface for KVRouterStore<T> {
...
                     .construct_new()
                     .await
                     .change_context(StorageError::EncryptionError)?;
-
-                let diesel_payment_intent = payment_intent
-                    .clone()
-                    .convert()
-                    .await
-                    .change_context(StorageError::EncryptionError)?;
+                let diesel_payment_intent = new_payment_intent.clone().into();
```
hrithikesh026: avoid .into() calls. use 'PaymentIntentNew::from()'

hrithikesh026: Addressed by replacing the implicit `.into()` with `DieselPaymentIntent::from(new_payment_intent.clone())`.

XyneSpaces:
[should-fix] `crates/storage_impl/src/payments/payment_intent.rs:112` — The new KV insert path goes through `insert_resource`, which decrypts using `key_store.merchant_id` as the key-manager identifier. The PR correctly overrides the identifier to `processor_merchant_id` for update/find, but not for insert. Confirm `merchant_id` is the intended identifier for payment-intent inserts, or propagate `processor_merchant_id` here as well to avoid decryption failures when the two ids differ.

crates/storage_impl/src/payments/payment_intent.rs
```diff
@@ -2,7 +2,9 @@
 use api_models::payments::{AmountFilter, Order, SortBy, SortOn};
 #[cfg(feature = "olap")]
 use async_bb8_diesel::{AsyncConnection, AsyncRunQueryDsl};
-use common_utils::ext_traits::{AsyncExt, Encode};
+use common_utils::ext_traits::AsyncExt;
+#[cfg(feature = "v2")]
+use common_utils::ext_traits::Encode;
 #[cfg(feature = "v2")]
 use common_utils::fallback_reverse_lookup_not_found;
```
SanchithHegde: Nit, optional change: Can combine imports.

crates/storage_impl/src/payments/payment_intent.rs
SanchithHegde: If we're anyway calling `insert_resource()`, shouldn't we be dropping the `decide_storage_scheme()` call and the `match storage_scheme` logic?

crates/storage_impl/src/payments/payment_intent.rs
```diff
@@ -247,80 +246,45 @@ impl<T: DatabaseStore> PaymentIntentInterface for KVRouterStore<T> {
...
+        let drainer_query = diesel_intent_update
+            .clone()
+            .generate_drainer_update_query(
+                &mut query_gen_conn,
+                origin_diesel_intent.payment_id.clone(),
+                origin_diesel_intent.processor_merchant_id.clone(),
+            )
+            .await
+            .change_context(StorageError::KVError)
+            .attach_printable("Failed to generate payment intent update query")?;
```
SanchithHegde: I just realized, using `insert_resource()` / `update_resource()` would require us to eagerly generate the drainer query, even though we may not use / need it at all, based on which storage scheme is chosen.

How do you suggest we handle this?

hrithikesh026: I can accept a new future that generates the update and insert queries in `insert_resource` and `update_resource` functions. Instead of calculating them eagerly.

What do you think?

crates/storage_impl/src/payments/payment_attempt.rs
```diff
@@ -1154,10 +1154,7 @@ impl<T: DatabaseStore> PaymentAttemptInterface for KVRouterStore<T> {
 
         let mut query_gen_conn = pg_connection_write(self).await?;
         let drainer_query = payment_attempt_internal
```
SanchithHegde: Nit: Should we consider naming this `drainer_query_fut` instead, to clearly indicate that the value is a future?

crates/storage_impl/src/kv_router_store.rs
```diff
@@ -57,8 +57,11 @@ impl<T: DatabaseStore> KVRouterStore<T> {
     }
 }
 
-pub struct InsertResourceParams<'a> {
-    pub drainer_query: kv::SerializableQuery,
+pub struct InsertResourceParams<'a, DrainerQuery>
+where
+    DrainerQuery: futures::Future<Output = diesel_models::StorageResult<kv::SerializableQuery>>,
```
SanchithHegde: Nit: Similarly, should we call this `DrainerQueryFut`?

SanchithHegde (commented):

SanchithHegde (approved):

crates/router_derive/src/macros/from_new.rs
```diff
@@ -0,0 +1,59 @@
...
+/// - Both `Entity` and `EntityNew` share the same field names and compatible types.
+pub fn from_new_derive_inner(input: DeriveInput) -> syn::Result<TokenStream> {
+    let struct_name = &input.ident;
+    let new_type_name = syn::Ident::new(&format!("{}New", struct_name), struct_name.span());
+
+    let fields = match &input.data {
+        syn::Data::Struct(data) => match &data.fields {
+            Fields::Named(fields_named) => &fields_named.named,
+            _ => {
+                return Err(syn::Error::new_spanned(
```
kashif-m, line 18: nit: we can avoid early returns

kashif-m (approved):

PR #13192, opened 2026-07-07
feat(adyen/repeat_payment): map network_txn_link_id from UCS RepeatPayment response

### LINKED ISSUE #13191: adyen repeat_payment: network_txn_link_id dropped in UCS shadow response
## Bug Description
On adyen `repeat_payment` (recurring/MIT charge), the Unified Connector Service (UCS) response
mapping in `crates/router/src/core/unified_connector_service/transformers.rs` always sets
`network_txn_link_id: None`, even though adyen returns `additionalData.transactionLinkId` for
these transactions and the value is available (the field is already correctly mapped for the
`Authorize` flow).

## Root cause
The gRPC `RecurringPaymentServiceChargeResponse` proto message had no field to carry
`network_txn_link_id`, so HS's response transformer for the RepeatPayment/Charge RPC discarded
it. Fixed upstream in connector-service by adding the field to the proto and threading it
through `generate_repeat_payment_response`.

## Fix
- connector-service: adds `network_txn_link_id` to `RecurringPaymentServiceChargeResponse`
  and wires it through in `domain_types`.
- hyperswitch: maps the new field in the RepeatPayment response handler and bumps the pinned
  UCS dependency revision.

AmitsinghTanwar007:
## Summary
- Maps the new `network_txn_link_id` field on `RecurringPaymentServiceChargeResponse` into `PaymentsResponseData::TransactionResponse.network_txn_link_id` for the RepeatPayment/UCS charge response handler (`crates/router/src/core/unified_connector_service/transformers.rs`).
- Bumps the pinned `connector-service` dependency (`unified-connector-service-client` / `unified-connector-service-cards`) from `tag = "2026.07.02.1"` to `rev = "e49eef2ba197c3cd0edd632181d1f4c346dcc0e7"` (head of connector-service PR juspay/hyperswitch-prism#1839) across the three Cargo.tomls, so the generated proto types include the new field.
- **Depends on connector-service PR #1839** — the `rev` pin above is interim; once #1839 merges to `main`, this pin should move to a `tag`/main commit rather than staying on a raw `rev`.
- The rev bump also pulled in unrelated proto/domain drift accumulated on connector-service `main` since the previous tag; fixed the resulting compile breaks minimally:
  - `PaymentMethodAuthenticationServiceAuthenticateRequest` gained a `webhook_url` field (set to `None`, mirroring the existing `CompleteAuthorize` precedent — `PaymentsAuthenticateData` has no webhook URL).
  - `PaymentMethodAuthenticationServicePreAuthenticateRequest` gained a `merchant_transaction_id` field (set from `connector_request_reference_id`, same value already used for `merchant_order_id` elsewhere in this file).
  - `MandateAmountData` gained 6 subscription-related fields (`initial_billing_amount`, `external_subscription_id`, `mandate_status`, `next_billing_date`, `billing_cycle`, `description`); set to `None`/`Unspecified` since none of that data exists on the plain mandate-amount path.
  - `redirect_form::FormType` gained a `HostedIframe` variant; added a not-implemented arm mirroring the existing `Uri` arm (HS's `RedirectForm` domain enum has no hosted-iframe representation yet).

## Root cause
Detected via HS↔UCS shadow-validation diff on adyen `repeat_payment`: `router.typeDiff:response.Ok.TransactionResponse.network_txn_link_id` (hyperswitch: string, ucs: null). Verified that adyen's repeat-payment connector-integration code in connector-service already parses `additionalData.transactionLinkId` into the domain response; the value was only lost because (a) the proto response message had no slot for it and (b) HS's mapping hardcoded `None`.

Closes #13191

## Test plan
- [x] `cargo build` (default v1 features)
- [ ] `cargo build --no-default-features --features "release,v2,redis-rs"`
- [ ] `cargo clippy --all-targets -- -D warnings`
- [ ] `cargo fmt --all -- --check`
- [x] Verified locally: adyen repeat_payment shadow-validation diff on `network_txn_link_id` disappears (typeDiff empty) after this fix + connector-service PR #1839.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1550,7 +1550,7 @@ dependencies = [
  "bitflags 2.9.0",
  "cexpr",
  "clang-sys",
- "itertools 0.11.0",
+ "itertools 0.10.5",
  "log 0.4.27",
  "prettyplease",
  "proc-macro2",
@@ -2264,7 +2264,7 @@ dependencies = [
 [[package]]
 name = "config_patch_derive"
 version = "0.1.0"
-source = "git+https://github.com/juspay/connector-service?tag=2026.07.02.1#a070ef74a8d36aeca23f8e90d4d8bb0cc4db77ca"
+source = "git+https://github.com/juspay/connector-service?rev=e49eef2ba197c3cd0edd632181d1f4c346dcc0e7#e49eef2ba197c3cd0edd632181d1f4c346dcc0e7"
 dependencies = [
  "proc-macro2",
  "quote",
@@ -3982,7 +3982,7 @@ checksum = "32619942b8be646939eaf3db0602b39f5229b74575b67efc897811ded1db4e57"
 [[package]]
 name = "grpc-api-types"
 version = "0.1.0"
-source = "git+https://github.com/juspay/connector-service?tag=2026.07.02.1#a070ef74a8d36aeca23f8e90d4d8bb0cc4db77ca"
+source = "git+https://github.com/juspay/connector-service?rev=e49eef2ba197c3cd0edd632181d1f4c346dcc0e7#e49eef2ba197c3cd0edd632181d1f4c346dcc0e7"
 dependencies = [
  "axum 0.8.4",
  "bytes 1.10.1",
@@ -6836,7 +6836,7 @@ dependencies = [
  "regex-cache",
  "serde",
  "serde_derive",
- "strum 0.26.3",
+ "strum 0.25.0",
  "thiserror 1.0.69",
 ]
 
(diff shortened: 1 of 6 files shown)


XyneSpaces:
[blocking] Cargo.lock downgrades from connector-service revision switch

`Cargo.lock`: switching `connector-service` from tag `2026.07.02.1` to rev `e49eef2ba1...` forces transitive downgrades (`itertools 0.11.0 → 0.10.5`, `strum 0.26.3 → 0.25.0`, `heck 0.5.0 → 0.4.1`). This can break code relying on newer APIs and introduces unrelated dependency churn. Pin a compatible `connector-service` revision or explicitly pin these workspace crates to prevent regressions.

Shubhodip900:
closing this PR as this changes are merged via https://github.com/juspay/hyperswitch/pull/13394 PR

PR #13206, opened 2026-07-07
fix(ucs): preserve construct-time amount_captured on psync when UCS reports none (#18021)

10X-GRACE:
## Summary

Fixes the shadow router-data diff reported in juspay/hyperswitch-cloud#18021 — `router.typeDiff:amount_captured` on cybersource **psync** for merchant flowbird (prod).

On the UCS psync gateway, `router_data.amount_captured` / `minor_amount_captured` were **unconditionally** overwritten with `PaymentServiceGetResponse.captured_amount`:

```rust
router_data.amount_captured = payment_get_response.captured_amount;      // clobbers to None
router_data.minor_amount_captured = payment_get_response.captured_amount.map(MinorUnit::new);
```

Several connectors (e.g. **cybersource**) return **no captured amount on sync** — the cybersource TSS transaction-details response carries only a status (`{"id":…,"applicationInformation":{"status":…},"processorInformation":…}`, no `orderInformation`/`amountDetails`). So `captured_amount` is `None` and the field got nulled out.

The **Direct** psync path never overwrites `amount_captured`; it keeps the value populated at construct time from the payment intent (`payment_intent.amount_captured`). For an already-captured payment being synced, Direct therefore has a number and UCS had `null` → the shadow router-data type diff.

Fix: only override `amount_captured`/`minor_amount_captured` when UCS actually reports a captured amount. This matches the Direct gateway and also fixes UCS-**primary** psync correctness (previously a UCS-primary sync would wrongly null out a captured amount).

## Root cause (RCA)

- HS Direct `amount_captured` on psync = `payment_intent.amount_captured` (set in `construct_payment_router_data`, `transformers.rs:2163`), never touched by the cybersource sync transformer (`..item.data`).
- UCS shadow: `psync_gateway.rs` overwrote it with `payment_get_response.captured_amount`, which is `None` because prism's cybersource psync transformer has no captured amount to report (the connector doesn't return one; the synced status is often `PENDING`/`TRANSMITTED`, so it can't be derived from status either).
- Prod evidence (Loki): every flowbird cybersource psync TSS response contains no amount fields; `captured_amount: None`. The reported diff is identical across samples: `typeDiff: amount_captured {hyperswitch: number, ucs: null}` + `minor_amount_captured`.

## Verification (local shadow stack, clean `main` + this fix)

Repro: an auto-captured cybersource payment (`amount_captured = 2000` in the intent) synced with `force_sync=true&all_keys_required=true` so the connector is actually called.

**Before (clean main, x-req `019f3d57-1d9c-7160-8421-a0ef11f632aa`)** — compare-service `VS_48`:
```
differences_found: true, total_type_differences: 2
typeDiff: { "amount_captured": {"hyperswitch":"number","ucs":"null"},
            "minor_amount_captured": {"hyperswitch":"number","ucs":"null"} }
```

**After (this fix, x-req `019f3d60-db58-7b30-83a2-f3496bdd6b35`)** — compare-service `VS_46`:
```
Router data comparison completed successfully - no differences found
total_key_differences: 0, total_value_differences: 0, total_type_differences: 0
hyperswitch_data_keys: 56, ucs_data_keys: 56   (both statuses: pending)
```

No proto or connector-service change is required — the captured amount is hyperswitch state, not something the connector returns on sync.


10X-GRACE:
## Also fixes juspay/hyperswitch-cloud#18022 — `router.typeDiff:minor_amount_captured` (same root cause, same fix)

Issue **#18022** (`[prod] Validation router diff: flowbird / cybersource / psync`, fingerprint `8c303e5e69a4…`) is the **`minor_amount_captured` sibling facet of the exact same shadow diff** this PR already fixes for #18021. This PR resolves both fields in the same `if let Some(captured_amount)` guard, so it covers #18022 with no additional change.

### They are the same underlying diff (proven from prod)
Grafana/Loki RCA on #18022's sample request-id `019f197d-992b-7482-bafc-6411cc349581` (decoded UUIDv7 txn time `2026-06-30T17:04:49Z`) found the **single** VS_48 router-data comparison event (`{app="validation-service"}`, `x-connector: cybersource`, `x-sub-flow: PSync`, prod `payment_id: pay_5am68N0jqwa4qtRHF8ph`):

```
"summary":{"differences_found":true,"total_key_differences":0,"total_value_differences":0,"total_type_differences":2}
"typeDiff":{
  "amount_captured":       {"hyperswitch":"number","ucs":"null"},
  "minor_amount_captured": {"hyperswitch":"number","ucs":"null"}
}   keyDiff={}  valueDiff={}
```

`amount_captured` (#18021) and `minor_amount_captured` (#18022) are the **two type-diffs of one and the same comparison event** (56 keys each side) — the shadow-validation pipeline filed them as two issues. UCS-side prod logs (`{app="connector-service-grpc"}`) show the outbound cybersource TSS `GET /tss/v2/transactions/{id}` carries **no amount fields** → `captured_amount = None`, so the pre-fix gateway nulled both `amount_captured` and `minor_amount_captured`.

### Fresh local verification of #18022 (this branch, rebased on latest `main` @ `aa8d7c3ea7`)
Repro: auto-captured cybersource payment `pay_Qcz7PlUSn5yNSGeuKsCR` (`payment_intent.amount_captured = 2000`), synced with `force_sync=true&all_keys_required=true` so the connector is actually called. Cybersource TSS sync response confirmed amount-less: `{"id":…,"applicationInformation":{"status":"PENDING"},"processorInformation":…,"clientReferenceInformation":…}`.

**Before** = the prod diff above (VS_48, `minor_amount_captured {hyperswitch:number, ucs:null}`).

**After (this fix, x-req `019f3d6d-5515-7fd0-ac7b-373bb597ede2`)** — compare-service `VS_46`:
```
Router data comparison completed successfully - no differences found
total_key_differences: 0, total_value_differences: 0, total_type_differences: 0
hyperswitch_data_keys: 56, ucs_data_keys: 56   (both statuses: pending)
```

`minor_amount_captured` now matches (construct-time `2000` preserved on both sides instead of being clobbered to `null`). No proto/prism change needed — the captured amount is hyperswitch DB state, not returned by cybersource on sync.

Please add `Fixes juspay/hyperswitch-cloud#18022` alongside #18021 on merge.

XyneSpaces:
[should-fix] `crates/router/src/core/payments/gateway/psync_gateway.rs:113-116` and `:296-300` — add a regression test for the `captured_amount == None` guard.

The new `if let Some(...)` paths preserve the construct-time captured amount when UCS reports `None`, but the diff does not include a test asserting this behavior. Add a unit or mocked-UCS integration test that verifies both `UCSConsumeResponse` and `Trigger` PSync paths keep `router_data.amount_captured` / `minor_amount_captured` unchanged when `PaymentServiceGetResponse.captured_amount` is `None`.

Shubhodip900 (approved):

AmitsinghTanwar007 (approved):

sai-harsha-vardhan (approved):

PR #13208, opened 2026-07-07
fix(unified_connector_service): treat empty merchant_recurring_payment_id as None in SetupMandate response

### LINKED ISSUE #13207: authorizedotnet: SetupMandate response leaks empty connector_response_reference_id via UCS shadow path
In the HS↔UCS shadow-validation pipeline, authorizedotnet's SetupMandate (zero-auth mandate setup) success response shows a type mismatch on `connector_response_reference_id`: the HS-native leg correctly reports `None`, while the UCS-shadow leg (parsed from the gRPC `PaymentServiceSetupRecurringResponse`) reports `Some("")`.

Root cause: the UCS gRPC response field `merchant_recurring_payment_id` is a plain (non-optional) proto3 `string`, which defaults to `""` when the connector doesn't supply a reference id. On the HS side, `crates/router/src/core/unified_connector_service/transformers.rs` unconditionally wraps this in `Some(...)` for the success branch of the `PaymentServiceSetupRecurringResponse` → `PaymentsResponseData` conversion — unlike the adjacent error branch, which already does `.filter(|s| !s.is_empty())` to guard against exactly this.

Fix: apply the same empty-string filter to the success branch so `""` maps to `None`, matching HS-native behavior.

AmitsinghTanwar007:
## Description

Fixes a shadow-validation diff between HS-native and UCS-shadow responses for
`authorizedotnet` `SetupMandate` (zero-auth mandate-setup) calls.

**Diff signature:** `router.typeDiff:response.Ok.TransactionResponse.connector_response_reference_id`
(`hyperswitch: null`, `ucs: string`)

### Root cause

The UCS gRPC response field `PaymentServiceSetupRecurringResponse.merchant_recurring_payment_id`
is a non-optional proto3 `string`, which serializes to `""` (empty string) whenever the
connector doesn't supply a reference id — as is the case for authorizedotnet's
`createCustomerPaymentProfileRequest`-based SetupMandate response.

On the HS side, `crates/router/src/core/unified_connector_service/transformers.rs` maps
this field back into `PaymentsResponseData::TransactionResponse.connector_response_reference_id`.
The **error**-branch mapping already guards against this with `.filter(|s| !s.is_empty())`,
but the **success**-branch mapping wrapped the value unconditionally in `Some(...)`, so an
empty string became `Some("")` instead of `None` — diverging from the HS-native leg, which
produces `None` directly from authorizedotnet's response (no transaction id to map for a
CIM profile call).

### Fix

Apply the same `.filter(|s| !s.is_empty())` guard to the success-branch mapping, so `""`
consistently maps to `None`, matching HS-native behavior and the sibling error-branch code.

This is a pure HS→UCS response-mapping fix; no proto or UCS-side change was needed.

### Repro / verification (local stack, HS:8080 + UCS:8001 + MITM + validation-service)

Fired 3x `SetupMandate` (zero-auth, `off_session`, multi-use mandate) payments against
`authorizedotnet` with the merchant enrolled in the UCS shadow rollout, and polled the
validation-service router-data comparison for each `request_id`.

**Before fix:** every SetupMandate sample showed
```
typeDiff: {"response.Ok.TransactionResponse.connector_response_reference_id": {"hyperswitch": "null", "ucs": "string"}}
```

**After fix (rebuilt HS router on this branch, same repro re-run):**
```
differences_found: False
valueDiff: {}
keyDiff:   {}
typeDiff:  {}
```
for all SetupMandate samples, with no new diffs introduced.

### Checks run locally
- `cargo build --bin router` — clean
- `cargo clippy --package router -- -D warnings` — clean
- `cargo fmt --all -- --check` — clean
- `cargo check -p router --no-default-features --features "release,v2,redis-rs"` — clean (V2 feature build)

Closes #13207


Shubhodip900:
This change is done via https://github.com/juspay/hyperswitch/pull/13233 PR

PR #13213, opened 2026-07-08
fix(unified_connector_service): scope Google Pay decrypted-token substitution to connectors that consume it

### LINKED ISSUE #13212: fiuu (and other non-GPay-decrypt connectors): UCS shadow mapping wrongly substitutes decrypted Google Pay token
## Bug

When a merchant's `fiuu` connector account has `connector_wallets_details.google_pay` configured
(e.g. purely to enable the Google Pay JS session-token button), HS's shadow-validation mapping to
the Unified Connector Service (UCS) unconditionally swaps in a server-side-decrypted Google Pay
token (`PaymentMethodToken::GooglePayDecrypt`) whenever one is available on the router data —
regardless of whether the destination connector's real transformer ever reads it.

`fiuu`'s real `WalletData::GooglePay` transformer (`crates/hyperswitch_connectors/src/connectors/fiuu/transformers.rs`)
never inspects `payment_method_token`; it always builds the connector request from the
client-submitted (typically `Encrypted`) `tokenization_data`. So HS's real outbound request to
Fiuu always succeeds using the encrypted token. But the UCS shadow-mapping (`get_google_pay_tokenization_data`
in `crates/router/src/core/unified_connector_service.rs`) blindly substitutes the decrypted
variant, which Fiuu's UCS-side transformer cannot build a request from (missing `token`/`token_type`
fields that only exist on the `Encrypted` variant) — producing a structural request-body diff
between HS and UCS purely for the shadow leg.

## Fix

Make the substitution connector-aware: only apply it for connectors whose real HS transformer
actually consumes `PaymentMethodToken::GooglePayDecrypt` for their `WalletData::GooglePay` flow
(confirmed via code read: cybersource, aci, tesouro, worldpayxml, stripe, adyen, nmi,
worldpayvantiv, checkout, nuvei). Every other connector (including fiuu) keeps using the original
tokenization data, matching real production behavior.

AmitsinghTanwar007:
## Diff signature (shadow-validation issue)

fiuu / authorize, connector-side request body:

```
body.keyDiff.GooglePay = {"hyperswitch":false,"ucs":true}
body.keyDiff.GooglePay[apiVersionMinor] = {"hyperswitch":true,"ucs":false}
body.keyDiff.GooglePay[apiVersion] = {"hyperswitch":true,"ucs":false}
body.keyDiff.GooglePay[paymentMethodData][info][cardDetails] = {"hyperswitch":true,"ucs":false}
```

## Root cause

`fiuu`'s real (non-UCS) `WalletData::GooglePay` transformer
(`crates/hyperswitch_connectors/src/connectors/fiuu/transformers.rs`) never reads
`payment_method_token` — it always builds the connector request from the client-submitted
`tokenization_data` (normally `Encrypted`), producing flat, bracket-keyed form fields like
`GooglePay[apiVersion]`, `GooglePay[apiVersionMinor]`, `GooglePay[paymentMethodData][info][cardDetails]`
via `build_form_from_struct`'s generic `serde_json::to_value` → flatten-to-form-fields path.

The HS→UCS shadow gRPC mapping (`get_google_pay_tokenization_data` in
`crates/router/src/core/unified_connector_service.rs`), however, unconditionally substitutes a
server-side-decrypted Google Pay token (`PaymentMethodToken::GooglePayDecrypt`) into the payload
sent to UCS whenever one happens to be present on the router data — regardless of whether the
destination connector's real transformer ever consumes it. This decryption is driven purely by
whether the connector's `connector_wallets_details` metadata has Google Pay merchant/session-token
details configured (e.g. to power the Google Pay JS button), independent of whether the connector
itself uses the decrypted payload for the actual authorization call.

Fiuu's UCS-side connector transformer (byte-identical struct/serde layout to HS's — confirmed via
`git blame`, both traced to the same original commit) cannot build a `FiuuGooglePayData` request
from the `Decrypted` variant: `GpayTokenizationData::get_encrypted_google_pay_token()` returns an
error for `Decrypted` (it only exists on the `Encrypted` variant), so the request-building step
fails for the shadow leg only — producing exactly the observed diff (UCS body missing the
bracket-keyed GooglePay fields that HS always sends).

Confirmed via code read that this connector-blind substitution is legitimate for some connectors
(their real HS transformer *does* consume `PaymentMethodToken::GooglePayDecrypt` for
`WalletData::GooglePay`: `cybersource`, `aci`, `tesouro`, `worldpayxml`, `stripe`, `adyen`, `nmi`,
`worldpayvantiv`, `checkout`, `nuvei`) but wrong for every other connector (`fiuu`, `rapyd`,
`worldpay`, `barclaycard`, `finix`, …), whose real transformer only ever calls
`get_encrypted_google_pay_token()`/errors on decrypted tokens.

## Fix

Gate the decrypted-token substitution behind an allowlist of connectors whose real transformer
actually consumes it, threading `connector_name` through
`build_unified_connector_service_payment_method` (10 call sites in
`unified_connector_service/transformers.rs`, all of which already have `router_data.connector`
available). For every other connector (including fiuu), the shadow mapping now keeps the original
tokenization data — matching what HS's real request actually sends.

## Verification

- Full crypto end-to-end reproduction isn't practical locally: triggering the real substitution
  requires a genuinely Google-signed decrypted Google Pay payload (same class of limitation as
  other Apple-Pay/Google-Pay-decrypt shadow-diff issues — the decrypted material only exists from
  real client-side/Google infrastructure).
- Instead verified the fix directly at the unit level against the exact pure functions changed
  (`crates/router/src/core/unified_connector_service.rs`, `google_pay_shadow_mapping_tests` module):
  - allowlisted connectors (`cybersource`, `aci`, `tesouro`, `worldpayxml`, `stripe`, `adyen`, `nmi`,
    `worldpayvantiv`, `checkout`, `nuvei`) still get the decrypted substitution when a decrypted
    token is present.
  - `fiuu` (and any non-allowlisted connector) keeps the original `Encrypted` tokenization data
    even when a decrypted token is present on the router data — this is the regression test for
    the exact diff in this issue.
  - `cargo test --lib -p router google_pay_shadow_mapping_tests`: 4/4 passed.
- `cargo build --bin router`: clean.
- `cargo fmt --all -- --check`: clean.
- `cargo clippy -p router --all-targets -- -D warnings`: clean.
- `cargo check --no-default-features --features "release,v2,redis-rs"` (the CI V2-features check):
  clean (only pre-existing warnings).

## Category

HS→UCS gRPC mapping layer (`crates/router/src/core/unified_connector_service/`) — no UCS-side or
HS connector-transformer changes needed; `fiuu`'s transformers on both sides are untouched and
byte-identical to each other.

Closes #13212


XyneSpaces:
[should-fix] Hardcoded connector string allowlist bypasses compile-time safety

`connector_supports_google_pay_decrypted_flow` matches raw strings like `"cybersource"` instead of using the typed connector enum. A typo or rename will silently break the allowlist, and adding a new connector that consumes decrypted Google Pay data won't be forced to register here.

Use `common_enums::Connector` (or the existing `router_data.connector` type) in the match so mismatches are caught at compile time.

Location: `crates/router/src/core/unified_connector_service.rs`, function `connector_supports_google_pay_decrypted_flow` (~line 100).

crates/router/src/core/unified_connector_service.rs
```diff
@@ -3569,3 +3575,77 @@ pub async fn call_unified_connector_service_for_notify_connector(
...
+    }
+
+    #[test]
+    fn keeps_encrypted_data_when_no_connector_metadata() {
+        let token = decrypted_token();
+        let result =
+            get_google_pay_tokenization_data(&encrypted_tokenization_data(), Some(&token), None);
+        assert!(matches!(result, GpayTokenizationData::Encrypted(_)));
+    }
+}
```
Shubhodip900: can we remove this `#[cfg(test)]` portion

Shubhodip900 (dismissed):

gopikrishna000 (dismissed):

Shubhodip900 (approved):

gopikrishna000 (approved):

10X-GRACE:
## Post-merge validation against shadow-validation issue #20872 (fiuu / authorize / `request_diff`)

Re-validated this merged PR from a clean `main` as part of resolving
[hyperswitch-cloud#20872](https://github.com/juspay/hyperswitch-cloud/issues/20872). **No new PR opened** —
this change already covers the issue.

### Merge state

| | |
|---|---|
| Merge commit | `924753b583a77556fe66b18934cdbef693d5e0b3` |
| Merged at | `2026-07-21T08:57:40Z` |
| On `origin/main` | ✅ ancestor of `d5188edcae` (`git merge-base --is-ancestor` → true) |
| `cargo build -p router` from clean `main` | ✅ exit 0 |
| CI on the PR | all substantive checks pass (Build hyperswitch server, MSRV ×2, V2-features, fmt, wasm, CodeQL). Only the non-blocking `Spell check` job is red. |

### Before (issue diff signature)

```
body.keyDiff.GooglePay                                            = {"hyperswitch":false,"ucs":true}
body.keyDiff.GooglePay[apiVersionMinor]                           = {"hyperswitch":true,"ucs":false}
body.keyDiff.GooglePay[apiVersion]                                = {"hyperswitch":true,"ucs":false}
body.keyDiff.GooglePay[paymentMethodData][info][cardDetails]      = {"hyperswitch":true,"ucs":false}
body.keyDiff.GooglePay[paymentMethodData][info][cardNetwork]      = {"hyperswitch":true,"ucs":false}
body.keyDiff.GooglePay[paymentMethodData][tokenizationData][token]= {"hyperswitch":true,"ucs":false}
body.keyDiff.GooglePay[paymentMethodData][tokenizationData][type] = {"hyperswitch":true,"ucs":false}
body.keyDiff.GooglePay[paymentMethodData][type]                   = {"hyperswitch":true,"ucs":false}
```

merchant `my_edge_gt`, request-id `019f40c1-f83d-7b12-8623-2cd1e7dd0bd2`, occurrences 1.

### Root cause — independently re-confirmed by reading both repos

The 7 `{"hyperswitch":true,"ucs":false}` facets are exactly the fields of `FiuuGooglePayData`, whose
serde renames are **character-identical in both repos**:

* HS Direct — `crates/hyperswitch_connectors/src/connectors/fiuu/transformers.rs:417-442` (struct),
  builder at `:709-750`, which calls `get_encrypted_google_pay_token()` at `:730` and
  `get_encrypted_token_type()` at `:737`.
* UCS/prism — `crates/integrations/connector-integration/src/connectors/fiuu/transformers.rs:458-484`
  (struct), builder at `:1148-1192`, same two calls at `:1171` / `:1179`.

Both getters **hard-error on the `Decrypted` variant** — same guard, same message
(`"Encrypted Google Pay payment data is mandatory"`): HS `crates/common_types/src/payments.rs:638-648`,
prism `crates/types-traits/domain_types/src/payment_method_data.rs:1183-1193`. Neither repo's fiuu
GooglePay path has any pre-decrypt fallback branch, and neither reads
`PaymentMethodToken::GooglePayDecrypt` for `WalletData::GooglePay` (HS rejects it outright in the
ApplePay arm at `fiuu/transformers.rs:550-564`).

So pre-fix, the shadow leg received `GpayTokenizationData::Decrypted` (substituted by
`get_google_pay_tokenization_data`) — a variant fiuu's UCS transformer provably cannot build a
request from — while the Direct leg kept the client's `Encrypted` data and emitted all 8
bracket-keyed form fields. Hence the one-sided `keyDiff`.

### Why the merged gate is the correct fix (structural equivalence)

The merged version gates the substitution on `is_googlepay_predecrypted_flow_supported(...)`
(`crates/router/src/core/unified_connector_service.rs:96`) — which is the **same predicate, applied to
the same argument**, that the real Direct path already uses to decide whether a pre-decrypted Google
Pay token is admissible:

* Direct: `crates/router/src/core/payments.rs:7570-7571` → `is_googlepay_predecrypted_flow_supported(merchant_connector_account.get_metadata())`
* Shadow (this PR): `unified_connector_service.rs:96` ← `router_data.connector_meta_data`, which is
  assigned from `merchant_connector_account.get_metadata()` at
  `crates/router/src/core/payments/transformers.rs:302, 544, 940, 1371, 1624`.

Both resolve to `GooglePayDetailsWrapper::is_predecrypted_token_supported()`
(`crates/api_models/src/payments.rs:9950-9952`, `support_predecrypted_token.unwrap_or(false)`).

That makes the shadow leg's tokenization data **equal by construction** to what the Direct leg
carries, for every connector and every merchant — strictly better than the connector allowlist the PR
originally proposed, since it needs no maintenance as connectors are added. For fiuu/`my_edge_gt` the
flag is not enabled (otherwise the Direct leg would itself have taken the `Decrypted` branch at
`payments.rs:7567`), so the gate returns `false`, the substitution is skipped, the `Encrypted` data is
preserved, and both legs serialize the identical 8 `GooglePay[...]` form fields.

### Verification — and its honest limits

* ✅ Fix confirmed present on `origin/main`; `cargo build -p router` from clean `main` exits 0.
* ✅ Root cause re-derived independently from both repos' source (citations above) and it matches the
  reported signature field-for-field.
* ⚠️ **A full end-to-end local repro was not run, and I am not claiming a green `keyDiff: {}` from the
  validation service.** Triggering the substitution requires the router to perform real server-side
  Google Pay decryption, which needs a genuinely Google-signed payload plus GPay decryption keys on
  the MCA — that material cannot be synthesized locally. The unit tests that covered this were
  removed from the PR before merge (commit `9316e236`), so no automated regression guard remains for
  this specific gate.
* ℹ️ One facet whose exact wire mechanism I could **not** pin: `body.keyDiff.GooglePay =
  {"hyperswitch":false,"ucs":true}` (a bare `GooglePay` key present only on the UCS side). Neither
  repo's fiuu code emits a form key named exactly `GooglePay` — `grep '"GooglePay"'` over both
  connector directories returns nothing — so this is most likely an artifact of how the comparator
  normalises bracket-notation form keys when the UCS side has no `GooglePay[...]` children at all,
  rather than an additional code divergence. Flagging it as inference, not verified fact. It does not
  affect the classification or the fix: it is the same single failure (UCS could not build the
  GooglePay request) viewed from the other side.

### Suggested follow-up (not blocking)

Since the regression tests were dropped at merge, consider re-adding a small unit test over
`get_google_pay_tokenization_data` asserting that a non-pre-decrypt-enabled MCA metadata preserves the
`Encrypted` tokenization data. Happy to open that separately if useful.

PR #13161, opened 2026-07-03
fix(connector): [Adyen] Handle non-JSON error responses gracefully

Vani-1107:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

This PR updates Adyen error response handling to avoid failing with ResponseDeserializationFailed when the connector returns an empty or non-JSON error body.

### Additional Changes

- [ ] This PR modifies the API contract
- [ ] This PR modifies the database schema
- [ ] This PR modifies application configuration/environment variables

## How did you test it?

Sanity Testing for failed cases - 

Request:

```

{
    "amount": 509,
    "order_tax_amount": null,
    "currency": "USD",
    "amount_to_capture": null,
    "shipping_cost": null,
    "merchant_id": null,
    "routing": null,
    "connector": null,
    "capture_method": "automatic",
    "authentication_type": null,
    "billing": {
        "address": {
            "city": "SANTA MARIA",
            "country": "US",
            "line1": "jhjk",
            "line2": null,
            "line3": null,
            "zip": "123456",
            "state": "hk",
            "first_name": "jgh",
            "last_name": "ggg",
            "origin_zip": null
        },
        "phone": null,
        "email": "test@gmail.com"
    },
    "capture_on": null,
    "confirm": true,
    "customer": {
        "id": "01a2e234c162994dd550c67a0119aba9",
        "name": "vhh jjk",
        "email": "test@gmail.com",
        "phone": null,
        "phone_country_code": null,
        "tax_registration_id": null,
        "document_details": null
    },
    "customer_id": null,
    "email": null,
    "name": null,
    "phone": null,
    "phone_country_code": null,
    "off_session": null,
    "description": null,
    "return_url": "https://api.convesiopay.com/psp/v1/threeds-redirect/4844b9e0b7584d3eb6981c4e2b30bc01",
    "setup_future_usage": "off_session",
    "payment_method_data": {
        "card": {
            "card_number": "4242424242424242",
            "card_exp_month": "12",
            "card_exp_year": "30",
            "card_holder_name": "Test Card",
            "card_cvc": "121",
            "card_issuer": null,
            "card_network": null,
            "card_type": null,
            "card_issuing_country": null,
            "card_issuing_country_code": null,
            "bank_code": null,
            "nick_name": null
        },
        "billing": null
    },
    "payment_method": "card",
    "payment_token": null,
[...]

PR #12877, opened 2026-06-22
refactor(adyen): gracefully handle unknown response enum variants and missing optional fields

sai-harsha-vardhan:
## Summary

Adyen has caused 7 incidents by adding new status/event values our enums didn't handle. This PR makes the Adyen direct connector resilient to unknown response values by applying two patterns from the connector response deserialization resilience design:

- **Pattern 1**: Add `#[serde(other)] Unknown` to response enums so unrecognised values deserialize gracefully instead of hard-failing
- **Pattern 3**: Loosen unused response struct fields so shape changes don't cause failures

### Pattern 1 — Unknown enum variants

`WebhookEventCode` already had `#[serde(other)]`. Added it to the remaining response enums:

| Enum | Response struct | `Unknown` handling |
|------|----------------|--------------------|
| `AdyenStatus` | `AdyenResponse`, `RedirectionResponse`, `PresentToShopperResponse`, `QrCodeResponseResponse` | Returns `prev_status` with warn log |
| `AdyenWebhookStatus` | `AdyenWebhookResponse` | `WebhookBodyDecodingFailed` |
| `CancelStatus` | `AdyenCancelResponse` | Stored, not consumed in business logic |
| `ActionType` | Redirect/QR action structs | Stored, not consumed in business logic |
| `DisputeStatus` | Dispute webhook structs | Absorbed by existing `Some(_)` catch-all arms |

**`AdyenStatus::Unknown` — `prev_status` threading:**

`get_adyen_payment_status()` now takes `prev_status: AttemptStatus` and returns it unchanged when `Unknown` is received (with a warn log so the occurrence is observable). All five response helper functions thread it from `item.data.status` at the `ForeignTryFrom` call site:

```
Adyen returns unknown status string
        │
        ▼
#[serde(other)] Unknown — deserialization succeeds
        │
        ▼
get_adyen_payment_status(..., prev_status)
        │
AdyenStatus::Unknown => prev_status   ← DB state unchanged, no HE_00
```

Mirrors the same pattern already in place for UCS (`RefundStatus::Unspecified → prev_status`).

### Pattern 3 — Unused fields loosened

Fields parsed from Adyen responses but never read in business logic are changed to accept any shape:

| Struct | Field | Before | After |
|--------|-------|--------|-------|
| `AdyenRefundResponse` | `merchant_account` | `Secret<String>` | `Secret<serde_json::Value>` |
| `AdyenRefundResponse` | `payment_psp_reference` | `String` | `Option<String>` |
[...]

PR #13214, opened 2026-07-08
fix(novalnet): add connector response deserialization resilience

sai-harsha-vardhan:
## Summary

Applies all 8 connector resilience patterns to the Novalnet connector, matching the same hardening done for Adyen (ref: #12877, #13161).

**transformers.rs**
- Added `#[serde(other)] Unknown` catch-all to all response enums: `NovalNetPaymentTypes`, `NovalnetTransactionStatus`, `NovalnetAPIStatus`, `WebhookEventType`
- `NovalnetTransactionStatus::Unknown` → `AttemptStatus::Unspecified` (with `warn!` log) in payment status mapping; → `RefundStatus::Pending` (with `warn!` log) in refund status mapping
- `NovalnetAPIStatus::Unknown` handled in all 5 `TryFrom` match arms (payments, psync, capture, refund execute, refund sync, cancel) — treated as failure with `warn!` log
- `WebhookEventType::Unknown` → `IncomingWebhookEvent::EventNotSupported` in `get_incoming_webhook_event`
- `get_novalnet_dispute_status` return type changed to `Result<WebhookDisputeStatus, _>` — `Unknown` and non-dispute event types return `Err(WebhookBodyDecodingFailed)` instead of silently mapping to an assumed dispute state

**novalnet.rs**
- `build_error_response`: added empty body guard; JSON parse changed to `Result<T, _>`; `Err` arm writes to `event_builder` via `set_error`, logs `router_env::logger::error!`, then delegates to `handle_json_response_deserialization_failure`
- `get_webhook_event_type`: added empty body guard returning `Ok(EventNotSupported)`; added early return for `Unknown` event type
- `get_dispute_details`: propagates `get_novalnet_dispute_status` error with `?`

Cypress Test report
image


XyneSpaces:
[blocking] Raw connector response body stored in event on deserialization failure

The `Err` branch in `build_error_response` writes `res.response.escape_ascii().to_string()` into the connector event. Connector error bodies can contain PANs, secrets, or PII; capturing them in observability events is a data-leakage risk.

Omit the raw body from the event or redact sensitive fields before logging. Keep only `status_code` and the parsed deserialization error context.

Location: `crates/hyperswitch_connectors/src/connectors/novalnet.rs`, `Err(error_msg) =>` branch in `build_error_response` (~line 190).

swangi-kumari (approved):

Sakilmostak (approved):

PR #13228, opened 2026-07-08
fix(connector): [noon] gracefully handle unknown response enum variants and missing optional fields

AkshayaFoiger:
## Type of Change

- [x] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

- Catch-all Unknown variants added to every response-deserialized enum, so unrecognized values from Noon fail to deserialize gracefully instead of erroring out the whole payload:
  - NoonPaymentStatus
  - NoonRevokeStatus
  - RefundStatus
  - (NoonWebhookEventTypes already had this)
- Payment status: get_payment_status now treats NoonPaymentStatus::Unknown the same way it already treats the existing Locked case — logs a warning and retains the current/previous AttemptStatus rather than assuming any concrete outcome.
- Refund status: extracted the inline From<RefundStatus> for enums::RefundStatus conversion into a free function get_refund_status((RefundStatus, enums::RefundStatus)), now threading the pre-call refund status (RefundsData.refund_status) through both the Execute and RSync handlers. RefundStatus::Unknown logs a warning and retains the previous refund status instead of assuming Pending. (Note: a trait-based tuple From impl isn't possible here — enums::RefundStatus and From are both foreignrameter trips Rust's orphan rule (E0117); a plain function,matching the existing get_payment_status convention in this file, sidesteps that.)
- Mandate revoke status: NoonRevokeStatus::Unknown returns an::ResponseHandlingFailed, viareport!(...).attach_printable(...)) rather than silently assuming the mandate was revoked — an incorrect assumption here could mask a failed revocation.
- Unused response fields loosened: class_description: String Response, RefundResponse, and RefundSyncResponse — verified via grep that this field is never read anywhere in the transform logic, so a future missing/renamed field on Noon's side won't break deserialization.

## How did you test it?

Screenshot 2026-07-08 at 8 11 30 PM

The threeds check failure is an existing behavior 

## Checklist


XyneSpaces:
[should-fix] Make `message` optional alongside `class_description`

`RefundResponse`, `RefundSyncResponse`, and `NoonErrorResponse` now use `class_description: Option<String>`, but `message` remains a required `String`. If Noon can omit `class_description`, it can also omit `message`, so a missing field will still cause deserialization to fail. Consider changing `message` to `Option<String>` as well, or document why it is guaranteed to always be present.

swangi-kumari (dismissed):

sai-harsha-vardhan (dismissed):

swangi-kumari (approved):

sai-harsha-vardhan (approved):

PR #13242, opened 2026-07-09
fix(unified_connector_service): complete PayPal wallet redirect over UCS

shuklatushar226:
## What

Fixes the PayPal **wallet redirect** flow over the Unified Connector Service (UCS): after the buyer approves at PayPal and returns, the payment looped at `requires_customer_action` instead of completing.

## Root cause

On the `CompleteAuthorize` → `PaymentService/Authorize` return, UCS was **creating a brand-new PayPal order** (`POST /v2/checkout/orders`) instead of **capturing the existing, approved order**, because the pending order id was never threaded onto the request the completion gateway actually builds.

Note: the completion gateway (`complete_authorize_gateway.rs`) calls the **tuple-arg** `ForeignTryFrom<(&RouterData<CompleteAuthorize,…>, CallConnectorAction)>` builder — not the single-arg one — so the fix had to live there.

## Changes (`crates/router/src/core/unified_connector_service/transformers.rs`)

**CompleteAuthorize → Authorize (tuple-arg builder):**
- Thread `connector_order_id` from `connector_transaction_id` so UCS captures the **existing** order (`/v2/checkout/orders/{id}/capture`) instead of creating a new one.
- Reconstruct a minimal `PaypalRedirect` `payment_method` on the bare redirect-return callback (where `payment_method_data` is `None`) so the authorize call is valid.

**RecurringPaymentServiceCharge (wallet-MIT redirect parity):**
- Carry `complete_authorize_url` on the charge request and map `redirection_data` from the charge response, so a redirected MIT surfaces `next_action` and can complete.
- Fall back to `recurring_mandate_payment_data.payment_method_type` for MIT charges.

## Companion / dependency

Pairs with connector-service `feat/paypal_grace` (PayPal completion sends an **empty** capture body; adds `redirection_data`/`complete_authorize_url` to the recurring-charge proto). The MIT proto-dependent lines here compile once those land and the pinned `unified-connector-service-client` tag is bumped; the core CIT fix (`connector_order_id` + `payment_method` reconstruct) uses existing fields.

## Testing

Local HS → UCS, PayPal sandbox: wallet CIT (`setup_future_usage=off_session`) → approve → **`succeeded`** (order captured on the existing id, vault `VAULTED`, `amount_capturable=0`). Previously looped at `requires_customer_action`.


crates/router/src/core/unified_connector_service/transformers.rs
```diff
@@ -595,10 +595,24 @@ impl
...
-            .request
-            .payment_method_data
-            .clone()
+        // On a redirect-complete callback (e.g. PayPal wallet), the inbound request is a bare
+        // GET carrying only redirect params, so `payment_method_data` is None. UCS/prism still
+        // requires `payment_method` on the authorize call, so reconstruct a minimal payment
+        // method from the attempt's `payment_method_type` for redirect-based APMs. Mirrors the
+        // CompleteAuthorize reconstruct used for other redirect connectors.
+        let payment_method_data = router_data.request.payment_method_data.clone().or_else(|| {
+            use hyperswitch_domain_models::payment_method_data::{
```
pixincreate: move this to the top so that it will become easier to track.
and remove unnecessary explanatory comments

crates/router/src/core/unified_connector_service/transformers.rs
```diff
@@ -595,10 +595,24 @@ impl
...
+                PaymentMethodData, PaypalRedirection, WalletData,
+            };
+            match router_data.request.payment_method_type {
+                Some(common_enums::PaymentMethodType::Paypal) => Some(PaymentMethodData::Wallet(
+                    WalletData::PaypalRedirect(PaypalRedirection { email: None }),
```
pixincreate: why hardcoding email to `None`? you should be getting that data in the redirection data, right?

pixincreate: also, will we be doing the same / similar change for other payment methods in future? if yes, i think, it is better to move it.

shuklatushar226: PayPal's redirect back is a bare `GET` to the complete-authorize URL carrying only `token` and `PayerID` — `redirect_response.payload` is `None`, and `ConnectorRedirectResponse::get_flow_type` (`paypal.rs:2553-2568`) discards both the query params and the payload and just returns `Trigger`. The payer email only comes back in the order *response*, never in the redirection data.

`request.email` is populated though — set from customer data at `core/payments/transformers.rs:7231` — and it's the same `common_utils::pii::Email` type, so I've switched to that instead of hardcoding `None`. It reaches the wire via the existing mapping at `core/unified_connector_service.rs:1482-1490`.

shuklatushar226: Yes — likely for other redirect-based APMs over UCS. Extracted it into `reconstruct_payment_method_data_for_redirect_completion` in `core/unified_connector_service.rs`, sitting next to `build_unified_connector_service_payment_method` since that's what consumes its output. Adding another APM is now a single match arm.

crates/router/src/core/unified_connector_service/transformers.rs
shuklatushar226: Done — the inline `use` and the explanatory comment are both gone; the call site is now a single delegating call and the rationale lives on the helper's doc comment.

One note on why the helper landed in `core/unified_connector_service.rs` rather than hoisting the imports to the top of `transformers.rs`: `unused_qualifications` is a workspace-level lint (`Cargo.toml:15`), and `transformers.rs:2348` already uses `hyperswitch_domain_models::payment_method_data::PaymentMethodData::MandatePayment` fully qualified. Importing the type at module level would have turned that existing line into a lint failure under `-D warnings`. Keeping the types in the other module avoids touching it. Also trimmed the other comments this PR added down to just the why.

shuklatushar226: Correction to my previous reply — I was wrong about `unused_qualifications`, and I should not have stated it without checking.

I tested it: with `PaymentMethodData` imported at module level, rustc emits **no** `unnecessary qualification` warning for `transformers.rs:2348`. It flags the *import* as unused instead, because a fully-qualified path to an enum variant does not count as using the import and does not trip the lint. So hoisting the import as you suggested would have compiled cleanly — my reasoning there was wrong.

The helper extraction stands on its own merits (your third comment, re: other payment methods), so the current code is unchanged. But the justification I gave on this thread was incorrect and I did not want to leave it standing.

crates/router/src/core/unified_connector_service/transformers.rs
```diff
@@ -794,7 +798,8 @@ impl
                 .map(payments_grpc::Tokenization::foreign_from)
                 .map(Into::into),
             l2_l3_data: None,
-            connector_order_id: None,
+            // Captures the order created before the redirect instead of creating a new one.
```
pixincreate, line 801: nit: we don't need these comments.

pixincreate (approved):
lgtm.

Nithin1506200 (approved):

PR #13251, opened 2026-07-09
fix(nmi): harden connector against unknown enum variants and non-JSON error bodies

sai-harsha-vardhan:
## Summary
- Add `#[serde(other)] Unknown` catch-alls to NMI's `Response`, `NmiWebhookEventType`, and `NmiActionType` enums so a new connector-side status/event/action code degrades gracefully instead of hard-failing deserialization (mirrors the pattern already merged for Adyen in #12877, #13161).
- Unknown payment/refund statuses now retain the pre-call `AttemptStatus`/`RefundStatus` (threaded through as `prev_status` / `request.refund_status`) with a warn log, rather than assuming success or failure.
- `build_error_response` now guards against an empty body and falls back to `handle_json_response_deserialization_failure` for non-JSON error responses instead of panicking (NMI is known to return non-JSON bodies on some failure paths).
- `get_webhook_event_type` guards against an empty webhook body.
- Reviewed but not changed: `NmiStatus` (already resilient via its existing `From<String>` fallback to `Pending`), dispute handling (NMI has none), UCS/hyperswitch-prism gap (not applicable — no such code in this repo), unused response fields (none found), `deny_unknown_fields` (not used anywhere in this connector).

Cypress Tests
image

### DIFF
--- a/crates/hyperswitch_connectors/src/connectors/nmi.rs
+++ b/crates/hyperswitch_connectors/src/connectors/nmi.rs
@@ -41,6 +41,7 @@ use hyperswitch_interfaces::{
         ConnectorSpecifications, ConnectorValidation,
     },
     configs::Connectors,
+    consts::{NO_ERROR_CODE, NO_ERROR_MESSAGE},
     errors::ConnectorError,
     events::connector_api_logs::ConnectorEvent,
     types::{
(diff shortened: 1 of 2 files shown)


AkshayaFoiger (approved):

XyneSpaces:
:warning: **[should-fix]** `NmiActionType::Unknown` returns `Err(ConnectorError::WebhooksNotImplemented)` in `get_webhook_object_reference_id`, while `NmiWebhookEventType::Unknown` is mapped to `EventNotSupported` in `get_webhook_event_type`.

If an unknown-but-well-formed webhook reaches the object-reference step, this error will cause the webhook endpoint to return a non-2xx response. Connectors typically retry on non-2xx, leading to duplicate delivery attempts for events that the integration intentionally does not handle. Align the two handlers: either map `NmiActionType::Unknown` to an acknowledged-but-no-op path, or ensure the framework never calls `get_webhook_object_reference_id` for `EventNotSupported` webhooks.

swangi-kumari (approved):

PR #13252, opened 2026-07-09
Revert: "feat(core/connector): [SANTANDER] Add UpdatePostConfirm Flow for Payments Update API"

bsayak03:
Reverts juspay/hyperswitch#13230


Vani-1107 (approved):

XyneSpaces:
:rotating_light: **[blocking]** `PaymentIntentUpdateFields` and `PaymentIntentUpdateInternal` drop `billing_descriptor` and `connector_metadata` fields, and every `From<PaymentIntentUpdate>` arm now hard-codes them to `None`.

Removing these fields from the update carrier structs is a backwards-incompatible change: any existing code or persisted data that sets/expects these fields through the generic `PaymentIntentUpdate::Update` path will silently lose them. The PR description frames this as metadata-update cleanup, but it does not show a migration strategy or a transition period where both shapes are accepted.

Either preserve the fields in `PaymentIntentUpdateFields`/`PaymentIntentUpdateInternal` (and thread them through) until a migration is ready, or document the migration plan and confirm that all producers/consumers of these structs have been updated.

PR #13253, opened 2026-07-09
fix(connector): [fiuu] gracefully handle unknown response enum variants and missing optional fields

AkshayaFoiger:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [x] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

- build_error_response: empty-body guard, Result-based parse instead of .change_context(...)?, logs to event_builder before falling back to handle_json_response_deserialization_failure.
- get_webhook_event_type: empty body → IncomingWebhookEvent::EndpointVerification (fiuu-specific, per your instruction).
- Added #[serde(other)] Unknown to FiuuRecurringStautus, StatCode, StatName, FiuuPaymentWebhookStatus, FiuuRefundsWebhookStatus, RefundStatus.
- Threaded prev_status through all AttemptStatus mappings (recurring, sync, webhook) — Unknown retains the pre-call status instead of guessing.
- Refund status mappings (get_refund_status_from_sync, get_refund_status_from_webhook) now take prev_status: enums::RefundStatus (sourced from item.data.request.refund_status) and log it on Unknown, still falling back to Pending — message text is now "...; previous status was {:?}, marking refund as pending", dropping the old "for resync" phrasing.
- Unknown webhook status codes map to EventNotSupported, never Err.
- Loosened confirmed-unused mandatory fields (reference_no, txn_type, txn_currency, txn_amount, txn_channel, FiuuRecurringResponse.order_id, nbcb, channel) to Option; deleted the now-dead RequestType enum.

## How did you test it?

Screenshot 2026-07-09 at 6 51 41 PM

## Checklist


XyneSpaces:
Automated review completed: no blocking or should-fix findings identified.

crates/hyperswitch_connectors/src/connectors/fiuu/transformers.rs
```diff
@@ -1929,13 +1959,21 @@ impl TryFrom<RefundsResponseRouterData<RSync, FiuuRefundSyncResponse>>
...
+        RefundStatus::Pending => enums::RefundStatus::Pending,
+        RefundStatus::Success => enums::RefundStatus::Success,
+        RefundStatus::Rejected => enums::RefundStatus::Failure,
+        RefundStatus::Processing => enums::RefundStatus::Pending,
+        RefundStatus::Unknown => {
+            router_env::logger::warn!(
+                "Fiuu returned unknown refund sync status; previous status was {:?}, marking refund as pending",
+                prev_status
+            );
+            enums::RefundStatus::Pending
```
sai-harsha-vardhan: We can use prev status

sai-harsha-vardhan (approved):

swangi-kumari (approved):

XyneSpaces:
Review findings:

[should-fix] `crates/hyperswitch_connectors/src/connectors/fiuu.rs` – When error-response deserialization fails, the connector-event payload records only a static string and status code. Peer connectors include the raw response body, which is essential for debugging unexpected error payloads.
```rust
event.set_error(serde_json::json!({
    "error": res.response.escape_ascii().to_string(),
    "status_code": res.status_code
}));
```

[should-fix] `crates/hyperswitch_connectors/src/connectors/fiuu/transformers.rs` – The `Unknown` refund-sync branch logs `"marking refund as pending"` but returns the previous status, which may be `Success` or `Failure`. Update the log to say the status is being retained.

[should-fix] `crates/hyperswitch_connectors/src/connectors/fiuu.rs` – The empty-body guard was added only to `get_webhook_event`. If the framework calls `verify_webhook_source` or `get_webhook_source_verification_message` first on an endpoint-verification ping, those methods will still fail to parse an empty body before the event path can return `EndpointVerification`. Consider adding the same early return there.

PR #13256, opened 2026-07-09
fix(payload): harden connector against unknown enum variants and non-JSON error bodies

sai-harsha-vardhan:
## Summary
- Add `#[serde(other)] Unknown` catch-alls to Payload's `PayloadPaymentStatus`, `RefundStatus`, and `PayloadWebhooksTrigger` enums (plus a fallback landing on `AvsResponse`'s existing `Unknown` variant) so a new connector-side status/event code degrades gracefully instead of hard-failing deserialization — mirroring the pattern already merged for Adyen (#12877, #13161) and NMI (#13251).
- Unknown payment/refund statuses now retain the pre-call `AttemptStatus`/`RefundStatus` (threaded through as `prev_status` / `request.refund_status`) with a warn log, rather than assuming success or failure. An unknown post-capture-void status defaults to `Pending` (its own safe default) rather than `Failed`.
- Unknown webhook triggers resolve to `EventNotSupported` instead of failing deserialization of the whole webhook body, and are rejected as `WebhooksNotImplemented` at the object-reference/dispute-status lookup sites rather than surfacing as a hard error.
- `build_error_response` now guards against an empty body and falls back to `handle_json_response_deserialization_failure` for non-JSON error responses instead of panicking.
- `get_webhook_event_type` guards against an empty webhook body.
- Loosened two response structs with unused-but-mandatory fields: `PayloadRefundResponse.ledger` (never read anywhere) is now `Option<serde_json::Value>` instead of a strictly-typed `Vec<RefundsLedger>` (the now-unused `RefundsLedger` struct was removed), and three unused fields on `PayloadPaymentMethodResponse` (`customer_id`, `verification_status`, `status`) are now optional.
- Reviewed but not changed: dispute status is inferred purely from the webhook trigger (no separate dispute-status enum), already covered by the trigger `Unknown` handling; no UCS/hyperswitch-prism code exists in this repo; `deny_unknown_fields` is not used anywhere in this connector.

## Cypress Tests
image


XyneSpaces:
:warning: [should-fix] `#[serde(other)] Unknown` catch-all loses real connector status values

In `crates/hyperswitch_connectors/src/connectors/payload/responses.rs`, the new `Unknown` variants for `PayloadPaymentStatus`, `RefundStatus`, and `PayloadWebhooksTrigger` use `#[serde(other)]`. Any unexpected status string from Payload will be folded into `Unknown`, discarding the actual value. The transformers then retain the previous payment/refund status or treat webhooks as unsupported, which can mask real state transitions (for example, a later update arriving with a new terminal status that the mapping does not yet recognize).

Consider enumerating the known statuses explicitly, or preserve the raw string so the unexpected value is visible in logs/events. If graceful degradation is required, logging the original connector value alongside the warning would make these failures observable.

AkshayaFoiger (approved):

swangi-kumari (approved):

PR #13262, opened 2026-07-10
fix(cypress): accept 422 status for invalid currency in forex convert tests

### LINKED ISSUE #13263: fix(cypress): forex convert tests expect 400 but API returns 422 for invalid currencies
## Description

The Cypress forex tests in `00003-ForexRates.cy.js` for "should fail to convert with invalid source currency" and "should fail to convert with invalid target currency" are failing because the `convertCurrency` command in `cypress-tests/cypress/support/commands.js` only accepts HTTP 400 as a valid error status, but the API returns **422 Unprocessable Entity** for invalid currencies.

### Root Cause

In the Rust backend (`crates/router/src/core/currency.rs`), the `convert_forex` function maps `ForexCacheError::CurrencyNotAcceptable` to `ApiErrorResponse::InvalidRequestData`. The `InvalidRequestData` variant maps to `AER::Unprocessable` (`crates/hyperswitch_domain_models/src/errors/api_error_response.rs:572`), which in turn maps to `StatusCode::UNPROCESSABLE_ENTITY` (422) in `crates/api_models/src/errors/actix.rs:13`.

The test command only handles `response.status === 200` (success) and `response.status === 400` (expected error), so the 422 response falls into the `else` branch and throws an unexpected error.

### Error Response

```json
{
  "error": {
    "type": "invalid_request",
    "message": "The provided currency is not acceptable",
    "code": "IR_06"
  }
}
```

### Affected Tests

- `should fail to convert with invalid source currency`
- `should fail to convert with invalid target currency`

### Fix

Update the `convertCurrency` command to also accept `422` as a valid error response status:

```diff
-          } else if (response.status === 400) {
+          } else if (response.status === 400 || response.status === 422) {
```

### Additional Context

This was introduced in PR #12415 which added both the proper error handling in the backend and the Cypress tests in the same commit. The test author assumed the error would return 400, but `InvalidRequestData` has always mapped to 422 in the codebase's error mapping.

venkatakarthikm-maker:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

### Additional Changes

## Motivation and Context

Test 1: "should convert small amount from USD to EUR"
- amount=100, fromCurrency="USD", toCurrency="EUR"
- API returns 200 with converted_amount and currency
- Hits the if (response.status === 200) branch → asserts converted_amount and currency exist → passes

Test 2: "should fail to convert with negative amount"
- amount=-100, fromCurrency="USD", toCurrency="EUR"
- API returns 400
- Hits the else if (response.status === 400) branch → asserts response.body exists → passes

Test 3: "should fail to convert with invalid source currency"
- amount=100, fromCurrency="INVALID", toCurrency="USD"
- API returns 422 with {"error":{"type":"invalid_request","message":"The provided currency is not acceptable","code":"IR_06"}}
- Without the OR: 422 !== 200 and 422 !== 400 → falls into else → throws "Unexpected forex convert status: 422" → fails
- With the OR: 422 !== 200 but 422 === 422 → asserts response.body exists → passes

Test 4: "should fail to convert with invalid target currency"
- amount=100, fromCurrency="USD", toCurrency="INVALID"
- API returns 422 (same body as above)
- Same as Test 3 → fails without the OR, passes with it

So if we dropped 400 and only kept 422, Test 2 (negative amount) would fall into the else and throw. We genuinely need both.

## Checklist

### DIFF
--- a/cypress-tests/cypress/support/commands.js
+++ b/cypress-tests/cypress/support/commands.js
@@ -10464,7 +10464,7 @@ Cypress.Commands.add(
               "cli_log",
               `Converted ${amount} ${fromCurrency} → ${response.body.converted_amount} ${response.body.currency}`
             );
-          } else if (response.status === 400) {
+          } else if (response.status === 400 || response.status === 422) {
             expect(response.body).to.exist;
           } else {
             throw new Error(


XyneSpaces:
[should-fix] The negative-test assertion in `cypress-tests/cypress/support/commands.js` now accepts either HTTP 400 or 422 for an invalid URL path. This relaxes the contract compared to the previous exact-message check and could mask a regression if the connector starts returning 422 for malformed request paths that previously returned 400. Please pin the assertion to the status documented for this scenario, or split the expectation per test case.

Gnanasundari24 (approved):

likhinbopanna (approved):

PR #13278, opened 2026-07-13
feat(airwallex): route Airwallex via UCS (order-id threading, customer auth, Klarna PMT)

shuklatushar226:
## What
Enables the **Airwallex** connector to run through the Unified Connector Service (UCS) from the HS router. Pairs with the connector-service PR on `feat/airwallex_grace` (adds the airwallex 3DS `confirm_continue`, redirect surfacing, mandate reference, and PMT mappings on the UCS side).

## Changes
- **Route via UCS**: add `airwallex` to `ucs_only_connectors` across `config/development.toml`, `config.example.toml`, and the `env_specific`/`integration_test`/`production`/`sandbox` deployment configs.
- **Thread the CreateOrder intent id** into the UCS payment requests (airwallex is a two-step connector — CreateOrder → confirm/confirm_continue address the payment intent by id):
  - `connector_order_id` from `request.order_id` in the Authorize + ExternalVaultProxy builders.
  - `connector_order_id` from `request.connector_transaction_id` on the CompleteAuthorize leg (the 3DS `confirm_continue` continuation).
- **Connector-customer auth**: forward the already-fetched Bearer access token to `CustomerServiceCreate` via `connector_feature_data` (that request has no `state` field); the UCS airwallex customer flow reads `connector_feature_data["access_token"]` as its auth fallback.
- **Klarna PMT**: map `PaymentMethodType::Klarna → Unspecified` for CreateOrder (the grpc-client `PaymentMethodType` enum has no Klarna variant; the Authorize leg still carries the actual method via the `PaymentMethod::Klarna` oneof).

## Testing
Validated end-to-end HS → UCS → Airwallex sandbox: card (no-3DS + full 3DS via `confirm_continue`), PayPal, Skrill, Trustly, iDEAL, BLIK; Capture/Void/Refund/PSync/RSync; connector-customer create on CIT.


XyneSpaces:
[should-fix] Mapping `PaymentMethodType::Klarna` to `Unspecified` in the UCS transformer (`transformers.rs`) is a workaround for a missing gRPC enum variant. If the downstream `CreateOrder` path ever needs the exact payment method type, it will silently fall back. Consider adding the `Klarna` variant to the rust-grpc-client `PaymentMethodType` enum so the mapping is explicit rather than relying on the Authorize leg to carry it later.

crates/router/src/core/unified_connector_service/transformers.rs
```diff
@@ -2420,7 +2426,9 @@ impl
                 .shipping_cost
                 .map(|shipping_cost| shipping_cost.get_amount_as_i64()),
             authentication_data,
-            connector_feature_data: None,
+            connector_feature_data: router_data.request.order_id.as_ref().map(|order_id| {
+                Secret::new(serde_json::json!({ "connector_order_id": order_id }).to_string())
+            }),
```
AmitsinghTanwar007, line 2431: can you reverify seems some specific logic to a connector

shuklatushar226: Reverified — you're right, that was Airwallex-specific knowledge sitting in the connector-agnostic transformer. Fixed at the root rather than documented as a workaround.

**What it was:** `RecurringPaymentServiceChargeRequest` has no `connector_order_id` field in the proto — unlike `PaymentServiceAuthorizeRequest`, where this PR sets it directly at three sites. So the order id was being smuggled through `connector_feature_data` as `{"connector_order_id": "..."}`, a magic key only the UCS Airwallex impl knew how to read. Every other `connector_feature_data` site in this file is an opaque pass-through of `router_data.request.connector_meta`, so this gave one field two contracts with nothing but a string literal holding the sides together.

**Fix** — juspay/hyperswitch-prism#2028 adds the missing field (`connector_order_id = 39`), maps it into `PaymentFlowData`, and deletes the fallback read from the connector. Here it becomes `connector_order_id: router_data.request.order_id.clone()`, identical to the Authorize builders, and `connector_feature_data` goes back to `None`.

**Same treatment for the sibling case you didn't flag:** `CustomerServiceCreateRequest` was carrying the Bearer token as `{"access_token": ...}` in the same blob, because that message had no `state` field. Added `state = 10` to the proto too, so this now uses the standard `ConnectorState::foreign_from` pattern like every other builder — and the token stops travelling in a metadata blob that several code paths log as `metadata = ?response.connector_feature_data`. That also turned up a real gap: the composite authorize path had no way to pass a token to customer-create at all, which the prism PR closes.

Both are new `optional` proto tags, so it's wire-compatible in both directions.

**Ordering:** I'm holding the change on this branch until prism#2028 merges and a tag is cut — this PR pins `2026.07.07.2`, which has neither field, so pushing it now would just turn CI red. Verified locally against that tag plus the two new fields: `cargo check -p router --features v1` and `--no-default-features --features "release,v2,redis-rs"` both clean, and `cargo clippy --workspace --all-targets` clean on the prism side.

Unrelated, but it'll bite on the next rebase: this branch is ~3 weeks behind main on the UCS pin (`2026.07.07.2` vs `2026.07.28.0`). Taking main's tag will require filling in `auth_type`, `complete_authorize_url`, and `payment_channel` on the `RecurringPaymentServiceChargeRequest` literal.

PR #13279, opened 2026-07-13
fix(router): allow dashboard (JWT) access to legacy customer/payment-method routes

ShankarSinghC:
## Type of Change
- [x] Bugfix

## Description

`validate_legacy_endpoint_access` was introduced to block **merchants** from calling deprecated customer and payment-method routes once the modular payment-method service is enabled. However, the dashboard drives these same routes and authenticates via **JWT**, so it was hitting `AccessForbidden { resource: "Deprecated route" }`.

This change restricts the block to merchant (API-key) traffic only. JWT-authenticated (dashboard) requests are allowed through.

### What changed

- `validate_legacy_endpoint_access` now takes an `is_jwt: bool` and returns `Ok(())` early when the request is JWT-authenticated, before running the feature-config lookup.
- **Customer endpoints** (`create` / `retrieve` / `update` / `delete`) support JWT auth (`auth_type(ApiKeyAuth, JWTAuth, …)` and the explicit `JWTAuth` retrieve branch), so they pass the real `auth::is_jwt_auth(req.headers())` result.
- **Payment-method endpoints** (`create` / `save` / `list_customer` / `retrieve` / `update` / `delete`) have no JWT auth path (API-key / SDK / ephemeral only), so they pass a hardcoded `false` — no needless header parsing. The block behaves exactly as before for them.

`is_jwt_auth` checks for a `Bearer` authorization header or a JWT cookie — exactly how the dashboard authenticates.

## Motivation and Context

The dashboard still relies on these legacy routes; blocking them broke dashboard flows for merchants with the modular payment-method service enabled.

## How did you test it?

- `cargo check -p router` passes cleanly.
- Manual verification of auth-scheme mapping per endpoint (confirmed `check_sdk_auth_and_get_auth` never yields JWT auth, so payment-method endpoints correctly pass `false`).

#### customer entry in the db
image

#### customer retrieve api via api key
```
curl -s -w "\nHTTP %{http_code}\n" \
  "http://127.0.0.1:8080/customers/0a_cus_019f48406b347862ad15ba69cb854f9d" \
  -H "api-key: dev_TSA4yXwedCirVKhxL0bAAw7XyScj2EU3JrFvDbqGCGZD7KxVj4hpv7aH0BEvwRI9"
```
```
{"error":{"type":"invalid_request","message":"Access forbidden. Not authorized to access this resource Deprecated route","code":"IR_22"}}
HTTP 403
```
#### customer retrieve api via jwt
```
curl -s -w "\nHTTP %{http_code}\n" \
  "http://127.0.0.1:8080/customers/0a_cus_019f48406b347862ad15ba69cb854f9d" \
  -H "Authorization: Bearer eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJ1c2VyX2lkIjoidGVzdF9sb2NhbF91c2VyIiwibWVyY2hhbnRfaWQiOiJjb25uZWN0ZWRfMTc4MzYyMzg3M18yIiwicm9sZV9pZCI6Im1lcmNoYW50X2FkbWluIiwiZXhwIjoxNzg2NTI3MDE0LCJvcmdfaWQiOiJvcmdfWU1HMHRGTzZTUUNoNmU2aXpvUTciLCJwcm9maWxlX2lkIjoicHJvX1VnMlRnTWMwOGZiZHE3aFNlTWdVIiwidGVuYW50X2lkIjpudWxsfQ.wg7BpAzTpxvOghqdVsuO_qtwRXSbZG36NpZ_B-ghISM"
```
```
{"customer_id":"0a_cus_019f48406b347862ad15ba69cb854f9d","name":"ext-3ds-customer","email":"test@example.com","phone":null,"phone_country_code":null,"description":null,"address":null,"created_at":"2026-07-09T19:00:09.142Z","metadata":null,"default_payment_method_id":null,"tax_registration_id":null,"document_details":null}
HTTP 200
```

## Checklist

### DIFF
--- a/crates/router/src/core/utils.rs
+++ b/crates/router/src/core/utils.rs
@@ -121,20 +121,25 @@ pub async fn get_feature_config(
 pub async fn validate_legacy_endpoint_access<E>(
     state: &SessionState,
     platform: &domain::Platform,
-    is_jwt: bool,
 ) -> error_stack::Result<(), E>
 where
     E: From<errors::ApiErrorResponse> + error_stack::Context,
 {
+    // The dashboard authenticates via JWT and still relies on these legacy
+    // endpoints, so only merchant (API key) traffic is blocked from deprecated
+    // routes.
+    let is_dashboard_access = matches!(
+        platform.get_initiator(),
+        Some(domain::Initiator::Jwt { .. })
+    );
+
     let dimensions = dimension_state::Dimensions::new()
         .with_processor_merchant_id(platform.get_processor().get_processor_merchant_id())
         .with_provider_merchant_id(platform.get_provider().get_provider_merchant_id());
 
     let feature_config = get_feature_config(state, platform, &dimensions).await;
-    // The dashboard authenticates via JWT and still relies on these legacy
-    // endpoints, so only block merchant (API key) traffic from deprecated routes.
     common_utils::fp_utils::when(
-        feature_config.is_payment_method_modular_allowed && !is_jwt,
+        feature_config.is_payment_method_modular_allowed && !is_dashboard_access,
         || {
             Err(error_stack::report!(E::from(
                 errors::ApiErrorResponse::AccessForbidden {
--- a/crates/router/src/core/utils.rs
+++ b/crates/router/src/core/utils.rs
@@ -121,6 +121,7 @@ pub async fn get_feature_config(
 pub async fn validate_legacy_endpoint_access<E>(
     state: &SessionState,
     platform: &domain::Platform,
+    is_jwt: bool,
 ) -> error_stack::Result<(), E>
 where
     E: From<errors::ApiErrorResponse> + error_stack::Context,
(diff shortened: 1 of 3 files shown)


sai-harsha-vardhan (dismissed):

crates/router/src/core/utils.rs
```diff
@@ -121,6 +121,7 @@ pub async fn get_feature_config(
 pub async fn validate_legacy_endpoint_access<E>(
     state: &SessionState,
     platform: &domain::Platform,
+    is_jwt: bool,
```
kashif-m: We can re-use `AuthenticationType` enum in auth layer, which covers all the auth types

And have a `From` impl on it to validate if it needs to be considered as a deprecated route

sai-harsha-vardhan (approved):

kashif-m (approved):

PR #13285, opened 2026-07-13
feat(deja): [1/5] optional dependency, feature flags, typed settings + feature-off purity proof

maverox:
Closes #13301

# feat(deja) [1/5] — foundation: optional dependency, feature flags, typed settings + feature-off purity proof

## TL;DR
This is **part 1 of a 5-PR stack** that re-slices #13237 (the feature-gated **déjà** record/replay instrumentation) into small, independently-reviewable layers. **Same net change as #13237**, just split so each PR tells one story and can be reviewed on its own. #13237 stays open as the reference.

This PR adds **only** the dependency + feature-flag + typed-config surface and the test that proves it's inert when the feature is off. It contains **no instrumentation and no runtime behavior**.

---

## About this stack (read once)

**What déjà is.** [Déjà](https://github.com/juspay/deja) is a deterministic **record/replay harness for service boundaries**. With `--features deja` and `[deja] mode = "record"`, every storage / cache / crypto / id / time / HTTP boundary call emits a structured `BoundaryEvent` that is published to Kafka and landed in object storage. A separate harness then replays a recorded request stream against a candidate build and scores divergences — a regression gate that catches behavior changes the HTTP response alone wouldn't reveal (a dropped cache write, an altered DB insert payload, a different nonce).

**Why it's split.** #13237 is one 80-file, ~5.9k-line change. That's very hard to review in one pass. This stack breaks it into 5 layers along **boundary class**, each of which compiles and can be reasoned about independently:

| Part | PR | Scope | ~size |
|------|----|-------|-------|
| 1/5 | **this** | optional dep + feature flags + typed settings + feature-off purity proof | 20 files, +725 |
| 2/5 | #13286 | per-request correlation substrate + ingress boundary + clock/id seams | 7 files, +1213 |
| 3/5 | #13287 | database read/write capture + per-correlation replay routing | 32 files, +1154 |
| 4/5 | #13288 | external I/O + crypto capture (redis, HTTP, RNG/nonce/hash) | 13 files, +1669 |
| 5/5 | #13289 | Kafka sink + boot install + Superposition sampler + enable in release | 8 files, +1106 |

**The invariant every PR upholds.** Each PR compiles **both** with the feature off (the default — and byte-identical to today) **and** with `--features deja`. Nothing records at runtime until PR5 installs the boot hook, so even feature-**on**, PRs 2–4 are inert (every boundary is a pass-through) until the very last PR flips it on.

[...]

PR #13286, opened 2026-07-13
feat(deja): [2/5] per-request correlation substrate + ingress boundary + clock/id seams

### LINKED ISSUE #13302: [deja] ART stacked split 2/5 — correlation substrate + ingress boundary + clock/id seams
Tracking issue for **part 2 of 5** of the déjà record/replay instrumentation, split out of the monolithic #13237 into a reviewable stacked series.

**Resolved by:** #13286
**ART "PR review" track:** juspay/hyperswitch-cloud#20222 &nbsp;(epic: juspay/hyperswitch-cloud#20220)

### Scope — correlation substrate + ingress boundary + clock/id seams
- Request-correlation substrate (`request_id` propagation across stores).
- The ingress boundary + deterministic clock / id seams that recording hangs off.

### Guarantee
Compiles both **feature-off** (default; byte-identical to today) and **feature-on** (`--features deja`).

**Stack (review top-down):** #13285 ← #13286 ← #13287 ← #13288 ← #13289

maverox:
Closes #13302

# feat(deja) [2/5] — per-request correlation substrate + ingress boundary + clock/id seams

## TL;DR
**Part 2 of a 5-PR stack** re-slicing #13237 (feature-gated **déjà** record/replay instrumentation) for review. **Base this PR on `deja/pr1-foundation` (#13285, merged).** This layer adds the runtime **identity substrate** every downstream boundary anchors to. All deja surface is `#[cfg(feature = "deja")]`; compiles feature-off (byte-identical) and `--features deja`; runtime-inert until PR5 installs the hook.

## The stack
1. #13285 — foundation (dep + feature flags + typed settings + purity proof) ← **full context lives here**
2. **this (#13286)** — correlation substrate + ingress boundary + clock/id seams
3. #13287 — DB capture + per-correlation replay routing
4. #13288 — external I/O + crypto capture
5. #13289 — Kafka sink + boot install + sampler + enable in release

> **What déjà is:** a deterministic record/replay harness for service boundaries — with the feature + record mode, each boundary emits a structured event to Kafka; a harness replays a recorded request stream against a candidate build and scores divergences. See #13285 for the full description and the split rationale.

---

## What's in THIS PR (part 2)

This is the "stable identity + logical clock" layer. An ingress request gets a correlation id; that id is mirrored into deja-context so spawned tasks inherit it; and every clock read / generated id becomes a recorded seam so replay reproduces the exact recorded value.

### `router_env/src/request_id.rs` (the bulk, ~960 lines)
- A `boundary` module implementing the **`http_incoming` ingress boundary**: `RecordingBody` body-buffering, a `RecordedIncomingFuture` that captures/replays the incoming request, and the anchor point where a request's correlation id is established.
- A per-request **recording-sampler hook** (`RequestRecordingSampler` trait + `with_recording_sampler` builder + `RequestRecordingFacts`) — the seam PR5's Superposition sampler plugs into.
- The per-request recording decision is cleared by an **RAII guard**, not a straight-line call after the handler: a cancelled or panicking record-mode request drops its future mid-flight and would otherwise skip the trailing clear and leak its decision-registry entry forever. The guard covers the normal return, the `?` early-return, and unwind/future-drop alike.
- A `uuid_v7` **determinism seam** (record/substitute the generated id).

### `router_env/src/logger/setup.rs`
- Installs two tracing layers on the subscriber — but only when the process is actually recording or replaying. Both installs key off the **process runtime mode** (a pure peek at the installed hook; never a per-request decision), so a `mode = disabled` process runs the exact pre-existing subscriber chain: the correlation layer (mirrors the ingress `request_id` span field into deja-context so boundary events fired from spawned tasks inherit the request correlation) and the execution-graph layer are simply absent.

### Clock / id determinism seams
- `common_utils/src/lib.rs` — gated `deja::time` boundaries on `date_time::now` / `now_unix_timestamp` / `date_as_yyyymmddthhmmssmmmz` / `now_rfc7231_http_date`, and gated `deja::id` boundaries on the nanoid generators.
- `router/src/utils.rs` (router-side nanoid), `router/src/types/domain/user.rs` (`generate_user_id`), `router/src/core/admin.rs` (merchant publishable key) — gated `deja::id` seams.

### Fire-and-forget lineage
- `router/src/core/payments/operations/payment_response.rs` — the two background-task sites (`save_payment_method`, `update_gateway_score`) spawn via `deja::spawn_fork` under the feature (`tokio::spawn` otherwise). **`spawn_fork` is a pure lineage stamp** (task/bucket provenance on emitted events): recording never defers, joins, or reschedules anything.

## The pinned library revision
This PR pins `juspay/deja @ 2ae840aa2b0a5160dc1719c83ad6d3c6bbcfa716`. Two library properties matter to this layer and were verified empirically on that revision:

- **The boundary wrapper hoists a single "is observation active?" gate above all capture work.** With the feature compiled in but no hook installed (the production `mode = disabled` state), a boundary adds **~1.5–2.0 ns** over the un-instrumented function (release build, pinned-core medians). A record-mode request that the sampler *skips* adds **~11 ns** — one thread-local decision read — and records nothing (writer stats `attempted=0`, zero tape lines).
- **Capture preparation runs inside a panic firewall.** A panicking argument/state-key expression degrades that one event's capture; the request itself always runs the original body.

## Feature-off guarantee — byte-identical, hardened during the split
All deja surface is gated. **Three spots that were unconditional in #13237 are gated here** so the layer is genuinely inert when off:
- **Tracing layers** — added only under the feature; the subscriber chain is cfg-shadowed, so a feature-off subscriber is exactly what it was before.
- **`#[track_caller]`** (~20 sites across `request_id.rs` + `common_utils/lib.rs`) — changed to `#[cfg_attr(feature = "deja", track_caller)]` so feature-off codegen is unchanged.
- **`Debug` derives** — `RequestIdentifier` / `RequestIdMiddleware` keep their derived `Debug` feature-off; the manual impls exist only under the feature, where the non-`Debug` sampler field is present.

Build proof on this branch: a feature-off release `router` binary contains **zero** deja code paths (`strings` scan for deja source-path markers = 0; feature-on the same scan finds them).

The `generate_user_id` and `payment_response` future extractions are kept: they're the minimal structure needed to attach a gated boundary / select the spawn function, and are behavior-identical feature-off.

## How to review
- The three gating fixes above are the only things that could perturb feature-off — confirm each.
- The ingress-boundary body buffering (`RecordingBody` / `RecordedIncomingFuture`).
- The RAII decision guard (leak-on-cancellation is the failure mode it closes).
- The sampler API (`RequestRecordingSampler` / `with_recording_sampler`) — its concrete impl arrives in PR5.

## Verification
- `cargo check -p router` (feature-off) ✅ · `cargo check -p router --features deja` ✅
- Feature-off release binary: zero deja code strings ✅
- End-to-end on the integrated branch (all five parts + the pinned library): record → replay of a 9-request payment workload at concurrency 4 reproduces **7/7 correlations with zero side-effect / value / body divergences**; injected faults (a changed DB write value, a silently dropped cache write) each surface as explicit divergences.


crates/router_env/src/request_id.rs
```diff
@@ -36,15 +36,445 @@ use std::{
...
+        hook().map(Arc::as_ref).is_some_and(DejaHook::is_active)
+    }
+
+    pub(super) fn is_record_mode() -> bool {
+        hook()
+            .map(Arc::as_ref)
+            .is_some_and(|hook| hook.mode().is_record())
+    }
+
+    pub(super) fn record_id_generation(
```
jarnura, line 104: Make all serde values as secret.

crates/router_env/src/request_id.rs
```diff
@@ -36,15 +36,445 @@ use std::{
...
+    /// In replay mode the active hook is a lookup-table hook; this returns the
+    /// recorded id for this call site so the candidate reproduces the recorded
+    /// run byte-for-byte instead of minting a fresh uuid. The hook emits an
+    /// `ObservedCall` either way (hit or miss) for the divergence scorer.
+    ///
+    /// Returns `None` when recording (so the caller generates live) or on a
+    /// replay miss (the scored divergence; the caller then generates live too).
+    /// The recorded result shape mirrors `record_id_generation`'s `response`:
+    /// `{"generated_value": "<id>"}`.
+    pub(super) fn replay_id_generation(
```
jarnura, line 141: Make all serde values as secret.

crates/router_env/src/request_id.rs
```diff
@@ -36,15 +36,445 @@ use std::{
...
+            caller_location: Some(caller),
+        })?;
+        recorded
+            .get("generated_value")
+            .and_then(serde_json::Value::as_str)
+            .map(ToOwned::to_owned)
+    }
+
+    #[derive(Debug)]
+    pub(super) struct IncomingHttpRecord {
```
jarnura, line 168: Make all serde values as secret.

jarnura (changes requested):

maverox:
Please add the required configs in sandbox.toml files as well

jarnura (dismissed):

ThisIsMani (dismissed):
Dashboard specific changes looks good to me.

ThisIsMani (dismissed):

jarnura (dismissed):

crates/router_env/src/request_id.rs
```diff
@@ -36,15 +36,465 @@ use std::{
...
+        deja::http::headers(headers.iter().map(|(name, value)| {
+            let value = value
+                .to_str()
+                .map(str::to_string)
+                .unwrap_or_else(|_| format!("{value:?}"));
+            (name.as_str().to_string(), value)
+        }))
+    }
+
+    fn body_json(bytes: &Bytes) -> serde_json::Value {
```
gopikrishna000, line 248: is this expected., to send raw http body?

maverox: yeah bro for now, as anyways this will only be enabled in sbx, solving the encryption part is deferred before we enable it on prod

gopikrishna000 (approved):

ThisIsMani (approved):

ShankarSinghC (approved):

jarnura (approved):

PR #13288, opened 2026-07-13
feat(deja): [4/5] external I/O + crypto capture (redis, HTTP, RNG/nonce/hash)

### LINKED ISSUE #13304: [deja] ART stacked split 4/5 — redis + HTTP + crypto capture
Tracking issue for **part 4 of 5** of the déjà record/replay instrumentation, split out of the monolithic #13237 into a reviewable stacked series.

**Resolved by:** #13288
**ART "PR review" track:** juspay/hyperswitch-cloud#20222 &nbsp;(epic: juspay/hyperswitch-cloud#20220)

### Scope — redis + HTTP + crypto capture
- Redis boundary capture, HTTP egress capture, and crypto boundary capture.

### Guarantee
Compiles both **feature-off** (default; byte-identical to today) and **feature-on** (`--features deja`).

**Stack (review top-down):** #13285 ← #13286 ← #13287 ← #13288 ← #13289

maverox:
Closes #13304

# feat(deja) [4/5] — external I/O + crypto capture (redis, outbound HTTP, in-memory cache, RNG/nonce/hash)

## TL;DR
**Part 4 of a 5-PR stack** re-slicing #13237 (feature-gated **déjà** record/replay instrumentation) for review. **Base this PR on `deja/pr3-db-capture` (#13287).** This layer instruments the second big boundary class — cache, egress, and cryptographic nondeterminism. Compiles feature-off (byte-identical) and `--features deja`; inert until PR5 installs the hook.

## The stack
1. #13285 — foundation ← **full context** · 2. #13286 — correlation substrate · 3. #13287 — DB capture · **4. this (#13288)** — I/O + crypto capture · 5. #13289 — sink/boot.

> **What déjà is:** a deterministic record/replay harness for service boundaries; see #13285.

---

### Redis command boundaries + typed wire codec
- `redis_interface/src/module/fred/commands.rs` and `module/redis_rs/commands.rs` — instrument the redis command surface on **both** backends as record/replay boundaries with a typed value codec (a `DejaRedisValue` proxy that round-trips `RedisValue`/`redis::Value`), explicit **multi-key read attribution**, and **SADD write attribution** (the boundary sits on `sadd` itself, with its own key attribution). Correlation-namespaced `add_prefix`.
- **Typed error round-trip**: all 22 boundary sites (15 fred + 7 redis-rs) use a typed `ResultCodec<_, RedisError>` — a recording that threw (e.g. `NotFound`, `SetNxFailed`) replays as the **same `RedisError` context** instead of an unreconstructable sentinel. `RedisError` gains feature-gated serde derives.
- `redis_interface/src/types.rs` — `RedisKey::as_str` + gated reply serde derives. `redis_interface/src/metrics.rs` — inverse-gates the `SerializeAndSetKeyWithExpiry` metric variant. `storage_impl/src/redis/kv_store.rs` — a `DejaKvValue` codec-capability marker (inert blanket-impl when off).

### In-memory-cache lookup boundary — `storage_impl/src/redis/cache.rs`
- The process-global moka L1 cache would break per-correlation replay isolation: a warm hit silently skips the instrumented redis/db boundary underneath, so two replayed requests could see each other's cached values. The L1 **lookup itself** becomes a boundary whose `Option` outcome is substituted on replay — neutral during recording, isolated during replay, and sequential (no orphaned downstream events).

### Outbound HTTP boundary
- `external_services/src/http_client/boundary.rs` (new, whole-file gated) + `http_client.rs` — a record/replay boundary around the outbound `send_request` path, with **eager body capture** for faithful replay. Captured request/response payloads are **`Secret`-wrapped** so they cannot leak through Debug/log formatting on the capture path.

### Cryptographic nondeterminism seams
- `common_utils/src/crypto.rs` — AEAD **nonce** bytes + CSPRNG (the ciphertext-reproducibility keystone).
- `router/src/utils/user/password.rs` — Argon2 hash. `router/src/services.rs` — AES-256 key. `router/src/services/jwt.rs` — JWT expiry. All gated `deja::id`/`deja::time` seams.
- `hyperswitch_domain_models/src/type_encryption.rs`, `router/src/core/api_locking.rs` — explanatory comments only (non-boundary notes).

## The pinned library revision
This PR pins `juspay/deja @ 2ae840aa2b0a5160dc1719c83ad6d3c6bbcfa716`. The library-side guarantees that matter here (measured on that revision): a boundary with the feature on but no hook installed costs **~1.5–2.0 ns**; a sampled-out request costs **~11 ns** and records nothing; all capture preparation runs inside a panic firewall, so a panicking capture expression can never fail the request.

## Feature-off guarantee
Fully gated, plus a few **behavior-preserving** ungated restructures that lower identically off (kept as the minimal structure for a gated boundary): the `add_prefix` restructure in both `commands.rs`, the response-`match` restructure in `http_client.rs`, and the `generate_password_hash_inner` split in `password.rs`.

**`fred/commands.rs` fix:** #13237 **ungated-replaced** `set_hash_fields` / `set_hash_field_if_not_exist`'s call to the instrumented `self.set_expiry(...)` with a raw `pool.expire`, which drops `set_expiry`'s `#[instrument]` DEBUG span in default builds. Restored: feature-off now calls `set_expiry` exactly as `main`; feature-on keeps the raw inline (needed so the EXPIRE isn't orphaned under the method's own replay boundary — a pure replay concern).

Build proof on this branch: a feature-off release `router` binary contains **zero** deja code paths (`strings` scan for deja source-path markers = 0).

## How to review
- **Redis proxy round-trip losslessness** — `into_supported_redis_value` can `Err` on exotic values (BigNumber/ServerError); check that a decodable read doesn't become a `GetFailed` under replay.
- The **typed `RedisError` round-trip** — error-path recordings replay as errors of the same shape (unit variants as bare strings; `InvalidConfiguration` keeps its payload).
- The **nonce byte-serde** — this is what makes recorded ciphertext reproducible on replay.
- The **in-memory-cache boundary** — the isolation argument (a warm L1 hit must not bypass replay).
- **HTTP full-body buffering** as a streaming/perf consideration (only active when the hook is installed).
- The `DejaKvValue` bound on `kv_wrapper` `T` (a feature-on `Serialize` requirement) — verify no caller type is left out.

## Verification
- `cargo check -p router` (feature-off) ✅ · `cargo check -p router --features deja` ✅
- Feature-off release binary: zero deja code strings ✅
- End-to-end on the integrated branch (all five parts + the pinned library): record → replay at concurrency 4 reproduces 7/7 correlations with zero divergences — including redis get/set/del/setnx, in-memory-cache hits, and AES/GCM crypto events on the tape; an injected silently-dropped cache write surfaces as an explicit divergence.

### DIFF
--- a/crates/external_services/src/http_client/semantic_boundary.rs
+++ b/crates/external_services/src/http_client/semantic_boundary.rs
@@ -5,7 +5,7 @@ use common_utils::{
 use deja::DejaHook;
 use error_stack::ResultExt;
 use hyperswitch_interfaces::errors::HttpClientError;
-use hyperswitch_masking::{Maskable, PeekInterface};
+use hyperswitch_masking::{ExposeInterface, Maskable, PeekInterface, Secret};
 use reqwest::ResponseBuilderExt;
 use serde_json::json;
 
@@ -64,7 +64,7 @@ pub(super) fn request_id(request: &Request) -> Option<String> {
     })
 }
 
-pub(super) fn request_args(request: &Request, timeout_secs: Option<u64>) -> serde_json::Value {
+pub(super) fn request_args(request: &Request, timeout_secs: Option<u64>) -> Secret<serde_json::Value> {
     // Header storage iterates in non-deterministic (HashMap) order, so the raw
     // sequence differs between record and replay even when the header SET is
     // identical. The args matcher compares serialized JSON arrays
@@ -82,7 +82,7 @@ pub(super) fn request_args(request: &Request, timeout_secs: Option<u64>) -> serd
         .into_iter()
         .map(|(key, value)| json!({ "key": key, "value": value }))
         .collect();
-    json!({
+    Secret::new(json!({
         "method": format!("{:?}", request.method),
         "url": request.url.as_str(),
         "request_id": request_id(request),
@@ -95,17 +95,21 @@ pub(super) fn request_args(request: &Request, timeout_secs: Option<u64>) -> serd
             "certificate_key": request.certificate_key.is_some(),
             "ca_certificate": request.ca_certificate.is_some(),
         },
+    }))
+}
+
+// Captured payloads travel as `Secret<serde_json::Value>` between helpers so
+// their `Debug` output is redacted (response bodies/headers can carry PII or
+// credentials); they are `.expose()`d only at the deja handoff, where the tape
+// keeps full fidelity.
+fn captured_body_json(response: &reqwest::Response) -> Secret<serde_json::Value> {
+    Secret::new(match response.extensions().get::<CapturedResponseBody>() {
+        Some(CapturedResponseBody(bytes)) => deja::http::body(bytes),
+        None => deja::http::missing_body("response body not captured (missing extension)"),
     })
 }
 
-fn captured_body_json(response: &reqwest::Response) -> serde_json::Value {
-    match response.extensions().get::<CapturedResponseBody>() {
-        Some(CapturedResponseBody(bytes)) => deja::http::body(bytes),
-        None => deja::http::missing_body("response body not captured (missing extension)"),
-    }
-}
-
-fn response_headers_json(response: &reqwest::Response) -> serde_json::Value {
+fn response_headers_json(response: &reqwest::Response) -> Secret<serde_json::Value> {
     let mut map = serde_json::Map::new();
     for (key, value) in response.headers() {
         if let Ok(value) = value.to_str() {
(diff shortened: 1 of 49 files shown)


crates/external_services/src/http_client/semantic_boundary.rs
```diff
@@ -0,0 +1,222 @@
...
+        "request_body": request.body.as_ref().map(request_body),
+        "client_tls": {
+            "certificate": request.certificate.is_some(),
+            "certificate_key": request.certificate_key.is_some(),
+            "ca_certificate": request.ca_certificate.is_some(),
+        },
+    })
+}
+
+fn captured_body_json(response: &reqwest::Response) -> serde_json::Value {
```
jarnura: All these function are not recorded and need not be serde_json value, make it secret.

jarnura: Other places too

jarnura (changes requested):

jarnura (approved):

PR #13291, opened 2026-07-13
Emit request-scoped events for Redis calls

### LINKED ISSUE #13090: feat(observability): capture per-command Redis call latency and errors as observable events

ThisIsMani:
**Redis calls can now emit external-service events correlated with the request that triggered them.** Request identity is copied into a lightweight command handle, keeping physical Redis resources shared and request-agnostic while avoiding mutable context on the startup-scoped pool.

### How context flows

```
HTTP request
  └─▶ SessionState::add_request_id
        ├─▶ tenant store
        └─▶ global store
              └─▶ RedisConnectionPool + ExecutionContext
                    └─▶ RedisConnection (owned request-id snapshot)
                          └─▶ track_redis_call
                                └─▶ ExternalServiceCall
```

The command handle is the boundary between lifecycle and execution concerns:

- `RedisConnectionPool` retains initialization, health monitoring, pub/sub resources, and shared backend clients.
- `RedisConnection` performs commands with an optional request ID and event emitter.
- `RouterStore` and `KVRouterStore` provide the reusable `ExecutionContext` contract.
- Both `redis-rs` and `fred` emit operation, latency, success, timestamp, and request ID without exposing Redis keys or values.
- Startup and background work use context-free handles and therefore do not emit request-scoped events.

> **Validation is intentionally pending.** Formatting was applied, but compilation, Cargo checks, and tests were not run at the requester’s direction. This remains a draft until both Redis backends are validated.

_Generated by Pi (model `unknown`)._


crates/common_utils/src/external_service.rs
```diff
@@ -28,6 +28,15 @@ pub struct ExternalServiceCall {
...
     fn emit_external_service_call(&self, event: ExternalServiceCall);
+
+    /// Whether this emitter actually emits events. Defaults to `true`.
+    ///
+    /// `NoOpEventEmitter` returns `false`, letting hot paths (notably Redis,
+    /// called far more often than any HTTP service) skip event construction and
+    /// the request-id lookup entirely when emission is disabled.
+    fn is_enabled(&self) -> bool {
+        true
+    }
```
ThisIsMani, line 39: Do we need this?

crates/common_utils/src/request_id_context.rs
ThisIsMani, line 1: Maybe RequestContext is a better name... We might add things related to request later on...

crates/common_utils/src/request_id_context.rs
```diff
@@ -0,0 +1,10 @@
+//! Request correlation context shared by infrastructure clients.
+
+/// Minimal request correlation context for infrastructure calls.
+///
+/// This intentionally avoids the name `Context` to prevent confusion with
+/// `error_stack::Context`, and carries the request ID only.
+pub trait RequestIdContext: Send + Sync {
```
ThisIsMani, line 7: This as well.

crates/drainer/src/health_check.rs
```diff
@@ -162,7 +162,7 @@ impl HealthCheckInterface for Store {
         &self,
         _conf: &Settings,
     ) -> CustomResult<(), HealthCheckRedisError> {
-        let redis_conn = self.redis_conn.clone();
+        let redis_conn = self.redis_conn.get_connection();
```
ThisIsMani, line 165: You can get proper redis connection since this is implemented on store.

crates/drainer/src/stream.rs
```diff
@@ -29,6 +29,7 @@ impl Store {
 
         match self
             .redis_conn
+            .get_connection()
```
ThisIsMani, line 32: Here as well I believe... It's better not to use get_connection() since it's basically a backdoor...

ThisIsMani: Please check the other places as well... And if this is feasible, we can make redis_conn private field to avoid confusion.

crates/redis_interface/src/metrics.rs
```diff
@@ -1,5 +1,17 @@
...
+//! `track_redis_call` wraps every raw Redis client future — i.e. every actual
+//! network roundtrip. It records the OTel latency histogram (gated behind the
+//! `metrics` feature) and, independently, emits one `ExternalServiceCall` event
+//! per roundtrip reflecting that exact command's latency and outcome.
+//!
+//! Emitting at the roundtrip layer (rather than once per logical command
+//! method) means multi-roundtrip methods — multitenancy-fallback retries,
+//! `HSET` + `EXPIRE`, cluster parallel `GET`, `SCAN`/`HSCAN` cursor pages —
+//! each produce one event per real roundtrip with correct wall-clock latency,
+//! instead of a single event whose latency sums or hides the constituent calls.
```
ThisIsMani, line 12: I think we can remove these from here... Some of these can be placed at track_redis_call maybe...

PR #13300, opened 2026-07-14
feat(router): add Prop, PrivateLabel, and Dinacard card networks

ayush22667:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

Adds three new `CardNetwork` variants — `Prop` (proprietary), `PrivateLabel` (store/closed-loop cards), and `Dinacard` (Serbia's national scheme) — sourced from BIN enrichment. They deserialize from the enrichment source via `PROP`, `PRIVATE LABEL`, and `DINACARD` serde aliases, and are threaded through every exhaustive `CardNetwork` match in the codebase.

### Additional Changes

## Motivation and Context

BIN enrichment can report these networks; without dedicated variants they cannot be represented or routed on. Adding them lets the system parse and carry the correct network end-to-end rather than losing the classification.

## Linked Issues

- Issue: https://github.com/juspay/hyperswitch-cloud/issues/21527

## How did you test it?

`cargo check` on the affected crates (`common_enums`, `kgraph_utils`, `hyperswitch_connectors`, `router`) passes. The changes are exhaustive-match additions and enum classification; no behavioural change for existing networks.

## Checklist


XyneSpaces:
[should-fix] `is_signature_network` classification mismatch
The PR description states the new networks are classified as non-signature (`is_signature_network = false`), but the diff adds `Prop`, `PrivateLabel`, and `Dinacard` to the `true` arm in `crates/common_enums/src/enums.rs`. This will cause them to be treated as signature networks by routing/fee logic. Please move these variants into the `false` arm.

[should-fix] PeachPayments mapping for unsupported networks
The description says PeachPayments drops the scheme field for unsupported networks, yet the diff still adds the three new variants to `CardNetworkLowercase` and maps them via `From<CardNetwork>` in `crates/hyperswitch_connectors/src/connectors/peachpayments/transformers.rs`. Payments with these networks routed to PeachPayments will send a scheme value the connector likely does not recognize. Please remove these arms or convert the mapping to an `Option`-returning helper so unsupported networks omit the scheme.

hrithikesh026 (approved):

swangi-kumari (approved):

ShankarSinghC (approved):

PR #13313, opened 2026-07-14
ci(cypress-tests-runner): check for 'S-test-ready' label at runtime

SanchithHegde:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [x] CI/CD

## Description

This PR updates the `S-test-ready` label detection in the Cypress workflow to happen at runtime, so that users can retry the Cypress tests alone after adding the label. As of today, if users want to re-trigger the check, they are required to push a blank commit, which would unfortunately re-trigger all checks on commit push. The changes in this PR removes that requirement.

## Motivation and Context

This should help avoid unnecessary workflow runs with respect to triggering Cypress checks.

## How did you test it?

Initially, hadn't attached the `S-test-ready` label to the PR. I attached the label and retried the job from the UI, was able to re-trigger the check without having to push new commits.

## Checklist


XyneSpaces:
[should-fix] The `check-cypress-prerequisites` job exits with code 1 when the `S-test-ready` label is absent. This causes every unlabeled PR to show a failing check, which is noisy and can interact badly with branch protection rules.

Return success after setting `run-checks=false` so the absence of the label is treated as a skip rather than a failure.

.github/workflows/cypress-tests-runner.yml
```diff
@@ -97,11 +97,49 @@ jobs:
...
+          if [[ "${{ github.event_name }}" == "merge_group" ]]; then
+            echo "run-checks=true" >> "${GITHUB_OUTPUT}"
+            exit 0
+          fi
+
+          HEAD_REPO="${{ github.event.pull_request.head.repo.full_name }}"
+          BASE_REPO="${{ github.event.pull_request.base.repo.full_name }}"
+          # Repository is a fork, ignore checks 
+          if [[ "${HEAD_REPO}" != "${BASE_REPO}" ]]; then
+            echo "run-checks=true" >> "${GITHUB_OUTPUT}"
```
pixincreate, line 124: shouldn't this be `false`?

SanchithHegde: The `RUN_TESTS` based check (in the later jobs) would succeed the check in case of forks anyway. Didn't see much point in skipping checks for forks only when the label is present.

pixincreate (approved):

NISHANTH1221 (approved):

sai-harsha-vardhan (approved):

PR #13334, opened 2026-07-15
perf(core): Fetch MCA list by profile_id

AnuthaDev:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [x] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

Fetch merchant connector accounts (MCAs) filtered by `(merchant_id, profile_id)` at the database level, instead of fetching every MCA belonging to a merchant and filtering by profile in memory.

Adds two storage methods:

- `list_merchant_connector_accounts_without_encrypted_including_disabled_by_merchant_id_profile_id`
- `list_enabled_merchant_connector_accounts_without_encrypted_by_merchant_id_profile_id` — same, plus `disabled = false`

Call sites migrated: payment session (v1 and v2), Apple Pay retryable connectors, allowed-payment-method-type validation, cgraph cache refresh, `get_active_mca_ids`, payout links, superposition SDK config, PM-auth validation, routing config validation, merchant-enabled-PM context, and customer payment method listing.

Since profile filtering now happens in SQL, the in-memory helpers shrink accordingly:

- `filter_based_on_profile_and_connector_type` removed (v1) / renamed to `filter_by_connector_type` (v2)
- `is_merchant_connector_account_id_in_connector_mandate_details` drops its `profile_id` parameter and internal disabled/profile filtering — callers now pass an already enabled- and profile-scoped list
- `get_active_mca_ids` gains a `profile_id` parameter
- `get_mca_status` takes `Option<&MerchantConnectorAccounts>`, which is `None` when the intent has no `profile_id` — preserving the previous "no mandate match" behavior in that case

**Two behavior changes worth reviewer attention:**

1. `get_active_mca_ids` was previously merchant-wide, so an MCA belonging to a *different* profile counted as "active" during cgraph filtering — even though the cgraph itself is already profile-scoped. It is now profile-scoped, making the two consistent.
2. PM-auth validation given an `mca_id` from another profile now returns `"payment method auth connector account not found"` rather than `"payment method auth profile_id differs from connector profile_id"`. The request is still rejected; only the diagnostic is less specific.

No new index is required — the existing `UNIQUE (profile_id, connector_label)` constraint provides a `profile_id`-leading index in both v1 and v2.

### Additional Changes

## Motivation and Context

Hot paths — payment session, routing, payment method listing — fetched every MCA under a merchant and discarded most of them in memory. MCA rows carry large `payment_methods_enabled` and `metadata` JSON columns, so for merchants with many profiles this transferred and deserialized substantially more data than needed on every request. Pushing the `profile_id` predicate into SQL cuts those rows at the database.

It also removes a latent failure mode in v1: `MerchantConnectorAccountWithoutEncrypted::try_from` errors with `MissingRequiredField { profile_id }` on a NULL `profile_id`, and `RouterStore` collects into `Result<Vec<_>>` — so a single legacy MCA with a NULL `profile_id` anywhere under a merchant would fail the *entire* fetch. `WHERE profile_id = $1` excludes such rows before conversion.

## Checklist

### DIFF
--- a/crates/diesel_models/src/query/merchant_connector_account.rs
+++ b/crates/diesel_models/src/query/merchant_connector_account.rs
@@ -310,7 +310,12 @@ impl MerchantConnectorAccount {
         merchant_id: &common_utils::id_type::MerchantId,
         profile_id: &common_utils::id_type::ProfileId,
     ) -> StorageResult<Vec<Self>> {
-        generics::generic_filter::<<Self as HasTable>::Table, _, _, _>(
+        generics::generic_filter::<
+            <Self as HasTable>::Table,
+            _,
+            <<Self as HasTable>::Table as Table>::PrimaryKey,
+            _,
+        >(
             conn,
             dsl::merchant_id
                 .eq(merchant_id.to_owned())
--- a/crates/diesel_models/src/query/merchant_connector_account.rs
+++ b/crates/diesel_models/src/query/merchant_connector_account.rs
@@ -170,6 +170,51 @@ impl MerchantConnectorAccount {
         )
         .await
     }
+
+    pub async fn list_merchant_connector_accounts_without_encrypted_including_disabled_by_merchant_id_profile_id(
+        conn: &PgPooledConn,
+        merchant_id: &common_utils::id_type::MerchantId,
+        profile_id: &common_utils::id_type::ProfileId,
+    ) -> StorageResult<Vec<Self>> {
+        generics::generic_filter::<
+            <Self as HasTable>::Table,
+            _,
+            <<Self as HasTable>::Table as Table>::PrimaryKey,
+            _,
+        >(
+            conn,
+            dsl::merchant_id
+                .eq(merchant_id.to_owned())
+                .and(dsl::profile_id.eq(profile_id.to_owned())),
+            None,
+            None,
+            None,
+        )
+        .await
+    }
+
+    pub async fn list_enabled_merchant_connector_accounts_without_encrypted_by_merchant_id_profile_id(
+        conn: &PgPooledConn,
+        merchant_id: &common_utils::id_type::MerchantId,
+        profile_id: &common_utils::id_type::ProfileId,
+    ) -> StorageResult<Vec<Self>> {
+        generics::generic_filter::<
+            <Self as HasTable>::Table,
+            _,
+            <<Self as HasTable>::Table as Table>::PrimaryKey,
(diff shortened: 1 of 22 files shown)


crates/diesel_models/src/query/merchant_connector_account.rs
```diff
@@ -170,6 +170,51 @@ impl MerchantConnectorAccount {
...
+        generics::generic_filter::<
+            <Self as HasTable>::Table,
+            _,
+            <<Self as HasTable>::Table as Table>::PrimaryKey,
+            _,
+        >(
+            conn,
+            dsl::merchant_id
+                .eq(merchant_id.to_owned())
+                .and(dsl::profile_id.eq(profile_id.to_owned())),
```
hrithikesh026, line 188: please create respective db indexes

crates/router/src/core/payment_methods/cards.rs
```diff
@@ -5781,21 +5781,30 @@ pub async fn list_customer_payment_method(
...
-        .change_context(errors::ApiErrorResponse::MerchantConnectorAccountNotFound {
-            id: platform
-                .get_processor()
-                .get_account()
-                .get_id()
-                .get_string_repr()
-                .to_owned(),
-        })?;
+    // The MCA list is used to evaluate if the payment method has connector_mandate_details for any active MCA
+    let merchant_connector_accounts = if let Some(profile_id) = profile_id {
```
hrithikesh026: profile_id can be found in business_profile that is fetched before.

crates/diesel_models/src/query/merchant_connector_account.rs
AnuthaDev: There is already a DB index on `merchant_id` for MCA

Sakilmostak (dismissed):

crates/diesel_models/src/query/merchant_connector_account.rs
hrithikesh026: an index on merchant_id alone won't help in this query right.

AnuthaDev: added composite index

hrithikesh026 (approved):

Sakilmostak (approved):

prajjwalkumar17 (approved):

PR #13335, opened 2026-07-15
refactor(router): separate DB pool configs for master, accounts, and global schemas

sai-harsha-vardhan:
## Summary
- Previously the accounts schema (merchant accounts, profiles, MCAs, key stores) and the global schema (users, roles) both reused `master_database`'s pool config, coupling their connection-pool sizing to regular per-tenant payment traffic even though they have very different access patterns.
- Adds dedicated `accounts_database` and `global_database` config blocks for the write/master side of those two pools, sized independently going forward.
- `replica_database` stays shared across all three read paths (tenant schema, accounts schema, global schema), since reads don't need the same per-role isolation.
- `get_store()` now picks `global_database` when building the global tenant's store, and `master_database`/`accounts_database` otherwise.
- Updated all deployment config files (`config.example.toml`, `development.toml`, `docker_compose.toml`, `config/deployments/env_specific.toml`, `loadtest/config/development.toml`) with the new sections, defaulted to mirror `master_database`'s values so existing deployments aren't broken until operators choose to tune them.

## Test plan
- [x] `cargo check -p router -p storage_impl` — clean
- [x] `cargo clippy -p router -p storage_impl --no-deps` — clean, no new warnings
- [ ] Deploy to a non-prod environment and confirm router/producer/consumer start up and connect using the new config sections
- [ ] Confirm accounts/global schema queries still succeed end-to-end (merchant account CRUD, user/role login flows)

Ran Cypress Locally
image

image


XyneSpaces:
[blocking] Please make the new `accounts_database` and `global_database` settings backward-compatible, or gate rollout on all deployed config providers being migrated first. This PR makes those settings required and startup now validates/decrypts them, so an existing environment that still only provides `master_database` will fail during boot instead of falling back to the current single-pool behavior.

SanchithHegde (approved):
We could have also structured it a bit differently, with per-tenant database credentials (host, username, password) being provided while having per-schema connection configs (schema name, pool sizes, timeouts, etc.).

ShankarSinghC (approved):

PR #13346, opened 2026-07-16
fix(core): moved merchant fingerprint secret from superposition to merchant account table

sahkal:
## Type of Change

- [x] Refactoring

## Description

moved merchant fingerprint secret from superposition to merchant account table

### Additional Changes

### **Overview**
The PR moves the `fingerprint_secret` for merchants from the dynamic configuration service (Superposition) into the primary database as a direct column on the `merchant_account` table. This simplifies data retrieval and reduces the dependency on external config services for fingerprinting payments (used in blocklists).

### **Key Changes**

1. **Database Schema & Migrations**
   - Added a new database migration (`2026-07-16-000001_add_fingerprint_secret_to_merchant_account`) to `ADD COLUMN IF NOT EXISTS fingerprint_secret VARCHAR(128)` to the `merchant_account` table.
   - Updated `schema.rs` and `schema_v2.rs` in `diesel_models` to reflect the new column.

2. **Model Updates**
   - Modified the `MerchantAccount`, `MerchantAccountSetter`, and `MerchantAccountNew` structs in both `diesel_models` and `hyperswitch_domain_models` to include the `fingerprint_secret: Option<Secret<String>>` field.
   - Updated the mapping and conversion logic between domain and database models to account for this new field.

3. **Core Logic Refactoring (Blocklist & Routing)**
   - **Removed Superposition Usage:** Removed `insert_merchant_configs_with_superposition` from `crates/router/src/core/admin.rs`.
   - **Account Creation:** Instead of making a separate API call to Superposition when creating a merchant account, a securely generated `fingerprint_secret` is now immediately embedded and saved with the new `MerchantAccount` record.
   - **Blocklist & Fingerprinting:** Refactored `get_merchant_fingerprint_secret` and `generate_payment_fingerprint` in `crates/router/src/core/blocklist/utils.rs`. They no longer require Superposition dimensions/state to fetch the secret; they simply extract it directly from the passed `MerchantAccount` object.

4. **Configuration & Cleanup**
   - Removed the `FINGERPRINT_SECRET` entry from `config/superposition_seed.toml`.
   - Removed `FINGERPRINT_SECRET` macros and writable config implementations in `dimension_config.rs`.
   - Removed related constants and helper methods (`get_merchant_fingerprint_secret_key`) across `consts.rs` and `merchant.rs`.

## Motivation and Context

Currently we are pushing merchant fingerprint to superposition and then fetching from superposition for further usecase, But merchant fingerprint is a merchant account property and not superposition property so we are moving back merchant fingerprint to merchant account table from superposition.

## How did you test it?

Create merchant account

```shell
curl --location 'localhost:8080/accounts' \
--data-raw '{
"merchant_id": "new_sahkal",
  "locker_id": "m0010",
  "merchant_details": {
    "website": "www.example.com",
    "address": {
      "line1": "1467",
      "line2": "Harrison Street",
      "line3": "Harrison Street",
      "city": "San Fransico",
      "state": "California",
      "zip": "94122",
      "country": "US"
    }
  },
  "webhook_details": {
    "webhook_version": "1.0.1",
  },
  "metadata": {
    "city": "NY",
    "unit": "245"
  },
  "primary_business_details": [
    {
      "country": "US",
      "business": "default"
    }
  ]
}'
```

Checked if fingerprint secret is set in the table or not

Screenshot 2026-07-16 at 3 29 10 PM

## Checklist

### DIFF
--- a/crates/router/src/core/blocklist/utils.rs
+++ b/crates/router/src/core/blocklist/utils.rs
@@ -10,6 +10,7 @@ use crate::{
     core::{
         configs::dimension_state,
         errors::{RouterResult, StorageErrorExt},
+        metrics,
         payments::PaymentData,
     },
     logger,
@@ -231,15 +232,32 @@ pub async fn insert_entry_into_blocklist(
     Ok(blocklist_entry.foreign_into())
 }
 
-pub fn get_merchant_fingerprint_secret(
+pub async fn get_merchant_fingerprint_secret(
+    state: &SessionState,
     merchant_account: &domain::MerchantAccount,
 ) -> RouterResult<String> {
-    merchant_account
-        .fingerprint_secret
-        .as_ref()
-        .map(|secret| secret.peek().clone())
-        .ok_or(errors::ApiErrorResponse::InternalServerError)
-        .attach_printable("fingerprint_secret not found in merchant account")
+    match merchant_account.fingerprint_secret.as_ref() {
+        Some(secret) => Ok(secret.peek().clone()),
+        None => {
+            logger::warn!(
+                merchant_id = ?merchant_account.get_id(),
+                "fingerprint_secret missing from merchant account; falling back to Superposition"
+            );
+            metrics::FINGERPRINT_SECRET_SUPERPOSITION_FETCH_COUNT.add(1, &[]);
+            let dimensions = dimension_state::Dimensions::new()
+                .with_processor_merchant_id(merchant_account.get_id().clone().into());
+            let secret = dimensions
+                .get_fingerprint_secret(&*state.store, state.superposition_service.as_ref(), None)
+                .await;
+
+            match secret.is_empty() {
+                false => Ok(secret),
+                true => Err(errors::ApiErrorResponse::InternalServerError).attach_printable(
+                    "fingerprint_secret not found in merchant account or Superposition",
+                ),
+            }
+        }
+    }
 }
 
 async fn duplicate_check_insert_bin(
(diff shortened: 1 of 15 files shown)


hrithikesh026 (dismissed):

NISHANTH1221 (dismissed):

crates/router/src/core/blocklist/utils.rs
```diff
@@ -231,24 +231,15 @@ pub async fn insert_entry_into_blocklist(
...
-            None, // No targeting key needed for merchant-level config
-        )
-        .await;
-
-    match secret.is_empty() {
-        false => Ok(secret),
-        true => Err(errors::ApiErrorResponse::InternalServerError)
-            .attach_printable("fingerprint_secret not found in Superposition for merchant"),
-    }
+    merchant_account
```
AnuthaDev: We should have superposition as a fallback for now if the value is not found. We already have several `fingerprint_secret`s stored in superposition. We will need to migrate them to the merchant table before we can deploy this change.

This is easy enough for us, but this will also impact the self hosted merchants who are using superposition. If they miss the migration before the new deployment they might start experiencing transaction failures.

Ideally migrations like this should be automatically done in the background on application startup or as a background job and should be transparent to the self hosted merchant.

AnuthaDev (changes requested):

XyneSpaces:
[blocking] Please add a migration/backfill or a read fallback for existing merchants before switching fingerprint generation to `merchant_account.fingerprint_secret`. The new column is nullable and the migration only adds it, so existing merchant rows will keep `NULL`; `get_merchant_fingerprint_secret` now returns `InternalServerError` for those merchants, breaking blocklist/fingerprint generation paths until every row is populated.

XyneSpaces:
[blocking] The merchant-account migration adds `fingerprint_secret` as nullable but does not backfill existing merchants, while the new blocklist/fingerprint path now errors when that field is `None`. Existing merchants that run card fingerprinting can start receiving internal server errors until each row gets a secret.

Please either backfill `merchant_account.fingerprint_secret` in the migration, or add a lazy generation/update path for `None` before removing the Superposition fallback.

crates/router/src/core/blocklist/utils.rs
sahkal: correct, will make it backward compatible on fingerprint fetch!

AnuthaDev (approved):

hrithikesh026 (approved):

PR #13351, opened 2026-07-16
perf(core): Cache business_profile

AnuthaDev:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

This PR adds two-tier caching (in-memory `moka` + Redis) for `business_profile` reads, with a split-storage model: **the encrypted diesel row lives in Redis, the decrypted domain model lives in process memory**. Unlike the existing merchant account / merchant connector account caches — which store the encrypted diesel row in both tiers and pay a decryption call on every cache hit — an in-memory hit here skips both the DB query and the decryption call entirely, while Redis itself never holds plaintext.

**New cache primitive (`crates/storage_impl/src/redis/cache.rs`)**
- `get_or_populate_in_memory_with_transform<T, R>`: a generic two-tier read-through cache where Redis stores type `R` and the in-memory cache stores type `T`, with a caller-supplied async `transform_func: R -> T` applied once when populating the in-memory tier. The lookup order is: in-memory hit (fastest, no transform) → Redis hit (deserialize `R`, transform, populate memory) → DB hit (fetch via `fetch_func`, serialize `R` to Redis, transform, populate memory). This is fully generic — extending the same pattern to merchant account or organization later only needs their decrypting `convert` passed as the transform.
- Redis TTL is set to `2×` the in-memory cache's TTL, so the Redis entry outlives its in-memory copy. When the memory entry expires, the refill is served from Redis (one decrypt call) instead of the database; the DB is only touched when both tiers miss.
- `Cache` now retains its `time_to_live` (`get_time_to_live()`), and `get_or_populate_redis` accepts an optional TTL via `serialize_and_set_key_with_expiry`. Existing `get_or_populate_in_memory` now forwards the cache's TTL to Redis instead of silently inheriting the pool-wide `default_ttl` (300s) — this applies to **all** existing caches (accounts, config, routing, …), moving their Redis expiry from 300s to their cache TTL (30 min for all current caches).

**Caching (`crates/storage_impl/src/business_profile.rs`)**
- `find_business_profile_by_profile_id` is cached under `business_profile_{profile_id}` in `ACCOUNTS_CACHE`, gated behind the `accounts_cache` feature (uncached path preserved). The `fetch_func` returns the raw `diesel::Profile` (no decryption); the transform closure performs the decrypting `convert` into `domain::Profile`. Per-lookup cost: memory hit = 0 decrypt, Redis hit = 1 decrypt, cold = 1 DB + 1 decrypt.
- `find_business_profile_by_merchant_id_profile_id` reuses the same profile_id-keyed entry and enforces merchant ownership in application code (`profile.merchant_id == merchant_id`, returning `ValueNotFound` on mismatch — matching the empty-result behavior of the compound SQL query it replaces). `profile_id` is the table's sole primary key (`VARCHAR(64) PRIMARY KEY`, `pro_`-prefixed nanoid — globally unique), so this check is authorization, not disambiguation. A single cache key per profile also avoids composite-key ambiguity and warms both the auth-path and data-path lookups on the same entry.
- `update_profile_by_profile_id` and `delete_profile_by_profile_id_merchant_id` redact the cache key after the DB write via the existing `redact_from_redis_and_publish` pub/sub flow.
- `find_business_profile_by_profile_name_merchant_id` and `list_profile_by_merchant_id` are intentionally left uncached. `MockDb`, `KVRouterStore`, and `KafkaStore` forwarding impls are untouched.

**Serialization (`crates/diesel_models/src/business_profile.rs`)**
- Added `serde::Serialize`/`serde::Deserialize` to the v1 and v2 diesel `Profile` structs. The diesel model's encrypted columns are the `Encryption` newtype, which serializes as ciphertext — so the Redis entry contains only ciphertext, never plaintext. The domain model needs no serde at all: `moka` stores it as a typed in-process object and never serializes it, which also means ciphertext and decryption context are embedded in the cached struct itself (no risk of writing a cache-reconstructed entity back with corrupted encrypted columns).

**Behavior notes**
- 🔒 Redis and its RDB/AOF snapshots contain only the encrypted diesel row (ciphertext). Decrypted values (webhook auth headers, network tokenization credentials) exist only in process memory for the cache TTL.
- Rolling deploys: old pods never touch these cache keys (the whole path is feature-gated), so updates from old pods during a deploy do not invalidate new entries — profiles updated mid-deploy may be served stale for up to the Redis TTL.
- Cache hits from Redis (but not memory) still pay one encryption-service decryption call; only in-memory hits are fully free of both DB and decrypt. This is the deliberate trade vs. storing decrypted models in Redis.

### Additional Changes

## Motivation and Context

`business_profile` is fetched on every payment confirm. On top of DB fetch overhead we also have encryption service call pverhead. Caching it will improve payments latency

## How did you test it?

- Tested Payments API locally, verified redis cache is created properly.

## Checklist

### DIFF
--- a/crates/storage_impl/src/business_profile.rs
+++ b/crates/storage_impl/src/business_profile.rs
@@ -32,6 +32,7 @@ use crate::{metrics, RedisConnInterface};
 #[cfg(feature = "accounts_cache")]
 const PROFILE_REDIS_CACHE_TTL: i64 = 60 * 60;
 
+/// Cache key for a profile, shared by both the profile-scoped and merchant-scoped lookups.
 #[cfg(feature = "accounts_cache")]
 fn profile_cache_key(profile_id: &common_utils::id_type::ProfileId) -> String {
     format!("business_profile_{}", profile_id.get_string_repr())
@@ -223,7 +224,7 @@ impl<T: DatabaseStore> ProfileInterface for RouterStore<T> {
                 .get_keymanager_state()
                 .attach_printable("Missing KeyManagerState")?
                 .clone();
-            let key = merchant_key_store.key.clone();
+
             let identifier = merchant_key_store.merchant_id.clone().into();
 
             cache::get_or_populate_in_memory_with_transform(
@@ -233,7 +234,7 @@ impl<T: DatabaseStore> ProfileInterface for RouterStore<T> {
                 fetch_func,
                 |diesel_profile| async move {
                     diesel_profile
-                        .convert(&state, &key, identifier)
+                        .convert(&state, &merchant_key_store.key, identifier)
                         .await
                         .change_context(StorageError::DecryptionError)
                 },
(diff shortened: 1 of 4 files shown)


XyneSpaces:
[blocking] The new cache serialization path stores decrypted `Encryptable` inner values in Redis alongside ciphertext. Any Redis reader or leaked Redis backup can recover plaintext values such as webhook headers, card-testing secrets, or network-tokenization credentials without KMS access.

Please keep cached profile data ciphertext-only and decrypt inside the application after fetching; if the goal is to avoid KMS round trips, use an envelope-encrypted cache representation rather than plaintext.

XyneSpaces:
[blocking] Please avoid writing decrypted `Encryptable` contents into the distributed accounts cache. With `accounts_cache` enabled, `encryptable_cache_serde` serializes `inner: value.get_inner().peek().clone()` and `find_business_profile_by_profile_id` stores the decrypted `Profile` through `get_or_populate_in_memory`, so Redis receives plaintext webhook headers / card-testing keys / network-tokenization credentials; cache only the encrypted DB representation or keep decrypted values in memory-only storage.

crates/storage_impl/src/business_profile.rs
```diff
@@ -156,12 +249,32 @@ impl<T: DatabaseStore> ProfileInterface for RouterStore<T> {
...
+                diesel::Profile::find_by_merchant_id_profile_id(&conn, merchant_id, profile_id),
+            )
+            .await
+        }
+
+        #[cfg(feature = "accounts_cache")]
+        {
+            // Reuses the profile_id-keyed cache entry; the merchant ownership check that the
+            // uncached query performs via its compound WHERE clause is enforced here instead
+            let business_profile = self
```
hrithikesh026: if two merchants have same profile_id, the new one will always override the old value

AnuthaDev: Changed the code to use the original DB query, fetching by merchant_id, profile_id

crates/storage_impl/src/business_profile.rs
```diff
@@ -233,7 +234,7 @@ impl<T: DatabaseStore> ProfileInterface for RouterStore<T> {
                 fetch_func,
                 |diesel_profile| async move {
                     diesel_profile
-                        .convert(&state, &key, identifier)
+                        .convert(&state, &merchant_key_store.key, identifier)
```
hrithikesh026, line 237: don't we need cache support for find_business_profile_by_profile_id????
Or do we even need find_business_profile_by_profile_id??? If not being called anywhere or if this function can be replaced with find_business_profile_by_merchant_id_profile_id at all call site, we can simply remove this.

AnuthaDev: Both `profile_id` fetch and `merchant_id_profile_id` fetch are used extensively in the codebase. Analyzing and removing redundancy can eb taken up in a separate PR.

hrithikesh026 (approved):

Sakilmostak (approved):

PR #13374, opened 2026-07-17
perf(routing): Reduce redundant MCA DB fetches

AnuthaDev:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [x] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

This PR removes redundant merchant connector account (MCA) fetches in the payment routing hot path and in contract-based routing config validation.

**1. Single MCA fetch in `perform_eligibility_analysis_with_fallback` (`crates/router/src/core/payments/routing.rs`)**

Previously, `perform_eligibility_analysis` and `perform_fallback_routing` each independently called `get_active_mca_ids`, firing two identical `list_enabled_merchant_connector_accounts_without_encrypted_by_merchant_id_profile_id` queries per call. On a cgraph cache miss, `refresh_cgraph_cache` fired a third identical query.

The MCA list is now fetched once in `perform_eligibility_analysis_with_fallback` and threaded down as a mandatory `&MerchantConnectorAccountsWithoutEncrypted` parameter through `perform_eligibility_analysis` / `perform_fallback_routing` → `perform_cgraph_filtering` → `get_merchant_cgraph` → `refresh_cgraph_cache`. This path runs on every payment confirm, v1/v2 payment routing, and payout routing: **3 DB queries reduced to 1** on a cold cgraph cache, 2 to 1 on a warm one.

Supporting refactors:
- The `active_mca_ids: HashSet<MerchantConnectorAccountId>` parameter was removed everywhere — it was always derived from the same MCA list, and is now computed at its single consumption point inside `perform_cgraph_filtering`.
- `get_active_mca_ids` is replaced by `get_active_merchant_connector_accounts`, which returns the full list. Session routing paths now pass this list down (`SessionRoutingInput.active_mca_ids` → `merchant_connector_accounts`), so their cgraph cache misses also reuse the fetch instead of querying again.
- `refresh_cgraph_cache` no longer queries the DB and drops its unused `key_store` / `profile_id` parameters.
- The MCA fetch error is no longer silently swallowed into an empty list (which, combined with fetch reuse, could have cached an empty constraint graph). It now propagates as a new `RoutingError::MerchantConnectorAccountsFetchFailed` variant.

**2. Batch MCA validation in contract-based routing (`crates/router/src/core/routing.rs`)**

`contract_based_dynamic_routing_setup` and `contract_based_routing_update_configs` validated each `label_info` entry with an individual `find_by_merchant_connector_account_merchant_id_merchant_connector_id` call (N queries for N connectors). Both now fetch all MCAs once via `list_merchant_connector_accounts_without_encrypted_including_disabled_by_merchant_id_profile_id` — the same pattern used by `validate_connectors_in_routing_config` — and validate against an in-memory `FxHashMap`: **N queries reduced to 1**.

**Behavior changes to note:**
- A storage error while fetching MCAs during eligibility analysis now fails the routing call instead of silently routing with an empty active-MCA set (the cgraph-cache-miss half of this path already hard-errored before).
- Contract-based routing validation is now profile-scoped: a `label_info` entry referencing an MCA belonging to a different profile of the same merchant is now rejected with `MerchantConnectorAccountNotFound` (previously accepted). This matches the profile-scoped validation used elsewhere in routing config validation.

### Additional Changes

## Motivation and Context

`list_enabled_merchant_connector_accounts_without_encrypted_by_merchant_id_profile_id` was being executed 2–3 times with identical arguments within a single routing invocation, on the hottest path in the router (every payment confirm). Contract-based routing config create/update similarly issued one MCA lookup per connector in the config. This PR eliminates the redundant round trips, reducing DB/cache load and payment confirm latency.

## How did you test it?

- `cargo check -p router` (v1 default features) passes.
- Compared query behavior by code inspection: the eligibility + fallback path now issues exactly one MCA list query (previously two, three on cgraph cache miss); session routing issues one (previously two on cgraph cache miss).
- Existing payment confirm / routing flows exercise these paths in Cypress and integration suites; no functional change expected for successful routing outcomes since the same list feeds the same derivations.

## Checklist

### DIFF
--- a/crates/router/src/core/payments.rs
+++ b/crates/router/src/core/payments.rs
@@ -12396,12 +12396,15 @@ where
 {
     let chosen = connectors.apply_filter_for_session_routing();
 
-    let active_mca_ids =
-        routing::get_active_mca_ids(&state, processor.get_key_store(), business_profile.get_id())
-            .await
-            .change_context(errors::ApiErrorResponse::GenericNotFoundError {
-                message: "Active mca_ids not found".to_string(),
-            })?;
+    let merchant_connector_accounts = routing::get_active_merchant_connector_accounts(
+        &state,
+        processor.get_key_store(),
+        business_profile.get_id(),
+    )
+    .await
+    .change_context(errors::ApiErrorResponse::GenericNotFoundError {
+        message: "Active merchant connector accounts not found".to_string(),
+    })?;
 
     let session_input = routing::SessionRoutingInput {
         state: &state,
@@ -12410,7 +12413,7 @@ where
         merchant_account: processor.get_account(),
         transaction_type: &transaction_type,
         chosen: &chosen,
-        active_mca_ids: &active_mca_ids,
+        merchant_connector_accounts: &merchant_connector_accounts,
         default_config: &fallback_config,
         backend_input: &mut backend_input,
     };
--- a/crates/router/src/core/payments.rs
+++ b/crates/router/src/core/payments.rs
@@ -12396,15 +12396,15 @@ where
 {
     let chosen = connectors.apply_filter_for_session_routing();
 
-    let merchant_connector_accounts = routing::get_active_merchant_connector_accounts(
+    // Session token routing must never hard-fail on a transient MCA fetch error: degrade
+    // to an empty active set so the merchant still gets session tokens via the fallback
+    // config, instead of returning a client-facing error for an infra failure.
+    let active_mca_ids = routing::get_active_mca_ids_for_session(
         &state,
         processor.get_key_store(),
         business_profile.get_id(),
     )
-    .await
-    .change_context(errors::ApiErrorResponse::GenericNotFoundError {
-        message: "Active merchant connector accounts not found".to_string(),
-    })?;
+    .await;
 
     let session_input = routing::SessionRoutingInput {
         state: &state,
@@ -12413,7 +12413,7 @@ where
         merchant_account: processor.get_account(),
         transaction_type: &transaction_type,
         chosen: &chosen,
-        merchant_connector_accounts: &merchant_connector_accounts,
+        active_mca_ids: &active_mca_ids,
         default_config: &fallback_config,
         backend_input: &mut backend_input,
     };
--- a/crates/router/src/core/payments.rs
+++ b/crates/router/src/core/payments.rs
@@ -12396,9 +12396,10 @@ where
 {
     let chosen = connectors.apply_filter_for_session_routing();
 
-    // Session token routing must never hard-fail on a transient MCA fetch error: degrade
-    // to an empty active set so the merchant still gets session tokens via the fallback
-    // config, instead of returning a client-facing error for an infra failure.
+    // Degrade to an empty active set on a transient MCA fetch error, with an explicit
+    // log, instead of returning a client-facing error for an infra failure. Note the
+    // empty set filters out every MCA-carrying choice, so a warm-cache request yields no
+    // session tokens and a cold-cache refresh can still hard-error.
     let active_mca_ids = routing::get_active_mca_ids_for_session(
         &state,
         processor.get_key_store(),
--- a/crates/router/src/core/payments/routing.rs
+++ b/crates/router/src/core/payments/routing.rs
@@ -904,8 +904,7 @@ pub struct SessionRoutingInput<'a> {
     pub merchant_account: &'a domain::MerchantAccount,
     pub transaction_type: &'a api_enums::TransactionType,
     pub chosen: &'a api::SessionConnectorDatas,
-    pub active_mca_ids:
-        &'a std::collections::HashSet<common_utils::id_type::MerchantConnectorAccountId>,
+    pub merchant_connector_accounts: &'a domain::MerchantConnectorAccountsWithoutEncrypted,
     pub default_config: &'a Vec<routing_types::RoutableConnectorChoice>,
     pub backend_input: &'a mut backend::BackendInput,
 }
@@ -1022,7 +1021,7 @@ impl RoutingStage for SessionRoutingStage {
                     None,
                     profile_id,
                     input.transaction_type,
-                    input.active_mca_ids,
+                    input.merchant_connector_accounts,
                 )
                 .await?;
 
@@ -1035,7 +1034,7 @@ impl RoutingStage for SessionRoutingStage {
                         None,
                         profile_id,
                         input.transaction_type,
-                        input.active_mca_ids,
+                        input.merchant_connector_accounts,
                     )
                     .await?
                 } else {
@@ -2026,6 +2025,7 @@ pub async fn get_merchant_cgraph(
     key_store: &domain::MerchantKeyStore,
     profile_id: &common_utils::id_type::ProfileId,
     transaction_type: &api_enums::TransactionType,
+    merchant_connector_accounts: &domain::MerchantConnectorAccountsWithoutEncrypted,
 ) -> RoutingResult<Arc<hyperswitch_constraint_graph::ConstraintGraph<euclid_dir::DirValue>>> {
     let merchant_id = &key_store.merchant_id;
 
@@ -2064,7 +2064,13 @@ pub async fn get_merchant_cgraph(
     let cgraph = if let Some(graph) = cached_cgraph {
         graph
     } else {
-        refresh_cgraph_cache(state, key_store, key.clone(), profile_id, transaction_type).await?
+        refresh_cgraph_cache(
+            state,
+            key.clone(),
+            transaction_type,
+            merchant_connector_accounts,
+        )
+        .await?
     };
 
     Ok(cgraph)
@@ -2073,19 +2079,11 @@ pub async fn get_merchant_cgraph(
 // #[cfg(feature = "v1")]
 pub async fn refresh_cgraph_cache(
     state: &SessionState,
-    key_store: &domain::MerchantKeyStore,
     key: String,
-    profile_id: &common_utils::id_type::ProfileId,
     transaction_type: &api_enums::TransactionType,
+    merchant_connector_accounts: &domain::MerchantConnectorAccountsWithoutEncrypted,
 ) -> RoutingResult<Arc<hyperswitch_constraint_graph::ConstraintGraph<euclid_dir::DirValue>>> {
-    let mut merchant_connector_accounts = state
-        .store
-        .list_enabled_merchant_connector_accounts_without_encrypted_by_merchant_id_profile_id(
-            &key_store.merchant_id,
-            profile_id,
-        )
-        .await
-        .change_context(errors::RoutingError::KgraphCacheRefreshFailed)?;
+    let mut merchant_connector_accounts = merchant_connector_accounts.clone();
 
     match transaction_type {
         api_enums::TransactionType::Payment => {
@@ -2182,7 +2180,7 @@ pub async fn perform_cgraph_filtering(
     eligible_connectors: Option<&Vec<api_enums::RoutableConnectors>>,
     profile_id: &common_utils::id_type::ProfileId,
     transaction_type: &api_enums::TransactionType,
-    active_mca_ids: &std::collections::HashSet<common_utils::id_type::MerchantConnectorAccountId>,
+    merchant_connector_accounts: &domain::MerchantConnectorAccountsWithoutEncrypted,
 ) -> RoutingResult<Vec<routing_types::RoutableConnectorChoice>> {
     let context = euclid_graph::AnalysisContext::from_dir_values(
         backend_input
@@ -2190,7 +2188,19 @@ pub async fn perform_cgraph_filtering(
             .change_context(errors::RoutingError::KgraphAnalysisError)?,
     );
 
-    let cached_cgraph = get_merchant_cgraph(state, key_store, profile_id, transaction_type).await?;
+    let cached_cgraph = get_merchant_cgraph(
+        state,
+        key_store,
+        profile_id,
+        transaction_type,
+        merchant_connector_accounts,
+    )
+    .await?;
+
+    let active_mca_ids = merchant_connector_accounts
+        .iter()
+        .map(|mca| mca.get_id())
+        .collect::<std::collections::HashSet<_>>();
 
     let mut final_selection = Vec::new();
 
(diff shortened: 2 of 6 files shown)


crates/router/src/core/payments/routing.rs
```diff
@@ -2073,19 +2079,11 @@ pub async fn get_merchant_cgraph(
...
 ) -> RoutingResult<Arc<hyperswitch_constraint_graph::ConstraintGraph<euclid_dir::DirValue>>> {
-    let mut merchant_connector_accounts = state
-        .store
-        .list_enabled_merchant_connector_accounts_without_encrypted_by_merchant_id_profile_id(
-            &key_store.merchant_id,
-            profile_id,
-        )
-        .await
-        .change_context(errors::RoutingError::KgraphCacheRefreshFailed)?;
+    let mut merchant_connector_accounts = merchant_connector_accounts.clone();
```
prajjwalkumar17: **Stale cgraph cache risk.** Now that `refresh_cgraph_cache` builds the graph from a caller-supplied list instead of re-querying the DB, there's a race that can poison the shared `CGRAPH_CACHE`:

1. A payment flow snapshots the MCA list at the top of `perform_eligibility_analysis_with_fallback`.
2. An admin disables/deletes an MCA → `redact_cgraph_cache` evicts the graph for this merchant+profile.
3. The in-flight payment hits the now-cold cache and re-populates it with a graph built from its pre-change snapshot — the disabled connector is back in the cached graph.

The race is self-arming (the admin action itself creates the cache miss the stale snapshot fills), and the poisoned graph is served to *all* subsequent payments until the next eviction/expiry. Previously refresh always re-fetched at refresh time, so this couldn't happen.

Suggestion: keep the DB fetch inside `refresh_cgraph_cache` (the extra query only happens on a cache miss, so the hot-path win of this PR is preserved), or document/accept the staleness window explicitly.

crates/router/src/core/payments/routing.rs
```diff
@@ -4085,25 +4105,17 @@ where
...
-        .unwrap_or_else(|_| {
-            hyperswitch_domain_models::merchant_connector_account::MerchantConnectorAccountsWithoutEncrypted::new(
-                vec![],
-            )
-        });
-
-    let active_mca_ids: std::collections::HashSet<_> =
-        db_mcas.iter().map(|mca| mca.get_id().clone()).collect();
-    Ok(active_mca_ids)
+        .change_context(errors::RoutingError::MerchantConnectorAccountsFetchFailed)
```
prajjwalkumar17, line 4154: **This breaks routing's no-hard-fail guarantee — worst case must still be fallback, not a failed payment.** Routing should never fail a payment; in the worst case we degrade to fallback routing. Pre-PR, on a warm cgraph cache, a transient storage error on this fetch was swallowed into an empty active-MCA set and the payment still proceeded via fallback. Post-PR the error is `?`-propagated at the very top of `perform_eligibility_analysis_with_fallback` — *before* `perform_eligibility_analysis` or `perform_fallback_routing` ever runs — so a single DB blip on this query now fails the payment outright, and the fallback path (whose entire purpose is surviving exactly this kind of failure) never gets a chance to execute. Same applies to the fetches at the top of both `perform_session_flow_routing` variants.

Suggestion: keep `get_active_merchant_connector_accounts` fallible as written, but handle the `Err` at the `perform_eligibility_analysis_with_fallback` boundary — log it and fall through to the merchant default/fallback config (skipping cgraph filtering for that request) instead of aborting. The one hard rule should be: never populate the cgraph cache from a failed or empty fetch.

crates/router/src/core/payments.rs
```diff
@@ -12396,12 +12396,15 @@ where
...
-                message: "Active mca_ids not found".to_string(),
-            })?;
+    let merchant_connector_accounts = routing::get_active_merchant_connector_accounts(
+        &state,
+        processor.get_key_store(),
+        business_profile.get_id(),
+    )
+    .await
+    .change_context(errors::ApiErrorResponse::GenericNotFoundError {
+        message: "Active merchant connector accounts not found".to_string(),
```
prajjwalkumar17: **Session-token routing should degrade, not fail the call.** This `change_context(GenericNotFoundError)` was effectively dead code before (the old `get_active_mca_ids` swallowed storage errors into an empty set, so the session call always proceeded). Now that the fetch hard-errors, a replica timeout during `/payments/session_tokens` fails the entire call — and with a client-facing 404 at that.

Per the principle that routing must never hard-fail (worst case = fallback), the fetch failure shouldn't propagate here at all: on `Err`, log and continue the session flow the way the old code effectively did, so the merchant still gets session tokens (possibly via fallback config). If a failure path is kept for truly unrecoverable cases, it must be a 5xx — an empty MCA list is `Ok(vec![])`, not an error, so the only thing that can reach this mapping is an infra failure, and returning 404 for that will make merchants and 5xx-keyed alerting read an outage as missing configuration.

crates/router/src/core/routing.rs
```diff
@@ -2221,6 +2223,21 @@ pub async fn contract_based_dynamic_routing_setup(
...
     let mut contained_mca = Vec::new();
     if let Some(info_vec) = &config.label_info {
+        let all_mcas = db
+            .list_merchant_connector_accounts_without_encrypted_including_disabled_by_merchant_id_profile_id(
+                processor.get_account().get_id(),
+                &profile_id,
+            )
+            .await
+            .change_context(errors::ApiErrorResponse::MerchantConnectorAccountNotFound {
+                id: processor.get_account().get_id().get_string_repr().to_owned(),
```
prajjwalkumar17: **Wrong error class + misleading id on DB failure.** A list-by-merchant/profile call never fails with "not found" for an empty result (that's `Ok(vec![])`), so the only errors reaching this `change_context` are real storage errors — which will now be returned as `404 "Merchant connector account <merchant_id> does not exist"`. Two problems: the status class is wrong (should be 5xx for an infra failure), and the `id` in the message is the *merchant* id presented as a missing merchant-connector-account, which will confuse users and support tooling.

crates/router/src/core/routing.rs
```diff
@@ -2233,38 +2250,24 @@ pub async fn contract_based_dynamic_routing_setup(
...
-                        },
-                    ))
-                })?;
+            let mca_connector_name = mca_map.get(&info.mca_id).ok_or_else(|| {
+                error_stack::Report::new(
+                    errors::ApiErrorResponse::MerchantConnectorAccountNotFound {
+                        id: info.mca_id.get_string_repr().to_owned(),
+                    },
+                )
+            })?;
```
prajjwalkumar17: **Error-precedence change visible to API clients.** Previously, duplicate detection ran to completion over the whole `label_info` list *before* any existence checks (which then ran via `try_join_all`). Now the checks are interleaved per item, so for `label_info = [A (nonexistent), B, B]` the API returns `MerchantConnectorAccountNotFound` where it previously returned `InvalidRequestData("Duplicate mca configuration received")`.

crates/router/src/core/payments/routing.rs
```diff
@@ -2182,15 +2180,27 @@ pub async fn perform_cgraph_filtering(
...
+    let cached_cgraph = get_merchant_cgraph(
+        state,
+        key_store,
+        profile_id,
+        transaction_type,
+        merchant_connector_accounts,
+    )
+    .await?;
+
+    let active_mca_ids = merchant_connector_accounts
```
prajjwalkumar17: **Minor perf give-back:** this HashSet (cloning every MCA id) is now rebuilt on *every* `perform_cgraph_filtering` call. In the session flow that's up to 2× per payment-method type (chosen + fallback), where pre-PR the set was built once per request and passed down. Since the list is fetched once and immutable, consider building the set once at the fetch site (or a small cached `get_ids()` helper on `MerchantConnectorAccountsWithoutEncrypted`). Relatedly, the `.clone()` of the full decoded MCA list in `refresh_cgraph_cache` (line 2086) copies everything before `retain`/`filter_by_connector_type` discard half of it — `iter().filter(...).cloned()` would clone only the survivors. Both are far cheaper than the DB queries this PR removes, so non-blocking.

crates/router/src/core/routing.rs
```diff
@@ -2324,19 +2327,31 @@ pub async fn contract_based_routing_update_configs(
...
     if let Some(info_vec) = &request.label_info {
+        let all_mcas = db
+            .list_merchant_connector_accounts_without_encrypted_including_disabled_by_merchant_id_profile_id(
+                processor.get_account().get_id(),
+                &profile_id,
+            )
+            .await
+            .change_context(errors::ApiErrorResponse::MerchantConnectorAccountNotFound {
+                id: processor.get_account().get_id().get_string_repr().to_owned(),
+            })?;
```
prajjwalkumar17: **Duplicated validation block, already drifting.** This ~25-line fetch + `FxHashMap` build + per-entry validation block is a near-verbatim copy of the one in `contract_based_dynamic_routing_setup`, and the two copies have already diverged in check ordering (setup checks duplicates before the map lookup, update checks after the label check) — so the same bad request already yields different errors on the two endpoints. Could we extract a shared helper (e.g. next to `validate_connectors_in_routing_config` in `routing/helpers.rs`) taking `(db, merchant_id, profile_id, info_vec)`? Any fix to the error-mapping/scoping comments above would otherwise need to be applied twice.

prajjwalkumar17 (commented):
There are some changes in the PR that brings the routing to a hot failure path, which shouldn't happen, let's refactor all of these.

crates/router/src/core/payments/routing.rs
AnuthaDev: Removed caching from this case

crates/router/src/core/payments/routing.rs
AnuthaDev: Done

crates/router/src/core/payments.rs
AnuthaDev: Done

crates/router/src/core/routing.rs
AnuthaDev: Fixed

crates/router/src/core/routing.rs
AnuthaDev: Fixed

crates/router/src/core/payments/routing.rs
AnuthaDev: Fixed

crates/router/src/core/routing.rs
AnuthaDev: Added helper

crates/router/src/core/payments/routing.rs
```diff
@@ -2415,13 +2437,33 @@ pub async fn perform_eligibility_analysis_with_fallback(
...
+            .await
+        {
+            Ok(merchant_connector_accounts) => merchant_connector_accounts.get_ids(),
+            Err(err) => {
+                logger::error!(
+                    error = ?err,
+                    "euclid_routing: failed to fetch active merchant connector accounts; \
+                     degrading to unfiltered fallback config"
+                );
+                return get_fallback_config(state, transaction_data, business_profile).await;
```
prajjwalkumar17: **The degrade path returns the fallback config completely unfiltered — it should at least keep the `eligible_connectors` intersection.** This early return discards the `eligible_connectors` list computed just above via `update_eligible_connectors_for_installments`, and skips all cgraph/active-MCA filtering. Two concrete consequences in the degrade window:

1. An installment payment can be routed to a fallback connector that doesn't support installments — the eligibility filter existed precisely to prevent that, and it's a pure in-memory intersection, so there's no reason to lose it here.
2. On the payout path (`payouts/helpers.rs` calls this with `check_eligibility` + a straight-through connector), the requested connector is discarded and the payout proceeds on whatever `connectors.first()` of the merchant default config happens to be — money moved via a connector the caller didn't ask for.

Suggestion: when `eligible_connectors` is `Some`, retain only fallback entries whose connector is in that list before returning (empty result is fine — that matches what pre-PR degrade produced and callers already handle it). Degrading coarsely is right; degrading *unconstrained* is more than we need.

crates/router/src/core/payments/routing.rs
```diff
@@ -4085,25 +4129,39 @@ where
...
+        .change_context(errors::RoutingError::MerchantConnectorAccountsFetchFailed)
+}
 
-    let active_mca_ids: std::collections::HashSet<_> =
-        db_mcas.iter().map(|mca| mca.get_id().clone()).collect();
-    Ok(active_mca_ids)
+/// Fetches the set of active MCA ids for session routing, degrading to an empty set on
+/// failure. Session token issuance must never be hard-failed by a transient MCA fetch
+/// error: an empty active set filters the request down to the merchant's fallback
+/// config, matching the pre-existing error-swallowing behaviour but with an explicit log.
```
prajjwalkumar17: ### NIT
**Doc overstates what the empty set does.** "an empty active set filters the request down to the merchant's fallback config" — actually, in `perform_cgraph_filtering` the `mca_active` check is `.unwrap_or(false)` for choices carrying an mca_id, so an empty set drops *all* such choices, fallback-config ones included: with a warm cache the merchant gets zero session tokens, not "tokens via the fallback config". And with a cold cache, `refresh_cgraph_cache`'s own DB fetch still hard-errors (`KgraphCacheRefreshFailed` → 500). Both behaviors are byte-for-byte identical to pre-PR `main`, so no code change needed — but the comment (and its twin at 
`payments.rs:12399`, plus the absolute "Routing must never hard-fail" phrasing at line 2440 — the `?` at line 2468 can still propagate `KgraphCacheRefreshFailed` in the entry-fetch-succeeds-then-DB-flakes window) should say "restores the pre-existing degrade behavior" rather than promising tokens/no-failure the code doesn't guarantee.

crates/router/src/core/routing/helpers.rs
```diff
@@ -464,6 +464,72 @@ impl RoutingAlgorithmHelpers<'_> {
...
+/// Shared by the contract-based setup and update endpoints so both apply the same
+/// checks in the same order:
+/// 1. duplicate `mca_id` in the request -> `InvalidRequestData` ("Duplicate ...")
+/// 2. `mca_id` not present for the profile -> `MerchantConnectorAccountNotFound`
+/// 3. label not matching the connector name -> `InvalidRequestData` ("Incorrect ...")
+///
+/// A failure of the list fetch itself is an infra error (an empty result is `Ok(vec![])`,
+/// not an error), so it maps to `InternalServerError` rather than a 404.
+#[cfg(all(feature = "v1", feature = "dynamic_routing"))]
+pub async fn validate_contract_based_label_info(
```
prajjwalkumar17, line 479: ### NIT
Small note on the "Fixed" for error precedence: the helper *unifies* the check order across both endpoints (duplicate → existence → label, per item) rather than restoring either endpoint's pre-PR order — e.g. setup with `[A (nonexistent), B, B]` still returns 404 where pre-PR returned the duplicate 400. Unified is arguably the better end state, so no change requested — just flagging that clients/tests asserting on the old error codes may still see differences, same caveat as before.

crates/router/src/core/payments/routing.rs
AnuthaDev: Fixed

prajjwalkumar17 (approved):

sai-harsha-vardhan (approved):

PR #13379, opened 2026-07-17
fix(refunds): serialize concurrent refund creation to prevent over-refund (TOCTOU

### LINKED ISSUE #13182: Concurrent refunds can exceed captured payment amount (TOCTOU race condition)
### What we observed

When two refund requests for the same payment are sent simultaneously, both can succeed even when their combined total exceeds the captured amount. The refund validation checks the sum of existing refunds before creating a new one, but two concurrent requests both read the same refund history and both pass validation independently.

**1. Both refunds succeed, total exceeds capture**

We created a $100 payment and sent two $60 refund requests simultaneously. Expected at least one to be rejected, since their combined total exceeds the captured amount. Both returned 200 with `status: "succeeded"`. We verified at the database level that the total refunded amount ($120) exceeds the $100 capture.

### Impact

- Merchants lose real money — both refunds are forwarded to the payment connector and processed independently, resulting in the merchant being debited more than the original payment
- The race is trivially exploitable by anyone with a valid payment ID and API key, since no rate limiting or locking prevents concurrent refund submissions
- The bug exists entirely within Hyperswitch's validation layer, before any connector is called, so it affects all configured connectors (Stripe, Adyen, Braintree, etc.)

### Steps to reproduce

**Environment:**

- Hyperswitch payment router (`docker.juspay.io/juspaydotin/hyperswitch-router:latest`)
- PostgreSQL 15 with Hyperswitch schema migrations applied
- Redis
- A payment connector configured in test mode (we used Stripe, but any connector works since the bug is pre-connector)

**Setup:**

Create a merchant account, API key, and configure a payment connector. These are standard Hyperswitch setup steps:

1. Create a merchant account via the admin API:

```
POST /accounts
Content-Type: application/json
api-key: {admin_api_key}

{
  "merchant_id": "race-test-merchant",
  "merchant_name": "Race Condition Test",
  "merchant_details": {
    "primary_contact_person": "Test",
    "primary_email": "test@example.com"
  }
}
```

Returns 200 with the merchant ID.

2. Create an API key for the merchant:

```
POST /api_keys/{merchant_id}
Content-Type: application/json
api-key: {admin_api_key}

{
  "name": "race-test-key",
  "expiration": "never"
}
```

Returns 200 with an `api_key` value. Use this key for all subsequent requests.

3. Configure a payment connector (Stripe shown, but any connector works):

```
POST /account/{merchant_id}/connectors
Content-Type: application/json
api-key: {api_key}

{
  "connector_name": "stripe",
    "auth_type": "HeaderKey",
    "api_key": "sk_test_..."
  },
  "test_mode": true,
  "payment_methods_enabled": [
    {
      "payment_method": "card",
      "payment_method_types": [
        {
          "card_networks": ["Visa"],
          "minimum_amount": 1,
          "maximum_amount": 68607706
        }
      ]
    }
  ]
}
```

**Control — single refund within limits:**

Create a $100.00 payment and confirm it:

```
POST /payments
Content-Type: application/json
api-key: {api_key}

{
  "amount": 10000,
  "currency": "USD",
  "confirm": true,
  "payment_method": "card",
  "payment_method_data": {
    "card": {
      "card_exp_month": "12",
      "card_exp_year": "2030",
      "card_holder_name": "Test User",
      "card_cvc": "123"
    }
  }
}
```

Returns 200 with `status: "succeeded"`. Extract the `payment_id`.

Submit a single $60.00 refund:

```
POST /refunds
Content-Type: application/json
api-key: {api_key}

{
  "payment_id": "{payment_id}",
  "amount": 6000
}
```

Returns 200 with `status: "succeeded"`. This is correct — $60 is within the $100 captured amount.

A second $60.00 refund submitted sequentially after the first correctly returns 400 because $60 + $60 = $120 exceeds the $100 capture. The validation works for sequential requests.

**Bug — two refunds sent simultaneously:**

Repeat the setup with a fresh $100.00 payment. Then send two $60.00 refund requests at the same time (concurrent HTTP requests, not sequential):

Request A:
```
POST /refunds
Content-Type: application/json
api-key: {api_key}

{
  "payment_id": "{payment_id}",
  "amount": 6000
}
```

Request B (sent simultaneously):
```
POST /refunds
Content-Type: application/json
api-key: {api_key}

{
  "payment_id": "{payment_id}",
  "amount": 6000
}
```

Both return HTTP 200 with `status: "succeeded"`. Neither is rejected.

**Database verification:**

```sql
SELECT
  COALESCE(SUM(refund_amount), 0)::text AS total_refunded,
  COUNT(*)::text AS refund_count
FROM refund
WHERE payment_id = '{payment_id}'
  AND refund_status NOT IN ('failure', 'transaction_failure');
```

Result: `total_refunded = 12000`, `refund_count = 2`. The payment has been refunded $120.00 against a $100.00 capture.

### Reproducing with Dokkimi

[Dokkimi](https://dokkimi.com) is an open source testing tool that stands up isolated Docker environments from declarative YAML definitions — services, databases, seed data, test steps, and assertions all in one file. The definition for this bug is in [dokkimi/dokkimi-in-the-wild/.dokkimi/hyperswitch](https://github.com/dokkimi/dokkimi-in-the-wild/tree/main/.dokkimi/hyperswitch) — `dokkimi run` reproduces it from scratch.

PrazwalR:
Fixes a TOCTOU race in refund creation that can let the total refunded amount exceed the captured amount.

`validate_and_create_refund` performs three unsynchronized steps: (1) read all existing refunds, (2) validate `refund_amount <= amount_captured - Σ(non-failed refunds)`, (3) insert the new refund. No lock, DB transaction, `SELECT … FOR UPDATE`, or idempotency guard sits between steps 1 and 3. Two concurrent refunds for the same payment interleave between the read and the insert, both see the same stale balance, both pass validation, and both insert — a double payout. With `refund_id` omitted it is auto-generated per request, so the unique key gives no dedup either.

Payments already serialize per `payment_id` via `api_locking::LockAction::Hold` (`server_wrap_util` acquires before the handler and frees after, on success and error paths). All three refund-create entry points instead passed `LockAction::NotApplicable`.

### Fix
- Implement `GetLockingInput` for `RefundRequest` (v1) and `RefundsGenericRequestWithResourceId` (v2), keyed on `payment_id`.
- Wire the lock into all three create routes: native v1, native v2 (`routes/refunds.rs`), and stripe-compat (`compatibility/stripe/refunds.rs`).

Concurrent refunds on the same payment now serialize; the second request re-reads the balance (including the first, still-`Pending` refund) and is rejected if it would exceed captured. The insert is inside the locked section and before the connector call, so the critical section is DB-only and stays within the redis lock TTL. Retrieve/list/update/aggregate endpoints are unchanged (reads / non-amount mutations).

Closes #13182

## How did you test it?

Unit tests (all passing):

**`validate_refund_amount` edge cases** — boundary allow, +1 reject, partial sums, Pending reserves balance, Failure/TransactionFailure excluded, mixed statuses.
**Lock wiring** — v1 key == payment_id; v2 Some→Hold, None→NotApplicable.
cargo test -p router --lib refund

test result: ok. 8 passed; 0 failed
cargo test -p router --lib --no-default-features --features "v2,redis-rs" refund_locking

test result: ok. 2 passed; 0 failed
Manual E2E (running stack): create a succeeded payment, fire two full-amount refunds concurrently with no `refund_id`. Before: both can succeed (Σ > captured). After: one succeeds, one 4xx (`ResourceBusy` / `RefundAmountExceedsPaymentAmount`).
## Checklist
- [x] `cargo fmt --all`
- [x] `cargo clippy`
- [x] Added unit tests
- [ ] Modified API contract (no — internal locking only)

### DIFF
--- a/crates/router/src/compatibility/stripe/refunds.rs
+++ b/crates/router/src/compatibility/stripe/refunds.rs
@@ -6,7 +6,10 @@ use router_env::{instrument, tracing, Flow, Tag};
 
 use crate::{
     compatibility::{stripe::errors, wrap},
-    core::{api_locking, refunds},
+    core::{
+        api_locking::{self, GetLockingInput},
+        refunds,
+    },
     logger, routes,
     services::{api, authentication as auth},
     types::api::refunds as refund_types,
@@ -34,6 +37,7 @@ pub async fn refund_create(
     let create_refund_req: refund_types::RefundRequest = payload.into();
 
     let flow = Flow::RefundsCreate;
+    let locking_action = create_refund_req.get_locking_input(flow.clone());
 
     Box::pin(wrap::compatibility_api_wrap::<
         _,
@@ -56,7 +60,7 @@ pub async fn refund_create(
             allow_connected_scope_operation: false,
             allow_platform_self_operation: false,
         }),
-        api_locking::LockAction::NotApplicable,
+        locking_action,
     ))
     .await
 }
--- a/crates/router/src/core/utils/refunds_validator.rs
+++ b/crates/router/src/core/utils/refunds_validator.rs
@@ -299,3 +299,115 @@ pub fn validate_xendit_charge_refund(
         }
     }
 }
+
+#[cfg(all(test, feature = "v1"))]
+mod refund_amount_validation_tests {
+    use common_utils::types::{ConnectorTransactionId, MinorUnit};
+    use diesel_models::refund as diesel_refund;
+
+    use super::*;
+
+    /// Builds a `Refund` row with the given amount and status; all other fields are irrelevant to
+    /// `validate_refund_amount` and are filled with placeholders.
+    fn refund(amount: i64, status: enums::RefundStatus) -> diesel_refund::Refund {
+        let now = common_utils::date_time::now();
+        diesel_refund::Refund {
+            internal_reference_id: "ref_internal".to_string(),
+            refund_id: "ref_1".to_string(),
+            payment_id: common_utils::id_type::PaymentId::default(),
+            merchant_id: common_utils::id_type::MerchantId::default(),
+            connector_transaction_id: ConnectorTransactionId::from("txn_1".to_string()),
+            connector: "stripe".to_string(),
+            connector_refund_id: None,
+            external_reference_id: None,
+            refund_type: enums::RefundType::InstantRefund,
+            total_amount: MinorUnit::new(amount),
+            currency: enums::Currency::USD,
+            refund_amount: MinorUnit::new(amount),
+            refund_status: status,
+            sent_to_gateway: false,
+            refund_error_message: None,
+            metadata: None,
+            refund_arn: None,
+            created_at: now,
+            modified_at: now,
+            description: None,
+            attempt_id: "att_1".to_string(),
+            refund_reason: None,
+            refund_error_code: None,
+            profile_id: None,
+            updated_by: String::new(),
+            merchant_connector_id: None,
+            charges: None,
+            organization_id: common_utils::id_type::OrganizationId::default(),
+            connector_refund_data: None,
+            connector_transaction_data: None,
+            split_refunds: None,
+            unified_code: None,
+            unified_message: None,
+            processor_refund_data: None,
+            processor_transaction_data: None,
+            issuer_error_code: None,
+            issuer_error_message: None,
+            processor_merchant_id: None,
+            created_by: None,
+        }
+    }
+
+    #[test]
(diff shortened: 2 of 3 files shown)


XyneSpaces:
[blocking] The new route-level `LockAction::Hold` makes `server_wrap` hold the per-payment refund lock for the entire `refund_create_core` handler, including the outbound connector call. Please move the lock inside the DB read/validate/insert critical section, or explicitly handle the Redis TTL/request-id mismatch risk for slow connector responses.

PR #13395, opened 2026-07-21
fix(router): eliminate duplicate DB connection pools from building each tenant's store twice

sai-harsha-vardhan:
## Summary
Merge the per-tenant `stores`/`accounts_store` construction into a single `get_store_interface_maps` call so each tenant's `Store`/`ReplicaStore` (and its 4 bb8 pools) is built once instead of twice — previously `get_store_interface_map` and `get_accounts_store_interface_map` each independently called `AppState::get_store_interface`, throwing away half of every build.

## Problem
AppState::with_storage used to call two separate methods per tenant — get_store_interface_map (for stores) and get_accounts_store_interface_map (for accounts_store) — and each one independently built a full Store/ReplicaStore (4 bb8 connection pools: master, replica, accounts_master, accounts_replica) via AppState::get_store_interface. Since both calls were for the same tenant, every tenant's DB store was built twice, with each build only ever using half its own pools.

## Changes
Merged both into a single get_store_interface_maps call that builds the store once per tenant, then derives both stores and accounts_store from that one instance via .get_storage_interface() / .get_accounts_storage_interface() — cheap Arc-clone accessors, not new pools. Same data reachable through both maps as before; purely removes the redundant second build.

## Test plan
- [x] `cargo check -p router -p storage_impl`
- [x] `cargo clippy -p router -p storage_impl --no-deps`

Ran Cypress Locally
image

image


crates/router/src/configs/settings.rs
```diff
@@ -469,79 +469,54 @@ pub struct DecisionConfig {
...
+                    cache_store.clone(),
+                    testable,
+                ))
+                .await;
+                (tenant_name.clone(), store)
+            }))
+            .await;
+
+        let mut stores = HashMap::new();
+        let mut accounts_store = HashMap::new();
```
SanchithHegde: Nit: Can optionally use `HashMap::with_capacity()`, since we know the capacity (number of tenants) beforehand.

SanchithHegde (dismissed):

crates/router/src/configs/settings.rs
```diff
@@ -469,79 +469,54 @@ pub struct DecisionConfig {
...
+                    tenant,
+                    conf.master_database.clone().into_inner(),
+                    conf.accounts_database_config(),
+                    cache_store.clone(),
+                    testable,
+                ))
+                .await;
+                (tenant_name.clone(), store)
+            }))
+            .await;
```
hrithikesh026, line 511: we can do idiomatic .iter().map().collect().

hrithikesh026 (approved):

SanchithHegde (approved):

PR #13396, opened 2026-07-21
feat(locker): admin batch API to provision entities in the card vault

ShankarSinghC:
## Summary
- Adds an admin-api-key-gated `POST /entities_migration` endpoint that takes a list of `merchant_ids` and calls the locker's `POST /entity` per id, provisioning a key-holder record for each.
- One-off migration tool for backfilling locker entities (e.g. for merchants missing an entity record) — not a general-purpose feature.
- Reuses the existing locker `VaultingInterface` + `call_to_vault` client so JWE/JWS encryption and the `x-tenant-id` header are handled the same way as every other locker call — no hand-rolled HTTP.
- Batches ids in chunks of 10 (`futures::future::join_all`) and never fails the whole request on a single-id error; response reports per-id success/error plus aggregate counts.
- Idempotent by construction — the locker's `/entity` endpoint returns the existing record on repeat calls, so safe to re-run for only the failed ids.

## Request / response
```
POST /entities_migration
api-key: <admin_api_key>

{ "merchant_ids": ["merchant_1234", "merchant_5678"] }
```
```json
{
  "total": 2,
  "succeeded": 2,
  "failed": 0,
  "results": [
    { "merchant_id": "merchant_1234", "status": "success", "created_at": "2026-07-21T06:38:05.618Z", "error_code": null, "error_message": null }
  ]
}
```

## Test plan
- [x] `cargo check -p router` clean on this branch
- [x] Manually exercised end-to-end against a local locker (`mock_locker = false`): single id, idempotent re-run (same `created_at` returned), blank-id rejection, and a 10-id batch — all behaved as expected
- [ ] Reviewer to confirm auth/route placement conventions match team expectations for one-off admin/migration endpoints


crates/router/src/types/payment_methods.rs
```diff
@@ -167,6 +167,30 @@ impl VaultingInterface for VaultDelete {
...
+
+#[derive(Debug, serde::Serialize)]
+pub struct EntityCreateRequest {
+    pub entity_id: String,
+}
+
+#[derive(Debug, serde::Deserialize)]
+pub struct EntityCreateResponse {
+    pub entity_id: String,
+    pub created_at: String,
```
SanchithHegde, line 191: Any reason `created_at` is being treated as a string?

SanchithHegde (commented):

PR #13409, opened 2026-07-21
feat(locker): create locker entity during merchant account creation

### LINKED ISSUE #13411: Add support to create entity in locker in merchant create call
Add a locker.create_entity_on_merchant_create config flag (default false). When enabled alongside the existing locker.locker_enabled, merchant account creation calls the locker's entity-create API (POST /entity) before any DB writes or key-manager transfer — a locker failure aborts the merchant creation entirely, so a merchant account is never left without a locker entity.

ShankarSinghC:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

Adds a `locker.create_entity_on_merchant_create` config flag (default `false`). When enabled alongside the existing `locker.locker_enabled`, merchant account creation calls the locker's entity-create API (`POST /entity`) before any DB writes or key-manager transfer — a locker failure aborts the merchant creation entirely, so a merchant account is never left without a locker entity.

### Additional Changes

New key `create_entity_on_merchant_create` under `[locker]` in `config/development.toml`, `config/config.example.toml`, `config/docker_compose.toml`, `config/deployments/env_specific.toml`, `loadtest/config/development.toml`, and `crates/router/src/configs/settings.rs`/`defaults.rs`. Defaults to `false` everywhere, so existing environments are unaffected unless explicitly opted in.

## Motivation and Context

Ensures every merchant has a corresponding locker entity from the moment it's created, instead of relying on a separate backfill/migration step.

## How did you test it?

Manually, against a local router + real (JWE-enabled) locker + Postgres, with `create_entity_on_merchant_create = true`:

Success path:
```bash
curl -s -X POST http://localhost:8080/accounts \
  -H "api-key: test_admin" \
  -H "content-type: application/json" \
  -d '{"merchant_id": "entity_on_create_test_merchant"}'
# -> 200, merchant created; locker entity confirmed to exist (idempotent re-check returns the same created_at)
```

Failure path (locker killed before the call):
```bash
curl -s -X POST http://localhost:8080/accounts \
  -H "api-key: test_admin" \
  -H "content-type: application/json" \
  -d '{"merchant_id": "entity_on_create_should_fail"}'
# -> 500, "Failed to create entity in locker for merchant" in the error chain
```
Confirmed via direct SQL that the failure path leaves **zero rows** in `merchant_account`, `merchant_key_store`, and `business_profile` for that merchant_id — no partial state on locker failure.

## Checklist

### DIFF
--- a/config/development.toml
+++ b/config/development.toml
@@ -153,6 +153,7 @@ host = "http://127.0.0.1:3000"
 mock_locker = true
 locker_enabled = true
 ttl_for_storage_in_secs = 220752000
+create_entity_on_merchant_create = false
 
 [forex_api]
 api_key = ""
--- a/crates/router/src/core/admin.rs
+++ b/crates/router/src/core/admin.rs
@@ -38,7 +38,7 @@ use crate::{
         disputes,
         encryption::transfer_encryption_key,
         errors::{self, RouterResponse, RouterResult, StorageErrorExt},
-        payment_methods::{cards, transformers},
+        payment_methods::{cards, transformers, vault},
         payments::helpers::{self},
         pm_auth::helpers::PaymentAuthConnectorDataExt,
         routing, utils as core_utils,
@@ -313,6 +313,14 @@ pub async fn create_merchant_account(
 
     let key_manager_state: &KeyManagerState = &(&state).into();
     let merchant_id = req.get_merchant_reference_id();
+
+    if state.conf.locker.locker_enabled && state.conf.locker.create_entity_on_merchant_create {
+        vault::create_entity_in_locker(&state, &merchant_id)
+            .await
+            .change_context(errors::ApiErrorResponse::InternalServerError)
+            .attach_printable("Failed to create entity in locker for merchant")?;
+    }
+
     let identifier = km_types::Identifier::Merchant(merchant_id.clone());
     keymanager::transfer_key_to_key_manager(
         key_manager_state,
--- a/crates/router/src/types/payment_methods.rs
+++ b/crates/router/src/types/payment_methods.rs
@@ -182,7 +182,7 @@ impl VaultingInterface for EntityCreate {
 
 #[derive(Debug, serde::Serialize)]
 pub struct EntityCreateRequest {
-    pub entity_id: String,
+    pub entity_id: id_type::MerchantId,
 }
 
 #[derive(Debug, serde::Deserialize)]
--- a/crates/router/src/types/payment_methods.rs
+++ b/crates/router/src/types/payment_methods.rs
@@ -167,6 +167,30 @@ impl VaultingInterface for VaultDelete {
     }
 }
 
+#[derive(Debug, serde::Deserialize, serde::Serialize)]
+pub struct EntityCreate;
+
+impl VaultingInterface for EntityCreate {
+    fn get_vaulting_request_url() -> &'static str {
+        consts::LOCKER_ENTITY_CREATE_REQUEST_URL
+    }
+
+    fn get_vaulting_flow_name() -> &'static str {
+        consts::LOCKER_ENTITY_CREATE_FLOW_TYPE
+    }
+}
+
+#[derive(Debug, serde::Serialize)]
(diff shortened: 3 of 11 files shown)


config/development.toml
```diff
@@ -153,6 +153,7 @@ host = "http://127.0.0.1:3000"
 mock_locker = true
 locker_enabled = true
 ttl_for_storage_in_secs = 220752000
+create_entity_on_merchant_create = false
```
SanchithHegde, line 156: Can keep `true` for development maybe?

crates/router/src/types/payment_methods.rs
```diff
@@ -167,6 +167,30 @@ impl VaultingInterface for VaultDelete {
     }
 }
 
+#[derive(Debug, serde::Deserialize, serde::Serialize)]
+pub struct EntityCreate;
```
SanchithHegde: Nit: Why does this type need `Serialize` / `Deserialize`?

crates/router/src/types/payment_methods.rs
```diff
@@ -167,6 +167,30 @@ impl VaultingInterface for VaultDelete {
...
+
+#[derive(Debug, serde::Serialize)]
+pub struct EntityCreateRequest {
+    pub entity_id: String,
+}
+
+#[derive(Debug, serde::Deserialize)]
+pub struct EntityCreateResponse {
+    pub entity_id: String,
+    pub created_at: String,
```
SanchithHegde: Any reason we're treating this as a string rather than `PrimitiveDateTime`?

config/development.toml
```diff
@@ -153,6 +153,7 @@ host = "http://127.0.0.1:3000"
 mock_locker = true
 locker_enabled = true
```
SanchithHegde, line 154: Both of these being present could cause some confusion though. We need to figure out a way to better restructure this.

crates/router/src/core/admin.rs
SanchithHegde, line 334: What would happen if entity was successfully created in locker but not in encryption service? Would we remain with orphan entities in locker which don't map to a valid entry in the application database? Are we okay with that?

config/development.toml
ShankarSinghC: The reason I kept it as false in local is even though `locker_enabled ` is true in local it does not make a add card request to locker utile `mock_locker ` made as false. As you [mentioned](https://github.com/juspay/hyperswitch/pull/13409#discussion_r3623775193) we would have to restructure it

config/development.toml
ShankarSinghC: Yes will have to re-visit the usecase and `restructure ` it

crates/router/src/core/admin.rs
ShankarSinghC: If merchant account creation fails for some other reason, the entity will remain orphaned in the locker. However, if someone attempts to create a merchant again using the same merchant ID, it should not cause any issues.

The main objective here is to ensure that a merchant account is not created unless the corresponding entity has been created successfully in the locker. That said, is there a better way to handle failure scenarios?

One thing that can think of is maybe delete the entry or schedule a task for if the merchant account create api fails

SanchithHegde (dismissed):

crates/router/src/types/payment_methods.rs
```diff
@@ -167,6 +167,31 @@ impl VaultingInterface for VaultDelete {
...
+    }
+
+    fn get_vaulting_flow_name() -> &'static str {
+        consts::LOCKER_ENTITY_CREATE_FLOW_TYPE
+    }
+}
+
+#[derive(Debug, serde::Serialize)]
+pub struct EntityCreateRequest {
+    pub entity_id: String,
```
kashif-m: This can remain `MerchantId`

kashif-m (approved):

sai-harsha-vardhan (approved):

PR #13410, opened 2026-07-21
feat(paysafe): support Apple Pay / Google Pay recurring (CIT→MIT)

shuklatushar226:
## What

Wires the hyperswitch side of **Paysafe wallet recurring** so a wallet CIT vaults a reusable payment method that a later MIT can replay. Card recurring already worked; this closes the gaps for Apple Pay and Google Pay. Pairs with the UCS-side connector change (juspay/hyperswitch-prism#1986).

## Changes

- **`paysafe.rs`** — `should_call_connector_customer` fires for Apple Pay / Google Pay too (not just Card), so wallets get a Paysafe connector customer to vault under. New `requires_wallet_vault_conversion` capability opts Paysafe into the conversion leg.
- **`ConnectorSpecifications`** — adds `requires_wallet_vault_conversion` (default `false`) + `ConnectorEnum` forwarding, so core stays connector-agnostic instead of hardcoding a connector name in the token gateway.
- **`payment_method_token_create_gateway.rs`** — after leg-1 mints a single-use wallet handle, a second tokenize (in its **own** UCS logging wrapper) converts it into a customer-vaulted MULTI_USE handle via `connector_feature_data`. Fully **fail-soft**: any leg-2 failure keeps the single-use handle so the payment still succeeds (only a future MIT is not replayable). Requires `Clone` on `GrpcHeadersUcs` / `UcsReferenceId` / `UcsResourceId`.
- **`payments.rs`** — runs connector tokenization for a one-shot `PaymentCreate` with `confirm=true` (not just standalone `Confirm`), via a new `is_operation_create` helper with a unit test. ⚠️ **Behavior change:** this now triggers connector tokenization on one-shot create+confirm for **any** connector that has a `[tokenization]` entry configured — a general fix, but worth noting for connectors beyond Paysafe.
- **`tokenization.rs`** — under `mock_locker` (local dev only; deployment configs omit it → prod/CI use the real locker) the wallet vault call has no backend; degrade to a locker-less payment method so the mandate (`connector_mandate_details`) is still saved.
- **config** — enable `google_pay` in the paysafe `[tokenization]` filter, add `paysafe` to the `apple_pay`/`google_pay` mandate connector lists, and route paysafe **UCS-only** (the native connector does not implement the wallet flows) — applied to `development.toml` **and all deployment configs** (`integration_test`/`sandbox`/`production`). If your org uses a separate deployment-config repo, those need the same edits.

## Testing

Verified end-to-end (hyperswitch → UCS → Paysafe sandbox):
- **Google Pay** real DIRECT-decrypt token: CIT `succeeded` with `payment_method_id` (mandate is a converted MULTI_USE handle) → MIT `succeeded`.
- **Apple Pay** real Safari PKPaymentToken: CIT `succeeded` with `payment_method_id` → MIT ×2 `succeeded`.
- **Card** CIT→MIT regression: green (confirms the capability flag correctly skips conversion for cards and the one-shot gate works).

## Closes

Closes juspay/hyperswitch-cloud#22283


config/deployments/integration_test.toml
```diff
@@ -1027,7 +1027,7 @@ connector_list = "juspaythreedsserver, ctp_mastercard, ctp_visa"
 connector_list = "worldpayvantiv"
 
 [grpc_client.unified_connector_service]
-ucs_only_connectors = "imerchantsolutions, paytm, phonepe, hyperpg, revolv3, fiservcommercehub, absa_sanlam, interpayments, payconex, dlocal, barclaycard, tsys_transit, jpmorgan, datatrans"    # Comma-separated list of connectors that use UCS only
+ucs_only_connectors = "imerchantsolutions, paytm, phonepe, hyperpg, revolv3, fiservcommercehub, absa_sanlam, interpayments, payconex, dlocal, barclaycard, tsys_transit, jpmorgan, datatrans, paysafe"    # Comma-separated list of connectors that use UCS only
```
AmitsinghTanwar007: is it confirmed that no merchant uses paysafe through HS

shuklatushar226: yes

JeevaRamu0104 (dismissed):

crates/router/src/core/payments/tokenization.rs
```diff
@@ -1363,20 +1363,66 @@ pub async fn save_in_locker_internal(
...
+                // Dev-only scaffolding: `mock_locker` has no real wallet-vaulting
+                // backend, so `add_wallet_to_locker` always fails there (it tries
+                // to encrypt for a vault that isn't running). The recurring (MIT)
+                // flow only needs the payment_method row's connector_mandate_details,
+                // so degrade to a locker-less payment method instead of dropping the
+                // mandate. Gated on `mock_locker` (true only in local dev configs;
+                // deployed envs use the real locker), so this never runs in
+                // integration-test/sandbox/production. The catch is intentionally
+                // broad: under `mock_locker` every wallet vault attempt fails the
+                // same way, so there is no narrower error to match on.
```
hrithikesh026: Mock locker behaviour must be changed to accommodate wallet use case as well. instead of handling here separately.

crates/router/src/core/payments/gateway/payment_method_token_create_gateway.rs
```diff
@@ -181,7 +212,97 @@ where
...
+                        .to_string()
+                        .into(),
+                );
+                let leg2_header_payload = state
+                    .get_grpc_headers_ucs(unified_connector_service_execution_mode)
+                    .external_vault_proxy_metadata(None)
+                    .merchant_reference_id(leg2_merchant_reference_id)
+                    .resource_id(leg2_resource_id)
+                    .lineage_ids(leg2_lineage_ids);
+                let leg2_outcome = Box::pin(unified_connector_service::ucs_logging_wrapper_granular(
```
hrithikesh026: Here we are moving from a granular approach to composite. Should not be done. Core must handle this.

crates/router/src/core/payments/tokenization.rs
shuklatushar226: Done — moved the mock support into the vault layer instead of the call site. `vault::call_to_vault` now has a `mock_locker` branch that synthesizes the fingerprint / add-vault responses, mirroring the legacy card locker's `add_card_to_vault` mock branch. Since wallets, bank_debit and v2 all share `call_to_vault`, this fixes them together at one choke point, so the special-case wallet fallback (and the broad `Err(_) if mock_locker` catch) in `tokenization.rs` is removed entirely. Pushed in 01b77ac.

crates/router/src/core/payments/gateway/payment_method_token_create_gateway.rs
shuklatushar226: Done — the second (wallet vault-conversion) call is no longer in the gateway. `execute()` is a single granular UCS call again, matching every other granular gateway. Core now sequences the conversion: `tokenization::add_payment_method_token` runs a second `execute_payment_gateway` after leg-1, gated by the `requires_wallet_vault_conversion` capability flag, fully fail-soft. The leg-1 token is handed to the second call via a new `connector_feature_data` field on `PaymentMethodTokenizationData` (mapped in the tokenize request builder) instead of the gateway mutating the gRPC request — so each gateway call stays one gRPC call and core owns the sequence, mirroring the `create_order_at_connector` → `add_payment_method_token` idiom. Pushed in 01b77ac.

crates/router/src/core/payment_methods/vault.rs
```diff
@@ -1996,6 +1996,30 @@ pub async fn call_to_vault<V: pm_types::VaultingInterface>(
     let locker = &state.conf.locker;
     let jwekey = state.conf.jwekey.get_inner();
 
+    // `mock_locker` has no real vault backend, so synthesize the decrypted
+    // responses the callers parse — mirroring the mock branch the legacy card
+    // locker already has (`add_card_to_vault`). This makes wallet, bank_debit and
+    // v2 vaulting work in local dev instead of failing on a live call to a vault
+    // that isn't running. Only the flows the tokenization path exercises are
+    // mocked; the rest fall through to the real call (unchanged behaviour).
+    if locker.mock_locker {
```
hrithikesh026: this again seems like a workaround style implementation.

shuklatushar226: Done — moved the mock off inline synthesis onto the established DB-backed convention. The `mock_locker` branch in `call_to_vault` now delegates to a dedicated `mock_call_to_vault` that persists to / reads back from the `locker_mock_up` table via `StorageInterface`, mirroring the legacy card locker's `mock_call_to_locker_hs`. So an add actually inserts a row (genuinely retrievable and deletable) instead of returning throwaway ids, and all five vault flows are covered (add / get-fingerprint / retrieve / delete / entity-create).

Kept it central in `call_to_vault` rather than at the typed wrappers because ~15 call sites hit `call_to_vault` directly (`cards.rs`, `payment_methods.rs`, `tokenization.rs`, workflows) — the choke point is the only place that covers them uniformly.

Verified under `mock_locker = true`: a wallet CIT inserts a `locker_mock_up` row holding the decrypted vaulting data keyed by the vault id, and CIT→MIT still replays. Pushed in 231d213.

hrithikesh026 (dismissed):

deepanshu-iiitu (dismissed):
Connector changes look good

hrithikesh026 (dismissed):

crates/router/src/core/payments.rs
```diff
@@ -9248,6 +9248,15 @@ where
...
 
+    // Connector tokenization must also run for a one-shot PaymentCreate with
+    // confirm=true (the payment reaches the connector in the same call), not just
+    // the standalone Confirm operation — otherwise a connector that mints a
+    // pre-authorize token (e.g. Paysafe payment handles) gets an Authorize with
+    // none. Mirrors the wallet-predecrypt gate in
+    // get_decrypted_wallet_payment_method_token.
+    let is_confirm_leg = is_operation_confirm(&operation)
+        || (is_operation_create(&operation) && payment_data.get_payment_attempt().confirm);
+
```
ShankarSinghC: In case of Payment create with confirm = true / s2s flows the operation will be changed to confirm. Check `if_not_create_change_operation` in payment create `get_trackers`

crates/router/src/core/payment_methods/vault.rs
```diff
@@ -1996,6 +1996,15 @@ pub async fn call_to_vault<V: pm_types::VaultingInterface>(
...
     let jwekey = state.conf.jwekey.get_inner();
 
+    // `mock_locker` has no real vault backend. Delegate to a dedicated, DB-backed
+    // mock that persists to / reads back from the `locker_mock_up` table through
+    // `StorageInterface` — the same mechanism the legacy card locker uses
+    // (`mock_call_to_locker_hs`) — so an add is genuinely retrievable and
+    // deletable instead of returning throwaway ids.
+    if locker.mock_locker {
+        return mock_call_to_vault::<V>(state, payload).await;
+    }
```
ShankarSinghC: Why do we want to have `mock_locker` flow ?

crates/router/src/core/payments/tokenization.rs
```diff
@@ -1824,6 +1824,78 @@ pub async fn add_payment_method_token<F: Clone, T: types::Tokenizable + Clone>(
...
+                        match Box::pin(gateway::execute_payment_gateway(
+                            state,
+                            conversion_integration,
+                            &conversion_router_data,
+                            payments::CallConnectorAction::Trigger,
+                            None,
+                            None,
+                            gateway_context.clone(),
+                        ))
+                        .await
```
ShankarSinghC: I was thinking that any additional connector calls for `setup_future_usage` could be handled within the Payments Core itself. With this modular flow, these calls are expected to be handled by the Payments Core logic rather than by the Payment Methods layer.

shuklatushar226:
UCS follow-up is ready: https://github.com/juspay/hyperswitch-prism/pull/2023

This HS branch now runs the Paysafe wallet conversion only after a successful CIT, persists only the validated MULTI_USE handle, and leaves every other connector on the default no-op path. The rebased branch passed `cargo +nightly fmt --all -- --check` and `cargo check -p router`; the paired Apple Pay and Google Pay CIT-to-MIT flows were validated locally.

PR #13413, opened 2026-07-21
feat(tsys_transit): add billing line1 + zip as required dynamic fields (CIT + MIT)

awasthi21:
## Summary

`tsys_transit` hard-requires `billing.address.line1` and `billing.address.zip`
(AVS) on **every** card authorization — for both CIT and MIT — but had **no
entry** in the connector required-fields config. As a result the SDK never
collected them and payments failed at the connector with:

```
UCS_400: Missing required field: billing.address.line1
```

This adds a `TsysTransit` entry to `payment_connector_required_fields.rs` with
`card_basic()` + `BillingAddressLine1` + `BillingAddressZip` (common fields), so
the SDK surfaces and collects them for both credit and debit cards.

## Why line1 + zip (and only these)

The connector's `extract_for_authorize` extracts `billing.address.line1` and
`billing.address.zip` unconditionally (no channel gate) and errors if absent —
they map to the TSYS `<addressLine1>` / `<zip>` AVS fields. City/state/country
are not sent on the wire, so they are intentionally not required.

Verified against the TSYS cert script (Test Card Data → AVS data: Address 8320,
Zip 85284) and live: a CIT/MIT without billing fails with the missing-field
error; with line1 + zip it returns `A0000`.

## Test

Extends `test_required_fields_to_json` to assert `TsysTransit` exposes
`billing.address.line1` + `billing.address.zip` (and card fields) on both the
Credit and Debit payment method types.

### DIFF
--- a/crates/payment_methods/src/configs/payment_connector_required_fields.rs
+++ b/crates/payment_methods/src/configs/payment_connector_required_fields.rs
@@ -4267,9 +4267,7 @@ fn test_required_fields_to_json() {
                     .fields
                     .get(&Connector::TsysTransit)
                     .expect("tsys_transit required fields should be present");
-                assert!(tsys_fields
-                    .common
-                    .contains_key("billing.address.line1"));
+                assert!(tsys_fields.common.contains_key("billing.address.line1"));
                 assert!(tsys_fields.common.contains_key("billing.address.zip"));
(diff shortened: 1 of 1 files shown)


awasthi21:
Superseded by #13416 (combined into a single PR).

PR #13427, opened 2026-07-22
fix(payment_methods): reject delete for already-redacted payment methods

ShankarSinghC:
## Summary
- `delete_payment_method_core` only guarded against `status == Inactive`, but a successful delete sets `status = Redacted` (not `Inactive`), so the guard never caught an already-deleted payment method.
- Repeated `DELETE /v1/payment-methods/{id}` calls on the same id kept returning `200`, silently re-running the status update and re-issuing a vault delete request with the stale `locker_id`.
- Added `ApiErrorResponse::PaymentMethodRedacted` (code `IR_11`, reusing the code already used by `CustomerRedacted` for the analogous customer-redaction case, following the existing pattern of `ClientSecretNotGiven`/`ClientSecretExpired` sharing `IR_08`) and reject re-deletion once the payment method is `Redacted`:
```json
{"error":{"type":"invalid_request","message":"Payment method has already been redacted","code":"IR_11"}}
```

## Test plan
- [ ] `DELETE` an active payment method — succeeds (200), status becomes `Redacted`.
- [ ] `DELETE` the same payment method again — now returns the `IR_11` error instead of a phantom `200`.

### DIFF
--- a/crates/hyperswitch_domain_models/src/errors/api_error_response.rs
+++ b/crates/hyperswitch_domain_models/src/errors/api_error_response.rs
@@ -203,6 +203,8 @@ pub enum ApiErrorResponse {
     MandateActive,
     #[error(error_type = ErrorType::InvalidRequestError, code = "IR_11", message = "Customer has already been redacted")]
     CustomerRedacted,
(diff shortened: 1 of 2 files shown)


ShankarSinghC:
Consolidated into #13423 — closing this in favor of that PR.

PR #13433, opened 2026-07-22
feat(payments): add offer engine foundation and connectivity check

apoorvdixit88:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

Foundation for the Juspay Offer Engine integration (Milestone 1). This lays down
config, config resolution, an API-key-authenticated client, and a dev/admin-only
connectivity check. **No payment-flow behavior changes** — no eligibility /
confirm / apply / list / notify wiring yet.

**What's included**

- **Static application config** `[offer_engine]` (`base_url`, `api_key`,
  `merchant_id`). Optional via serde defaults — a missing or partial section
  defaults to empty strings and never fails startup.
- **Two Superposition runtime configs** (defaults live in
  `config/superposition_seed.toml`):
  - `offer_engine_enabled` — boolean master gate / kill switch, default `false`.
  - `offer_engine_credential_source` — string enum `none | application`, default
    `none` (extensible to org/merchant/profile sources later).
- **`resolve_offer_engine_config`** — evaluates the kill switch **first**
  (short-circuits and returns `Ok(None)` when off, without reading anything
  else), then the credential source. Returns `Some(config)` only when both allow
  a call, so callers never re-check enablement. Returns `Err` only when a
  required Superposition value or the application config is missing/invalid.
- **`OfferEngineClient`** — thin client over `external_services::http_client`
  with HTTP Basic API-key auth; no custom retries; never logs the key.
- **`POST /offer_engine/connectivity`** — dev/admin-only (AdminApiAuth) endpoint
  that verifies infra allowlisting / reachability to Offer Engine without running
  any business flow.

**Fail-safe by design:** when disabled (the default), the resolver returns
`Ok(None)` before touching credentials, and a forgotten/unseeded environment
behaves exactly like "Offer Engine off" — no offers, zero payment impact.

### Additional Changes

Config files touched:
- `config/config.example.toml`, `config/docker_compose.toml` — new `[offer_engine]` section
- `config/superposition_seed.toml` — `offer_engine_enabled`, `offer_engine_credential_source` defaults
- `crates/router/src/configs/settings.rs` — `OfferEngineConfig`

## Motivation and Context

Milestone 1 issue: juspay/hyperswitch-cloud#18111

First milestone of the Offer Engine integration: give Hyperswitch the ability to
read Offer Engine application config, fetch the optional runtime credential
source (and a kill switch) from Superposition, decide whether Offer Engine should
be called, and make a basic API-key-authenticated HTTP call — verified via a
protected connectivity endpoint. Subsequent milestones add the eligibility /
apply payment-flow integration on top of this foundation.

## How did you test it?

Manually via the connectivity endpoint against a live Offer Engine sandbox,
toggling the two Superposition configs. The endpoint resolves config exactly like
future payment-flow callers will, so it exercises the full gate + credential-source
+ client-auth + reachability path.

**Request**
```bash
curl -X POST http://localhost:8080/offer_engine/connectivity \
  -H 'api-key: <admin-api-key>' -H 'Content-Type: application/json'
```

**Responses**

| # | `offer_engine_enabled` | `offer_engine_credential_source` | Response |
|---|---|---|---|
| A — fully enabled | `true` | `application` | `{"enabled":true,"reachable":true,"status_code":200,"detail":"Reached Offer Engine (HTTP 200)"}` |
| B — kill switch engaged | `false` | `application` | `{"enabled":false,"reachable":null,"status_code":null,"detail":"Offer Engine is not enabled for this context (offer_engine_enabled is false or credential source is none)"}` |
| C — gate open, no source | `true` | `none` | `{"enabled":false,"reachable":null,"status_code":null,"detail":"Offer Engine is not enabled for this context (offer_engine_enabled is false or credential source is none)"}` |

Scenario **B** confirms `offer_engine_enabled` acts as a true kill switch — it
disables Offer Engine even when a valid credential source is configured.

## Checklist


crates/router/src/core/offer_engine/config.rs
```diff
@@ -0,0 +1,71 @@
...
+    context: Option<&ConfigContext>,
+    targeting_key: Option<&String>,
+) -> CustomResult<Option<ResolvedOfferEngineConfig>, OfferEngineError> {
+    let enabled: bool = state
+        .superposition_service
+        .get_config_value(
+            superposition_consts::OFFER_ENGINE_ENABLED,
+            context,
+            targeting_key,
+        )
```
NISHANTH1221: we should define the dimensions on the `dimension_config.rs` and call it using the dimensions wherever we calling this function. If there are no dimension we can create an empty dimesnion type and call it..

crates/router/src/core/offer_engine/config.rs
```diff
@@ -0,0 +1,71 @@
...
+        return Ok(None);
+    }
+
+    let raw_source: String = state
+        .superposition_service
+        .get_config_value(
+            superposition_consts::OFFER_ENGINE_CREDENTIAL_SOURCE,
+            context,
+            targeting_key,
+        )
```
NISHANTH1221: same for this config too..

crates/router/src/configs/settings.rs
```diff
@@ -194,6 +194,8 @@ pub struct Settings<S: SecretState> {
     #[serde(default)]
     pub enhancement: Option<HashMap<String, String>>,
     pub superposition: SecretStateContainer<SuperpositionClientConfig, S>,
+    #[serde(default)]
+    pub offer_engine: OfferEngineConfig,
```
ThisIsMani: Can be option.

config/superposition_seed.toml
```diff
@@ -116,6 +116,18 @@ schema = { type = "boolean" }
...
 
+[default-configs.offer_engine_enabled]
+value = false
+schema = { type = "boolean" }
+description = "Master gate for the Offer Engine integration. When false (default) Offer Engine is never called, regardless of offer_engine_credential_source. Set true (per tenant/env override) to allow Offer Engine calls."
+change_reason = "Offer Engine disabled by default; enable via Superposition override"
+
+[default-configs.offer_engine_credential_source]
+value = "none"
+schema = { type = "string", enum = ["none", "application"] }
```
ThisIsMani: Can be Option<String>

apoorvdixit88: Used a string enum with an explicit `none` variant so the off state is explicit and exhaustive rather than an ambiguous null, and it stays extensible to more credential sources later.

ThisIsMani (dismissed):

NISHANTH1221 (dismissed):

crates/router/src/routes/offer_engine.rs
```diff
@@ -0,0 +1,26 @@
...
+use router_env::{instrument, tracing, Flow};
+
+use super::app;
+use crate::{
+    core::{api_locking, offer_engine},
+    services::{api, authentication as auth},
+};
+
+#[instrument(skip_all, fields(flow = ?Flow::OfferEngineConnectivityCheck))]
+pub async fn offer_engine_connectivity_check(
```
hrithikesh026, line 11: should we include this in deep health check???

apoorvdixit88: Kept it out of the deep health check. Right now offer engine is optional and gated (default off), so its reachability should not gate instance readiness; if it were unreachable while disabled the instance is still perfectly healthy. The dedicated POST /offer_engine/connectivity (admin) can be leveraged.

hrithikesh026 (dismissed):

crates/router/src/core/offer_engine/client.rs
```diff
@@ -0,0 +1,137 @@
...
+use external_services::http_client;
+use hyperswitch_masking::{Mask, PeekInterface};
+
+use super::types::{OfferEngineError, ResolvedOfferEngineConfig};
+use crate::{consts::BASE64_ENGINE, routes::SessionState};
+
+const OFFERS_LIST_PATH: &str = "v1/offers/list";
+
+#[derive(Debug, Clone)]
+pub struct OfferEngineClient {
```
jarnura, line 15: Already there is micro service frame work is there use that instead of creating new

crates/router/src/core/offer_engine/client.rs
```diff
@@ -0,0 +1,137 @@
...
+}
+
+impl OfferEngineClient {
+    pub fn new(config: ResolvedOfferEngineConfig) -> Self {
+        Self { config }
+    }
+
+    fn build_url(&self, path: &str) -> CustomResult<url::Url, OfferEngineError> {
+        let mut base_url = self.config.base_url.clone();
+        if !base_url.path().ends_with('/') {
```
jarnura: The config builder in appstate should validated this already, redundant validation is not required

crates/router/src/core/offer_engine/config.rs
```diff
@@ -0,0 +1,68 @@
...
+use error_stack::ResultExt;
+use hyperswitch_masking::ExposeInterface;
+
+use super::types::{OfferEngineCredentialSource, OfferEngineError, ResolvedOfferEngineConfig};
+use crate::{core::configs::dimension_state::Dimensions, routes::SessionState, utils};
+
+pub async fn resolve_offer_engine_config(
+    state: &SessionState,
+) -> CustomResult<Option<ResolvedOfferEngineConfig>, OfferEngineError> {
+    let dimensions = Dimensions::new();
```
jarnura: Dimensions scope is not available in the parent function?

jarnura (changes requested):

crates/router/src/core/offer_engine/config.rs
apoorvdixit88: This endpoint is only a temporary connectivity check for  testing, not a merchant/profile-specific flow.

 This is admin api only for this we just need to verify that the app-wide Offer Engine config can be resolved and that the service is reachable, so the config lookup is intentionally done at the global scope. I made this explicit by using `DimensionsGlobal`, and `resolve_offer_engine_config` now receives the scope from the caller.
 
Once integration testing is done, this can be removed. In milestone 2, the actual eligibility flow will call Offer Engine from the payment path where the real runtime context exists.

crates/router/src/core/offer_engine/client.rs
apoorvdixit88: Addressed, moved to MicroserviceClient; `OfferEngineClient` now implements `MicroserviceClient`, and the connectivity call is wired through `impl_microservice_flow!`

crates/router/src/core/offer_engine/config.rs
```diff
@@ -0,0 +1,55 @@
...
+        .offer_engine
+        .as_ref()
+        .filter(|config| config.is_configured())
+        .ok_or_else(|| {
+            error_stack::report!(OfferEngineError::MissingApplicationConfig(
+                "offer_engine application config is not set".to_string()
+            ))
+        })?;
+
+    let base_url = url::Url::parse(&app_config.base_url).change_context(
```
jarnura: In app_config itself this should be url only

jarnura (changes requested):

jarnura (approved):

ThisIsMani (approved):

PR #13434, opened 2026-07-23
feat(vault): route HyperswitchVault calls through an optional egress proxy

ShankarSinghC:
## Summary
- Self-hosted Hyperswitch deployments without direct internet egress need the injector's HyperswitchVault detokenize and charge calls to route through an egress proxy (e.g. Squid), the same way VGS traffic already does via `proxy_url`.
- Adds optional `proxy_url` / `certificate` fields to the `hyperswitch_vault` connector's MCA metadata (`ExternalVaultConnectorMetadata`), threads them through the gRPC `x-external-vault-metadata` header into UCS, and into the injector's HTTP client for both the detokenize call (`transformers.rs::get_permanent_pm_id_from_temporary_token`) and the injector's own proxy request (`injector.rs::make_hyperswitch_vault_request`).
- Falls back to direct/global-proxy behavior when `proxy_url` is unset, so existing in-cluster deployments are unaffected.
- Reorders the untagged `ExternalVaultProxyMetadata` enum (HyperswitchVault before VGS) to avoid deserialization ambiguity now that both variants can structurally overlap.

Companion PR in connector-service: juspay/hyperswitch-prism#<fill-in>

## Test plan
- [x] `cargo check -p injector`, `cargo check -p router --features v1` — clean
- [x] Verified locally end-to-end against a mock SaaS vault behind a local Squid proxy: both the detokenize `GET` and the injector's charge `POST` appear in Squid's access log, authenticated with the credentials from the vault MCA's `proxy_url` metadata
- [ ] Backward-compat check (no `proxy_url` set → no Squid traffic, direct call) — pending

### DIFF
--- a/crates/external_services/src/grpc_client/unified_connector_service.rs
+++ b/crates/external_services/src/grpc_client/unified_connector_service.rs
@@ -155,10 +155,11 @@ pub struct VaultConnectorAuth {
 #[derive(Debug, Clone, serde::Deserialize, serde::Serialize)]
 #[serde(untagged)]
 pub enum ExternalVaultProxyMetadata {
+    /// HyperswitchVault data variant — must be first so serde matches vault_endpoint+vault_auth_data
+    /// before falling through to VgsMetadata (which only needs proxy_url+certificate)
+    HyperswitchVaultMetadata(HyperswitchVaultMetadata),
     /// VGS proxy data variant
     VgsMetadata(VgsMetadata),
-    /// HyperswitchVault data variant
-    HyperswitchVaultMetadata(HyperswitchVaultMetadata),
 }
 
 /// Complete external vault proxy configuration to be serialized and sent to UCS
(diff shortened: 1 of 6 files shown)


XyneSpaces:
[should-fix] Propagate invalid vault MCA metadata instead of converting the parse failure to `None`. With the added `.ok()`, a merchant that configures malformed `ExternalVaultConnectorMetadata` silently bypasses the per-merchant egress proxy and sends the detokenize call via the default path.

Location: `crates/router/src/core/payments/operations/payment_confirm_external_vault_proxy.rs` (`parse_value::<ExternalVaultConnectorMetadata>(...).ok()`).

XyneSpaces:
[should-fix] The detokenize path parses vault MCA metadata with `.ok()` and forwards only `proxy_url`, so malformed proxy metadata silently falls back to direct routing and any configured CA certificate is dropped. Please propagate metadata parse errors and pass the parsed `certificate` into `get_permanent_pm_id_from_temporary_token`, populating `Proxy::mitm_ca_certificate` / `mitm_enabled` instead of hard-coding them to `None`.

XyneSpaces:
[should-fix] Preserve the metadata parse error instead of dropping it with `.ok()`. With malformed HyperswitchVault MCA metadata, this path silently treats `proxy_url` as absent and falls back to the router-wide proxy, so a merchant configured to detokenize through a specific egress proxy can be routed incorrectly.

Reference: `crates/router/src/core/payments/operations/payment_confirm_external_vault_proxy.rs`, added `.parse_value::<ExternalVaultConnectorMetadata>(...).ok()`.

XyneSpaces:
[should-fix] `parse_value` failures are silently discarded in `crates/router/src/core/payments/operations/payment_confirm_external_vault_proxy.rs` via `.ok()`, so malformed MCA metadata causes the request to bypass the configured per-merchant egress proxy instead of surfacing the configuration error. Propagate the parse error (as the UCS-side `build_unified_connector_service_external_vault_proxy_metadata_*` functions already do with `.transpose()?`) or at least log the failure before falling back.

PR #13443, opened 2026-07-23
feat(user): copy wallet, FRM and additional merchant data while cloning a connector

### LINKED ISSUE #13449: feat(user): copy wallet, FRM and additional merchant data while cloning a connector
copy wallet, FRM and additional merchant data while cloning a connector

kanikabansal08:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

Connector cloning copies a merchant connector account from one profile to another within the same merchant. `build_cloned_connector_create_request` hardcoded four fields to `None`, so anything held in them had to be configured again by hand on the destination connector.

Three of those four are now carried over:

| Field | Cloned | Reasoning |
|---|---|---|
| `connector_wallets_details` | yes | Wallet certificates and credentials. Owned by the merchant, and both profiles belong to the same merchant. |
| `frm_configs` | yes | `FrmConfigs.gateway` names a connector, it is not a merchant connector account id, so nothing points back at the source profile. |
| `additional_merchant_data` | yes | Open banking recipient data. The locker entry is keyed by merchant id, and the connector recipient id is issued against the merchant's connector account, which is already cloned. |
| `pm_auth_config` | **no** | `PaymentMethodAuthConnectorChoice` carries an `mca_id`, and `validate_pm_auth` rejects a reference whose profile differs from the connector being created. Cloning it would hard-fail the request. |

`metadata` and `connector_account_details` were already copied verbatim and are unchanged.

The allowlist in the local development configs is widened alongside this, adding several connectors and admitting `bank_transfer`, `bank_redirect`, `pay_later` and `wallet` payment method types. `bank_debit` is deliberately left out: it is the payment method that relies on `pm_auth_config`, which is the one field still stripped.

### Additional Changes

`config/development.toml` and `config/docker_compose.toml` — `clone_connector_allowlist`.

These are local development defaults only. Deployed environments are configured separately and are unaffected by this PR.

## Motivation and Context

Cloning a connector into a new profile left the merchant re-entering configuration they had already supplied on the source connector, even though every value involved belongs to the same merchant that owns both profiles. Copying these three fields removes that manual step and makes the clone a usable starting point rather than a partial one.

Two behaviours worth recording for reviewers:

**Apple Pay and domain verification.** `applepay_verified_domains` is written only by the Apple Pay verification flow and is never copied by a clone. Apple validates the Apple Pay session against the domain the checkout page is served from, so if the destination profile serves a different domain from the source, `POST /verify/apple_pay/{merchant_id}` has to be run against the new merchant connector account before Apple Pay works there. Same-domain profiles need nothing. Google Pay has no equivalent registration step.

**Open banking recipients.** For the `AccountData` variant, `create_connector` routes `additional_merchant_data` through `process_open_banking_connectors`, which registers a recipient rather than simply storing the value. A clone therefore provisions a new recipient instead of reusing the existing one. The account details are identical either way, so funds settle to the same merchant account. No open banking connector is currently in the allowlist.

## How did you test it?

Static checks only so far:

- `cargo check -p router --features v1` — passes
- `cargo +nightly fmt --all -- --check` — clean
- `cargo clippy -p router --features v1` — no new lints
- Every connector name and payment method type added to the allowlist was parsed through `Connector::from_str` and `PaymentMethodType::from_str`, the same `FromStr` implementations the config deserializer uses, to confirm the block loads. Worth noting that `deserialize_hashset_inner` fails the whole allowlist if any single value is unparseable.

Manual verification of an end-to-end clone is still pending and should happen before merge:

1. Configure a connector in a source profile with wallets, FRM configs and payment methods enabled.
2. Clone it into a destination profile.
3. Confirm `connector_wallets_details`, `frm_configs` and `additional_merchant_data` are populated on the destination merchant connector account.
4. Confirm a session token call succeeds for the cloned wallet, re-running Apple Pay domain verification first if the destination profile serves a different domain.

## Checklist


ThisIsMani (dismissed):

crates/router/src/utils/user.rs
```diff
@@ -454,6 +454,45 @@ pub async fn build_cloned_connector_create_request(
...
+            serde_json::Value::parse_value(
+                merchant_data.into_inner().expose(),
+                "AdditionalMerchantData",
+            )
+            .change_context(UserErrors::InternalServerError)
+            .attach_printable("Unable to deserialize additional_merchant_data")
+        })
+        .transpose()?;
+
+    let connector_wallets_details = source_mca
```
hrithikesh026: there is an foreign_try_from impl to convert from `DomainMerchantConnectorAccount` to `MerchantConnectorResponse`.  All these conversion are already there in that implementation. You can reuse them.

hrithikesh026 (approved):

ThisIsMani (approved):

PR #13452, opened 2026-07-24
feat(routing): Decision Engine cut-over routing entry and per-card SSO handoff

### LINKED ISSUE #13460: feat(routing): Decision Engine cut-over routing entry and per-card SSO handoff
## Problem

When a merchant's profile is **cut over to the Decision Engine** (`routing_result_source_<profile_id> = decision_engine`), Hyperswitch needs to:

1. Route that profile's live payments via the Decision Engine — **without** requiring a parallel Hyperswitch-side routing algorithm on the profile, and
2. Hand the merchant off from the Control Center's **Workflow → Routing** screen to the specific Decision Engine dashboard page, authenticated.

Today there is no backend endpoint for the dashboard handoff, and payment-time DE routing is only consulted when the profile *also* has an active HS-side rule-based algorithm ("piggyback"), so a profile managed purely on the DE is never evaluated.

## Proposed change

- Add **`POST /routing/entry`**:
  - returns `{ "is_cutover": bool }` for the profile's routing source (a cheap source check, no SSO minting), and
  - on `?target=<volume|rule|multi_objective|debit>` for a cut-over profile, mints a **fresh one-time SSO code** on the Decision Engine and returns `{ "redirect_url": "<dashboard>/routing/<page>?code=..." }` deep-linking to that page. DE merchants are provisioned on demand (best-effort create + retry on first mint).
- **Payment-time full handoff:** when a profile is cut over, hand the full static/rule routing decision to the Decision Engine regardless of whether Hyperswitch also has an active local algorithm; fall back to the local static/default connectors if DE returns nothing or errors.
- Add **`open_router.dashboard_url`** config — the DE dashboard (browser) URL used to build the SSO redirect.

## Notes

- The SSO handoff code is **single-use and short-lived** (60s TTL); no session token is ever placed in a URL.
- Companion Control Center change: juspay/hyperswitch-control-center#5243

prajjwalkumar17:
Closes #13460

## Type of Change

- [x] New feature

## Description

Adds the backend for routing a profile that is **cut over to the Decision Engine (DE)**, driven per-profile by the `routing_result_source_<profile_id>` config (`decision_engine` vs `hyperswitch_routing`).

**1. `POST /routing/entry` (dashboard handoff)**
- Plain call → `{ "is_cutover": true | false }` — a cheap source check with **no** SSO minting. The Control Center uses it to decide, per profile, whether its routing cards should deep-link to DE.
- `?target=<volume|rule|multi_objective|debit>` on a cut-over profile → mints a **fresh one-time SSO code** on the DE and returns `{ "redirect_url": "{dashboard_url}/routing/<page>?code=..." }` deep-linking to that specific DE page. DE merchants are provisioned on demand (best-effort create + retry on the first mint). The code is single-use and short-lived; no session token is ever placed in a URL.

**2. Payment-time full handoff (no piggyback)**
- When a profile is cut over, hand the full static/rule routing decision to the DE **regardless** of whether Hyperswitch also has an active local algorithm (removes the earlier requirement that HS carry a `RuleBasedRouting` algorithm for DE to be consulted). If DE returns nothing or errors, fall back to the local static / default connectors.

**3. Config**
- Add `open_router.dashboard_url` — the DE dashboard (browser) URL used to build the SSO redirect.
- Stop seeding `routing_result_source` in `superposition_seed.toml` so the per-profile DB config decides the source (a seeded default would shadow the per-profile config).

## Motivation and Context

Merchants cut over to the Decision Engine should manage routing in the DE dashboard, and their live traffic should be routed by the DE — without needing a parallel Hyperswitch-side algorithm. This wires the source check, the authenticated per-card handoff, and the payment-time delegation, with a safe fallback to local routing on any DE failure.

Companion Control Center PR: juspay/hyperswitch-control-center#5243

## How did you test it?

Manually, locally, against a profile with `routing_result_source_<profile_id> = decision_engine`:

- `POST /routing/entry` → `{ "is_cutover": true }`; `?target=volume|rule|multi_objective|debit` → returns a `redirect_url` deep-linking to the corresponding DE page with a freshly minted `?code=`, which the DE SPA exchanges for a session.
- Non-cut-over profile → `{ "is_cutover": false }`, `redirect_url` null.
- Payments on a cut-over profile with no HS-side algorithm → DE is consulted; when DE returns no decision, the attempt falls back to the default connectors (verified via `payment_attempt.routing_approach` / `connector`).
- DE unreachable → mint/route errors are caught and the flow falls back to local routing (no payment failure).

## Checklist

### DIFF
--- a/config/development.toml
+++ b/config/development.toml
@@ -1561,6 +1561,7 @@ connector_list = "juspaythreedsserver, ctp_mastercard, ctp_visa"
 dynamic_routing_enabled = false
 static_routing_enabled = false
 url = "http://localhost:8080"
+dashboard_url = "http://localhost:5173"
 
 [l2_l3_data_config]
 enabled = "true"
--- a/crates/router/src/configs/settings.rs
+++ b/crates/router/src/configs/settings.rs
@@ -407,6 +407,9 @@ pub struct OpenRouter {
     pub dynamic_routing_enabled: bool,
     pub static_routing_enabled: bool,
     pub url: String,
+    /// Browser-facing Decision Engine dashboard base URL, used for the merchant SSO redirect.
+    #[serde(default)]
+    pub dashboard_url: String,
 }
 
 #[derive(Debug, Deserialize, Clone, Default)]
--- a/crates/router/src/core/routing.rs
+++ b/crates/router/src/core/routing.rs
@@ -70,6 +70,57 @@ use crate::{
     utils::{self, OptionExt},
 };
 
+/// Dashboard routing entry: reports the profile's routing source and, for a cut-over profile with a `target`, returns a one-time DE dashboard deep-link.
+#[cfg(all(feature = "olap", feature = "v1"))]
+pub async fn routing_entry(
+    state: SessionState,
+    platform: domain::Platform,
+    profile_id: Option<common_utils::id_type::ProfileId>,
+    target: Option<routing_types::DecisionEngineRoutingTarget>,
+) -> RouterResponse<routing_types::RoutingEntryResponse> {
+    let profile_id = profile_id.get_required_value("profile_id")?;
+
+    let dimensions = dimension_state::Dimensions::new()
+        .with_processor_merchant_id(platform.get_processor().get_processor_merchant_id())
+        .with_provider_merchant_id(platform.get_provider().get_provider_merchant_id())
+        .with_profile_id(profile_id.clone());
+
+    use crate::core::payments::routing::utils::get_routing_result_source;
+    let routing_source = get_routing_result_source(&state, &dimensions).await;
+    let is_cutover = matches!(
+        routing_source,
+        routing_types::RoutingResultSource::DecisionEngine
+    );
+
+    // Mint a fresh one-time code only when a card was clicked (`target`) on a cut-over profile.
+    let redirect_url = match (is_cutover, target) {
+        (true, Some(target)) => {
+            // First mint may fail if the profile has no DE merchant yet; create it best-effort and retry once.
+            let code = match helpers::mint_decision_engine_sso_code(&state, &profile_id).await {
+                Ok(code) => code,
+                Err(_) => {
+                    let _ = helpers::create_decision_engine_merchant(&state, &profile_id).await;
+                    helpers::mint_decision_engine_sso_code(&state, &profile_id).await?
+                }
+            };
+            Some(format!(
+                "{}/{}?code={}",
+                state.conf.open_router.dashboard_url,
+                target.dashboard_path(),
+                code
+            ))
+        }
+        _ => None,
+    };
+
+    Ok(service_api::ApplicationResponse::Json(
+        routing_types::RoutingEntryResponse {
+            is_cutover,
+            redirect_url,
+        },
+    ))
+}
+
 pub enum TransactionData<'a> {
     Payment(PaymentsDslInput<'a>),
     #[cfg(feature = "payouts")]
--- a/crates/router/src/core/routing.rs
+++ b/crates/router/src/core/routing.rs
@@ -85,7 +85,6 @@ pub async fn routing_entry(
         .with_provider_merchant_id(platform.get_provider().get_provider_merchant_id())
         .with_profile_id(profile_id.clone());
 
-    use crate::core::payments::routing::utils::get_routing_result_source;
     let routing_source = get_routing_result_source(&state, &dimensions).await;
     let is_cutover = matches!(
         routing_source,
@@ -95,12 +94,27 @@ pub async fn routing_entry(
     // Mint a fresh one-time code only when a card was clicked (`target`) on a cut-over profile.
     let redirect_url = match (is_cutover, target) {
         (true, Some(target)) => {
-            // First mint may fail if the profile has no DE merchant yet; create it best-effort and retry once.
             let code = match helpers::mint_decision_engine_sso_code(&state, &profile_id).await {
                 Ok(code) => code,
-                Err(_) => {
+                // Provision the DE merchant only when it does not exist yet (DE returns 404), then retry once.
+                Err(err)
+                    if matches!(
+                        err.current_context(),
+                        errors::RoutingError::RoutingEventsError {
+                            status_code: 404,
+                            ..
+                        }
+                    ) =>
+                {
                     let _ = helpers::create_decision_engine_merchant(&state, &profile_id).await;
-                    helpers::mint_decision_engine_sso_code(&state, &profile_id).await?
+                    helpers::mint_decision_engine_sso_code(&state, &profile_id)
+                        .await
+                        .change_context(errors::ApiErrorResponse::InternalServerError)?
+                }
+                Err(err) => {
+                    return Err(err)
+                        .change_context(errors::ApiErrorResponse::InternalServerError)
+                        .attach_printable("Failed to mint Decision Engine SSO code");
                 }
             };
             Some(format!(
--- a/crates/router/src/core/routing/helpers.rs
+++ b/crates/router/src/core/routing/helpers.rs
@@ -79,6 +79,7 @@ pub const DECISION_ENGINE_RULE_GET_ENDPOINT: &str = "rule/get";
 pub const DECISION_ENGINE_RULE_DELETE_ENDPOINT: &str = "rule/delete";
 pub const DECISION_ENGINE_MERCHANT_BASE_ENDPOINT: &str = "merchant-account";
 pub const DECISION_ENGINE_MERCHANT_CREATE_ENDPOINT: &str = "merchant-account/create";
+pub const DECISION_ENGINE_MERCHANT_TOKEN_ENDPOINT: &str = "auth/admin/merchant-token";
 
 /// Provides us with all the configured configs of the Merchant in the ascending time configured
 /// manner and chooses the first of them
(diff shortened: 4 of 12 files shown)


XyneSpaces:
Automated review findings:

[should-fix] `crates/router/src/core/routing.rs` — fragile dashboard URL construction  
The added redirect URL uses `format!("{}/{}?code={}", state.conf.open_router.dashboard_url, ...)`. Because `dashboard_url` has `#[serde(default)]`, a missing config becomes `""`, producing a relative URL like `/routing/volume?code=...` and breaking the merchant SSO redirect. Please join the URL with `reqwest::Url` (or `url::Url`) and normalize the base path, or make `dashboard_url` required/validated.

[should-fix] `crates/router/src/core/routing.rs` — first mint error and merchant-creation result are discarded  
The `Err(_) => { let _ = helpers::create_decision_engine_merchant(...).await; helpers::mint_decision_engine_sso_code(...).await? }` branch swallows the original failure and the outcome of `create_decision_engine_merchant`. If the DE token endpoint is down or returns an auth error, this retry masks the root cause. Please log the original error and the create-call outcome, or only retry when the error specifically signals a missing merchant.

[should-fix] `crates/router/src/core/payments/routing.rs` — static-routing gate removed  
The prior guard restricted the static request to `RuleBasedRouting`; the change removes that guard so any profile with `static_routing_enabled` now evaluates static rules on DE. If this is intentional for the DE cutover, please add an explicit guard or update the comment to clarify that the change affects all static-approach profiles.

XyneSpaces:
[should-fix] Please preserve the error from `create_decision_engine_merchant` instead of discarding it before retrying SSO code minting.

In the new fallback path, `let _ = helpers::create_decision_engine_merchant(...).await;` loses the concrete merchant-creation failure, so the second mint failure is the only error surfaced even when creation is the root cause.

XyneSpaces:
[should-fix] Decision Engine cut-over / SSO entry issues

- `OpenRouter.dashboard_url` defaults to an empty string and is interpolated directly into `redirect_url`. If the config key is omitted, the redirect becomes `/{path}?code=...`, which is malformed. Use `Option<String>` or validate at startup.

- `routing_entry` swallows the first `mint_decision_engine_sso_code` error and the result of `create_decision_engine_merchant`, so a persistent failure surfaces only as the second mint failing without the original cause. Log/return the underlying error instead of discarding it.

- `DecisionEngineMerchantTokenResponse.code` and `RoutingEntryResponse.redirect_url` are plain `String` in derived-`Debug` structs; they can leak the one-time SSO code in traces/metrics. Consider wrapping them in `Secret<String>`.

- In `HybridRoutingStage`, the guard `input.static_approach == RoutingApproach::RuleBasedRouting` was removed, so any static approach now sends a static request to the DE rule path when `static_routing_enabled` is true. Please confirm this is intentional; otherwise non-rule static routing may hit the wrong DE endpoint.

XyneSpaces:
Automated review findings for this PR:

[blocking] `crates/router/src/core/routing/helpers.rs:2741` — `DecisionEngineMerchantTokenResponse.code` should be `Secret<String>`
`code` carries a one-time SSO token from the Decision Engine. Store it as `Secret<String>` so it is masked in logs, metrics, and `Debug` output; comparable token/code fields elsewhere in the codebase already use `Secret<String>`.

```suggestion
    pub code: Secret<String>,
```
Expose the value with `.expose_secret()` only when formatting the redirect URL.

[blocking] `crates/router/src/core/routing.rs:102-103` — retry branch silently discards errors
The `Err(_)` arm ignores the original mint failure and `let _ = helpers::create_decision_engine_merchant(&state, &profile_id).await;` discards the merchant-create failure. This makes misconfigurations undebuggable. Propagate the error, or match only the specific "merchant not found" case and attach a printable context.

XyneSpaces:
Review findings:

[blocking] `crates/router/src/core/routing.rs` (~line 76): `routing_entry` silently retries on any mint failure. The `Err(_)` arm discards the original error and blindly calls `create_decision_engine_merchant` before retrying. This masks transient/network errors and can create duplicate merchants. Only retry when the error explicitly indicates the merchant is missing; otherwise propagate.

[should-fix] `crates/router/src/routes/routing.rs` (~line 37): Query parsing errors are silently ignored. `web::Query::from_query(...).ok()` treats malformed `target` values as absent. Return a 400 for invalid enum values instead of falling back to `None`.

[should-fix] `crates/router/src/core/routing.rs` (~line 76): Missing `#[instrument]` on the async core function. Add `#[instrument(skip_all, fields(profile_id))]`.

[should-fix] `crates/router/src/configs/settings.rs` (~line 410): `dashboard_url` defaults to an empty string. An empty default produces a relative redirect URL. Use `Option<String>` and fail fast if unset, or provide a sensible default.

config/development.toml
jagan-jaya, line 1: any other toml file we need to add this dashboard_url?

crates/router/src/configs/settings.rs
```diff
@@ -407,6 +407,9 @@ pub struct OpenRouter {
     pub dynamic_routing_enabled: bool,
     pub static_routing_enabled: bool,
     pub url: String,
+    /// Browser-facing Decision Engine dashboard base URL, used for the merchant SSO redirect.
+    #[serde(default)]
+    pub dashboard_url: String,
```
jagan-jaya, line 412: Why is this mandatory? Will this affect older deployments?

crates/router/src/core/routing/helpers.rs
```diff
@@ -2733,6 +2734,45 @@ pub async fn create_decision_engine_merchant(
...
+            None,
+            None,
+        )
+        .await
+        .change_context(errors::ApiErrorResponse::InternalServerError)
+        .attach_printable("Failed to mint SSO code on decision engine")?
+        .response;
+
+    let code = response
+        .ok_or(errors::ApiErrorResponse::InternalServerError)?
```
jagan-jaya: add attach printable

crates/router/src/core/routing.rs
```diff
@@ -70,6 +70,57 @@ use crate::{
...
+    target: Option<routing_types::DecisionEngineRoutingTarget>,
+) -> RouterResponse<routing_types::RoutingEntryResponse> {
+    let profile_id = profile_id.get_required_value("profile_id")?;
+
+    let dimensions = dimension_state::Dimensions::new()
+        .with_processor_merchant_id(platform.get_processor().get_processor_merchant_id())
+        .with_provider_merchant_id(platform.get_provider().get_provider_merchant_id())
+        .with_profile_id(profile_id.clone());
+
+    use crate::core::payments::routing::utils::get_routing_result_source;
```
jagan-jaya: why inline import here?

crates/router/src/core/routing.rs
```diff
@@ -70,6 +70,57 @@ use crate::{
...
+    let routing_source = get_routing_result_source(&state, &dimensions).await;
```
jagan-jaya, line 88: Is this from the cache or the DB? Can we cache it if it in calling DB always?

crates/router/src/core/routing.rs
```diff
@@ -70,6 +70,57 @@ use crate::{
...
+        routing_types::RoutingResultSource::DecisionEngine
+    );
+
+    // Mint a fresh one-time code only when a card was clicked (`target`) on a cut-over profile.
+    let redirect_url = match (is_cutover, target) {
+        (true, Some(target)) => {
+            // First mint may fail if the profile has no DE merchant yet; create it best-effort and retry once.
+            let code = match helpers::mint_decision_engine_sso_code(&state, &profile_id).await {
+                Ok(code) => code,
+                Err(_) => {
```
jagan-jaya: let's create it only if the merchant_not_found error is thrown

crates/router/src/routes/routing.rs
```diff
@@ -31,6 +31,41 @@ use crate::{
     services::{api as oss_api, authentication as auth, authorization::permissions::Permission},
     types::domain,
 };
+
+/// Dashboard routing entry: returns the profile's routing source and, with `?target`, a DE deep-link.
+#[cfg(all(feature = "olap", feature = "v1"))]
+#[instrument(skip_all)]
+pub async fn routing_entry(state: web::Data<AppState>, req: HttpRequest) -> impl Responder {
+    let flow = Flow::DecisionEngineSsoRedirect;
+    let target = web::Query::<routing_types::RoutingEntryRequest>::from_query(req.query_string())
```
jagan-jaya: Check if the library itself provides the query param from app.rs instead of manually parsing it

jagan-jaya (changes requested):

config/development.toml
prajjwalkumar17: Only `development.toml` and `config.example.toml` carry an `[open_router]` section today (both have `dashboard_url`). `sandbox.toml` / `production.toml` / `docker_compose.toml` / `integration_test.toml` don't have an `[open_router]` block at all — those environments configure `open_router` via env vars, so `dashboard_url` is provided the same way as `open_router.url` (`ROUTER__OPEN_ROUTER__DASHBOARD_URL`). So no other repo toml needs it.

crates/router/src/configs/settings.rs
prajjwalkumar17: It's not mandatory — the field is `String` with `#[serde(default)]`, so a config that omits `dashboard_url` deserializes to `""` and boots fine. Older deployments are unaffected; the value is only read on the DE cut-over redirect path.

crates/router/src/core/routing/helpers.rs
prajjwalkumar17: Added `.attach_printable("Decision engine returned an empty SSO code response")` on the empty-code path. The mint fn also now surfaces the underlying `RoutingError`/status instead of flattening to `InternalServerError`. Done in `9610856`.

crates/router/src/core/routing.rs
prajjwalkumar17: Removed — `get_routing_result_source` is already in scope via the module-level `utils::*` import, so the inline `use` was redundant. `9610856`

crates/router/src/core/routing.rs
prajjwalkumar17: It's already cached, not a per-request DB call. The superposition path serves from an `open_feature` provider snapshot (15s polling interval), and the DB fallback (`find_config_by_key`) hits the in-memory `CONFIG_CACHE` (moka, ~30-min TTL). So it doesn't hit the DB on every invocation — no extra caching needed here.

crates/router/src/core/routing.rs
prajjwalkumar17: Done — the DE merchant is now created only when minting fails with a 404 (merchant-not-found), and other mint errors propagate instead of blindly creating + retrying. `9610856`

crates/router/src/routes/routing.rs
prajjwalkumar17: Switched to actix's `web::Query<RoutingEntryRequest>` extractor as a handler argument (auto-wired by the framework), removing the manual `from_query`. `app.rs` needs no change since the extractor is resolved from the request. `9610856`

jagan-jaya (approved):

VenuMadhav2541 (approved):
Analytics LGTM

Issue #13458, opened 2026-07-26
bug(analytics): user_id missing from api_events for report-generation flows


muditbhutani:
### Summary

Dashboard/JWT-authenticated API calls are supposed to carry the acting user's `user_id` into the `api_events` analytics/audit stream via `AuthenticationType::MerchantJwt { user_id, .. }`. For one specific authentication code path, `user_id` is hardcoded to `None` instead of the real value from the JWT payload — even though the same value is correctly extracted and used elsewhere in the same function.

### Root cause

`crates/router/src/services/authentication.rs`, in the `#[cfg(feature = "v1")]` impl of `AuthenticateAndFetch<AuthenticationDataWithUserId, A> for JWTAuth` (around line 5182-5261):

```rust
Ok((
    (auth.clone(), Some(payload.user_id.clone())),   // real user_id, used for authz/query filtering
    AuthenticationType::MerchantJwt {
        merchant_id: payload.merchant_id,
        user_id: None,                                 // <-- same payload, discarded here
    },
))
```

`payload.user_id` is cloned and correctly passed through in the returned tuple (used by the calling analytics handlers for permission/query-scoping), but the `AuthenticationType` value — which feeds the `ApiEvent`/`api_events` audit stream — hardcodes `user_id: None`.

This `AuthenticateAndFetch<AuthenticationDataWithUserId, _> for JWTAuth` impl is used exclusively by the report-generation routes in `crates/router/src/analytics.rs`.

Notably, the `#[cfg(feature = "v2")]` counterpart of the exact same impl (a few lines below, same struct, same callers) correctly does `user_id: Some(payload.user_id)` — confirming this is an oversight in the v1 impl rather than an intentional design choice. Git blame traces the `None` back to the original PR that introduced this impl in Jan 2024 (#3469, "feat(users): Added blacklist for users"); every other `AuthenticationType` construction site in the file (~27 others) correctly forwards the real user_id.

### Impact / blast radius

Verified against internal analytics data. The impact is precise and limited: every event on the report-generation flows (`GeneratePaymentReport`, `GeneratePayoutReport`, `GenerateRefundReport`, `GenerateDisputeReport`, `GenerateAuthenticationReport`) was missing `user_id`, while every other `merchant_jwt`-authenticated flow (which uses different, unaffected `authenticate_and_fetch` impls) correctly carried it.

[...]

PR #13462, opened 2026-07-26
feat(analytics): add auth_user_id to api_events

### LINKED ISSUE #13461: feat(analytics): expose user_id as a top-level field on api_events for audit logging
### Background

`ApiEvent` (`crates/router/src/events/api_logs.rs`) embeds `AuthenticationType` via `#[serde(flatten)]` to carry auth context into the emitted analytics/audit event. `AuthenticationType` itself is serialized with:

```rust
#[serde(tag = "api_auth_type", content = "authentication_data", rename_all = "snake_case")]
```

This is an *adjacently tagged* enum, so when flattened into `ApiEvent`, its variant fields (including `user_id`, present on `MerchantJwt`, `OrganizationJwt`, `UserJwt`, `MerchantJwtWithProfileId`, `SinglePurposeJwt`, `SinglePurposeOrLoginJwt`) do not reach the JSON top level — they get nested one level down, inside an `authentication_data` object (e.g. `{"api_auth_type": "merchant_jwt", "authentication_data": {"merchant_id": "...", "user_id": "..."}}`).

The `api_events` ClickHouse schema (`crates/analytics/docs/clickhouse/scripts/api_events.sql`) has always had a dedicated **top-level** `user_id Nullable(String)` column, alongside a separate `authentication_data Nullable(String)` blob column. Because of the serialization shape above, `user_id` was never populated in that top-level column, even for dashboard/JWT-authenticated calls that do carry a real user id — it silently ended up buried in the `authentication_data` blob string instead, unusable for direct SQL filtering/grouping.

This is separate from and in addition to the fix in #13458 / the linked PR there, which addressed a specific code path that discarded `user_id` entirely before it even reached `AuthenticationType`.

### What this change does

- Adds `get_user_id(&self) -> Option<String>` to `AuthenticationType` (`authentication.rs`), extracting the user id independent of the enum's serde tagging.
- Adds a plain, non-flattened `user_id: Option<String>` field directly on `ApiEvent` (`api_logs.rs`), populated via `auth_type.get_user_id()` at construction time in `ApiEvent::new`.

With this, `user_id` now serializes at the top level of the emitted event JSON (matching the flat `user_id` column in ClickHouse), in addition to still being present inside `authentication_data` (left as-is, harmless duplication).

### Why this matters

This directly unblocks building a **user audit log feature for the dashboard**: to show "which user did what" (e.g. who created an API key, who exported a report, who changed a config), we need to filter/group `api_events` by `user_id` directly in SQL, which wasn't possible while it was buried inside a JSON blob string. With this change, `user_id` is a first-class, directly queryable column for every dashboard/JWT-authenticated action, while correctly remaining `NULL` for merchant API-key-authenticated traffic (which has no associated user).

### Verification

Tested locally end-to-end:
- Dashboard (JWT-authenticated) calls now show a real `user_id` at the top level of the emitted `api_events` Kafka message / ClickHouse row, matching the value already present in `authentication_data.user_id`.
- Merchant API-key-authenticated calls (e.g. `CustomersList`, `CustomersCreate`, `PaymentsList`) correctly show `user_id` as `NULL` at the top level, since `AuthenticationType::ApiKey` carries no user id at all.

muditbhutani:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

`AuthenticationType` is serialized with `#[serde(tag = "api_auth_type", content = "authentication_data")]` (adjacently tagged). When it is flattened into `ApiEvent`, the authenticated dashboard user's ID remains nested inside the `authentication_data` JSON blob instead of being available as a directly queryable actor field.

The existing top-level `user_id` column cannot be reused because it is also populated by `ApiEventsType` for resource-specific user IDs. This PR therefore introduces a separate `auth_user_id` field for the authenticated actor.

### Add `auth_user_id` as a directly queryable actor field

This PR:

- Adds `AuthenticationType::get_user_id(&self) -> Option<String>` to extract the authenticated user ID independently of the enum's serde representation.
- Adds a plain, non-flattened `auth_user_id: Option<String>` field to `ApiEvent`.
- Populates `auth_user_id` at the `ApiEvent::new` call site in `services/api.rs`.
- Adds `auth_user_id Nullable(String)` to `api_events_queue` and `api_events`.
- Updates `api_events_mv` to copy `auth_user_id` from the Kafka queue into `api_events`.

The existing `user_id` field remains unchanged and continues to represent resource-specific user IDs where applicable. `auth_user_id` represents the authenticated user who initiated the request.

### Additional Changes

## Motivation and Context

Fixes #13461

This is a prerequisite for a dashboard user activity log. To show which authenticated user performed an action—such as creating an API key, exporting a report, or changing a configuration—`api_events` must expose the actor ID as a dedicated, directly queryable column.

Related to, but independent of, #13458 / #13459, which fixed a specific code path where the user ID was discarded before it reached `AuthenticationType`.

## How did you test it?

Tested locally end-to-end against a local Kafka + ClickHouse `api_events` pipeline:

- Dashboard JWT-authenticated requests populate `auth_user_id` with the same user ID present in `authentication_data.user_id`.
- Merchant API-key-authenticated requests such as `CustomersList`, `CustomersCreate`, and `PaymentsList` leave `auth_user_id` as `NULL`, because `AuthenticationType::ApiKey` has no associated dashboard user.
- The existing top-level `user_id` field remains independent from `auth_user_id`.

Also confirmed against production ClickHouse that the actor ID was previously present inside `authentication_data` but was not available as a dedicated top-level actor column.

```sql
SELECT
    created_at,
    api_flow,
    api_auth_type,
    auth_user_id,
    user_id AS resource_user_id,
    JSONExtractString(authentication_data, 'user_id') AS nested_auth_user_id
FROM api_events
ORDER BY created_at DESC
LIMIT 20
```

## Checklist

- [x] I formatted the code with `cargo +nightly fmt --all`
- [x] I addressed lints reported by `cargo clippy`

Screenshot 2026-07-29 at 7 54 51 PM


XyneSpaces:
Automated review finding:

[should-fix] Duplicate `user_id` key in `ApiEvent` serialization
`ApiEvent` already flattens `auth_type: AuthenticationType`, and several of its variants already include a `user_id` field. Adding a top-level `user_id: Option<String>` introduces a duplicate JSON key in the event schema. Drop the top-level field and read it from the flattened `auth_type`, or rename it (e.g., `auth_user_id`) to keep the schema unambiguous.

[possible issue — worth verifying] Plain-string user identifier in Kafka event
The new `user_id` is stored as `Option<String>` in `ApiEvent`. If this event is shipped to Kafka/other sinks, consider whether it should be wrapped in the project's masking type (e.g., `Option<Secret<String>>`) so the identifier is not emitted verbatim in logs.

crates/router/src/events/api_logs.rs
```diff
@@ -34,6 +34,7 @@ pub struct ApiEvent {
     status_code: i64,
     #[serde(flatten)]
     auth_type: AuthenticationType,
+    user_id: Option<String>,
```
ThisIsMani: Rename this to `auth_user_id` to avoid conflicts with the existing one.

crates/router/src/events/api_logs.rs
```diff
@@ -67,6 +68,7 @@ impl ApiEvent {
         http_method: &http::Method,
         infra_components: Option<serde_json::Value>,
     ) -> Self {
+        let user_id = auth_type.get_user_id();
```
ThisIsMani: Take this as parameter instead of extracting here.

crates/router/src/events/api_logs.rs
muditbhutani: Renamed to `auth_user_id` (06d2bf1). Kept `#[serde(rename = "user_id")]` on it though — the `api_events_queue` Kafka table ingests via `JSONEachRow`, which maps JSON keys to ClickHouse columns by exact name, and the column is called `user_id`. If we rename the wire key too, the `user_id` column goes back to never being populated (verified 0% populated today across all JWT auth types on prod, which is the exact bug this PR fixes), and `auth_user_id` would just be an unmapped/dropped key. So: Rust identifier renamed for clarity, serialized key intentionally left as `user_id` to match the existing column. Let me know if you'd actually want the column itself renamed — that would need a separate schema migration.

crates/router/src/events/api_logs.rs
muditbhutani: Done (06d2bf1) — moved `auth_type.get_user_id()` out of `ApiEvent::new` and into the call site in `services/api.rs` (before `auth_type` is moved into the constructor), passing it in as a new `auth_user_id` parameter.

crates/router/src/events/api_logs.rs
ThisIsMani: The current user_id column is being populated from the ApiEventMetrics which contains user_id as a column. If an API implements it, this will have conflicts.

ThisIsMani (dismissed):

VenuMadhav2541 (approved):

VenuMadhav2541 (changes requested):
`auth_user_id` is added only to `api_events`, but API-event log retrieval uses `api_events_audit` / `api_events_payout_audit` (`AnalyticsCollection::ApiEvents` maps to `api_events_audit`). Since the audit tables and their materialized views are not updated, payment/payout API log queries will still drop the authenticated user id. Please add `auth_user_id` to the audit tables/MVs and to `ApiLogsResult` if this should be returned by the API.

hrithikesh026 (approved):

VenuMadhav2541 (approved):

JeevaRamu0104 (approved):

apoorvdixit88 (approved):

PR #13464, opened 2026-07-27
perf(core): Make locker call async in case of mandates

AnuthaDev:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

### Additional Changes

## Checklist

### DIFF
--- a/crates/router/src/core/payments/operations/payment_response.rs
+++ b/crates/router/src/core/payments/operations/payment_response.rs
@@ -695,9 +695,7 @@ impl<F: Send + Clone> PostUpdateTracker<F, PaymentData<F>, types::PaymentsAuthor
                 ..
             } = sync_response;
 
-            payment_data.payment_method_info = if let Some(payment_method_id) =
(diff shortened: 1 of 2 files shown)


XyneSpaces:
[blocking] `crates/router/src/core/payments/tokenization.rs:1596` should return the original error when connector/locker tokenization fails instead of `Ok` with an empty `SavePaymentMethodDataResponse`. Callers treat this as success, so a failed tokenization can continue with no `payment_method_id` or connector mandate reference.

[should-fix] `crates/router/src/core/payments/tokenization.rs:1797` should not only log DB failures while updating the async locker id. If the payment method lookup/update fails, the locker id and network-token mapping can be permanently lost with no retry path.

AnuthaDev:
Mandates need to be synchronous. This is a domain behaviour, we can't make it async

PR #13472, opened 2026-07-27
feat(observability): add context aware database connections

### LINKED ISSUE #13501: feat(observability): emit per-query database latency and error events
## Problem

Database calls dominate the request path for payments, refunds, disputes, and payouts, but none of that time is currently attributable. A request trace shows total latency with no breakdown of how much came from Postgres, and there's no way to tell a slow filtered scan from a cheap primary-key read, or to compare error rates across tables and operations.

## Requirement

Wrap query execution in the shared `diesel_models` query layer so that each database call reports an `ExternalServiceCall` event with:
- the table it ran against
- the kind of operation (insert, update, delete, single-row find, filtered read, count, ...)
- how long it took
- whether it succeeded
- the request ID it belongs to, so it can be joined against the rest of that request's events

Calls made outside of a request — the drainer, scheduled jobs, and similar background work — carry no request ID and should emit nothing, since there is no API request to correlate them to.

This follows the same event shape and delivery path (Kafka → ClickHouse) as the Redis instrumentation in #13090, so database and Redis latency end up queryable and joinable the same way.

## Acceptance Criteria

- Every query built through the generic query helpers (`crates/diesel_models/src/query/generics.rs`) reports its latency and outcome as an event.
- The smaller set of hand-written queries outside that shared layer (aggregate/count queries in payment attempts, payouts, disputes, refunds) do the same.
- Reporting isn't opt-in per call site: the connection type every query function takes already carries the request ID and the event sink, so a query can't be run without that context being available.
- Background connections (no request in flight) produce no events.
- Current database test coverage keeps passing.

spritianeja03:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [x] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

This PR introduces explicit, request-scoped database connection context required for database-call observability. Following review feedback, request correlation is carried through contextual connection wrappers instead of task-local storage. The PR is intentionally limited to this foundation so that the connection hierarchy and call-site migration can be reviewed before per-query event emission is added.

### Key Changes

1. **Database context abstraction**

   - Adds `DatabaseEventContext`, which carries the request ID and external-service event emitter
   - Adds the `DatabaseConnection` trait, exposing the raw Diesel connection and its associated event context
   - Updates model-layer connection arguments to accept the contextual database connection abstraction

2. **Context-aware PostgreSQL pools and connections**

   - Changes `PgPool` from a raw BB8 type alias into a wrapper containing the raw pool and event emitter
   - Keeps shared pools request-agnostic and snapshots the request ID into `DatabaseConnectionWithContext` when a connection is checked out
   - Adds `DatabaseTransactionConnectionWithContext` for transaction-scoped raw connections
   - Adds `DatabaseStoreWithContext` to provide read, write, accounts-read, and accounts-write connections from request-scoped stores

3. **Explicit request-context propagation**

   - Avoids task-local storage; `RouterStore` and `KVRouterStore` provide request context explicitly during connection checkout
   - Preserves the same event context when entering Diesel transactions
   - Updates router, storage, and model call sites to unwrap the raw connection only at the Diesel execution boundary

4. **Background-worker compatibility**

   - Updates drainer to use the shared contextual PostgreSQL connection types
   - Drainer connections use a no-op event emitter and no request ID because they execute outside an API request

### Deferred to the follow-up implementation

- Wiring database event emission into `track_database_call`
- Emitting per-query latency, success, and error information
- Adding database operation and table metadata to emitted events
- Instrumentation-specific tests

### Additional Changes

## Motivation and Context

Database calls currently receive raw pooled connections, so the model layer cannot access request-scoped observability information. This change establishes an explicit connection hierarchy that carries the request ID and event emitter from the router store to the database execution boundary without relying on task-local state, while retaining support for transactions and background workers.

Splitting the work at this boundary allows the contextual design and call-site impact to be reviewed before database-call tracking behavior is introduced.

Closes #13501 

## How did you test it?

- Ran `cargo check -p router`
- Ran `cargo check -p drainer --features v1`
- Ran `cargo fmt -p drainer -- --check`
- Ran `git diff --check`

## Checklist

- [x] I formatted the code


crates/storage_impl/src/utils.rs
```diff
@@ -26,36 +20,22 @@ pub async fn pg_connection_read<T: DatabaseStore>(
...
-    let pool = store.get_master_pool();
-
-    pool.get()
-        .await
-        .change_context(StorageError::DatabaseConnectionError)
+    store.get_write_connection().await
 }
 
-pub async fn pg_connection_write<T: DatabaseStore>(
+pub async fn pg_connection_write<T: DatabaseStoreWithContext>(
```
ThisIsMani: We can try to remove DatabaseStoreWithContext and use DatabaseStore + RequestContext.

crates/storage_impl/src/database/store.rs
```diff
@@ -1,20 +1,106 @@
...
+
+    pub async fn get_without_context(
+        &self,
+    ) -> Result<DatabaseConnectionWithContext, bb8::RunError<ConnectionError>> {
+        self.get_with_request_id(None).await
+    }
+
+    async fn get_with_request_id(
+        &self,
+        request_id: Option<String>,
```
ThisIsMani: Should be RequestContext.

crates/diesel_models/src/query/user/sample_data.rs
```diff
@@ -30,7 +30,7 @@ pub async fn insert_payment_intents(
     logger::debug!(query = %debug_query::<diesel::pg::Pg,_>(&query).to_string());
 
     query
-        .get_results_async(conn)
+        .get_results_async(conn.raw_connection())
         .await
         .change_context(errors::DatabaseError::Others)
         .attach_printable("Error while inserting payment intents")
```
ThisIsMani, line 36: I assume these won't be tracked since `track_database_call` is not being called?

crates/diesel_models/src/lib.rs
```diff
@@ -65,12 +65,64 @@ pub mod user_authentication_method;
...
+    pub fn request_id(&self) -> Option<&str> {
+        self.request_id.as_deref()
+    }
+
+    pub fn event_emitter(&self) -> &dyn ExternalServiceEventEmitter {
+        self.event_emitter.as_ref()
+    }
+}
+
+pub type PgPooledConn = DatabaseConnectionWithContext;
```
ThisIsMani: Add a comment that we are doing this to reduce blast radius.

crates/drainer/src/health_check.rs
```diff
@@ -119,15 +119,21 @@ impl HealthCheckInterface for Store {
...
         let conn = pg_connection(&self.master_pool).await;
 
-        conn
-            .transaction_async(|conn| {
+        // The connection handed to the closure is another handle to the connection `conn` already
+        // holds, so queries issued through `conn` run within this transaction.
+        let conn = &conn;
+        conn.raw_connection()
+            .transaction_async(move |_| {
                 Box::pin(async move {
```
ThisIsMani, line 127: I assume transactions also won't be tracked?

crates/drainer/src/connection.rs
```diff
@@ -1,11 +1,9 @@
-use bb8::PooledConnection;
 use common_utils::DbConnectionParams;
 use diesel::PgConnection;
+use storage_impl::database::store::{DatabaseConnectionWithContext, PgPool};
 
 use crate::{settings::Database, Settings};
 
-pub type PgPool = bb8::Pool<async_bb8_diesel::ConnectionManager<PgConnection>>;
-
```
ThisIsMani: Looks like drainer was defining it's own PgPool previously, should we keep it same instead of adding storage_impl as a dependency?

ThisIsMani: Another option is to keep the drainer as is, as in the same types that it had previously since we are not going to emit events from drainer anyways.

crates/router/src/db/health_check.rs
```diff
@@ -23,39 +23,47 @@ impl HealthCheckDbInterface for Store {
...
-            let query = diesel::select(diesel::dsl::sql::<diesel::sql_types::Integer>("1 + 1"));
-            let _x: i32 = query.get_result_async(&conn).await.map_err(|err| {
-                logger::error!(read_err=?err,"Error while reading element in the database");
-                errors::HealthCheckDBError::DBReadError
-            })?;
+        // The connection handed to the closure is another handle to the connection `conn` already
+        // holds, so queries issued through `conn` run within this transaction.
+        let conn = &conn;
+        conn.raw_connection()
+            .transaction_async(move |_| async move {
```
ThisIsMani, line 30: We could see if it's possible to wrap this in track_database_call...

crates/drainer/src/connection.rs
ThisIsMani: I think it's fine. Let's keep it this way for now... Previously we were defining the same types in router and drainer, this would make it uniform. I using storage_impl as dependency is correct.

crates/router/src/connection.rs
```diff
@@ -1,44 +1,15 @@
...
-    }
-}
+pub type PgPooledConn = diesel_models::PgPooledConn;
 
-pub async fn pg_connection_read<T: storage_impl::DatabaseStore>(
+pub async fn pg_connection_read<T: DatabaseStore + RequestContext>(
     store: &T,
 ) -> errors::CustomResult<
-    PooledConnection<'_, async_bb8_diesel::ConnectionManager<PgConnection>>,
+    storage_impl::database::store::DatabaseConnectionWithContext,
```
ThisIsMani: NIT: Too much qualification.

crates/diesel_models/src/query/user/sample_data.rs
spritianeja03: yes, track_database_call is not being called in a lot of places. Should we go ahead and add it wherever possible or just leave it as is ?

crates/router/src/connection.rs
```diff
@@ -57,21 +56,25 @@ pub async fn pg_connection_read<T: storage_impl::DatabaseStore>(
     let pool = store.get_master_pool();
 
     #[cfg_attr(not(feature = "deja"), allow(unused_mut))]
-    let mut conn = pool
-        .get()
+    let mut connection = pool
+        .pg_pool
+        .get_owned()
```
ThisIsMani: Let's use get.

crates/storage_impl/src/payments/payment_attempt.rs
```diff
@@ -622,12 +622,7 @@ impl<T: DatabaseStore> PaymentAttemptInterface for RouterStore<T> {
         card_discovery: Option<Vec<common_enums::CardDiscovery>>,
         _storage_scheme: MerchantStorageScheme,
     ) -> CustomResult<i64, errors::StorageError> {
-        let conn = self
-            .db_store
-            .get_replica_pool()
-            .get()
-            .await
-            .change_context(errors::StorageError::DatabaseConnectionError)?;
+        let conn = pg_connection_read(self).await?;
```
ThisIsMani: Need to change this to get only replica.

ThisIsMani (approved):

ThisIsMani (dismissed):

jarnura (dismissed):

Sakilmostak (dismissed):

Sakilmostak (dismissed):

jarnura (dismissed):

ThisIsMani (dismissed):

ThisIsMani (approved):

jarnura (approved):

Sakilmostak (approved):

PR #13481, opened 2026-07-28
refactor(api_models): use Connector enum for Apple Pay session token connector field

### LINKED ISSUE #13123: [FIX] add variant for ApplepaySessionTokenResponse.connector instead of string
```rs
pub struct ApplepaySessionTokenResponse {
    /// Session object for Apple Pay
    /// The session_token_data will be null for iOS devices because the Apple Pay session call is skipped, as there is no web domain involved
    #[serde(skip_serializing_if = "Option::is_none")]
    #[smithy(value_type = "Option<ApplePaySessionResponse>")]
    pub session_token_data: Option<ApplePaySessionResponse>,
    /// Payment request object for Apple Pay
    #[smithy(value_type = "Option<ApplePayPaymentRequest>")]
    pub payment_request_data: Option<ApplePayPaymentRequest>,
    /// The session token is w.r.t this connector
    #[smithy(value_type = "String")]
    pub connector: String, // make it as variant
    /// Identifier for the delayed session response
    #[smithy(value_type = "bool")]
    pub delayed_session_token: bool,
    /// The next action for the sdk (ex: calling confirm or sync call)
    #[smithy(value_type = "SdkNextAction")]
    pub sdk_next_action: SdkNextAction,
    /// The connector transaction id
    #[smithy(value_type = "Option<String>")]
    pub connector_reference_id: Option<String>,
    /// The public key id is to invoke third party sdk
    #[smithy(value_type = "Option<String>")]
    pub connector_sdk_public_key: Option<String>,
    /// The connector merchant id
    #[smithy(value_type = "Option<String>")]
    pub connector_merchant_id: Option<String>,
}
```

Utkal059:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [x] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

`ApplepaySessionTokenResponse.connector` was a `String`, even though the value always
comes from the `Connector` enum. `ConnectorData.connector_name` is already a
`types::Connector`, and `session_flow.rs` was calling `.to_string()` on it purely to fit
the `String` field.

This changes the field to `api_enums::Connector` and drops that conversion.

Updated construction sites:

- `session_flow.rs` — `create_apple_pay_session_response` now takes `types::Connector`;
  both callers no longer call `.to_string()`.
- `bluesnap`, `payme`, `trustpay` transformers — the hardcoded strings `"bluesnap"`,
  `"payme"`, `"trustpay"` are now the matching enum variants.
- `braintree` transformer — this one used `data.connector.clone()` rather than a literal,
  so it is now `Connector::Braintree`. Same value at runtime, since that code is only
  reachable from braintree, but flagging it as the one change that isn't a direct
  substitution.
- `unified_connector_service/transformers.rs` — here the connector arrives from protobuf
  as a `String`, so it is parsed with `Connector::from_str` and mapped to
  `ParsingFailed`, matching how `Currency` and `CountryAlpha2` are already handled in
  that file.

**The API response is unchanged.** `Connector` uses `#[serde(rename_all = "snake_case")]`
and `#[strum(serialize_all = "snake_case")]`, and each of the 8 variants with a custom
name has both a `#[serde(rename = "...")]` and a matching `#[strum(serialize = "...")]`.
I checked all 150 variants — what serde writes is identical to what `.to_string()` used
to produce, so existing clients see the same JSON.

Worth noting: the struct also derives `Deserialize`, so a connector value that isn't a
known variant will now be rejected instead of accepted as an arbitrary string. That looks
like the intent of the issue, and I couldn't find a path that deserializes this struct
from stored data, but happy to revisit if that's a concern.

### Additional Changes

The OpenAPI specs were regenerated with
`cargo run -p openapi --features v1 && cargo run -p openapi --features v2`. The only
change in each file is:

```diff
 "connector": {
-  "type": "string",
-  "description": "The session token is w.r.t this connector"
+  "$ref": "#/components/schemas/Connector"
 }
```

The description disappearing is just how utoipa renders scalar enum fields here — the
same is true of `MerchantConnectorCreate.connector_name` and `GsmCreateRequest.connector`.

## Motivation and Context

Closes #13123.

## How did you test it?

Two tests, matching what was asked for in review:

1. `ucs_apple_pay_session_connectors_parse_into_connector_enum`
   (`crates/router/src/core/unified_connector_service/transformers.rs`) — the UCS path is the
   only one that parses this field rather than constructing it, so this pins the connectors
   that actually build an `ApplepayClientAuthenticationResponse` in
   juspay/connector-service against `Connector::from_str`. There are two, and both already
   send the snake_case hyperswitch name: braintree (`BRAINTREE_CONNECTOR_NAME`, i.e.
   `"braintree"`) and trustpay (`"trustpay"`). It also asserts each string is the name
   hyperswitch serialises the variant as, so it fails if either side drifts.

2. `connector_serializes_exactly_as_it_displays`
   (`crates/common_enums/src/connector_enums.rs`) — walks every `Connector` variant and
   asserts `serde_json::to_value(connector)` equals `connector.to_string()`, the exact call
   this field used to be built from, and that the shared spelling is snake_case. That is
   what makes the response JSON byte-identical to before. It lives next to the enum because
   `Display` and serde are configured independently (`strum(serialize_all)` vs
   `serde(rename_all)`, plus a `strum(serialize)`/`serde(rename)` pair on each of the 8
   renamed variants), so that is where the invariant can break.

Also verified by hand across all 150 variants: no variant name contains consecutive
capitals, so heck (which strum uses) and serde cannot disagree on the snake_case
conversion, and every one of the 8 renamed variants carries both attributes with the same
value.

Other checks:

- `cargo +nightly fmt --all --check` passes.
- `api_models` builds under both `v1` and `v2`.
- OpenAPI regeneration succeeds for v1 and v2, and produces only the one hunk shown above
  in each file — no unrelated drift.

I wasn't able to run the full `cargo check` / `cargo clippy` matrix locally as the
workspace needs more memory than my machine has, so I'm relying on CI for those. I'll fix
anything that comes back red quickly.

## Checklist


Nithin1506200 (dismissed):

Utkal059:
Hi @AkshayaFoiger @kashif-m — could one of you take a look when you get a chance?

This is a small type-only refactor for #13123: `ApplepaySessionTokenResponse.connector`
moves from `String` to the `Connector` enum, which lets `session_flow.rs` drop the
`.to_string()` on `ConnectorData::connector_name`.

Two things are blocking it:

- The workflows are still awaiting maintainer approval, so no CI has run yet. I wasn't
  able to run the full `cargo check` / `clippy` matrix locally (the workspace needs more
  memory than my machine has), so CI is the only real signal on this one — approving the
  runs would help a lot.
- It needs a code owner review. There's currently one approval from @Nithin1506200.

On the `M-api-contract-changes` label: the response JSON is unchanged. `Connector` is
`snake_case` for both serde and strum, and each of the 8 variants with a custom name has
a matching `#[serde(rename = "...")]` and `#[strum(serialize = "...")]`, so all 150
variants serialize exactly as `.to_string()` did. The only behavioural difference is on
the `Deserialize` side, which will now reject an unknown connector string — that seemed
like the intent of the issue, but happy to revisit if there's a path I missed.

Happy to rebase on latest main if that helps.

Nithin1506200:
## test case
```json
{
    "payment_id": "pay_NMJUX9cZ5GLMxrn6ONHW",
    "client_secret": "pay_NMJUX9cZ5GLMxrn6ONHW_secret_4aldg0jQEFFeIfNneJJh",
    "session_token": [
        {
            "wallet_name": "apple_pay",
            "session_token_data": {
                "epochTimestamp": 1786347274135,
                "expiresAt": 1786350874135,
                "merchantSessionIdentifier": "SSH***27C24",
                "nonce": "06dbf4d8",
                "merchantIdentifier": "109A****D25F",
                "domainName": "hyperswitch-demo-store.netlify.app",
                "displayName": "applepay",
                "signature": "308006092a********0000000000",
                "operationalAnalyticsIdentifier": "applepay:109A289660******AD25F",
                "pspId": "109*****5F"
            },
            "payment_request_data": {
                "country_code": "US",
                "currency_code": "GBP",
                "total": {
                    "label": "applepay",
                    "type": "final",
                    "amount": "1.23"
                },
                "merchant_capabilities": [
                    "supports3DS"
                ],
                "supported_networks": [
                    "visa",
                    "masterCard",
                    "amex",
                    "discover"
                ],
                "merchant_identifier": "merchant.com.stripe.sang"
            },
            "connector": "worldpayxml", // --------
            "delayed_session_token": false,
            "sdk_next_action": {
                "next_action": "confirm",
                "should_block_confirm": null
            },
            "connector_reference_id": null,
            "connector_sdk_public_key": null,
            "connector_merchant_id": null
        }
    ],
    "vault_details": null
}
```

gopikrishna000 (dismissed):

Utkal059:
Thanks for the review and for running that test, @Nithin1506200 @gopikrishna000 🙏

I've merged latest `main` in, so the head is now `a9c34e7`. Two things I can't do from my side:

- The CI workflows on the new commit are stuck in "action required" — they need a maintainer to approve the runs.
- The merge box still says `REVIEW_REQUIRED`. CODEOWNERS points `crates/` and `api-reference/` at @juspay/hyperswitch-framework, and the whole diff sits in those paths, so it needs an approval from that team.

Heads up that `Spell check` will likely go red again once CI runs — it's the repo-wide `typos` scan, currently failing on every open PR, and nothing it flags is in this diff. The required `Spell check specific critical directories` passes.

crates/router/src/core/unified_connector_service/transformers.rs
```diff
@@ -5803,7 +5803,13 @@ impl transformers::ForeignTryFrom<payments_grpc::ClientAuthenticationTokenData>
                         .as_ref()
                         .map(ApplePayPaymentRequest::foreign_try_from)
                         .transpose()?,
-                    connector: apay_session_token_response.connector.clone(),
+                    connector: common_enums::connector_enums::Connector::from_str(
+                        &apay_session_token_response.connector,
```
hrithikesh026, line 5844: this should be an enum too.

Utkal059: Agreed it should be — but this field isn't ours. `apay_session_token_response` comes from `unified-connector-service-client` (juspay/connector-service, tag 2026.08.07.1), and its proto declares `string connector = 3;`.

That proto already defines `enum Connector`, so making it `Connector connector = 3;` would match how Currency / CountryAlpha2 / SdkNextAction already arrive here as proto enums. But it needs a change in connector-service plus a tag bump on this side, and string → enum is a breaking wire change, so it can't happen inside this PR.

Happy to raise it there. Would you rather this land with the boundary parse meanwhile, or hold it?

hrithikesh026: Idk if this is a possibility. there may exist a enum variant in UCS(unified connector service) which is not present in HS(Hyperswitch). In that case, this logic will definitely fail. 

Secondly, Connector::from_str() expects a `snake_case` format.  But UCS maynot use the same serialization format.

cc: @Nithin1506200

Nithin1506200:
```code
there may exist a enum variant in UCS(unified connector service) which is not present in HS(Hyperswitch). In that case, this logic will definitely fail.
```
Yes, there may be connector enum variants in UCS that are not present in Hyperswitch. But that path should not occur here, because Hyperswitch only calls UCS for connectors selected/configured within Hyperswitch.

  
```code
Secondly, Connector::from_str() expects a snake_case format. But UCS maynot use the same serialization format.
```
On the serialization format: `Connector::from_str()` expects the Hyperswitch snake_case connector name. If UCS  returns a different format, failing at this boundary is still better than passing an invalid connector through to the SDK response and failing later during confirm.


-------
  I agree that the ideal fix is to make this field an enum in UCS as well, with explicit HS <-> UCS connector conversion, and throw errors if there is unknown connector or enum mismatch. That is a broader connector-service/proto change, so we can scope it to a follow-up PR and audit other places where UCS still uses `String` for connector and look why the decision was made.

- In either case this PR has safe changes, which the api schema expects proper routable enum - connector
- @Utkal059  you can add tow test case here 
1. Run ucs connectors which implements session for applePay and it doesn't fail in HS during this conversion. Ideally it should be snake case ! 
2. Ensuring the string conversion output from previous and your current changes are having correct Case Sensitive representation.

Nithin1506200: HI @Utkal059 , can you raise a pr in UCS making similar changes of connector Enum here. We will get it merged and you can bump it here later

Utkal059: Sure @Nithin1506200, I'll raise the connector enum PR in UCS shortly and link it here. Once it's merged I'll bump the dependency in this PR. Also adding the two test cases you mentioned.

Utkal059:
Thanks @Nithin1506200 @hrithikesh026 — both of the tests you asked for are in, and the UCS PR is up: juspay/hyperswitch-prism#2128.

### 1. UCS connectors that serve an Apple Pay session don't fail this conversion

`ucs_apple_pay_session_connectors_parse_into_connector_enum` in
`crates/router/src/core/unified_connector_service/transformers.rs`.

Going through connector-service, exactly two connectors build an
`ApplepayClientAuthenticationResponse` today, and both already send the snake_case
hyperswitch name:

| connector | what UCS sends | where |
|---|---|---|
| braintree | `"braintree"` | `BRAINTREE_CONNECTOR_NAME` — `connectors/braintree/transformers.rs:44` |
| trustpay | `"trustpay"` | literal — `connectors/trustpay/transformers.rs:2488` |

The test pins that table against `Connector::from_str`, and also asserts each string is the
name hyperswitch itself serialises the variant as, so it fails if either side drifts.

Worth noting the other `connector` fields on that same message are Google Pay and PayPal,
not Apple Pay — `BRAINTREE_CONNECTOR_NAME` is used at three sites in braintree and only the
first is the Apple Pay one.

### 2. Case-sensitive representation is unchanged

`connector_serializes_exactly_as_it_displays` in
`crates/common_enums/src/connector_enums.rs`.

It walks every `Connector` variant and asserts `serde_json::to_value(connector)` equals
`connector.to_string()` — the exact call the field used to be built from — and that the
shared spelling is snake_case. I put it next to the enum rather than in the UCS file
because `Display` and serde are configured independently there (`strum(serialize_all)` vs
`serde(rename_all)`, plus a `strum(serialize)`/`serde(rename)` pair on each of the 8
renamed variants), so that's where the invariant can actually break.

That is what makes the response JSON byte-identical to before, which is the claim in the
PR description — now checked by CI instead of by hand.

### 3. UCS enum change

> can you raise a pr in UCS making similar changes of connector Enum here

Done — juspay/hyperswitch-prism#2128. It moves
`ApplepayClientAuthenticationResponse.connector` to the proto `Connector` enum, makes the
domain field `ConnectorEnum`, and adds an explicit `ConnectorEnum -> proto Connector`
conversion that errors instead of falling back to `CONNECTOR_UNSPECIFIED`.

Two things to flag there:

- It's a wire-breaking change on field 3, so `buf breaking` will go red and it needs the
  `proto-breaking-approved` label. The additive alternative (new field + deprecate the
  string) is written up in that PR if you'd rather go that way.
- While checking the mapping was total I found `ConnectorEnum::RazorpayV2` and
  `ConnectorEnum::Affirm` have no variant in the proto `Connector` enum, so
  `grpc_connector_from_connector_enum` silently maps them to `CONNECTOR_UNSPECIFIED`
  today. Neither serves Apple Pay, so nothing regresses, but it looked worth raising.

@hrithikesh026 on your two points — you were right that nothing guarantees the formats
match. Both UCS Apple Pay connectors happen to send snake_case today, and test 1 pins that
so it can't drift silently; the UCS PR removes the string from the wire entirely. On an
enum variant existing in UCS but not HS: the failure is a `ParsingFailed` on the session
call rather than an invalid connector reaching the SDK, which is the tradeoff
@Nithin1506200 described.

Once the UCS PR is merged and tagged I'll bump `unified-connector-service-client` here and
read the enum directly, which drops the `from_str` at the boundary.

Utkal059:
Hi @Nithin1506200, @hrithikesh026, and @gopikrishna000 👋 Just a quick gentle ping on this one!

I've just updated the branch with the latest main. A couple of quick items to help unblock this:

CI Runs: The CI workflows are currently in "action required" state and need a maintainer to approve the runs. Could one of you trigger them when you have a moment?

Next Steps: The related UCS PR (juspay/hyperswitch-prism#2128) is up. Do you want to wait for that to be merged and tagged so I can bump the unified-connector-service-client dependency in this PR? Or would you prefer to merge this PR now (with the safe boundary parse tests passing) and do the dependency bump in a follow-up PR?

Happy to proceed whichever way works best for the team. Thanks!

Utkal059:
Status so far: two approvals (@Nithin1506200, @gopikrishna000), but neither counts as a code owner. CODEOWNERS points `crates/` and `api-reference/` at @juspay/hyperswitch-framework, and the whole diff is in those paths, so this needs one review from that team specifically. The 8 workflows are also still in "action required" and have never run.

@Nithin1506200 on the sequencing you proposed — the UCS PR is up and its code side is done: juspay/hyperswitch-prism#2128, Compilation Check and Clippy green. It's waiting on the `proto-breaking-approved` label plus code owner review. Once it lands and gets tagged I'll bump `unified-connector-service-client` here and drop the `from_str` at the boundary.

Happy either way on ordering: hold this for the bump, or merge the boundary parse now (it's covered by the two tests you asked for) and do the bump in a follow-up.

PR #13471, opened 2026-07-27
feat(superposition): moved superposition configs to folder based rather than plaintext for RBAC

sahkal:
## Type of Change

- [x] Refactoring

## Description

moved superposition configs to folder based rather than plaintext for RBAC

### Additional Changes

- [ ] This PR modifies the API contract
- [ ] This PR modifies the database schema
- [ ] This PR modifies application configuration/environment variables

### **Overview**
This PR updates dynamic configuration management with **Superposition** by introducing **foldered (namespaced / dot-delimited) keys** for grouping configurations (e.g., `payments.requires_cvv`, `vaulting.fingerprint_secret`) and adding a **graceful fallback mechanism to legacy keys** to prevent downtime during Superposition workspace migrations.

---

#### 1. **Superposition Client Fallback Logic & Legacy Key Resolution**
* **File:** [`crates/external_services/src/superposition.rs`](file:///Users/sahkal.poddar/hyperswitch/crates/external_services/src/superposition.rs#L1021-L1120)
* Added `Send` bound to `Config::Output`.
* Added `legacy_superposition_key()` default method to the `Config` trait, which extracts the trailing key name after the last dot (e.g., `"payments.requires_cvv"` $\rightarrow$ `"requires_cvv"`).
* Updated `Config::fetch` to attempt primary lookup with the foldered key first. If that misses (e.g., before the Superposition workspace is migrated), it automatically retries with the legacy un-foldered key while logging a warning.

#### 2. **Superposition Key Constants Namespacing**
* **File:** [`crates/router/src/consts.rs`](file:///Users/sahkal.poddar/hyperswitch/crates/router/src/consts.rs#L368-L480)
* Prefix-namespaced all Superposition string constants into logical domain folders:
  * **`payments.*`**: `payments.requires_cvv`, `payments.implicit_customer_update`, `payments.should_call_gsm`, `payments.should_perform_eligibility`, `payments.poll_config_external_three_ds`, etc.
  * **`vaulting.*`**: `vaulting.fingerprint_secret`, `vaulting.should_disable_vault_tokenization`, `vaulting.should_perform_sdk_vaulting`, `vaulting.should_trigger_fingerprint_migration`, `vaulting.save_wallet_decrypted_data`.
  * **`payouts.*`**: `payouts.payout_tracker_mapping`, `payouts.max_auto_payout_retries`, `payouts.gsm_payout_call`.
  * **`routing.*`**: `routing.routing_result_source`, `routing.threeds_routing_region_uas`.
[...]

PR #13488, opened 2026-07-29
feat(account_updater): add Superposition gating and credential resolution

kanikabansal08:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

Adds the configuration foundation for the Account Updater flow: the application config section, its startup validation and secret management, and the Superposition keys that will gate the flow. No payment-flow behaviour changes — nothing reads these values yet.

**Application config.** New optional `[account_updater.juspay]` section carrying `base_url`, `api_key`, `merchant_id`, and the three fields for the card-sync call (`euler_encryption_public_key`, `au_decryption_pvt_key`, `card_sync_key_id`). Omitting the section leaves the feature off.

`AccountUpdaterConfig` is an externally tagged enum with a single `Juspay` variant, so the provider is named by the TOML table itself and provider-specific fields never sit on the shared type. Adding a second provider is a new variant plus a new section, with no change to existing keys.

When the section is present it must be complete: `validate` rejects a `base_url` without a trailing slash and any empty required field, at startup, so the eventual call path never re-checks individual fields.

`api_key`, `euler_encryption_public_key` and `au_decryption_pvt_key` are `Secret<String>` and resolved through `SecretsHandler`, so they can be held in the secrets manager rather than in plaintext config.

**Superposition keys**, with defaults in `config/superposition_seed.toml`:

| Key | Type | Default |
| --- | --- | --- |
| `account_updater.enabled` | boolean | `false` |
| `account_updater.credential_source` | `none \| application` | `none` |

Both default off, so an unseeded environment resolves as disabled. Keys are namespaced under `account_updater.` following #13471; the legacy DB config-table keys stay flat, as they do there.

Registering them through `config!` generates `DimensionsGlobal::get_account_updater_enabled` and `get_account_updater_credential_source`, each falling back to the DB-backed config and then the default when Superposition is unavailable. `AccountUpdaterCredentialSource` is a string enum (`none` | `application`) so future org/merchant/profile sources can extend it alongside their credential storage.

### Additional Changes

| File | Change |
| --- | --- |
| `config/config.example.toml` | new `[account_updater.juspay]` section |
| `config/superposition_seed.toml` | the two defaults above |
| `crates/router/src/configs/settings.rs` | `AccountUpdaterConfig` / `JuspayAccountUpdaterConfig` |
| `crates/router/src/configs/validations.rs` | startup validation |
| `crates/router/src/configs/secrets_transformers.rs` | secrets-manager resolution for the three secret fields |
| `crates/router/src/consts.rs` | the two Superposition key constants |
| `crates/router/src/core/configs/dimension_config.rs` | `config!` registrations |
| `crates/router/src/core/account_updater/types.rs` | `AccountUpdaterCredentialSource` |

## Motivation and Context

Card networks reissue cards continuously through expiry updates and replacement PANs, and a stored card silently goes stale until a payment fails. The Account Updater flow checks a stored card against the network before it is charged.

This PR is the configuration layer for that flow, split out so the gating and credential plumbing can be reviewed independently of the call path that will consume it.

## How did you test it?

No automated tests — this PR adds no executable path, only config types, validation and key registrations. `cargo clippy -p router --lib` is clean.

The behaviour to exercise on startup is:

| `[account_updater.juspay]` | Expected |
| --- | --- |
| omitted | boots, feature off |
| present and complete | boots |
| `base_url` without trailing slash | startup error `account_updater.juspay.base_url must end with a trailing slash` |
| any required field empty | startup error `account_updater.juspay.<field> must not be empty` |

The Superposition keys are not seeded in any environment yet, so both resolve to their defaults (disabled).

## Checklist


crates/router/src/configs/settings.rs
```diff
@@ -386,6 +387,16 @@ pub struct OnSessionConfig {
...
 
+#[derive(Debug, Deserialize, Clone)]
+pub struct AccountUpdaterConfig {
+    pub base_url: url::Url,
+    pub api_key: Secret<String>,
+    pub merchant_id: String,
+    pub euler_encryption_public_key: Secret<String>,
+    pub au_decryption_pvt_key: Secret<String>,
+    pub card_sync_key_id: String,
+}
```
ThisIsMani, line 404: Can we make this an enum and put these under JuspayAccountUpdater or something? These are too specific to be put in AccountUpdaterConfig.

ThisIsMani (commented):
Other than that looks good to me.

crates/router/src/configs/secrets_transformers.rs
```diff
@@ -412,6 +412,32 @@ impl SecretsHandler for settings::OidcSettings {
...
+
+        Ok(value.transition_state(|account_updater| {
+            let Self::Juspay(juspay) = account_updater;
+            Self::Juspay(settings::JuspayAccountUpdaterConfig {
+                api_key,
+                euler_encryption_public_key,
+                au_decryption_pvt_key,
+                ..juspay
+            })
+        }))
```
ThisIsMani: Impl this on Juspay Config. Similar to what you did in validation.

config/config.example.toml
```diff
@@ -1517,6 +1517,15 @@ base_url = "https://offer-engine.example.com/" # Base URL of the Offer Engine se
...
 merchant_id = ""                              # Offer Engine merchant id sent in request bodies
 
+[account_updater] # Credentials forwarded to UCS for the Account Updater flow; omit the section entirely to keep it off
+provider = "juspay"                               # Account Updater provider
+base_url = "https://account-updater.example.com/" # Account Updater base URL
+api_key = ""                     # API key for the Account Updater provider
+merchant_id = ""                 # Merchant id sent in Account Updater request bodies
+euler_encryption_public_key = "" # Provider RSA public key, used to seal outbound card data
+au_decryption_pvt_key = ""       # Our RSA private key, used to open the response payload
+card_sync_key_id = ""            # Id of our public key as registered with the provider
```
ThisIsMani, line 1527: This should be moved to `[account_updater.juspay]` I believe.

crates/router/src/consts.rs
```diff
@@ -371,6 +371,11 @@ pub mod superposition {
     pub const OFFER_ENGINE_ENABLED: &str = "offer_engine_enabled";
     /// Offer Engine credential source key: `"none"` skips Offer Engine, `"application"` uses the static app config.
     pub const OFFER_ENGINE_CREDENTIAL_SOURCE: &str = "offer_engine_credential_source";
+    /// Account Updater master gate key: `false` (default) disables all Account Updater calls
+    pub const ACCOUNT_UPDATER_ENABLED: &str = "account_updater_enabled";
+    /// Account Updater credential source key: `"none"` skips Account Updater, `"application"` uses
+    /// the static application config
+    pub const ACCOUNT_UPDATER_CREDENTIAL_SOURCE: &str = "account_updater_credential_source";
```
NISHANTH1221: can we group them under a name similar to how it is done in this pr: #13471

ThisIsMani (approved):

NISHANTH1221 (approved):

jarnura (approved):

PR #13510, opened 2026-07-31
feat(account_updater): add force_sync trigger and UCS refresh call

kanikabansal08:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

Adds a `force_sync` query parameter to the v2 payment method retrieve endpoint. When set, the stored card is evaluated against the Account Updater over UCS.

**This is shadow mode. The refreshed card is discarded.** UCS returns a `RefreshResult` carrying `CardRefreshOutcome`; we read the outcome and nothing else. Even on `ACCOUNT_UPDATED` — "card reissued, a new card is available" — nothing is persisted. No locker write, no `payment_method` update, no fingerprint recompute. The `GET` response is byte-identical whether `force_sync` is set or not. The only output is a log line.

The whole module is behind `#[cfg(feature = "v2")]`.

### Pipeline

`crates/router/src/core/account_updater/`, entered through `account_updater::run_account_updater`:

| Stage | Module | Responsibility |
|---|---|---|
| Gate | `config` | Superposition enable flag + credential source; resolves application credentials |
| Eligibility + unvault | `eligibility` | Payment method must be active and a card, stored network must be Visa/Mastercard; then unvaults the card and builds the UCS `PaymentMethod` |
| Refresh | `refresh` | UCS `PaymentMethodService::Refresh` under an explicit 5s `grpc-timeout` |

`run_account_updater` never propagates a failure to the caller — the outcome is logged as `account_updater_outcome`, and every early exit is logged and swallowed, so `force_sync` can never fail a retrieve.

### Exit paths

Everything funnels into `AccountUpdaterError` (`types.rs`):

- **Gate** — the config resolver returns `None` (feature disabled, or credential source is `none`) and the pipeline never starts; `MissingApplicationConfig` when the source is `application` but the `[account_updater]` section is absent
- **Never reached UCS** — `PaymentMethodNotACard`, `PaymentMethodNotActive`, `UnsupportedNetwork`, `CardUnusable` (unvault failed, or the unvaulted data carries no usable card / network)
- **Reached UCS** — `RefreshCallFailed` (client missing, metadata build failed, transport error or deadline), `RefreshReturnedError` (UCS returned an error, or neither a result nor an error)
- **Success** — a `CardRefreshOutcome`, logged via `as_str_name()`

### Juspay as an Account Updater connector

Account Updater credentials come from application config rather than an MCA, so the plumbing is slightly off the usual path:

- `Connector::Juspay` added to the connector enum, explicitly non-routable and not constructible as a `ConnectorData` (`connector_mapping.rs`, `connector_transformers.rs`, `euclid/enums.rs`, `feature_matrix.rs`); `connector_configs` reports it as "Configured through application config"
- `JuspayMetadata` + a `ConnectorSpecificConfig::Juspay` arm in `unified_connector_service/connector_config.rs`, so the UCS `x-connector-config` header can be built from `HeaderKey` auth plus metadata
- `build_unified_connector_service_auth_metadata_without_mca` in `core/unified_connector_service.rs` — the existing MCA-based builder was split so the auth mapping (`build_connector_auth_metadata`) is shared between the two entry points
- `UnifiedConnectorServiceClient::payment_method_refresh` in `external_services` wraps the `PaymentMethodService::Refresh` RPC with the standard header/timeout/error handling

### Additional Changes

**API contract** — `force_sync` added to `PaymentMethodRetrieveRequest` (`crates/api_models/src/payment_methods.rs`), v2 only. Optional, `#[serde(default)]`, defaults to `false`, so existing callers are unaffected. `retrieve_payment_method` now takes the request struct instead of the single `fetch_raw_detail` flag. Response schema unchanged. Both OpenAPI specs regenerated (`force_sync` in v2, the `juspay` connector enum value in both).

**Runtime gating** — the Superposition keys `account_updater_enabled` and `account_updater_credential_source` are gated to `v2` in `dimension_config.rs`, and the dimensions are built from processor merchant / org / profile at the retrieve call site.

**Already on main** — the `[account_updater]` settings section, its validation and secret transformers, `config/config.example.toml`, and the `unified-connector-service-client` bump that carries the `Refresh` RPC all landed separately and are not part of this diff.

### Access matrix — both query params, both auth types

`force_sync` decides whether the refresh runs (internal auth only). `fetch_raw_detail` decides only what the response carries. They are independent.

| auth | `force_sync` | `fetch_raw_detail` | AU runs | locker reads | CVC consumed | raw details in response |
|---|---|---|---|---|---|---|
| external | false | false | no | 0 | no | none |
| external | false | true | no | 0 | no | none — grant denied by org config |
| external | true | false | no | 0 | no | none |
| external | true | true | no | 0 | no | none — `force_sync` ignored, not rejected |
| internal | false | false | no | 0 | no | none |
| internal | false | true | no | 1 | yes | card + CVC |
| internal | true | false | **yes** | **1** | **no** | none |
| internal | true | true | **yes** | **1** | yes | card + CVC |

Row 7 is the case this PR fixes. The unvault is escalated to `Allowed` so AU gets the card; the response is filtered back to the caller's grant afterwards, so no PAN leaks and the CVC is never touched.

## Motivation and Context

Stored cards go stale — reissued, re-expired, closed. Account Updater asks the networks whether a stored card has changed. This PR builds the full path to that answer and stops short of acting on it, so the gate, eligibility rules, unvaulting, and UCS call can be exercised against real traffic before any write path exists. The logged outcome shows what the write path would have done.

### Not covered

No unit or integration tests added. Outcomes other than `EXPIRY_UPDATED` and `NOT_FOUND` were not exercised against the sandbox.

## Checklist


crates/router/src/core/unified_connector_service/transformers.rs
```diff
@@ -847,6 +847,7 @@ impl
                 .map(payments_grpc::PaymentMethodType::foreign_try_from)
                 .transpose()?
                 .map(|payment_method_type| payment_method_type.into()),
+            order_details: build_ucs_order_details(router_data.request.order_details.as_deref()),
```
kanikabansal08, line 857: this line is collateral from a dependency bump, not an Account Updater change.

api-reference/v2/openapi_spec_v2.json
```diff
@@ -23129,6 +23129,10 @@
         "properties": {
           "fetch_raw_detail": {
             "type": "boolean"
+          },
+          "force_sync": {
+            "type": "boolean",
+            "description": "Trigger an Account Updater check for the stored card."
```
ThisIsMani: Should this description be card specific?

crates/api_models/src/payment_methods.rs
```diff
@@ -116,11 +116,14 @@ pub struct PaymentMethodCreate {
 }
 
 #[cfg(feature = "v2")]
-#[derive(Debug, serde::Deserialize, serde::Serialize, Clone, ToSchema)]
+#[derive(Debug, Default, serde::Deserialize, serde::Serialize, Clone, ToSchema)]
```
ThisIsMani: I don't think we need to add Default here...

crates/router/src/core/account_updater.rs
```diff
@@ -1,3 +1,18 @@
 pub mod types;
 
+#[cfg(feature = "v2")]
```
ThisIsMani: Can't we move account updater itself to v2?

crates/router/src/core/unified_connector_service/connector_config.rs
```diff
@@ -1659,16 +1647,20 @@ pub fn build_connector_config_header(
...
     ))?;
 
-    let config_json = serde_json::to_value(&config)
+    serialize_connector_config(&config).map(Some)
+}
+
+/// Serializes a [`ConnectorSpecificConfig`] into the `{"config": {"ConnectorName": {...}}}`
+/// envelope UCS expects for the connector config header.
+pub fn serialize_connector_config(config: &ConnectorSpecificConfig) -> RouterResult<String> {
+    let config_json = serde_json::to_value(config)
```
ThisIsMani: Do we need these changes?

crates/router/src/core/unified_connector_service/connector_config.rs
```diff
@@ -10,33 +10,12 @@ use common_utils::ext_traits::ValueExt;
...
 };
 
-/// Connector-specific configuration wrapper for UCS.
-/// Serializes as: `{"config": {"ConnectorName": {...}}}`
-#[derive(Debug, Serialize)]
-pub struct UcsConnectorConfig {
-    pub config: serde_json::Map<String, serde_json::Value>,
-}
-
-impl UcsConnectorConfig {
```
ThisIsMani, line 27: This wasn't being used before?

crates/api_models/src/payment_methods.rs
kanikabansal08: we are using this in by `retrieve_payment_method_olap `(core/payment_methods.rs:5627)`PaymentMethodRetrieveRequest::default()`
can switch to explicit literal if you'd prefer it visible

crates/router/src/core/unified_connector_service/connector_config.rs
kanikabansal08: yes, `UcsConnectorConfig` had zero call sites. it was also unusable as written as it was producing , `{"config":{"Juspay":{"Juspay":{...}}}}` , a duplicated tag UCS can't deserialize.

crates/router/src/core/unified_connector_service/connector_config.rs
kanikabansal08: yes, just kept it took this out so that Account Updater can call it. Without the extraction, AU would need its own copy of serialize the config, wrap it under a config key and stringify.

crates/router/src/core/account_updater/eligibility.rs
```diff
@@ -0,0 +1,44 @@
...
+    payment_method: &domain::PaymentMethod,
+) -> Result<EligibleCard, SkipReason> {
+    evaluate_stored_card(
+        payment_method.get_payment_method_type(),
+        payment_method.status,
+        payment_method
+            .payment_method_data
+            .as_ref()
+            .map(|payment_method_data| payment_method_data.get_inner()),
+    )
```
ThisIsMani: Can we do a match on PaymentMethodType and call evaluate_stored_card with card details.

crates/router/src/core/account_updater/eligibility.rs
```diff
@@ -0,0 +1,44 @@
...
+    status: PaymentMethodStatus,
+    payment_method_data: Option<&PaymentMethodsData>,
+) -> Result<EligibleCard, SkipReason> {
+    if payment_method_type != Some(PaymentMethod::Card) {
+        return Err(SkipReason::PaymentMethodNotACard);
+    }
+
+    if status != PaymentMethodStatus::Active {
+        return Err(SkipReason::PaymentMethodNotActive);
+    }
```
ThisIsMani: This is not specific to cards. Can be moved to the above function.

crates/router/src/core/account_updater/evaluation.rs
```diff
@@ -0,0 +1,79 @@
...
+    unvault::fetch_card_for_sync,
+};
+use crate::{
+    core::{configs::dimension_state, metrics},
+    routes::SessionState,
+    types::domain,
+};
+
+#[instrument(skip_all)]
+pub async fn run(
```
ThisIsMani: These function names are bit too generic.

crates/router/src/core/account_updater/unvault.rs
ThisIsMani, line 1: This is a bit too specific to be a separate file imo... Maybe rename to utils?

crates/router/src/core/account_updater/types.rs
```diff
@@ -15,3 +18,98 @@ pub enum AccountUpdaterCredentialSource {
...
+#[strum(serialize_all = "snake_case")]
+pub enum RefreshOutcome {
+    AccountUpdated,
+    ExpiryUpdated,
+    NoChange,
+    Closed,
+    NotFound,
+    ContactIssuer,
+    Unspecified,
+}
```
ThisIsMani: Is it possible to use the existing grpc types instead of creating new one?

crates/router/src/core/payment_methods.rs
```diff
@@ -5508,11 +5509,23 @@ pub async fn retrieve_payment_method(
...
     )?;
 
+    if request.force_sync {
+        Box::pin(account_updater::run(
+            &state,
+            &platform,
+            &profile,
+            &payment_method,
+            storage_type,
+            &dimensions.without_provider_merchant_id(),
```
ThisIsMani: Should have org, merchant and profile as dimensions as well.

crates/router/src/core/unified_connector_service.rs
```diff
@@ -1847,6 +1847,25 @@ fn get_ucs_client(
         })
 }
 
+pub fn build_unified_connector_service_auth_metadata_without_mca(
+    connector_name: String,
+    auth_type: String,
```
ThisIsMani: Construct ConnectorAuthType and pass it here.

crates/router/src/core/unified_connector_service.rs
```diff
@@ -1847,6 +1847,25 @@ fn get_ucs_client(
...
+    processor_merchant_id: &id_type::MerchantId,
+    connector_config: Option<Secret<String>>,
```
ThisIsMani: Don't accept string. Instead take type and serialize it.

crates/router/src/core/account_updater/eligibility.rs
```diff
@@ -0,0 +1,44 @@
+use common_enums::{CardNetwork, PaymentMethod, PaymentMethodStatus};
+use hyperswitch_domain_models::payment_method_data::PaymentMethodsData;
+
+use super::types::{EligibleCard, SkipReason};
+use crate::types::domain;
+
+pub fn evaluate_eligibility(
+    payment_method: &domain::PaymentMethod,
+) -> Result<EligibleCard, SkipReason> {
```
ThisIsMani: Keep the result generic.

ThisIsMani (dismissed):

JeevaRamu0104 (dismissed):

crates/router/src/core/account_updater/eligibility.rs
```diff
@@ -0,0 +1,118 @@
...
+    storage_type: StorageType,
+) -> CustomResult<payments_grpc::PaymentMethod, AccountUpdaterError> {
+    let card_details = match payment_method
+        .payment_method_data
+        .as_ref()
+        .map(|payment_method_data| payment_method_data.get_inner())
+    {
+        Some(PaymentMethodsData::Card(card_details)) => card_details,
+        _ => {
+            return Err(report!(AccountUpdaterError::PaymentMethodNotACard)
```
jarnura: for return with Err use question mark operator ?

crates/router/src/core/account_updater/eligibility.rs
```diff
@@ -0,0 +1,118 @@
...
+                .attach_printable("Stored payment method data holds no card details"))
+        }
+    };
+
+    if !matches!(
+        card_details.card_network,
+        Some(CardNetwork::Visa | CardNetwork::Mastercard)
```
jarnura: Lets not hard code this data in code, instead make as a config. Not a database config. Either superposition or Env

crates/router/src/core/account_updater/eligibility.rs
```diff
@@ -0,0 +1,118 @@
...
+        Some(RawPaymentMethodData::Card(card_details)) => card_details,
+        Some(RawPaymentMethodData::CardWithNT(details)) => details.card_details,
+        Some(RawPaymentMethodData::BankDebit(_) | RawPaymentMethodData::ProxyCard(_)) | None => {
+            return Err(report!(AccountUpdaterError::CardUnusable)
+                .attach_printable("Unvaulted payment method data holds no card"))
+        }
+    };
+
+    let card_number = CardNumber::from_str(&card_details.card_number.get_card_no())
+        .map_err(|_| report!(AccountUpdaterError::CardUnusable))
```
jarnura: Don't discard errors, do change context

crates/router/src/core/account_updater/refresh.rs
```diff
@@ -0,0 +1,94 @@
...
+use unified_connector_service_client::payments as payments_grpc;
+
+use super::types::{AccountUpdaterError, ResolvedAccountUpdaterConfig};
+use crate::{
+    core::unified_connector_service::build_unified_connector_service_auth_metadata_without_mca,
+    routes::SessionState, types::domain,
+};
+
+/// Per-call `grpc-timeout` deadline for the refresh request.
+const REFRESH_TIMEOUT: Duration = Duration::from_secs(5);
```
jarnura: Move this to env

crates/router/src/core/unified_connector_service/connector_config.rs
```diff
@@ -1689,6 +1707,28 @@ impl ForeignTryFrom<(Connector, &ConnectorAuthType, Option<&serde_json::Value>)>
                 }),
                 _ => Err(err("Givepayments requires HeaderKey auth type")),
             },
+            Connector::Juspay => match auth {
+                ConnectorAuthType::HeaderKey { api_key } => {
+                    let juspay_meta = metadata
+                        .map(|meta| {
+                            serde_json::from_value::<JuspayMetadata>(meta.clone())
+                                .map_err(|_| err("Invalid Juspay metadata format"))
```
jarnura: Don't discard errors, do change context

jarnura (changes requested):

jarnura (dismissed):

ThisIsMani (dismissed):

prajjwalkumar17 (dismissed):

JeevaRamu0104 (dismissed):

crates/router/src/core/account_updater/config.rs
```diff
@@ -0,0 +1,75 @@
...
+            store,
+            superposition,
+            dimensions,
+            None,
+        )
+        .await;
+
+    if !enabled {
+        return Ok(None);
+    }
```
ShankarSinghC: Avoids using early return

crates/router/src/core/account_updater/config.rs
```diff
@@ -0,0 +1,75 @@
...
+    let source = configs::fetch_db_config_for_string_enum::<
+        dimension_config::AccountUpdaterCredentialSource,
+        AccountUpdaterCredentialSource,
+    >(store, superposition, dimensions, None)
+    .await
+    .unwrap_or(AccountUpdaterCredentialSource::None);
+
+    match source {
+        AccountUpdaterCredentialSource::None => Ok(None),
+        AccountUpdaterCredentialSource::Application => resolve_application_config(state).map(Some),
```
ShankarSinghC, line 44: Why do we have AccountUpdaterCredentialSource as enum ? What does None mean here? Are there going to be more variants?

crates/router/src/core/account_updater/config.rs
```diff
@@ -0,0 +1,75 @@
...
+        settings::AccountUpdaterConfig::Juspay(juspay) => {
+            Ok(ResolvedAccountUpdaterConfig::Juspay(JuspayConfig {
+                base_url: juspay.base_url.clone(),
+                api_key: juspay.api_key.clone(),
+                merchant_id: juspay.merchant_id.clone(),
+                euler_encryption_public_key: juspay.euler_encryption_public_key.clone(),
+                au_decryption_pvt_key: juspay.au_decryption_pvt_key.clone(),
+                card_sync_key_id: juspay.card_sync_key_id.clone(),
+                supported_card_networks: juspay.supported_card_networks.clone(),
+                refresh_timeout: Duration::from_secs(juspay.refresh_timeout_in_secs),
```
ShankarSinghC: Can we have a from impl for this ?

crates/router/src/core/account_updater/eligibility.rs
```diff
@@ -0,0 +1,121 @@
...
+        .as_ref()
+        .filter(|card_network| config.supported_card_networks().contains(*card_network))
+        .ok_or_else(|| report!(AccountUpdaterError::UnsupportedNetwork))
+        .attach_printable("Stored card network is not configured for Account Updater")?;
+
+    let raw_payment_method_data = RawPaymentMethodFetchAccess::Allowed
+        .get_raw_payment_method_data(state, platform, profile, payment_method, storage_type)
+        .await
+        .change_context(AccountUpdaterError::CardUnusable)
+        .attach_printable("Account Updater could not unvault the stored card")?;
```
ShankarSinghC: The `get_raw_payment_method_data` function internally calls `retrieve_payment_method_data_from_storage`, which in turn calls `retrieve_and_delete_cvc_from_payment_token`. Since the latter deletes the CVC after retrieving it, the CVC may be null when constructing the force-sync retrieve response.

crates/router/src/core/account_updater/eligibility.rs
```diff
@@ -0,0 +1,121 @@
...
+        .attach_printable("Stored payment method data holds no card details")?;
+
+    card_details
+        .card_network
+        .as_ref()
+        .filter(|card_network| config.supported_card_networks().contains(*card_network))
+        .ok_or_else(|| report!(AccountUpdaterError::UnsupportedNetwork))
+        .attach_printable("Stored card network is not configured for Account Updater")?;
+
+    let raw_payment_method_data = RawPaymentMethodFetchAccess::Allowed
```
ShankarSinghC: Do we want to allow the account updater flow for all types of auth, or should it be limited to internal API key auth? That way, we only perform this check when the MIT payments server makes a retrieve call.

crates/router/src/core/account_updater.rs
```diff
@@ -1,3 +1,81 @@
...
+        &config,
+    )
+    .await;
+
+    match outcome {
+        Ok(refresh_outcome) => logger::info!(
+            account_updater_outcome = refresh_outcome.as_str_name(),
+            "Account Updater refresh completed"
+        ),
+        Err(error) => logger::info!(?error, "Account Updater refresh did not complete"),
```
ShankarSinghC: Should this be a info log or warning log ?

crates/router/src/core/account_updater/eligibility.rs
```diff
@@ -0,0 +1,121 @@
...
+        .get_raw_payment_method_data(state, platform, profile, payment_method, storage_type)
```
ShankarSinghC: If `force_sync` is true, we make the Locker call twice: once as part of the Account Updater flow, and once afterward while constructing the retrieve response.

crates/router/src/core/payment_methods.rs
```diff
@@ -5766,11 +5767,29 @@ pub async fn retrieve_payment_method(
         },
     )?;
 
+    if request.force_sync {
+        let processor = platform.get_processor();
+        let account_updater_dimensions = dimension_state::Dimensions::new()
+            .with_processor_merchant_id(processor.get_processor_merchant_id())
+            .with_organization_id(processor.get_account().get_org_id().clone())
+            .with_profile_id(profile.get_id().clone());
```
ShankarSinghC: Should this be at provider or processor ?
cc: @apoorvdixit88

crates/router/src/core/unified_connector_service/connector_config.rs
```diff
@@ -1689,6 +1707,29 @@ impl ForeignTryFrom<(Connector, &ConnectorAuthType, Option<&serde_json::Value>)>
...
+                        .ok_or_else(|| err("Juspay requires metadata"))?;
+
+                    Ok(Self::Juspay {
+                        api_key: api_key.clone(),
+                        merchant_id: juspay_meta.merchant_id,
+                        base_url: juspay_meta.base_url,
+                        juspay_encryption_public_key: juspay_meta.juspay_encryption_public_key,
+                        response_decryption_private_key: juspay_meta
+                            .response_decryption_private_key,
+                        card_sync_key_id: juspay_meta.card_sync_key_id,
```
ShankarSinghC: We can have a from impl for this

crates/router/src/core/account_updater/config.rs
kanikabansal08: It's an enum because it answers where credentials come from, not whether the feature is on.
There can be more variants in future. Future org/merchant/profile sources will extend the enum alongside their credential storage. Mentioned in `superposition_seed.toml`

crates/router/src/core/payment_methods.rs
apoorvdixit88: It should be at provider.

crates/router/src/core/payment_methods.rs
```diff
@@ -5792,16 +5793,22 @@ pub async fn retrieve_payment_method(
...
-            storage_type,
-        ))
-        .await
-        .attach_printable("Failed to get raw payment method data")?;
+    let account_updater_access = get_account_updater_access(api_key_type, request.force_sync);
+
+    let unvault_access = match account_updater_access {
+        true => RawPaymentMethodFetchAccess::Allowed,
+        false => raw_payment_method_fetch_access,
+    };
```
ThisIsMani: Should we move this to `get_raw_payment_method_data_fetch_access` and pass `account_updater_access` as parameter?

crates/router/src/core/payment_methods.rs
```diff
@@ -5830,6 +5837,66 @@ pub async fn retrieve_payment_method(
...
+                RawPaymentMethodFetchAccess::Denied.get_raw_payment_method_data(
+                    &state,
+                    &platform,
+                    &profile,
+                    &payment_method,
+                    storage_type,
+                ),
+            )
+            .await
+            .attach_printable("Failed to get raw payment method data")?;
```
ThisIsMani: Are we supposed to call this again? Don't we have this data already above?

crates/router/src/core/payment_methods.rs
```diff
@@ -5792,16 +5793,22 @@ pub async fn retrieve_payment_method(
...
-            &platform,
-            &profile,
-            &payment_method,
-            storage_type,
-        ))
-        .await
-        .attach_printable("Failed to get raw payment method data")?;
+    let account_updater_access = get_account_updater_access(api_key_type, request.force_sync);
+
+    let unvault_access = match account_updater_access {
```
ShankarSinghC: can we call this as retrieve raw card or something ?

crates/router/src/core/payment_methods.rs
ShankarSinghC: +1

crates/router/src/core/payment_methods.rs
ShankarSinghC: Yes, we can use some sort of simplified access method instead of adding a separate access function for each feature in the core logic.

ShankarSinghC (dismissed):

ShankarSinghC (approved):

JeevaRamu0104 (approved):

ThisIsMani (approved):

prajjwalkumar17 (approved):

jarnura (approved):

PR #13514, opened 2026-08-03
feat(account_updater): emit synchronization observation event

kanikabansal08:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

Milestone 2 runs Account Updater in shadow mode: no locker write, no payment method update, no change to the API response. Nothing about a run is observable from the outside. This PR adds the one artifact that makes it observable — a Kafka event emitted once per evaluation, on **every** exit path, including the ones that never reach UCS.

### What the event carries

`KafkaAccountUpdaterEvent` (`crates/router/src/events/account_updater.rs`):

| field | notes |
| --- | --- |
| `request_id` | correlates with `api_logs` / `connector_events` |
| `merchant_id`, `profile_id` | the processor account and profile the run was resolved for |
| `payment_method_id` | also the Kafka key, so one payment method's events stay ordered within a partition |
| `card_network` | read from the **stored** payment method, not the refresh response |
| `updater_outcome` | set only when the UCS call returned an outcome |
| `error_category` | set only when the evaluation failed |
| `latency_ms` | measured around the whole evaluation, so it is the wall time the caller paid |
| `created_at` | unix nanos |

`updater_outcome` and `error_category` are the two halves of `CustomResult<CardRefreshOutcome, AccountUpdaterError>`, so exactly one of them is ever populated:

- **`updater_outcome`** — `CARD_REFRESH_ACCOUNT_UPDATED`, `CARD_REFRESH_EXPIRY_UPDATED`, `CARD_REFRESH_NO_CHANGE`, `CARD_REFRESH_CLOSED`, `CARD_REFRESH_NOT_FOUND`, `CARD_REFRESH_CONTACT_ISSUER`, `CARD_REFRESH_OUTCOME_UNSPECIFIED`
- **`error_category`** — `missing_application_config`, `payment_method_not_a_card`, `payment_method_not_active`, `unsupported_network`, `card_unusable`, `refresh_call_failed`, `refresh_returned_error`

Keeping them in separate columns is the point. Collapsed into one, `payment_method_not_active` would sit next to `no_change` in the same enum and the stale-card rate would silently include runs where we never asked the issuer anything.

### Additional Changes

`account_updater_topic` added to `config/config.example.toml`, `config/development.toml`, `config/docker_compose.toml`, `config/deployments/env_specific.toml`.

The event is not behind a dedicated emit flag. It is one fire-and-forget `log_event` call on the retrieve path, already gated by the Superposition Account Updater switch and by internal auth with `force_sync`, so it follows `[events] source` like the other event types. The topic emptiness check lives in `KafkaSettings::validate` alongside the other topics, so it runs unconditionally.

## Motivation and Context

Part of milestone 2, subissue 3. Follows #13510, which is merged.

## How did you test it?

Screenshot 2026-08-17 at 2 59 03 PM

## Checklist


XyneSpaces:
[should-fix] The new account-updater event import is line-specific and should be gated in `crates/router/src/core/payment_methods.rs`: `events::account_updater` is only declared under `#[cfg(feature = "v2")]`, but `payment_methods.rs` is compiled for v1 too, so a v1 build fails with an unresolved import. Please gate the import and the event-construction block with the same `v2` cfg, or make the module available for v1 as well.

XyneSpaces:
[blocking] Guard the new Account Updater event import and emission block behind `#[cfg(feature = "v2")]`. `events::account_updater` is v2-gated, so the default v1 build will fail when `retrieve_payment_method` references it unconditionally.

XyneSpaces:
[blocking] Gate the new account-updater event import and emission with the same `v2` cfg as the module, or make the module available for both feature sets. `crates/router/src/events.rs` adds `#[cfg(feature = "v2")] pub mod account_updater;`, but `crates/router/src/core/payment_methods.rs` now imports and uses `events::account_updater` unconditionally, so a v1-only build cannot resolve the module.

VenuMadhav2541 (approved):

jarnura (approved):

PR #13515, opened 2026-08-03
feat(external-3ds-vgs): Netcetera external 3DS authentication over VGS external vault (v1)

ShankarSinghC:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

Implements Netcetera external 3DS authentication for payments where the card lives in an external vault (VGS), with connector calls routed through UCS (connector-service). This reimplements an earlier POC, but reuses the existing non-proxy authentication functions (`create_new_authentication`, `utils::update_trackers`) and the standard tracker-persist ordering wherever the connector-native and UCS transports allow, instead of duplicating auth-tracking logic per proxy step or bypassing the normal update flow.

Key pieces:
- `perform_pre_authentication_proxy` / `perform_authentication_proxy` / `perform_post_authentication_proxy` in `core/authentication.rs` — proxy analogues of the non-proxy pre-authentication/authenticate/post-authentication functions, driving Netcetera over UCS with a vaulted `ExternalVaultPaymentMethodData` alias instead of a real card, but routing their results through the same `utils::update_trackers` the non-proxy flow uses (including its `Err(error) => ErrorUpdate` handling, so a single code path persists both success and connector-reported failure).
- `resolve_proxy_ucs_gateway_and_vault_mca` — extracted shared helper for the UCS gateway-decision + external-vault MCA resolution that was previously duplicated between the pre- and post-authentication functions.
- `PaymentExternalVaultProxyConfirm` (`payment_confirm_external_vault_proxy.rs`) — the proxy confirm operation, wiring pre-/post-authentication into the standard `should_continue_confirm_transaction` gate so trackers persist before the connector-call decision, rather than an early-return that could skip the normal `ConfirmUpdate`.
- UCS request/response plumbing for `PreAuthenticate`/`Authenticate`/`PostAuthenticate` over the external-vault-proxy card representation (`unified_connector_service.rs`, `unified_connector_service/transformers.rs`, `connector_config.rs`).

### Fixes found and applied during end-to-end testing

- **Gateway-stickiness poisoning**: an interleaved `PSync` (no dedicated UCS rollout config, defaults to Shadow) could overwrite the per-payment `gateway_system` in `payment_intent.feature_metadata`, permanently downgrading a subsequent Authorize/ExternalVaultProxy call to the legacy Direct connector integration. Fixed by excluding `PSync` from persisting its gateway decision.
- **`customer_acceptance` never persisted on the proxy confirm attempt** — `get_trackers` only read it from the current request, so a 3DS resume (which doesn't repeat `customer_acceptance`) always saw `None`, silently skipping PM vaulting. Now falls back to and persists onto `payment_attempt.customer_acceptance`, matching the non-proxy `payment_confirm.rs`.
- **Duplicate PM creation on 3DS resume** — the resume call had nothing to fetch (no `VaultCardTokenData` on the request), so `create_payment_method` fell through to creating a second PM instead of reusing the one from the original confirm. Now also checks `payment_attempt.payment_method_id` and re-fetches by id.
- **`RawPaymentMethodFetchAccess::Allowed` didn't special-case proxy-vaulted cards** — it tried a live vault retrieve (`connector_vault_id` from `locker_id`, which proxy cards never have), always failing. Now short-circuits to the stored `external_vault_token_data`, same as the `Denied` branch already did.
- **`is_terminal_status()` gating missing in `perform_post_authentication_proxy`** — restructured to mirror the non-proxy `perform_post_authentication`'s `if !is_terminal_status() && is_pull_mechanism_enabled { live call } else { pass through }` shape. Previously, a resume that merely raced ahead of its own RRes webhook (not yet terminal, pull disabled) was permanently marked `Failed` instead of being left as-is for a later correctly-timed resume.
- `ExternalVaultCard.card_cvc` made mandatory (was `Option`) after confirming CVC is always available on every reachable construction path in this flow; the one branch that could never recover a real CVC (re-resolving from a saved-PM token) now fails loudly instead of silently building an invalid card.
- Removed unused `payment_method`/`payment_method_type` fields from `PaymentAttemptUpdate::AuthenticationUpdate` — these were a workaround for a POC-era function that skipped the normal `ConfirmUpdate` on a challenge halt; that function no longer exists in the reimplemented flow.

### Known limitations (tracked in `config/external-vault-proxy-pending-items.md`, not committed)

- Only Checkout validated end-to-end through VGS; Stripe/Adyen have known connector-side gaps.
- No duplicate-card detection on the proxy save-card path (VGS aliases aren't stable across re-entries of the same physical card, so the existing PAN-based fingerprint mechanism doesn't apply).
- Nickname given at save time doesn't appear in the saved-cards list for proxy cards (`RawProxyCardDataResponse` has no `nick_name` field).

## Motivation and Context

External-vault (VGS) merchants need 3DS authentication support equivalent to the direct-card flow. Netcetera is the 3DS connector; UCS (connector-service) is the transport for both the payment connector and the auth connector in this flow, since the real card never reaches hyperswitch.

## How did you test it?

Manually, end-to-end, locally against a VGS test vault + Netcetera sandbox + Checkout sandbox: fresh card entry with challenge flow (pre-auth → SDK AReq → challenge → RRes webhook → resume authorize with CAVV), save-card + off-session repeat, and the `force_3ds_challenge`/`request_external_three_ds_authentication` bypass path. Confirmed via router logs and direct DB inspection (`payment_attempt`, `authentication`, `payment_methods` tables) at each fix.

## Checklist


XyneSpaces:
**Review note (line-independent):** This PR touches multiple payment-critical paths (3DS, UCS, webhooks, payment methods). Please ensure the gateway-stickiness and CAVV vaulting changes have targeted runtime/test coverage before merge; inline findings could not be posted from this scheduled run.

crates/router/src/core/payments.rs
```diff
@@ -3082,190 +3082,258 @@ where
...
-            // this flow, so set it explicitly — the confirm update tracker reads it back.
-            payment_data
-                .set_connector_in_payment_attempt(Some(connector.connector_name.to_string()));
+    tokenize_in_router_when_external_authentication_proxy(
+        state,
+        &mut payment_data,
+        &business_profile,
+        platform.get_processor().get_key_store(),
+    )
+    .await?;
```
sahkal, line 3092: Can we also include exemption manager changes aswell

crates/router/src/core/payments.rs
```diff
@@ -3082,190 +3082,258 @@ where
...
+                    &merchant_connector_account,
+                    &connector,
+                    &mut payment_data,
+                    op_ref,
+                    Some(header_payload.clone()),
+                )
+                .await?;
+            }
+
+            utils::trigger_payments_webhook(
```
sahkal, line 3295: in both the cases we are calling `trigger_payments_webhook` we can keep it outside of condisional statements then

crates/router/src/core/payments.rs
```diff
@@ -5225,7 +5314,78 @@ impl PaymentRedirectFlow for PaymentAuthenticateCompleteAuthorize {
...
+                .external_vault_details
+                .is_external_vault_enabled()
+                && match payment_attempt.payment_token.as_ref() {
+                    Some(token) => read_external_vault_alias_from_temp_locker(
+                        state,
+                        token,
+                        platform.get_processor().get_key_store(),
+                    )
+                    .await
+                    .is_ok(),
```
sahkal, line 5343: can we also `inspect_err` and log before moving ahead

crates/router/src/core/payments.rs
```diff
@@ -6169,8 +6329,17 @@ where
...
         execution_mode,
     };
-    // Update feature metadata to track Direct routing usage for stickiness
-    update_gateway_system_in_feature_metadata(payment_data, gateway_context.get_gateway_system())?;
+    // Update feature metadata to track routing usage for stickiness. Read-only status-check
+    // flows (e.g. PSync) almost never have their own rollout config and default to Shadow
+    // (which resolves to GatewaySystem::Direct); persisting that ambient fallback here would
+    // clobber the sticky decision a concurrently in-flight Authorize/ExternalVaultProxy call
+    // relies on, silently downgrading it to the legacy Direct connector integration.
+    if core_utils::get_flow_name::<F>().unwrap_or_default() != "PSync" {
```
sahkal, line 6331: why do we need this?

crates/router/src/core/payments.rs
```diff
@@ -12681,6 +12850,142 @@ pub async fn route_connector_v1_for_payouts(
...
+    let is_external_authentication_requested = payment_data
+        .get_payment_intent()
+        .request_external_three_ds_authentication
+        .unwrap_or(false);
+    if !is_external_authentication_requested || payment_data.get_token().is_some() {
+        return Ok(());
+    }
+    let Some(external_vault_pmd) = payment_data.get_external_vault_pmd().cloned() else {
+        return Ok(());
+    };
```
sahkal: please dont use early returns

crates/router/src/core/payments.rs
```diff
@@ -12681,6 +12850,142 @@ pub async fn route_connector_v1_for_payouts(
...
+
+    let alias_token = tokenize_external_vault_alias_proxy(
+        state,
+        &external_vault_pmd,
+        payment_method,
+        business_profile,
+        key_store,
+    )
+    .await
+    .attach_printable("Failed to mint external vault alias token ahead of 3DS eligibility check")?;
```
sahkal: What is `mint`?

crates/router/src/core/authentication.rs
```diff
@@ -369,3 +351,1383 @@ pub async fn perform_pre_authentication(
...
+        authenticate_tracker_router_data,
+        authentication,
+        None,
+        key_store,
+        authentication_info,
+        storage_scheme,
+    ))
+    .await?;
+
+    let attempt_update = crate::types::storage::PaymentAttemptUpdate::AuthenticationUpdate {
```
sahkal: can we use absolute imports

crates/router/src/core/authentication.rs
```diff
@@ -369,3 +351,1383 @@ pub async fn perform_pre_authentication(
...
+
+    // Self-processing `Platform` reconstructed from `processor` — see doc comment on
+    // `perform_pre_authentication_proxy` below.
+    let platform = domain::Platform::new(
+        processor.get_account().clone(),
+        processor.get_key_store().clone(),
+        processor.get_account().clone(),
+        processor.get_key_store().clone(),
+        initiator.cloned(),
+    );
```
sahkal: Why are we creating new platform??

crates/router/src/core/authentication.rs
```diff
@@ -369,3 +351,1383 @@ pub async fn perform_pre_authentication(
...
+        )
+        .await?;
+    let execution_mode = match execution_path {
+        common_enums::ExecutionPath::UnifiedConnectorService => {
+            common_enums::ExecutionMode::Primary
+        }
+        common_enums::ExecutionPath::ShadowUnifiedConnectorService => {
+            common_enums::ExecutionMode::Shadow
+        }
+        common_enums::ExecutionPath::Direct => return Ok(None),
```
sahkal: Can we avoid early return

crates/router/src/core/authentication.rs
```diff
@@ -369,3 +351,1383 @@ pub async fn perform_pre_authentication(
...
+    };
+
+    let ucs_authentication_data = match pre_authenticate_router_data.response {
+        Ok(core_types::PaymentsResponseData::TransactionResponse {
+            authentication_data,
+            ..
+        }) => authentication_data.map(|boxed| *boxed),
+        Ok(_) => None,
+        Err(err) => return Ok(Err(err)),
+    };
```
sahkal: Can we evaluate this error at the end of definition

sahkal: Also we can merge the statements into single variable

crates/router/src/core/authentication.rs
```diff
@@ -369,3 +351,1383 @@ pub async fn perform_pre_authentication(
...
+
+    let default_message_version = common_utils::types::SemanticVersion::new(2, 2, 0);
+    let message_version = ucs_authentication_data
+        .as_ref()
+        .and_then(|data| data.message_version.clone())
+        .unwrap_or(default_message_version);
```
sahkal: We need not  create this extra memory, incase of not required `default_message_version`

crates/router/src/core/authentication.rs
```diff
@@ -369,3 +351,1383 @@ pub async fn perform_pre_authentication(
...
+    let threeds_server_transaction_id = ucs_authentication_data
+        .as_ref()
+        .and_then(|data| data.threeds_server_transaction_id.clone())
+        .unwrap_or_default();
```
sahkal: we can error out if we get `TransactionResponse` in this flow which has `threeds_server_transaction_id` missing

crates/router/src/core/authentication.rs
```diff
@@ -369,3 +351,1383 @@ pub async fn perform_pre_authentication(
...
+    let connector_authentication_id = ucs_authentication_data
```
sahkal: we can map directly to `threeds_server_transaction_id`

crates/router/src/core/authentication.rs
```diff
@@ -369,3 +351,1383 @@ pub async fn perform_pre_authentication(
...
+        .as_ref()
+        .and_then(|data| data.ds_trans_id.clone());
+
+    Ok(Ok(
+        core_types::authentication::AuthenticationResponseData::PreAuthNResponse {
+            threeds_server_transaction_id,
+            maximum_supported_3ds_version: message_version.clone(),
+            connector_authentication_id,
+            three_ds_method_data: None,
+            three_ds_method_url: None,
```
sahkal: Why are these empty, we need to do DDC before authenticating the transaction.

crates/router/src/core/authentication.rs
```diff
@@ -369,3 +351,1383 @@ pub async fn perform_pre_authentication(
...
+        Option<String>,
+    ),
+    ApiErrorResponse,
+> {
+    let external_vault_pmd = payment_data
+        .external_vault_pmd
+        .clone()
+        .ok_or(ApiErrorResponse::InternalServerError)
+        .attach_printable(
+            "external_vault_pmd missing in pre-authentication (guaranteed present by eligibility)",
```
sahkal: `(guaranteed present by eligibility)` better error message please

crates/router/src/core/authentication.rs
```diff
@@ -369,3 +351,1383 @@ pub async fn perform_pre_authentication(
...
+        business_profile,
+        payment_data.payment_attempt.payment_id.clone(),
+        merchant_connector_id,
+        payment_data.payment_attempt.organization_id.clone(),
+        payment_data.payment_intent.force_3ds_challenge,
+        payment_data.payment_intent.psd2_sca_exemption_type,
+        processor,
+        initiator,
+        &domain::Card::default(),
+        None,
```
sahkal: can we add `browser_info` as well

crates/router/src/core/authentication.rs
```diff
@@ -369,3 +351,1383 @@ pub async fn perform_pre_authentication(
...
+        }));
+    }
+
+    Ok(Ok(
+        core_types::authentication::AuthenticationResponseData::PostAuthNResponse {
+            trans_status,
+            authentication_value: cavv.map(hyperswitch_masking::Secret::new),
+            eci,
+            challenge_cancel: None,
+            challenge_code_reason: None,
```
sahkal: can we also map these values, incase of failures

crates/router/src/core/authentication.rs
```diff
@@ -369,3 +351,1383 @@ pub async fn perform_pre_authentication(
...
+        .unwrap_or(common_enums::TransactionStatus::VerificationNotPerformed);
+    let eci = post_auth_authentication_data
+        .as_ref()
+        .and_then(|data| data.eci.clone())
+        .or_else(|| authentication.eci.clone());
+    let cavv = post_auth_authentication_data
+        .as_ref()
+        .and_then(|data| data.cavv.as_ref().map(|cavv| cavv.peek().clone()))
+        .or_else(|| authentication.cavv.clone());
+
```
sahkal, line 1008: you can directly map cavv to `PostAuthNResponse` struct `authentication_value`

crates/router/src/core/authentication.rs
```diff
@@ -369,3 +351,1383 @@ pub async fn perform_pre_authentication(
...
+        .to_not_found_response(ApiErrorResponse::InternalServerError)
+        .attach_printable("Error while fetching external vault authentication record")?;
+
+    let ucs_authentication_data =
+        hyperswitch_domain_models::router_request_types::UcsAuthenticationData {
+            eci: authentication.eci.clone(),
+            cavv: authentication
+                .cavv
+                .clone()
+                .map(hyperswitch_masking::Secret::new),
```
sahkal: we are not storing cavv authentication table

crates/router/src/core/authentication.rs
```diff
@@ -369,3 +351,1383 @@ pub async fn perform_pre_authentication(
...
+                );
+        }
+    };
+
+    let lineage_ids = external_services::grpc_client::LineageIds::new(
+        business_profile.merchant_id.clone(),
+        business_profile.get_id().clone(),
+    );
+
+    let netcetera_notification_url = Some(common_utils::types::Url::wrap(
```
sahkal: why is connector name mentioned in core flows?

crates/router/src/core/authentication.rs
```diff
@@ -369,3 +351,1383 @@ pub async fn perform_pre_authentication(
...
+        url::Url::parse(&payments_core::helpers::create_authorize_url(
+            &state.base_url,
+            payment_attempt,
+            &psp_connector_name,
+        ))
+        .change_context(ApiErrorResponse::InternalServerError)?,
+    ));
+
+    let netcetera_acquirer_metadata =
```
sahkal: +1

crates/router/src/core/authentication.rs
```diff
@@ -369,3 +351,1383 @@ pub async fn perform_pre_authentication(
...
+        psp_merchant_connector_account
+            .get_metadata()
+            .and_then(|metadata| {
+                let value = metadata.expose();
+                let obj = value.as_object()?;
+                let acquirer: serde_json::Map<String, serde_json::Value> = [
+                    "acquirer_bin",
+                    "acquirer_merchant_id",
+                    "acquirer_country_code",
```
sahkal: take the parsing logic from `get_payment_external_authentication_flow_during_confirm`, also take the fall back logic to get acquirer incase not present. Also ideally we should check `Merchant Configuration ID` in authentication connector mca! if found we need not check for any acquirer details set at our end and directly proceed with the transaction

crates/router/src/core/authentication.rs
```diff
@@ -369,3 +351,1383 @@ pub async fn perform_pre_authentication(
...
+fn parse_ucs_authenticate_response(
+    response: &Result<core_types::PaymentsResponseData, core_types::ErrorResponse>,
+    fallback_threeds_server_transaction_id: Option<String>,
+) -> CustomResult<ParsedAuthenticateResponse, ApiErrorResponse> {
+    #[derive(serde::Deserialize)]
+    struct AppChallengeAcsMetadata {
+        acs_signed_content: Option<String>,
+        acs_reference_number: Option<String>,
+        acs_trans_id: Option<String>,
+    }
```
sahkal: define the struct outside

crates/router/src/core/authentication.rs
```diff
@@ -369,3 +351,1383 @@ pub async fn perform_pre_authentication(
...
+            }) => (
+                Some(endpoint.clone()),
+                form_fields
+                    .get(crate::consts::CREQ_CHALLENGE_REQUEST_KEY)
+                    .cloned(),
+                form_fields.get("acsSignedContent").cloned(),
+                form_fields.get("acsReferenceNumber").cloned(),
+                form_fields.get("acsTransID").cloned(),
+            ),
+            _ => (None, None, None, None, None),
```
sahkal: we need not form RedirectForm for Areq response you can refer `crates/hyperswitch_connectors/src/connectors/netcetera/transformers.rs` for further details

crates/router/src/core/authentication.rs
```diff
@@ -369,3 +351,1383 @@ pub async fn perform_pre_authentication(
...
+    let trans_status = areq_authentication_data
+        .as_ref()
+        .and_then(|data| data.trans_status.clone())
+        .unwrap_or_else(|| {
+            if acs_url.is_some() || acs_signed_content.is_some() {
+                common_enums::TransactionStatus::ChallengeRequired
+            } else {
+                common_enums::TransactionStatus::VerificationNotPerformed
+            }
+        });
```
sahkal: you should directly map `trans_status` need judge based on presence of data

crates/router/src/core/authentication.rs
```diff
@@ -369,3 +351,1383 @@ pub async fn perform_pre_authentication(
...
+    let eci = areq_authentication_data
+        .as_ref()
+        .and_then(|data| data.eci.clone());
+    let ds_trans_id = areq_authentication_data
+        .as_ref()
+        .and_then(|data| data.ds_trans_id.clone());
+    let three_ds_server_transaction_id = areq_authentication_data
+        .as_ref()
+        .and_then(|data| data.threeds_server_transaction_id.clone())
+        .or(fallback_threeds_server_transaction_id);
```
sahkal: we don't need `fallback_threeds_server_transaction_id` we should always get `three_ds_server_transaction_id` in case of 200 response

crates/router/src/core/authentication.rs
```diff
@@ -369,3 +351,1383 @@ pub async fn perform_pre_authentication(
...
+        }
+        common_enums::DecoupledAuthenticationType::Frictionless => {
+            hyperswitch_domain_models::router_request_types::authentication::AuthNFlowType::Frictionless
+        }
+    };
+
+    let authenticate_response_data =
+        core_types::authentication::AuthenticationResponseData::AuthNResponse {
+            authn_flow_type,
+            authentication_value: areq_cavv.map(hyperswitch_masking::Secret::new),
```
sahkal: you can directly map from authenticate response

crates/router/src/core/authentication.rs
```diff
@@ -369,3 +351,1383 @@ pub async fn perform_pre_authentication(
...
+            trans_status: trans_status.clone(),
+            connector_metadata: None,
+            ds_trans_id,
+            eci,
+            challenge_code: None,
+            challenge_cancel: None,
+            challenge_code_reason: None,
+            message_extension: None,
+        };
```
sahkal: please map the values if found

crates/router/src/core/unified_connector_service/connector_config.rs
```diff
@@ -618,6 +618,8 @@ pub enum ConnectorSpecificConfig {
     },
     /// Givepayments connector configuration
     Givepayments { api_key: Secret<String> },
+    /// Netcetera authentication connector configuration (no connector-specific config needed)
+    Netcetera,
```
sahkal, line 651: can we atleast map and send the values, we can take the fix later on ucs for consumption and deprecate the plain text creds sent in request.

crates/router/src/core/unified_connector_service/transformers.rs
```diff
@@ -1238,6 +1238,117 @@ impl
...
+                .transpose()?,
+            browser_info: router_data
+                .request
+                .browser_info
+                .clone()
+                .map(payments_grpc::BrowserInformation::foreign_try_from)
+                .transpose()?,
+            connector_feature_data: None,
+            capture_method: capture_method.map(|capture_method| capture_method.into()),
+            webhook_url: None,
```
sahkal: why is webhook_url `None`

crates/router/src/core/unified_connector_service/transformers.rs
```diff
@@ -2791,6 +3094,12 @@ impl
                 },
                 None => (None, None),
             },
+            // NOTE: the POC surfaced Netcetera acquirer/results-url metadata echoed back via a
+            // `connector_feature_data` field on this response, but that field does not exist on
+            // `PaymentServiceAuthorizeResponse` in the connector-service proto version this repo
+            // is currently pinned to (tag 2026.07.21.0) — it requires a proto bump upstream in
+            // connector-service first. Left as `(None, None)` pending that; not required for the
+            // pre/authenticate/post-authenticate flow itself.
```
sahkal: Remove the comment

crates/hyperswitch_domain_models/src/router_request_types.rs
```diff
@@ -953,6 +956,9 @@ pub struct PaymentsAuthenticateData {
     pub minor_amount: Option<MinorUnit>,
     pub capture_method: Option<storage_enums::CaptureMethod>,
     pub authentication_data: Option<UcsAuthenticationData>,
+    pub sdk_information: Option<api_models::payments::SdkInformation>,
+    pub device_channel: Option<api_models::payments::DeviceChannel>,
+    pub webhook_url: Option<String>,
```
sahkal, line 961: can we make it `url::Url`

crates/hyperswitch_domain_models/src/router_request_types.rs
```diff
@@ -1437,6 +1446,10 @@ pub struct UcsAuthenticationData {
     pub trans_status: Option<common_enums::TransactionStatus>,
     pub transaction_id: Option<String>,
     pub ucaf_collection_indicator: Option<String>,
+    pub challenge_code: Option<String>,
+    pub challenge_cancel: Option<String>,
+    pub challenge_code_reason: Option<String>,
+    pub message_extension: Option<pii::SecretSerdeValue>,
```
sahkal, line 1452: shouldn't this be `SemanticVersion`

crates/router/src/core/authentication.rs
```diff
@@ -369,3 +356,1406 @@ pub async fn perform_pre_authentication(
         },
     )
 }
+
+#[cfg(feature = "v1")]
+struct ProxyUcsGatewayContext {
```
sahkal, line 380: can we move this stuct to authentication type files

crates/router/src/core/authentication.rs
```diff
@@ -369,3 +356,1406 @@ pub async fn perform_pre_authentication(
...
+    };
+
+    let post_auth_authentication_data = match post_authenticate_router_data.response {
+        Ok(core_types::PaymentsResponseData::TransactionResponse {
+            authentication_data,
+            ..
+        }) => authentication_data.map(|boxed| *boxed),
+        Ok(_) => None,
+        Err(err) => return Ok(Err(err)),
+    };
```
sahkal, line 994: Please avoid early return statements

crates/router/src/core/authentication.rs
```diff
@@ -369,3 +356,1406 @@ pub async fn perform_pre_authentication(
...
+    .attach_printable(
+        "UCS is not enabled for this connector; external vault 3DS authenticate over UCS is unavailable",
+    )?;
+
+    let lineage_ids = external_services::grpc_client::LineageIds::new(
+        business_profile.merchant_id.clone(),
+        business_profile.get_id().clone(),
+    );
+
+    let notification_url = Some(common_utils::types::Url::wrap(
```
sahkal, line 1412: why do we need to `wrap` just parsing should be enough right

crates/router/src/core/authentication.rs
```diff
@@ -369,3 +356,1406 @@ pub async fn perform_pre_authentication(
...
+                .or_else(|| {
+                    business_profile.get_default_acquirer_details_from_network(card_network)
+                })?;
+            Some(AcquirerMetadata {
+                acquirer_bin: acquirer_config.acquirer_bin,
+                acquirer_merchant_id: acquirer_config.acquirer_assigned_merchant_id,
+                acquirer_country_code: acquirer_config.acquirer_country_code,
+            })
+        })
+        .and_then(|metadata| serde_json::to_value(metadata).ok());
```
sahkal, line 1452: These checks should be done only when merchant configuration id is null in authentication connector metadata

crates/router/src/core/authentication.rs
```diff
@@ -369,3 +356,1406 @@ pub async fn perform_pre_authentication(
...
+    Ok(api::authentication::AuthenticationResponse {
+        trans_status,
+        acs_url,
+        challenge_request,
+        acs_reference_number,
+        acs_trans_id,
+        three_dsserver_trans_id: three_ds_server_transaction_id,
+        acs_signed_content,
+        challenge_request_key: None,
+        // A connector-level error is now a hard `Err` above, not a value ending up here.
```
sahkal, line 1789: remove unnecessary comments

crates/router/src/core/payments.rs
```diff
@@ -3082,190 +3082,255 @@ where
...
-    // The external vault proxy flow currently supports only a single, predetermined connector.
-    let (connector, merchant_connector_account, router_data) = match connector {
-        Some(ConnectorCallType::PreDetermined(connector_routing_data)) => {
-            let connector = connector_routing_data.connector_data;
-
-            // `choose_connector` does not persist the chosen connector in the payment attempt for
-            // this flow, so set it explicitly — the confirm update tracker reads it back.
-            payment_data
-                .set_connector_in_payment_attempt(Some(connector.connector_name.to_string()));
+    tokenize_in_router_when_external_authentication_proxy(
```
sahkal, line 3085: can we also add exemption manager evaluation in these flows just like we have in payments. For reference check `apply_three_ds_authentication_strategy`

sahkal (dismissed):

crates/hyperswitch_domain_models/src/router_request_types.rs
hrithikesh026, line 1452: Please add documentation for each new fields

crates/hyperswitch_domain_models/src/router_request_types/authentication.rs
```diff
@@ -185,7 +185,7 @@ pub struct AuthenticationStore {
     pub authentication: authentication::Authentication,
 }
 
-#[derive(Clone, Debug)]
+#[derive(Clone, Debug, Default)]
```
hrithikesh026: Lets not have Default trait implemented for this type. Future field addition to the struct will go unnoticed by developers

crates/router/src/core/payment_methods.rs
```diff
@@ -5851,6 +5851,25 @@ impl RawPaymentMethodFetchAccess {
...
+                // card lives at the external vault (e.g. VGS) and is never given a `locker_id`, so
+                // a live `retrieve_payment_method_data_from_storage` call always fails with a
+                // missing `connector_vault_id`. Short-circuit the same way `Denied` does and hand
+                // back the stored alias as a `ProxyCard` instead of attempting a vault fetch.
+                let proxy_card_data = payment_method
+                    .external_vault_token_data
+                    .clone()
+                    .map(|enc| enc.into_inner());
+                if let Some(external_vault_token_data) = proxy_card_data {
+                    return Ok(Some(payment_methods::RawPaymentMethodData::ProxyCard(
```
hrithikesh026, line 6119: Add logs for this decision

crates/router/src/core/authentication.rs
```diff
@@ -369,3 +356,1411 @@ pub async fn perform_pre_authentication(
...
+    Ok(
+        hyperswitch_domain_models::router_request_types::authentication::AuthenticationStore {
+            cavv: tokenized_data.map(|data| hyperswitch_masking::Secret::new(data.value1)),
+            authentication,
+        },
+    )
+}
+
+#[cfg(feature = "v1")]
+#[allow(clippy::too_many_arguments)]
```
hrithikesh026: you don't need this for structs

crates/router/src/core/authentication.rs
```diff
@@ -369,3 +356,1411 @@ pub async fn perform_pre_authentication(
...
+        .as_ref()
+        .and_then(|data| data.cavv.clone());
+
+    let authentication_type = match trans_status {
+        common_enums::TransactionStatus::ChallengeRequired
+        | common_enums::TransactionStatus::ChallengeRequiredDecoupledAuthentication => {
+            common_enums::DecoupledAuthenticationType::Challenge
+        }
+        _ => common_enums::DecoupledAuthenticationType::Frictionless,
+    };
```
hrithikesh026: this can be a from conversion

crates/router/src/core/authentication.rs
```diff
@@ -369,3 +356,1411 @@ pub async fn perform_pre_authentication(
...
+        acs_trans_id,
+        three_ds_server_transaction_id,
+        acs_signed_content,
+    } = parse_ucs_authenticate_response(authenticate_response_data_ref)?;
+
+    let mut authenticate_tracker_router_data: core_types::RouterData<
+        api::Authentication,
+        (),
+        core_types::authentication::AuthenticationResponseData,
+    > = transformers::construct_router_data(
```
hrithikesh026, line 1715: why are we constructing router data here?
Ideally router data must be constructed only to do a connection call

crates/router/src/core/authentication.rs
```diff
@@ -369,3 +356,1411 @@ pub async fn perform_pre_authentication(
...
+            payment_attempt.clone(),
+            attempt_update,
+            storage_scheme,
+            key_store,
+        )
+        .await
+        .to_not_found_response(ApiErrorResponse::PaymentNotFound)
+        .attach_printable(
+            "Error while updating the payment_attempt for external vault authenticate",
+        )?;
```
hrithikesh026, line 1778: all DB update should happen in update_trackers.

crates/router/src/core/authentication.rs
```diff
@@ -369,3 +356,1411 @@ pub async fn perform_pre_authentication(
...
+    }
+}
+
+/// Drives the live UCS authenticate (AReq) call; response interpretation is
+/// `parse_ucs_authenticate_response`'s job.
+#[cfg(feature = "v1")]
+#[allow(clippy::too_many_arguments)]
+async fn call_ucs_authenticate_proxy(
+    state: &SessionState,
+    platform: &domain::Platform,
```
hrithikesh026, line 1206: this should be Processor

crates/router/src/core/authentication.rs
```diff
@@ -369,3 +356,1411 @@ pub async fn perform_pre_authentication(
...
+                .find_by_merchant_connector_account_merchant_id_merchant_connector_id(
+                    platform.get_provider().get_account().get_id(),
+                    &external_vault_mca_id,
+                    platform.get_provider().get_key_store(),
+                )
+                .await
+                .to_not_found_response(ApiErrorResponse::MerchantConnectorAccountNotFound {
+                    id: external_vault_mca_id.get_string_repr().to_string(),
+                })?,
+        ));
```
hrithikesh026, line 1260: merchant account is already available. No need to fetch it from DB

crates/router/src/core/authentication.rs
```diff
@@ -369,3 +356,1411 @@ pub async fn perform_pre_authentication(
...
+        &psp_router_data,
+    )
+    .await?
+    .ok_or_else(|| {
+        error_stack::report!(ApiErrorResponse::PreconditionFailed {
+            message: "External 3DS authentication was requested for this payment, but UCS is \
+                      not enabled for this connector; cannot proceed without performing 3DS"
+                .to_string(),
+        })
+    })?;
```
hrithikesh026, line 683: how can we conclude that `UCS is not enabled for this connector` just because this function returned None.

crates/router/src/core/authentication.rs
ShankarSinghC: authenticate_tracker_router_data gets built via construct_router_data purely to satisfy update_trackers's signature

crates/router/src/core/authentication.rs
ShankarSinghC: The same inline update is present in the `payment_external_authentication`

crates/router/src/core/authentication.rs
ShankarSinghC: The external vault merchant connector is configured at the provider level, so we need platform input here

crates/router/src/core/authentication.rs
ShankarSinghC: We would already have auth connector here but the external vault connector still needs to be fetched

sahkal (dismissed):

crates/router/src/core/authentication.rs
ShankarSinghC: This function returns None only in case of ucs config is not found (basically direct mode)

sahkal (approved):

crates/router/src/core/payments/helpers.rs
```diff
@@ -3112,11 +3112,21 @@ pub async fn resolve_provider_profile(
...
+                    provider.get_key_store(),
+                    provider.get_account().get_id(),
+                )
+                .await
+                .change_context(errors::ApiErrorResponse::InternalServerError)
+                .attach_printable("Failed to list provider merchant business profiles")?;
+            profiles.into_iter().next().ok_or_else(|| {
+                report!(errors::ApiErrorResponse::InternalServerError)
+                    .attach_printable("Provider merchant has no business profile configured")
+            })
```
hrithikesh026, line 3129: there is no need for listing profiles here. Find one should work

ShankarSinghC: There we are trying to list a profile with provider merchant id
We don't have find one for it

Sakilmostak (approved):

swangi-kumari (approved):

hrithikesh026 (approved):
Approving to unblock — didn't do a full in-depth review given the timeline.

PR #13468, opened 2026-07-27
feat(router): add support for payload split payments 

AkshayaFoiger:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [x] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

Add support for split payments via payload 

Hyperswitch request will additionally accept this value for split payments field.  

```
 "split_payments": {
        "payload_split_payment": {
            "ledger": [
                {
                    "receiver_id": "acct_3eox**********",
                    "amount": 10000
                },
                {
                    "receiver_id": "acct_3***********",
                    "amount": 900
                }
            ]
        }
    }
```

Note that the value sent 

### Additional Changes

- [x] This PR modifies the API contract
- [ ] This PR modifies the database schema
- [ ] This PR modifies application configuration/environment variables

## How did you test it?

Create a payload split payment 

```
curl --location 'http://localhost:8080/payments' \
--header 'Content-Type: application/json' \
--header 'Accept: application/json' \
--header 'api-key: dev_ixwpVfjtOacO9qz8nqJBOhictVZoQm9g8Hf0A4eNbiMnY48DIcGN4xAbQhsXAAyP' \
--data-raw '{
    "amount": 10900,
    "currency": "USD",
    

    
    
    
    
    "confirm": true,
    "customer": {
        "id": "2343",
        "name": "John Doe",
        "email": "john@example.com",
        "tax_registration_id": "TAX123456789"
    },
    "merchant_order_reference_id": "ORD-L3-0001",
    "payment_method": "card",
    "payment_method_data": {
        "card": {
            "card_number": "371449635392376",
            "card_exp_month": "12",
            "card_exp_year": "28",
            "card_holder_name": "John Doe",
            "card_cvc": "9997",
            "card_network": "AMEX"
        }
    },
    

    
    "billing": {
        "address": {
            "line1": "123 Main St",
            
            
            "zip": "10001"
            
            
        }
    },
    
    "billing_descriptor": {
        "statement_descriptor": "ABCD"
    }
    ,    "split_payments": {
        "payload_split_payment": {
            "ledger": [
                {
                    "receiver_id": "acct_3eoxafCHioIB3jNMKJev4",
                    "amount": 10000
                },
                {
                    "receiver_id": "acct_3epxuNyAtNd77zShIaaL1",
                    "amount": 900
                }
            ]
        }
    }
    
}'

```

Response

```
{
[...]

PR #13518, opened 2026-08-03
test(cypress): add split payments coverage for payload connector

### LINKED ISSUE #13520: Extend Cypress split_payments coverage to stripeconnect, adyen, and xendit
### Problem

`crates/common_types/src/payments.rs::SplitPaymentsRequest` supports 4 connectors — `StripeSplitPayment`, `AdyenSplitPayment`, `XenditSplitPayment`, `PayloadSplitPayment` — but Cypress split-payments coverage (`cypress-tests/cypress/e2e/spec/Payment/54-SplitPayments.cy.js`, added in #13518) currently only has config data + assertions for `payload`.

`stripeconnect` already has `split_payments` wired into several `Request` configs in `StripeConnect.js`, but no dedicated split-payment assertions or spec coverage. `adyen` and `xendit` have no split-payments config data at all yet.

### Ask

- [ ] Add `split_payments`-specific Request/Response configs for `adyen` (`AdyenSplitPayment`) and `xendit` (`XenditSplitPayment`) connectors, mirroring `Payload.js`'s `*SplitPayment` configs.
- [ ] Add equivalent coverage/assertions for `stripeconnect` (`StripeSplitPayment`) alongside its existing `split_payments` Request data.
- [ ] Extend `CONNECTOR_LISTS.INCLUDE.SPLIT_PAYMENTS` in `Utils.js` to include each connector as its coverage lands, instead of the current `["payload"]`-only scope.

### Context

Follow-up from #13518 (Cypress coverage for `payload_split_payment`), scoped down to `payload`-only for now since the other 3 connectors had no test data ready.

venkatakarthikm-maker:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [x] CI/CD

## Description

Adds Cypress coverage for the `payload_split_payment` feature (#13468), which had no automated test coverage before this PR.

- `commands.js`: `split_payments` injection was hardcoded to only fire for `stripeconnect` in 4 call sites (`confirmCallTest`, `createConfirmPaymentTest`, `citForMandatesCallTest`, `mitForMandatesCallTest`). Generalized to `supportsSplitPayments(globalState)`, driven by `CONNECTOR_LISTS.INCLUDE.SPLIT_PAYMENTS` in `Utils.js` instead of a single hardcoded connector.
- `Utils.js`: added `SPLIT_PAYMENTS: ["payload"]` to `CONNECTOR_LISTS.INCLUDE`. Backend split-payments support (`crates/common_types/src/payments.rs::SplitPaymentsRequest`) actually spans 4 connectors (`stripe`/`stripeconnect`, `adyen`, `xendit`, `payload`), but only `payload` has Cypress config data today — scoped the list down to what's actually tested for now. Extending to the other 3 is tracked in #13520.
- `Payload.js`: added `No3DSAutoCaptureSplitPayment`, `MandateSingleUseNo3DSAutoCaptureSplitPayment`, and `MITAutoCaptureSplitPayment` configs, each asserting the split ledger is echoed back unchanged in the response.
- `54-SplitPayments.cy.js` (new, connector-agnostic — no hardcoded connector name, gated via the same `SPLIT_PAYMENTS` inclusion list):
  - Create+confirm payment with a split ledger → verify `succeeded` and ledger echoed back.
  - Create+confirm payment without `split_payments` → regression check that the field is optional.
  - CIT (mandate) + split ledger → verify mandate created and ledger applied.
  - MIT (recurring) + split ledger → verify ledger applied on the off-session charge.

### Additional Changes

## Motivation and Context

#13468 shipped `payload_split_payment` with only manual curl verification and no automated regression coverage; the request/response contract (nested `split_payments.payload_split_payment.ledger`) is easy to silently break in future refactors without a test asserting it round-trips correctly.

While manually testing this coverage against the sandbox `payload` connector, found that an under-sum ledger (sum < payment amount) passes hyperswitch's own validation but fails opaquely at the Payload connector (`InvalidAttributes` / `UE_9000`). Filed as #13519 — not fixed in this PR, since it's a backend validation change, not test coverage. This PR's Cypress spec only covers the exact-sum happy path, matching current backend behavior.

Related: #13520 tracks extending this same Cypress coverage pattern to `stripeconnect`, `adyen`, and `xendit`, which also support split_payments on the backend but have no test data yet.

## How did you test it?

Manually verified against `integ.hyperswitch.io` (sandbox `payload` connector) before writing the Cypress assertions:
- Exact-sum ledger (`10000 + 900 = 10900` = amount) → `succeeded`, ledger echoed back exactly matching request.
- Under-sum ledger (`8000 + 1000 = 9000` < amount `10900`) → unexpectedly fails at the connector rather than being caught by hyperswitch validation (see #13519).
- Over-sum ledger case not yet manually re-verified against sandbox; hyperswitch's own validation (`crates/router/src/core/payments/helpers.rs`) already rejects it with `PreconditionFailed` before reaching the connector.

Also ran `eslint`/`prettier` on all touched Cypress files (clean, 0 errors). Cypress spec itself has not yet been executed against a live `payload` connector run (requires `CYPRESS_CONNECTOR=payload` credentials).

## Checklist

image
image
image

Regression for stripe connect

image
image
image


XyneSpaces:
[blocking] Fix the split-payments support check so it returns true for connectors in the split-payments include list. `shouldIncludeConnector` is already negated for inclusion-list filtering, so this currently marks every connector except Payload as supporting split payments and will drive the new split-payments setup down the wrong path.

cypress-tests/cypress/support/commands.js
```diff
@@ -3101,7 +3114,7 @@ Cypress.Commands.add(
       delete confirmBody.setup_future_usage;
     }
 
-    if (reqData?.split_payments && isStripeConnect(globalState)) {
+    if (reqData?.split_payments && supportsSplitPayments(globalState)) {
```
Gnanasundari24, line 3171: We already have stripe connect for splitpayments. Split payments shouldn't be a seperate spec it should include all the testcase with split object. So handle accordingly. Dont remove stripeconnect

Gnanasundari24 (changes requested):

cypress-tests/cypress/e2e/configs/Payment/PayloadConnect.js
```diff
@@ -0,0 +1,898 @@
...
+        payment_method_data: {
+          card: successfulThreeDSTestCardDetails,
+        },
+        currency: "USD",
+        customer_acceptance: null,
+        setup_future_usage: "on_session",
+      },
+      Response: {
+        status: 400,
+        body: {
```
likhinbopanna, line 185: Make it as const and use it wherever is called

cypress-tests/cypress/e2e/configs/Payment/PayloadConnect.js
```diff
@@ -0,0 +1,898 @@
...
+    PaymentIntentWithBillingDescriptor: {
+      Configs: {
+        DELAY: {
+          STATUS: true,
+          TIMEOUT: DUPLICATION_TIMEOUT,
+        },
+      },
+      Request: {
+        currency: "USD",
+        billing_descriptor: {
```
likhinbopanna: Make this as const and use it everywhere required

likhinbopanna (dismissed):

cypress-tests/cypress/e2e/configs/Payment/Utils.js
```diff
@@ -545,6 +561,13 @@ export const CONNECTOR_LISTS = {
       "tsys_transit",
     ],
     SAVE_CARD: ["helcim"],
+    // payloadconnect always splits the full payment amount to receivers via
+    // the ledger, leaving nothing in the merchant's transaction balance for
+    // a standard refund. Payload's connector integration has no split-aware
+    // refund mechanism (crates/hyperswitch_connectors/src/connectors/payload
+    // /transformers.rs builds PayloadRefundRequest from amount +
+    // connector_transaction_id only), so refunds are unsupported here.
+    SPLIT_PAYMENTS_REFUND: ["payloadconnect"],
```
Gnanasundari24: Why this exclusion required for payload connect

Gnanasundari24 (changes requested):

likhinbopanna (approved):

Gnanasundari24 (approved):

PR #13521, opened 2026-08-03
fix(common_utils): add TCP keepalive params to DB URL so pool connections recover from RDS failover

sai-harsha-vardhan:
## Type of Change

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

After an Aurora/RDS failover the router could require a manual restart to recover.

Root cause is in `async-bb8-diesel`: its `ConnectionManager` wraps `r2d2::ConnectionManager` in a single `Arc<Mutex<..>>` shared by the entire pool, and both `connect()` and `is_valid()` acquire it for the full duration of their blocking libpq call. Aurora failover leaves old connections silently unresponsive (no RST/FIN), so that call never returns, the shared lock is never released, and every subsequent connection check *and* new-connection attempt for the whole pool queues behind it indefinitely.

We can bound the failure at the OS layer: libpq passes these connection-string parameters straight to setsockopt(), so the kernel detects and tears down the dead socket on its own. bb8's existing "is_valid() errors -> mark invalid -> discard -> reconnect" logic then works unmodified.

Both `storage_impl::config::Database` and `drainer::settings::Database` implement `DbConnectionParams` without overriding `get_database_url`, so the router pool and the drainer's separate pool are both covered.

`tcp_user_timeout` is deliberately 5s, below the existing 10s bb8 `connection_timeout`. At 10s (equal) the outer timeout wins the race, bb8 never marks the connection invalid, and it returns to the idle queue looking healthy - measured as one hard checkout failure at 10.002s and ~16s to recover, versus zero failures and ~12s at 5s.

### Additional Changes

## How did you test it?

Tested manually
Validated on Linux against a genuine iptables partition of the old instance plus DNS re-point to a healthy replacement, using unmodified bb8 0.8.6 and async-bb8-diesel 0.2.1:
  - in-flight queries at failover: all fail cleanly at ~11s, pool fully recovered, 16/16 follow-up probes succeed
  - all-idle pool at failover: 20/20 probes succeed, no checkout timeouts

## Checklist

### DIFF
--- a/config/config.example.toml
+++ b/config/config.example.toml
@@ -58,6 +58,10 @@ connection_timeout = 10   # Timeout for database connection in seconds
 queue_strategy = "Fifo"   # Add the queue strategy used by the database bb8 client
 idle_timeout = 300        # Maximum idle duration in seconds before a connection is closed
 max_lifetime = 1800       # Maximum lifetime in seconds before a connection is reaped
+keepalives_idle = 5       # Idle seconds before the OS sends TCP keepalive probes (detects silently-dead connections, e.g. after an RDS failover)
+keepalives_interval = 2   # Seconds between keepalive probes when one goes unanswered
+keepalives_count = 3      # Unanswered keepalive probes tolerated before the OS declares the connection dead
+tcp_user_timeout_ms = 5000 # Milliseconds of unacknowledged sent data before the OS closes the connection; must stay well below connection_timeout
 
 # Replica SQL data store credentials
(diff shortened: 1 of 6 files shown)


sai-harsha-vardhan:
closing this for now, as issue is solved by setting max_lifetime of connections to 120 secs

PR #11704, opened 2026-04-05
feat(superposition): move db configs to superposition

aniketburman014:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

This PR migrates payment-related runtime configurations from direct database lookups to the dimension-aware Superposition configuration layer while retaining legacy database fallback behavior.

### Key changes

- Extends the typed configuration dimensions with:
  - `transaction_type`
  - `payment_method_type`

- Adds Superposition-backed resolution for:
  - `step_up_enabled`
  - `authentication_service_eligible`
  - `pre_routing_disabled_pm_pmt`
  - `blocklist_guard`
  - `skip_saving_wallet_at_connector`
  - `connector_api_version`
  - `payment_update_enabled_for_client_auth`

- Migrates the corresponding payment, retry, routing, card, and refund flows away from direct database reads.

- Preserves legacy array-based database behavior for:
  - Connectors enabled for step-up authentication
  - Payment method types for which wallet saving at the connector must be skipped

- Adds JSON object and array serialization support for Superposition documents.

- Handles arrays encoded by the OpenFeature provider as indexed `StructValue` objects, including empty arrays.

- Removes obsolete merchant configuration-key helpers after migrating their call sites.

- Updates `config/superposition_seed.toml` with the new configuration defaults and `transaction_type` dimension.

- Adapts the new configuration keys to the folder-based Superposition naming convention introduced on `main`:
  - `payments.*`
  - `routing.*`
  - `vaulting.*`

### Additional Changes

- [ ] This PR modifies the API contract
- [ ] This PR modifies the database schema
- [ ] This PR modifies application configuration/environment variables

## Checklist

- [ ] I formatted the code `cargo +nightly fmt --all`
- [ ] I addressed lints thrown by `cargo clippy`
- [ ] I reviewed the submitted code
- [ ] I added unit tests for my changes where possible

PR #13529, opened 2026-08-03
feat(superposition): migrate blocklist guard toggle to superposition

ayush22667:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

The blocklist guard is currently a DB-only runtime flag. `POST /blocklist/toggle?status=true` writes a row into the `configs` table keyed `guard_blocklist_for_{merchant_id}`, and two payment-time call sites read that row back with a hand-rolled `find_config_by_key_unwrap_or(..., "false")` + `serde_json::from_str` pair.

This moves the flag onto the standard resolution ladder — **Superposition → `configs` table → hardcoded default** — matching `payments.should_call_gsm`, `payments.should_perform_eligibility` and `payments.enable_extended_card_bin`.

Concretely:

- Adds `payments.guard_blocklist` (boolean, default `false`) to `config/superposition_seed.toml`
- Adds the `GUARD_BLOCKLIST` key constant to `crates/router/src/consts.rs`
- Defines the config in `crates/router/src/core/configs/dimension_config.rs` via the `config!` macro — dimensions `DimensionsWithProcessorAndProviderMerchantId`, targeting key `PaymentId` — plus an `impl DatabaseBackedConfig` whose `db_key()` reconstructs `guard_blocklist_for_{merchant_id}`
- Re-exports `GuardBlocklist` from `crates/router/src/core/configs.rs`
- Replaces both read sites in `crates/router/src/core/payments.rs` with `dimensions.get_guard_blocklist(...)`: `blocklist_guard()` (payment confirm) and `BlockListCheck::should_run()` (payments eligibility API)

Two supporting changes worth calling out for review:

- `blocklist_guard()` is now gated with `#[cfg(feature = "v1")]`. Its only caller (`call_connector_service_prerequisites`) is already v1-gated, and `PaymentIntent::get_id()` returns `GlobalPaymentId` under v2, which would not match the `PaymentId` targeting key.
- `EligibilityCheck::should_run` gains a `payment_elgibility_data` parameter so the eligibility read site can pass a real targeting key instead of `None`, keeping both read sites bucketing consistently under an experiment. `CardTestingCheck::should_run` ignores it.

**Backwards compatibility.** `POST /blocklist/toggle`, `MerchantId::get_blocklist_guard_key()`, and the `core/blocklist/utils.rs` write path are all unchanged. `db_key()` reproduces the existing key format exactly, and `impl ConfigType for bool` parses via `str::parse::<bool>()` — the same `"true"` / `"false"` strings the toggle already writes. Existing merchant settings keep resolving through the DB rung with no data migration.

### Additional Changes

- `config/superposition_seed.toml` — adds the `payments.guard_blocklist` default

## Motivation and Context

As a DB-only flag, the blocklist guard cannot be defaulted per environment, scoped by org/profile, rolled out gradually, or inspected alongside the other runtime configs. Every comparable payment-time boolean has already been migrated to Superposition; this closes the gap for the last of the blocklist flags while keeping the existing toggle API working untouched.

## Linked Issues

- Issue: https://github.com/juspay/hyperswitch-cloud/issues/22466

## How did you test it?

Built under both feature sets: `just check && just check_v2`, `just clippy && just clippy_v2`.

Manually exercised all three rungs of the resolution ladder, confirming which one answered via the `source = ...` log line emitted by `core/configs.rs`:

1. **Superposition** — set a `payments.guard_blocklist` override for the merchant, sent a confirm with a blocklisted card, and saw the `HE-03` / `PaymentBlockedError` rejection (`source = "superposition"`).
2. **DB fallback** — with no Superposition override present, enabled the flag through `POST /blocklist/toggle?status=true` and confirmed the same card is still rejected (`source = "database"`). This is the case that proves `db_key()` reproduces `guard_blocklist_for_{merchant_id}` byte-for-byte.
3. **Default** — a merchant with neither an override nor a `configs` row skips the guard.

`cypress-tests/cypress/e2e/spec/Payment/35-PaymentsEligibilityAPI.cy.js` seeds `guard_blocklist_for_${merchantId}` through the configs API and passes unmodified, covering the eligibility read site against the legacy row.

## Checklist


hrithikesh026 (dismissed):

ayush22667:
closed in favor of [pr](https://github.com/juspay/hyperswitch/pull/11704)

PR #13540, opened 2026-08-04
fix: raise minimum password length to 12 characters

### LINKED ISSUE #13552: Minimum password length does not meet PCI DSS requirement (8 chars, needs 12)
## Bug Description

`MIN_PASSWORD_LENGTH` in `crates/router/src/consts/user.rs` is currently set to `8`. PCI DSS requires a minimum password length of 12 characters, so the current setting falls short of compliance.

## Expected Behavior

Password-creation flows (signup, change-password, rotate-password, forgot-password reset, internal/tenant user creation) should reject passwords shorter than 12 characters.

## Actual Behavior

Passwords as short as 8 characters are currently accepted.

## Additional Context

- `MIN_PASSWORD_LENGTH` is only consumed by `UserPassword::new`, so login (`signin_token_only_flow`) and old-password verification (`compare_password`) are unaffected — they only do an Argon2 hash comparison.
- The frontend (`hyperswitch-control-center`) also enforces an 8-character minimum client-side and should be updated separately to match, to avoid a client/server validation mismatch on signup/reset forms.

muditbhutani:
## Type of Change

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

Raises `MIN_PASSWORD_LENGTH` from 8 to 12 characters (`crates/router/src/consts/user.rs`) to meet the PCI DSS minimum password length requirement.

`MIN_PASSWORD_LENGTH` is only consumed by `UserPassword::new`, which runs on password-creation paths: signup, change-password, rotate-password, forgot-password reset, and internal/tenant user creation. Login (`signin_token_only_flow`) and old-password verification (`compare_password`) only perform an Argon2 hash comparison against the stored hash — they never re-validate password length or format. Existing users with passwords shorter than 12 characters are therefore unaffected and continue to log in normally; the new 12-character minimum only applies the next time a user sets or changes their password.

This is a backend-only change. The frontend (`hyperswitch-control-center`) enforces the same 8-character minimum client-side and should be updated separately to match, to avoid a client/server validation mismatch on signup/reset forms.

### Additional Changes

## Motivation and Context

PCI DSS requires a 12-character minimum password length; the current 8-character minimum falls short of compliance.

Closes #13552

## How did you test it?

Manually tested against a local router build (pre- and post-change) with a local Postgres instance:

**Baseline (pre-change, `MIN_PASSWORD_LENGTH = 8`)**
- Created two users via `POST /user/signup`: one with an 8-char password (`Pass@123`), one with a 9-char password (`Pass@1234`). Both succeeded (`200`).

**After raising the limit to 12 (`MIN_PASSWORD_LENGTH = 12`)**
- [x] Verified existing users with shorter passwords can still log in — `POST /user/signin` with the 8-char and 9-char passwords created above both succeeded (`200`), confirming login only compares against the stored Argon2 hash and doesn't re-validate length.
- [x] Verified signup/change-password/reset-password reject passwords shorter than 12 characters:
  - `POST /user/signup` with a 10-char password → rejected, `400 { "code": "UR_09", "message": "Invalid Password" }`.
  - Boundary check: an 11-char password → rejected (`400 UR_09`); a 12-char password → accepted (`200`).
  - `POST /user/change_password` for the existing 8-char-password user: changing to a 9-char password → rejected (`400 UR_09`); changing to a 12-char password → succeeded (`200`).
  - After the change, the old 8-char password no longer authenticates (`401 UR_01 "Incorrect email or password"`), and the new 12-char password logs in successfully — confirming the update took effect and didn't affect unrelated accounts.
- Confirmed `new_password_without_validation` (the only bypass of `UserPassword::new`'s validation) is unused dead code — `grep` shows no call sites — so there is no path that skips the new 12-character minimum.

## Checklist


crates/router/src/consts/user.rs
```diff
@@ -23,7 +23,7 @@ pub const RECOVERY_CODE_MAX_ATTEMPTS: u8 = 4;
 pub const ORG_LIST_LIMIT_FOR_TENANT: u32 = 20;
 
 pub const MAX_PASSWORD_LENGTH: usize = 70;
-pub const MIN_PASSWORD_LENGTH: usize = 8;
+pub const MIN_PASSWORD_LENGTH: usize = 12;
```
apoorvdixit88, line 26: What will be the behaviour of existing passwords? Will we get an error? Or when we rotate the password, then only user have to set new password of 12 length and old passwords will work as it is? 
Is this a backward compatible change, have we tested? Can we please add the tests.

apoorvdixit88 (approved):

VenuMadhav2541 (approved):

PR #13564, opened 2026-08-05
feat(routing): add automatic kill switch for decision engine routing diffs

### LINKED ISSUE #13565: [FEATURE] Automatic kill switch to cut a profile back to Hyperswitch routing on repeated Decision Engine diffs
## Feature Description

Hyperswitch shadow-compares its static routing evaluation against the Decision Engine (DE) and logs divergence under the `decision_engine_euclid` tag (`is_equal=false`). Today this divergence is only caught when someone reacts to a Grafana alarm on that log, and reverting an affected profile back to Hyperswitch routing is a manual operation.

Requested behavior:
1. The diff log should carry an `is_volume` key so that expected-random diffs from volume-split selections (where both engines pick randomly) can be distinguished from real divergence.
2. A per-profile Redis counter (e.g. `routing_decision_engine_{profile_id}_diff_count`) should count real (non-volume) routing diffs, with a configurable threshold. Once the threshold is met within a window, the profile should be automatically cut over back to Hyperswitch routing (kill switch), without waiting for manual intervention. Volume-based routing diffs must not feed this counter.

## Possible Implementation

- Add `is_volume` to `compare_and_log_result` and have it return the comparison outcome.
- Count non-volume `evaluate_routing` diffs per profile in Redis with a TTL window; on crossing a configurable threshold, set a latched kill-switch key honored at routing-time (forcing `HyperswitchRouting` over the configured `routing_result_source`), with an error-level log and metric for alerting. Manual reset by deleting the Redis key.

## Have you spent some time checking if this feature request has been raised before?

## Have you read the Contributing Guidelines?

- [x] I have read the Contributing Guidelines

## Are you willing to submit a PR?

Yes, PR incoming.

prajjwalkumar17:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

Today, divergence between Hyperswitch's static routing evaluation and the Decision Engine (DE) is only observable through the `decision_engine_euclid` diff log (`is_equal=false`) and is caught only when someone reacts to a Grafana alarm. This PR enriches the diff log and adds an automatic, per-profile kill switch that cuts a diverging profile back to Hyperswitch routing:

1. **`is_volume` in the diff log** — `compare_and_log_result` now logs `is_volume` alongside `is_equal` / `is_equal_length`, so expected-random diffs (volume-split selections, where both sides pick randomly) are distinguishable from real divergence. Volume detection covers both `VolumeSplit` algorithms and **advanced rules whose evaluated output is a `volume_split`** (the DSL executor now surfaces this).

2. **Automatic kill switch (counter overlay, no config write)** — every non-volume `evaluate_routing` diff on a cut-over profile increments a per-profile Redis counter `routing_decision_engine_{profile_id}_diff_count` (HINCRBY). The counter is a **lifetime total** (no TTL); it stays until cleared, so a tripped profile stays cut over until an operator resets it (the reset API is added in the stacked follow-up #13599; until then, clear the Redis key). At the threshold, a one-shot error-level `auto_cutover` log and `DECISION_ENGINE_KILL_SWITCH_TRIGGERED` metric fire for alerting.

   Enforcement is a **read-time overlay**, not a config mutation: `get_routing_result_source` resolves the configured `routing_result_source` as usual, then — if it resolved to `DecisionEngine` and the profile's diff counter has reached the threshold — returns `HyperswitchRouting` instead. The stored `routing_result_source` is **left untouched** (no Superposition write). Because **both payment routing and the dashboard entry (`/routing/entry` → `is_cutover`) resolve through this one function**, a tripped profile serves Hyperswitch on both surfaces at once; clearing the counter reverts both to the Decision Engine. This was confirmed against the control-center, which renders Hyperswitch's own routing cards and reads only `is_cutover` (it never reads the raw `routing_result_source`).

Safety properties:
- **Fail-open**: any Redis error is logged and swallowed — diff accounting and the overlay check never break payment routing (on a Redis blip the overlay reports "not tripped" and the configured source is honored).
- **No false trips**: on the hybrid path only DE **static** outcomes are counted (`RoutingApproach::is_de_static_result()`); dynamic selections (SR exploration/exploitation, debit routing, and any unmapped DE approach strings, which fall back to `Default`) legitimately reorder connectors and are excluded. Volume-split selections are excluded per design.
- **Every divergence class counts, each with a dedicated prefixed log + metric reason**: element/order mismatch → `decision_engine_result_mismatch`; a non-empty DE result shorter than the HS result → `decision_engine_length_mismatch`; an **empty** DE result (outage/error/no rule) → `decision_engine_unresponsive`. Each counted diff emits a WARN with that prefix and tags `DECISION_ENGINE_ROUTING_DIFF` with a `reason` attribute, so Grafana can distinguish rule divergence from DE availability issues.
- **Overlay cost is bounded**: the counter is read in `get_routing_result_source` only when the resolved source is `DecisionEngine` (Hyperswitch-source profiles skip the read entirely).

3. **Shadow mode for non-cutover profiles** — with `[open_router] shadow_routing_enabled = true` (new global flag, requires `static_routing_enabled`), profiles whose `routing_result_source` is still `hyperswitch_routing` get the same diff signal *before* cutover: a detached `tokio::spawn` task calls the DE static evaluation (`routing/evaluate`) and logs the same `evaluate_routing` / `is_equal` / `is_volume` comparison, while the Hyperswitch result serves the payment with **zero added latency**. Shadow mode is **observation-only**: it always runs when enabled and never feeds the kill switch — the kill switch acts solely on profiles configured to use DE results. Validation workflow: enable shadow globally → watch `is_equal` per profile in Grafana → cut over only profiles with clean diffs.

The feature is fully disabled by default (`[open_router.diff_kill_switch] enabled = false`, `shadow_routing_enabled = false`).

> A follow-up PR (#13599, stacked on this one) adds an admin API to reset the counter for self-service re-enablement.

### Additional Changes

New `[open_router.diff_kill_switch]` section (`enabled`, `diff_count_threshold`) and the `[open_router] shadow_routing_enabled` flag:
- [config/development.toml](https://github.com/juspay/hyperswitch/blob/main/config/development.toml)
- [config/config.example.toml](https://github.com/juspay/hyperswitch/blob/main/config/config.example.toml)
- Struct: `crates/router/src/configs/settings.rs` (`DecisionEngineDiffKillSwitch`, nested under `OpenRouter`)

## Motivation and Context

Closes #13565.

When a profile is cut over to DE routing and DE starts diverging from Hyperswitch's evaluation (rule sync issues, DE regressions), the merchant keeps getting mis-routed until a human notices the `is_equal=false` alarm and manually reverts. This PR turns that manual runbook into an automatic, bounded-blast-radius cutover: after N real diffs, the affected profile alone serves Hyperswitch routing (and its dashboard reflects that) while its `routing_result_source` config is left unchanged. The counter is a lifetime total, so the profile stays cut over until it is explicitly cleared — the stacked follow-up #13599 adds a reset API for that. Shadow mode extends the same diff signal to profiles not yet cut over, so a profile can be validated before enabling DE for it.

The overlay approach (vs. writing the config) was chosen so the application never writes to Superposition, and so the kill switch is a transparent, reversible overlay on top of the operator-owned `routing_result_source` rather than a mutation of it.

## How did you test it?

- `cargo check -p router` (v1 + redis-rs, and v2) — clean; `cargo clippy -p router` — no warnings in touched files; `cargo +nightly fmt --all` applied.
- **Verified end-to-end against a live local stack** (router + real Decision Engine, kill switch at `threshold=2`, a cut-over profile with divergent DE rules):
  - Payments 1–2 routed via the DE result (`paypal_test`); each diff logged `is_equal=false, is_volume=false` and the counter reached `2`; the one-shot `auto_cutover` alarm fired.
  - Payment 3 routed via the **Hyperswitch** result (`stripe_test`) with `overlaying Hyperswitch routing result` logged, and the counter **stayed at 2** (DE no longer consulted). The DB `routing_result_source` stayed `decision_engine` throughout — no config write.
  - Clearing the counter reverted routing to the DE result on the next payment — confirming the overlay is fully reversible.
  - Diff-reason coverage: divergent-order rule → `decision_engine_result_mismatch`; strict-prefix rule (`[stripe_test]` vs `[stripe_test, paypal_test]`) → `decision_engine_length_mismatch`; DE URL pointed at a dead port → `decision_engine_unresponsive` — each counted and tripped the overlay.
  - Volume-split probe: a 50/50 volume rule logged diffs with `is_volume=true` (including a genuinely divergent one) and the counter stayed untouched.
  - Shadow mode: a non-cutover profile served the HS result while shadow tasks logged `is_equal=false` diffs, and the counter stayed **absent** — observation-only confirmed.
- Dashboard: confirmed in the control-center that `is_cutover` (from `/routing/entry`, which uses this same `get_routing_result_source`) is the only routing-source signal it reads, so the overlay flips the dashboard along with routing.

## Checklist


XyneSpaces:
[should-fix] `record_de_diff_and_maybe_trip_kill_switch` should treat length mismatches as diffs, not only `!comparison.is_equal`. With the current guard, a DE result that is a prefix of the HS result (for example DE `[stripe]` vs HS `[stripe, adyen]`) has `is_equal=true` and `is_equal_length=false`, so the kill switch never counts that routing diff.

```suggestion
if !config.enabled || is_volume || (comparison.is_equal && comparison.is_equal_length) {
    return;
}
```

crates/router/src/core/payments/routing/utils.rs
```diff
@@ -1054,10 +1063,291 @@ pub fn compare_and_log_result<T: RoutingEq<T> + Serialize>(
...
+        override_values: serde_json::json!({
+            crate::consts::superposition::ROUTING_RESULT_SOURCE:
+                api_routing::RoutingResultSource::HyperswitchRouting.to_string()
+        }),
+        description: format!(
+            "Routing diff kill switch cutover for profile {}",
+            profile_id.get_string_repr()
+        ),
+        change_reason: "hyperswitch routing diff kill switch tripped".to_string(),
+    };
```
NISHANTH1221: can we use writable sdk rather than directly calling superposition. This is a reference pr: #11454

prajjwalkumar17: Refactored to SuperpositionClient::set_config_value with a WritableConfig impl on the existing RoutingResultSource config marker, same pattern as [#11454](https://github.com/juspay/hyperswitch/pull/11454). The raw HTTP call is removed; write failures still log-and-degrade to the Redis latch as the enforcement.

config/config.example.toml
```diff
@@ -1423,10 +1423,16 @@ billing_connectors_which_require_payment_sync = "stripebilling, recurly" # List
 [open_router]
 dynamic_routing_enabled = true                          # Enable or disable Open Router for dynamic routing
 static_routing_enabled = true           # Enable or disable Open Router for static routing
+shadow_routing_enabled = false  # Shadow-evaluate static routing on DE for non-cutover profiles (diff logging + kill switch only); requires static_routing_enabled
 url = "http://localhost:8080"           # Open Router URL
 dashboard_url = "http://localhost:5173" # Decision Engine dashboard (browser) URL for merchant SSO redirect
 admin_secret = ""                       # Shared secret sent as x-admin-secret on Decision Engine admin endpoints; empty disables the header
 
+[decision_engine_diff_kill_switch]
```
jagan-jaya:
```suggestion
[open_router.diff_kill_switch]
```

jagan-jaya (changes requested):

crates/router/src/core/payments/routing/utils.rs
```diff
@@ -1054,10 +1100,245 @@ pub fn compare_and_log_result<T: RoutingEq<T> + Serialize>(
...
+    let context =
+        crate::core::configs::ConfigContext::new().with("profile_id", profile_id.get_string_repr());
+    let value = api_routing::RoutingResultSource::HyperswitchRouting.to_string();
+
+    match state
+        .superposition_service
+        .set_config_value::<crate::core::configs::dimension_config::RoutingResultSource>(
+            &value, &context,
+        )
+        .await
```
NISHANTH1221: dimension can be built and dimension.set_routing_result_source can be used to create an override

crates/router/src/core/payments/routing/utils.rs
```diff
@@ -1054,10 +1100,241 @@ pub fn compare_and_log_result<T: RoutingEq<T> + Serialize>(
...
+    business_profile: domain::Profile,
+    payment_id: String,
+    backend_input: BackendInput,
+    fallback_config: Vec<RoutableConnectorChoice>,
+    hs_connectors: Vec<RoutableConnectorChoice>,
+    is_volume: bool,
+) {
+    let profile_id = business_profile.get_id().clone();
+
+    if is_de_kill_switch_active(&state, &profile_id).await {
```
jagan-jaya: why do we need a kill switch for shadow mode? 
In shadow mode, all transactions can go via DE in async request. diff checker should always run in shadow mode. kill switch should work only when the merchant is configured to use DE results.

crates/router/src/core/payments/routing/utils.rs
```diff
@@ -1054,10 +1100,241 @@ pub fn compare_and_log_result<T: RoutingEq<T> + Serialize>(
...
+fn de_diff_count_key(profile_id: &id_type::ProfileId) -> String {
+    format!(
+        "routing_decision_engine_{}_diff_count",
+        profile_id.get_string_repr()
+    )
+}
+
+fn de_kill_switch_key(profile_id: &id_type::ProfileId) -> String {
+    format!(
+        "routing_decision_engine_{}_kill_switch",
```
jagan-jaya: This key is equivalent to the use_decision_engine_results config, right? If so, could you explain why we need this?

In live mode, if the results from DE are not equal to the results from HS, we just toggle use_decision_engine_results, and the merchant will start using the HS dashboard and HS routing for live decisions. why do we need this different key?

jagan-jaya (changes requested):

crates/router/src/core/payments/routing/utils.rs
prajjwalkumar17: This was done in order to answer for the latency that can be exerted by Superposition (which is ~15secs) but this is fixed and we entirely rely on Superposition now.

crates/router/src/core/payments/routing/utils.rs
prajjwalkumar17: fixed

jagan-jaya (dismissed):

NISHANTH1221 (dismissed):

jarnura (approved):

jagan-jaya (approved):

PR #13569, opened 2026-08-05
feat(revenue_recovery): support Chargebee payment methods

NISHANTH1221:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

Extends Chargebee revenue recovery webhook handling beyond card-only transactions.

- Maps supported Chargebee payment methods to Hyperswitch payment method data.
- Accepts optional payment method details for non-card transactions.
- Extracts card information only when the payment method supplies card details.
- Supports composite and bare connector payment method reference IDs.
- Safely represents unknown payment methods and gateways without failing webhook deserialization.
- Logs the Chargebee gateway account reference used to verify billing connector configuration.

This PR is the first change in the revenue recovery micro-PR stack and is based directly on `main`.

### Additional Changes

- [ ] This PR modifies the API contract
- [ ] This PR modifies the database schema
- [ ] This PR modifies application configuration/environment variables

## Motivation and Context

Chargebee recovery webhooks were modelled as card-only payloads and assumed payment method details were always populated. Valid non-card events could consequently fail conversion before recovery processing began.

The transformer now accepts the payload shapes emitted for supported recovery payment methods while preserving the existing card flow.

## How did you test it?

- `cargo +nightly fmt --all -- --check`
- `cargo check --package hyperswitch_connectors --no-default-features --features v2,revenue_recovery`
- `cargo check --package router --no-default-features --features release,v2,redis-rs`
- `git diff --check origin/main..HEAD`

Validation was run on the complete stacked branch.

## Checklist

- [x] I formatted the code `cargo +nightly fmt --all`
- [ ] I addressed lints thrown by `cargo clippy`
- [x] I reviewed the submitted code
- [ ] I added unit tests for my changes where possible

## Related Issue

Closes #13653

PR #13570, opened 2026-08-05
fix(revenue_recovery): honor billing retry threshold

### LINKED ISSUE #13654: Honor the billing connector retry threshold in revenue recovery
## Bug Description

Revenue recovery waits for the billing connector while the retry count is less than or equal to its configured threshold. This delays the Hyperswitch recovery handoff by one attempt. New recovery intents without retry metadata also fail instead of starting at zero, and non-actionable webhook events continue through the recovery path.

## Expected Behavior

- Wait only while the retry count is strictly below the billing threshold.
- Start missing retry metadata at zero.
- Return `NoEffect` for events without a valid recovery action.

## Actual Behavior

The threshold comparison allows one extra billing retry, missing retry metadata produces an error, and invalid actions continue through recovery processing.

## Steps To Reproduce

1. Configure a billing connector retry threshold.
2. Process recovery webhooks until the retry count equals the threshold.
3. Observe that recovery is still deferred.
4. Process a new intent without retry metadata or a webhook without a valid action and observe the flow error or continue unnecessarily.

## Context For The Bug

The handoff boundary must match the merchant's configured threshold so Hyperswitch recovery begins at the intended attempt. Implemented by #13570.

## Environment

Hyperswitch revenue recovery v2 flow on the current `main` branch.

## Checks

- [x] I checked and didn't find a similar issue.
- [x] I have read the contributing guidelines.
- [x] I am willing to submit a PR.

NISHANTH1221:
## Type of Change

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

Corrects how incoming revenue recovery webhooks apply the billing connector retry threshold.

- Returns `NoEffect` for webhook events that do not resolve to a valid recovery action.
- Treats missing retry metadata on newly created recovery intents as zero retries.
- Waits for the billing connector only while the retry count is strictly below its configured threshold.
- Allows Hyperswitch recovery to proceed once the threshold is reached.

This PR is intentionally limited to the incoming recovery webhook decision path.

### Stack Dependency

Depends on [#13569](https://github.com/juspay/hyperswitch/pull/13569).

### Additional Changes

## Motivation and Context

The previous `less than or equal to` comparison deferred recovery for one additional attempt after the configured billing threshold had already been reached. Missing retry metadata also caused new recovery intents to fail instead of starting from zero, and irrelevant events continued through a flow that had no actionable recovery operation.

These changes align the handoff boundary with the configured threshold and make the webhook flow tolerant of newly created intents and non-actionable events.

## How did you test it?

Tested it locally.

## Checklist

## Related Issue

Closes #13654


XyneSpaces:
[should-fix] The retry threshold comparison now uses `intent_retry_count < mca_retry_threshold`, so when the count equals the configured threshold Hyperswitch stops waiting for the billing connector one retry earlier than the previous `<=` behavior. If the threshold represents the maximum retry count the billing connector is allowed to perform, keep the boundary inclusive to avoid scheduling recovery while the connector's final retry is still expected.

```suggestion
    (intent_retry_count <= mca_retry_threshold)
```

Context: this should be an inline comment on `crates/router/src/core/webhooks/recovery_incoming.rs`, but the available GitHub write tool in this run only supports PR-level comments.

crates/router/src/core/webhooks/recovery_incoming.rs
```diff
@@ -179,17 +179,29 @@ pub async fn recovery_incoming_webhook_flow(
         action: RecoveryAction::get_action(event_type, attempt_triggered_by),
     };
 
+    // Return early, before the retry threshold is fetched below. Events mapping to `InvalidAction`
```
sai-harsha-vardhan: Do we need all these comments?

crates/router/src/core/webhooks/recovery_incoming.rs
```diff
@@ -179,17 +179,29 @@ pub async fn recovery_incoming_webhook_flow(
...
+    // are not revenue recovery events at all, so they must not fail the webhook when the billing
+    // connector account has no revenue recovery config (and hence no retry threshold) set up.
+    if matches!(
+        recovery_action.action,
```
sai-harsha-vardhan: this check can be an impl on action

crates/router/src/core/webhooks/recovery_incoming.rs
```diff
@@ -257,11 +269,12 @@ async fn handle_schedule_failed_payment(
...
 
-    // When intent_retry_count is less than or equal to threshold
-    (intent_retry_count <= mca_retry_threshold)
+    // When intent_retry_count is strictly less than threshold, billing connector
+    // hasn't retried enough times yet — let it handle retries and wait.
+    (intent_retry_count < mca_retry_threshold)
         .then(|| {
-            logger::error!(
-                "Payment retry count {} is less than threshold {}",
+            logger::info!(
```
sai-harsha-vardhan, line 269: should this be a warn log instead?

NISHANTH1221: By nature these are not warning logs these just provide info for a particular webhook whether we have started the recovery process or not

crates/router/src/core/webhooks/recovery_incoming.rs
NISHANTH1221: Moved it to a impl

crates/router/src/core/webhooks/recovery_incoming.rs
NISHANTH1221: Removed them..

sai-harsha-vardhan (approved):

aniketburman014 (approved):

PR #13583, opened 2026-08-06
docs(openapi): add API reference for list payment methods client SDK endpoint

### LINKED ISSUE #13582: add `/client` endpoint in the api ref
add `/client` endpoint in the api ref

ShankarSinghC:
## Summary
- Adds the missing OpenAPI `utoipa::path` documentation for `GET /payments/{payment_id}/client` (`list_payment_methods_for_payments_client`), which was routed and had a doc comment on the handler but was absent from the generated API reference.
- Registers the endpoint's response schema (`ClientPaymentMethodsListResponse`) and its nested types in `components(schemas(...))` so the spec resolves fully.

## Test plan
- [ ] Regenerate/inspect the OpenAPI spec and confirm `GET /payments/{payment_id}/client` appears under "Payment Methods" with the correct request params and response schema.
image


Sakilmostak (dismissed):

sai-harsha-vardhan (dismissed):

XyneSpaces:
[should-fix] Please make the documented client payment-methods endpoint path consistent across the OpenAPI route and schemas, and fix the `accepted_countries` parameter description to say country code instead of currency code.

The generated path is `/payments/{payment_id}/client`, but the new response schema text documents `GET /payments/{payment_id}/payment-methods/client`; SDK/API consumers following the schema description will call a different route.

Sakilmostak (approved):

PR #13586, opened 2026-08-06
feat(payouts): surface vop response, store VOP

heykulthe:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring
- [x] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

 Surfaces the connector's payout eligibility verdict to merchants, and stops UCS payout declines being reported as internal errors.

### Additional Changes

## Motivation and Context

A payee refusal was indistinguishable from any other ineligible payout the merchant received a status and nothing else, with no way to tell whether the IBAN failed to match the payee name or the payee could not be verified at all. A close match is also a success but not an exact one, and that distinction was being discarded entirely.

## How did you test it?

Refer PR: [2074](https://github.com/juspay/hyperswitch-prism/pull/2074)

## Checklist


XyneSpaces:
[should-fix] Preserve non-object merchant metadata when merging connector metadata. If a payout was created with valid JSON metadata that is not an object (for example a string or array), the new `unwrap_or_default()` path drops it and returns only connector metadata in the payout response instead of preserving the merchant-supplied value.

```suggestion
    let Some(mut merged) = merchant_metadata
        .as_ref()
        .and_then(|metadata| metadata.peek().as_object().cloned())
    else {
        return merchant_metadata.or_else(|| Some(Secret::new(serde_json::Value::Object(connector_details))));
    };
```

Note: this is intended for `crates/router/src/core/payouts/helpers.rs` in `merge_connector_metadata`, on the added `unwrap_or_default()` line. The available API in this run only supports PR-level comments, not inline diff comments.

crates/router/src/core/unified_connector_service/transformers.rs
```diff
@@ -7991,14 +8003,37 @@ impl
...
+                        .and_then(|metadata| {
+                            serde_json::from_str::<serde_json::Value>(metadata.peek()).ok()
+                        })
+                        .and_then(|value| value.as_object().cloned())
+                        .unwrap_or_default();
+
+                    if let Some(reference_id) = response.connector_eligibility_reference_id.clone()
+                    {
+                        details.insert(
+                            consts::PAYOUT_CONNECTOR_ELIGIBILITY_REFERENCE_ID_KEY.to_string(),
```
hrithikesh026: lets not deal with raw serde values. Introduce dedicated fields if needed.

AkshayaFoiger (dismissed):
Connector Change Looks Good

crates/hyperswitch_interfaces/src/unified_connector_service/transformers.rs
```diff
@@ -1020,6 +1020,12 @@ impl ForeignTryFrom<payments_grpc::RedirectForm> for RedirectForm {
                 )
                 .into(),
             ),
+            Some(payments_grpc::redirect_form::FormType::Script(_)) => Err(
+                UnifiedConnectorServiceError::RequestEncodingFailedWithReason(
+                    "Script form type is not implemented".to_string(),
+                )
+                .into(),
+            ),
```
Sakilmostak: is this due to ucs bump?

crates/router/src/core/errors/utils.rs
```diff
@@ -484,6 +486,24 @@ impl<T> ConnectorErrorExt<T> for error_stack::Result<T, errors::ConnectorError>
...
+                    "code": connector_error.code,
+                    "message": connector_error.message,
+                    "reason": connector_error.reason,
+                    "connector": connector_error.connector,
+                    "status_code": connector_error.status_code,
+                });
+                return err
+                    .change_context(errors::ApiErrorResponse::PayoutFailed { data: Some(data) });
+            }
+
```
Sakilmostak: this is inappropriate place to handle this. please don't add new scope for a specifically defined function (in this case conversion from ConnectorError to ApiErrorResponse).

this should be handled at the UCSError to ConnectorError transition

crates/router/src/core/payouts/helpers.rs
```diff
@@ -1817,3 +1817,51 @@ pub fn should_continue_payout<F: Clone + 'static>(
 ) -> bool {
     router_data.response.is_ok()
 }
+
+#[derive(Debug, serde::Serialize)]
+struct PayoutResponseMetadata {
+    #[serde(skip_serializing_if = "Option::is_none")]
+    connector_eligibility_reference_id: Option<String>,
```
Sakilmostak: why this is in metadata?

crates/router/src/core/payouts/helpers.rs
```diff
@@ -1817,3 +1817,51 @@ pub fn should_continue_payout<F: Clone + 'static>(
...
+
+fn metadata_object(
+    metadata: Option<&pii::SecretSerdeValue>,
+) -> serde_json::Map<String, serde_json::Value> {
+    metadata
+        .and_then(|metadata| metadata.peek().as_object().cloned())
+        .unwrap_or_default()
+}
+
+pub fn merge_connector_metadata(
```
Sakilmostak, line 1821: this could be under impl of the respective struct

deepanshu-iiitu (dismissed):

crates/router/src/core/payouts/helpers.rs
```diff
@@ -1817,3 +1817,51 @@ pub fn should_continue_payout<F: Clone + 'static>(
 ) -> bool {
     router_data.response.is_ok()
 }
+
+#[derive(Debug, serde::Serialize)]
+struct PayoutResponseMetadata {
```
Sakilmostak: rename it to 'MergePayoutMetadata' or something better to avoid any confusion

crates/router/src/core/unified_connector_service/transformers.rs
```diff
@@ -7943,7 +7949,12 @@ impl
         let status = common_enums::PayoutStatus::foreign_try_from(response.payout_status())
             .unwrap_or(prev_status);
 
-        let router_response = if let Some(error_info) = response.error {
+        let is_payee_refusal = response.payout_eligible == Some(false);
+        let connector_error = response.error.clone();
+
+        let router_response = if let Some(error_info) =
+            connector_error.clone().filter(|_| !is_payee_refusal)
```
Sakilmostak: is this case valid? do we throw error from ucs if payout_eligible is true?

crates/hyperswitch_interfaces/src/unified_connector_service/transformers.rs
heykulthe: yes it was due to ucs bump, but after recent changes in HS main, this is not required. so removed it.

crates/router/src/core/payouts/helpers.rs
heykulthe: renamed this

crates/router/src/core/unified_connector_service/transformers.rs
heykulthe: no, its not. removed it

crates/hyperswitch_interfaces/src/unified_connector_service/transformers.rs
```diff
@@ -1339,8 +1339,18 @@ impl ErrorSwitch<ConnectorError> for UnifiedConnectorServiceError {
...
+                    "code": inner.code,
+                    "message": inner.message,
+                    "reason": inner.reason,
+                    "connector": inner.connector,
+                    "status_code": inner.status_code,
+                });
+                ConnectorError::ProcessingStepFailed(Some(bytes::Bytes::from(
+                    error_body.to_string(),
+                )))
+            }
```
Sakilmostak: connect with prism team to check why this wasn't handled before

crates/router/src/core/payouts/helpers.rs
```diff
@@ -1817,3 +1817,55 @@ pub fn should_continue_payout<F: Clone + 'static>(
 ) -> bool {
     router_data.response.is_ok()
 }
+
+/// The `metadata` returned on a payout response.
+#[derive(Debug, serde::Serialize)]
+pub struct MergedPayoutMetadata {
+    #[serde(skip_serializing_if = "Option::is_none")]
+    connector_eligibility_reference_id: Option<String>,
```
Sakilmostak: why are we putting connector_eligibility_reference_id in metadata as a separate field? metadata shouldn't handle specific cases at core. since it is properly introduced into payout_attempt table, add it as a separate field in the response directly.

Cargo.toml
```diff
@@ -2,7 +2,7 @@
 resolver = "2"
 members = ["crates/*"]
 package.edition = "2021"
-package.rust-version = "1.85.0"
+package.rust-version = "1.92.0"
```
AmitsinghTanwar007: this might not be needed as i have updated ucs so that dependency graph will not affect this rust version because of ucs

heykulthe: ok, i'll remove this

crates/router/src/core/payouts/helpers.rs
```diff
@@ -1817,3 +1817,37 @@ pub fn should_continue_payout<F: Clone + 'static>(
...
+    let mut merged = merchant_metadata
+        .as_ref()
+        .and_then(|metadata| metadata.peek().as_object().cloned())
+        .unwrap_or_default();
+
+    for (key, value) in connector_details {
+        merged.entry(key).or_insert(value);
+    }
+
+    Some(Secret::new(serde_json::Value::Object(merged)))
```
Sakilmostak: nit: please avoid early return, use match for logic handling

deepanshu-iiitu (dismissed):

crates/hyperswitch_interfaces/src/unified_connector_service/transformers.rs
gopikrishna000: https://github.com/juspay/hyperswitch/pull/13709/changes
this and other error mapping is already handled in this recent pr
its not yet merged

We can remove this change from this pr

crates/router/src/core/payouts/helpers.rs
heykulthe: done

Sakilmostak (dismissed):

AkshayaFoiger (dismissed):

crates/router/src/core/payouts/gateway/eligibility_gateway.rs
```diff
@@ -135,6 +135,9 @@ where
 
                     router_data.connector_http_status_code = Some(ucs_data.status_code);
 
+                    router_data.request.connector_eligibility_reference_id =
```
hrithikesh026: should not populate the request field using response info.  Any connector response field should be propagated through RouterData::response field.

hrithikesh026: include this field in `struct PayoutsResponseData`

crates/router/src/core/payouts.rs
```diff
@@ -2995,14 +3016,18 @@ pub async fn response_handler(
         billing_descriptor: payouts.billing_descriptor.to_owned(),
         entity_type: payouts.entity_type.to_owned(),
         recurring: payouts.recurring,
-        metadata: payouts.metadata.clone(),
+        metadata: helpers::merge_connector_metadata(
```
hrithikesh026: the response `metadata` field should directly map with the metadata column in payout db record. This should be updated in DB.

crates/router/src/core/unified_connector_service.rs
```diff
@@ -3366,6 +3366,7 @@ impl
         Ok(Self {
             router_data_response,
             status_code,
+            connector_eligibility_reference_id: response.connector_eligibility_reference_id,
```
hrithikesh026: this response field can be accommodated in PayoutsResponseData

crates/router/src/core/unified_connector_service/transformers.rs
```diff
@@ -7977,70 +7980,30 @@ impl
...
-                payout_eligible: response.payout_eligible,
-                should_add_next_step_to_process_tracker: false,
-                error_code: None,
-                error_message: None,
-                payout_connector_metadata: None,
-            })
-        };
-
-        Ok(router_response)
+        let connector_details = response
```
hrithikesh026, line 7987: we are no longer returning an error in case of error response from prism.

crates/router/src/types.rs
```diff
@@ -833,6 +833,7 @@ pub struct UcsPayoutTransferResponseData {
 pub struct UcsPayoutEligibilityResponseData {
     pub router_data_response: Result<PayoutsResponseData, ErrorResponse>,
     pub status_code: u16,
+    pub connector_eligibility_reference_id: Option<String>,
```
hrithikesh026: this field needs to be in PayoutsResponseData

crates/router/src/core/unified_connector_service/transformers.rs
heykulthe: That path is a payee refusal, not an error, it's already modelled as NotPermitted

crates/router/src/core/payouts/gateway/eligibility_gateway.rs
heykulthe: Done

hrithikesh026 (approved):

deepanshu-iiitu (approved):

Sakilmostak (approved):

PR #13069, opened 2026-06-29
fix(events): suppress shadow UCS connector logs

gopikrishna000:
## Type of Change

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

Restores the primary-mode guard around UCS connector event emission and restores the UCS config override that disables UCS-side event publishing for shadow mode.

Original PR: https://github.com/juspay/hyperswitch/pull/13068
Tracking issue: https://github.com/juspay/hyperswitch-cloud/issues/17446

Follow-up steps to re-enable shadow UCS connector events:

1. Add and deploy the ClickHouse ingestion migration for connector analytics so `source`, `destination`, and `execution_mode` are persisted for connector event rows.
2. Update dashboard/cloud analytics queries to use the new dimensions, especially `execution_mode`, so shadow UCS rows do not mix with primary connector metrics.
3. Once those are live, remove this temporary suppression and re-enable shadow UCS connector event emission.

### Additional Changes

- [ ] This PR modifies the API contract
- [ ] This PR modifies the database schema
- [ ] This PR modifies application configuration/environment variables

## Motivation and Context

Shadow UCS connector events should not be emitted until downstream ClickHouse storage can reliably distinguish them using the new event metadata fields. Without this guard, shadow events can be published into the connector events stream before the ClickHouse schema migration and dashboard query changes are available.

Cloud tracking issue: https://github.com/juspay/hyperswitch-cloud/issues/17446

## How did you test it?

- `cargo check -p router`

## Checklist

- [ ] I formatted the code `cargo +nightly fmt --all`
- [ ] I addressed lints thrown by `cargo clippy`
- [x] I reviewed the submitted code
- [ ] I added unit tests for my changes where possible

PR #13611, opened 2026-08-10
fix(ucs): allow shadow event publishing

gopikrishna000:
## Summary
- remove the shadow-only `events.enabled=false` config override sent to UCS
- remove the HS-side `ExecutionMode::Primary` guards around UCS connector-event emission
- this fully reverts the shadow event suppressions introduced in #13069

## Why
- prod certification needs shadow connector events persisted in ClickHouse for both HS and Prism/UCS event streams
- HS analytics APIs still filter shadow connector events out of dashboard responses through `connector_events_core(...).filter(|event| !event.is_shadow())`

## Verification
- `cargo fmt --all --check`
- verified both connector-event API routes (`Hyperswitch` and `Prism`) go through the shared `connector_events_core` shadow filter
- verified HS shadow rows are tagged via `EventExecutionMode::from(execution_mode)`
- verified UCS still receives `x-shadow-mode` and no longer receives the events-disabled config override

### DIFF
--- a/crates/router/src/core/unified_connector_service.rs
+++ b/crates/router/src/core/unified_connector_service.rs
@@ -2974,23 +2974,21 @@ where
         }
     };
 
-    if let ExecutionMode::Primary = execution_mode {
-        emit_ucs_connector_event(
-            state,
-            std::any::type_name::<T>(),
-            connector_name,
-            payment_id,
-            merchant_id,
-            refund_id,
-            dispute_id,
-            payout_id,
-            grpc_request_body,
-            status_code,
-            response_body,
-            external_latency,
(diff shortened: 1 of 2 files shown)


JeevaRamu0104 (approved):

XyneSpaces:
[should-fix] `crates/router/src/routes/app.rs:204`

Replacing the shadow-only `events.enabled=false` override with `config_override(None)` removes explicit suppression. Add a comment stating that shadow events are intentionally emitted and filtered downstream, so future refactors do not accidentally re-enable the suppression.

su-shivanshmathur (approved):

PR #13614, opened 2026-08-10
ci(deja): gate PRs on the deja-art label and the replay verdict

iemyashasvi:
The gate fails until the deja-art label is present, then polls the 'Deja ART Replay' check-run that Jenkins posts on the head commit and passes only when it concludes success. Uses only the built-in token (checks: read); no internal endpoints are touched.

## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [x] CI/CD

### Additional Changes

## Checklist

### DIFF
--- a/.github/workflows/deja-art-gate.yml
+++ b/.github/workflows/deja-art-gate.yml
@@ -1,0 +1,81 @@
+name: Deja ART Gate
+
+on:
+  pull_request:
+    types:
+      - opened
+      - reopened
+      - synchronize
+      - labeled
+      - unlabeled
+
+permissions:
+  contents: read
+  checks: read
+
+concurrency:
+  group: deja-art-gate-${{ github.event.pull_request.number }}
+  cancel-in-progress: true
+
+jobs:
+  deja-art-gate:
+    name: Deja ART verdict
+    runs-on: ubuntu-latest
+    timeout-minutes: 180
+
+    steps:
+      - name: Require the deja-art label
+        env:
+          LABELS: ${{ toJSON(github.event.pull_request.labels.*.name) }}
+        run: |
+          if echo "${LABELS}" | jq -e 'index("deja-art")' > /dev/null; then
+            echo "deja-art label present."
+          else
+            echo "::error::Add the 'deja-art' label to run a Deja ART replay for this pull request."
+            exit 1
+          fi
+
+      - name: Wait for the Deja ART Replay verdict
+        env:
+          GH_TOKEN: ${{ github.token }}
+          HEAD_SHA: ${{ github.event.pull_request.head.sha }}
+        run: |
+          echo "Waiting for the 'Deja ART Replay' check on ${HEAD_SHA}..."
+          ABSENT_POLLS=0
+          # 30s x 340 polls stays inside the 180-minute job timeout.
+          for _ in $(seq 1 340); do
+            RUN=$(
+              gh api "repos/${{ github.repository }}/commits/${HEAD_SHA}/check-runs?check_name=Deja%20ART%20Replay" \
+                --jq '.check_runs | sort_by(.started_at) | last'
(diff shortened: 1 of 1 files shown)


.github/workflows/deja-art-gate.yml
```diff
@@ -0,0 +1,81 @@
+name: Deja ART Gate
+
+on:
+  pull_request:
+    types:
+      - opened
+      - reopened
+      - synchronize
+      - labeled
+      - unlabeled
```
SanchithHegde, line 10: From what I remember, these triggers cause the workflow to be triggered in case any label is added or removed. Is that the intended behavior?

.github/workflows/deja-art-gate.yml
```diff
@@ -0,0 +1,81 @@
...
+          fi
+
+      - name: Wait for the Deja ART Replay verdict
+        env:
+          GH_TOKEN: ${{ github.token }}
+          HEAD_SHA: ${{ github.event.pull_request.head.sha }}
+        run: |
+          echo "Waiting for the 'Deja ART Replay' check on ${HEAD_SHA}..."
+          ABSENT_POLLS=0
+          # 30s x 340 polls stays inside the 180-minute job timeout.
```
SanchithHegde, line 45: Are we wasting runner minutes just to poll? Can we find a better solution?

.github/workflows/deja-art-gate.yml
```diff
@@ -0,0 +1,81 @@
...
+          for _ in $(seq 1 340); do
+            RUN=$(
+              gh api "repos/${{ github.repository }}/commits/${HEAD_SHA}/check-runs?check_name=Deja%20ART%20Replay" \
```
SanchithHegde, line 48: Nit, optional change: Can use hyphenated strings/names (`Deja-ART-Replay`) if that helps simplify things.

XyneSpaces:
[should-fix] `.github/workflows/deja-art-gate.yml:9`

`labeled`/`unlabeled` triggers run the workflow on every label change, not just `deja-art`. Add a job-level `if: contains(github.event.pull_request.labels.*.name, 'deja-art')` to avoid wasted runs and noisy failures.

[should-fix] `.github/workflows/deja-art-gate.yml:45`

Busy-polling for up to 170 minutes ties up a GitHub-hosted runner. Replace active polling with a `workflow_run`-triggered workflow or a check-run webhook so the gate is event-driven.

maverox:
Closing this one and keeping https://github.com/juspay/hyperswitch/pull/14181 as the label check.

Three pull requests are open for the same gate: this one, https://github.com/juspay/hyperswitch/pull/14098 and https://github.com/juspay/hyperswitch/pull/14181. Only one can define the check, and two of them share a head branch name under different authors, which is why they have read as a single entry in listings and sat open beside each other for weeks. A required check that never merges is indistinguishable from one that did until somebody goes and looks at the gate itself, so these are better closed than left open.

The reason this is not the survivor is not the label step, which is sound. It is that a required check has to report a result in the merge queue, and this workflow has no `merge_group` trigger, so an entry reaching the queue waits on a check that will never run there. Both of the others handle that case explicitly. Separately, the verdict wait holds a runner for up to three hours polling for another check's conclusion.

The verdict gate itself is still wanted, and this is the only one of the three that attempted it — the label alone says a replay was asked for, not that it passed. It belongs as a required status check on the replay's own check-run, reported by whatever runs the replay, rather than as a job that polls for it. That is a different change from this one, and worth raising separately once the replay side is ready to report a verdict.

Noting also that the review point left here, that `labeled` and `unlabeled` fire on every label change rather than only on `deja-art`, is a good one and still applies to the surviving workflow.

PR #13629, opened 2026-08-11
fix(db): recover from Aurora/RDS failover hangs via tcp_user_timeout

### LINKED ISSUE #13631: AWS DB failover resilience for Hyperswitch application
# DB failover resilience
## Summary

During RDS/Aurora failover testing (both simulated network partitions and a
real Multi-AZ failover against a dev Aurora cluster), we found **two
distinct, independently-caused** ways the router's database layer misbehaves
after a failover — one that can leave the process permanently stuck until a
manual restart, and one that causes silent, ongoing write failures without
ever hanging. They have different root causes and different fixes:

| | Problem | Root cause | Fix |
|---|---|---|---|
| 1 | Pool-wide hang, doesn't self-recover | `async-bb8-diesel`'s shared connection-manager mutex | `tcp_user_timeout` on the DB connection string ([hyperswitch#13629](https://github.com/juspay/hyperswitch/pull/13629)) |
| 2 | Connects to the old (demoted) writer, writes fail | No connection ever proactively expires for 30 mins; health checks don't test writability | `max_lifetime` on the pool config of 2 mins |

---

### Symptom

After some failovers, `hyperswitch` stops serving any DB-backed traffic and
does not recover on its own — a manual process restart is required. Locker,
hit by the same failover, never needs this.

### Root cause

`async-bb8-diesel` (the sync-to-async bridge `hyperswitch` needs, since
`diesel::PgConnection` is blocking and the router is async) wraps its
connection manager in **one mutex shared by the entire pool**:

```rust
// async-bb8-diesel-0.2.1/src/connection_manager.rs
pub struct ConnectionManager<T> {
    inner: Arc<Mutex<r2d2::ConnectionManager<T>>>,   // ONE instance, whole pool
}

async fn run_blocking<R, F>(&self, f: F) -> R {
    let cloned = self.inner.clone();
    tokio::task::spawn_blocking(move || f(&*cloned.lock().unwrap())).await.unwrap()
}

impl<T> bb8::ManageConnection for ConnectionManager<T> {
    async fn connect(&self) -> ... { self.run_blocking(|m| m.connect()) }   // locks it
    async fn is_valid(&self, conn) -> ... { self.run_blocking(...) }        // locks the SAME lock
    fn has_broken(&self, _: &mut Self::Connection) -> bool { false }        // no cheap check exists
}
```

Aurora/RDS failover leaves connections to the old writer **silently
unresponsive** — no TCP RST/FIN (AWS's own documented behavior: "the client
no longer gets TCP traffic ... it's up to the client to time out"). When
that happens to a connection mid-query, or mid-health-check, the blocking
libpq call underneath hangs **forever**, holding the shared mutex. Every
other pool operation — including brand-new, otherwise-healthy connections —
queues behind a lock that will never release, because:

- `has_broken()`, bb8's cheap non-blocking pre-check, is hardcoded `false`,
  so there's no fast path to skip a bad connection.
- bb8's own self-healing Reaper task (which fires every 30s to spawn
  replacement connections) doesn't help either — every replacement attempt
  calls the same `connect()`, which locks the same mutex.

**Locker has no equivalent lock.** `diesel-async` is natively async (backed
by `tokio-postgres`), so it needs no sync-to-async bridge, so there's no
shared resource for `deadpool`'s per-connection `recycle()`/`create()` calls
to contend on. One dead connection only affects whoever was using it.

### Validated by direct reproduction

Using a real Linux container with genuine `iptables`-level packet loss (not
just a proxy simulation, which can't test TCP keepalive behavior — a proxy's
own kernel still ACKs keepalive probes even when the app logic is frozen):

- One dead bb8 connection, pool otherwise healthy: **zero** new connections
  ever succeeded, regardless of pool size or `min_idle`/`max_size` tuning.
- Same test against deadpool with spare capacity: 5 brand-new connections
  succeeded in ~24ms while the 5 originals sat permanently stuck.
- Full end-to-end replication of the actual Aurora failover shape (old
  instance silently dies, a second healthy instance stands in for the
  promoted writer, DNS re-points at the failover moment) — using
  hyperswitch's **exact, unmodified** dependency versions
  (`async-bb8-diesel = "0.2.1"`, `bb8 = "0.8.6"`), zero application code
  changed: all 5 in-flight queries against the dying instance failed on
  their own after ~11s via a genuine OS-level error
  (`"could not receive data from server: Connection timed out"`), then all
  16/16 subsequent probes against the new writer succeeded immediately.

Full investigation, including several disproven hypotheses (TCP keepalives
alone are not sufficient; `tcp_user_timeout` alone bounds a fresh connect but
not an already-established connection's stuck query) is in
`RCA-db-pool-recovery.md` and `INCIDENT-postmortem-db-failover-recovery.md`
in this directory.

### Fix

Add `tcp_user_timeout` to the Postgres connection string — bounds how long
*newly transmitted* data (notably the pool's own checkout health-check ping)
may go unacknowledged before the OS forcibly closes the socket, producing a
real, catchable error instead of an infinite hang. This lets bb8's **already
correct, unmodified** "mark connection invalid → discard → reconnect" logic
actually fire.

**Sizing constraint (validated, not assumed):** `tcp_user_timeout` must stay
comfortably below the pool's `connection_timeout` (10s in hyperswitch's
config everywhere). `connection_timeout` wraps the *entire* checkout,
including the health-check ping this bounds — if the outer timeout wins that
race, the pool's "mark invalid" branch never executes, and the dead
connection is returned to the idle queue looking healthy. Measured directly:
at `tcp_user_timeout = 10_000ms` (equal to `connection_timeout`), one
checkout hard-failed at exactly `10.002s` and recovery took ~16s; at
`5_000ms` there were zero failures and recovery took ~12s.

**Scope note:** an earlier draft of this fix also added TCP keepalive
tuning (`keepalives_idle`/`interval`/`count`) alongside `tcp_user_timeout`,
on the theory that keepalives cover an *already in-flight* query going
silent, while `tcp_user_timeout` alone only covers a *newly-idle* connection
being pinged. Keepalives were deliberately **left out of the shipped PR** to
keep the change minimal and easiest to reason about/review; `tcp_user_timeout`
alone is what's proven necessary for the specific "connection goes silent,
next thing sent to it never gets acknowledged" failure this issue is about.
Revisit keepalive tuning as a separate, follow-up change if the in-flight-query
path turns out to need it independently.

PR: **https://github.com/juspay/hyperswitch/pull/13629**

---

### Symptom

Distinct from problem 1 — **no hang at all.** The pool stays fully
responsive throughout the failover, but some writes fail (Postgres rejects
them: "cannot execute INSERT/UPDATE in a read-only transaction"), because
the connection serving them is still talking to the instance that *used* to
be the writer.

Reproduced directly against locker's own vault (Aurora Multi-AZ, real
failover, no simulation) while load-testing the `payments/confirm` →
`setup_future_usage=off_session` → locker-save path: **5 requests failed in
a ~5 second window** (`18:13:47`–`18:13:52`) with locker's own error
response:

```
Locker error response: {"code":"TE_01","message":"Failed while connecting to database"}
Locker error response: {"code":"TE_01","message":"Failed while inserting data into locker"}
```

Confirmed against the actual affected payments (looked up by `payment_id`
extracted from the request logs): all 5 came back `status: succeeded` with
`payment_method_id: null` — **the charge succeeded, but the card silently
failed to save.** Nothing about the payment's own status reflects this;
without the `payment_method_id` check we added to the load-test script
(`load-tests/payments-flow.js`, `hyperswitch-infra` repo) afterward, this
would be invisible.

### Root cause

Neither `bb8`'s `is_valid()` nor `deadpool`'s `recycle()` health-check tests
**writability** — both just run a cheap `SELECT 1`. That succeeds fine
against a demoted-but-reachable old primary (Postgres only rejects writes on
a read replica, not reads), so both pools confidently keep handing out a
connection that "passes health check" but is pointed at the wrong node. The
error only surfaces at the SQL layer, on the next actual write.

**Without an enforced connection lifetime, a stale connection like this can
survive indefinitely** — DNS re-resolution for Aurora's writer endpoint only
happens when a *new* physical connection is made; an existing, still-alive
TCP session never re-resolves on its own no matter how long it's held open.

- `bb8` has **no periodic expiry unless `max_lifetime` is explicitly set** —
  confirmed as the actual gap in `hyperswitch`'s config (was unset
  everywhere prior to this fix).
- `deadpool` has **no `max_lifetime` concept at all** — confirmed against
  its own source (`PoolConfig`/`Builder` expose only `max_size` and
  `Timeouts{wait, create, recycle}`, no lifetime cap).

So in principle, both stacks share this exact blind spot during the window
before the old instance's session is actually torn down. The reason
locker's observed impact was ~5 seconds rather than open-ended is most
likely `diesel-async`'s `is_broken()` check, which (unlike bb8's
hardcoded-`false` `has_broken()`) includes a real socket-liveness check:

```rust
// diesel-async 0.5.0, src/pg/mod.rs:836-839
fn is_broken(&mut self) -> bool {
    Self::TransactionManager::is_broken_transaction_manager(self) || self.conn.is_closed()
}
```

Once AWS's failover process actually closes the old instance's lingering
sessions (not instantaneous with the role flip, but not indefinite either),
`is_closed()` flips true and — because `deadpool` has no shared mutex
(problem 1) — that connection gets discarded and replaced independently,
without blocking the rest of the pool. This is a plausible mechanism, not a
fully isolated proof — see "Open questions" below.

**Note:** we also checked whether locker's card-insert path wraps writes in
an explicit `diesel-async` `.transaction()` (which would make the *other*
half of `is_broken()`, the transaction-manager-error check, relevant too).
It doesn't — `insert_locker` in `hyperswitch-card-vault/src/storage/db.rs`
is a single autocommit `diesel::insert_into(...).get_result()` call, so that
mechanism doesn't apply to this specific path.

### Fix

Add `max_lifetime` to `master_database`'s pool config, forcing periodic
connection expiry so the pool is guaranteed to make fresh connections on a
schedule — each of which re-resolves the (Aurora-managed) writer-endpoint
DNS and lands on whichever instance is currently the writer.

sai-harsha-vardhan:
## Problem

After some RDS/Aurora failovers, the router stops serving DB-backed traffic and does not recover on its own — a manual process restart is required.

## Root cause

`async-bb8-diesel` (the sync-to-async bridge between diesel's `PgConnection` and `bb8`, needed because the router is async but diesel's connection is blocking) serializes **every** connection health-check and **every** new-connection creation behind one mutex shared by the entire pool:

```rust
// async-bb8-diesel-0.2.1/src/connection_manager.rs
pub struct ConnectionManager<T> {
    inner: Arc<Mutex<r2d2::ConnectionManager<T>>>,   // ONE instance, whole pool
}
impl<T> bb8::ManageConnection for ConnectionManager<T> {
    async fn connect(&self) -> ... { self.run_blocking(|m| m.connect()) }   // locks it
    async fn is_valid(&self, conn) -> ... { self.run_blocking(...) }        // locks the SAME lock
    fn has_broken(&self, _: &mut Self::Connection) -> bool { false }        // no cheap check exists
}
```

Aurora/RDS failover leaves connections to the old writer **silently unresponsive** (no TCP RST/FIN — AWS's own documented behavior). When that happens to a connection mid-query or mid-health-check, the blocking libpq call underneath hangs indefinitely while holding the shared mutex — wedging every other pool operation, including brand-new connections that would otherwise succeed instantly. `has_broken()` (bb8's cheap, non-blocking pre-check) is hardcoded `false` in this bridge, so there's no fast path around it, and bb8's own self-healing Reaper task doesn't help either, since its replacement `connect()` calls route through the same lock.

Verified by direct reproduction: a real Linux container with genuine `iptables`-level packet loss showed one dead bb8 connection freezing the **entire** pool (0/N new connections ever succeeded), while an equivalent `diesel-async`/`deadpool` stack (no shared lock — natively async, no bridge needed) recovered new connections in ~24ms.

## Fix

Add `tcp_user_timeout` to the Postgres connection string. It bounds how long *newly-transmitted* data — notably the pool's own checkout health-check ping — may go unacknowledged before the OS forcibly closes the socket, producing a real, catchable error instead of an infinite hang. This lets bb8's already-correct, unmodified "mark connection invalid → discard → reconnect" logic actually fire — no fork, no application-level timeout wrapper, hyperswitch's existing `bb8`/`async-bb8-diesel` versions unchanged.

**Sizing (validated, not guessed):** set to `5000ms`, which must stay comfortably below the pool's `connection_timeout` (10s everywhere in this codebase). `connection_timeout` wraps the *entire* checkout including this ping — if the outer timeout won that race instead, the pool's "mark invalid" branch would never execute and the dead connection would be returned to the idle queue looking healthy. Measured directly against a real failover reproduction: at `10_000ms` (equal to `connection_timeout`) one checkout hard-failed at exactly `10.002s`; at `5_000ms` there were zero failures and recovery was faster.

**Scope:** deliberately limited to `tcp_user_timeout` alone, without TCP keepalive tuning, to keep this change minimal and easy to review. Keepalives cover a different path (an already in-flight query going silent before this ping-based detection engages) and can be addressed separately if that path is shown to need it independently.

## Testing
- Sanity Payments Testing

### DIFF
--- a/crates/common_utils/src/lib.rs
+++ b/crates/common_utils/src/lib.rs
@@ -392,9 +392,30 @@ pub trait DbConnectionParams {
     fn get_host(&self) -> &str;
     fn get_port(&self) -> u16;
     fn get_dbname(&self) -> &str;
+
+    /// Milliseconds of unacknowledged *transmitted* data tolerated before the
+    /// OS forcibly closes a database connection. This bounds how long a
+    /// connection checkout's health-check ping (`bb8`'s `is_valid()`, a
+    /// `SELECT 1`) can hang against a peer that went silent without ever
+    /// sending a TCP RST/FIN -- exactly what Aurora/RDS failover does to
+    /// connections left open against the old writer. Without this, such a
+    /// checkout hangs indefinitely.
+    ///
+    /// Must stay comfortably below the pool's `connection_timeout` (10s by
+    /// default): `connection_timeout` wraps the entire checkout, including
+    /// this ping. If the outer timeout wins that race, the pool's own "mark
+    /// connection invalid" path never runs, and the dead connection is
+    /// returned to the idle queue looking healthy. Measured against a real
+    /// failover with an all-idle pool: at 10_000ms (equal to
+    /// `connection_timeout`) one checkout hard-failed at 10.002s; at
+    /// 5_000ms there were no failures and recovery was faster.
+    fn get_pool_tcp_user_timeout_ms(&self) -> u64 {
+        5_000
+    }
+
     fn get_database_url(&self, schema: &str) -> String {
         format!(
-            "postgres://{}:{}@{}:{}/{}?application_name={}&options=-c%20search_path%3D{}",
+            "postgres://{}:{}@{}:{}/{}?application_name={}&options=-c%20search_path%3D{}&tcp_user_timeout={}",
             self.get_username(),
             self.get_password().peek(),
             self.get_host(),
@@ -402,6 +423,7 @@ pub trait DbConnectionParams {
             self.get_dbname(),
             schema,
             schema,
+            self.get_pool_tcp_user_timeout_ms(),
         )
     }
 }


XyneSpaces:
[should-fix] `crates/common_utils/src/lib.rs:414`

The 5-second `tcp_user_timeout` is hard-coded in the default trait method. Make it configurable under `[database]` (similar to `connection_timeout`) so operators can tune or disable it per environment without recompiling.

```suggestion
// Example: read from config
let tcp_user_timeout = config.tcp_user_timeout.unwrap_or(Duration::from_secs(5));
```

abhisheksharma2411 (commented):
This is the right mechanism for the failure it describes, and the docstring is unusually good — a silent peer with no RST/FIN is exactly what an Aurora writer failover leaves behind, and bounding it at the socket rather than hoping the application notices is the correct layer. The measured comparison (10_000ms → a checkout hard-failing at 10.002s; 5_000ms → no failures) is the part that makes this reviewable rather than a guess.

One finding I think matters, and three smaller notes.

## The ordering invariant is stated but not enforced, and config can silently invert it

The docstring makes the safety property explicit:

> Must stay comfortably below the pool's `connection_timeout` (10s by default) … If the outer timeout wins that race, the pool's own "mark connection invalid" path never runs, and the dead connection is returned to the idle queue looking healthy.

I checked the default and it holds — `connection_timeout: 10` in `crates/storage_impl/src/config.rs:82`, used at `store.rs:172`. But `connection_timeout` is operator-configurable, and `get_pool_tcp_user_timeout_ms()` is a trait default returning a hardcoded `5_000` that cannot see it. So an operator who tunes `connection_timeout = 3` inverts the invariant, and the failure mode is the one the docstring warns about: dead connections returning to the idle queue looking healthy — worse than not having the setting at all, and with nothing logged to say so.

`Database` already carries `connection_timeout`, and both implementors (`storage_impl/src/config.rs:38`, `drainer/src/settings.rs:91`) could override the default to derive it, e.g. clamp to half the checkout budget:

```rust
fn get_pool_tcp_user_timeout_ms(&self) -> u64 {
    (self.connection_timeout * 1000 / 2).clamp(1_000, 5_000)
}
```

or keep the constant and assert the relationship at startup. Either turns a documented assumption into an enforced one. As it stands the comment is the only thing holding it.

## It lands on three pools, and the justification only analyses one

`get_database_url` feeds `storage_impl/src/database/store.rs:166`, `drainer/src/connection.rs:26`, and `analytics/src/sqlx.rs:59`. The reasoning is written in terms of `bb8`'s `is_valid()` ping and `bb8`'s `connection_timeout`; the analytics pool is **sqlx**, with `.acquire_timeout(...)` and no `is_valid()` in that sense.

It happens to be fine — sqlx reads the same `conf.connection_timeout`, so the ordering holds by coincidence of sharing one config field. Worth a sentence saying so, because the next person to change either pool's timeout independently will not know the two are coupled through this constant.

## Two portability notes

- **`tcp_user_timeout` is effectively Linux-only.** libpq documents it as ignored on platforms without `TCP_USER_TIMEOUT`, so this is safely a no-op on macOS/BSD dev machines — but it also means the failover behaviour cannot be reproduced locally there, which is worth knowing before someone tries and concludes the setting does nothing.
- **It requires libpq ≥ 12.** An older client library rejects the unknown keyword at connect time rather than ignoring it. Almost certainly a non-issue in your build environments, but it is a hard failure rather than a soft one if it ever isn't.

None of this blocks the change — an unbounded hang on failover is strictly worse than any of the above. The first point is the one I would want addressed before merge, since it is the difference between a guarantee and a comment.

PR #13639, opened 2026-08-11
feat(ucs): add automatic kill switch for deterministic UCS failures

### LINKED ISSUE #13645: feat(ucs): add automatic kill switch for UCS failures
## Description

Add an automatic kill switch that diverts a `ucs_rollout_config` scope back to the direct connector integration when UCS fails for it.

Hyperswitch's direct connector integrations have served live merchants for years. UCS is being rolled out behind `ucs_rollout_config` keys, and promoting a key from shadow to primary is the point at which a UCS regression starts reaching a merchant. This kill switch bounds that exposure — when a rollout scope fails, its traffic is cut back to the direct integration without waiting for a human.

## Key Design Decisions

- **Off by default.** The `ucs_kill_switch_enabled` config key turns it on at runtime, with no redeploy. While off, the failure metric is still emitted so the classifier can be validated against real traffic.
- **Conservative tripping.** All errors trip the kill switch, including `ConnectorError`. A false bypass to the battle-tested direct path is safe, while a missed trip leaves merchants on a potentially broken UCS path.
- **Cutover state is separate from rollout intent.** Trips live in Redis under `ucs_kill_switch_<scope>`. The `ucs_rollout_config` rows are never written to.
- **Shadow fallback.** A tripped scope serves from the direct integration but keeps mirroring to UCS, preserving the comparison signal needed to confirm a fix.
- **Admin endpoints.** `GET /unified-connector-service/kill-switch` lists tripped scopes. `DELETE /unified-connector-service/kill-switch/{scope}` clears one.

## Motivation

Rollout keys are being promoted from shadow to primary in batches. This is the guardrail for that promotion.

gopikrishna000:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

Adds an automatic kill switch that serves a `ucs_rollout_config` scope from the direct connector
integration once UCS has failed for that scope.

Promoting a rollout key from `shadow` to `primary` is the point at which a UCS regression starts
reaching a merchant. Hyperswitch's direct integrations have served live merchants for years; this
bounds the exposure of moving off them, without requiring a human to hand-edit config rows during
an incident.

**Off by default.** The `ucs_kill_switch_enabled` config key turns it on at runtime, with no
redeploy. While off, the failure metric is still emitted, so the classifier can be validated
against real traffic before the switch is turned on.

**The lookup only runs once a rollout config has already resolved to `primary`.** Shadow traffic
never reaches it, so the added redis read is scoped to exactly the scopes being protected rather
than to every request. The write happens at most once per scope.

**Propagation needs no invalidation.** The trip is read straight from redis with nothing cached in
front of it, so a trip written by one pod is visible to every other pod on its next request. There
is no cache to go stale and no invalidation channel to depend on.

### Trip state is separate from rollout intent

The trip lives in redis under `ucs_kill_switch_<scope>`. The `ucs_rollout_config_*` rows are never
written to.

A rollout row declares what the migration *intends*; a trip records what *happened*. Keeping them
apart buys two properties:

- **Re-applying the intended config set cannot silently undo a trip.** A bulk enable of the next
  batch, a runbook re-run, or a restore from a known-good list would otherwise write
  `execution_mode: primary` back over a tripped scope, removing protection with no signal to
  anyone.
- **A defect in this module cannot corrupt a row that decides routing.** The worst it can do is
  trip a healthy scope, which serves that scope from the integration it used before UCS.

`SET NX` makes the write exactly-once fleet-wide: concurrent failures all attempt it, one wins, the
rest are a no-op. Trips carry a 7 day TTL and are cleared by an operator through the endpoint
below.

### It fires on the first failure, not on a count

A threshold measures volume, not health. At any realistic threshold a fully broken low-volume scope
never reaches it, so the long tail of a rollout would be silently unprotected while busy healthy
scopes trip on background noise. There is also no production data to calibrate a threshold against:
every rollout key provisioned today is in shadow, so the switch lands before the first promotion.

### Every UCS failure trips

`ucs_kill_switch_reason` in `hyperswitch_interfaces` classifies the failure. Reasons are named for
where the failure originated, so the metric and the log line say who to go and look at:

| reason | origin |
|---|---|
| `hyperswitch_request_invalid` | Hyperswitch could not build the request; UCS was never reached |
| `hyperswitch_response_undecodable` | Hyperswitch could not decode what UCS returned |
| `ucs_rejected_request` | UCS rejected what Hyperswitch sent it |
| `ucs_flow_unsupported` | UCS does not implement this flow for this connector |
| `ucs_internal_error` | UCS failed internally, or failed the flow without saying more |
| `ucs_unreachable` | UCS could not be reached, or was too unwell to answer |
| `connector_outcome` | the connector answered with an error |

A connector rejection may be a legitimate decline or a request UCS built wrongly, and the two are
indistinguishable at this layer, so it trips like the rest: a false trip costs a scope reverting to
the integration it was already using, while a missed trip leaves a merchant on a broken UCS path.

The bias throughout is deliberate. False positives are acceptable; false negatives are not.

The match is exhaustive by design — a new `UnifiedConnectorServiceError` variant fails to compile
here and forces an explicit decision, rather than falling into a catch-all that picks a side
silently.

### It diverts to shadow, not to direct

A tripped scope is served by the direct integration but keeps mirroring to UCS, so the comparison
signal needed to confirm a fix survives the trip. Shadow falls back to direct on its own when no
proxy override is configured.

### Two paths are never diverted

- **UCS-only connectors** (`ConnectorIntegrationType::UcsConnector`) have no direct path to fall
  back to. Diverting them would not protect a merchant; it would be the outage.
- **`CallConnectorAction::UCSConsumeResponse`** carries a response UCS has already produced.
  Handing it to another gateway errors out.

Both are covered by `is_kill_switch_applicable`.

### Scope granularity

The trip is keyed on the **rollout scope** — the rollout config key without its prefix — so it
targets exactly the key that enabled the traffic:

```
merchant_1_cybersource_card_Authorize                 card
merchant_1_cybersource_wallet_google_pay_Authorize    wallet, trips independently
merchant_1_cybersource_Execute                        refunds carry no payment method
```

Card and wallet are enabled by separate rollout keys and have different failure characteristics, so
they trip separately. For payments and refunds, `build_rollout_keys_by_precedence` and the kill
switch both derive their keys from `build_merchant_rollout_scope`, so a trip targets exactly the
key that enabled the traffic.

Payouts are the exception: `build_rollout_keys_for_payouts` keys on payment method *type* where the
scope keys on payment *method*, so a tripped payout scope does not read back as its rollout config
key. Enforcement is unaffected — the recording and enforcement sites build the scope the same way —
but an operator clearing one has to map it to the config row by hand.

## What this does not cover

- **Semantically wrong but well-formed responses.** A status mapped incorrectly deserializes
  cleanly and errors nowhere, so nothing here observes it. Shadow comparison is what catches that,
  and a primary scope has no shadow to compare against.
- **Webhooks.** `ucs_webhook_logging_wrapper` has no `merchant_id` at the call site, so no scope
  can be built.

A kill switch is damage-limiting, not damage-preventing.

## Additional Changes

New config key (runtime, in the `configs` table, not a file): `ucs_kill_switch_enabled`.

New admin endpoints:

```
GET    /unified-connector-service/kill-switch           list tripped scopes
DELETE /unified-connector-service/kill-switch/{scope}   clear one
```

`{scope}` takes either form the list endpoint hands back: the whole redis key, or the scope alone.
The list answers with whole keys so one can be pasted back verbatim, and so the `ucs_kill_switch`
prefix makes it obvious which namespace is being cleared.

The list endpoint is not a convenience — with a large number of provisioned rollout keys, an
on-call engineer cannot reconstruct which scopes are tripped from logs.

Metrics: `UCS_KILL_SWITCH_FAILURE` (qualifying failures seen, emitted whether or not the switch is
on) and `UCS_KILL_SWITCH_TRIPPED` (scopes diverted). Each trip also logs at `warn` with the scope
and reason, so an alert can be built on either.

## Motivation and Context

Rollout keys are being promoted from shadow to primary in batches. This is the guardrail for that
promotion.

## How did you test it?

Unit tests cover the classifier boundary and the scope shape, which is where the risk sits:

- every `UnifiedConnectorServiceError` shape maps to its stated reason, including gRPC status codes
  across the availability, rejection, unimplemented and internal groups
- connector errors trip, including the synthetic 504 a connector timeout produces
- a connector error carrying no status code is classified as UCS-side rather than connector-side
- reason tags are unique, so the metric cannot silently merge two causes
- the scope matches the rollout key shape, including the wallet payment-method-type variant and the
  refund shape that carries no payment method
- independently enabled keys (card vs wallet, different merchant / connector / flow) get
  independent scopes
- the trip key cannot collide with a rollout config key

```
cargo test -p router --features v1 --lib kill_switch                  7 passed
cargo test -p hyperswitch_interfaces --features v1 --lib kill_switch  2 passed
```

`just clippy`, `just clippy_v2` and `just clippy fred` are clean.

Closes #13645


crates/router/src/core/unified_connector_service/kill_switch.rs
```diff
@@ -0,0 +1,437 @@
...
+    context: UcsFailureContext<'_>,
+    execution_mode: ExecutionMode,
+    error: &UnifiedConnectorServiceError,
+) {
+    // Only the path serving merchant traffic can trip.
+    if !matches!(execution_mode, ExecutionMode::Primary) {
+        return;
+    }
+
+    let Some(reason) = error.ucs_kill_switch_reason() else {
```
AmitsinghTanwar007: as here we are avoiding connector error but still we can have observability by logging and also adding this to metrics
and as discussed yesterday can we have the connector error also covered

crates/router/src/core/unified_connector_service/kill_switch.rs
```diff
@@ -0,0 +1,437 @@
...
+    };
+
+    // Carries enough to find the originating request.
+    let record = serde_json::json!({
+        "reason": reason.as_str(),
+        "error": error.to_string(),
+        "request_id": state.request_id.as_ref().map(|id| id.to_string()),
+        "tripped_at": common_utils::date_time::now_unix_timestamp(),
+    })
+    .to_string();
```
AmitsinghTanwar007: i think this is still not counter based but its ok for now but need to be updated later

crates/hyperswitch_interfaces/src/unified_connector_service/transformers.rs
```diff
@@ -1398,3 +1398,221 @@ impl ErrorSwitch<ConnectorError> for UnifiedConnectorServiceError {
...
+            Self::HyperswitchRequestInvalid => "hyperswitch_request_invalid",
+            Self::HyperswitchResponseUndecodable => "hyperswitch_response_undecodable",
+            Self::UcsRejectedRequest => "ucs_rejected_request",
+            Self::UcsFlowUnsupported => "ucs_flow_unsupported",
+            Self::UcsInternalError => "ucs_internal_error",
+            Self::UcsUnreachable => "ucs_unreachable",
+            Self::ConnectorOutcome => "connector_outcome",
+        }
+    }
+}
```
kashif-m: nit: this can be added as a macro to convert a CamelCase to snake_case str

JeevaRamu0104 (dismissed):

crates/router/src/core/payments/helpers.rs
```diff
@@ -2501,23 +2501,20 @@ pub fn decide_payment_method_retrieval_action(
     }
 }
 
-pub async fn is_ucs_enabled(state: &SessionState, config_key: &str) -> bool {
+pub async fn is_config_flag_enabled(state: &SessionState, config_key: &str) -> bool {
```
kashif-m, line 2504: nit: check if we can move this flag to superposition

crates/hyperswitch_interfaces/src/helpers.rs
```diff
@@ -147,6 +157,9 @@ pub async fn send_comparison_data(
     connector_name: String,
     sub_flow_name: Option<String>,
     request_id: Option<String>,
+    merchant_id: Option<id_type::MerchantId>,
+    payment_method: Option<common_enums::enums::PaymentMethod>,
+    payment_method_type: Option<common_enums::enums::PaymentMethodType>,
 ) -> common_utils_errors::CustomResult<(), errors::HttpClientError> {
```
kashif-m: Pass references instead of cloned values

`Option<&Type>`

crates/hyperswitch_interfaces/src/helpers.rs
```diff
@@ -175,6 +188,27 @@ pub async fn send_comparison_data(
...
+        );
+    }
+
+    if let Some(pmt) = payment_method_type {
+        request.add_header(
+            X_PAYMENT_METHOD_TYPE,
+            hyperswitch_masking::Maskable::Normal(pmt.to_string()),
+        );
+    }
+
```
hrithikesh026, line 212: why are these needed for killswitch implementation?

crates/hyperswitch_interfaces/src/unified_connector_service/transformers.rs
```diff
@@ -1398,3 +1398,221 @@ impl ErrorSwitch<ConnectorError> for UnifiedConnectorServiceError {
...
+    UcsInternalError,
+    /// UCS could not be reached, or was too unwell to answer.
+    UcsUnreachable,
+    /// The connector rejected the request. May be a legitimate decline or a request UCS built
+    /// wrongly — indistinguishable at this layer, so we trip conservatively because falling back
+    /// to the battle-tested direct path is always safe.
+    ConnectorOutcome,
+}
+
+impl UcsKillSwitchReason {
```
hrithikesh026: you can make use if strum::Display for this. Rename to snake_case

crates/router/src/core/unified_connector_service.rs
```diff
@@ -536,20 +591,58 @@ fn decide_execution_path(
...
+pub fn build_merchant_rollout_scope(
+    merchant_id: &str,
+    connector_name: &str,
+    flow_name: &str,
+    payment_method: common_enums::PaymentMethod,
+    payment_method_type: Option<PaymentMethodType>,
+) -> String {
+    // Refund keys carry no payment method.
+    if matches!(flow_name, "Execute" | "RSync") {
+        return format!("{merchant_id}_{connector_name}_{flow_name}");
```
kashif-m: Avoid early returns

crates/router/src/core/unified_connector_service/kill_switch.rs
```diff
@@ -0,0 +1,466 @@
...
+    format!("{}_{rollout_scope}", consts::UCS_KILL_SWITCH_REDIS_PREFIX)
+}
+
+/// Whether the kill switch has tripped for this scope.
+///
+/// Fails closed: a redis error routes to the direct integration. Only reached once the rollout
+/// config resolved to primary, so shadow traffic never pays for the lookup.
+pub async fn is_tripped(state: &SessionState, rollout_scope: &str) -> bool {
+    if !is_config_flag_enabled(state, consts::UCS_KILL_SWITCH_ENABLED).await {
+        return false;
```
kashif-m: Avoid early returns in the entire module

crates/hyperswitch_interfaces/src/helpers.rs
JeevaRamu0104: earlier we are not passing the X_PAYMENT_METHOD_TYPE X_PAYMENT_METHOD for  router data diff to so we are not able to correlate the data in validation service

crates/router/src/core/payments/helpers.rs
JeevaRamu0104: we will all these config to superposition and entire shadow enablement using supoerposition

gopikrishna000: we will be using superposition soon, along with other ucs rollout config., then we will deprecate this kill switch

crates/router/src/core/unified_connector_service/kill_switch.rs
JeevaRamu0104: this is expected behaviour

crates/router/src/core/unified_connector_service.rs
JeevaRamu0104: this is expected behaviour

crates/router/src/core/unified_connector_service/kill_switch.rs
jarnura: Avoid early return here means, if should have a else block, because of return that is skipped. Standard practices to flow

crates/hyperswitch_interfaces/src/unified_connector_service/transformers.rs
gopikrishna000: Done in `6084bea3e5` — derived `strum::Display` with `#[strum(serialize_all = "snake_case")]` and dropped the hand-written `as_str` match. Labels are unchanged (`hyperswitch_request_invalid`, `ucs_unreachable`, ...).

crates/hyperswitch_interfaces/src/unified_connector_service/transformers.rs
gopikrishna000: Covered by `6084bea3e5` — `#[strum(serialize_all = "snake_case")]` now derives the labels, so the CamelCase→snake_case match is gone rather than replaced with a macro.

crates/router/src/core/unified_connector_service/kill_switch.rs
jarnura: Code should not have early jumps and error can short circuit that also can be covered by ?

crates/router/src/core/unified_connector_service/kill_switch.rs
```diff
@@ -0,0 +1,551 @@
...
+        .attach_printable("Failed to delete the UCS kill switch trip")?;
+
+    match reply {
+        redis_interface::DelReply::KeyDeleted => {
+            logger::info!(
+                rollout_scope = %rollout_scope,
+                "ucs_kill_switch: trip cleared via api"
+            );
+
+            Ok(crate::services::ApplicationResponse::StatusOk)
```
kashif-m, line 329: Avoid crate based imports

jarnura (approved):

hrithikesh026 (approved):

crates/router/src/core/unified_connector_service.rs
```diff
@@ -3093,6 +3150,34 @@ where
                 router_env::metric_attributes!(("connector", connector_name.clone(),)),
             );
 
+            // Every payment and payout gateway funnels through here with the UCS error still
+            // typed, so this is the one place the kill switch has to observe.
+            match get_flow_name::<T>() {
```
kashif-m, line 3155: Do we not need to record failures for cases where UCS returns `ConnectorError` in `Ok` branch?

kashif-m (approved):

PR #13497, opened 2026-07-29
feat(core): Align UCS connector timeout handling with Hyperswitch direct connector timeout response

Shubhodip900:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

Align UCS connector timeout handling with Hyperswitch direct connector timeout response

### Additional Changes

- [ ] This PR modifies the API contract
- [ ] This PR modifies the database schema
- [ ] This PR modifies application configuration/environment variables

## How did you test it?

Tested Via creating a mockserver, compared Direct Hyperswitch and direct Hyperswitch-prism responses. Now bot are identical.

Before this change Direct Hyperswitch Prism response:-

```json
{
    "error": {
        "type": "invalid_request",
        "message": "Payment failed during authorization with connector. Retry payment",
        "code": "CE_01",
        "data": {
            "code": "UCS_504",
            "message": "Server responded with Request Timeout",
            "status_code": 504
        }
    }
}
```

Direct Hyperswitch Response:-
```json
{
    "payment_id": "pay_BPw71s88fyIQw7QJ7dK1",
    "merchant_id": "merchant_1785324054",
    "status": "processing",
    "amount": 100,
    "net_amount": 100,
    "shipping_cost": null,
    "amount_capturable": 100,
    "amount_received": null,
    "processor_merchant_id": "merchant_1785324054",
    "initiator": null,
    "sdk_authorization": "cHJvZmlsZV9pZD1wcm9feTU2RVd3MzlQRDJyUWJxZnBmYmwscHVibGlzaGFibGVfa2V5PXBrX2Rldl9iMDcxZTBjNmMzNGI0MWJmOWIzOGYxZmVlMmZiNzg4YSxjbGllbnRfc2VjcmV0PXBheV9CUHc3MXM4OGZ5SVF3N1FKN2RLMV9zZWNyZXRfeHZVeTFwWjNVUDNUWWdmcnFDQ0wsY3VzdG9tZXJfaWQ9Y3VzX0N0TG1IOFJQd3JMZ3d3LGNsaWVudF9zZXNzaW9uX2lkPWNsaWVudF9zZXNzX0syYnBQTE9hSDFYOFM3UTVndWkxLHBheW1lbnRfaWQ9cGF5X0JQdzcxczg4ZnlJUXc3UUo3ZEsx",
    "connector": "stripe",
    "state_metadata": null,
[...]

PR #13647, opened 2026-08-12
feat(ucs): add automatic kill switch for deterministic UCS failures

### LINKED ISSUE #13645: feat(ucs): add automatic kill switch for UCS failures

gopikrishna000:
Hotfix port of #13639 onto `hotfix-2026.07.29.2`, as the merged squash commit rather than the
individual branch commits.

Four cherry-picks, in order. The first three are already on `main` and are prerequisites the kill
switch builds on; without them this base does not compile.

| commit | why it is needed here |
|---|---|
| #13500 `feat: add payment method headers to connector requests` | defines `X_PAYMENT_METHOD` / `X_PAYMENT_METHOD_TYPE`, which the comparison-service headers use |
| #13497 `feat(core): Align UCS connector timeout handling` | maps a connector timeout to a `ConnectorError` carrying the synthetic 504, so the classifier labels it as a connector outcome rather than `ucs_unreachable` |
| #13597 `fix(ucs): Add missing error mapping in HS <> UCS` | completes the tonic-to-HTTP mapping and routes UCS server failures to `ResponseHandlingFailed` |
| #13639 `feat(ucs): add automatic kill switch` | the merged squash — the kill switch itself |

Every kill-switch file is byte-identical to what merged to `main`.

## Why

Over a thousand `ucs_rollout_config` keys are provisioned in shadow. Promoting any of them to primary puts UCS in front of a connector integration that has served production for years, and the config cardinality is too high to promote one at a time and watch. This adds an automatic revert so a promotion that goes wrong is bounded without waiting for a human.

## What it does

When a UCS call fails in a way that could not have come from the connector, the affected rollout scope is written to Redis and every subsequent request for that scope is served by the direct integration instead. The scope is `merchant + connector + flow + payment_method + payment_method_type`, so a trip does not take down unrelated traffic for the same merchant.

Only primary is diverted, and it is diverted to shadow rather than to direct. The merchant is served by the direct integration either way, but shadow keeps mirroring, so a tripped scope still produces the comparison signal needed to confirm a fix before it is reset.

## Failure classification

`ucs_kill_switch_reason` in `hyperswitch_interfaces` decides what counts. Reasons are named for where the failure originated:

| Reason | Origin |
|---|---|
| `hyperswitch_request_invalid` | Hyperswitch could not build the request; UCS was never reached |
| `hyperswitch_response_undecodable` | Hyperswitch could not decode what UCS returned |
| `ucs_rejected_request` | UCS rejected what Hyperswitch sent it |
| `ucs_flow_unsupported` | UCS does not implement this flow for this connector |
| `ucs_internal_error` | UCS failed internally, or failed the flow without saying more |
| `ucs_unreachable` | UCS could not be reached, or was too unwell to answer |
| `connector_outcome` | the connector answered with an error |

Every failure trips. A connector rejection may be a legitimate decline or a request UCS built wrongly, and the two are indistinguishable at this layer, so it trips like the rest.

The bias is deliberate: a false trip costs a scope reverting to the integration it was already using, a missed trip costs merchant-visible behaviour change. False positives are acceptable, false negatives are not.

## Carve-outs

- `ConnectorIntegrationType::UcsConnector` — connectors that exist only on UCS have no direct integration to fall back to
- `CallConnectorAction::UCSConsumeResponse` — no call was made
- shadow — nothing merchant-facing to protect

## Operating it

- `UCS_KILL_SWITCH_ENABLED` config flag gates enforcement. Off, the classifier still runs and still emits metrics, so it can be validated against real traffic before it is turned on.
- `GET /unified-connector-service/kill-switch/{scope}` reports whether a scope is tripped and what tripped it; `DELETE` on the same path clears it. Both admin-authenticated, both a single redis read on a known key. `{scope}` takes either the `rollout_scope` the trip logs or the whole redis key.
- Clearing a scope that is not tripped returns **404**, not a 200 — redis reports a missing key as a successful delete of nothing, and reporting that as success would tell an operator a scope is back on UCS when a mistyped scope left it tripped.
- Trips carry a 7 day TTL. Each qualifying failure emits one structured line, `ucs_kill_switch`, carrying `rollout_scope`, `reason`, `outcome`, merchant, connector, flow, payment id and request id. Alert on `outcome`: `tripped` means a scope left UCS, `write_failed` means one should have and did not, `switch_off` is the calibration feed while enforcement is disabled.
- Metrics: `UCS_KILL_SWITCH_FAILURE` (qualifying failures seen) and `UCS_KILL_SWITCH_TRIPPED` (scopes cut over).

## Verification

- `cargo check -p router -p hyperswitch_interfaces --features v1 --all-targets` on this base — clean
- 9 unit tests (7 router, 2 hyperswitch_interfaces), `just clippy`, `just clippy_v2`, `just clippy fred` — all clean
- Both endpoints exercised against a running server and a real redis: status reports untripped and tripped correctly, accepts the bare scope and the whole redis key alike, `DELETE` on a mistyped scope returns 404, `DELETE` on a real one returns 200 and the key is gone

## Note

`hotfix/ucs_kill_switch` on this same base carries a database-backed kill switch for the same gate. The two overlap in `should_call_unified_connector_service` and should not both land.

### DIFF
--- a/crates/common_utils/src/consts.rs
+++ b/crates/common_utils/src/consts.rs
@@ -173,6 +173,12 @@ pub const X_FLOW_NAME: &str = "x-flow";
 /// Connector name
 pub const X_CONNECTOR_NAME: &str = "x-connector";
 
+/// Payment method
+pub const X_PAYMENT_METHOD: &str = "x-payment-method";
+
+/// Payment method type
+pub const X_PAYMENT_METHOD_TYPE: &str = "x-payment-method-type";
+
 /// Sub-flow name
 pub const X_SUB_FLOW_NAME: &str = "x-sub-flow";
 
--- a/crates/hyperswitch_interfaces/src/api/gateway.rs
+++ b/crates/hyperswitch_interfaces/src/api/gateway.rs
@@ -410,6 +410,9 @@ where
             let return_raw_connector_response_clone = return_raw_connector_response;
             let context_clone = context;
             let connector_name = router_data.connector.clone();
+            let merchant_id = router_data.merchant_id.clone();
+            let payment_method = router_data.payment_method;
+            let payment_method_type = router_data.payment_method_type;
             tokio::spawn(
                 async move {
                     let gateway: Box<
@@ -442,6 +445,9 @@ where
                         connector_name,
                         direct_for_compare,
                         ucs_for_compare,
+                        Some(&merchant_id),
+                        Some(payment_method),
+                        payment_method_type,
                     )
                     .await;
                 }
@@ -544,6 +550,9 @@ where
             let return_raw_connector_response_clone = return_raw_connector_response;
             let context_clone = context;
             let connector_name = router_data.connector.clone();
+            let merchant_id = router_data.merchant_id.clone();
+            let payment_method = router_data.payment_method;
+            let payment_method_type = router_data.payment_method_type;
             tokio::spawn(
                 async move {
                     let gateway: Box<
@@ -576,6 +585,9 @@ where
                         connector_name,
                         direct_for_compare,
                         ucs_for_compare,
+                        Some(&merchant_id),
+                        Some(payment_method),
+                        payment_method_type,
                     )
                     .await;
                 }
--- a/crates/hyperswitch_interfaces/src/api_client.rs
+++ b/crates/hyperswitch_interfaces/src/api_client.rs
@@ -3,13 +3,15 @@ use std::{
     time::{Duration, Instant},
 };
 
-use common_enums::ApiClientError;
+use common_enums::{ApiClientError, PaymentMethod, PaymentMethodType};
 #[cfg(feature = "ext_services_latency")]
 use common_utils::consts::EXTERNAL_CALL_TAG;
 use common_utils::{
-    consts::{X_CONNECTOR_NAME, X_FLOW_NAME, X_REQUEST_ID},
+    consts::{
+        X_CONNECTOR_NAME, X_FLOW_NAME, X_PAYMENT_METHOD, X_PAYMENT_METHOD_TYPE, X_REQUEST_ID,
+    },
     errors::CustomResult,
-    request::{Request, RequestContent},
+    request::{Headers, Request, RequestContent},
 };
 use error_stack::{report, ResultExt};
 use futures::FutureExt;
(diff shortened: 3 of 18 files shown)


XyneSpaces:
:warning: `kill_switch.rs:read_trip(...)` uses `unwrap_or_else(|_| true)` on an external Redis call. Make the fail-closed path explicit with a match so the safety invariant is obvious to future readers.

:warning: `kill_switch.rs:is_ucs_only_connector` returns `false` when the connector name fails to parse, silently hiding parse failures. Log the error before returning the default.

:warning: `unified_connector_service.rs:build_merchant_rollout_scope` matches flow names as `&str` literals (`"Execute"`, `"RSync"`). Use the typed `Flow` enum to avoid stringly-typed keys and silent mismatches.

:warning: `transformers.rs:_ => 500` in `tonic_to_http_status` swallows future `tonic::Code` variants. Enumerate all variants explicitly or document why a catch-all is safe.

jarnura (dismissed):

AmitsinghTanwar007 (approved):

shuklatushar226 (approved):

JeevaRamu0104 (approved):

ShankarSinghC (approved):

PR #13667, opened 2026-08-12
fix(revenue_recovery): index retry ladder by total invoice attempts

### LINKED ISSUE #13665: Index revenue recovery retry schedules by total invoice attempts
## Bug Description

The revenue recovery calculate workflow seeds its retry ladder with a fixed value instead of the total number of attempts already made on the invoice. Billing-connector retries can therefore be replayed by Hyperswitch, and an exhausted ladder can remain scheduled indefinitely.

## Expected Behavior

- Index the retry ladder using the invoice's total attempt count.
- Preserve that count when reopening the calculate workflow after a failed attempt.
- Finish the calculate process with `RETRIES_EXCEEDED` when no retry slot remains.

## Actual Behavior

The calculate tracker starts with a hard-coded retry count, reopening derives the count from the execute tracker, and an empty retry schedule is treated like a temporary absence of work.

## Steps To Reproduce

1. Configure a retry ladder and allow the billing connector to consume one or more attempts.
2. Hand the invoice to Hyperswitch revenue recovery.
3. Observe that scheduling starts from the fixed tracker value instead of the invoice total.
4. Exhaust the configured ladder and observe that the calculate job can remain alive without scheduling another payment.

## Context For The Bug

Billing and Hyperswitch retries share one invoice-level attempt budget. Scheduling must account for attempts already consumed and terminate deterministically when the ladder ends.

## Environment

Hyperswitch revenue recovery v2 flow on the current `main` branch.

## Checks

- [x] I checked and didn't find a similar issue.
- [x] I have read the contributing guidelines.
- [x] I am willing to submit a PR.

Implemented by #13667.

NISHANTH1221:
## Type of Change

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

Aligns the revenue recovery retry ladder with the total attempts already made on an invoice.

- Seeds the calculate process tracker from `intent_retry_count` rather than a fixed value.
- Reopens the calculate workflow using the payment intent's total recovery retry count.
- Adds a distinct `RetriesExhausted` token-selection result when the configured schedule has no remaining slot.
- Finishes the calculate workflow with `RETRIES_EXCEEDED` instead of rescheduling indefinitely.
- Adds regression coverage for a ladder shared between billing-connector and Hyperswitch attempts.

### Stack Dependency

Depends on [#13719](https://github.com/juspay/hyperswitch/pull/13719).

### Additional Changes

## Motivation and Context

The billing connector and Hyperswitch consume attempts from the same invoice-level retry budget. Starting the calculate tracker from a fixed value replays ladder positions already consumed by the billing connector. Separately, treating an exhausted schedule as a temporary lack of work keeps a process alive even though it can never schedule another payment.

Using the invoice total as the ladder index preserves the intended retry schedule across the handoff and makes exhaustion terminal.

## How did you test it?

- `cargo +nightly fmt --all -- --check`
- `cargo check --package router --no-default-features --features release,v2,redis-rs`
- `git diff --check origin/main..HEAD`

The focused scheduler test is included in this PR. Running the scheduler package standalone is currently blocked on the latest `main` by existing feature-propagation compilation failures before the test binary is built.

## Checklist

## Related Issue

Closes #13665


abhisheksharma2411 (commented):
Nice change — indexing the ladder by invoice total rather than by our own tracker's count is clearly the right call, and adding a distinct `RetriesExhausted` instead of letting the calculate job reschedule forever is a good catch. The `walk_pcr_ladder` test is a genuinely useful thing to have; ladder arithmetic is exactly the kind of thing that drifts silently.

Two observations, neither blocking.

**1. The fallback quietly changes what the number means.**

```rust
.map(|recovery_metadata| i32::from(recovery_metadata.total_retry_count))
.unwrap_or(process.retry_count + 1);
```

The primary path yields an **absolute total across the invoice**; the fallback yields **our local tracker's count, incremented**. Those aren't the same quantity, and the comment right above says why the second one is wrong — the EXECUTE tracker only counts our own attempts.

So if `payment_revenue_recovery_metadata` is ever absent on an invoice that has genuinely been retried, this silently reinstates the bug the PR is fixing, and the ladder restarts from our local index. Nothing would signal it.

Might be worth either logging when the fallback fires, or treating missing metadata as an error at this point — if we've reached `reopen_calculate_workflow_on_payment_failure` the invoice has been through a recovery attempt, so is metadata actually optional here, or is it optional only in the type?

**2. Small mismatch between the test's comment and its assertion.**

```rust
/// 15 retries are budgeted on top of the initial charge...
assert_eq!(attempts, 15, "ladder must cover the initial charge plus 14 retries");
```

The doc says 15 retries *plus* the charge (16 total), the assertion says 15 total (charge + 14). The assertion matches the mapping — `start_after` + 4 + 10 = 15 attempts, landing at 1 + 11 + 16 = day 28 — so the assertion is right and the comment is one out.

Trivial on its own, but in a PR specifically about off-by-one attempt counting, the comment is the thing a future reader will trust.

**One question rather than a finding:** `reopen_calculate_workflow_on_payment_failure` reads the total from intent metadata and then checks whether a tracker entry already exists. If two failures for the same invoice land close together, can both read the same `total_retry_count` and both get past the existence check, or is that insert guarded somewhere I haven't read? Not asserting there's a race — I only have the diff — just that it's the shape where one usually hides.

Rest looks good to me 🙂

abhisheksharma2411:
Following up on my 12 Aug notes — apologies for the gap, that is on me. Three commits landed on 14 Aug and I did not come back to look.

Two things, one mechanical and one still open.

**The branch is conflicted.** `mergeable_state` is `dirty`, so it cannot merge as it stands regardless of review.

**The comment/assertion mismatch is still there, and it now carries more weight than when I first mentioned it.** In `walk_pcr_ladder`:

```rust
/// 15 retries are budgeted on top of the initial charge. The billing connector takes the first
/// and holds the last back for day 30, so the ladder covers the 14 in between and then stops.
```

The mapping is `start_after` plus `1+1+1+1+10 = 14` frequency slots, and by the helper's own doc that is **the charge plus 14 retries = 15 attempts** — which is exactly what the first assertion says:

```rust
assert_eq!(attempts, 15, "ladder must cover the initial charge plus 14 retries");
```

So "15 retries on top of the initial charge" is one out; it should read 14. The sentence is also internally inconsistent — if 15 were budgeted and the connector took the first and held the last, the in-between would be 13, not 14.

The arithmetic in the code is right, and the new `assert_eq!(attempts, 13, ...)` for `attempts_already_made = 2` follows correctly from it (14 - 1 = 13). It is only the prose that is off. I would normally not push on a comment, but this is a PR about off-by-one attempt counting, and that doc line is what the next reader will trust over the mapping.

**And the fallback is unchanged:**

```rust
.map(|recovery_metadata| i32::from(recovery_metadata.total_retry_count))
.unwrap_or(process.retry_count + 1);
```

Still two different quantities — an absolute invoice total on the primary path, our local tracker's count on the fallback. Entirely reasonable if you concluded metadata is genuinely always present there; I just could not tell from the diff, and if it can be absent this silently reinstates the ladder restart the PR exists to fix, with nothing to signal it. A log line when it fires would be enough.

Happy to re-read properly once it is off conflict — ping me and I will turn it around quickly this time.

hrithikesh026 (dismissed):

rohanm13 (dismissed):

rohanm13 (approved):

hrithikesh026 (approved):

abhisheksharma2411:
Came back as promised now that the conflict is cleared — `mergeable` is clean and it's approved, so this is just a last read rather than a request to hold anything up.

The arithmetic is right and the new `13` assertion follows correctly. Both small things from my last pass are still open on `71680b8`, so flagging once more and then leaving it with you:

**The doc line above the test.** `crates/scheduler/src/utils.rs:523`:

```rust
/// 15 retries are budgeted on top of the initial charge. The billing connector takes the first
/// and holds the last back for day 30, so the ladder covers the 14 in between and then stops.
```

The frequency slots are `1+1+1+1+10 = 14`, and the assertion three lines down reads:

```rust
assert_eq!(attempts, 15, "ladder must cover the initial charge plus 14 retries")
```

15 *attempts* = charge + 14 retries. "15 retries on top of the initial charge" would be 16. The second sentence doesn't survive it either — 15 budgeted, minus the connector's first and the day-30 last, leaves 13 in between, not the 14 it claims. Swapping the one digit makes the whole line agree with both the mapping and the assertion. Only pushing on prose because this is a PR about off-by-one attempt counting, and that comment is what the next reader will trust over the slot arithmetic.

**The fallback**, `crates/router/src/core/revenue_recovery/types.rs:1486`, unchanged:

```rust
.unwrap_or(process.retry_count + 1)
```

Still an absolute invoice total on the primary path and our local tracker's count on the fallback. Fine if metadata is genuinely always present there — you're better placed to know than I am from the diff. If it can be absent, this quietly restarts the ladder that the PR exists to stop, with nothing in the logs to say it happened. A single `log` line where it fires would make that visible without changing behaviour.

Neither is a correctness bug in the shipped path, and I'm not a maintainer here — merge over both if you'd rather not respin an approved branch. Sorry again for the slow first loop on this one.

PR #13557, opened 2026-08-05
feat(connector): add support for santander pix payouts

aniketburman014:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

Integrates Santander Brazil as a payout connector supporting Pix payments (key-based, QR/EMV, and bank account transfers) via the management payments partners API with mTLS.

### Additional Changes

- [ ] This PR modifies the API contract
- [ ] This PR modifies the database schema
- [ ] This PR modifies application configuration/environment variables

## Motivation and Context

UCS pr - https://github.com/juspay/hyperswitch-prism/pull/2061
### Connector create

```
'{
  "connector_type": "payout_processor",
  "connector_name": "santander",
  "profile_id": "pro_lcr5iL63I2S4p2GmRmPQ",
  "connector_account_details": {
    "auth_type": "CertificateAuth",
    
    
    "certificate": "{{CERTIFICATES_BASE64}}",
    "private_key": "{{CERTIFICATE_KEY_BASE64}}"
  },
  "metadata": {
    "pix_payout": {
      "workspace_id": "{{WORKSPACE_ID}}",
      "client_id": "{{CLIENT_ID}}",
      "client_secret": "{{CLIENT_SECRET}}"
    }
  },
  "payment_methods_enabled": [
    {
      "payment_method": "bank_transfer",
      "payment_method_types": [
        {
          "payment_method_type": "pix",
          "minimum_amount": 1,
          "maximum_amount": 68607706,
          "recurring_enabled": false,
          "installment_payment_enabled": false
        },
        {
          "payment_method_type": "pix_key",
          "minimum_amount": 1,
          "maximum_amount": 68607706,
          "recurring_enabled": false,
          "installment_payment_enabled": false
        },
        {
          "payment_method_type": "pix_emv",
          "minimum_amount": 1,
          "maximum_amount": 68607706,
          "recurring_enabled": false,
          "installment_payment_enabled": false
        }
      ]
    }
  ]
}'
```
### Request
```
{
  "amount": 1500,
  "currency": "BRL",
  "connector": ["santander"],
  "customer_id": "cus_bw5T9ISg3KE6XP0pripr",
  "payout_type": "bank",
     "payout_method_data": {
      "bank_transfer": {
        "payout_method_type": "pix",
        "bank_branch": "0001",
        "bank_account_number": "20294925",
[...]

PR #13671, opened 2026-08-12
fix(ucs): suppress X-Connector-Config header for Netcetera

ShankarSinghC:
## Summary
- `connector-service`'s `ConnectorSpecificConfig` proto oneof has never had a `netcetera` case (confirmed across `main` and every relevant feature/fix branch).
- Since #13557, router unconditionally builds and sends the typed `x-connector-config` header for `Connector::Netcetera` (`ConnectorAuthType::CertificateAuth`), which UCS now hard-fails to deserialize (`unknown variant \`Netcetera\`, expected one of \`Adyen\`, ...`) instead of falling back to the legacy `x-connector-auth` header.
- Confirmed via integ logs (build `2026.08.12.0-dirty-eefdfe9b7`) and reproduced by diffing against sandbox's custom branch (`sbx-custom-cug-hyperswitch-server`, commit `3ae23cd`), which never picked up the #13557 change and therefore still suppresses the header for Netcetera, hitting UCS's legacy-header fallback path successfully.
- `Connector::Netcetera`'s own doc comment already says "no connector-specific config needed" — so the correct fix is for `build_connector_config_header` to return `Ok(None)` for Netcetera, restoring the legacy-header path that sandbox proves works, rather than touching connector-service's proto.

## Test plan
- [ ] `cargo check -p router` / `just clippy` (v1 + v2)
- [ ] Manual: run a Netcetera external-3ds-over-VGS payment against integ-matching UCS build and confirm PreAuthenticate/Authenticate no longer error with `unknown variant \`Netcetera\`` and instead log `Typed connector config headers not found, falling back to legacy headers`

### DIFF
--- a/crates/router/src/core/unified_connector_service/connector_config.rs
+++ b/crates/router/src/core/unified_connector_service/connector_config.rs
@@ -1762,6 +1762,14 @@ pub fn build_connector_config_header(
         merchant_account_metadata,
     ))?;
 
+    // Netcetera has no connector-specific config on the wire (connector-service's
+    // `ConnectorSpecificConfig` oneof has no `netcetera` case), so sending this header makes
+    // UCS hard-error on deserialization instead of falling back to the legacy auth header.
+    // Suppress it here so UCS takes the legacy-header path, which does work for Netcetera.
+    if matches!(config, ConnectorSpecificConfig::Netcetera) {
+        return Ok(None);
+    }
+
     let config_json = serde_json::to_value(&config)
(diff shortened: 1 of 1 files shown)


shuklatushar226 (approved):

AmitsinghTanwar007 (approved):

ShankarSinghC:
closing the pr as the changes are present in the https://github.com/juspay/hyperswitch/pull/13673

PR #13673, opened 2026-08-12
fix(ucs): correct Netcetera auth-header injection and suppress its X-Connector-Config

ShankarSinghC:
## Summary
- Netcetera is UCS's one authentication-only, no-key connector: its mTLS cert is presented on the VGS outbound route, not via UCS auth headers, and connector-service explicitly opts it out of the `CertificateAuth`->`ConnectorSpecificConfig` conversion (`ConnectorEnum::Netcetera => Err(...)` in connector-service). Sending it as `multi-auth-key` (correct for Santander, which does need real cert/key transmitted) makes UCS reject pre-authenticate with `Failed to convert legacy auth for connector: netcetera`. Special-cased Netcetera to send `x-auth: no-key` instead.
- UCS's `multi-auth-key` header parsing unconditionally requires `x-key2`/`x-api-secret` to be present, but `CertificateAuth` only carries two real values (`certificate`, `private_key`). Duplicated `private_key` into both to satisfy UCS's presence check for connectors that do need the multi-auth-key path (e.g. Santander).
- Suppressed the `X-Connector-Config` header for Netcetera: UCS's `ConnectorSpecificConfig` oneof has no `Netcetera` case, so sending it makes UCS hard-fail deserialization instead of falling back to the legacy auth-header path, which works once the above is fixed.

Verified end-to-end locally: Netcetera pre-authenticate -> AReq/challenge -> Checkout authorize with CAVV now succeeds (`payment_intent.status = succeeded`, `authentication.status = success`, `trans_status = Success`).

## Test plan
- [x] `cargo check -p router --features v1`
- [x] Local end-to-end payment through Netcetera external 3DS + VGS external vault + Checkout succeeded


shuklatushar226 (approved):

AmitsinghTanwar007 (approved):

crates/router/src/core/unified_connector_service.rs
```diff
@@ -2061,17 +2064,47 @@ pub fn build_unified_connector_service_auth_metadata(
...
-            api_key: Some(certificate.clone()),
-            key1: Some(private_key.clone()),
-            key2: None,
-            api_secret: None,
-            auth_key_map: None,
-            merchant_id: Secret::new(merchant_id.to_string()),
-            connector_config,
-        }),
+        } => {
+            if matches!(connector, Connector::Netcetera) {
```
AmitsinghTanwar007, line 2068: Nit: if you have connector enum then can you have match here

crates/router/src/core/unified_connector_service/connector_config.rs
```diff
@@ -1757,20 +1748,24 @@ impl ForeignTryFrom<(Connector, &ConnectorAuthType, Option<&serde_json::Value>)>
...
         merchant_account_metadata,
     ))?;
 
+    // Netcetera has no connector-specific config on the wire (connector-service's
+    // `ConnectorSpecificConfig` oneof has no `netcetera` case), so sending this header makes
+    // UCS hard-error on deserialization instead of falling back to the legacy auth header.
+    // Suppress it here so UCS takes the legacy-header path, which does work for Netcetera.
+    if matches!(config, ConnectorSpecificConfig::Netcetera) {
+        return Ok(None);
+    }
```
AmitsinghTanwar007, line 1767: here also same thing you can have match

PR #13674, opened 2026-08-12
fix(ucs): add missing CertificateAuth arm for header injection, no-key for Netcetera

ShankarSinghC:
## Summary
- This hotfix branch was cut from a point that predates `CertificateAuth` support in `build_unified_connector_service_auth_metadata` (added to `main` later via a separate PR), so Netcetera's `CertificateAuth` MCA always fell into the catch-all `_ => Err(FailedToObtainAuthType)`, failing pre-authenticate before ever reaching UCS ("Unsupported ConnectorAuthType for header injection").
- Netcetera is UCS's one authentication-only, no-key connector: its mTLS cert is presented on the VGS outbound route, not via UCS auth headers, and connector-service explicitly opts it out of the `CertificateAuth`->`ConnectorSpecificConfig` conversion (`ConnectorEnum::Netcetera => Err(...)` in connector-service). Routed it via the `x-auth: no-key` shortcut instead, matching connector-service's documented design.
- The generic `multi-auth-key` path is kept for any other `CertificateAuth` connector, with `key2`/`api_secret` duplicated from `private_key` to satisfy UCS's presence check on those headers.

This builds on top of `aa957b1c05` (X-Connector-Config suppression for Netcetera) already on this branch — both are needed together for Netcetera's pre-authenticate call to succeed.

## Test plan
- [x] `cargo check -p router --features v1`
- [x] Local end-to-end payment through Netcetera external 3DS + VGS external vault + Checkout succeeded

### DIFF
--- a/crates/router/src/core/unified_connector_service.rs
+++ b/crates/router/src/core/unified_connector_service.rs
@@ -2033,6 +2033,44 @@ pub fn build_unified_connector_service_auth_metadata(
             merchant_id: Secret::new(merchant_id.to_string()),
             connector_config,
         }),
+        // Netcetera is UCS's one authentication-only, no-key connector: its mTLS cert is
+        // presented on the VGS outbound route, not via UCS auth headers, and connector-service
+        // explicitly opts it out of the CertificateAuth->ConnectorSpecificConfig conversion
+        // (see `ConnectorEnum::Netcetera => Err(...)` in connector-service's router_data.rs),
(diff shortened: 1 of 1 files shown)


ShankarSinghC:
Consolidated into #13672 — cherry-picked this branch's commit directly onto `hotfix/netcetera-connector-config-header-2026.07.29.2` instead of stacking a separate PR on top.

PR #13683, opened 2026-08-13
fix(ucs): fire kill switch on connector errors in Ok(RouterData)

### LINKED ISSUE #13721: fix(ucs): fire kill switch on connector errors wrapped in Ok(RouterData)
## Description

When UCS returns a `ConnectorError`, gateway handlers catch it and set `router_data.response = Err(ErrorResponse)`, returning `Ok(router_data)`. The kill switch currently only fires in the `Err` arm of the result, which means the `connector_outcome` reason becomes dead code — connector errors wrapped inside `Ok(RouterData)` silently bypass the kill switch.

## Expected Behavior

The kill switch should detect `Err(ErrorResponse)` inside the `Ok(RouterData)` response and call `kill_switch::record_failure`, so that connector errors are properly tracked regardless of how they are wrapped.

## Changes Required

- Add a check in the `Ok` arm of both `ucs_logging_wrapper` and `ucs_logging_wrapper_granular` to detect `Err(ErrorResponse)` inside the response and call `kill_switch::record_failure`
- Add `From<&ErrorResponse> for ConnectorErrorInner` conversion to reconstruct the error for kill switch classification
- Update kill switch threshold default value for better error handling

## Impact

Without this fix, connector errors routed through UCS are not counted toward the kill switch threshold, potentially allowing a degraded connector to continue receiving traffic.

AmitsinghTanwar007:
## Summary

Fixes #13721

- When UCS returns a `ConnectorError`, gateway handlers catch it and set `router_data.response = Err(ErrorResponse)`, returning `Ok(router_data)`. The kill switch only fired in the `Err` arm, making the `connector_outcome` reason dead code.
- Adds a check in the Ok arm of both `ucs_logging_wrapper` and `ucs_logging_wrapper_granular` to detect `Err(ErrorResponse)` inside the response and call `kill_switch::record_failure`.
- Adds `From<&ErrorResponse> for ConnectorErrorInner` conversion to reconstruct the error for kill switch classification.

## Test plan
- [ ] Verify successful UCS payment does not trigger kill switch
- [ ] Verify connector decline (e.g. card declined) via UCS triggers kill switch with `connector_outcome` reason
- [ ] Verify refund decline via UCS triggers kill switch (goes through `ucs_logging_wrapper`)
- [ ] Verify shadow mode does not trigger kill switch (early return on non-Primary execution mode)
- [ ] Verify kill switch trip is visible via admin API (`/ucs/kill_switch/tripped`)


crates/router/src/core/payments/helpers.rs
```diff
@@ -2566,6 +2578,8 @@ impl Default for RolloutExecutionResult {
             should_execute: false,
             proxy_override: None,
             execution_mode: ExecutionMode::NotApplicable,
+            kill_switch_enabled: true,
+            kill_switch_threshold: 0,
```
gopikrishna000: Should we have default kill switch disabled 
Anyone in integ sbx testing with older primary key don't expect it to kill switch 
That too if both kill switch enabled and threshold zero as default, then all older config without these fields will immediately fall back to direct even without error right

crates/router/src/core/unified_connector_service/kill_switch.rs
```diff
@@ -44,49 +46,66 @@ fn rollout_scope_in(key_or_scope: &str) -> &str {
...
+pub async fn is_kill_switched(
+    state: &SessionState,
+    rollout_scope: &str,
+    kill_switch_enabled: bool,
+    kill_switch_threshold: u64,
+) -> bool {
+    if kill_switch_enabled {
+        match read_counter(state, rollout_scope).await {
+            Ok(count) => {
+                let exceeded = count > kill_switch_threshold;
```
gopikrishna000: should it be >= ?
like if we have threadhold 0., it mean immediate kill switch. but it will be consider as exceeded only when it is 1 as per this code?

crates/router/src/core/unified_connector_service/kill_switch.rs
```diff
@@ -44,49 +46,66 @@ fn rollout_scope_in(key_or_scope: &str) -> &str {
...
-    if tripped {
-        // Every request for a tripped scope reaches here; the trip is logged once elsewhere.
-        logger::debug!(
-            rollout_scope = %rollout_scope,
-            "ucs_kill_switch: scope is tripped, routing to the direct integration"
-        );
-    }
+        .get_hash_field(&key, COUNTER_FIELD)
+        .await
+        .unwrap_or(0);
```
gopikrishna000, line 112: why unwrap or 0?
caller function has handling for RedisError right

crates/router/src/core/unified_connector_service/kill_switch.rs
AmitsinghTanwar007: the statement says exceeded if count = 0 and threshold = 0 then as 0 ! greater than 0 we will make it false

crates/router/src/core/unified_connector_service/kill_switch.rs
gopikrishna000: `get_hash_field` is generic over `V`, and `V = u64` here, so this swallows two different things:

- nil (key/field absent) → `u64::from_value(Nil)` fails → `Err` → 0. Intended.
- Redis timeout / connection reset → `Err` → 0. Not intended.

Both arrive as `RedisError::GetHashFieldFailed` — the transport error and the parse both `change_context` to it in `fred/commands.rs:1333` — so they can't be told apart downstream.

Net effect: the `Err(error) => true` arm in `is_kill_switched` is unreachable for HGET failures; only `get_redis_conn()?` can reach it. During a Redis outage every scope reads 0, `0 > threshold` is false, and all traffic stays on UCS — the opposite of the "fails closed" the doc comment promises, at the moment it matters most.

Typing the get is enough:

```rust
let count: u64 = conn
    .get_hash_field::<Option<u64>>(&key, COUNTER_FIELD)
    .await?
    .unwrap_or(0);
```

nil → `None` → 0, real errors propagate. `?` compiles as-is since `storage_impl::errors::RedisError` re-exports `redis_interface::errors::RedisError`. Same shape as `is_de_diff_threshold_exceeded` in #13663.

crates/router/src/core/unified_connector_service/kill_switch.rs
```diff
@@ -314,56 +309,87 @@ pub async fn reset(state: SessionState, listed_key: String) -> errors::RouterRes
...
         .get_redis_conn()
         .change_context(errors::ApiErrorResponse::InternalServerError)
-        .attach_printable("Failed to get a redis connection to read the UCS kill switch trip")?
-        .get_key::<Option<String>>(&trip_key(rollout_scope).as_str().into())
+        .attach_printable("Failed to get a redis connection to read the UCS kill switch counter")?
+        .get_hash_field(&key, COUNTER_FIELD)
         .await
-        .change_context(errors::ApiErrorResponse::InternalServerError)
-        .attach_printable("Failed to read the UCS kill switch trip")?;
+        .unwrap_or(0);
```
gopikrishna000: Second site with the same `.unwrap_or(0)` — see the thread on `read_counter`.

Here the consequence is different: during a Redis outage this endpoint reports `counter: 0`, telling an on-call engineer the scope is clean at exactly the moment the answer is actually unknown. For a diagnostic endpoint a 500 is more useful than a confident zero.

```rust
.get_hash_field::<Option<u64>>(&key, COUNTER_FIELD)
.await
.change_context(errors::ApiErrorResponse::InternalServerError)?
.unwrap_or(0);
```

crates/router/src/core/unified_connector_service/kill_switch.rs
gopikrishna000: Correction to the file reference above: `redis_interface` defaults to the redis-rs backend (`Cargo.toml`: `default = ["redis-rs"]`), so the relevant path is `module/redis_rs/commands.rs:1209`, not the fred one. Behaviour is identical on both, so the point stands either way:

- **redis-rs** — `Value::Nil` falls to the `_ =>` arm: `invalid_type_error!("Response type not convertible to numeric.")` (`redis-1.2.0/src/types.rs:1719`)
- **fred 8.0.6** — `RedisValue::Null => Err(NotFound, "Cannot convert nil to number.")`. The `default-nil-types` feature that would turn this into `Ok(0)` is not enabled anywhere in the workspace.

And `Option<T>` maps nil to `Ok(None)` on both (`redis-1.2.0/src/types.rs:2366`, fred `response.rs:319`), so `get_hash_field::<Option<u64>>` is the fix regardless of which backend is compiled.

crates/router/src/core/unified_connector_service/kill_switch.rs
```diff
@@ -314,56 +309,90 @@ pub async fn reset(state: SessionState, listed_key: String) -> errors::RouterRes
...
+                        config_key = %config_key,
+                        "ucs_kill_switch: failed to parse RolloutConfig for threshold lookup"
+                    );
+                })
+                .ok()
+        })
+        .map(|rc| rc.kill_switch_threshold);
+
+    let tripped = match threshold {
+        Some(t) => counter_value > t,
```
gopikrishna000: should it be `>=`

gopikrishna000: can it be insdie function in killswitch.rs and used by actual call and here

crates/router/src/core/unified_connector_service/kill_switch.rs
```diff
@@ -314,56 +315,90 @@ pub async fn reset(state: SessionState, listed_key: String) -> errors::RouterRes
...
+/// Reads the counter so an on-call engineer can inspect it without Redis or log access.
 pub async fn trip_status(
     state: SessionState,
     key_or_scope: String,
 ) -> errors::RouterResponse<KillSwitchStatusResponse> {
     let rollout_scope = rollout_scope_in(&key_or_scope);
+    let key: redis_interface::RedisKey = counter_key(rollout_scope).as_str().into();
 
-    let record = state
+    let counter_value: u64 = state
```
gopikrishna000: should this use exsiting read counter function and also update that fucntion with this fix

JeevaRamu0104 (dismissed):

sai-harsha-vardhan (dismissed):

gopikrishna000 (approved):

JeevaRamu0104 (approved):

sai-harsha-vardhan (approved):

PR #13709, opened 2026-08-14
fix(ucs): surface UCS client errors as proper 4xx API responses

### LINKED ISSUE #13782: fix(ucs): surface UCS client errors as proper 4xx API responses instead of generic CE_01/HE_00
## Description

When UCS returns 4xx client errors (InvalidArgument, FailedPrecondition, etc.), the current error handling funnels them through the `ConnectorError` path, which loses the original error message and status code — returning generic `CE_01` or `HE_00` errors to the API consumer.

## Expected Behavior

UCS client errors should be surfaced as proper 4xx API responses with the actual error message and status code preserved.

## Current Behavior

All UCS errors are converted through the lossy `ConnectorError` → `ErrorSwitch` path, resulting in:
- `CE_01` (ProcessingFailed) or `HE_00` (InternalServerError) for what should be validation/client errors
- Original error details (message, status code) are discarded

## Fix

- Intercept 4xx UCS errors in `ucs_logging_wrapper_granular` and convert to `Ok(RouterData)` with `Err(ErrorResponse)`
- Fix `ErrorSwitch<ConnectorError>` to serialize error details into `ProcessingStepFailed(Some(bytes))` instead of discarding them
- Exclude `MissingRequiredField`/`MissingRequiredFields` from interception (they already map correctly via `ErrorSwitch` to `IR_04`/`IR_21`)

AmitsinghTanwar007:
Closes #13782

### Problem
When UCS (Unified Connector Service / Prism) returns errors, the previous implementation funnelled everything through a lossy `ConnectorError` path that turned all errors into generic `CE_01` or `HE_00` — losing the original error message, status code, and error semantics.

### What this PR does
- Intercept 4xx UCS errors (InvalidArgument, FailedPrecondition, etc.) in `ucs_logging_wrapper_granular` and convert to `Ok(RouterData)` with `Err(ErrorResponse)` — preserving error message and status code
- Fix `ErrorSwitch<ConnectorError>` for the `ConnectorError` variant to serialize error details into `ProcessingStepFailed(Some(bytes))` instead of discarding them to bare `ResponseHandlingFailed`

### Why typed `UnifiedConnectorServiceError` variants are needed

UCS `IntegrationError` variants (e.g., `MissingRequiredField`, `CurrencyNotSupported`, `InvalidWallet`) are connector-related errors that happen during request transformation inside UCS — for example, UCS says "Missing required field: card.card_number" because the **connector** (e.g., Cybersource) requires that field.

These typed variants ensure UCS produces the **same error codes** as Hyperswitch's direct (non-UCS) connector path for the same class of errors:

| Error type | Typed variant path | Without typed variants (TonicStatus catch-all) |
|---|---|---|
| `MissingRequiredField` | **`IR_04`** (400) — "Missing required param: card.card_number" | `CE_01` (400) — generic "Payment failed during authorization" |
| `MissingRequiredFields` | **`IR_21`** (400) — specific fields listed | `CE_01` (400) — generic |
| `InvalidWallet` / wallet errors | **`IR_07`** (400) — "Invalid value: payment_method_data" | `CE_01` (400) — generic |
| `CurrencyNotSupported` / `FlowNotSupported` | **`IR_00`** (501) — "Not implemented" | `CE_01` (400) — wrong HTTP status |
| `CaptureMethodNotSupported` | **`IR_00`** (501) | `CE_01` (400) — wrong HTTP status |

Without these typed variants, all UCS validation errors become a generic `CE_01` and merchants can't tell if they're missing a field, using an unsupported currency, or sending invalid wallet data.

### Forward compatibility
New `IntegrationError` variants added in UCS are handled via `#[serde(other)]` → `UcsErrorCode::Unknown` → falls through to the `TonicStatus` catch-all, which still preserves the gRPC code and error message in the API response (as `CE_01` with the message in the `data` field). No Hyperswitch changes needed for new variants to surface meaningful errors — typed variants can be added later for specific error code parity.

## Test plan
- [ ] Send a payment through UCS that triggers a validation error (e.g., missing billing email for Cybersource) — should return `4xx` with the actual error message instead of `CE_01`
- [ ] Send a payment through UCS that triggers a connector error (card declined) — should still return proper connector error details
- [ ] Send a payment through UCS that triggers a server error (UCS unavailable) — should still return `500 InternalServerError`
- [ ] Verify `MissingRequiredField` errors still return `IR_04` with field name


JeevaRamu0104 (dismissed):

crates/router/src/core/unified_connector_service/kill_switch.rs
```diff
@@ -344,6 +344,7 @@ pub async fn trip_status(
     key_or_scope: String,
 ) -> errors::RouterResponse<KillSwitchStatusResponse> {
     let rollout_scope = rollout_scope_in(&key_or_scope);
+    let key: redis_interface::RedisKey = counter_key(rollout_scope).as_str().into();
```
gopikrishna000: where is this key used

crates/hyperswitch_interfaces/src/unified_connector_service/transformers.rs
```diff
@@ -1152,30 +1253,50 @@ impl UnifiedConnectorServiceError {
...
-        }
-
-        Self::TonicStatus {
-            code: status.code(),
-            message: status.message().to_string(),
-        }
+        Self::decode_connector_error_response(status, connector_name)
+            .or_else(|| Self::decode_connector_timeout(status, connector_name))
+            .or_else(|| Self::decode_integration_error(status))
+            .or_else(|| Self::decode_from_plain_text_message(status))
```
gopikrishna000: decode_from_plain_text_message is for?

crates/router/src/core/unified_connector_service/kill_switch.rs
AmitsinghTanwar007: Good catch — this `key` was unused dead code. `read_counter()` constructs its own key internally. Removed it.

crates/hyperswitch_interfaces/src/unified_connector_service/transformers.rs
AmitsinghTanwar007: This is a plain-text fallback for version skew / proto decode failures. In normal operation, UCS always sends structured proto details (`IntegrationError` or `ConnectorError` via `Status::with_details`), so steps 1–3 handle everything.

`decode_from_plain_text_message` fires only when:
1. **Version skew** — an older UCS sends `IntegrationError` as plain-text message only (without proto details), and this recognizes known `#[error(...)]` patterns (e.g., `"Missing required field: card.card_number"`)
2. **Proto decode failure** — if `prost::decode` in step 3 fails due to proto schema mismatch between UCS and Hyperswitch versions

In a matched-version deployment, this should never fire — it's a resilience layer so error semantics aren't lost during rolling upgrades.

crates/hyperswitch_interfaces/src/unified_connector_service/transformers.rs
```diff
@@ -1272,6 +1391,226 @@ impl UnifiedConnectorServiceError {
...
+const UCS_PREFIX_CURRENCY_NOT_SUPPORTED: &str = "The given currency is not configured";
+const UCS_PREFIX_NOT_SUPPORTED: &str = " is not supported by ";
+const UCS_PREFIX_FLOW_NOT_SUPPORTED: &str = " flow not supported by ";
+const UCS_PREFIX_NOT_IMPLEMENTED: &str = "This feature is not implemented: ";
+const UCS_PREFIX_INVALID_CONNECTOR_CONFIG: &str = "Invalid connector configuration: ";
+const UCS_PREFIX_INVALID_WALLET_TOKEN: &str = "Failed to parse ";
+const UCS_PREFIX_MISSING_RELATED_TXN_ID: &str = "Missing connector related transaction ID: ";
+const UCS_PREFIX_FIELD_TOO_LONG: &str = "Field '";
+const UCS_PREFIX_MANDATE_MISMATCH: &str = "Field ";
+
```
AmitsinghTanwar007: we dont need to have exact same structure of hs 
ucs can have its own way to represent the error

JeevaRamu0104 (dismissed):

gopikrishna000 (dismissed):

hrithikesh026 (dismissed):

JeevaRamu0104 (dismissed):

maverox (dismissed):

Sakilmostak (dismissed):
payout looks good

Vani-1107 (dismissed):

jagan-jaya (dismissed):

JeevaRamu0104 (approved):

Sakilmostak (approved):

Vani-1107 (approved):

hrithikesh026 (approved):

gopikrishna000 (approved):

maverox (approved):

jagan-jaya (approved):

PR #13724, opened 2026-08-16
fix(hyperswitch_domain_models): include order_tax_amount in v2 net am…

### LINKED ISSUE #13712: [BUG] V2 net_amount omits order_tax_amount
### Bug Description

In V2, `AmountDetails::calculate_net_amount()` omits `order_tax_amount`. The request model accepts the tax. Request conversion stores it in the default tax field. The domain calculation adds only the order amount, shipping cost, surcharge, and tax on surcharge.

The helper is also used by V2 attempt creation, intent responses, payment method amount filtering, session routing, and split payment allocation. This means the inconsistency is not limited to one response field.

The V1 equivalent still includes the tax, because `NetAmount::get_total_amount()` in `crates/hyperswitch_domain_models/src/payments/payment_attempt.rs` sums `order_tax_amount` alongside shipping cost, surcharge, and tax on surcharge. The V2 attempt also carries `net_amount` and `order_tax_amount` in the same shape.

Merged PR #5844 established that `net_amount` includes both `shipping_cost` and `order_tax_amount`, and merged PR #10805 later fixed recalculation of those fields in V1. I did not find an open or closed issue or pull request for the current V2 omission.

### Expected Behavior

For a V2 payment with `order_amount=10000` and `order_tax_amount=800`, `calculate_net_amount()` should return `10800` when no other amount components are present.

### Actual Behavior

These valid V2 details produce `10000`. The order tax remains available in the default tax field, but it is absent from the calculated net amount.

### Steps To Reproduce

On `main` at `b68f937887b78b577958ff29c97610c61a20fe2e`, add this integration test as `crates/hyperswitch_domain_models/tests/v2_order_tax_net_amount.rs`:

```rust
use api_models::payments;
use hyperswitch_domain_models::payments as domain_payments;

#[test]
fn v2_net_amount_includes_order_tax() {
    let request_amounts: payments::AmountDetails = serde_json::from_value(serde_json::json!({
        "order_amount": 10_000,
        "currency": "USD",
        "order_tax_amount": 800,
        "skip_external_tax_calculation": "skip",
        "skip_surcharge_calculation": "skip"
    }))
    .expect("valid V2 amount details");

    let domain_amounts: domain_payments::AmountDetails = request_amounts.into();

    assert_eq!(domain_amounts.calculate_net_amount().get_amount_as_i64(), 10_800);
}
```

Run:

```text
cargo test -p hyperswitch_domain_models --no-default-features --features v2,revenue_recovery --test v2_order_tax_net_amount -- --nocapture
```

Current result:

```text
assertion `left == right` failed
  left: 10000
 right: 10800
```

### Context For The Bug

I found this while checking amount composition invariants across the V2 payment path. A temporary local change that adds the default order tax in `calculate_net_amount()` makes the regression pass; I removed that change after verification.

That change is probably not the right shape, because `create_attempt_amount_details` resolves the attempt's own `order_tax_amount` through `TaxDetails::get_tax_amount(payment_method_subtype)`, which prefers a matching payment method type tax and falls back to the default, while `calculate_net_amount()` receives no subtype and can only reach `get_default_tax_amount()`. The helper also reads no calculation override, whereas the attempt path populates `order_tax_amount` only under `Skip`. Should the V2 net amount use the default tax, or resolve the same payment method aware value the attempt already uses?

I am willing to prepare the bounded fix and regression once that is settled.

### Environment

Not a hosted instance.

1. Operating system: Apple silicon (`arm64`)
2. Rust version: `rustc 1.97.1 (8bab26f4f 2026-07-14)`
3. Source version: `main` at `b68f937887b78b577958ff29c97610c61a20fe2e`

Mohil-Ahuja:
## Type of Change

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

`AmountDetails::calculate_net_amount()` (V2) summed the order amount, shipping cost, surcharge and
tax on surcharge, but never added `order_tax_amount`. The V1 equivalent,
`NetAmount::get_total_amount()`, does include it, son what a payment
is worth.

The fix resolves the order tax amount and the net amount together, so that the net amount always
includes exactly the `order_tax_amount` that is repo

- `get_order_tax_amount(payment_method_subtype)` resolves the tax that applies. `None` yields the
  payment-method-agnostic default tax, which is what already falls
  back to.
- `calculate_net_amount_with_order_tax(order_tax_amount)` does the summation once, from a tax
  amount the caller supplies.
- `calculate_net_amount()` keeps its signature and behaviour for intent level callers, now passing
  the default tax.
- `create_attempt_amount_details` and `proxy_create_attempt_amount_details` pass the
  payment-method-aware amount they already resolve fy persist, instead
  of recomputing a different one.

Choosing the default tax at intent level follows what the codebase already does for the sibling
field: `PaymentAmountDetailsResponse::foreign_from` x_amount()` for
`order_tax_amount` when there is no attempt (`transformers.rs:7477-7482`), and five other call
sites use the same fallback for the same reason.

Behaviour is deliberately unchanged when `skip_external_tax_calculation` is `Calculate`: the
external tax provider has not been called yet, so the attempt reports no `order_tax_amount` and the
net amount does not include one either.

### Additional Changes

- [ ] This PR modifies application configuration/env

No API contract change: the shape of `PaymentAmountDetailsResponse` is untouched. The `net_amount`
value it carries becomes correct, which is the point

## Motivation and Context

Fixes #13712.

`calculate_net_amount()` has eight call sites, and t one of them:

| Call site | Effect |
| --- | --- |
| `AmountDetails::create_attempt_amount_details` | attempt created, and payment authorized, for less than the order is worth |
| `AmountDetails::proxy_create_attempt_amount_detailrm path |
| `core/split_payments.rs:151` | gift card allocation under-covers the order, `remaining_to_allocate` starts below the real total |
| `core/payments/routing.rs:2694` | session routing rules evaluated against an amount lower than what will be charged |
| `core/payments/payment_methods.rs:558` | `filter_amount_based` can keep a payment method whose `maximum_amount` cannot cover the
order |
| `core/payments/payment_methods.rs:716` | a tax-only intent is mistaken for a zero amount (mandate) intent |
| `core/payments/transformers.rs:7493` | `net_amountamount` reported beside it |
| `core/payments/transformers.rs:3175` | same, and the value is also reused as `amount_captured` for split payments |

`AttemptAmountDetails::net_amount` also documents itself as "The total amount for this payment
attempt. This includes all the surcharge and tax amounts", so V2 was not meeting the contract
declared by its own type.

## How did you test it?

Added unit tests in `crates/hyperswitch_domain_models/src/payments.rs`:

```
cargo test -p hyperswitch_domain_models --no-defaultt_details_tests
```

- `net_amount_includes_default_order_tax` — the case from the issue: `order_amount` 10000 with an
  `order_tax_amount` of 800 now gives 10800, not 10000.
- `net_amount_without_tax_details_is_the_order_amoung intents that
  carry no tax at all.
- `net_amount_sums_every_component` — order amount, shipping, tax, surcharge and tax on surcharge.
- `net_amount_ignores_payment_method_specific_tax` — a tax calculated for one specific payment
  method must not be charged before a payment method
- `attempt_net_amount_matches_the_reported_order_tax_amount` — the invariant, on the attempt path.
- `attempt_excludes_tax_that_is_yet_to_be_calculatedalculation` set to
  `Calculate`, neither field carries a tax amount.

## Checklist

- [x] I formatted the code `cargo +nightly fmt --all


Mohil-Ahuja:
Refreshed against main and the checks are green. Same ask as the other one: this is a narrow v2 amount calculation fix and should be a short read whenever a reviewer has a moment.

PR #13727, opened 2026-08-16
feat(offers): notify offer engine via process tracker

apoorvdixit88:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

**Offer Engine: asynchronous offer revocation on failure / refund.**

After `/apply`, an offer is already marked availed (its transaction sits in the Offer Engine `PENDING` state and its counters are held). Hyperswitch therefore does **not** notify on a successful/charged payment — that outcome is already covered by `/apply`. Instead, when a payment that has an applied offer reaches a **terminal failure**, or a **refund** on such a payment succeeds, Hyperswitch notifies the Offer Engine via `POST /offers/notify` to **revoke** the offer (rolling its counters back / freeing any once-per-card or budget slot).

Delivery is fully asynchronous through **Process Tracker**, so Offer Engine availability never affects the primary payment or refund flow (best-effort scheduling; a failure to schedule or deliver never changes the payment/refund result).

**Outcome → Offer Engine mapping**

| Hyperswitch outcome | Notifies? | `txn_status` | offer `status` |
|---|---|---|---|
| Payment charged / success | **No** (already covered by `/apply`) | — | — |
| In-flight / ambiguous attempt states | **No** | — | — |
| Terminal payment failure (`Failure`, `AuthenticationFailed`, `AuthorizationFailed`, `RouterDeclined`, `Voided`, `VoidFailed`, `CaptureFailed`, `Expired`) | **Yes** | `FAILURE` | `REVOKED` |
| Connector-driven auto-refund (`AutoRefunded` attempt) | **Yes** | `FAILURE` | `REVOKED` |
| Refund success (full or partial) | **Yes** | `FAILURE` | `REVOKED` |

From the post-`/apply` `PENDING` state the only valid revoke transitions are `FAILURE` and `AUTO_REFUNDED` (`REFUNDED` requires a prior `CHARGED`, which we never send — verified live: notifying a `PENDING` txn as `REFUNDED` returns `422 Status Update Not Allowed`). We report **`FAILURE`** for every case — terminal failure, refund, and connector auto-refund — since it is a single sufficient revoke from `PENDING` and rolls the counters back identically. Every notification carries offer `status = REVOKED`.

**What's included**

- v1-only `OfferEngineNotifyWorkflow` Process Tracker workflow + `ProcessTrackerRunner::OfferEngineNotifyWorkflow` variant + scheduler dispatch arm.
- Best-effort scheduling hooks:
  - Payment post-update: `attempt_notify_status(status) -> Option<OfferTxnStatus>` returns `Some(FAILURE)` for terminal outcomes (including connector auto-refund) and `None` for success/in-flight (so no task is scheduled on success).
  - Refund success path: schedules `FAILURE` + `REVOKED`.
- **Idempotent** scheduling via a deterministic task id (`offer_engine_notify_payment_{attempt_id}` / `offer_engine_notify_refund_{refund_id}`); a unique-constraint violation on re-processing (webhook / sync / retry) suppresses the duplicate.
- At execution: reconstruct the request from the persisted `applied_offer_details` (no raw `/list`/`/apply` responses stored, no sensitive PM/customer data in the payload), resolve Offer Engine config at **processor + org + profile** scope (same scope used at `/apply`), call `POST /offers/notify`, retry transient failures via the existing Process Tracker retry policy, and terminate on invalid state / missing credentials.
- Slimmed DTOs: `OfferTxnStatus` = `{ Failure }` (kept as an extensible enum), `OfferNotifyStatus` = `{ Revoked }`. The optional `origin` field is omitted entirely — it's `Maybe` in the engine and never read (no branching, not persisted), so we don't send it. Plus notify client flow, counters, and structured logs.

No new public API surface, no DB schema change (reuses the `process_tracker` table), and no new configuration (reuses the M1 Offer Engine config + enablement keys).

### Additional Changes

## Motivation and Context

`/apply` already tells the Offer Engine an offer was availed (transaction `PENDING`, counters held). What the Offer Engine still needs is to be told when that availed offer should be **undone** — a terminal payment failure or a refund — so it can revoke the offer and roll the counters back (freeing once-per-card / budget slots). Doing this synchronously in the payment/refund path would couple Hyperswitch's core flow to a third party's availability, so revocations are delivered out-of-band through Process Tracker.

Internal tracking: Offer Engine Milestone 3 (`juspay/hyperswitch-cloud#21535`); M1 PR #13433.

## How did you test it?

Manually, end-to-end, against the **real Offer Engine sandbox** (`sandbox.juspay.in`) with the router + Process Tracker producer/consumer running, and the enum/request shapes cross-checked against the Offer Engine source. (Debug builds need a larger stack: `RUST_MIN_STACK=33554432`.) API keys redacted.

```bash
cargo build --bin router --bin scheduler --features v1
# Offer Engine enabled via Superposition: offer_engine_enabled=true, offer_engine_credential_source="application"
RUST_MIN_STACK=33554432 ./target/debug/router
SCHEDULER_FLOW=producer ROUTER__SCHEDULER__SERVER__PORT=3002 ./target/debug/scheduler
RUST_MIN_STACK=33554432 SCHEDULER_FLOW=consumer ROUTER__SCHEDULER__SERVER__PORT=3001 ./target/debug/scheduler
```

Each flow is create → `POST /payments/{id}/eligibility` (BIN `424242`) → `POST /payments/{id}/confirm` with the returned `offer_details.offer_quote_ids`.

### 1. Success does NOT notify

Charged payment with an applied offer (net reduced by the discount) scheduled **no** offer-notify task — only `PAYMENTS_SYNC`. Confirms success is left to `/apply`.

### 2. Refund → revoke / unblock

Payment with offer applied → full refund. The refund hook scheduled `offer_engine_notify_refund_<refund_id>`:

```
process_tracker: status=finish, business_status=COMPLETED_BY_PT, retry_count=0
consumer log:    POST https://sandbox.juspay.in/offers/notify  status_code: 200  (errorcode: "")
                 Offer engine notification delivered  (txn_status FAILURE, offer REVOKED)
```

Re-running `/eligibility` for the same card afterward returned the offer as available again — the revoke freed the slot (unblock).

### 3. Failure → revoke (payment-level), budget not consumed

Declined card (`4000000000000002`) on an offer that was applied at `/apply` then failed at the connector (`status=failed`, "card was declined"). The payment post-update hook scheduled `offer_engine_notify_payment_<attempt_id>`:

```
process_tracker: status=finish, business_status=COMPLETED_BY_PT
consumer log:    POST /offers/notify 200 (txn_status FAILURE, offer REVOKED)
```

Verified the failed payment did **not** consume the offer's budget (a subsequent good payment still availed) — i.e. the failure revoke behaves exactly like a refund revoke.

### 4. Blocking / unblocking against real Offer Engine counters

Created counter-limited offers in the sandbox to exercise revoke end-to-end:

- **Count budget** (`OFFER_AMOUNT`/`COUNT`, max 4): 4 avails consumed the counter `0 → 4`; the 5th was `INELIGIBLE` ("COUNTER … has reached Max") and fell back to another offer. Refunding one avail drove the counter `4 → 3` (via the notify revoke) and the offer became eligible again.
- **Amount budget** (`OFFER_AMOUNT`, cap read live from `/offers/list.offer_amount_available`): flat $10 discount, $50 cap → remaining stepped `50 → 40 → 30 → 20 …` per avail and blocked at exhaustion; a refund/failure revoke returned budget.

### 5. Idempotency

Re-processing the same outcome (force-sync) did not create a duplicate — the deterministic task id + unique-violation suppression kept the task count at exactly 1.

### 6. Wire-shape correctness cross-checked against Offer Engine source

- `txn_status` (`FAILURE`) and offer `status` (`REVOKED`) are in Offer Engine's `FromJSON` accepted sets (OE rejects unknowns). `txn_status` is mandatory — omitting it returns HTTP 400 `MISSING_FIELD` — which our non-optional DTO field guarantees.
- `validateReqStatus` permits a revoke only from `PENDING`/`FAILURE`/`AUTO_REFUNDED` with all offers `REVOKED` — which is exactly what we send. `REVOKED` is also the idempotency-safe status (the engine's already-terminal guards key off it), so a re-delivered notify is a no-op rather than a double-decrement.

## Checklist


XyneSpaces:
[blocking] `schedule_refund_notification` classifies a refund as full only when the single refund row's `refund_amount >= total_amount`, but `refund_amount` is the amount of this refund and `total_amount` is the original payment amount. A payment fully refunded through multiple partial refunds will keep sending `PARTIALLY_REFUNDED`/`REVOKED`; please base the full-refund notification on cumulative refunded amount (or the post-refund payment/refund state) instead of this single-row comparison.

XyneSpaces:
[blocking] Gate the new payment-response scheduling call with `#[cfg(feature = "v1")]` as well. The import/export for `offer_engine::schedule_payment_notification_for_attempt` is v1-only, so a v2-only build will fail to resolve that symbol after this PR.

[blocking] Add coverage for the new Offer Engine Process Tracker notification flow. This PR adds scheduling, retry, and terminal-failure behavior for payment/refund notifications, but no router/process-tracker tests exercise success, retry, missing-offer-details, or config-disabled paths.

XyneSpaces:
[blocking] This adds a new Offer Engine process-tracker notification flow without any corresponding tests in the PR. Please add coverage for scheduling on terminal payment/refund outcomes and for `execute_notification` success/retry/terminal-failure behavior so regressions in this new async delivery path are caught.

crates/router/src/core/offer_engine/notify.rs
```diff
@@ -0,0 +1,410 @@
...
+        Ok(None) => false,
+        Err(err) => {
+            logger::warn!(
+                ?err,
+                "Offer Engine config could not be resolved while scheduling notification"
+            );
+            false
+        }
+    }
+}
```
ThisIsMani: Should we stop notification based on the current config? If a offer is applied before when config was enabled and at notify before notifying, and we don't notify, isn't that wrong?

jarnura: +1

crates/router/src/core/offer_engine/notify.rs
```diff
@@ -0,0 +1,410 @@
...
+    let config = match resolve_offer_engine_config(state, &dimensions).await {
+        Ok(Some(config)) => config,
+        other => {
+            logger::error!(
+                ?other,
+                process_tracker_id = %process.id,
+                "Offer Engine disabled or credentials unavailable; terminating notify task"
+            );
+            metrics::OFFER_ENGINE_NOTIFY_TERMINAL_FAILURES.add(1, &attributes);
+            return db
```
jarnura: Avoid return based code

crates/router/src/core/offer_engine/notify.rs
```diff
@@ -0,0 +1,410 @@
...
+        logger::warn!(
+            payment_id = %tracking_data.payment_id.get_string_repr(),
+            payment_attempt_id = %tracking_data.payment_attempt_id,
+            refund_id = ?tracking_data.refund_id,
+            ?notification_type,
+            process_tracker_id = %process.id,
+            "applied_offer_details missing; terminating offer engine notify task"
+        );
+        metrics::OFFER_ENGINE_NOTIFY_TERMINAL_FAILURES.add(1, &attributes);
+        return db
```
jarnura: Avoid return based code

jarnura (changes requested):

ThisIsMani (approved):

jarnura (approved):

PR #13750, opened 2026-08-18
feat(account_updater): apply card updates to stored payment methods

kanikabansal08:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

Stacked on #13510. That PR gets an answer from UCS; this one acts on it. **v2 only, `v2_compatible` migration track, generic locker only.**

- When Account Updater reports `AccountUpdated` or `ExpiryUpdated`, the returned card is stored by forking the payment method record: a new row carrying the refreshed card is written under the **same id**, so every reference the merchant already holds keeps working. The force-sync retrieve responds with the row that was actually written.
- `run_account_updater` only reports what the issuer holds — it performs no writes. The retrieve matches on the result and drives the vault and database work itself, so a future refreshable payment method kind is a compile error at that call site rather than a silent gap.
- Write order is vault → insert the new row as `New` → redact the superseded row and drop its locker entry → promote the new row to `Active`. Nothing is destroyed until its replacement exists, so a failure at any step leaves the merchant with the card they already had and needs no compensating write. A failed promotion is logged and left for the first payment to correct, as it would be for any newly added card.
- The new card takes its number and expiry from the issuer and its holder name and nickname from the stored row. Network, issuer, type and country are left unset and re-derived by the BIN lookup, since the stored values describe the old number.
- One id can now span several rows, held up by two indexes that are not the same rule: `UNIQUE (id, locker_fingerprint_id)` stops the same card being stored twice under one id, and `UNIQUE (id) WHERE status = 'active'` is what carries "one live card per id". The partial index is what forces the write order above.
- Redacted rows carry a **NULL** fingerprint rather than the old `FINGERPRINT_ID_REDACTED` sentinel, which would collide under the composite index on the second redaction. `StatusAndFingerprintUpdate.locker_fingerprint_id` becomes `Option<Option<String>>` so the changeset can express "write NULL" as distinct from "leave alone"; `GenericUpdate` maps through `.map(Some)` and so can never clear it.
- Read paths stop assuming one row per id: `find_by_id` prefers a live row over a redacted one and then the newest, so a card change that stops part way resolves back to the row that was already serving. `update_with_id` is pinned to the row the caller loaded via `id = $1 AND locker_fingerprint_id IS NOT DISTINCT FROM $2`, since the fork deliberately holds two live rows and the redaction must land on the older one.
- Deployments that cannot express the design are rejected before the UCS call: the legacy locker never populates `locker_fingerprint_id`, and an external vault profile stores the card at the vault connector rather than the locker this flow writes to.

### Additional Changes

Four migrations on the `v2_compatible` track:

- `000001_prepare_payment_methods_for_id_index_swap` — clears existing sentinels to NULL. Expected to affect zero rows on a healthy database; it exists so the `CONCURRENTLY` build that follows cannot meet a violation.
- `000002` / `000003` — build `UNIQUE (id, locker_fingerprint_id)` and `UNIQUE (id) WHERE status = 'active'`.
- `000004` — drops the old `payment_methods_id_index`.

Both new indexes are built **before** the old one is dropped, so a failed build leaves the existing uniqueness guarantee in place.

Adds `CreatedBy::AccountUpdater`, so writes made while applying an issuer-reported change are attributed to the system rather than to the caller who triggered the force-sync.

## Motivation and Context

Closes juspay/hyperswitch-cloud#22846.

A reported card change must be stored without changing the payment method id the merchant holds, and without ever leaving two live cards under one id.

**Deploy precondition:** `000003` fails if any id still has two or more active rows. It is a `CONCURRENTLY` build, so there is no rollback — a failure leaves an INVALID index that costs writes and is never used by the planner, and recovery is a manual `DROP INDEX CONCURRENTLY payment_methods_id_active_index` before retrying. Existing duplicate rows must be cleared before this ships.

## How did you test it?

Manually, end-to-end against the **real Juspay Account Updater sandbox**, through a local stack — router (v2), UCS, card vault, Kafka, Postgres on the `v2_compatible` migration track. Test card `5204500130000005` (Mastercard, stored as `12/2028`); the sandbox reports an updated expiry for it. Payment method id `0a_pm_01a01eda564e7a93b94615fca897d0f7` throughout. Account Updater only runs for internal auth, so every force-sync below carries `X-Internal-Api-Key` / `X-Merchant-Id` / `X-Profile-Id`; the merchant-key call is included deliberately as the negative case.

### Migrations

Run against a database seeded **before** the migration with the rows it is meant to meet: two redacted rows carrying the `FINGERPRINT_ID_REDACTED` sentinel, an **active** row also carrying the sentinel (must be left alone), an inactive row, a `new` row with no fingerprint, and two real vaulted payment methods redacted the old way.

```
--- 1. sentinel gone from redacted rows (000001)
 sentinel_cleared | redacted_rows_still_carrying_sentinel
------------------+---------------------------------------
 PASS             |                                     0

--- 2. non-redacted rows untouched by 000001
                  id                   |  status  |  locker_fingerprint_id  | verdict
---------------------------------------+----------+-------------------------+---------
 12345_pm_seed_active_0000000000000001 | active   | seed_fp_1               | PASS
 12345_pm_seed_inactv_000000000000004  | inactive | seed_fp_4               | PASS
 12345_pm_seed_new_00000000000000006   | new      |                         | PASS
 12345_pm_seed_redact_000000000000002  | redacted |                         | PASS
 12345_pm_seed_redact_000000000000003  | redacted |                         | PASS
 12345_pm_seed_weird_0000000000000005  | active   | FINGERPRINT_ID_REDACTED | PASS

--- 3. indexes
 CREATE UNIQUE INDEX payment_methods_id_active_index ON public.payment_methods USING btree (id) WHERE ((status)::text = 'active'::text)
 CREATE UNIQUE INDEX payment_methods_id_locker_fingerprint_id_index ON public.payment_methods USING btree (id, locker_fingerprint_id)

 composite_unique_present | partial_active_present | old_id_index_dropped
--------------------------+------------------------+----------------------
 PASS                     | PASS                   | PASS

--- 4. no index left INVALID          PASS
--- 5. __diesel_schema_migrations     820260818000001..820260818000004
```

`000001` cleared the sentinel on redacted rows only — the active row that carries it is untouched, which is the case the `WHERE status = 'redacted'` filter exists for.

**1. `GET /v2/payment-methods/{id}` — baseline, no force sync**

Response (trimmed):
```json
{ "id": "0a_pm_01a01eda564e7a93b94615fca897d0f7",
  "payment_method_data": { "card": { "last4_digits": "0005", "expiry_month": "12", "expiry_year": "2028",
    "card_network": "Mastercard", "issuer_country": "US", "card_issuer": "Test Bank", "card_type": "CREDIT" } } }
```
```
db: active | fp=JSyawRJuK3GxyVSre5zo | locker=01a01eda-569e-7032-a58a-9d7c33f71ca5 | by=-
```

**2. `?force_sync=true` with the MERCHANT key → Account Updater must not run**

Response (trimmed): identical to step 1 — `12/2028`, same fingerprint.
```
db: active | fp=JSyawRJuK3GxyVSre5zo | locker=01a01eda-569e-7032-a58a-9d7c33f71ca5 | by=-
kafka: no event
```
The parameter is accepted but inert without internal auth: no UCS call, no event, no write.

**3. `?force_sync=true&fetch_raw_detail=true` with internal auth → `ExpiryUpdated` applied**

Response (trimmed):
```json
{ "id": "0a_pm_01a01eda564e7a93b94615fca897d0f7",
  "payment_method_data": { "card": { "last4_digits": "0005", "expiry_month": "12", "expiry_year": "2029",
    "card_network": "Mastercard", "issuer_country": "US", "card_issuer": "Test Bank", "card_type": "CREDIT" } },
  "raw_payment_method_data": { "card": { "card_number": "5204500130000005", "card_exp_month": "12",
    "card_exp_year": "2029", "card_holder_name": "AU Tester", "nick_name": "au-mc_pr",
    "card_issuing_country": "US", "card_network": "Mastercard", "card_issuer": "Test Bank", "card_type": "credit" } } }
```
```
db: active   | fp=kM0t84arvup9bUYTtDuf | locker=01a01eda-73ac-7da2-9569-495593764a5e | by=AccountUpdater:juspay
    redacted | fp=NULL                 | locker=01a01eda-569e-7032-a58a-9d7c33f71ca5 | by=-
vault: new entry 01a01eda-73ac-7da2-9569-495593764a5e present, superseded entry gone
```
Same id, same PAN, expiry `12/2028 → 12/2029`. Two rows under one id, exactly one `active`; the superseded row is `redacted` with a **NULL** fingerprint, not the sentinel — which is what lets the same id be redacted a second time in step 4. The write is attributed to `AccountUpdater:juspay`, and the response is the row that was actually written.

**4. Second refresh on an already-updated row**

Same call again. Response (trimmed): `expiry_year: "2032"`, `card_network: "Mastercard"`.
```
db: active   | fp=HRZcee1oIYF26stJqn9i | locker=01a01eda-76d3-7f31-9a15-83496b46355f | by=AccountUpdater:juspay
    redacted | fp=NULL                 | locker=01a01eda-73ac-7da2-9569-495593764a5e | by=AccountUpdater:juspay
    redacted | fp=NULL                 | locker=01a01eda-569e-7032-a58a-9d7c33f71ca5 | by=-
```
Three rows under one id, still exactly one active, two redacted rows both holding NULL. This is the case the composite index would have rejected under the old sentinel, and the case an AU-written row must survive: the replacement row carries the stored network forward, so it stays eligible for the next refresh instead of failing the network gate.

**5. `DELETE /v2/payment-methods/{id}` after a card change**

Response: `200`, `{ "id": "0a_pm_01a01eda564e7a93b94615fca897d0f7" }`
```
db: redacted | fp=NULL | locker=01a01eda-76d3-7f31-9a15-83496b46355f | by=AccountUpdater:juspay
    redacted | fp=NULL | locker=01a01eda-73ac-7da2-9569-495593764a5e | by=AccountUpdater:juspay
    redacted | fp=NULL | locker=01a01eda-569e-7032-a58a-9d7c33f71ca5 | by=-
vault: 0 entries for this id
```
Delete reaches **every** row under the id, not just the live one, and clears the vault entries with them. Retrieve and force-sync afterwards both return `404 HE_02 "Payment method does not exist in our records"`.

**6. Observation events**

One event per refresh on `hyperswitch-account-updater-events`, keyed by payment method id:
```json
{"request_id":"01a01eda-6e53-72c2-ad5e-68883095d4cd","merchant_id":"aum3merchant_l43yomG4MCulG4nutlIy","profile_id":"pro_Se2uqmgknV6Npay9zl4O","payment_method_id":"0a_pm_01a01eda564e7a93b94615fca897d0f7","card_network":"Mastercard","updater_outcome":"CARD_REFRESH_EXPIRY_UPDATED","error_category":null,"latency_ms":1332,"created_at":1787224028065314000}
{"request_id":"01a01eda-73f5-7a91-9b93-3d606807ca2b","merchant_id":"aum3merchant_l43yomG4MCulG4nutlIy","profile_id":"pro_Se2uqmgknV6Npay9zl4O","payment_method_id":"0a_pm_01a01eda564e7a93b94615fca897d0f7","card_network":"Mastercard","updater_outcome":"CARD_REFRESH_EXPIRY_UPDATED","error_category":null,"latency_ms":712,"created_at":1787224028874849000}
```

### Other outcomes

**`NoChange`** — Mastercard `5567630000003002` `12/2028`, sandbox reports no change. Card and expiry unchanged, **one** row, fingerprint and vault entry untouched, event `CARD_REFRESH_NO_CHANGE`. The no-write path stays a no-write path.

**Unsupported network** — Amex `378282246310005`. Refused by the eligibility gate before the UCS call; card unchanged, one row, event carries `"updater_outcome": null, "error_category": "unsupported_network"`, log `Account Updater refresh did not complete, error: Card network is not supported by Account Updater`.

Result: a reported card change is stored under the same id with one live card at all times, the superseded row is redacted with a NULL fingerprint so the id can be updated again, delete reaches the whole chain, and every attempt — applied, no-change, or refused — emits exactly one observation event.

Static checks: `cargo check -p router` (v2 feature set), `cargo clippy`, `cargo +nightly fmt --all --check` — all clean.

## Checklist


v2_compatible_migrations/2026-08-18-000001_prepare_payment_methods_for_id_index_swap/up.sql
```diff
@@ -0,0 +1,4 @@
+UPDATE payment_methods
+SET locker_fingerprint_id = NULL
+WHERE locker_fingerprint_id = 'FINGERPRINT_ID_REDACTED'
```
ThisIsMani: Can we remove this?

v2_compatible_migrations/2026-08-18-000001_prepare_payment_methods_for_id_index_swap/down.sql
```diff
@@ -0,0 +1,5 @@
+UPDATE payment_methods
+SET locker_fingerprint_id = 'FINGERPRINT_ID_REDACTED'
+WHERE locker_fingerprint_id IS NULL
+  AND locker_id IS NOT NULL
```
ThisIsMani: Do we need this?

v2_compatible_migrations/2026-08-18-000002_add_payment_methods_id_locker_fingerprint_id_index/up.sql
ThisIsMani, line 1: Should we put this in v2 migrations as well?

crates/diesel_models/src/payment_method.rs
```diff
@@ -1244,7 +1244,7 @@ impl From<PaymentMethodUpdate> for PaymentMethodUpdateInternal {
                 network_token_requestor_reference_id,
                 network_token_locker_id,
                 network_token_payment_method_data,
-                locker_fingerprint_id,
+                locker_fingerprint_id: locker_fingerprint_id.map(Some),
```
ThisIsMani: NIT: Can we use Some(..) instead.

crates/common_utils/src/types.rs
```diff
@@ -1588,6 +1588,11 @@ impl_enum_str!(
             /// merchant id of creator.
             merchant_id: String,
         },
+        /// AccountUpdater variant, for writes made while applying a reported card change
+        AccountUpdater {
+            /// merchant id whose payment method was updated.
+            merchant_id: String,
```
ThisIsMani: Should you have the service instead? In our case, Juspay.

crates/diesel_models/src/query/payment_method.rs
```diff
@@ -269,8 +271,23 @@ impl PaymentMethod {
...
-        generics::generic_find_one::<<Self as HasTable>::Table, _, _>(conn, pm_id.eq(id.to_owned()))
-            .await
+        generics::generic_filter::<<Self as HasTable>::Table, _, _, Self>(
+            conn,
+            pm_id.eq(id.to_owned()),
+            Some(1),
+            None,
+            Some((
+                dsl::status
+                    .ne(storage_enums::PaymentMethodStatus::Redacted)
```
ThisIsMani: I think this shouldn't be there.

crates/diesel_models/src/payment_method.rs
kanikabansal08: this is needed when input is None, .map(some) will keep it none whereas some(None) will set it as null.

cypress-tests/cypress/support/commands.js
ThisIsMani, line 1: These are formatting changes?

crates/diesel_models/src/query/payment_method.rs
```diff
@@ -333,12 +345,20 @@ impl PaymentMethod {
         conn: &PgPooledConn,
         payment_method: payment_method::PaymentMethodUpdateInternal,
     ) -> StorageResult<Self> {
```
ThisIsMani, line 357: Should we name this function `update_with_id_and_locker_fingerprint_id`?

crates/diesel_models/src/payment_method.rs
ThisIsMani: I think it's better to make the Generic Update locker_fingerprint_id to `Option<Option<String>>`.

crates/router/src/core/account_updater.rs
```diff
@@ -27,24 +28,25 @@ pub async fn run_account_updater<D>(
     payment_method: &domain::PaymentMethod,
     raw_payment_method_data: Option<&RawPaymentMethodData>,
     dimensions: &D,
-) where
+) -> Option<RefreshResult>
```
ThisIsMani: Convert this to Result.

crates/router/src/core/account_updater/types.rs
```diff
@@ -112,6 +120,18 @@ impl From<&settings::AccountUpdaterConfig> for ResolvedAccountUpdaterConfig {
...
+    Card(CardRefreshResult),
+}
+
+pub struct CardRefreshResult {
+    pub outcome: payments_grpc::CardRefreshOutcome,
+    /// Carries a card only on the outcomes that supersede the stored one.
+    pub refreshed_card: Option<payments_grpc::CardDetailsWithNoCvc>,
+    /// The service that reported the change, attributed on the rows written to apply it.
+    pub service: Connector,
+}
```
ThisIsMani: Let's tie refreshed_card_data to outcome.

crates/router/src/core/account_updater/store.rs
```diff
@@ -0,0 +1,321 @@
...
+        )
+        .await
+        .change_context(AccountUpdaterError::StoreFailed)
+}
+
+fn account_updater_created_by(service: Connector) -> common_utils::types::CreatedBy {
+    common_utils::types::CreatedBy::AccountUpdater {
+        service: service.to_string(),
+    }
+}
```
ThisIsMani: Please remove this function.

crates/router/src/core/account_updater/store.rs
```diff
@@ -0,0 +1,321 @@
...
+        created_at: now,
+        last_modified: now,
+        last_modified_by: Some(created_by),
+        updated_by: Some(created_by_string),
+        payment_method_subtype: payment_method_subtype.or(payment_method.payment_method_subtype),
+
+        connector_mandate_details: None,
+        client_secret: None,
+
+        ..payment_method.clone()
```
ThisIsMani: I think it's better not to use this...

crates/router/src/core/account_updater/refresh.rs
```diff
@@ -75,15 +78,40 @@ pub async fn request_account_updater_refresh(
...
-                payments_grpc::CardRefreshOutcome::try_from(card.outcome)
-                    .unwrap_or(payments_grpc::CardRefreshOutcome::Unspecified)
-            }
-        })
         .ok_or_else(|| report!(AccountUpdaterError::RefreshReturnedError))
-        .attach_printable("UCS returned neither a result nor an error")
+        .attach_printable("UCS returned neither a result nor an error")?;
+
+    match result {
+        payments_grpc::refresh_result::Result::Card(card) => Ok(RefreshResult::Card(
```
ThisIsMani: NIT: Reduce the import size.

crates/router/src/core/account_updater/refresh.rs
```diff
@@ -75,15 +78,40 @@ pub async fn request_account_updater_refresh(
...
+            build_card_refreshed_data(card, config.service())?,
+        )),
+    }
+}
+
+fn build_card_refreshed_data(
+    card_result: payments_grpc::CardRefreshResult,
+    service: Connector,
+) -> CustomResult<CardRefreshedData, AccountUpdaterError> {
+    use payments_grpc::CardRefreshOutcome as Outcome;
```
ThisIsMani: Don't write imports here.

crates/router/src/core/account_updater/store.rs
```diff
@@ -0,0 +1,332 @@
...
+                common_enums::PaymentMethodStatus::Inactive,
+            )
+            .await
+            .attach_printable("Failed to deactivate the closed payment method")
+            .map(Some);
+        }
+        None => return Ok(None),
+    };
+
+    let stored_card = payment_method
```
ThisIsMani, line 69: It would be better if you could move everything below this to a function. And the above match would get cleaner.

cypress-tests/cypress/support/commands.js
ThisIsMani: Are we not reverting these changes?

ThisIsMani (dismissed):
Minor concerts. Other than that, LGTM.

ThisIsMani (dismissed):

cypress-tests/cypress/support/commands.js
kanikabansal08: added by the bot again.

crates/diesel_models/src/query/payment_method.rs
```diff
@@ -269,8 +271,18 @@ impl PaymentMethod {
         conn: &PgPooledConn,
         id: &common_utils::id_type::GlobalPaymentMethodId,
     ) -> StorageResult<Self> {
-        generics::generic_find_one::<<Self as HasTable>::Table, _, _>(conn, pm_id.eq(id.to_owned()))
-            .await
+        generics::generic_filter::<<Self as HasTable>::Table, _, _, Self>(
```
jarnura, line 284: Why this is not find_one? by the condition there should be only one active.

crates/diesel_models/src/query/payment_method.rs
```diff
@@ -328,17 +340,25 @@ impl PaymentMethod {
...
             <Self as HasTable>::Table,
             _,
             _,
             _,
-        >(conn, pm_id.eq(self.id.to_owned()), payment_method)
+        >(
+            conn,
+            pm_id
+                .eq(self.id.to_owned())
+                .and(dsl::locker_fingerprint_id.is_not_distinct_from(locker_fingerprint_id)),
```
jarnura, line 369: What is usage of is_not_distinct_from?

crates/router/src/core/account_updater/store.rs
```diff
@@ -0,0 +1,349 @@
...
+
+    let auxiliary_fingerprint_id = auxiliary_fingerprint_id_result
+        .change_context(AccountUpdaterError::StoreFailed)
+        .attach_printable("Failed to compute the auxiliary fingerprint for the refreshed card")?;
+
+    if payment_method.locker_fingerprint_id.as_deref() == Some(locker_fingerprint_id.as_str()) {
+        logger::info!(
+            "Account Updater reported a change whose fingerprint matches the stored card; nothing written"
+        );
+        return Ok(None);
```
jarnura: Avoid writing early return based flow

crates/router/src/core/account_updater/store.rs
```diff
@@ -0,0 +1,349 @@
...
+        auxiliary_fingerprint_id,
+        external_vault_source,
+        bin_enriched.payment_method_subtype,
+    )
+    .await?;
+
+    let inserted_payment_method =
+        insert_refreshed_payment_method(state, platform, refreshed_payment_method).await?;
+
+    pm_core::delete_payment_method_by_record(
```
jarnura: why we are deleting a payment_method record here?

crates/router/src/core/account_updater/store.rs
```diff
@@ -0,0 +1,349 @@
...
+        common_enums::PaymentMethodStatus::Active,
+    )
+    .await
+    .inspect_err(|error| {
+        logger::warn!(
+            ?error,
+            "Account Updater stored the refreshed card but could not activate the row"
+        )
+    })
+    .unwrap_or(inserted_payment_method);
```
jarnura: What is reason behind doing unwrap_or and treating as this operation succeeds but actually not

crates/router/src/core/account_updater.rs
```diff
@@ -27,24 +29,25 @@ pub async fn run_account_updater<D>(
...
+) -> CustomResult<RefreshResult, AccountUpdaterError>
+where
     D: dimension_state::DimensionsBase,
 {
     let config = match resolve_account_updater_config(state, dimensions).await {
         Ok(Some(config)) => config,
         Ok(None) => {
             logger::debug!("Account Updater is not enabled for these dimensions");
-            return;
+            return Err(report!(AccountUpdaterError::NotEnabled));
```
jarnura:
```suggestion
           Err(report!(AccountUpdaterError::NotEnabled))?;
```

crates/router/src/core/account_updater.rs
```diff
@@ -27,24 +29,25 @@ pub async fn run_account_updater<D>(
...
         }
         Err(error) => {
             logger::warn!(?error, "Account Updater config could not be resolved");
-            return;
+            return Err(error);
```
jarnura:
```suggestion
            Err(error)?;
```

crates/router/src/core/payment_methods.rs
```diff
@@ -6738,14 +6782,16 @@ pub async fn delete_payment_method_by_record(
...
     // Soft delete - mark as Redacted (terminal state, no transitions allowed)
     let pm_update = storage::PaymentMethodUpdate::StatusAndFingerprintUpdate {
         status: Some(enums::PaymentMethodStatus::Redacted),
-        last_modified_by: platform
-            .get_initiator()
-            .and_then(|initiator| initiator.to_created_by())
-            .map(|last_modified_by| last_modified_by.to_string()),
-        locker_fingerprint_id: Some(PAYMENT_METHOD_REDACTED_FINGERPRINT_ID.to_string()),
+        last_modified_by,
+        locker_fingerprint_id: Some(None),
```
jarnura, line 6823: Add comments saying setting the locker_fingerprint_id as null

jarnura (changes requested):

crates/diesel_models/src/query/payment_method.rs
kanikabansal08: Because id is no longer unique after this PR. Redacted rows can have the same id, so find_one could return a redacted row. Also, filtering by status = active isn't sufficient since the row can be New before confirmation or Inactive after closure. Hence, we filter and pick the latest row by `created_at`.

crates/router/src/core/account_updater/store.rs
kanikabansal08: It's the shared soft delete , the row goes to Redacted, its fingerprint is nulled, and the locker entry is removed.

crates/diesel_models/src/query/payment_method.rs
kanikabansal08: It’s NULL-safe equality (`IS NOT DISTINCT FROM`). Since id is no longer unique, we scope the UPDATE to (id, locker_fingerprint_id). Plain = won’t match NULL fingerprints.

crates/router/src/core/account_updater/store.rs
kanikabansal08: it was me not wanting a failed activation to fail the whole retrieve, but it can be the wrong trade. By that point the new row is already inserted and the old one redacted with its vault entry gone, so swallowing the error hands back a New row as if it were live and the caller carries on with a card that never went active.
I can change to ?, keeping the warn log.

jarnura (dismissed):

VenuMadhav2541 (dismissed):

crates/router/src/core/payment_methods.rs
```diff
@@ -5950,24 +5948,70 @@ pub async fn retrieve_payment_method(
...
+                &state,
+                &platform,
+                &profile,
+                &payment_method,
+                card_result,
+            ))
+            .await
+            .change_context(errors::ApiErrorResponse::InternalServerError)
+            .attach_printable("Account Updater failed while applying a card change")?
+        }
```
prasunna09, line 6011: in retrieve_payment_method flow, why should update card details be in a critical path?
we can update asynchronously , and also error shouldn't be propagated.

v2_migrations/2026-08-18-000001_add_payment_methods_id_locker_fingerprint_id_index/up.sql
```diff
@@ -0,0 +1,2 @@
+CREATE UNIQUE INDEX CONCURRENTLY IF NOT EXISTS payment_methods_id_locker_fingerprint_id_index
```
prasunna09: The v2_migrations side never drops the payment_methods PRIMARY KEY (id) (added by 2024-08-28-081838_update_v2_primary_key_constraints/up.sql:72-74, and v2_migrations run last per the justfile, re-imposing it after the compat index swap).

 So on a pure-v2 deployment, insert_refreshed_payment_method inserting a second row with the same id while the old row still exists (delete comes after insert, no on_conflict) hits a duplicate-key violation → StoreFailed → every retrieve of that PM 500s deterministically and the card update can never apply.

prasunna09 (changes requested):

crates/router/src/core/payment_methods.rs
kanikabansal08: I think the write needs to stay on the critical path here. This only runs when the caller explicitly passes `force_sync=true` with internal auth, essentially asking us to check with the issuer and return the current card.

The response is also built from the row we write, not directly from the issuer response.  get_raw_payment_method_data then reads the card from the vault using the new row's locker_id If we background the write, we'd either return the old card or return data that hasn't been committed yet. If the background write fails, we'd only know after the response has already gone out. Since the purpose of this call is to return the current card, I think keeping the refresh and write synchronous makes more sense.

v2_migrations/2026-08-18-000001_add_payment_methods_id_locker_fingerprint_id_index/up.sql
kanikabansal08: This is expected and was discussed with @jarnura. In our current setup, there isn't a concept of a pure v2 deployment. Could you please confirm, @jarnura?

kanikabansal08: Also i think we don't need to add the index to `v2_migrations` separately. `migrate_v2` runs `v2_compatible_migrations` before `v2_migrations`, so the index will already exist by the time the v2 migration runs. I can remove it from there as well.

jarnura: With Account Updater feature, the payment_methods id is no more the primary key, this table will have composite key only even with a pure v2 build also. In pure v2 build the refreshed_payment_method should be inserted only and it shouldn't be a duplicate error. @kanikabansal08 I hope these are covered.

kanikabansal08: Added the patch file.

ThisIsMani: Do we need this in v2_migrations any more since we know that v2 migrations already include v2_compatible, which already includes this index?

crates/router/src/core/payment_methods.rs
ThisIsMani: This is in critical path because, account updater said the existing card has been updated with new details and those should be used. If account updater itself fails, we are not throwing error (in the above function call), but if account updater gives new data, we can't give out old details without updating since those are stale.

ThisIsMani (approved):

kanikabansal08:
**`migration_verify` guarantees that stays consistent.** It's a required check, it regenerates the schema from a fresh database through the patch, and `--locked-schema` fails on any divergence from the committed file naming the exact column

VenuMadhav2541 (approved):

jarnura (approved):

crates/router/src/core/account_updater/refresh.rs
```diff
@@ -75,15 +80,39 @@ pub async fn request_account_updater_refresh(
...
+    card_result: payments_grpc::CardRefreshResult,
+    service: Connector,
+) -> CustomResult<CardRefreshedData, AccountUpdaterError> {
+    let reported_outcome = card_result.outcome();
+
+    let outcome = match (reported_outcome, card_result.card) {
+        (Outcome::CardRefreshAccountUpdated, Some(card)) => CardOutcome::AccountUpdated(card),
+        (Outcome::CardRefreshExpiryUpdated, Some(card)) => CardOutcome::ExpiryUpdated(card),
+        (Outcome::CardRefreshAccountUpdated | Outcome::CardRefreshExpiryUpdated, None) => {
+            return Err(report!(AccountUpdaterError::RefreshReturnedError)
```
kashif-m: nit: avoid early returns

prasunna09 (approved):

PR #13770, opened 2026-08-19
fix(scheduler): skip acknowledge and delete when the consumer batch is empty

### LINKED ISSUE #13642: Scheduler consumer logs an ERROR on every poll of an empty stream since v1.124.0 (xack with no IDs)
### Summary

Since `v1.124.0`, the scheduler consumer logs an `ERROR` on every polling cycle whenever the stream has no pending entries. Task processing itself is unaffected, but on an idle deployment it produces a continuous error stream — measured at 40 lines per minute with the default 3-second `loop_interval`, or about 21% of all consumer log output — which buries genuine scheduler failures.

`v1.123.1` does not do this. Same deployment, same Redis, same config; only the image tag differs.

### What the consumer logs

```
[FETCH_CONSUMER_TASKS - EVENT] Error acknowledging batch in stream
Failed to acknowledge Redis stream entry
├╴at crates/redis_interface/src/module/redis_rs/commands.rs:1002:10
╰─▶ ResponseError: wrong number of arguments for 'xack' command

[CONSUMER_OPERATION_HANDLER - EVENT] Failed to perform consumer operation
Failed to update process batch in stream
├╴at crates/scheduler/src/utils.rs:256:19
```

Two lines per poll — once from the Redis layer, once from the scheduler wrapper above it.

### Cause

In `crates/scheduler/src/utils.rs`, `get_batches` returns early when the read reports the stream is empty:

```rust
Err(error) => {
    if let redis_interface::errors::RedisError::StreamEmptyOrNotAvailable = error.current_context() {
        logger::debug!("No batches processed as stream is empty");
        return Ok(Vec::new());
    }
    ...
}
```

The `stream_read_with_options` signature changed — it now takes a slice of stream keys and a `Vec` of stream IDs:

```rust
// v1.123.1
.stream_read_with_options(stream_name, RedisEntryId::UndeliveredEntryID, Some(1), None, Some((group_name, consumer_name)))

// v1.124.0 and v1.125.0
.stream_read_with_options(&[stream_name.into()], vec![RedisEntryId::UndeliveredEntryID.to_stream_id()], Some(1), None, Some((group_name, consumer_name)))
```

On an empty stream the new implementation returns `Ok` with an empty result rather than `Err(StreamEmptyOrNotAvailable)`, so the early return is unreachable. Execution continues to:

```rust
let entry_ids = entry_ids.into_iter().flatten().collect::<Vec<_>>();

conn.stream_acknowledge_entries(&stream_name.into(), group_name, entry_ids.clone())
```

with `entry_ids` empty, which issues `XACK <key> <group>` with no IDs. Redis requires at least one and rejects the command.

`metrics::BATCHES_CONSUMED` is also incremented before that point, so the counter now over-reports: it counts every empty poll as a consumed batch.

### Which versions are affected

`v1.124.0` and `v1.125.0`. Checked the tags directly — `v1.124.0`'s `get_batches` already pairs the new call signature with the old, now-unreachable error guard.

This appears to originate in #12162 (`feat(redis_interface): redis-rs and fred-rs merge`, merged 2026-05-12), which reworked `stream_read_with_options` as part of making the Redis backend selectable between the `redis` and `fred` crates. `v1.123.1` was released 2026-05-19, after that merge, but was cut from the `v1.123` line and does not contain it — `v1.124.0` (2026-06-30) is the first release that does.

Worth confirming whether the `fred` backend reports an empty stream the same way. If it does, the guard is dead for both backends.

### Impact

Processing is not broken. I verified this on `v1.125.0` by injecting a well-formed batch entry (empty `trackers` array) into the stream and watching the consumer group:

| | before | after |
|---|---|---|
| `last-delivered-id` | `0-0` | `1786483270926-0` |
| `entries-read` | – | `1` |
| `pending` | `0` | `0` |
| `XLEN` | `0` | `0` |

The entry was read, acknowledged and deleted, and the group's pending count stayed at zero. So the failure is confined to the empty-batch case and no task is left stuck. The cost is log noise and a misleading metric.

It is not self-limiting, though: an empty stream is the normal resting state of an idle scheduler, and real traffic does not clear it either, since a poll only avoids the error if it happens to catch an entry inside its particular 3-second window.

There is also no safe way for an operator to filter it out — both lines are emitted at `ERROR` from the same modules that report genuine acknowledge failures, so suppressing them would suppress the real thing.

### Suggested fix

Guard the acknowledge and delete on a non-empty `entry_ids`, and return before `BATCHES_CONSUMED` is incremented:

```rust
if entry_ids.is_empty() {
    logger::debug!("No batches processed as stream is empty");
    return Ok(Vec::new());
}
```

Restoring `StreamEmptyOrNotAvailable` in the new `stream_read_with_options` would also work, but the explicit emptiness check is robust regardless of which backend is compiled in and regardless of how each one chooses to report an empty read.

### Environment

- Reproduced on router `v1.125.0` (`5b9f9a5`); `v1.124.0` affected by inspection; `v1.123.1` unaffected at runtime
- Self-hosted on Kubernetes, consumer and producer as separate deployments
- Redis 7, single instance, TLS via a local proxy
- Default `[scheduler] loop_interval = 3000`

abhisheksharma2411:
## Type of Change

- [x] Bugfix

## Description

`get_batches` decides "the stream is empty, there is nothing to do" from `RedisError::StreamEmptyOrNotAvailable`. `XREADGROUP` does not report an exhausted stream that way — it returns an empty reply — so that branch never runs for the consumer's read. Execution falls through to `stream_acknowledge_entries` with an empty `entry_ids`, and `XACK key group` with no IDs is a syntax error. Every poll of an idle stream logs an `ERROR` pair, one from the Redis layer and one from the scheduler wrapper above it.

The fix returns once `entry_ids` is known to be empty. That holds whichever way a backend reports an empty read, so it is not sensitive to how `redis-rs` and `fred` each choose to surface it.

**Three things worth calling out:**

**1. `XDEL` is affected the same way, which the issue did not cover.** `stream_delete_entries` runs on the same empty `entry_ids`, and `XDEL` rejects an empty ID list for exactly the same reason. It is invisible today only because the acknowledge fails first and returns via `?`. A fix that made `stream_acknowledge_entries` tolerate an empty list would have moved the error one line down rather than removing it — which is part of why the guard belongs in `get_batches`.

**2. The existing error branch is kept rather than removed — and the two backends genuinely differ.** The issue asks: *"Worth confirming whether the fred backend reports an empty stream the same way. If it does, the guard is dead for both backends."* I ran it. **It does not**, so the guard is not dead:

| backend | exhausted `XREADGROUP` returns | existing guard | affected? |
|---|---|---|---|
| `redis-rs` | `Ok`, empty reply | never fires | **yes** |
| `fred` | `Err(StreamEmptyOrNotAvailable)` | fires correctly | no |

So this bug is **redis-rs-only**, and removing the error branch would have broken the fred path. Keeping both is not belt-and-braces — each is load-bearing for one backend. That is also why the fix guards on `entry_ids.is_empty()` rather than on which error came back: it is the one condition that holds either way.

I found this the hard way. My first version of the test asserted the `redis-rs` shape and went red on fred, which is how the divergence surfaced; the test now asserts the union both backends satisfy.

**3. `metrics::BATCHES_CONSUMED` moves below the check.** It was incremented immediately after the read, so every empty poll counted as a consumed batch — on an idle deployment that is the entire series. A side effect worth noting explicitly: a batch that fails to deserialize now returns before the increment too, where previously it was counted. That seems right for a counter named "batches consumed", but say the word if you would rather keep the read-attempt semantics and I will split it out.

## Motivation and Context

Fixes #13642.

The reporter's analysis was accurate and I have not had to correct any of it — this PR is essentially their suggested fix, plus the `XDEL` half and the metric.

## How did you test it?

**Reproduced the mechanism against Redis 7 first**, so the failure is observed rather than inferred:

```
2. first XREADGROUP — delivers the entry (scheduler has work)
   repro13642_stream / 1787101118373-0 / task / process_payment

3. second XREADGROUP — nothing undelivered left
   (empty reply, NOT an error => StreamEmptyOrNotAvailable never fires,
    so get_batches() falls through with entry_ids == [])

4. XACK repro13642_stream repro13642_group   (no IDs)
   ERR wrong number of arguments for 'xack' command

5. XDEL repro13642_stream                    (no IDs)
   ERR wrong number of arguments for 'xdel' command
```

Step 3 is the crux: the empty read is an empty *reply*, not an error. Step 5 is the masked second failure.

**Added a test** — `test_stream_read_with_options_xreadgroup_empty` in `crates/redis_interface/src/test.rs`. It builds the realistic case (append, create the group, drain it) rather than reading a stream that was never written to, and asserts the contract both backends actually satisfy: an exhausted read reports *no work* — as an empty reply or as `StreamEmptyOrNotAvailable` — and never reports entries that are not there.

To be straight about what it does and does not do: it passes both before and after this change, because it covers `redis_interface`, which the fix does not touch. It pins the divergence in the table above so that neither backend can drift into the other's shape unnoticed — the thing that would silently reintroduce this class of bug. The reproduction above is the evidence for the bug itself.

Placed there because that suite already requires a live Redis and runs against both backends; `crates/scheduler` has no Redis-backed test harness, and I did not want to make `cargo test -p scheduler` newly dependent on one. Happy to add one if you would prefer `get_batches` covered directly — that would be the test that goes red without the guard.

**Verified against a live Redis 7:**

| command | result |
|---|---|
| `cargo test -p redis_interface` (redis-rs) | **100 passed, 0 failed** |
| `cargo test -p redis_interface --no-default-features --features fred` | 74 passed, 1 failed |
| `just clippy` | clean |
| `cargo +nightly fmt --all` | clean |

The one fred failure is `test_scan_returns_matching_keys`, which is **pre-existing and unrelated** — I confirmed it by stashing my changes and re-running it on `main`, where it fails identically. It uses a `unique_test_id()`-scoped pattern, so nothing I added can reach it. Flagging it since it looks like a real fred/redis-rs behavioural difference in `scan` rather than a flake, and it may be worth its own issue; happy to open one if that would be useful.

Note that `cargo test` does not run in CI, so the new test will not appear in the checks on this PR — it is a developer-run suite, per the module docs in `test.rs`.

## Checklist


abhisheksharma2411:
Checking why this has been quiet, and it looks mechanical rather than anything to do with the change — so flagging it in case a maintainer can unblock it with two clicks.

**Every substantive workflow on this PR is sitting at `action_required`.** Querying the runs for head `8915f583`:

```
CI-pr                                  action_required   (pull_request)
CI Nix                                 action_required   (pull_request)
CI wasm-check                          action_required   (pull_request)
Validate Generated OpenAPI Spec File   action_required   (pull_request)
Migration Consistency Tests            action_required   (pull_request)
Run Cypress tests                      action_required   (pull_request)
PR Title Spell Check                   action_required   (pull_request)
Pull Request Convention Checks         success           (pull_request_target)
```

The one that ran is the only one on `pull_request_target`, which executes in the base-repo context and so doesn't need approval. Everything on `pull_request` is waiting for a maintainer to hit **Approve and run workflows** — standard GitHub behaviour for a fork PR from someone with no merged PR here yet.

The effect is that **no build, test, clippy, wasm, migration or cypress run has happened on this branch at all**. The 5 green checks showing on the PR are the convention/secret-scanning gates only, which is easy to mistake for a passing CI.

For comparison, the last eight merged PRs here all look like this:

```
#13806  checks=45   labels=['S-test-ready']
#13772  checks=91   labels=['S-test-ready']
#13808  checks=72   labels=['S-test-ready']
#13758  checks=53   labels=['S-test-ready']
#13535  checks=96   labels=['S-test-ready']
```

This PR has **5 checks and no labels**. So there are two things needed, and I can't do either from outside: workflow-run approval, and the `S-test-ready` label that gates `cypress-tests-runner.yml`.

Locally, for what it's worth: `cargo clippy --all-targets --all-features` clean across the workspace, and the new `redis_interface` test passes against both the `redis-rs` and `fred` backends — that divergence was the interesting part of the change and is described in the description.

Happy to rebase onto the latest `main` first if you'd rather approve a fresh run — just say so. And if the right move is for me to close and reopen, or if there's a reviewer I should be tagging instead, I'm glad to be pointed at the correct process.

abhisheksharma2411:
Rebased onto current `main` as offered — this was `mergeable_state: behind` by 27 commits. New head `4e13d00c7`, clean rebase, no conflicts.

Re-verified rather than trusting a clean rebase, since 27 commits of `main` can break a diff that still applies:

```
cargo clippy --all-targets --no-default-features --features <full workspace list>
  exit 0, no warnings from this change
  (one pre-existing future-incompat notice for upstream deps: buf_redux,
   grpc-api-types, multipart, num-bigint-dig, typemap)

cargo test -p redis_interface --features redis-rs   test_stream_read_with_options_xreadgroup_empty ... ok
cargo test -p redis_interface --features fred       test_stream_read_with_options_xreadgroup_empty ... ok
```

Both backends matter here and that's the point of the test: `XREADGROUP` on an exhausted stream returns an **empty reply** on `redis-rs` and **`Err(StreamEmptyOrNotAvailable)`** on `fred`, so the test asserts the union both satisfy. It still holds after the rebase, which is the thing I actually wanted to know.

The scheduler crate has two test functions in total and `get_batches` needs live Redis plus a database, so there is no unit test to run for the changed function itself — the workspace `--all-targets` clippy compiling its test target is the real compile-time gate, and I'd rather say that than imply a test suite covered it.

**The two clicks from the earlier comment are still what this needs**, and nothing has changed on that front in three days: no labels, no reviewer, and all seven `pull_request` workflows still at `conclusion=action_required`. The fresh head should make approving a run straightforward if someone has a moment.

One note on method, in case it's useful to anyone else hitting it: `just` isn't installed here, so I replicated the `clippy` recipe by hand *including its feature computation*. A narrower `cargo test -p scheduler --features redis-rs` fails with spurious unresolved imports — `-p` scopes features to the package, so the workspace feature list is rejected outright. That's presumably why the recipe runs workspace-wide.

hrithikesh026 (approved):

PR #13786, opened 2026-08-19
feat(account_updater): sync stored card before merchant-initiated payments

kanikabansal08:
## Type of Change

- [x] New feature

## Description

Runs the Account Updater against a stored card just before a merchant-initiated payment authorizes, so the payment is attempted with the card as the networks report it now rather than the card on record.

The v1 payments flow sets `force_sync` on the modular payment method retrieve when the request is off-session and draws on a stored payment method:

- `PaymentCreate` — only when the create also confirms (`confirm == true`), since a create that does not authorize will be confirmed separately and synced there.
- `PaymentConfirm` — when the resolved payment method reference is the one named by `recurring_details`. `payment_token` takes priority over `recurring_details` when resolving the reference, so the condition compares against the reference actually being fetched.

Every other caller passes `false`: eligibility checks are not payments, external vault cards are outside the Account Updater's reach, and customer-present flows can ask the cardholder for a new card directly.

Whether the sync actually happens is still governed entirely by the existing configuration — `account_updater.enabled` and `account_updater.credential_source`, both of which default to off. No new configuration is introduced.

### Additional Changes

`force_sync` is now always sent as a query parameter on the internal modular payment method retrieve (previously it was not sent by this client at all). The parameter already existed on the receiving endpoint and defaults to `false`, so this is additive.

## Motivation and Context

Stored cards go stale. They expire, get reissued after fraud, or change PAN while the mandate stays valid. A merchant-initiated payment has no cardholder present to ask for new details, so a stale card is simply a declined payment. This wires the Account Updater — already built in #13510 and #13750 — into the payment path so those payments authorize against current card details.

Stacked on #13750; this PR targets `feat/account-updater-apply-card-updates` and should be merged after it.

local clippy_v2 test:
Screenshot 2026-09-09 at 12 15 49 PM
Screenshot 2026-09-09 at 12 15 37 PM

## Checklist


crates/router/src/core/payments/operations/payment_confirm.rs
```diff
@@ -2539,13 +2539,21 @@ impl PaymentConfirm {
             }
         }
 
+        let is_off_session_mit_for_payment_method = req.off_session == Some(true)
+            && matches!(
+                req.recurring_details.as_ref(),
+                Some(RecurringDetails::PaymentMethodId(payment_method_id))
```
ThisIsMani: If possible let's move this to a function since this is present two times.

crates/api_models/src/payments.rs
```diff
@@ -1712,6 +1712,16 @@ impl PaymentsRequest {
...
     }
 
+    pub fn is_off_session_mit_for_payment_method(&self, payment_method_id: &str) -> bool {
+        self.confirm == Some(true)
+            && self.off_session == Some(true)
+            && matches!(
+                self.recurring_details.as_ref(),
+                Some(RecurringDetails::PaymentMethodId(recurring_payment_method_id))
+                    if recurring_payment_method_id.as_str() == payment_method_id
+            )
```
ThisIsMani: Could use `is_some_and` instead.

crates/api_models/src/payments.rs
```diff
@@ -1712,6 +1712,16 @@ impl PaymentsRequest {
         })
     }
 
+    pub fn is_off_session_mit_for_payment_method(&self, payment_method_id: &str) -> bool {
```
ThisIsMani: Should this be defined in api_models?

ThisIsMani (dismissed):

ShankarSinghC (dismissed):

ThisIsMani (approved):

ShankarSinghC (approved):

jarnura (approved):

PR #13789, opened 2026-08-19
fix(router): report the captured amount on v2 split payment responses

### LINKED ISSUE #13788: Bug: v2 split payment attempts duplicate the order's additional charges, and split responses always report the order as fully captured
## Bug Description

Two amount reporting defects in the V2 split payments path. Both are the same class of problem as
#13712: an amount field that contradicts the fields reported next to it. I found them while
tracing the call sites of `AmountDetails::calculate_net_amount()` for that issue, and they are
independent of it, so I am reporting them separately rather than widening that one.

Line references are against `main` at `182dd4cf5755d90329d24cc08a2d9eab5ea0eaaa`.

### 1. Every split attempt stores the order's additional charges in full

`AmountDetails::create_split_attempt_amount_details` (`crates/hyperswitch_domain_models/src/payments.rs:785`)
sets `net_amount` to that attempt's slice of the order, but copies `shipping_cost`,
`order_tax_amount`, `surcharge_amount` and `tax_on_surcharge` from the intent unchanged:

```rust
let net_amount = split_amount;
...
shipping_cost: self.shipping_cost,
order_tax_amount,
```

The slices are produced in `core/split_payments.rs:151`, where `remaining_to_allocate` starts at
the intent's full net amount and is decremented per payment method, so the slices sum to the order
total. The additional charges do not get split with them. With N payment methods, each charge is
persisted N times, and no attempt's `net_amount` equals its order slice plus the charges stored
beside it.

`AttemptAmountDetails::net_amount` documents itself as "The total amount for this payment attempt.
This includes all the surcharge and tax amounts" (`payment_attempt.rs:249`). For a split attempt it
is a bare slice that includes none of them.

### 2. Split payment responses always report the order as fully captured

`SplitPaymentResponseData::generate_response`
(`crates/router/src/core/payments/transformers.rs:3166`) builds the response as:

```rust
let net_amount = intent_amount_details.calculate_net_amount();
...
amount_captured: Some(net_amount),
```

`amount_captured` is set unconditionally to the intent's full net amount. Nothing gates it on
status or on capture method, and `SplitPaymentResponseData::get_intent_status`
(`core/split_payments.rs:42`) can return `Failed` or a mixed status from the underlying attempts.

Every other V2 response builder reads the persisted value instead
(`transformers.rs:7460` and `transformers.rs:7508`):

```rust
amount_captured: intent_amount_details.amount_captured,
```

## Expected Behavior

1. A split attempt's `net_amount` and the charges stored on that attempt describe the same money,
   so that the group's charges are counted once, not once per payment method.
2. `amount_captured` reflects what was actually captured: absent for a split payment that failed,
   and absent for a manual capture payment sitting at `requires_capture`.

## Actual Behavior

For an order of `order_amount` 10000, `shipping_cost` 500, `order_tax_amount` 800 (net 11300) split
across two gift cards of 6000 and 5300:

| | attempt 1 | attempt 2 | sum over the group |
| --- | --- | --- | --- |
| `net_amount` | 6000 | 5300 | 11300 (correct) |
| `shipping_cost` | 500 | 500 | 1000 |
| `order_tax_amount` | 800 | 800 | 1600 |

And on the confirm response for that payment, `amount_captured` is 11300 whether the split
succeeded, failed, or is awaiting a manual capture.

## Steps To Reproduce

Found by code inspection rather than against a running instance, so I am giving the regression
instead of a curl transcript. Defect 1 is reachable from a unit test, using the same fixture shape
as the tests in #13724:

```rust
#[test]
fn split_attempt_charges_are_not_duplicated_across_the_group() {
    let mut amount_details = amount_details(Some(default_tax(ORDER_TAX_AMOUNT)));
    amount_details.shipping_cost = Some(MinorUnit::new(500));

    let first = amount_details
        .create_split_attempt_amount_details(&confirm_intent_request(None), MinorUnit::new(6000));
    let second = amount_details
        .create_split_attempt_amount_details(&confirm_intent_request(None), MinorUnit::new(5300));

    // The slices add up to the order's net amount.
    assert_eq!(
        first.get_net_amount() + second.get_net_amount(),
        amount_details.calculate_net_amount()
    );

    // The charges that make up that net amount are counted twice.
    assert_eq!(
        first.get_shipping_cost().unwrap_or_default()
            + second.get_shipping_cost().unwrap_or_default(),
        MinorUnit::new(500)
    );
}
```

The second assertion currently fails with `left: 1000, right: 500`, and the same holds for
`get_order_tax_amount`.

## Context For The Bug

Both were introduced with the split payments flow in #9239 (2025-11-17) and predate #13712.

I have not proposed a shape for defect 1 here, because the choice is a real one and I would rather
have a maintainer's view first: the charges could be allocated to the primary attempt only, or
apportioned across the group, or kept on the intent and left off the split attempts entirely. Each
changes what per-attempt reporting means, and the split path currently has no test coverage that
would pin the intent.

Defect 2 looks unambiguous to me: reading `intent_amount_details.amount_captured` there would make
the split path agree with every other V2 response builder.

## Environment

Not a hosted instance.

Operating system: Windows 11 (x86_64)
Rust version: rustc 1.95.0 (59807616e 2026-04-14)
Source version: main at 182dd4cf5755d90329d24cc08a2d9eab5ea0eaaa

Have you spent some time checking if this bug has been raised before?
- I checked and didn't find a similar issue

Have you read the Contributing Guidelines?
- I have read the Contributing Guidelines

Are you willing to submit a PR?
Yes, I am willing to submit a PR, for defect 2 immediately, and for defect 1 once the intended
allocation is settled.

Mohil-Ahuja:
## Type of Change

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

`SplitPaymentResponseData::generate_response` reported `amount_captured` as the order's full net
amount on every split payment confirm response:

```rust
let net_amount = intent_amount_details.calculate_net_amount();
...
amount_captured: Some(net_amount),
```

Nothing gated that on the outcome. `SplitPaymentResponseData::get_intent_status` derives the
intent status from the underlying attempts and can return `Failed`, or a mixed status when the
attempts disagree, and the split path does not restrict the capture method. So a split payment
that failed, or one that is sitting at `requires_capture` awaiting a manual capture, still
reported the whole order as captured.

The captured amount is already tracked correctly. `get_payment_intent_update` accumulates it into
the intent as each attempt of the split is confirmed
(`crates/hyperswitch_domain_models/src/router_data.rs`):

```rust
let amount_captured = self.get_captured_amount(payment_data);
let intent_amount_captured = payment_data
    .payment_intent
    .amount_captured
    .map(|amount| amount + amount_captured.unwrap_or(MinorUnit::zero()))
    .or(amount_captured);
```

That accumulation exists for exactly this case, and the response was discarding it. This PR reads
it instead, which is what every other V2 response builder already does:
`PaymentAmountDetailsResponse::foreign_from` uses `intent_amount_details.amount_captured` in both
of its implementations.

### Additional Changes

The shape of `PaymentAmountDetailsResponse` is unchanged. `amount_captured` is already
`Option<MinorUnit>` and already absent on non split payments that have captured nothing, so
reporting it as absent here is within the documented contract rather than a change to it.

## Motivation and Context

Fixes #13788.

`amount_captured` is settlement accounting. Reporting an order as captured when it failed, or
before a manual capture has happened, is the same failure mode as a payment appearing settled at
authorization, and it misleads any reconciliation built on the field.

## How did you test it?

The behaviour here is one branch in a response builder that needs a `SessionState`, a `Platform`
and a `Profile` to construct, so this is not reachable from a unit test without standing up those
fixtures, and I did not want to add that scaffolding for a single field.

Reviewed by tracing the value instead:

- `amount_captured` on the response is now the persisted `payment_intent.amount_captured`, which
  `get_payment_intent_update` sets from `get_captured_amount` per attempt and accumulates across
  the attempts of the split.
- For a split where nothing captured, that value is `None`, so the field is omitted rather than
  reported as the order total.
- For a fully captured automatic split, each attempt contributes its captured slice, so the
  accumulated total is the order total, which is what the previous hardcoded value happened to be
  right about.

Manual verification against a live split payment would need gift card balances in Redis and a
connector sandbox, which I do not have set up. Happy to add an integration test if the team has a
fixture for the split flow, since it currently has none.

## Checklist


Mohil-Ahuja:
Refreshed against main, checks are green and it still applies cleanly. It is a small v2-only correctness fix, so it should be quick to read. Could someone from the payments side take a look when there is a gap?

PR #13790, opened 2026-08-19
refactor(common_enums): make diesel an optional dependency

shuklatushar226:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [x] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

Makes `diesel` an **optional (default-on)** dependency of `common_enums`, so that
consumers which never touch a database do not have to compile it.

`common_enums` annotates ~113 enums with `#[router_derive::diesel_enum(...)]`,
which expands to diesel `ToSql`/`FromSql`/`AsExpression` impls. Because the
dependency was mandatory, *every* consumer of these enums compiled diesel with
`128-column-tables` — including services that never issue a query.

Three changes:

1. **`crates/common_enums`** — `diesel` becomes `optional = true` behind a
   `diesel` feature that is **on by default**, so hyperswitch resolves exactly as
   before. The `diesel_enum` attributes, the `use diesel::{…}` import, the
   `diesel_exports` module and the handful of hand-written `ToSql`/`FromSql`
   impls (`MerchantCategoryCode`, `PixKey`) are gated with `cfg`/`cfg_attr`.

2. **`crates/common_utils`** — now declares `diesel` with `features = ["postgres"]`.
   It uses `diesel::pg::Pg` in 14 places but previously declared bare
   `diesel = "2.2.10"`; it only compiled because `common_enums` happened to enable
   `postgres` through feature unification. This is a latent under-declaration that
   the gating exposes. It also declares `common_enums` with
   `default-features = false` — this is the opt-out edge for downstream consumers.

3. **`crates/router_derive`** — one line in `[dev-dependencies]`. Please see the
   note below; this is the non-obvious part of the PR.

### Why the `router_derive` dev-dependency line is required

Cargo's resolver v2 resolves features **separately for the proc-macro/host
context**, so this workspace resolves two `diesel` lib units. `diesel_derives` is
a **single shared** proc-macro build carrying the highest tier requested anywhere,
while `diesel` only raises its own recursion limit when it holds the tier itself:

```rust
// diesel-2.2.10/src/lib.rs:219
#![cfg_attr(feature = "128-column-tables", recursion_limit = "256")]
```

`common_enums`'s previously-mandatory diesel was silently holding the 128 tier on
*both* resolution lines. Gating it dropped the proc-macro line to the 32 tier
while the shared derive crate stayed at 128, producing:

```
error: recursion limit reached while expanding `impl_sql_type!`
error: could not compile `diesel` (lib)
```

That line exists only because `router_derive` declares `common_utils` (and
transitively `common_enums`) in `[dev-dependencies]`, which `--all-targets`
builds. This is why `cargo check --workspace` passes but `just clippy` fails.
Pinning the tier there restores agreement. Dev-dependencies are never propagated
to consumers, so it has no effect downstream.

### Additional Changes

This is a compile-time change only — no runtime behaviour, API surface, schema or
config is affected. `payment_attempt` has 92 columns, so hyperswitch still
requires the `128-column-tables` tier and continues to use it.

## Motivation and Context

The Unified Connector Service (`hyperswitch-prism`) contains **zero** diesel code
— no `table!` blocks, no queries, no `diesel` entry in any of its manifests. It
nevertheless spent **119 s of a 234 s cold build (51 %)** compiling diesel, purely
because it depends on hyperswitch's `common_enums` for shared vocabulary and that
crate mandated diesel.

The cost is concentrated in the `128-column-tables` feature, which generates
trait impls for progressively wider tuples. Measured by compiling diesel in
isolation with nothing else in the dependency graph:

| feature | diesel compile time |
| --- | --- |
| `32-column-tables` | 6.95 s |
| `64-column-tables` | 23.15 s |
| `128-column-tables` | **115.64 s** |

### hyperswitch (this PR)

All run on this branch, rebased on `main`:

- `just clippy`, `just clippy fred`, `just clippy_v2`
- `cargo check --workspace`, `cargo check --features "release"`
- `cargo build --package router --bin router`; `cargo build --package drainer --no-default-features --features "redis-rs,v1"`
- `cargo build --package router --bin router --no-default-features --features "release,v2,redis-rs"` (full v2 build)
- `cargo check -p euclid_wasm --target wasm32-unknown-unknown --features "dummy_connector,v1"` (wasm path)
- `cargo hack check --each-feature --all-targets --exclude-features 'v2 payment_v2' -p common_enums -p common_utils`
- `cargo run -p openapi --features v1` and `--features v2`
- `cargo +nightly fmt --all --check` (clean for the files touched here)
- `cargo check -p common_enums --no-default-features` (diesel gated **off**) and
  `--features diesel` (**on**) — both sides of the gate compile

Resolution is verified structurally rather than by inspection, using
`cargo +nightly build --unit-graph -Z unstable-options --all-targets`: both
`diesel` lib units and the shared `diesel_derives` unit carry
`128-column-tables`, identical to a clean-`main` baseline captured before the
change.

### UCS (`hyperswitch-prism`)

`hyperswitch-prism` pins hyperswitch by git tag, so it was tested by pointing it
at this branch with cargo `[patch]` overrides — **prism's own tree was never
modified**:

```bash
cargo --config 'patch."https://github.com/juspay/hyperswitch".common_enums.path="…/crates/common_enums"' \
      --config 'patch."https://github.com/juspay/hyperswitch".common_utils.path="…/crates/common_utils"' \
      build --timings -p grpc-server
```

Verified first that the dependency actually changed shape:

- `cargo tree -e features -p grpc-server` reports only `32-column-tables`
  (previously `128-column-tables`)
- `cargo tree -i diesel -p grpc-server` no longer lists `common_enums` among
  diesel's parents

Then measured a cold build into a fresh target directory:

| metric | before | after |
| --- | --- | --- |
| cold build wall clock | 234 s | **137.7 s** (−41 %) |
| `diesel` compile unit | 118.8 s | **16.1 s** |
| peak RSS | 9.7 GB | **~5.1 GB** |

Measured with `cargo build --timings` on 24 cores / rustc 1.96.0, builds run
strictly serially so they could not contend with one another.

### Note on rollout

UCS only benefits once this is merged and tagged, and prism bumps its pin from
`2026.06.18.0`. **Both manifest changes must ship in the same tag** — the current
`common_utils` relies on `common_enums` supplying diesel's `postgres` feature via
unification, so pairing a new `common_enums` with an old `common_utils` will not
build.

## Checklist

No new unit tests: this is compile-time feature gating with no runtime behaviour
to assert. The meaningful assertions are that both sides of the gate compile and
that the resolved dependency graph is unchanged for hyperswitch — both covered
above.

### DIFF
--- a/crates/common_enums/Cargo.toml
+++ b/crates/common_enums/Cargo.toml
@@ -8,13 +8,22 @@ readme = "README.md"
 license.workspace = true
 
 [features]
+# Diesel ORM impls for these enums. Default-on, so hyperswitch resolves exactly
+# as before. Consumers that never touch a database opt out via
+# `default-features = false` (see common_utils), which is their only path here.
+# Keeping it default-on also keeps diesel on the 128-column tier in every
+# dependency context: diesel_derives is a single shared proc-macro build
+# carrying the highest tier, while diesel only raises its own recursion_limit
+# when it holds 128-column-tables. A mixed 32/128 split fails to compile.
+default = ["diesel"]
+diesel = ["dep:diesel"]
 dummy_connector = []
 openapi = []
 payouts = []
 v2 = []
 
 [dependencies]
-diesel = { version = "2.2.10", features = ["postgres", "128-column-tables"] }
+diesel = { version = "2.2.10", features = ["postgres", "128-column-tables"], optional = true }
 serde = { version = "1.0.219", features = ["derive"] }
 serde_json = "1.0.140"
 strum = { version = "0.26", features = ["derive"] }
--- a/crates/router_derive/Cargo.toml
+++ b/crates/router_derive/Cargo.toml
@@ -27,7 +27,13 @@ serde_json = "1.0.140"
 url = { version = "2.5.4", features = ["serde"] }
 utoipa = "4.2.3"
 
+# common_utils (and transitively common_enums) is a dev-dependency only, used by
+# this crate's own tests. Because --all-targets builds those tests, that chain
+# drags diesel onto the proc-macro/host resolution line; pin the 128-column tier
+# here so it matches the shared diesel_derives build. Dev-dependencies are never
+# propagated to consumers, so this does not affect downstream users.
 common_utils = { version = "0.1.0", path = "../common_utils" }
+common_enums = { version = "0.1.0", path = "../common_enums", features = ["diesel"] }
 
 [lints]
 workspace = true
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8395,6 +8395,7 @@ dependencies = [
 name = "router_derive"
 version = "0.1.0"
 dependencies = [
+ "common_enums",
  "common_utils",
  "diesel",
  "error-stack 0.4.1",
--- a/crates/common_enums/src/connector_enums.rs
+++ b/crates/common_enums/src/connector_enums.rs
@@ -23,7 +23,7 @@ pub use crate::PaymentMethodType;
     Hash,
     SmithyModel,
 )]
-#[router_derive::diesel_enum(storage_type = "text")]
+#[cfg_attr(feature = "diesel", router_derive::diesel_enum(storage_type = "text"))]
 #[serde(rename_all = "snake_case")]
 #[strum(serialize_all = "snake_case")]
 #[smithy(namespace = "com.hyperswitch.smithy.types")]
@@ -519,7 +519,7 @@ impl Connector {
     strum::EnumString,
     ToSchema,
 )]
-#[router_derive::diesel_enum(storage_type = "text")]
+#[cfg_attr(feature = "diesel", router_derive::diesel_enum(storage_type = "text"))]
 #[serde(rename_all = "snake_case")]
 #[strum(serialize_all = "snake_case")]
 pub enum InvoiceStatus {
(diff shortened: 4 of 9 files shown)


crates/router_derive/Cargo.toml
```diff
@@ -27,7 +27,13 @@ serde_json = "1.0.140"
 url = { version = "2.5.4", features = ["serde"] }
 utoipa = "4.2.3"
 
+# common_utils (and transitively common_enums) is a dev-dependency only, used by
+# this crate's own tests. Because --all-targets builds those tests, that chain
+# drags diesel onto the proc-macro/host resolution line; pin the 128-column tier
+# here so it matches the shared diesel_derives build. Dev-dependencies are never
+# propagated to consumers, so this does not affect downstream users.
 common_utils = { version = "0.1.0", path = "../common_utils" }
+common_enums = { version = "0.1.0", path = "../common_enums", features = ["diesel"] }
```
SanchithHegde, line 36: Why was this dependency added in this PR when it didn't exist previously?

crates/common_enums/Cargo.toml
```diff
@@ -8,13 +8,22 @@ readme = "README.md"
 license.workspace = true
 
 [features]
+# Diesel ORM impls for these enums. Default-on, so hyperswitch resolves exactly
```
SanchithHegde, line 11: Nit: Can probably get rid of the over-explanatory comment.

SanchithHegde (commented):

PR #13793, opened 2026-08-20
feat(revenue_recovery): Add support for storing retry stats in DB

AnuthaDev:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [x] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

Adds persistence of revenue-recovery retry-outcome statistics into a new revenue_recovery_retry_stats table, so that for every failed invoice we capture how retries against a given failure signature fared (success vs. failure), bucketed over time. This is the data foundation the retry math-model will read to decide when to retry.

Cluster key. Each outcome is recorded under a strictly-typed cluster key:

`v1|{standardised_error_code}/{card_type}/{issuer}`

- error_code → `common_enums::StandardisedCode`, `card_type → common_enums::CardType` (credit/debit), issuer → free-text bank name. The typed Dim<T> prevents putting a value in the wrong dimension at construction. * wildcards mark ancestor (root/mid) roll-up nodes; UNK marks an unresolved dim.
- Constructors for the retry algorithm (new, root_from_error_code) plus a reversible DB codec (as_db_string/from_db_string) with percent-escaping.

Where dimensions come from. All three dims are read from the previous (triggering) attempt, so the record is keyed by the error that motivated the retry, not the retry's own new error:
- error code: persisted standardised_code preferred, else a live GSM lookup on the Payment/Authorize flow (same key the attempt path persists under);
- card funding type: chargebee webhook funding_type (previously dropped in chargebee/transformers.rs, now populated);
- issuer: ISIN → cards_info.card_issuer lookup.

Recording points. A RetryOutcomeEvent is recorded when a retry's fate becomes known — internal execute response, and PSync resolution (success/failure). Recording only happens when a previous attempt exists (a retry outcome is meaningless without one).

Plumbing. prev_attempt_id is stamped on the CALCULATE tracking data and carried verbatim through EXECUTE → PSYNC, so recorders source the triggering attempt without extra process-tracker round-trips.

Write path. record.rs does a per-cluster-key Redis-locked read-merge-write of a JSONB distribution document (document.rs), single-row fetch on the write primary to avoid replica-lag duplicate-key races. Best-effort: failures are logged and swallowed, never affecting the payment/recovery flow.

### Additional Changes

## Motivation and Context

Revenue-recovery retries were fired without any historical signal on how a given failure signature responds to retrying. To drive a data-informed retry schedule (the math-model), we first need to accumulate per-signature success/failure distributions over time. This PR builds that recording layer end-to-end.

## How did you test it?

- Unit tests for the cluster-key codec (retry_stats_cluster_key::tests): leaf/root round-trips, root/mid/leaf chain(), issuer percent-escaping, foreign-version rejection, reserved-spelling normalization — all passing (cargo test -p hyperswitch_domain_models --no-default-features --features "v2,revenue_recovery" retry_stats_cluster_key).
- Full v2 build/lint passes (just clippy_v2, -D warnings).

## Checklist


crates/hyperswitch_domain_models/src/revenue_recovery/retry_stats_cluster_key.rs
```diff
@@ -0,0 +1,322 @@
...
+    pub issuer: Dim<String>,
+}
+
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub enum NodeDepth {
+    Root,
+    Mid,
+    Leaf,
+}
+
```
NISHANTH1221: This will be insignificant if we have more dimensions

crates/router/src/core/revenue_recovery/retry_stats/events.rs
```diff
@@ -0,0 +1,160 @@
...
+        let card_details = revenue_recovery_metadata
+            .billing_connector_payment_method_details
+            .as_ref()
+            .and_then(|details| details.get_billing_connector_card_info());
+
+        // `card_network` is only needed for the live GSM lookup; the middle dimension is
+        // the card funding type (credit/debit), taken from the revenue recovery
+        // metadata's payment method subtype.
+        let card_network = card_details.and_then(|card| card.card_network.clone());
+
```
NISHANTH1221: `card_network` can also be taken from payment_attempt's payment_method_data

XyneSpaces:
[blocking] Please coordinate this migration with #13804/#13802 before merging: this PR creates `revenue_recovery_retry_stats` with a `stats` JSONB column while those PRs create the same table with a `distribution` JSONB column. Whichever lands second will have Diesel models expecting a column that the already-applied migration did not create, so the table schema/model needs to be unified across the open retry-stats PRs.

crates/diesel_models/src/revenue_recovery_retry_stats.rs
```diff
@@ -0,0 +1,20 @@
...
+use crate::schema_v2::revenue_recovery_retry_stats;
+
+#[derive(
+    Clone, Debug, Queryable, Identifiable, Selectable, serde::Serialize, serde::Deserialize,
+)]
+#[diesel(table_name = revenue_recovery_retry_stats, check_for_backend(diesel::pg::Pg))]
+#[diesel(primary_key(cluster_key))]
+pub struct RevenueRecoveryRetryStats {
+    pub cluster_key: String,
+    pub stats: serde_json::Value,
```
jarnura: untyped value should be secret.

crates/diesel_models/src/revenue_recovery_retry_stats.rs
```diff
@@ -0,0 +1,20 @@
...
+}
+
+#[derive(Clone, Debug, diesel::Insertable, serde::Serialize, serde::Deserialize)]
+#[diesel(table_name = revenue_recovery_retry_stats)]
+pub struct RevenueRecoveryRetryStatsNew {
+    pub cluster_key: String,
+    pub stats: serde_json::Value,
```
jarnura: untyped value should be secret.

jarnura (changes requested):

crates/hyperswitch_domain_models/src/revenue_recovery/retry_stats_cluster_key.rs
```diff
@@ -0,0 +1,337 @@
...
+/// misreading them.
+pub const RETRY_STATS_KEY_VERSION: &str = "v1";
+pub const KEY_SEPARATOR: char = '|';
+pub const SEGMENT_SEPARATOR: char = '/';
+pub const WILDCARD_SEGMENT: &str = "*";
+pub const UNKNOWN_SEGMENT: &str = "UNK";
+/// Number of dimension segments in a key: `error_code` / `card_type` / `issuer`.
+/// Adding or removing a dimension is a layout change — bump
+/// [`RETRY_STATS_KEY_VERSION`] together with this count.
+const SEGMENT_COUNT: usize = 3;
```
jarnura: Why this is a hard-coded value?

jarnura (changes requested):

crates/router/src/core/revenue_recovery/retry_stats/events.rs
```diff
@@ -0,0 +1,177 @@
+use api_models::payments::PaymentRevenueRecoveryMetadata;
+// NOTE: `PaymentMethodType` is dropped from this import while the card_type dimension is
```
jarnura: remove comments like these, in other places too

jarnura (changes requested):

crates/hyperswitch_domain_models/src/revenue_recovery/retry_stats_cluster_key.rs
AnuthaDev: Moved to a macro based approach

crates/router/src/core/revenue_recovery.rs
```diff
@@ -188,6 +190,29 @@ pub async fn record_internal_attempt_and_execute_payment(
 ) -> Result<(), sch_errors::ProcessTrackerError> {
     let db = &*state.store;
 
+    // The attempt whose failure triggered this execute task — used by the
+    // retry-stats recorder for cluster dimensions (NOT the new attempt that
+    // `record_internal_attempt_api` is about to create). Absent it, there is no
+    // prior outcome to attribute this retry to and retry-stats are not recorded.
+    let prev_attempt = match tracking_data.prev_attempt_id.as_ref() {
+        Some(prev_attempt_id) => db
+            .find_payment_attempt_by_id(
```
jarnura: Revenue recovery system should not read anything directly from database for payments related things

jarnura (changes requested):

crates/router/src/core/revenue_recovery.rs
```diff
@@ -508,6 +540,32 @@ pub async fn perform_payments_sync(
...
     payment_intent: &PaymentIntent,
 ) -> Result<(), errors::ProcessTrackerError> {
+    let db = &*state.store;
+
+    // The failed attempt that motivated this retry chain. Once the Psync returns
+    // a terminal status, the error details of this attempt are used to update the
+    // retry_stats
+    let prev_attempt = match tracking_data.prev_attempt_id.as_ref() {
+        Some(prev_attempt_id) => db
+            .find_payment_attempt_by_id(
```
jarnura: Revenue recovery system should not read anything directly from database for payments

crates/router/src/core/revenue_recovery/retry_stats/record.rs
```diff
@@ -0,0 +1,199 @@
...
+};
+
+use super::events::RetryOutcomeEvent;
+use crate::{
+    core::configs::dimension_state,
+    routes::{app::SessionStateInfo, SessionState},
+};
+
+/// Short TTL on the per-key lock so a crashed writer never wedges a key.
+const LOCK_TTL_SECONDS: i64 = 10;
```
jarnura: What is this lock? Why it used? Move all the configs to env and don't hard code in code

jarnura (changes requested):

crates/router/src/core/revenue_recovery.rs
AnuthaDev: It was reading the error code of the previous attempt. Moved to storing the error code in process tracker data.

crates/router/src/core/revenue_recovery/retry_stats/record.rs
```diff
@@ -0,0 +1,202 @@
...
+        {
+            Ok(SetnxReply::KeySet) => {
+                acquired = true;
+                break;
+            }
+            Ok(SetnxReply::KeyNotSet) => {
+                actix_web::rt::time::sleep(wait_duration).await;
+            }
+            Err(error) => {
+                return Err(error).change_context(StorageError::KVError);
```
jarnura:
```suggestion
                Err(error).change_context(StorageError::KVError)?;
```

crates/router/src/core/revenue_recovery/retry_stats/record.rs
```diff
@@ -0,0 +1,202 @@
...
+    }
+
+    if !acquired {
+        // Lock was still contended after the retry budget; dropping this update is
+        // acceptable for best-effort stats.
+        logger::warn!(
+            cluster_key = %db_key,
+            "revenue_recovery_retry_stats: lock contended after retries, skipping this update"
+        );
+        return Ok(());
```
jarnura: Avoid return based code, use valid if else branches

crates/router/src/core/revenue_recovery/retry_stats/record.rs
```diff
@@ -0,0 +1,202 @@
...
+                    "revenue_recovery_retry_stats: unable to acquire redis connection"
+                );
+                return;
+            }
+        };
+        let store = state.store.get_revenue_recovery_retry_stats_store();
+
+        let db_key = self.key.as_db_string();
+
+        let lock_settings = state.conf().lock_settings;
```
jarnura: lock_settings is two generic and expressing the use-case, use a key which express the use-case and not a generic term

jarnura (changes requested):

jarnura (dismissed):

jarnura (approved):

NISHANTH1221 (approved):

PR #13802, opened 2026-08-20
feat(revenue-recovery) - add adaptive retry scheduling for revenue recovery

rohanm13:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [x] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

Adds an **adaptive retry scheduling** path to revenue recovery, behind a Superposition flag
that defaults to off.

On each scheduling decision the static cascading ladder and an adaptive algorithm are both
consulted. The **earlier calendar day wins, ties go to static**. A static ladder position is
consumed only when its time is actually used — if the adaptive time wins, the same position
is offered again next time, so inserted adaptive retries never exhaust the ladder early.

```
Smart arm
 ├─ adaptive_retry_enabled == false → existing decider path, unchanged
 └─ true →
      queried_rung  = recovery_schedule.next_rung()      // static_rung + 1
      static_time   = cascading ladder at queried_rung
      adaptive_time = adaptive algorithm
      decision      = earlier day wins, ties → static
      token gate    = same availability check cascading uses
      persist rung  only when a retry is actually scheduled
```

**Files**

| File | Change |
|---|---|
| `core/revenue_recovery/schedule.rs` | **New.** `RecoverySchedule`, `decide_next_retry`. Pure — no state, no I/O, no clock read |
| `workflows/revenue_recovery.rs` | `Smart` arm branches on the flag; `get_adaptive_retry_time`; token-availability block extracted into a shared helper |
| `core/revenue_recovery.rs` | Call site, `finish_calculate_workflow_with_schedule`, tracking-data construction sites |
| `core/revenue_recovery/types.rs` | `reopen_...` now destructures the tracking data exhaustively |
| `types/storage/revenue_recovery.rs` | `recovery_schedule` field on the workflow tracking data |
| `consts.rs`, `core/configs/dimension_config.rs` | Three Superposition config keys |

**Notes for reviewers**

- **`Cascading` is touched.** Its token-availability block moved into
  `get_token_availability_for_schedule_time`, now shared with the `Smart` arm. Same order,
  conditions, precedence and error mapping — the only behavioural difference is a log string
  (`"…for cascading retry"` → `"…for retry"`).
- **`static_rung` is deliberately not `process_tracker.retry_count`.** `retry_count` counts
  every attempt including inserted adaptive ones; feeding it to a short ladder runs off the
  end early and strands the invoice.
- **The exhaustive destructure in `reopen_calculate_workflow_on_payment_failure` is
  intentional.** That row is rebuilt on every failure, so anything not carried across is wiped
  exactly when it matters. It already earned its keep — it forced `prev_attempt_id` to be
  handled explicitly when that field landed, instead of silently defaulting to `None`.
- **This is currently a no-op.** The scoring function is owned separately and does not exist
  yet, so `get_adaptive_retry_time` returns `None` and every decision resolves to the static
  ladder. Even with the flag on, schedule times are identical to `Cascading`.

### Additional Changes

Three new Superposition keys, resolved on `processor_merchant_id` + `connector`, all with
database fallback:

| Key | Type | Default |
|---|---|---|
| `revenue_recovery.adaptive_retry_enabled` | `bool` | `false` |
| `revenue_recovery.grace_period_days` | `i64` | `30` |
| `revenue_recovery.max_retry_count` | `i64` | `15` |

Defined in `crates/router/src/consts.rs` and
`crates/router/src/core/configs/dimension_config.rs`. No `config/*.toml` changes.

**No schema change** — the scheduling state rides inside the existing
`process_tracker.tracking_data` JSONB, so there is no migration. `#[serde(default)]` means
rows written before this field still deserialize.

## Motivation and Context

Closes #

Revenue recovery retries every invoice on the same fixed ladder, regardless of why it
declined, what card it is, or when that card has historically succeeded. This adds the
scheduling half of an adaptive algorithm that can propose a better date per invoice, while
keeping the static ladder as a floor so an unproven model cannot make recovery worse.

Gating per merchant and connector, defaulting to off, means it can be piloted and rolled back
without a deploy.

## How did you test it?

**Unit tests** — 17 passing in `core::revenue_recovery`, 10 of them new, covering the whole
decision matrix:

```
cargo test --offline --no-default-features --features "v2,redis-rs" \
  -p router --lib core::revenue_recovery
```

The two carrying most weight:

- `stored_rung_advances_by_one` — the ladder index follows `static_rung`, not `retry_count`
- `a_rung_survives_repeated_adaptive_wins_and_is_spent_once` — adaptive → adaptive → static,
  three attempts, exactly one static position consumed

`decide_next_retry` is a pure function of its arguments, so the matrix is covered without a
scheduler or a database.

**Not yet tested end to end.** The adaptive path cannot produce a time until the scoring
function exists, and with the flag off nothing changes. A manual test plan (config setup,
side-by-side `Cascading` vs `Smart` comparison, assertions) is written up separately.

## Checklist

### DIFF
--- a/crates/diesel_models/src/lib.rs
+++ b/crates/diesel_models/src/lib.rs
@@ -7,6 +7,8 @@ pub mod capture;
 pub mod card_issuer;
 pub mod cards_info;
 pub mod configs;
+#[cfg(feature = "v2")]
+pub mod revenue_recovery_retry_stats;
 
 pub mod authentication;
 pub mod authorization;
--- a/crates/diesel_models/src/payment_attempt.rs
+++ b/crates/diesel_models/src/payment_attempt.rs
@@ -1064,6 +1064,7 @@ pub struct PaymentAttemptUpdateInternal {
     pub network_error_message: Option<String>,
     pub connector_request_reference_id: Option<String>,
     pub amount_captured: Option<MinorUnit>,
+    pub error_details: Option<ErrorDetails>,
 }
 #[cfg(feature = "v1")]
 #[derive(Clone, Debug, AsChangeset, router_derive::DebugAsDisplay)]
@@ -1260,6 +1261,7 @@ impl From<PaymentAttemptUpdate> for PaymentAttemptUpdateInternal {
                     cancellation_reason: None,
                     amount_captured,
                     payment_method_data: None,
+                    error_details: None,
                 }
             }
             PaymentAttemptUpdate::ErrorUpdate {
@@ -1312,6 +1314,7 @@ impl From<PaymentAttemptUpdate> for PaymentAttemptUpdateInternal {
                     cancellation_reason: None,
                     amount_captured,
                     payment_method_data: None,
+                    error_details: None,
                 }
             }
             PaymentAttemptUpdate::UnresolvedResponseUpdate {
@@ -1360,6 +1363,7 @@ impl From<PaymentAttemptUpdate> for PaymentAttemptUpdateInternal {
                     cancellation_reason: None,
                     amount_captured: None,
                     payment_method_data: None,
+                    error_details: None,
                 }
             }
             PaymentAttemptUpdate::PreprocessingUpdate {
@@ -1406,6 +1410,7 @@ impl From<PaymentAttemptUpdate> for PaymentAttemptUpdateInternal {
                     cancellation_reason: None,
                     amount_captured: None,
                     payment_method_data: None,
+                    error_details: None,
                 }
             }
             PaymentAttemptUpdate::ConnectorResponse {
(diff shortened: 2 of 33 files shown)


XyneSpaces:
[blocking] Please coordinate this migration with #13793 before merging: this PR creates `revenue_recovery_retry_stats` with a `distribution` JSONB column while #13793 creates the same table with a `stats` JSONB column. Whichever lands second will have Diesel models expecting a column that the already-applied migration did not create, so the table schema/model needs to be unified across the open retry-stats PRs.

PR #13804, opened 2026-08-20
feat(revenue_recovery): mathmodel retry-time predictor

### LINKED ISSUE #13805: [FEATURE] Data-driven retry-time prediction for Revenue Recovery (MathModel arm of Cascading)
Revenue Recovery currently reschedules failed subscription (MIT) retries on a **fixed static
schedule** (e.g. retry on days 1, 3, 8, 10, 12 after the failure). That schedule is identical for
every failure — it does not adapt to *why* the payment failed or *when* that kind of failure has
historically been most likely to recover. As a result, retries land at times that aren't
necessarily optimal, wasting scarce retry attempts (and, for some declines, risking escalation to
a hard decline).

The request is to make the retry **timing** data-driven: for each failure cluster, place the
retry on the day-of-week, day-of-month, and hour-of-day that have historically had the highest
recovery rate for *that* cluster, instead of a one-size-fits-all schedule. Requirements:

- The model should only ever pull a retry **sooner** than the static schedule, never later — a
  safe augmentation of the existing schedule, not a replacement.
- It should degrade gracefully to neutral (uniform) behaviour on sparse or missing data, and
  explore under-observed slots rather than committing to a fabricated estimate.
- It should respect the invoice's remaining retry budget and grace period.

Outcome: a higher recovery rate per retry attempt, with the static schedule retained as a safety
floor.

balakrishnak-ctrl:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [x] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

Implements the **MathModel** half of the Cascading revenue-recovery retry strategy —
a data-driven predictor of *when* to retry a failed subscription (MIT) payment, based on
a cluster's day-of-week / day-of-month / hour-of-day success statistics
(the dense-array `hyperswitch_domain_models::revenue_recovery::retry_stats_document::StatsDocument`).

Two functions in `crates/router/src/workflows/revenue_recovery.rs`:

**1. `compute_mathmodel_retry_time(stats: &StatsDocument, budget: u32, grace_days: u32) -> Option<OffsetDateTime>`** — the model.
- **Scoring:** per-slot confidence via Laplace `p̂ = (k+1)/(n+2)`, standard error, pairwise `C_ij = Φ((p̂ᵢ-p̂ⱼ)/√(ΣSE²))`, score = mean `logit(C_ij)`. `Φ` uses `libm::erf`.
- **Combine:** softmax each of the dow/dom axes over the candidate days, then `max` per day.
- **Pick:** budget-aware per-tick probabilistic day + hour selection with a runway guard (`PickDriver` attributes each pick as weights / runway_guard / exhausted for a future back-test).
- **Restricted exploration:** `n=0` (never attempted) and corrupt (`k>n`) slots are excluded from scoring so they can't hijack the pick; corrupt slots are debug-logged.
- **Window:** starts the *next* day (failure day excluded). `grace_days` counts the failure day, so the retriable window is `grace_days − 1` future days, capped at 31. Returns `None` when `grace_days ≤ 1` (v1 — grace-1-with-retries is a follow-up).

**2. `compute_adaptive_retry_time(state, error_code, remaining_grace_days, remaining_budget) -> Option<OffsetDateTime>`** — the seam between resolved retry inputs and the model. Fetches the cluster's stats for the error code (`find_revenue_recovery_retry_stats_by_error_code`) and hands them to the model. Returns `None` on any "no opinion" case (no history, lookup/parse failure, or the model declining) so the caller always falls back to the static ladder. Stays in explicit UTC.

Adds `libm` as a direct dependency of `router` (already present transitively) for a full-precision `erf`. Inline unit tests (`mod mathmodel_retry_time_tests`).

**Not in this PR:** the caller-side wiring in the Cascading branch — resolving grace/budget/error code and invoking `compute_adaptive_retry_time`, then `min()`-ing the UTC result against the static Superposition schedule — is a follow-up.

### Additional Changes

## Motivation and Context

Revenue Recovery currently reschedules failed MIT retries on a fixed static schedule.
To improve recovery rate we need a data-driven predictor that places each retry on the
day/hour most likely to succeed for that failure's cluster, while only ever pulling a
retry *sooner* than the static schedule (the caller `min()`s the two). This PR provides
that predictor (the MathModel arm of Cascading) and the fetch-by-error-code plumbing that
feeds it. It is self-contained and side-effect free; behaviour on sparse/corrupt data
degrades safely to neutral (uniform) rather than being steered by garbage.

## How did you test it?

- Inline unit tests (`mod mathmodel_retry_time_tests`) covering the scoring math
  (hand-checked values), the Monday=0 indexing contract, the next-day window, the grace
  cap, `grace ≤ 1 → None`, corrupt/`n=0` exclusion, the all-corrupt "still retries"
  invariant, hour fallback, and runway-guard attribution.
- `just clippy_v2` passes clean (no warnings/errors in the touched files).

## Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8437,6 +8437,7 @@ dependencies = [
  "josekit",
  "jsonwebtoken",
  "kgraph_utils",
+ "libm",
  "maud",
  "mimalloc",
  "mime 0.3.17",
--- a/config/config.example.toml
+++ b/config/config.example.toml
@@ -1485,6 +1485,7 @@ billing_connectors_which_requires_invoice_sync_call = "recurly" # List of billin
 monitoring_threshold_in_seconds = 60  # 60 secs , threshold for monitoring the retry system
 retry_algorithm_type = "cascading"  # type of retry algorithm
 redis_ttl_in_seconds=3888000  # ttl for redis for storing payment processor token details
+default_retry_hour_utc = 12  # fallback hour-of-day (UTC, 0-23) the mathmodel retries at when a cluster has no hour-of-day history
 
 # Card specific configuration for Revenue Recovery
 [revenue_recovery.card_config.amex]
--- a/config/deployments/env_specific.toml
+++ b/config/deployments/env_specific.toml
@@ -477,6 +477,7 @@ ucs_psync_disabled_connectors = "cashtocode, trustly"    # Comma-separated list 
 monitoring_threshold_in_seconds = 10368000
 retry_algorithm_type = "cascading"
 redis_ttl_in_seconds=3888000
+default_retry_hour_utc = 12 # fallback hour-of-day (UTC, 0-23) the mathmodel retries at when a cluster has no hour-of-day history
 
 # Card specific configuration for Revenue Recovery
 [revenue_recovery.card_config.amex]
--- a/config/development.toml
+++ b/config/development.toml
@@ -1598,6 +1598,7 @@ ucs_psync_disabled_connectors = "cashtocode, trustly"    # Comma-separated list 
 monitoring_threshold_in_seconds = 60
 retry_algorithm_type = "cascading"
 redis_ttl_in_seconds=3888000
+default_retry_hour_utc = 12 # fallback hour-of-day (UTC, 0-23) the mathmodel retries at when a cluster has no hour-of-day history
 
 [revenue_recovery.recovery_timestamp]
 initial_timestamp_in_seconds = 3600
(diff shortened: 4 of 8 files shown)


XyneSpaces:
[blocking] Please coordinate this migration with #13793 before merging: this PR creates `revenue_recovery_retry_stats` with a `distribution` JSONB column while #13793 creates the same table with a `stats` JSONB column. Whichever lands second will have Diesel models expecting a column that the already-applied migration did not create, so the table schema/model needs to be unified across the open retry-stats PRs.

jarnura (approved):

NISHANTH1221 (dismissed):

jarnura (approved):

jarnura (dismissed):

jarnura (approved):

NISHANTH1221 (approved):

PR #13810, opened 2026-08-20
feat(connector): [Airwallex] support external network token pass-through

### LINKED ISSUE #13600: [FEATURE] Airwallex: support NetworkToken payment method data (external network token pass-through)
### Feature Description

The Airwallex connector does not map PaymentMethodData::NetworkToken — it falls into the catch-all NotImplemented arm in crates/hyperswitch_connectors/src/connectors/airwallex/transformers.rs (L613–623 at v1.125.0), and airwallex is absent from [network_tokenization_supported_connectors] (config/config.example.toml L1305).

Airwallex's API supports externally-provisioned network tokens natively (number_type: EXTERNAL_NETWORK_TOKEN + cryptogram): https://www.airwallex.com/docs/payments/optimize-your-integration/network-tokenization — so this appears to be a transformer mapping gap rather than a PSP capability gap. The Adyen connector's AdyenNetworkTokenData mapping looks like the reference pattern.

Use case: self-hosted OSS deployment, externally-provisioned network tokens (own TSP/TRID via a PCI-proxy vendor, not the Juspay NT service) submitted through S2S confirm with payment_method_data.network_token, routed to an Airwallex MCA.

Related: extended_authentication_applicable is currently gated on payment_method == Card in the Airwallex transformer — worth deciding whether extended authorization should compose with NT payments as part of this mapping.

Confirmed as not-yet-implemented by a Hyperswitch contributor in the community Slack (2026-08-07), who suggested filing this issue.

### Possible Implementation

Follow the existing Adyen pattern (AdyenNetworkTokenData in crates/hyperswitch_connectors/src/connectors/adyen/transformers.rs), applied to the Airwallex transformer:

In crates/hyperswitch_connectors/src/connectors/airwallex/transformers.rs, add a match arm for PaymentMethodData::NetworkToken(token_data) in the payment-request builder (currently falls into the NotImplemented catch-all at L613–623), mapping to Airwallex's external network token API shape:
payment_method.card.number = the network token PAN, with number_type: EXTERNAL_NETWORK_TOKEN
payment_method.card.expiry_month / expiry_year = token expiry
payment_method_options.card.cryptogram = token_data.token_cryptogram (per-transaction TAVV)
Ref: https://www.airwallex.com/docs/payments/optimize-your-integration/network-tokenization
Add airwallex to [network_tokenization_supported_connectors] connector_list in the config (config/config.example.toml L1305 and deployment configs).
Decide whether extended_authentication_applicable (currently gated on payment_method == Card in the same transformer) should also apply to NT payments — Airwallex docs don't state whether extended pre-auth windows compose with external network tokens, so this may need confirmation with Airwallex; at minimum the gating decision should be explicit rather than an accident of the match arm.
No core/router changes expected — payment_method_data.network_token is already accepted by the v1 payments API and routed to connectors; this is purely a connector-crate mapping gap.

No, I don't have time to work on this right now

abhisheksharma2411:
## Type of Change

- [x] New feature

## Description

Maps `PaymentMethodData::NetworkToken` in the Airwallex transformer, where it previously fell into the catch-all `NotImplemented` arm, and adds `airwallex` to `network_tokenization_supported_connectors`.

The token travels in Airwallex's card object with `number_type: EXTERNAL_NETWORK_TOKEN` and no `cvc`; the per-transaction TAVV cryptogram rides in `payment_method_options.card.cryptogram`. New `AirwallexNetworkTokenCard` / `AirwallexNetworkTokenDetails` structs rather than widening `AirwallexCardDetails`, so the raw-card path is untouched — making `cvc` optional there would have started emitting `"cvc": null` for ordinary card payments.

### The extended-authorization question, and why I changed my answer

I said on the issue that I'd gate `extended_authentication_applicable` *off* for network tokens, since Airwallex doesn't document whether extended pre-auth composes with an external token. Writing it, that turns out to be the wrong call, for a reason I'd missed:

`build_airwallex_connector_response_data` reports extended authentication as applied for every `PaymentMethod::Card`, and it **cannot** tell a network token apart — it's `impl<F, T>` over the request type, so `payment_method_data` isn't reachable there without adding a trait bound. Sending `authorization_type: None` while that path reports `applied: true` would make the response contradict the request.

So both ends now agree: NT gets the same `authorization_type` treatment as a raw card, with a comment recording that the two sites have to change together, and that the response path needs the payment method data threaded in, if Airwallex confirms extended pre-auth doesn't apply. Happy to do that instead if someone has confirmation — it's a bigger diff, which is why I'd rather not do it speculatively.

### Scope I deliberately left out

- **`config/deployments/production.toml` is unchanged.** It carries a narrower list (`cybersource,peachpayments`) than sandbox — `adyen` and `trustpay` are in sandbox and not production — so enabling a newly mapped connector in production config looked like an operator decision rather than mine. The other four configs are updated.
- **Merchant-initiated NT without a cryptogram is still unsupported.** A missing cryptogram returns `MissingRequiredField { field_name: "payment_method_data.network_token.token_cryptogram" }` rather than sending a token Airwallex will reject opaquely. A stored-token MIT keyed on a network transaction id would need its own handling (Adyen does this through `MandateReferenceId::NetworkTokenWithNTI`); today that case returns `NotImplemented`, so this is never worse, just not yet better. Say the word if you'd like it in scope.

## Motivation and Context

Fixes #13600. The reporter said they had no time to work on it. I re-verified all three of their claims against `main` before starting: the variant was still in the `NotImplemented` arm, `airwallex` was still absent from `connector_list`, and `AdyenNetworkTokenData` is still the reference shape.

## How did you test it?

Two serialization unit tests in `crates/hyperswitch_connectors`, pinning the wire format — which is the entire substance of the change and is otherwise unverifiable without an Airwallex account:

- the token serializes with `number_type: "EXTERNAL_NETWORK_TOKEN"`, four-digit expiry, `type: "card"`, and **asserts no `cvc` key is emitted**
- the cryptogram lands at `payment_method_options.card.cryptogram`

They need no credentials and no services, which is why I added them here — I'm aware `cargo test` isn't part of CI, so a test requiring a live dependency wouldn't earn its place.

Also verified: `just clippy` and `cargo +nightly fmt` clean. `network_tokenization_supported_connectors` deserializes into `HashSet<enums::Connector>`, so the config string is all that's needed — no enum or registry change.

## Checklist


abhisheksharma2411:
Checking why this has been quiet, and it looks mechanical rather than anything to do with the change — so flagging it in case a maintainer can unblock it with two clicks.

**Every substantive workflow on this PR is sitting at `action_required`.** Querying the runs for head `c8d664b1`:

```
CI-pr                                  action_required   (pull_request)
CI Nix                                 action_required   (pull_request)
CI wasm-check                          action_required   (pull_request)
Validate Generated OpenAPI Spec File   action_required   (pull_request)
Migration Consistency Tests            action_required   (pull_request)
Run Cypress tests                      action_required   (pull_request)
PR Title Spell Check                   action_required   (pull_request)
Pull Request Convention Checks         success           (pull_request_target)
```

The one that ran is the only one on `pull_request_target`, which executes in the base-repo context and so doesn't need approval. Everything on `pull_request` is waiting for a maintainer to hit **Approve and run workflows** — standard GitHub behaviour for a fork PR from someone with no merged PR here yet.

The effect is that **no build, test, clippy, wasm, migration or cypress run has happened on this branch at all**. The 5 green checks showing on the PR are the convention/secret-scanning gates only, which is easy to mistake for a passing CI.

For comparison, the last eight merged PRs here all look like this:

```
#13806  checks=45   labels=['S-test-ready']
#13772  checks=91   labels=['S-test-ready']
#13808  checks=72   labels=['S-test-ready']
#13758  checks=53   labels=['S-test-ready']
#13535  checks=96   labels=['S-test-ready']
```

This PR has **5 checks and no labels**. So there are two things needed, and I can't do either from outside: workflow-run approval, and the `S-test-ready` label that gates `cypress-tests-runner.yml`.

Locally, for what it's worth: the full `just clippy` equivalent is clean across the workspace — including the two new serialization tests, which I had to rewrite with `.get()`/`.pointer()` rather than indexing to satisfy `clippy::indexing_slicing`.

Happy to rebase onto the latest `main` first if you'd rather approve a fresh run — just say so. And if the right move is for me to close and reopen, or if there's a reviewer I should be tagging instead, I'm glad to be pointed at the correct process.

abhisheksharma2411:
Rebased onto current `main` as offered — this was `mergeable_state: behind` by 19 commits. New head `440a8a74c`, clean rebase, no conflicts.

Re-verified rather than trusting a clean rebase, since 19 commits of `main` can break a diff that still applies:

```
cargo clippy --all-targets --no-default-features --features <full workspace list>
  exit 0, no warnings from this change
  (one pre-existing future-incompat notice for upstream deps: buf_redux,
   grpc-api-types, multipart, num-bigint-dig, typemap)
```

`--all-targets` matters rather than being belt-and-braces: without it the new serialization tests are never compiled, so an earlier `cargo check` on this branch reported success while proving nothing about them. The same run is what caught `clippy::indexing_slicing` in those tests originally, which is why they use `.get()`/`.pointer()`.

**The two clicks from the earlier comment are still what this needs**, and nothing has changed on that front in three days: no labels, no reviewer, and all seven `pull_request` workflows still at `conclusion=action_required`. The fresh head should make approving a run straightforward if someone has a moment.

One note on method, in case it's useful to anyone else hitting it: `just` isn't installed here, so I replicated the `clippy` recipe by hand *including its feature computation*. A narrower `cargo test -p scheduler --features redis-rs` fails with spurious unresolved imports — `-p` scopes features to the package, so the workspace feature list is rejected outright. That's presumably why the recipe runs workspace-wide.

abhisheksharma2411:
Rebased onto current `main` (`d8b6ebe77`) — it had gone into conflict, 125 commits behind. 0 behind now, mergeable.

**Checked the feature is still unlanded before doing the work.** `main`'s Airwallex transformer still has `PaymentMethodData::NetworkToken(_)` in the `NotImplemented` arm, so nothing has superseded this.

**The conflict was structural, so I re-applied rather than merged text.** `main` moved the whole `payment_method` construction inside an `is_mit_payment` conditional (mandate/MIT payments send only `payment_consent_id`). Taking either side of that hunk would have been wrong, so I took `main`'s structure wholesale and re-inserted the network-token arm into its new match, removed `NetworkToken(_)` from the `NotImplemented` list, and added `cryptogram: None` to `main`'s card arm.

One thing that only surfaced on compile: `ConnectorError::MissingRequiredField.field_name` became `Cow<'static, str>` in that window. Fixed with `.into()`, matching how the rest of this file already spells it (`"order_details".into()`, `"connector_mandate_id".into()`).

### Verification

```
cargo check  -p hyperswitch_connectors --features "v1,payouts,frm,revenue_recovery"   clean
cargo clippy -p hyperswitch_connectors --features "v1,payouts,frm,revenue_recovery"   no findings
cargo +nightly fmt --all --check                                                      clean
```

### Still blocked on the same thing

Every substantive workflow on this PR remains at `action_required` — not red, not queued, waiting on a maintainer to approve the run for a fork PR. That's been the state since 20 Aug and it's the only thing between this and a review. Another contributor hit the identical wall on a different repo today, so it isn't specific to this branch.

This is the third rebase against a moving `main`. Happy to keep it current, but a maintainer clicking approve once would be worth more than any of them — without CI nobody can see that it builds, which is the whole reason it hasn't been looked at.

PR #13822, opened 2026-08-24
feat(analytics): emit API events for internal microservice calls

prasunna09:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

Calls made from the router to internal microservices (currently the modular payment methods service) were not captured by any event pipeline: only the inbound HTTP request produced an `ApiEvent`, so a failing service-to-service call was invisible in analytics.

Adds a `MicroserviceEvent` that the shared microservice client pipeline emits once per call, carrying the operation, URL, method, masked request body, status code, latency, error body and the merchant / profile / request identifiers forwarded to the service. Response bodies are not recorded, since payment method responses can carry raw card data that cannot be masked from the serialized bytes.

The event is published on the new `microservice_logs_topic` Kafka topic, along with the ClickHouse queue table, target table and materialized view to ingest it.

### Additional Changes

## Checklist


XyneSpaces:
[should-fix] The new microservice event pipeline can emit upstream error bodies without masking. Please mask or structure these error payloads before sending them to Kafka/ClickHouse, because microservice error responses can include echoed request fields or other sensitive payment-method data.

XyneSpaces:
[blocking] Add `microservice_logs_topic` to the production, sandbox, and integration `[kafka]` configs before this ships. The new producer path reads this topic for microservice API events; environments that omit it will publish to an empty topic and drop those events.

XyneSpaces:
[blocking] `microservice_logs_topic` is a newly required Kafka setting but it is not defaulted or validated, so deployments that do not yet carry this config key will deserialize it as an empty string and publish `MicroserviceApiLogs` to an empty topic. Wire `#[serde(default = "default_microservice_logs_topic")]` on the field and add it to KafkaSettings validation before emitting these events.

crates/router/src/services/kafka.rs
```diff
@@ -167,6 +167,11 @@ pub struct KafkaSettings {
     revenue_recovery_topic: String,
     external_service_call_topic: String,
     account_updater_topic: String,
+    microservice_logs_topic: String,
+}
+
+fn default_microservice_logs_topic() -> String {
```
VenuMadhav2541, line 176: This function has no callers, there's no #[serde(default = "default_microservice_logs_topic")] on microservice_logs_topic

crates/analytics/docs/clickhouse/scripts/microservice_events.sql
```diff
@@ -0,0 +1,106 @@
...
+    INDEX serviceIndex service_name TYPE bloom_filter GRANULARITY 1,
+    INDEX flowIndex flow TYPE bloom_filter GRANULARITY 1,
+    INDEX requestIdIndex request_id TYPE bloom_filter GRANULARITY 1
+) ENGINE = MergeTree PARTITION BY toStartOfDay(created_at)
+ORDER BY
+    (
+        created_at,
+        service_name,
+        flow,
+        status_code
```
VenuMadhav2541: `status_code` is Nullable and it's in the sorting key,  so ClickHouse rejects this with "Sorting key contains nullable columns. Can we please add test screenshots?

crates/hyperswitch_interfaces/src/micro_service/executor.rs
```diff
@@ -114,55 +132,93 @@ impl<O: ClientOperation> TransformedRequest<O> {
...
-            .await
-            .map_err(|e| {
+        // Step 4: Build the API event for this call. Merchant/profile identifiers travel in
+        // the parent headers, so read them back from there.
+        let mut microservice_event = MicroserviceEvent::new(
+            state.get_tenant().tenant_id.clone(),
+            service_name.to_string(),
+            operation,
+            masked_request_body,
+            url.to_string(),
```
VenuMadhav2541: url was mutated in Step 1 to append the query string, so this records the full URL including query params into an unmasked String column with an 18-month TTL

PR #13834, opened 2026-08-25
feat(external_services): add GCP Cloud KMS as a secrets/encryption manager backend

itsharshvb:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

Adds GCP Cloud KMS as a fourth `secrets_management`/`encryption_management` backend,
alongside `NoEncryption`, `AwsKms`, and `HashiCorpVault`. Mirrors AWS KMS's shape
(encrypt/decrypt an opaque ciphertext blob against a key), not Secret Manager's
fetch-by-name shape, since that's what this subsystem already expects.

- New `GcpKmsConfig` / `GcpKmsClient` / `GcpKmsOperations` / `GcpKmsError` in
  `crates/external_services/src/gcp_kms/`, feature-gated behind a new `gcp_kms` Cargo
  feature.
- `GcpKmsOperations` is a small internal seam between `GcpKmsClient` and the real
  `google-cloud-kms` SDK client (wrapped via a `RealGcpKms` newtype, matching how
  `AwsKmsClient`/`HashiCorpVault` already wrap their own SDK clients) — added
  specifically so unit tests can substitute an in-process fake instead of hitting the
  real GCP API. This backend has 10 tests that run in CI with zero network/credential
  dependency, unlike `aws_kms`, whose only tests are live-network and excluded from CI.
- Wired into both `SecretsManagementConfig` and `EncryptionManagementConfig` as a
  `GcpKms` variant — one client backs both traits, same as `AwsKmsClient`.
- Auth is Application Default Credentials only — no credential fields in config.
- `GcpKmsConfig` requires all four fields (`project_id`, `location_id`, `key_ring_id`,
  `key_id`) for both managers uniformly, unlike AWS's decrypt/encrypt asymmetry — GCP
  ciphertext doesn't embed key identity, so there's nothing to make optional.
- Four new metrics: `GCP_KMS_{DECRYPTION,ENCRYPTION}_FAILURES` (counters),
  `GCP_KMS_{DECRYPT,ENCRYPT}_TIME` (histograms), matching the existing AWS ones.
  
### End-to-end flow

```mermaid
flowchart TD
    A["SecretsManagementConfig::GcpKms /<br/>EncryptionManagementConfig::GcpKms<br/>(managers/*.rs)"] --> B["GcpKmsConfig::validate()<br/>all 4 fields required, checked before<br/>a client is ever constructed"]
    B --> C["GcpKmsClient::new(config)"]
    C --> D["ClientConfig::default().with_auth()<br/>resolves Application Default Credentials"]
    D --> E["Client::new(client_config)<br/>real google-cloud-kms SDK client"]
    E --> F["GcpKmsClient { inner_client: client, key_name }<br/>key_name = projects/.../cryptoKeys/...<br/>Client stored directly — no wrapper, no Arc"]

    F --> G["GcpKmsClient::decrypt(data) / encrypt(data)<br/>— the public, base64-string boundary"]
    G --> H["base64 decode/encode<br/>consts::BASE64_ENGINE"]
    H --> I["self.inner_client.decrypt/encrypt(EncryptRequest / DecryptRequest)<br/>direct call — real SDK call over gRPC"]

    I --> J["on error: logger::error! + metrics,<br/>change_context into GcpKmsError"]
    J --> K["implementers.rs:<br/>SecretManagementInterface::get_secret<br/>EncryptionManagementInterface::encrypt/decrypt"]
```

### A real startup panic this PR introduces, and fixes, in the same diff

Adding `google-cloud-kms` pulls in `tonic` (via `google-cloud-gax`), which requests the
`ring` rustls crypto backend — the first thing in this codebase's dependency graph to
disagree with the `aws-lc-rs` backend every existing AWS SDK crate here already agrees
on (via `aws-smithy-http-client`). rustls 0.23 refuses to pick between two compiled-in
backends automatically and panics at the first TLS handshake unless one is installed
explicitly. Confirmed via `cargo tree` that `aws_kms`-only and default builds request
`aws-lc-rs` exclusively — no conflict, no panic — so this is a genuine, direct
consequence of adding `gcp_kms`, not a pre-existing unrelated issue.

Fixed with an explicit `CryptoProvider::install_default()` call at the top of
`main()` (`crates/router/src/bin/router.rs`), gated behind a new `gcp_kms` feature on
`router` itself (`crates/router/Cargo.toml`) so the fix — and the extra `rustls_0_23`
dependency it needs — is only compiled into binaries that actually enable GCP KMS.
`release` now activates `gcp_kms` instead of `external_services/gcp_kms` directly, so
this fix is automatically included wherever GCP KMS is.

### Additional Changes

## How did you test it?

**Unit tests** (`crates/external_services/src/gcp_kms/core.rs`, run under `cargo test -p
external_services --features gcp_kms`) — 10/10 passing, no network or credentials
required.

**Real GCP Cloud KMS, twice, against two different projects/keys** — encrypted a
plaintext value, got real ciphertext back from live Cloud KMS, decrypted that exact
ciphertext, got the original plaintext back both times.

**Full local router boot** with every `SecretsHandler`-processed config field (32
fields) GCP-KMS-encrypted: router booted clean and listened on `:8080`; `POST
/accounts` with the decrypted `admin_api_key` → `200` with a real merchant account
created (proves `master_database.password` decrypted correctly too, via a real
Postgres write); same request with a wrong key → `401`.

Screenshot 2026-08-24 at 4 08 07 PM
Screenshot 2026-08-24 at 4 08 54 PM

## Checklist


XyneSpaces:
[blocking] Avoid panicking during router startup when installing the rustls provider for the `gcp_kms` feature. `install_default()` returns `Err` if another crate has already installed a provider before `main`, so this feature can make startup fail depending on initialization order; handle the already-installed case without `expect`.

config/deployments/env_specific.toml
```diff
@@ -361,19 +361,31 @@ private_key = "/path/to/private_key.pem"
 certificate = "/path/to/certificate.pem"
 
 [secrets_management]
-secrets_manager = "aws_kms" # Secrets manager client to be used
+secrets_manager = "aws_kms" # Secrets manager client to be used (NoEncryption | aws_kms | hashicorp_vault | gcp_kms)
```
SanchithHegde: Nit: Isn't this supposed to be `no_encryption` instead of `NoEncryption`?

config/deployments/env_specific.toml
```diff
@@ -361,19 +361,31 @@ private_key = "/path/to/private_key.pem"
...
 
 [secrets_management.aws_kms]
 key_id = "kms_key_id" # The AWS key ID used by the KMS SDK for decrypting data.
 region = "kms_region" # The AWS region used by the KMS SDK for decrypting data.
 
+[secrets_management.gcp_kms]
+project_id = "gcp_project_id"
+location_id = "gcp_kms_location_id" # e.g. "global", "us-east1"
+key_ring_id = "gcp_kms_key_ring_id"
+key_id = "gcp_kms_key_id"
```
SanchithHegde, line 376: Can we add a comment on these 2 blocks that they would have to be included, only if `secrets_manager` is `aws_kms` or `gcp_kms` respectively?

crates/external_services/src/gcp_kms/core.rs
```diff
@@ -0,0 +1,514 @@
...
+            Err("GCP KMS key ID must not be empty")
+        })
+    }
+}
+
+/// A thin seam between [`GcpKmsClient`] and the real `google-cloud-kms` client, so that
+/// unit tests can substitute an in-process fake instead of hitting the real GCP KMS API
+/// over the network.
+#[async_trait::async_trait]
+pub(crate) trait GcpKmsOperations: Send + Sync {
```
SanchithHegde: Do we need this trait, and do we need the `FakeGcpKmsOperations` used in tests?

crates/router/Cargo.toml
```diff
@@ -245,6 +256,11 @@ unified-connector-service-cards = { git = "https://github.com/juspay/connector-s
 rustc-hash = "1.1.0"
 rustls = "0.22"
 rustls-pemfile = "2"
+rustls_0_23 = { package = "rustls", version = "0.23", optional = true, default-features = false, features = [
```
SanchithHegde: Can we bump to `rustls 0.23` across the repo instead of directly depending on 2 versions of the crate? Are there any blockers in updating?

SanchithHegde (approved):

gopikrishna000 (approved):

PR #378, opened 2023-01-14
build(deps): use utoipa crate upstream URL as git dependency

SanchithHegde:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [x] Dependency updates

## Description

This PR updates the repository URL for the `utoipa` crate to use the upstream repository URL instead of my fork.

## Motivation and Context

Since juhaku/utoipa#436 was merged, my fork is no longer required. However, since it could take some time for the next release of the crate, we'll be using it as a git dependency.

## How did you test it?

Regenerated the OpenAPI spec file, there are no changes in property order of the schemas.

## Checklist

- [x] I formatted the code `cargo +nightly fmt --all`
- [x] I addressed lints thrown by `cargo clippy`
- [x] I reviewed submitted code
- [ ] I added unit tests for my changes where possible
- [ ] I added a [CHANGELOG](/CHANGELOG.md) entry if applicable

PR #13846, opened 2026-08-25
fix(routing): send all dashboard routing dimensions to the Decision Engine

### LINKED ISSUE #13855: [BUG] Five dashboard routing dimensions never sent to the Decision Engine, so rules keyed on them could not match
## Summary

`convert_backend_input_to_routing_eval` omitted five of the routing dimensions the dashboard lets merchants build rules on, so any rule keyed on them could never match in the Decision Engine: DE fell through to `default_selection`, substituted the merchant fallback list, and diverged from the Hyperswitch result on every such payment — a permanent, unfixable shadow diff feeding the kill-switch counter (#13565) on otherwise healthy profiles.

## Detail

Now sent: `surcharge_amount`, `transaction_initiator`, `card_discovery`, `issuer_country` — plus `card_type` alongside the existing `card` key. `DirKeyKind::CardType` serialises as `card_type`, so dashboard-authored rules use that spelling while only `card` was being sent; the DE declares both keys, so emitting both fixes new rules without breaking rules authored under the older name.

**Deploy ordering:** the DE rejects an entire evaluate request on an unrecognised parameter key (`InvalidRequestParameter` → `fail_preview`), which Hyperswitch swallows into an empty result — so the engine must already carry all 43 `[routing_config.keys]` entries before this ships (`surcharge_amount` reached that config on 2026-08-18).

## Verification

At the DE contract: a surcharge-gated rule matches with the parameter present, falls through to the fallback without it, and `0` correctly fails `> 0` — real comparison semantics, not key-presence matching. End to end through Hyperswitch: a card payment matching a `card_type = credit` rule routes to the rule's connector with an `is_equal: true` diff log on both engines.

## PRs

- [x] #13846 — main
- [x] #13875 — hotfix backport (2026.08.26.1)

Part of #13762.

AnkitKmrGupta:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

`convert_backend_input_to_routing_eval` omitted five of the routing dimensions the dashboard lets merchants build rules on, so any rule keyed on them could never match in the Decision Engine. DE fell through to `default_selection` and substituted the merchant fallback list, diverging from the Hyperswitch result on every such payment.

Missing parameters now sent:

| Key | Source |
| --- | --- |
| `surcharge_amount` | `payment.surcharge_amount` |
| `transaction_initiator` | `payment.transaction_initiator` |
| `card_discovery` | `payment_method.card_discovery` |
| `issuer_country` | `issuer_data.country` |

Also emits `card_type` alongside the existing `card` key. `DirKeyKind::CardType` serialises as `card_type`, so dashboard-authored rules use that spelling while only `card` was being sent. The Decision Engine declares both keys, so emitting both fixes new rules without breaking any authored under the older name.

No Decision Engine change is required — `[routing_config.keys]` in both `config/development.toml` and `config/docker-configuration.toml` already declares all 43 keys, covering every dashboard dimension.

**Deploy ordering:** the Decision Engine rejects an entire evaluate request on an unrecognised parameter key (`InvalidRequestParameter` → `fail_preview`), which Hyperswitch swallows into an empty result. DE must already be on a build carrying all 43 `[routing_config.keys]` entries before this ships. `surcharge_amount` was added to that config on 2026-08-18.

### 1. Decision Engine evaluate — `surcharge_amount` present vs absent

Isolates exactly what this PR changes: the same active algorithm and the same payment, with the parameter as the only difference.

**Set up a rule that only matches when a surcharge is present**

```bash
curl -X POST http://127.0.0.1:8082/routing/create \
  -H 'Content-Type: application/json' -H 'x-admin-secret: test_admin' \
  -d '{
    "name": "surcharge-regression",
    "created_by": "pro_surcharge_demo",
    "algorithm_for": "payment",
    "algorithm": {"type":"advanced","data":{
      "globals":{}, "metadata":{}, "default_selection":{"priority":[]},
      "rules":[{"name":"rule_1","routing_type":"priority",
        "output":{"priority":[{"gateway_name":"stripe","gateway_id":"mca_surcharge_stripe"}]},
        "statements":[{"nested":null,"condition":[
          {"lhs":"surcharge_amount","comparison":"greater_than","value":{"type":"number","value":0},"metadata":{}}]}]}]}}}'
```

```json
{"rule_id":"routing_591814a0-7778-49fb-9b17-80a2e757c7e6","name":"surcharge-regression","algorithm_for":"payment"}
```

```bash
curl -X POST http://127.0.0.1:8082/routing/activate \
  -H 'Content-Type: application/json' -H 'x-admin-secret: test_admin' \
  -d '{"created_by":"pro_surcharge_demo","routing_algorithm_id":"routing_591814a0-7778-49fb-9b17-80a2e757c7e6"}'
# HTTP 200
```

**Case A — `surcharge_amount` absent (behaviour before this PR)**

```bash
curl -X POST http://127.0.0.1:8082/routing/evaluate \
  -H 'Content-Type: application/json' -H 'x-admin-secret: test_admin' \
  -d '{
    "created_by":"pro_surcharge_demo","payment_id":"pay_demo",
    "fallback_output":[{"gateway_name":"adyen","gateway_id":"mca_fallback_adyen"},
                       {"gateway_name":"cybersource","gateway_id":"mca_fallback_cyb"}],
    "parameters":{"amount":{"type":"number","value":1000},
                  "currency":{"type":"enum_variant","value":"USD"},
                  "payment_method":{"type":"enum_variant","value":"card"}}}'
```

```json
{
  "payment_id": "pay_demo",
  "status": "default_selection",
  "evaluated_output": [
    {"gateway_name": "adyen", "gateway_id": "mca_fallback_adyen"},
    {"gateway_name": "cybersource", "gateway_id": "mca_fallback_cyb"}
  ]
}
```

The rule cannot match, DE falls through to `default_selection` and substitutes the merchant fallback list. This is the divergence seen in sandbox.

**Case B — `surcharge_amount: 100` (behaviour after this PR)**

Same request with one extra parameter:

```bash
    "parameters":{..., "surcharge_amount":{"type":"number","value":100}}
```

```json
{
  "payment_id": "pay_demo",
  "status": "success",
  "evaluated_output": [
    {"gateway_name": "stripe", "gateway_id": "mca_surcharge_stripe"}
  ]
}
```

**Case C — `surcharge_amount: 0` (boundary)**

```json
{
  "payment_id": "pay_demo",
  "status": "default_selection",
  "evaluated_output": [
    {"gateway_name": "adyen", "gateway_id": "mca_fallback_adyen"},
    {"gateway_name": "cybersource", "gateway_id": "mca_fallback_cyb"}
  ]
}
```

`0` correctly fails `> 0`, so the fix restores real comparison semantics rather than matching on key presence.

### 2. End-to-end through Hyperswitch — `card_type`

`card_type` is one of the five keys this PR adds and is populated by an ordinary card payment, so it exercises the same code path end to end.

```bash
curl -X POST http://localhost:8080/routing \
  -H 'Content-Type: application/json' -H "api-key: $API_KEY" \
  -d '{
    "name":"cardtype-e2e","profile_id":"pro_M72RxtufoXilqEvQPrRa",
    "algorithm":{"type":"advanced","data":{
      "defaultSelection":{"type":"priority","data":[]}, "metadata":{},
      "rules":[{"name":"rule_1",
        "connectorSelection":{"type":"priority","data":[{"connector":"zift","merchant_connector_id":"mca_6cVktnfhgGuvuOyIjgLr"}]},
        "statements":[{"nested":null,"condition":[
          {"lhs":"card_type","comparison":"equal","value":{"type":"enum_variant","value":"credit"},"metadata":{}}]}]}]}}}'
```

```json
{"id":"routing_qurRwOA8M9PuX62vgM6S","profile_id":"pro_M72RxtufoXilqEvQPrRa","kind":"advanced",
 "algorithm_for":"payment","decision_engine_routing_id":"routing_qurRwOA8M9PuX62vgM6S"}
```

```bash
curl -X POST http://localhost:8080/routing/routing_qurRwOA8M9PuX62vgM6S/activate \
  -H 'Content-Type: application/json' -H "api-key: $API_KEY" -d '{}'
# HTTP 200 — synced to the Decision Engine, confirmed via
# POST /routing/list/active/pro_M72RxtufoXilqEvQPrRa which stores lhs "card_type"
```

```bash
curl -X POST http://localhost:8080/payments \
  -H 'Content-Type: application/json' -H "api-key: $API_KEY" \
  -d '{"amount":1000,"currency":"USD","confirm":true,"capture_method":"automatic",
       "profile_id":"pro_M72RxtufoXilqEvQPrRa","customer_id":"cus_surcharge_e2e",
       "payment_method":"card","payment_method_type":"credit",
       "payment_method_data":{"card":{"card_number":"4242424242424242","card_exp_month":"12",
         "card_exp_year":"2030","card_holder_name":"Test","card_cvc":"123"}},
       "billing":{"address":{"country":"US","line1":"1 Test St","city":"SF","zip":"94103",
         "state":"CA","first_name":"T","last_name":"U"}}}'
```

```json
{
  "payment_id": "pay_MV4CRUdmrrDmj1HW28vq",
  "status": "failed",
  "connector": "zift",
  "amount": 1000,
  "currency": "USD"
}
```

`connector: zift` shows the rule matched. (`status: failed` is the local dummy connector credentials — routing happens before the connector call.)

Router diff log for that payment:

```
decision_engine_euclid, routing_flow: "evaluate_routing",
  is_equal: true, is_equal_length: true, is_volume: false, is_de_result_empty: false,
  de_response: "[{\"connector\":\"zift\",\"merchant_connector_id\":\"mca_6cVktnfhgGuvuOyIjgLr\"}]",
  hs_response: "[{\"connector\":\"zift\",\"merchant_connector_id\":\"mca_6cVktnfhgGuvuOyIjgLr\"}]"
```

Both engines return the single rule connector rather than the profile's five-connector fallback, so this is a genuine rule match on both sides and not a fallback tie.

### Not covered

A full end-to-end surcharge payment is not included. The routing DSL sources `surcharge_amount` only from `payment_attempt.external_surcharge_details` (`routing.rs:658`), populated by `populate_external_surcharge_details` under `SurchargeMode::External`. The internal Surcharge Decision Manager resolves to `SurchargeMode::Internal` and deliberately does not feed it, and `POST /payments/{id}/eligibility` returns `surcharge_details: null` without an external surcharge connector (`connector_type = surcharge_processor`, e.g. interpayments) on the profile — which the local setup does not have. Test 1 covers the same behaviour at the Decision Engine contract instead.

---

## Decision Engine cutover — testing

Covers the cutover commits on this branch (`feat(routing): serve routing APIs and payment paths from the Decision Engine for cut-over profiles` and the two follow-up fix commits).

Local stack: `router` on `:8082` (this branch), `decision-engine` on `:8081` (Postgres, built from `feat/evaluate-algorithm-for`), shared local Postgres + Redis. Cutover is set per profile through `POST /configs/` (`routing_result_source_{profile_id} = decision_engine`); the router runs with a superposition seed that omits the key so the DB value is authoritative. Connectors are the dummy `stripe_test` / `paypal_test`, and `wise` for payouts, so the routed target is provable from the returned `connector` / stored `merchant_connector_id`.

Every scenario below is an automated assertion, not a manual spot check. Totals: **63 + 20 + 9 = 92 assertions, all passing**.

### Suite A — cutover behaviour, flag on (63 assertions)

| # | Scenario | What is asserted |
|---|---|---|
| 1 | Merchant **not** cut over (control) | Create dual-writes to DE and stores `decision_engine_routing_id`; activate updates `business_profile.routing_algorithm`; `/routing/active` lists the rule; payment routes by the HS rule (`paypal_test`). Legacy behaviour unchanged. |
| 2 | **New profile cut over before any rule exists** (the reported bug) | `/routing/entry` reports `is_cutover: true`; create returns 200 with a DE id and **zero** HS `routing_algorithm` rows; the rule exists on the DE; retrieve-by-id is served from the DE; activate sets the DE mapper; `/routing/active` lists the rule despite a NULL HS ref; **the payment is routed by the DE rule with no HS algorithm present**; re-activation is rejected with `already active`. |
| 3 | Payouts on a cut-over profile | Payout rule created DE-only and activated; DE payout mapper set; the payout is routed to the exact `wise` MCA the DE rule names; **the payment on the same profile still uses the payment rule** (transaction-type isolation). |
| 4 | Deactivate under cutover | Deactivate returns 200 via the DE; the DE payment mapper is cleared; the payment falls back to the merchant default; a second deactivate returns `already inactive`. |
| 5 | Rule created directly on the **DE dashboard** | A rule created straight on the DE (no HS row) is retrievable through HS, activatable through HS, and routes payments. |
| 6 | **Old merchant flipped to DE** | Pre-cutover payments use the HS rule; after the flip the same rule is served from the DE (ids align); routing is unchanged from the merchant's perspective. |
| 7 | Legacy rule with `decision_engine_routing_id = NULL` | Activation creates the rule on the DE on demand and then activates it; the payment routes by the re-created rule. |
| 8 | Guards | Profile `PATCH` with `routing_algorithm` is rejected for a cut-over profile and allowed for a non-cut-over one; `/routing/rule/evaluate` accepts the caller's own profile and rejects a foreign `created_by`. |
| 9 | List APIs | Merchant-level list is HS-only and paginated; the profile-scoped list serves DE-only rules; a non-cut-over profile's list is unchanged. |

### Suite B — regressions for the review fixes, flag on (20 assertions)

| # | Fix under test | What is asserted |
|---|---|---|
| 1 | Stale HS ref cleared | With a pre-cutover HS rule linked (`paypal_test`), activating a DE rule clears `business_profile.routing_algorithm`; payments route by the DE rule; after DE deactivation the ref stays clear, the payment falls to the **default fallback and not the frozen HS rule**, and `/routing/active` reports nothing active. |
| 2 | Migration can bootstrap | With the profile's DE rules deleted (mimicking a never-migrated profile the DE does not know), `POST /routing/rule/migrate` succeeds, creates the rules on the DE, and a re-run is idempotent (no 500). |
| 3 | Unknown rule id | `GET /routing/{unknown_id}` under cutover still returns **404**, not a DE error. |
| 4 | 3DS carve-out | A rule created with `transaction_type: three_ds_authentication` on a cut-over profile is written to **Hyperswitch**, not DE-only. |

### Suite C — global flag precedence, flag off (9 assertions)

With `open_router.static_routing_enabled = false` and a profile whose `routing_result_source` is still `decision_engine`:

- `/routing/entry` reports `is_cutover: false`
- create writes to Hyperswitch (legacy path), activate updates `business_profile.routing_algorithm`
- **the payment is routed by the HS rule**, i.e. the flag is a working revert lever on the payment path

### Control Center

The dashboard (`hyperswitch-control-center`) was run against this router and driven through the browser for a cut-over profile whose rules exist only on the Decision Engine: the Smart Routing page renders the cutover card set, and the **Active configuration** card, **Configuration History** tab, the rule detail page and the **Deactivate Configuration** action all reflect DE state (verified via the JWT `/routing/entry`, `/routing/active`, `/routing/list/profile` and `/routing/{id}` calls). Deactivating from the UI cleared the DE mapper and flipped the Active card to Default Fallback.

Known display gap: for an *advanced* rule that exists only on the DE, the detail page shows the rule's conditions but not the per-rule connector chips, because the DE returns its own program shape and there is no reverse AST transformer. Priority and volume-split rules render fully.

### Builds and static checks

- `cargo check -p router` (v1) and `cargo check -p router --no-default-features --features "v2,..."` both clean — the v2 build break found in review is fixed and verified against the failing command.
- `cargo clippy -p router` clean; `cargo +nightly fmt` applied.
- OpenAPI specs regenerated with `cargo run -p openapi --features v1 && cargo run -p openapi --features v2`. The only change against `main` is the optional `algorithm_for` request field on `RoutingEvaluateRequest`; v2 has no spec change at all. `oasdiff` reports zero breaking changes on both specs.

### Batch evaluation — session tokens and payment method list

Raised in review: these flows evaluate the rule once per payment method type (a rule may branch on the type), which was N serialized Decision Engine round trips per checkout request — and after the diff-checker change, for every merchant.

Landed in three steps:

1. **Off the request path for everyone not cut over** — the result is only load-bearing for a cut-over profile; for all other merchants it exists purely for the HS/DE diff, so it is spawned through the same shadow mechanism the payment path already uses (gated on `shadow_routing_enabled`). Zero added latency for the overwhelming majority of traffic.
2. **Concurrent for cut-over profiles** — per-type inputs built up front, calls issued together: ~1×RTT instead of N×RTT.
3. **One batch call** — `POST /routing/evaluate/batch` (paired change: [juspay/decision-engine#378](https://github.com/juspay/decision-engine/pull/378)) carries every wallet type's parameter set in one request; the engine resolves and parses the active algorithm once and returns positional results. N round trips and N rule lookups become one of each. **Deploy ordering is soft**: against an engine without the endpoint the load-bearing path degrades to the concurrent single calls from step 2, so either side can ship first. A failed *shadow* batch just logs empty diffs — the events layer strips status codes, so "endpoint absent" is indistinguishable from "no active rule", and for the latter single-call fallbacks fail identically.

#### Batch testing (all against locally running router + engine)

Decision Engine direct (engine build from #378), advanced rule branching on `payment_method_type` — 14/14:

| Assertion | Result |
|---|---|
| Single evaluate after the engine-side refactor: each arm routes correctly, fallback echo intact | pass |
| Batch of 3 returns positional results **byte-identical to the single calls** | pass |
| Cache-hit repeat identical; empty/oversized/invalid-enum/unknown-merchant guards all reject | pass |

Full stack (this branch's router + #378's engine) — 14/14:

| Assertion | Result |
|---|---|
| Cut-over card payment routes by the rule's default arm through the DE | pass |
| PML pre-routing on a cut-over profile sends **one batch evaluate** (proven by the engine's own `api_requests_total{endpoint="routing_evaluate_batch"}` counter) | pass |
| Non-cutover profile's PML sends a **spawned shadow batch**, off the request path | pass |

Regression: the 63-assertion cutover suite passes unchanged against the new binaries (63/63). The review-fixes suite passes 17/20 — the 3 failures assert the ref-clearing behaviour this PR deliberately reversed (the ref is now kept and serves as the DE-empty fallback), so the assertions are stale, not the code.

### Review follow-ups — cutover resolution on three read paths

Raised in review: three paths resolved rules from Hyperswitch alone, so a cut-over profile did not behave as one.

**`GET /routing/{id}`** queried the HS table first and only consulted the Decision Engine when that query errored. A migrated rule has an HS row, so retrieve served the HS copy while payments used the DE one. Since the DE can edit a rule in place, the two can diverge and the rule-detail page shows stale content. Retrieve now prefers the DE copy for a cut-over profile — after the profile fetch and auth check — and falls back to the HS row when the DE has no record, or when the DE call itself fails, so an engine outage degrades a read rather than failing it.

**Wallet session-token routing** and **payment-method-list pre-routing** read `business_profile.routing_algorithm` directly and never consulted cutover; with no algorithm id they fell through to `get_merchant_default_config`. This was not a rare edge — activation under cutover clears that ref, so for a cut-over profile it was every request. Both now evaluate on the DE and pick through `select_routing_result`, the same helper the payment path uses.

No new engine support was required. Session routing is the same rule evaluated with a partly populated input, and the evaluate request builder omits absent parameters, so the existing `routing/evaluate` call serves it.

| Assertion | Result |
|---|---|
| `just clippy` (v1, `-D warnings`) | pass |
| `just clippy_v2` (v2, `-D warnings`) | pass |
| `cargo test -p router --features v1 --lib de_program_round_trip` | 8 passed |

Shape of the fix, for reviewers:

- The DE is called **per payment method type**, because a rule can branch on the type. The cutover check is hoisted out of the loop, and a profile that is not cut over makes no DE call and resolves no extra config — but a cut-over merchant with three wallet types now makes three DE calls per session request. That is the real cost of this fix.
- **Only the connector list is swapped.** `RoutingApproach` is left as the Hyperswitch side derived it, matching `perform_static_routing_v1`. It is persisted on the attempt and consumed by analytics, so relabelling it here would shift those numbers for cut-over merchants as a side effect of a routing fix. The consequence is that a cut-over session payment still reports the Hyperswitch-derived approach; worth its own change rather than this one.
- The per-profile DE lookup is extracted so retrieve and the rule-id scan share one implementation.

### Suite D — Decision Engine advanced-rule conversion (8 unit tests)

Retrieving an advanced rule from a cut-over profile previously had no way to build a euclid AST, so the body was returned verbatim under an extra `RoutingAlgorithmWrapper::DecisionEngine` variant. That added a third arm to an untagged response union — which `oasdiff` flags as breaking — and left the dashboard able to render such a rule's conditions but not its connector chips. This adds the reverse of the existing euclid to Decision Engine transformer, so a DE-authored rule comes back in exactly the same typed shape as a Hyperswitch-authored one, and the extra variant is removed.

`cargo test -p router --features v1 de_program_round_trip` — 8 passed, 0 failed.

| Assertion | Result |
|---|---|
| Round trip over a program using all seven shared value types with nested if-statements is lossless | pass |
| Round trip is lossless for a volume-split program | pass |
| `Output::Single` widens to a one-element priority list (routes identically) | pass |
| A DE program with no metadata converts to empty metadata rather than erroring | pass |
| `Output::VolumeSplitPriority` is rejected (no euclid equivalent) | pass |
| `ValueType::GlobalRef` is rejected (no euclid equivalent) | pass |
| A number above `i64::MAX` is rejected instead of wrapping to a negative amount | pass |
| An unknown connector name is rejected | pass |

The four rejections matter more than the two round trips: the two type worlds are not isomorphic, and each gap fails loudly rather than round-tripping into a rule that would route differently. `Program.globals` is dropped, which is safe precisely because `GlobalRef` errors — globals are only reachable through it, so an unreferenced map carries no behaviour.

### Not covered

- The Routing cypress suite is not exercised: it runs only when the `ROUTING` env var is set and no workflow sets it, so these scenarios live in a local harness rather than CI. Porting them into cypress and enabling the job is the natural follow-up.
- The v2 `perform_session_flow_routing` still resolves rules from Hyperswitch alone for cut-over profiles. v1 is fixed (see below); v2 is unchanged and carries no regression.
- `card_discovery` is now populated on the v1 payment path only; the payout and v2 DSL builders have no attempt-level source for it.
- The paired Decision Engine change is required for the payout isolation asserted in Suite A #3: [juspay/decision-engine#376](https://github.com/juspay/decision-engine/pull/376). **Deploy the Decision Engine PR first** — until it ships, a profile with both payment and payout rules active can have the wrong transaction type's rule evaluated. Hyperswitch can merge independently: the engine ignores the new `algorithm_for` field, falling back to its current type-blind lookup.

## Checklist


crates/router/src/core/payments/routing/utils.rs
```diff
@@ -1476,7 +1595,7 @@ fn insert_dirvalue_param(params: &mut HashMap<String, Option<ValueType>>, dv: di
         }
         dir::DirValue::CardType(v) => {
             params.insert(
-                "card".to_string(),
```
jagan-jaya, line 1479: why 'card' is removed , is that intentional?

crates/router/src/core/payments/routing.rs
jagan-jaya, line 1: Wallet session-token routing and payment-method-list pre-routing lose the merchant's rule on cut-over profiles.

perform_session_flow_routing (/payments/routing.rs:2855) reads business_profile.routing_algorithm directly, never consults cutover, and with algorithm_id == None falls through to get_merchant_default_config (:3049-3059). Callers: hs:core/payments.rs:12924 and hs:core/payment_methods/cards.rs:4439.

crates/router/src/core/routing.rs
jagan-jaya, line 1: ` GET /routing/{id} retrieve` retrieve_routing_algorithm_from_algorithm_id (hs:core/routing.rs:1729-1751) queries the HS table first and only calls find_de_record_for_algorithm when that query errors. A migrated rule has an HS row, so under cutover, retrieve serves the HS copy — not DE.

 DE exposes /routing/update, which edits a rule in place; if anyone edits DE-side, HS's rule-detail page shows stale content while payments use the DE version.
 This should resolve cutover first and read from DE, falling back to the HS row only when DE has no record the same shape as link_routing_config already uses.

jagan-jaya (changes requested):

crates/router/src/core/payments/routing/utils.rs
prajjwalkumar17: Yes HS has renamed `card` to `card_type`.

crates/router/src/core/payments/routing.rs
prajjwalkumar17: Confirmed — and on this branch it was worse than latent. `clear_hs_active_algorithm_ref` clears the profile's algorithm ref when a rule is activated under cutover, so for a cut-over profile `algorithm_id` was always `None` and both paths fell to `get_merchant_default_config` on **every** request, not just occasionally.

Fixed in 372003230d.

This turned out to be two distinct v1 paths rather than one: `SessionRoutingStage::route` (wallet session tokens, via `perform_session_token_routing`) and `perform_session_routing_for_pm_type` (payment-method-list pre-routing, via `cards.rs`). Both now evaluate on the Decision Engine when the profile is cut over and pick the result through `select_routing_result`, the same helper `perform_static_routing_v1` uses. `_business_profile` is a real parameter again.

No new engine support was required: session routing is the same rule evaluated with a partly populated input, and `convert_backend_input_to_routing_eval` omits absent parameters rather than erroring, so the existing `routing/evaluate` call serves it as-is.

crates/router/src/core/routing.rs
prajjwalkumar17: Confirmed. Fixed in 372003230d.

Retrieve now prefers the DE copy for a cut-over profile and keeps the HS row as the fallback. The DE lookup happens *after* the profile fetch and `validate_profile_id_from_auth_layer`, so the auth checks are unchanged.

One deliberate deviation from the suggestion: it also falls back to the HS row when the **DE call itself fails**, not only when the DE has no record. A read endpoint degrading to a possibly-stale row seemed better than a DE outage turning rule-detail into a 500.

It cannot literally resolve cutover first, since the endpoint receives only an `algorithm_id` and `find_de_record_for_algorithm` fans out across every profile when the profile is unknown. So the HS query stays first purely to yield `profile_id`, and the DE is then consulted for that single profile — one targeted call, no scan. I extracted `find_de_record_in_profile` so retrieve and the rule-id scan share one implementation rather than duplicating the cutover-check-then-fetch logic.

crates/router/src/core/payments/routing.rs
jagan-jaya, line 1: let's add the diffchecker for sessions calls and pml call also

> │ payments/routing.rs:1034 SessionRoutingStage                 │ session tokens                   │ ❌          │

> │ payments/routing.rs:3140 perform_session_routing_for_pm_type │ session tokens + PML pre-routing │ ❌ |

crates/router/src/core/routing.rs
```diff
@@ -805,21 +1194,84 @@ pub async fn link_routing_config_under_profile(
...
+                .map_err(|error| {
+                    map_de_write_error(
+                        error,
+                        "decision_engine_euclid: rule activation failed on the Decision Engine",
+                    )
+                })?;
+
+                // The DE now owns this profile's active rule; a leftover HS ref would
+                // still be served whenever the DE result comes back empty.
+                clear_hs_active_algorithm_ref(
```
jagan-jaya: clear_hs_active_algorithm_ref — cutover destroys state it can't rebuild

When a rule is activated on a cut-over profile, HS nulls the profile's routing_algorithm ref. The reason is real: under cutover, if the DE returns an empty result, select_routing_result would otherwise serve the stale HS rule - so they delete the input rather than ignore it.

The cost is that the pointer to "which rule was active in HS" is gone. The routing_algorithm rows survive; the pointer doesn't. So turning cutover back off does not restore the previous behaviour - the profile now has no active HS algorithm and drops to the merchant fallback list. Recovery is re-activating each affected profile by hand, and only if you still know which id was active. Rules created after cutover are worse: they have no HS row at all, so there's nothing to re-activate.

Fix is small: under cutover, ignore the ref instead of clearing it - one match arm in perform_static_routing_v1 (_ if de_routing_effective => None). Same routing behaviour, nothing destroyed, and revert becomes a pure config flip.

jagan-jaya (changes requested):

crates/router/src/core/payments/routing.rs
prajjwalkumar17: Added in 7d411dae96 — both flows now call `compare_and_log_result`, the same helper the payment path uses.

This needed one behaviour change to be useful: the DE was previously only evaluated on these paths under cutover, so there was nothing to diff against for the profiles a shadow comparison actually exists for. It now runs whenever `open_router.static_routing_enabled` is set, matching `perform_static_routing_v1`. `is_volume` is derived from the routing approach rather than threaded through as a new flag.

Each flow also names itself now. `RoutingFlow` (`Payment` / `SessionToken` / `PaymentMethodList`) is carried into the diff log — whose `routing_flow` field existed but received the same value from every caller — and into the routing event name, which was hardcoded to `"DecisionEngine: Euclid Static Routing"` for all three. `Payment` keeps its exact historical strings (`"evaluate_routing"` and the original event name) so existing queries and dashboards do not move.

Two things worth flagging:

- **The kill switch is deliberately not fed from these flows.** It gates routing for the whole profile, so tripping it on a session-flow discrepancy would silently disable DE routing for payments too. Say the word if you would rather it counted.
- **Cost:** with the flag on, a session request now makes one DE call per payment method type for every merchant, not only cut-over ones. Per-type is required for correctness since a rule can branch on the type, so this is inherent to wanting the diff on this path.

crates/router/src/core/routing.rs
prajjwalkumar17: Good catch, and you read the reasoning correctly — that is exactly why it was clearing, and deleting the input was the wrong remedy. Fixed in 7d411dae96 using your suggestion: `clear_hs_active_algorithm_ref` and its three call sites are gone, and the ref is ignored via one arm in `perform_static_routing_v1` plus the same guard in the two session paths. `cached_algorithm` is `None` either way, so routing behaviour was unchanged and cutover became a reversible config flip again.

Then a follow-up in 98bcb84bc7 goes further, and it reverses your premise, so please push back if you disagree.

Rather than ignoring the ref, the Hyperswitch rule is now **evaluated** on cut-over profiles, so an empty DE result falls back to the merchant's own rule instead of the flat fallback list. You treated serving the stale rule as the bug; the argument the other way is the unmigrated cut-over profile — cut over before its rules reach the DE, it gets an empty result and today routes everything by fallback while looking healthy, which is the failure this PR exists to fix. Falling back to their real rule is the better answer there.

The cost is the one you identified: under DE-only writes the Hyperswitch rule is frozen at the moment of cutover, so an outage can route by stale logic. That is now a `warn` naming the profile rather than a `debug`, so it is visible instead of silent. Profiles whose rules only ever existed on the DE have no Hyperswitch row and still land on the fallback list.

Side benefit: it makes the diff meaningful under cutover. With the ref ignored, the Hyperswitch side of the comparison *was* the fallback list, so cut-over profiles logged DE-vs-fallback noise; it is now a real rule-vs-rule diff.

jagan-jaya (dismissed):

jagan-jaya (dismissed):

sahkal (approved):

GauravRawat369 (approved):

jagan-jaya (approved):

prajjwalkumar17 (approved):

PR #13831, opened 2026-08-24
feat(offers): add merchant-level Offer Engine credential source

apoorvdixit88:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

One of task to complete Offer Engine milestone 4 — move offer configuration from the **application level** to the **merchant level** so each merchant can use its own Offer Engine account instead of the single shared application-level credentials.

The `offer_engine.credential_source` Superposition config was intentionally introduced in the foundation milestone as an extensible enum (`none` / `application`) so that per-merchant/org/profile sources could be added later "alongside their credential storage". This PR adds the `merchant` source and that storage.

**What changes**

- **New `merchant` credential source** — `OfferEngineCredentialSource` gains a `Merchant` variant; the Superposition enum for `offer_engine.credential_source` now accepts `"merchant"`.
- **Per-merchant credential storage** — a new **encrypted** `offer_engine_config` column on `merchant_account` holds the merchant's Offer Engine `api_key` + Offer Engine `merchant_id` (encrypted via the merchant key store, exactly like `network_tokenization_credentials`). Settable through the merchant account update API (`POST /accounts/{id}`).
- **Resolution** — when the resolved source is `merchant`, credentials are read from the merchant account; `base_url` continues to come from the shared application config. Merchants without config fall through gracefully (offers disabled). An explicitly `merchant` source with missing/invalid config errors, mirroring the existing `application` behaviour.
- **Superposition key namespacing** — the keys are renamed to a dotted namespace (`offer_engine.enabled`, `offer_engine.credential_source`) so they group as an `offer_engine.` folder in the Superposition UI, consistent with the other config groups (`payments.`, `routing.`, …). The Rust key constants, DB-backed config keys, and the seed file are kept in sync.
- The `application` path is unchanged and both levels coexist: the default source can stay `application` while specific merchants are pointed at `merchant` via a Superposition override on `processor_merchant_id`.

### Additional Changes

- [x] This PR modifies the API contract — adds optional `offer_engine_config` to `admin::MerchantAccountUpdate` (+ `OfferEngineMerchantConfig` schema registered in `crates/openapi`).
- [x] This PR modifies the database schema — new `offer_engine_config BYTEA` column on `merchant_account` (`migrations/2026-08-24-100000_add_offer_engine_config_to_merchant_account`, `schema.rs`, `schema_v2.rs`).
[...]

PR #13858, opened 2026-08-26
feat(router): add token validation API for external services

spritianeja03:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

Adds `POST /external_service/validate/token`, which lets an external service verify a Hyperswitch dashboard JWT and find out who the user is and what they are allowed to do.

The endpoint is generic rather than Offer Engine specific. A `service` field in the request body selects how `merchant_id` and `permissions` are resolved, so another consumer can be added as a new match arm rather than a new endpoint.

**Request**

```json
{
  "token": "<hyperswitch dashboard JWT>",
  "service": "offer_engine"
}
```

**Response**

```json
{
  "merchant_id": "oe_merchant_abc",
  "context": "MERCHANT",
  "token": "<echoed back>",
  "permissions": ["ProfileOffersRead", "ProfileOffersWrite"]
}
```

The handler decodes the JWT, checks the blacklist so that signing out revokes a token that has not yet expired, verifies the tenant, and loads the user's role. For Offer Engine it then resolves the merchant id and the permissions.

A few notes on the behaviour:

- `merchant_id` is the **Offer Engine** merchant id, resolved through `resolve_offer_engine_config` — the same resolver the payment flow uses for `/offers/list` and `/offers/apply`, so the dashboard and payment paths cannot disagree on the mapping. That call doubles as the enablement gate: if Offer Engine is not enabled for the merchant there is no config to resolve, and the request is rejected with `403`.
- `context` is hardcoded to `MERCHANT` for now, until Offer Engine confirms which values they accept.
- Expired tokens return `IR_48` and invalid ones `IR_17`, kept distinct so the caller can tell a lapsed session apart from a bad token.

Also adds `Resource::Offers` with `OffersView` / `OffersManage` permission groups, wired the same way `Configurations` is. These are granted to `organization_admin`, `merchant_admin` and `profile_admin`. Every other predefined role gets an empty `permissions` array, which the caller is expected to treat as not authorised.

### Additional Changes

New endpoint only — no existing endpoint or response shape changes. The `OfferEngine` arm is added to an existing `#[serde(untagged)]` enum, so the Hypersense response is untouched.

`PermissionGroup` is stored as text, so the new variants need no migration. Worth noting it is roll-forward-only during a rolling deploy, since a role saved with the new groups cannot be deserialised by an older node.

## Motivation and Context

Offer Engine's offer-management components are being embedded in the Hyperswitch dashboard. The dashboard passes its session token to Offer Engine, which cannot read it because it is signed with Hyperswitch's secret, so it needs somewhere to have that token validated and to learn the user's merchant and permissions.

Hyperswitch already has this pattern for Hypersense, and previously had it for recon, but both minted a separate token first and then verified that. This validates the dashboard token directly, which means one endpoint instead of two and permissions computed fresh on every call rather than frozen into a token when it was issued.

Closes https://github.com/juspay/hyperswitch-cloud/issues/22988

## How did you test it?

Request
```
curl --location 'http://localhost:8080/external_service/validate/token' \
--data '{
      "token": "****",
      "service": "offer_engine"
  }'
```

Response - offer engine enabled
```json
{
    "merchant_id": "test_merchant",
    "context": "MERCHANT",
    "token": "****",
    "permissions": [
        "ProfileOffersRead",
        "ProfileOffersWrite"
    ]
}
```

Response - offer engine disabled
```json
{
    "error": {
        "type": "invalid_request",
        "message": "Access forbidden. Not authorized to access this resource offer_engine",
        "code": "IR_22"
    }
}
```

## Checklist


crates/router/src/core/external_service_auth.rs
```diff
@@ -92,3 +97,79 @@ pub async fn verify_external_token(
...
+    let role_info = authorization::get_role_info(&state, &payload).await?;
+
+    match req.service {
+        external_service_auth_api::ValidatingService::OfferEngine => {
+            let merchant_id = resolve_offer_engine_merchant_id(&state, &payload).await?;
+
+            Ok(service_api::ApplicationResponse::Json(
+                external_service_auth_api::ExternalVerifyTokenResponse::OfferEngine {
+                    merchant_id,
+                    context: "MERCHANT".to_string(),
```
apoorvdixit88: Please assign it to some const and then use it.

crates/router/src/core/external_service_auth.rs
```diff
@@ -92,3 +97,79 @@ pub async fn verify_external_token(
...
+async fn resolve_offer_engine_merchant_id(
+    state: &SessionState,
+    payload: &authentication::AuthToken,
+) -> RouterResult<String> {
+    let dimensions = Dimensions::new()
+        .with_processor_merchant_id(payload.merchant_id.clone().into())
+        .with_organization_id(payload.org_id.clone())
+        .with_profile_id(payload.profile_id.clone());
+
+    offer_engine::resolve_offer_engine_config(state, &dimensions)
```
apoorvdixit88, line 150: Here we will be getting conflicts with the PR, so we need to change accordingly: https://github.com/juspay/hyperswitch/pull/13831

crates/router/src/core/external_service_auth.rs
```diff
@@ -92,3 +97,79 @@ pub async fn verify_external_token(
...
+fn offer_engine_permissions(role_info: &roles::RoleInfo) -> Vec<String> {
+    [
+        Permission::ProfileOffersRead,
+        Permission::ProfileOffersWrite,
+    ]
+    .into_iter()
+    .filter(|permission| role_info.check_permission_exists(*permission))
+    .map(|permission| permission.to_string())
+    .collect()
+}
```
apoorvdixit88, line 172: nit: please check if we can derive these from Resource::Offers (via EnumIter on Permission) instead of hardcoding.

apoorvdixit88 (commented):
Please add the tests as well.

crates/api_models/src/external_service_auth.rs
```diff
@@ -24,12 +36,22 @@ pub enum ExternalVerifyTokenResponse {
         name: Secret<String>,
         email: pii::Email,
     },
+    OfferEngine {
+        merchant_id: String,
+        context: String,
+        token: Secret<String>,
+        /// Rendered from the router's `Permission` enum, e.g. `ProfileOffersRead`,
```
apoorvdixit88: nit: we can remove these comments

crates/router/src/core/external_service_auth.rs
```diff
@@ -1,17 +1,27 @@
...
     services::{
         api as service_api,
-        authentication::{self, ExternalServiceType, ExternalToken},
+        authentication::{self, blacklist::BlackList, ExternalServiceType, ExternalToken},
+        authorization::{self, permissions::Permission, roles},
     },
     SessionState,
 };
 
+/// Context reported to Offer Engine. Fixed until they confirm which values they accept.
```
apoorvdixit88: nit: please remove this comment

apoorvdixit88 (dismissed):
We can remove comments that over explain things

VenuMadhav2541 (dismissed):

crates/router/src/core/external_service_auth.rs
```diff
@@ -92,3 +102,76 @@ pub async fn verify_external_token(
...
+    payload: &authentication::AuthToken,
+) -> RouterResult<String> {
+    let dimensions = Dimensions::new()
+        .with_processor_merchant_id(payload.merchant_id.clone().into())
+        .with_organization_id(payload.org_id.clone())
+        .with_profile_id(payload.profile_id.clone());
+
+    offer_engine::resolve_offer_engine_config(state, &dimensions)
+        .await
+        .change_context(errors::ApiErrorResponse::InternalServerError)
```
prasunna09: A deployment with the flag enabled, source Application, but no offer_engine block in config gets a 500 here rather than the 403 documented for the not-enabled case — resolve_application_config returns MissingApplicationConfig, which is mapped to InternalServerError.

Every other caller of this resolver treats a resolution failure as "offers not available" (core/payments.rs:14971 and core/payment_methods/cards.rs:4922 both use .ok().flatten()). Worth folding the error into the same AccessForbidden path as the None case so the config gap doesn't 500 on every dashboard load.

apoorvdixit88 (dismissed):

prasunna09 (dismissed):

crates/router/src/routes/app.rs
```diff
@@ -1902,6 +1903,19 @@ impl Tokenization {
     }
 }
 
+pub struct ExternalService;
+
+impl ExternalService {
+    pub fn server(state: AppState) -> Scope {
+        web::scope("/external_service")
```
SanchithHegde: Consider using hyphenated path segments, that's what we've used with v2 APIs and newly introduced APIs.


```suggestion
        web::scope("/external-service")
```

Two other points, to keep consistency with v2 API endpoints:
- We _may_ have to use plurals, so we may have to use `external-services`. Not entirely sure of this.
- Use `/validate-token` over `/validate/token`, we've followed such convention with the v2 APIs. 
  Reference: https://github.com/juspay/hyperswitch/blob/32c30307aad3053d177998d0560d0f668ae7c834/crates/router/src/routes/app.rs#L859-L920

Please discuss these aspects about the API contract with the relevant folks before making the change.

crates/api_models/src/events/external_service_auth.rs
```diff
@@ -28,3 +28,9 @@ impl ApiEventMetric for ExternalSignoutTokenRequest {
         Some(ApiEventsType::ExternalServiceAuth)
     }
 }
+
+impl ApiEventMetric for ValidateTokenRequest {
+    fn get_api_event_type(&self) -> Option<ApiEventsType> {
+        Some(ApiEventsType::ExternalServiceAuth)
```
SanchithHegde: Nit: The `ExternalServiceAuth` could also include a field with the service name.

apoorvdixit88 (approved):

SanchithHegde (approved):

VenuMadhav2541 (approved):

prasunna09 (approved):

PR #13868, opened 2026-08-26
fix(connector-config): correct GoTyme payout auth key name

### LINKED ISSUE #13869: Replace the unsupported `profile_id` key with `key1` for the GoTyme Sanlam payout connector config

Saptak88:
## Type of Change

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

- Replace the unsupported `profile_id` key with `key1` for the GoTyme Sanlam payout connector
- Apply the correction across development, sandbox, and production connector configurations
- Align the configuration with the expected `BodyKey` authentication schema

### Additional Changes

## How did you test it?

Tested locally on control center
Screenshot 2026-08-26 at 6 55 01 PM

## Checklist

### DIFF
--- a/crates/connector_configs/toml/development.toml
+++ b/crates/connector_configs/toml/development.toml
@@ -9105,7 +9105,7 @@ key1 = "Worldpay Merchant ID"
 
 [gotyme_sanlam_payout.connector_auth.BodyKey]
 api_key = "API Key"
-profile_id = "Profile Id"
+key1 = "Profile Id"
 [[gotyme_sanlam_payout.bank_transfer]]
 payment_method_type = "payshap"
 [[gotyme_sanlam_payout.bank_transfer]]
--- a/crates/connector_configs/toml/production.toml
+++ b/crates/connector_configs/toml/production.toml
@@ -8161,7 +8161,7 @@ key1 = "Worldpay Merchant ID"
 
 [gotyme_sanlam_payout.connector_auth.BodyKey]
 api_key = "API Key"
-profile_id = "Profile Id"
+key1 = "Profile Id"
 [[gotyme_sanlam_payout.bank_transfer]]
 payment_method_type = "payshap"
 [[gotyme_sanlam_payout.bank_transfer]]
--- a/crates/connector_configs/toml/sandbox.toml
+++ b/crates/connector_configs/toml/sandbox.toml
@@ -9079,7 +9079,7 @@ key1 = "Worldpay Merchant ID"
 
 [gotyme_sanlam_payout.connector_auth.BodyKey]
 api_key = "API Key"
-profile_id = "Profile Id"
+key1 = "Profile Id"
 [[gotyme_sanlam_payout.bank_transfer]]
 payment_method_type = "payshap"
 [[gotyme_sanlam_payout.bank_transfer]]


Vani-1107 (approved):

Shubhodip900 (approved):

abhisheksharma2411 (commented):
This is correct, and I think it's considerably more important than the description makes it sound — worth saying so, because a three-line TOML change reads like a cosmetic label fix and this isn't one.

**`profile_id` doesn't merely fail to populate GoTyme's auth fields — it fails the whole file.**

`ConnectorAuthType` is externally tagged with `BodyKey { api_key, key1 }`, and `ConnectorConfig::new()` parses the entire TOML eagerly from a single `include_str!`:

```rust
let config = toml::from_str::<Self>(config_str);
```

The `toml` crate is stricter than plain serde here — it rejects unknown keys in a table rather than ignoring them. Running your exact before/after snippets through the same struct shapes:

```
BEFORE (profile_id, current main):
  PARSE FAILED -> TOML parse error at line 8, column 1
    | profile_id = "Profile Id"
    | ^^^^^^^^^^
    unexpected keys in table: profile_id, available keys: api_key, key1

AFTER (key1, this PR):
  OK -> gotyme=Some(BodyKey { api_key: "API Key", key1: "Profile Id" })
```

Because the parse is all-or-nothing, `ConnectorConfig::new()` returns `Err` for **every** connector while that key is present — so `get_connector_config` and `get_payout_connector_config` fail for Worldpay, Stripe and everything else, not just GoTyme Sanlam. That's a considerably bigger blast radius than "the GoTyme payout connector shows the wrong field", and it's worth putting in the PR body so this gets prioritised accordingly.

Caveat on my method: I reproduced against replica structs rather than compiling `connector_configs` (its pinned `toml 0.8.22` doesn't build standalone against current serde), using `toml 0.8.23` — same minor line, and the unknown-key rejection is 0.8 behaviour throughout. Happy for someone to confirm against the real crate; the direction won't change.

**The part I'd like to see added: nothing in CI can catch this class of bug.**

`crates/connector_configs` has **no tests at all** — `#[cfg(test)]` appears zero times across all five source files, and there's no `tests/` directory. The only consumers are `euclid_wasm` (the dashboard config surface) and the router's webhook-management transformer, both runtime paths. So a malformed TOML compiles cleanly, passes CI, and only surfaces when the dashboard asks for a connector config — which is exactly how this reached `main`.

A single smoke test would close it permanently:

```rust
#[test]
fn all_connector_configs_parse() {
    // The file is parsed all-or-nothing, so one bad key disables every connector.
    ConnectorConfig::new().expect("connector config TOML must parse");
}
```

run across the `production` / `sandbox` / default feature combinations so all three files are covered. Given that onboarding a connector means hand-editing the same stanza in three separate TOMLs, this will recur otherwise — and the three files can silently drift from each other too, which the same test shape would catch.

That's a follow-up, not a blocker. The change here is right and applied consistently to all three files; I'd merge it and open the test separately.

abhisheksharma2411:
Following up on my review with something I found while writing the test I suggested — it directly affects this PR, so flagging it here rather than leaving you to hit it.

**`production.toml` has a second, independent defect that this PR doesn't cover.** Running a parse test over all three files on `main` reports:

```
development.toml: unexpected keys in table: profile_id, available keys: api_key, key1
sandbox.toml:     unexpected keys in table: profile_id, available keys: api_key, key1
production.toml:  duplicate key `name` in table `santander.metadata.pix_automatico_push`
```

A missing `[[santander.metadata.pix_automatico_push]]` header merges two entries into one. `development.toml` and `sandbox.toml` both have that header, so it's a dropped line in one file.

The two are not alternatives — `production.toml` carries **both**, so it needs both fixes before it parses:

```
main as-is        ->  FAILED
your fix only     ->  FAILED  (production: duplicate key)
my fix only       ->  FAILED  (dev + sandbox: profile_id)
both              ->  ok
```

I've opened #13871 with the santander header plus the regression test, and deliberately left `profile_id` alone so there's no duplicate change — yours is the fix for that half and it should land as it is.

One practical note: my PR's test is red until this one merges, which is why I'd suggest merging yours first. Nothing needed from you beyond that; I'll rebase.

Your change is still correct and still the more urgent of the two, since it's the one affecting `development` and `sandbox` as well.

PR #13871, opened 2026-08-26
fix(connector-config): restore the santander pix_automatico_push array header in production.toml

### LINKED ISSUE #13874: connector_configs: production.toml fails to parse, disabling every connector's config lookup
## Bug

`crates/connector_configs/toml/production.toml` is missing the `[[santander.metadata.pix_automatico_push]]` header before the `pix_key_type` entry, so that entry merges into the preceding `client_secret` element and the file fails to parse:

```
production.toml: TOML parse error at line 6594, column 1
     |
6594 | name="pix_key_type"
     | ^
duplicate key `name` in table `santander.metadata.pix_automatico_push`
```

`development.toml` and `sandbox.toml` both carry the header in that position, so this is a dropped line in one file rather than an intentional difference.

## Impact is every connector, not santander

`ConnectorConfig::new()` parses a whole file in a single `toml::from_str`:

```rust
let config = toml::from_str::<Self>(config_str);
```

and both public entry points propagate the error with `?`. So while this is present, a **production** build cannot load *any* connector's config — `get_connector_config` and `get_payout_connector_config` fail for every connector, including ones with nothing wrong in their own stanza.

`toml` is stricter than plain serde here: it rejects unknown keys in a table rather than ignoring them, which is what turns a single malformed stanza into a total parse failure.

## Why CI does not catch this

`crates/connector_configs` has **no tests at all** — `#[cfg(test)]` appears zero times across all five source files, and there is no `tests/` directory. The only consumers are `euclid_wasm` and the router's webhook-management transformer, both runtime paths. So a malformed TOML compiles cleanly, passes CI, and surfaces only when the dashboard asks for a connector config.

This is the second such defect found in the same files. #13868 (merged) fixed an `unexpected keys in table: profile_id` error in all three TOMLs; this one is independent and affects `production.toml` only. Both were present simultaneously, so `production.toml` needed both fixes before it would parse.

## Steps to reproduce

```
cargo test -p connector_configs --features v1
```

with a parse test over all three shipped TOMLs. (`--features v1` is required — `-p` scopes features to the package, and without it `api_models` fails to compile with unrelated `E0432`/`E0599`.)

## Expected

All three shipped connector TOMLs deserialize, and a regression test fails loudly if one stops doing so.

## Environment

Reproduced on `main` at `4e31a5f25`, and on the current head after #13868 merged.

abhisheksharma2411:
## Type of Change

- [x] Bugfix

Closes #13874

## Description

`production.toml` is missing the `[[santander.metadata.pix_automatico_push]]` header before the `pix_key_type` entry, so that entry merges into the preceding `client_secret` element:

```toml
[[santander.metadata.pix_automatico_push]]
name="client_secret"
...
type="Text"
name="pix_key_type"          # <-- no header, so this joins the element above
```

`development.toml` and `sandbox.toml` both carry the header in that position, and the sibling `santander.metadata.pix_qr` block carries it too — so this is a dropped line in one file rather than an intentional difference. @bsayak03 [confirmed](https://github.com/juspay/hyperswitch/pull/13871#discussion_r3868662555) it was unintentional and got missed in #12185.

**The consequence is not scoped to santander.** `ConnectorConfig::new()` parses a whole file in a single `toml::from_str`, and both public entry points propagate the error with `?`:

```rust
let config = toml::from_str::<Self>(config_str);
```

So while this is present, a **production** build cannot load *any* connector's config — `get_connector_config` and `get_payout_connector_config` fail for every connector.

## Verification

Parsed both revisions with an independent TOML parser rather than asserting from inspection:

```
main (unfixed)   ->  Cannot overwrite a value (at line 6717, column 20)
with this fix    ->  parses; pix_automatico_push has 7 elements:
                     client_id, client_secret, pix_key_type, pix_key_value,
                     account_number, account_type, branch_code
```

Rebased onto current `main` (`7e51536dc`).

## Scope

One line. The regression test that originally came with this has been dropped at @pixincreate's request — happy to bring it back separately if it is ever wanted.

### DIFF
--- a/crates/connector_configs/toml/production.toml
+++ b/crates/connector_configs/toml/production.toml
@@ -6714,6 +6714,7 @@ label="Client Secret"
 placeholder="Enter your Client Secret"
 required=true
 type="Text"
+[[santander.metadata.pix_automatico_push]]
 name="pix_key_type"
 label="Chave Key Type"
 placeholder="Enter your Chave/Pix Key Type"


abhisheksharma2411:
Rebased onto current `main` now that **#13868 has merged** — this PR's test is green.

Re-ran the whole matrix on the rebased head rather than assuming, since the point of this PR was that the two defects are complements:

```
main (with #13868, without this PR)   ->  FAILED
                                          duplicate key `name` in table
                                          `santander.metadata.pix_automatico_push`

main (with #13868, with this PR)      ->  ok. 1 passed; 0 failed
```

The failing run is with my one-line header removed and the removal confirmed first (7 array headers → 6), so the test is still doing work rather than passing because the other half landed. `profile_id` is now absent from all three TOMLs, so #13868's half is confirmed in place too.

Net effect: with both merged, all three shipped connector TOMLs parse for the first time, and `crates/connector_configs` has a test that will say so if that stops being true.

Thanks to @Saptak88 — merging that one first is what unblocked this.

crates/connector_configs/src/connector.rs
pixincreate, line 1: we don't need this.

crates/connector_configs/toml/production.toml
```diff
@@ -6591,6 +6591,7 @@ label="Client Secret"
 placeholder="Enter your Client Secret"
 required=true
 type="Text"
+[[santander.metadata.pix_automatico_push]]
```
pixincreate, line 6717: @bsayak03, was this intentional?

bsayak03: non-intentional. Somehow got missed in this [PR](https://github.com/juspay/hyperswitch/pull/12185)

crates/connector_configs/src/connector.rs
abhisheksharma2411: Dropped it — thanks, and no argument from me; it is your call what the crate carries.

For what it is worth if it ever comes up again: that test is how I found this, and @bsayak03 has since confirmed the line was unintentional and got missed in #12185. Same shape of thing could go unnoticed again, since a malformed TOML compiles cleanly. But that is a separate conversation from this fix, and I would rather not hold up a one-line production fix over it. Happy to open it as its own PR if you ever want it, or to leave it alone.

abhisheksharma2411:
`Deja ART Replay` went red on this head at 07:24 today. Flagging it because I cannot diagnose it from here and do not want to guess in a way that wastes someone else's time.

What I can see: the check reports `Verdict: FAILED.` with **no annotations and no log** — `details_url` points back at this PR, and `hyperswitch-bot` has not left a comment. So there is nothing on my side to act on yet.

Two things that might help whoever owns the bot:

- This check does not appear on **any** other pull request I can see — not on the 15 most recently updated open PRs, nor on the 20 most recently merged. This PR is the only one carrying it, which makes me wonder whether it is newly enabled rather than newly broken.
- The diff is one line: a `[[santander.metadata.pix_automatico_push]]` table header. Worth noting that *without* it `production.toml` does not parse at all — `ConnectorConfig::new()` returns `Err` for every connector, which I checked against an independent TOML parser. So if the replay loads production connector configs, I would expect it to be failing on `main` too rather than only here. I genuinely cannot tell which, and that is a guess, not a finding.

Happy to re-run, rebase, or dig in properly if someone can point me at the replay log.

pixincreate (approved):

abhisheksharma2411:
Rebased onto current `main` — this was approved on 28 Aug and had drifted 170 commits behind, which is what `mergeable_state: behind` was reporting. Clean merge, no conflicts, and the diff is still the same single line.

**The reason it is worth landing rather than closing: `production.toml` on `main` does not parse right now.**

```python
>>> tomllib.load(open("crates/connector_configs/toml/production.toml","rb"))
TOMLDecodeError: Cannot overwrite a value (at line 6717, column 20)
```

Line 6717 is exactly the missing `[[santander.metadata.pix_automatico_push]]` header this restores. With the fix applied the file parses and `santander.metadata` has all four arrays:

```
pix_qr, pix_automatico_push (7 entries), pix_automatico_qr, boleto
```

Without it, those seven fields have no table header to belong to, so the parser hits the previous key and stops.

@bsayak03 confirmed in review that the omission was unintentional and got missed in #12185, and @pixincreate's point about `connector.rs` was taken — that file is out of the diff entirely, so this is `+1/-0` on `production.toml` alone. `cargo test -p connector_configs --features v1` passes on the merged head.

Worth noting the gap this sat in: the malformed config has been on `main` for the whole 25 days and nothing in CI reports it, which is what #14208 (still open) adds a build-time guard for. That one is the durable fix; this is the one-line repair.

Nothing outstanding from me — happy to rebase again if it drifts.

PR #13889, opened 2026-08-27
feat(connector): [WorldpayWPG][Checkout] Add support for AFT

Vani-1107:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

- Added funding_transaction_type and payment_purpose as enums under connector_metadata.worldpayxml, and purpose_of_payment under connector_metadata.checkout, since these values are connector-specific and not modelled in core
-  Passed down the business country because worldpayxml requires taxId only for Brazilian merchants
-  Checkout: sets processing.aft, and builds recipient, sender (from billing) and instruction.purpose; the account union flattens to Checkout's single account_number, rejecting the identifier types it doesn't document
- Worldpay WPG: emits <fundingTransfer> with sender and recipient <fundingParty> blocks, mapping each account variant to its accountType code; supported on authorize and complete-authorize

### Additional Changes

## How did you test it?

Can't test as we do not have the creds yet

## Checklist


XyneSpaces:
[blocking] `Order.funding_transfer` needs the XML element rename before this ships. Without `#[serde(rename = "fundingTransfer")]`, `quick-xml` will emit `<funding_transfer>`, so Worldpay will not receive the AFT funding-transfer block.

[blocking] `CheckoutSender.date_of_birth` needs explicit ISO date serialization. The connector expects a `YYYY-MM-DD` string; use the existing project serialization pattern for `time::Date` or convert to a string wrapper before serializing.

[blocking] `customer_date_of_birth` is hard-coded to `None` in every new `RouterData` constructor. Since Checkout and WorldpayWPG read this field for AFT sender DOB, the value must be populated from the request/customer data on all relevant construction paths.

deepanshu-iiitu (dismissed):

maverox (dismissed):

crates/hyperswitch_connectors/src/connectors/worldpayxml/transformers.rs
```diff
@@ -446,6 +449,103 @@ struct WorldpayxmlPayinAddress {
...
+
+#[derive(Debug, Serialize, Deserialize)]
+struct WorldpayxmlDate {
+    #[serde(rename = "@dayOfMonth")]
+    day_of_month: String,
+    #[serde(rename = "@month")]
+    month: String,
+    #[serde(rename = "@year")]
+    year: String,
+}
```
SanchithHegde, line 535: Consider wrapping these fields in `Secret` instead?

Vani-1107: Yes, have addressed it

deepanshu-iiitu (dismissed):

maverox (dismissed):

prasunna09 (dismissed):

prasunna09 (approved):

maverox (approved):

SanchithHegde (approved):

PR #13919, opened 2026-08-30
feat(revenue_recovery): add DisputeRecordBack flow and implement it for Chargebee

### LINKED ISSUE #13917: [FEATURE] Record a lost dispute back to the billing connector for revenue recovery
### Feature Description

When a revenue recovery payment succeeds, hyperswitch records it back to the billing connector, marking the invoice as paid. If a dispute on that payment is later lost, the money has left the merchant's account — but nothing tells the billing connector. The invoice still shows as paid and the merchant's books are wrong.

Chargebee will only accept an offline refund against the transaction id it returns from `record_payment`. Today that id is discarded: `ChargebeeRecordbackResponse` deserializes only the `invoice` object from a response that also carries a `transaction` object. Because the id is never stored, there is no way to correct the invoice after a dispute is lost.

We need to capture that identifier at record-back time and add a connector flow that spends it when a dispute is lost.

### Possible Implementation

1. **Capture the id.** Add `transaction` to `ChargebeeRecordbackResponse` and surface `connector_transaction_id` on `InvoiceRecordBackResponse`. Recurly and Stripebilling build the same response type and set it to `None`.

2. **Persist it.** Store it on `payment_attempt.feature_metadata.revenue_recovery`. That column is already `Jsonb`, so no migration is required. The write must merge into the existing recovery metadata rather than replace it, so `attempt_triggered_by` and `charge_id` survive. A persist failure should not fail the recovery flow — the payment succeeded and the invoice was recorded; losing the id degrades a future refund, it does not invalidate the payment.

3. **Add a `DisputeRecordBack` connector flow**, following `InvoiceRecordBack`: flow marker, request/response types, `FlowData`, `RevenueRecoveryDisputeRecordBack{,V2}` traits as supertraits of `RevenueRecovery{,V2}`, type aliases, and default unimplemented impls for every connector except the one implementing it.

4. **Implement it for Chargebee** against `v2/invoices/{invoice-id}/record_refund`. Note this is the invoice-level endpoint, not the transaction-level one: `v2/transactions/{id}/record_refund` caps its amount at the transaction's `amount_unused`, which is always zero for a payment applied to an invoice in full. Send `payment_method = chargeback` so Chargebee's reporting attributes the refund to the dispute.

5. **Gate the call on a configured connector list**, following the existing `[billing_connectors_invoice_sync]` pattern, so connectors without an offline-refund API are skipped rather than failing on the default unimplemented impl.

### Notes

Chargebee does not enable `record_refund` on live sites by default — their support must enable it per site before this works in production. The connector allowlist means the code can ship ahead of that.

NISHANTH1221:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

Adds a `DisputeRecordBack` connector flow for revenue recovery, implemented for Chargebee, so that a dispute lost on a recovery payment can be recorded back to the billing connector as an offline refund.

Three parts:

**1. Capture the billing connector's transaction id.**

When a recovery payment succeeds, `record_back_to_billing_connector` posts to Chargebee's `v2/invoices/{invoice-id}/record_payment`. Chargebee's response carries two top-level objects, `invoice` and `transaction`, but only `invoice` was deserialized — so `transaction.id`, Chargebee's handle for the payment we just recorded, was thrown away.

It is now surfaced on `InvoiceRecordBackResponse` and persisted onto `payment_attempt.feature_metadata.revenue_recovery.billing_connector_transaction_id`. The write **merges** into the existing recovery metadata rather than replacing it, so `attempt_triggered_by` and `charge_id` survive. A persist failure is logged and metered but does not fail the recovery flow — the payment succeeded and Chargebee recorded it; losing the id degrades a future dispute refund, it does not invalidate the payment.

Recurly and Stripebilling build the same response type and set the new field to `None`.

**2. The `DisputeRecordBack` flow.**

A new connector flow following `InvoiceRecordBack` exactly — flow marker, request/response types, `FlowData`, the `RevenueRecoveryDisputeRecordBack{,V2}` traits as supertraits of `RevenueRecovery{,V2}`, type aliases, and default unimplemented impls for every connector except Chargebee.

Chargebee targets **`v2/invoices/{invoice-id}/record_refund`**. This is deliberate and was corrected during development: the transaction-level `v2/transactions/{id}/record_refund` caps its amount at the transaction's `amount_unused`, which Chargebee defines as "the part of the `amount` which has not been invoiced yet". A payment created by `record_payment` is applied to an invoice in full, so `amount_unused` is always zero and that endpoint rejects every call with:

```
This operation is not supported as this payment does not have any unused amount.
```

The invoice-level endpoint is the one that reverses a payment already applied to an invoice. Its transaction fields nest under `transaction[...]`, matching `record_payment`. `payment_method` is sent as `chargeback` — one of Chargebee's accepted values and an accurate description of what happened — so Chargebee's own reporting attributes the refund to the dispute rather than lumping it under `other`. The stored transaction id travels as `transaction[reference_number]` so the refund record points back at the payment it reverses.

**3. A connector allowlist.**

Only some billing connectors expose an offline-refund API. Without a gate, a lost dispute on a Recurly or Stripebilling recovery payment would reach the default unimplemented impl and be metered as a failure. `[billing_connectors_dispute_record_back]` follows the existing `[billing_connectors_invoice_sync]` pattern.

There is **no caller in this PR**. The v2 dispute incoming webhook that invokes this flow is stacked on top of this branch in a separate PR, because the call site lives inside the webhook handler.

### Additional Changes

**No database migration.** `payment_attempt.feature_metadata` is already a `Jsonb` column, so the new field is additive within existing JSON. Rows written before this change deserialize the missing key as `None`; there is a unit test asserting exactly that rather than assuming it.

Configuration files changed:

- [config/config.example.toml](config/config.example.toml)
- [config/development.toml](config/development.toml)
- [config/docker_compose.toml](config/docker_compose.toml)
- [config/deployments/production.toml](config/deployments/production.toml)
- [config/deployments/sandbox.toml](config/deployments/sandbox.toml)
- [config/deployments/integration_test.toml](config/deployments/integration_test.toml)
- [loadtest/config/development.toml](loadtest/config/development.toml)
- [crates/router/src/configs/settings.rs](crates/router/src/configs/settings.rs)
- [crates/router/src/configs/secrets_transformers.rs](crates/router/src/configs/secrets_transformers.rs)

New block, identical in each:

```toml
[billing_connectors_dispute_record_back]
billing_connectors_which_requires_dispute_record_back_call = "chargebee"
```

## Motivation and Context

When a revenue recovery payment succeeds, hyperswitch tells the billing connector the invoice is paid. If a dispute on that payment is later lost, the money has left the merchant's account — but nothing tells Chargebee. The invoice still shows as paid and the merchant's books are wrong.

Chargebee will only accept an offline refund against an identifier we were discarding, so there was no way to correct this after the fact. This PR captures that identifier and adds the connector flow that spends it.

Closes #13917

## How did you test it?

**Unit tests** (10 added):

- `diesel_models::payment_attempt::recovery_data_tests` — recovery metadata written before this change deserializes with the new field as `None`; the merge preserves `attempt_triggered_by` and `charge_id`.
- `chargebee::transformers::chargebee_recordback_tests` — a `record_payment` body carrying both `invoice` and `transaction` parses, and one carrying only `invoice` still parses.
- `chargebee::transformers::chargebee_dispute_record_back_tests` — every transaction field nests under `transaction[...]` with no flat duplicate left behind; an absent comment is omitted; the wire body asserted through `serde_urlencoded` (what `RequestContent::FormUrlEncoded` actually serializes with) contains `transaction%5Bpayment_method%5D=chargeback` and no `=other`.
- `router::configs::settings::hashset_deserialization_test` — the shipped `"chargebee"` value parses and gates Recurly and Stripebilling out; an unknown connector name is rejected rather than silently ignored.

**Compile gates:** `just check_v2` and `just check` both pass. This branch was verified to build standalone, with no caller present.

**Not tested against a live Chargebee site.** Two things a reviewer should weigh:

1. Chargebee does **not** enable `record_refund` on live sites by default — their support must enable it per site. Until that is done, the call will fail in production regardless of this code. The connector allowlist exists partly so this can be shipped ahead of that.
2. `credit_note[create_reason_code]` is not sent. Chargebee's docs say it is required only if reason codes are mandatory in a given site's setup, which cannot be determined from here. If the first live call fails complaining about a reason code, that is the cause and it is a one-field addition.

## Checklist


XyneSpaces:
[blocking] `persist_billing_connector_transaction_id` silently swallows the `update_payment_attempt` DB error.

In `crates/router/src/core/revenue_recovery/types.rs` (~line 1544), the `if let Err(error) = state.store.update_payment_attempt(...)` block logs the error and returns `()`, so a failed write loses the billing-connector transaction id. A later lost-dispute webhook will then skip the record-back refund because `billing_connector_transaction_id` is `None`.

Change the function to return `RecoveryResult<()>` and propagate the error with `?` (or add a dedicated metric/alert if intentional degradation is required).

XyneSpaces:
[should-fix] Validate or truncate `reference_number` to Chargebee's 100-character cap.

`crates/hyperswitch_connectors/src/connectors/chargebee/transformers.rs` documents that Chargebee caps `transaction[reference_number]` at 100 characters, but the `TryFrom<&DisputeRecordBackRouterData>` impl clones the full `billing_connector_transaction_id` into `reference_number` without length guarding. A long transaction id will cause a connector-side rejection.

Fix: truncate or validate the value before assigning it.

[should-fix] Use a semantic `updated_by` value in `persist_billing_connector_transaction_id`.

`crates/router/src/core/revenue_recovery/types.rs` passes `storage_scheme.to_string()` as the `updated_by` field of `PaymentAttemptUpdate::RecordBackUpdate`. That field is meant to identify the actor/source of the update, not the storage backend. Use a constant such as `"revenue_recovery"` instead.

XyneSpaces:
[should-fix] The new async helper `persist_billing_connector_transaction_id` in `crates/router/src/core/revenue_recovery/types.rs` is missing `#[instrument(skip_all)]`. Async functions in the core layer should carry tracing instrumentation so the storage update and error paths are observable in production traces.

XyneSpaces:
[blocking] New required config field breaks existing deployments

`crates/router/src/configs/settings.rs` adds `billing_connectors_dispute_record_back` without `#[serde(default)]`. Any existing config file that does not include the `[billing_connectors_dispute_record_back]` table will fail to deserialize and the router will refuse to start.

Make the field default-capable so old configs keep working:
```rust
    #[serde(default)]
    pub billing_connectors_dispute_record_back: BillingConnectorDisputeRecordBackCall,
```

XyneSpaces:
[should-fix] Missing `#[instrument]` on new async core helper

`crates/router/src/core/revenue_recovery/types.rs` introduces `persist_billing_connector_transaction_id` as a new async function in the revenue-recovery core but does not annotate it with `#[instrument]`. This leaves it out of the router's tracing/observability setup.

Add the standard instrumentation:
```rust
#[instrument(skip_all)]
async fn persist_billing_connector_transaction_id(
    state: &SessionState,
    payment_attempt: &PaymentAttempt,
    revenue_recovery_payment_data: &storage::revenue_recovery::RevenueRecoveryPaymentData,
    connector_transaction_id: Option<&str>,
) {
```

XyneSpaces:
[should-fix] Persisting the billing connector transaction id silently drops DB errors

In `crates/router/src/core/revenue_recovery/types.rs`, `persist_billing_connector_transaction_id` logs a failure from `update_payment_attempt` but does not propagate it or surface any alert. Because this id is exactly what the later `DisputeRecordBack` flow needs to identify the payment transaction, a transient DB write failure here permanently loses the mapping.

Consider making the helper fallible (returning a `Result`) and letting the caller emit a metric or alarm, or at least add a dedicated counter/metric for persistence failures so operators can detect the inconsistency.

jarnura (approved):

rohanm13 (approved):

PR #13920, opened 2026-08-30
feat(router): implement the v2 dispute incoming webhook flow

### LINKED ISSUE #13918: [FEATURE] Implement the v2 incoming webhook flow for disputes
### Feature Description

The v2 incoming webhook dispatcher has no dispute handling. `api::WebhookFlow::Dispute` is a `todo!()` in `crates/router/src/core/webhooks/incoming_v2.rs`, so a dispute webhook arriving on a v2 merchant panics rather than being recorded.

v1 has had this flow for a long time (`crates/router/src/core/webhooks/incoming.rs`). v2 needs the equivalent: resolve the payment attempt, map the incoming event to a dispute status, and insert or update the dispute row.

Two things make this more than a copy of v1:

1. **The dispute model is version-split.** The row carries `PaymentId` and a `String` attempt id in v1, but `GlobalPaymentId` and `GlobalAttemptId` in v2. The model, the Kafka event structs and the storage interface all need splitting by API version.

2. **v2 can only resolve a payment attempt by connector transaction id.** v1 looks the dispute up by payment id; v2 has no such path from a webhook, so it needs a lookup by profile id and connector transaction id.

This is also the natural home for the dispute record-back call from #13917 — the trigger is a dispute transitioning to `DisputeLost`, which only this flow observes.

### Possible Implementation

1. **Split the dispute model by API version** — `diesel_models::dispute`, the Kafka `KafkaDispute` / `KafkaDisputeEvent` structs, and the storage interface. Use two whole `cfg`-gated definitions per version rather than field-level `cfg` attributes, following the precedent in `diesel_models/src/refund.rs`.

2. **Add the v2 dispute storage methods**, including lookup by profile id and connector transaction id.

3. **Implement `disputes_incoming_webhook_flow`** in `incoming_v2.rs`, replacing the `todo!()`. Resolve the attempt, map the event to a `DisputeStatus`, and insert or update the dispute row via `get_or_update_dispute_object`.

4. **Invoke the dispute record-back** on the transition into `DisputeLost`. This needs an idempotency guard: `validate_dispute_status` permits `DisputeLost -> DisputeLost`, so a redelivered webhook would otherwise issue a second refund. Evaluating `Dispute::is_not_lost_or_none` against the row *as it was before the update* makes the call fire exactly on the transition and never on a redelivery.

### Notes

The idempotency guard has a deliberate consequence worth recording: because a redelivery hits the guard and skips, **failing the webhook buys no retry**. A record-back failure should therefore be logged and metered rather than propagated. Moving to a process-tracker task later would change only the call site.

Out of scope for the initial implementation, and worth tracking separately: v2 sends no outgoing webhook to merchants on the dispute path (v1 does), and there are no v2 dispute REST endpoints — `impl Disputes` in `routes/app.rs` is gated `#[cfg(all(feature = "olap", feature = "v1"))]`.

NISHANTH1221:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

> **Stacked on #13919.** The base of this PR is `feat/v2-dispute-record-back`, not `main`. Review that one first — this PR's diff is only its own two commits. GitHub will retarget this to `main` automatically once #13919 merges.

Implements the v2 incoming webhook flow for disputes, replacing the `todo!()` on `api::WebhookFlow::Dispute`, and wires in the `DisputeRecordBack` flow added by #13919.

**1. The dispute model split by API version.**

The dispute row carries different id types per version — `PaymentId` and a `String` attempt id in v1, `GlobalPaymentId` and `GlobalAttemptId` in v2. The diesel model, the Kafka `KafkaDispute` / `KafkaDisputeEvent` structs and the storage interface are each split into two whole `cfg`-gated definitions rather than gating individual fields, following the precedent in `diesel_models/src/refund.rs`. `from_storage` and the `KafkaMessage` impls stay shared.

Adds the v2 dispute storage methods the flow needs, including lookup by profile id and connector transaction id — the only handle a v2 dispute webhook can resolve a payment attempt by.

**2. The webhook flow.**

`disputes_incoming_webhook_flow` resolves the payment attempt from the webhook's connector transaction id, maps the incoming event to a `DisputeStatus`, and inserts or updates the dispute row.

**3. The dispute record-back call.**

On the transition into `DisputeLost`, the flow records the dispute back to the billing connector as an offline refund, using the transaction id stored at record-back time. The call is gated three ways:

- the attempt must carry a stored `billing_connector_transaction_id` — absent for any payment that never went through revenue recovery, which is the common case and skips silently;
- the billing connector must appear in the configured `[billing_connectors_dispute_record_back]` list;
- the dispute must not already have been lost.

That last guard is the important one. `validate_dispute_status` permits `DisputeLost -> DisputeLost`, so a redelivered webhook passes validation and re-runs the flow — without a guard, that is a duplicate refund. `Dispute::is_not_lost_or_none` is evaluated against the row **as it was before the update**, so the refund fires exactly on the transition and never on a redelivery. v1 uses the same guard to gate its intent state-metadata update, so this is an established pattern here rather than a new invention.

Neither skip path counts toward `DISPUTE_RECORD_BACK_SUCCESS_METRIC`; only a call the billing connector accepted does.

### Additional Changes

No migration and no config changes. The dispute table already exists; this PR only splits how the row is modelled per API version. The configuration this flow reads (`[billing_connectors_dispute_record_back]`) is introduced by #13919.

## Motivation and Context

A dispute webhook arriving on a v2 merchant hits `todo!()` and panics. v2 has had no dispute handling at all, so disputes were neither recorded nor acted on.

This is also the only place that observes a dispute transitioning to `DisputeLost`, which is the trigger the record-back flow from #13919 needs.

Closes #13918

## How did you test it?

**Unit tests** (4 added):

- `connector_transaction_id_is_extracted` — the connector transaction id is pulled out of a `ConnectorTransactionId` object reference.
- `unsupported_object_reference_is_rejected` — every other `ObjectReferenceId` variant is rejected rather than silently mishandled, since v2 has no other lookup path.
- `every_dispute_event_maps_to_a_status` — all seven dispute events route to `WebhookFlow::Dispute` and have a `DisputeStatus`, so the flow cannot reject a webhook the dispatcher already committed to handling.
- `dispute_record_back_fires_only_on_the_transition_into_lost` — covers the guard's `None` and not-yet-lost branches.

**Compile gates:** `just check_v2`, `just check` and `just clippy_v2` all pass.

**Known gaps a reviewer should weigh:**

1. **The idempotency guard is not fully covered by tests.** The case that matters most — a stored row already `DisputeLost` suppressing a redelivery — is not exercised, because constructing a full `Dispute` row is required. It is covered indirectly by `is_not_lost_or_none`'s own contract. A `MockDb` test running the same webhook twice is the right way to close this and is the first thing I would add.

2. **The record-back has no retry, by design.** Because a redelivery hits the guard and skips, failing the webhook would buy no retry — only a duplicate dispute update. So a failure is logged at error level, increments `DISPUTE_RECORD_BACK_FAILURE_METRIC`, and returns success. This is alertable but not self-healing, and it is a deliberate trade for a simpler call path. Moving to a process-tracker task later would change only the call site; the guard, the connector flow and the stored id all carry over unchanged.

3. **No live connector test.** See #13919 for the Chargebee site prerequisite.

**Out of scope, and worth tracking separately:** v2 still sends no outgoing webhook to merchants on the dispute path, unlike v1, and there are no v2 dispute REST endpoints (`impl Disputes` in `routes/app.rs` is `#[cfg(all(feature = "olap", feature = "v1"))]`). So a v2 merchant has no way to learn a dispute exists. That predates this PR, but this PR does not fix it.

## Checklist


XyneSpaces:
Review findings from the v2 dispute incoming webhook flow:

[should-fix] `crates/router/src/core/webhooks/incoming_v2.rs` (~line 934): the original billing-connector error is logged but then replaced with a fresh `report!(...)`, dropping the error_stack chain. Use `.change_context(errors::ApiErrorResponse::WebhookProcessingFailure).attach_printable("billing connector rejected the dispute record-back")` instead.

[should-fix] `crates/router/src/core/webhooks/incoming_v2.rs` (~line 839): `record_dispute_back_to_billing_connector` is an async hot-path function without `#[instrument]`. Add `#[instrument(skip_all, fields(dispute_id = %dispute.dispute_id, payment_id = %payment_attempt.payment_id))]`.

[should-fix] `crates/router/src/core/webhooks/incoming_v2.rs` (~line 628): `get_or_update_dispute_object` is a non-trivial async helper without `#[instrument]`. Add `#[instrument(skip_all, fields(connector_dispute_id = %dispute_details.connector_dispute_id))]`.

XyneSpaces:
[should-fix] The new async helpers `get_or_update_dispute_object` and `record_dispute_back_to_billing_connector` in `crates/router/src/core/webhooks/incoming_v2.rs` are missing `#[instrument(skip_all)]`. Other async functions in this file (`incoming_webhooks_core`, `disputes_incoming_webhook_flow`) already carry it; add the annotation so these dispute-webhook paths are visible in traces.

```suggestion
#[instrument(skip_all)]
async fn get_or_update_dispute_object(...)

#[instrument(skip_all)]
async fn record_dispute_back_to_billing_connector(...)
```

XyneSpaces:
[blocking] Several new async functions in `crates/router/src/core/webhooks/incoming_v2.rs` are missing `#[instrument(skip_all)]`, which the baseline rules require for async core functions:

- `get_payment_attempt_from_object_reference_id`
- `get_or_update_dispute_object`
- `record_dispute_back_to_billing_connector`

Please add `#[instrument(skip_all)]` (with relevant fields such as `dispute_id`/`payment_id` where available) so the dispute webhook flow is observable in traces.

(Inline diff comments are not available in this environment; findings are posted as a PR-level comment with file references.)

PR #13924, opened 2026-08-31
test(cypress): add Ilixium connector coverage

### LINKED ISSUE #13925: test(cypress): follow-ups for Ilixium connector coverage (save card, partial amounts, non-card methods)
### Description

Follow-up items surfaced while adding initial Cypress coverage for the Ilixium connector (#13924). Ilixium is UCS-only — its local `crates/hyperswitch_connectors` transformer is a stub that always returns `NotImplemented`, so all behavior below was verified live against `integ.hyperswitch.io` rather than derived from source.

### Known gaps not covered by #13924

- **Save card flow**: excluded via `CONNECTOR_LISTS.EXCLUDE.SAVE_CARD` in `Utils.js` rather than implemented. Ilixium requires `customer.dateOfBirth` (schema-mandatory on the connector-service side) for save-card confirms, which isn't wired into the save-card test path. Needs either connector-service metadata support in that flow or a documented reason it can't apply.
- **Mandates**: not supported by the connector at all (`SetupMandate` is `not_implemented`, `MandateRevoke` is `not_supported` in `connector-service`'s `ilixium.rs`) — confirmed at the code level, no coverage expected here, just tracking that this was checked and is a genuine connector limitation, not a gap.
- **Bank transfers / bank redirects / wallets / UPI / etc.**: Ilixium only supports raw card payments on `POST /direct/auth`; non-card payment methods return a real "not implemented" error. `17-BankTransfers.cy.js` is excluded via `CYPRESS_EXCLUDE_SPEC_PATTERN`; other non-card specs (BankRedirect, Wallet, UPI, RewardPayment, CryptoPayment, GiftCardPayment, RealTimePayment) haven't been run against ilixium yet and may need the same treatment.
- **Partial capture / partial refund amounts**: `PartialCapture`'s exact `amount_capturable`/`amount_received` shape and `manualPaymentPartialRefund`'s `"succeeded"` status assumption are inferred from the full-capture/full-refund cases, not independently verified against a dedicated live run.
- **3DS refund flow** (`09-RefundPayment.cy.js`'s `"Card - Refund flow - 3DS"` describe block): now unskipped after the `handleRedirection` frictionless-3DS fix, but not yet re-verified end-to-end with a live run.
- **PSync/RSync accuracy**: Ilixium's sync flows are bulk, time-windowed reconciliation reports (`POST /history/operations`, max 24h window) filtered client-side, with a documented limitation distinguishing one partial refund from another by `merchantRef`/`gatewayRef` alone (see `transformers::IlixiumHistoryResponse::match_refund_operation` in connector-service). Not specifically stress-tested here.

### Motivation

Keep these follow-ups visible rather than losing them in PR discussion once #13924 merges.

venkatakarthikm-maker:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [x] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

Adds initial Cypress coverage for the Ilixium connector (UCS-only — its local `crates/hyperswitch_connectors` transformer is a stub that always returns `NotImplemented`, so this repo alone can't be used to derive its behavior).

Covered and live-verified against `integ.hyperswitch.io`:
- No3DS / 3DS manual-capture confirm, full Capture, partial Capture, Void
- Full and partial refund (No3DS path, confirm → capture → refund chain)
- Documented-broken automatic capture (No3DS and 3DS variants — the connector doesn't support `capture_method: automatic` with these credentials; encoded as an expected `"failed"`/`error_code: "4"` response rather than skipped, so a future connector fix would surface as a real, actionable test failure)
- `handleRedirection` fix: Ilixium's 3DS confirm can be frictionless (no challenge, no `next_action`) even though 3DS itself is genuinely implemented by the connector (see `ilixium.rs`'s `PreAuthenticate`/`Authorize` split in connector-service). The shared `handleRedirection` command previously fell back to a fake `example.com` redirect and asserted a host change unconditionally, even in live mode — that fallback is only meaningful for cassette record/replay (per its own comment). Scoped the fallback to record/replay mode only; live mode now skips cleanly on a genuinely missing redirect, matching the existing behavior of the two sibling commands `handleCardRedirectRedirection`/`handleBankRedirectRedirection`.

Not covered (tracked in #13925): save card (blocked on `ilixium_date_of_birth` metadata not being wired into that flow), mandates (confirmed unsupported by the connector itself, not a gap), non-card payment methods beyond bank transfers, and independent verification of partial-capture/partial-refund exact amounts.

### Additional Changes

## Motivation and Context

Ilixium had no Cypress coverage. Related follow-up work tracked in #13925.

## How did you test it?

All new/changed assertions were verified against real request/response pairs from live runs against `integ.hyperswitch.io`, not guessed from other connectors' defaults — see inline comments in `Ilixium.js` for the evidence trail on each entry.

## Checklist

image
image
image


cypress-tests/cypress/e2e/configs/Payment/Ilixium.js
```diff
@@ -0,0 +1,287 @@
...
+          card: verifiedCardDetails,
+        },
+        currency: "USD",
+        customer_acceptance: null,
+        metadata: ilixiumMetadata,
+      },
+      Response: {
+        status: 200,
+        body: {
+          status: "failed",
```
likhinbopanna, line 120: can you add the comment ?

cypress-tests/cypress/e2e/configs/Payment/Ilixium.js
```diff
@@ -0,0 +1,413 @@
...
+        payment_method: "card",
+        amount: 1000,
+        payment_method_data: {
+          card: verifiedCardDetails,
+        },
+        currency: "USD",
+        customer_acceptance: null,
+        metadata: ilixiumMetadata,
+      },
+      // Creds we currently have only supports manaul capture. Therefore mapped error code for auto capture.
```
likhinbopanna: spell check

likhinbopanna (approved):

vasanthp-jus (approved):

PR #13926, opened 2026-08-31
fix(revenue-recovery): invoke adaptive retry algo

### LINKED ISSUE #13923: [FEATURE] Invoke adaptive retry algorithm alongside static ladder in revenue recovery Smart arm
## Feature Description

Revenue recovery schedules every retry from a fixed cascading ladder — a per-merchant array of `(delay, count)` pairs stored in Superposition under `pt_mapping_pcr_retries`. The ladder is applied blindly: the Nth retry always waits the Nth configured delay, regardless of why the payment failed or when retries for that failure reason actually succeed.

Two pieces of the missing signal already exist but have no consumer:

- **#13793** records per-cluster retry outcomes in `revenue_recovery_retry_stats`, keyed by standardised decline code, card type and issuer, with day-of-week / day-of-month / hour-of-day success counts.
- **#13804** adds `compute_adaptive_retry_time`, a MathModel predictor that turns those statistics into a proposed retry time given a remaining grace window and retry budget.

Nothing calls the predictor. The `Smart` arm of `RevenueRecoveryAlgorithmType` exists but currently returns `None`, so every merchant falls back to cascading.

The gap is a scheduler that consults both and picks between them, without giving up the guarantees the static ladder provides: a bounded number of retries, spread across the grace period, at times the merchant configured.

### Goals

- Use the adaptive prediction when it lands on an **earlier calendar day** than the ladder would.
- Never let adaptive push a retry *later* than the ladder — it may only pull one forward.
- Spend each static position exactly once, however many adaptive retries are inserted between them.
- Count retries the billing connector already made against the invoice's allowance, so an invoice reaching recovery late resumes part-way down rather than starting over.
- Degrade to today's behaviour whenever the model has nothing to say — no decline code resolved, no recorded history for that cluster, or an exhausted grace window or budget.

### Non-goals

- Changing the `Cascading` arm. Merchants not opted into `Smart` are unaffected.
- Replacing the static ladder. It remains the floor.

---

### 1. A decision function

A pure function over the two candidates, so the policy is testable without a database:

```rust
pub fn decide_next_retry(
    schedule: &StaticLadderProgress,
    queried_rung: i32,
    static_time: Option<PrimitiveDateTime>,
    adaptive_time: Option<PrimitiveDateTime>,
) -> Option<ScheduleDecision>
```

| static | adaptive | outcome | ladder position |
|---|---|---|---|
| `Some` | `Some` | earlier calendar day; **tie goes to static** | advances iff static won |
| `Some` | `None` | static | advances |
| `None` | `Some` | adaptive | unchanged |
| `None` | `None` | nothing schedulable → invoice ends | — |

Comparing **calendar days** rather than instants is deliberate: if the ladder already covers a day, an earlier hour on that same day should not earn a second attempt. Ties going to static keeps the configured schedule authoritative.

A `static_time` of `None` means the ladder is spent. That should not be fatal — adaptive can carry the invoice for the rest of its grace window. Today it surfaces as `EApiErrorResponse` → `GLOBAL_ERROR` with no webhook, which is indistinguishable from a crash.

### 2. Tracking the ladder position

The position cannot be derived from `process_tracker.retry_count`, which counts every CALCULATE run including ones where adaptive won. Following it would exhaust a five-entry ladder long before the fifth static retry.

Store it in the CALCULATE row's tracking data instead — that row is reopened rather than recreated, so it survives the whole recovery lifecycle:

```rust
pub struct StaticLadderProgress {
    pub consumed_rungs: i32,
}
```

`Option<StaticLadderProgress>` with `#[serde(default)]`, so rows written before the field existed deserialize as `None` and are seeded on first use.

### 3. Seeding from what the connector already spent

A new field on the billing connector's `RevenueRecoveryMetadata`:

```rust
pub max_hybrid_cascading_retry_count: u16,
```

- **New invoice**, at row creation: `min(intent_retry_count, cap)`
- **Pre-existing invoice**, on first CALCULATE run: `min(intent_retry_count, cap - 1)`

The `- 1` for existing invoices is a hedge: by then `total_retry_count` has been incrementing on recovery's own retries too, and the connector's share can't be recovered from it. Holding one position back guarantees a cascading retry survives however inflated the count is.

### 4. Supplying the model's inputs

```
remaining_grace_days = (invoice_billing_started_at_time ?? intent.created_at)
                       + grace_period_days - now
remaining_budget     = max_retry_count - retry_count
```

Saturating throughout — an expired window or spent budget yields `0`, and the model declines rather than an unsigned subtraction wrapping to a value that reads as unlimited.

### 5. Configuration

| Value | Source |
|---|---|
| `adaptive_retry_enabled` | Superposition, per merchant + connector |
| `recovery_grace_period_days` | Superposition |
| `max_retry_count` | billing connector MCA |
| `max_hybrid_cascading_retry_count` | billing connector MCA |
| retry ladder | Superposition `pt_mapping_pcr_retries` |

No migration required: both new fields live in existing JSONB columns.

rohanm13:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

Wires the MathModel adaptive retry predictor into the revenue-recovery `Smart` arm, so an invoice's next retry is the earlier of the cascading (static) ladder and the model's prediction.

**Calling the predictor.** The `Smart` arm previously had a stubbed `adaptive_time = None`. It now calls `compute_adaptive_retry_time(state, error_code, remaining_grace_days, remaining_budget)` and converts the returned `OffsetDateTime` to `PrimitiveDateTime` at the schedule boundary. `prev_attempt_error_code` comes off the tracking data, already resolved via GSM at webhook ingestion.

**Deciding between the two.** `decide_next_retry` takes both candidates as `Option` and returns `Option<ScheduleDecision>`:

| static | adaptive | outcome | ladder position |
|---|---|---|---|
| `Some` | `Some` | earlier calendar day; **tie goes to static** | advances iff static won |
| `Some` | `None` | static | advances |
| `None` | `Some` | adaptive | unchanged |
| `None` | `None` | nothing schedulable → invoice ends | — |

An exhausted ladder is no longer fatal. Previously `get_delay` returning `None` became `EApiErrorResponse` → `GLOBAL_ERROR`; now adaptive can carry the invoice for the rest of its grace window, and the invoice terminates on grace/budget instead of a ladder cliff.

**Seeding the ladder.** `static_ladder_progress` is now `Option<StaticLadderProgress>` and seeded from what the billing connector already spent, so an invoice reaching recovery late resumes part-way down rather than replaying positions:

- new invoice, at row creation: `min(intent_retry_count, max_hybrid_cascading_retry_count)`
- pre-existing invoice, on first CALCULATE run: `min(intent_retry_count, cap − 1)` — one position held back, because for an invoice already in recovery `total_retry_count` includes our own retries and can't be separated from the connector's

**Allowances.** `get_adaptive_retry_allowances` computes `remaining_grace_days` from `(invoice_billing_started_at_time ?? payment_intent.created_at) + grace_period_days − now`, and `remaining_budget` from `max_retry_count − retry_count`. Saturating throughout, so an expired window or spent budget yields `0` and the model declines rather than wrapping.

### Additional Changes

`max_hybrid_cascading_retry_count: u16` added to `RevenueRecoveryMetadata` on the merchant connector account:

- `crates/api_models/src/admin.rs`
- `crates/diesel_models/src/merchant_connector_account.rs`
- `crates/hyperswitch_domain_models/src/merchant_connector_account.rs`

No migration: the field lives in the existing `feature_metadata` JSONB, and `static_ladder_progress` in `process_tracker.tracking_data`. Both carry `#[serde(default)]` on the storage layers so existing rows deserialize unchanged.

## Motivation and Context

Revenue recovery currently schedules retries from a fixed cascading ladder — a per-merchant array of `(delay, count)` pairs applied blindly. It ignores when retries for a given decline reason actually succeed.

The MathModel predictor (#13804) and its retry-stats store (#13793) supply that signal but had no caller. This connects them, while keeping the static ladder as a floor: adaptive only wins by proposing a **strictly earlier calendar day**, so it can pull a retry forward but never push one back, and every static position is still spent exactly once.

## How did you test it?

**Unit tests** — 19 in `crates/router/src/core/revenue_recovery/schedule.rs`, covering seeding (below cap / at-or-past cap, for both new and existing invoices, plus an unconfigured cap of `0` that must not underflow), the four decision arms including both `None` combinations, rung consumption across successive decisions, and the serde round-trip with absent fields.

```
cargo test --no-default-features --features "v2,redis-rs" -p router --lib core::revenue_recovery::schedule
test result: ok. 19 passed; 0 failed
```

**End-to-end**, against a local Postgres with router + producer + consumer, on a pre-existing CALCULATE row with no ladder state:

```
Adaptive retry decision  source: Static  queried_rung: 3
  static_time: 2026-08-31 18:45:35  adaptive_time: None
  error_code: Some(InsufficientFunds)  remaining_grace_days: 0  remaining_budget: 14
```

```
status | business_status              | progress
finish | CALCULATE_WORKFLOW_SCHEDULED | {"consumed_rungs": 3}
```

This confirmed back-fill seeding, the MCA budget read (`15 − 1`), error-code threading, and that the position is persisted in the same write that finishes the row. `adaptive_time` was `None` because the local `revenue_recovery_retry_stats` table was empty and no grace-period config was set — the adaptive branch itself has not yet been observed returning a time in a live run.

## Checklist

### DIFF
--- a/crates/router/src/core/revenue_recovery.rs
+++ b/crates/router/src/core/revenue_recovery.rs
@@ -121,6 +121,13 @@ pub async fn upsert_calculate_pcr_task(
                 payment_id.get_string_repr()
             );
 
+            let max_hybrid_cascading_retry_count = billing_connector_account
+                .get_max_hybrid_cascading_retry_count()
+                .ok_or(errors::RevenueRecoveryError::RetryCountFetchFailed)
+                .attach_printable(
+                    "Failed to get max hybrid cascading retry count from billing merchant connector account",
+                )?;
+
             // Create tracking data
             let calculate_workflow_tracking_data = pcr::RevenueRecoveryWorkflowTrackingData {
                 billing_mca_id: billing_connector_account.get_id(),
@@ -134,7 +141,7 @@ pub async fn upsert_calculate_pcr_task(
                 static_ladder_progress: Some(
                     schedule::StaticLadderProgress::seed_for_new_invoice(
                         intent_retry_count,
-                        billing_connector_account.get_max_hybrid_cascading_retry_count(),
+                        max_hybrid_cascading_retry_count,
                     ),
                 ),
             };
@@ -592,11 +599,9 @@ pub async fn perform_payments_sync(
 }
 
 /// `attempts_already_made` counts the initial charge, which is not a retry, so the retry about to
-/// be scheduled is retry number `attempts_already_made`. No ceiling configured means no gate.
-fn is_retry_budget_exhausted(attempts_already_made: i32, max_retry_count: Option<u16>) -> bool {
-    max_retry_count
-        .map(|max_retry_count| attempts_already_made > i32::from(max_retry_count))
-        .unwrap_or(false)
+/// be scheduled is retry number `attempts_already_made`.
+fn is_retry_budget_exhausted(attempts_already_made: i32, max_retry_count: u16) -> bool {
+    attempts_already_made > i32::from(max_retry_count)
 }
 
 pub async fn perform_calculate_workflow(
@@ -662,22 +667,31 @@ pub async fn perform_calculate_workflow(
     )
     .await?;
 
-    let static_ladder_progress = tracking_data
-        .static_ladder_progress
-        .clone()
-        .unwrap_or_else(|| {
+    let static_ladder_progress = match tracking_data.static_ladder_progress.clone() {
+        Some(static_ladder_progress) => static_ladder_progress,
+        None => {
+            let max_hybrid_cascading_retry_count = revenue_recovery_payment_data
+                .billing_mca
+                .get_max_hybrid_cascading_retry_count()
+                .ok_or(errors::RecoveryError::ValueNotFound)
+                .attach_printable(
+                    "Failed to get max hybrid cascading retry count from billing merchant connector account",
+                )?;
             schedule::StaticLadderProgress::seed_for_existing_invoice(
                 payment_intent.get_revenue_recovery_retry_count(),
-                revenue_recovery_payment_data
-                    .billing_mca
-                    .get_max_hybrid_cascading_retry_count(),
+                max_hybrid_cascading_retry_count,
             )
-        });
+        }
+    };
 
     // 2. Bound the invoice by the merchant's ceiling, independently of the retry ladder.
     let max_retry_count = revenue_recovery_payment_data
         .billing_mca
-        .get_max_retry_count();
+        .get_max_retry_count()
+        .ok_or(errors::RecoveryError::ValueNotFound)
+        .attach_printable(
+            "Failed to get max retry count from billing merchant connector account",
+        )?;
 
     let (payment_processor_token_response, next_static_ladder_progress) =
         if is_retry_budget_exhausted(process.retry_count, max_retry_count) {
@@ -702,7 +716,7 @@ pub async fn perform_calculate_workflow(
                 process.retry_count,
                 &tracking_data,
                 &static_ladder_progress,
-                revenue_recovery_payment_data.billing_mca.get_max_retry_count(),
+                max_retry_count,
         )
             .await
             {
--- a/crates/router/src/core/revenue_recovery.rs
+++ b/crates/router/src/core/revenue_recovery.rs
@@ -419,6 +419,7 @@ pub async fn perform_execute_payment(
                         storage::ProcessTrackerRunner::PassiveRecoveryWorkflow,
                         tracking_data.revenue_recovery_retry,
                         state.conf.application_source,
+                        tracking_data.static_ladder_progress.clone(),
                     )
                     .await?;
 
@@ -492,6 +493,7 @@ async fn insert_psync_pcr_task_to_pt(
     runner: storage::ProcessTrackerRunner,
     revenue_recovery_retry: diesel_enum::RevenueRecoveryAlgorithmType,
     application_source: common_enums::ApplicationSource,
+    static_ladder_progress: Option<schedule::StaticLadderProgress>,
 ) -> RouterResult<storage::ProcessTracker> {
     let task = PSYNC_WORKFLOW;
     let process_tracker_id = payment_attempt_id.get_psync_revenue_recovery_id(task, runner);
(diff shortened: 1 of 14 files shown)


XyneSpaces:
Automated review findings (posted as PR-level comment because inline review-comment creation is unavailable in this run):

1. `crates/router/src/workflows/revenue_recovery.rs:672-677`
:warning: Grace-window arithmetic failures are silently coerced to 0 via `.unwrap_or_default()`.

Replace the chained `.unwrap_or_default()` calls with explicit handling: log/alert on `checked_add` overflow and clamp negative day counts with `.max(0)` before converting to `u32`, so a misconfigured grace period does not hide as a zero remaining window.



2. `crates/router/src/workflows/revenue_recovery.rs:679-682`
:warning: `remaining_budget` integer conversion failure is silently discarded.

Handle the `try_into` result explicitly instead of using `.unwrap_or_default()`; the merchant's retry budget should not silently become 0 on an unexpected value.

crates/router/src/core/revenue_recovery.rs
```diff
@@ -486,7 +489,7 @@ async fn insert_psync_pcr_task_to_pt(
         revenue_recovery_retry,
         invoice_scheduled_time: Some(schedule_time),
         // PSYNC has its own row; scheduling state lives on CALCULATE.
-        static_ladder_progress: schedule::StaticLadderProgress::default(),
+        static_ladder_progress: None,
```
NISHANTH1221: Reopen caluclate happens at two p-lace one at the end of execute and one at the end of psync. Is this is being made to none, can we verify while reopenning calculate this field is being populated as none

crates/router/src/core/revenue_recovery.rs
```diff
@@ -1054,7 +1074,7 @@ async fn insert_execute_pcr_task_to_pt(
                 revenue_recovery_retry,
                 invoice_scheduled_time: Some(schedule_time),
                 // EXECUTE has its own row; scheduling state lives on CALCULATE.
-                static_ladder_progress: schedule::StaticLadderProgress::default(),
+                static_ladder_progress: None,
```
NISHANTH1221: Similar to psync

jarnura (dismissed):

NISHANTH1221 (dismissed):

NISHANTH1221 (approved):

jarnura (approved):

PR #13953, opened 2026-09-01
fix(ucs): let the gRPC boundary stand in for a transport replay never…

maverox:
… uses

build_grpc_client! connects each UCS service client eagerly and returns None from build_connections when that connect fails, so one unreachable service nulls the whole unified_connector_service_client. That is the right thing for a deployment: check_ucs_availability then reports Disabled, every request takes ExecutionPath::Direct, and the merchant is served by the direct connector integration rather than routed at a service the router cannot reach.

It is the wrong thing under replay. Every unary rpc is substituted at the deja boundary and never reaches the wire, so the connect is ceremony that cannot succeed in a namespace where UCS is deliberately absent, and its failure removes UCS shadow mode from the run. Replaying a tape recorded with shadow mode on then reports the shadow Authorize rpc, and the comparison POST that follows it, as omissions in every run — including runs driven by the tape's own revision. The comparison never ran, because the client was already None when the first request arrived.

The obvious fix is to build the channel lazily, and it is the wrong one. An eager connect is a legitimate choice, and a recorder that requires the host to abandon it has made itself an argument in the host's transport design. It would also put record and production on different paths: a recording taken over a lazy channel could contain UCS traffic that the eager deployment it claims to stand for would never have produced. A recorder that changes what it observes is a worse defect than the gap it closes.

So the accommodation belongs in the boundary, which is already the deja-specific part. DejaGrpcTransport now holds an optional inner transport. Under replay it never touched that transport in any case — substitution happens in call, before the inner service is reached — so the boundary can stand in for a transport that was never built. connect_or_substitute is generic over the transport and over the host's connect closure: it runs that connect verbatim in every mode except replay, where no rpc is issued live. The UCS constructor keeps its eager connect, keeps returning None when the connect fails, and no longer expresses any opinion about deja's modes.

A boundary with no transport needs its own fail-stop. Reusing the substitute-miss message would tell a reader that re-running is unsafe and to declare replay_strategy = Execute, a remedy that does not exist when there is nothing to run. The two states also put the burden on different people: a miss with the transport present may be a genuine novel call whose author needs to look at it, while a miss with the transport absent establishes nothing about the candidate at all. The new message says so, and names the rpc path and authority, because the boundary and component are module constants shared with the dynamic routing, health check and recovery decider pool and so cannot say which transport was missing.

poll_ready delegates to the inner service ungated, which is safe only because tonic answers readiness from its buffer rather than from a connection. The no-transport case does not delegate at all; the delegating case is left as it was, with the reasoning recorded at the call site.

## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

### Additional Changes

## Checklist

### DIFF
--- a/crates/external_services/src/grpc_client/deja_transport.rs
+++ b/crates/external_services/src/grpc_client/deja_transport.rs
@@ -40,17 +40,139 @@ const OPERATION: &str = "unary";
 /// The transport wrapper. Generic over the inner tower service so the same
 /// type serves the shared hyper pool (dynamic routing / health / recovery)
 /// and the UCS `tonic::transport::Channel`s.
+///
+/// The inner transport is optional because a host may build its transport
+/// eagerly and fallibly — `tonic::transport::Endpoint::connect` is a perfectly
+/// reasonable choice — and under replay that connect is ceremony: every unary
+/// rpc is substituted here and the transport beneath is never driven. Rather
+/// than ask such a host to become lazy (which would change what it does in
+/// production, and make a recording diverge from the thing it records), the
+/// boundary can stand in for a transport that was never built. See
+/// [`DejaGrpcTransport::substituted`] and [`connect_or_substitute`].
 #[derive(Debug, Clone)]
 pub struct DejaGrpcTransport<S> {
-    inner: S,
+    inner: Option<S>,
 }
 
 impl<S> DejaGrpcTransport<S> {
-    /// Wraps a transport. Call at the construction site, never per request.
+    /// Wraps a live transport. Call at the construction site, never per request.
     pub fn new(inner: S) -> Self {
-        Self { inner }
+        Self { inner: Some(inner) }
+    }
+
+    /// A boundary with no transport beneath it.
+    ///
+    /// Valid ONLY where every call is substituted, i.e. replay. Nothing here is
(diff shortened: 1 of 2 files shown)


maverox:
Superseded by a two-PR stack — closing this one.

This PR and #14001 became two live views of the same branch. `work/ucs-lazy-connect` now carries two commits instead of the original squashed one, so this PR was showing the whole +262/-43 change against `main` while #14001 shows only the second commit. Merging here would have landed the pair unsplit and bypassed the first PR entirely.

The work is unchanged — same final tree, split into two reviewable pieces and rebased onto current `main`:

- **#14000** — `refactor(deja): let the gRPC boundary hold an optional transport`. `deja_transport.rs` only, and **behaviour-neutral**: nothing constructs a transport-less boundary yet. The design argument lives here.
- **#14001** — `fix(ucs): keep the shadow comparison path reachable when UCS is unreachable`. `unified_connector_service.rs` only, stacked on #14000. The actual behaviour change, one file.

Both commits were gated independently (`just clippy`, `just clippy_v2` on rustc 1.97.1, and `cargo check --features release` on 1.85.0 — all exit 0), so #14000 is green on its own rather than only as half of the pair.

Please review #14000 first.

PR #13955, opened 2026-09-01
fix(router): populate card_network for wallet payment attempts

ayush22667:
## Type of Change

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

`payment_attempt.card_network` is a derived column — recomputed from `payment_method_data` on every
write, rather than set by anyone. Every derivation was hardcoded to the card shape:

```rust
.and_then(|pm| pm.get("card"))
.and_then(|card| card.get("card_network"))
```

`AdditionalPaymentData` is externally tagged, so a wallet attempt stores
`{"wallet": {"apple_pay": {"network": "MasterCard", ...}}}`. The `card` lookup misses, the chain
short-circuits, and the column is written `NULL`. `dsl::card_network.eq_any(...)` then matches
nothing, which is why filtering payments by card network returns no wallet results.

Two things were needed:

1. **Extraction** — read the wallet shape as well as the card shape. Apple Pay names the field
   `network`; Google Pay and Samsung Pay use `card_network`.
2. **Normalisation** — providers send their own spellings. The filter binds `CardNetwork`, whose
   Postgres text form is the bare variant name, so storing the raw `"MasterCard"` would never match
   `'Mastercard'`.

Spellings seen in production over 30 days, and their volume:

| Spelling | Attempts |
|---|---|
| `Visa` / `VISA` | 44,119 |
| `MasterCard` / `MASTERCARD` / `Mastercard` | 25,466 |
| `AMEX` / `AmEx` / `AMERICAN EXPRESS` | 1,782 |
| `DISCOVER` / `Discover` | 508 |
| `MAESTRO`, `Interac` | 5 |

Changes by crate:

- **`common_enums`** — `strum(serialize)` entries on `CardNetwork` for each spelling above.
  `to_string` is pinned to the variant name on every variant that gains a `serialize`, so `Display`
  (this enum's database representation) and `VariantNames` are unchanged; only `FromStr` widens, and
  only additively. Added `CardNetwork::from_payment_method_data` for the update path.
- **`api_models`** — `AdditionalPaymentData::get_card_network()`, covering the card variant and all
  three card-backed wallets.
- **`hyperswitch_domain_models`** — `extract_card_network()` now handles wallets for v1, and is
  implemented for v2 (it was `todo!()`).
- **`diesel_models`, `storage_impl`, kafka event builders** — use the shared extraction instead of
  their own inline card-only JSON walks.

### Additional Changes

The public API contract is deliberately unchanged: normalisation is applied via `strum`'s `FromStr`,
not serde, so the set of `card_network` values accepted on request bodies is exactly as before.

## Motivation and Context

Filtering payments by card network returns nothing for wallet payments. Measured on production over
30 days: **71,880 wallet attempts carry an extractable network, and none of them had
`card_network` populated.**

Worth flagging for review: `extract_card_network()` has 11 pre-existing callers in `retry.rs`,
`payment_response.rs`, `payments.rs`, `helpers.rs` and `payment_sync.rs`. All of them were silently
receiving `None` for wallet payments and will now receive a real network, so retry and GSM decisions
change for wallet traffic.

Two follow-ups, both out of scope here:

- Existing rows are not backfilled — the column starts populating for new attempts only.
- v2 has `card_network` commented out of `PaymentAttemptUpdateInternal`, so it still never updates
  the column after insert.

## Linked Issues

- Issue: https://github.com/juspay/hyperswitch-cloud/issues/22260

## How did you test it?

Verified against production data rather than synthetic cases: the accepted spellings were taken from
`hyperswitch_prod.payment_attempts` over 30 days, so the set covers real traffic instead of guesses.
The same query established the baseline for this bug — 71,880 wallet attempts with an extractable
network, none of them populated.

`just check` passes under `v1`.

Post-deploy this can be confirmed by re-running the baseline query and checking the populated count
moves off zero, with the empty bucket shrinking to roughly the PayPal + AliPay share (~9% of wallet
attempts, which legitimately carry no network):

```sql
SELECT card_network, count() AS attempts
FROM hyperswitch_prod.payment_attempts
PREWHERE created_at >= now() - INTERVAL 1 DAY AND payment_method = 'wallet'
GROUP BY card_network ORDER BY attempts DESC
```

## Checklist

### DIFF
--- a/crates/api_models/src/payments.rs
+++ b/crates/api_models/src/payments.rs
@@ -4314,6 +4314,36 @@ impl AdditionalPaymentData {
             _ => None,
         }
     }
+
+    /// Wallet providers report the network as a free-form string in their own spelling, so it is
+    /// normalised through `CardNetwork`'s `FromStr`. Apple Pay names the field `network`, the others
+    /// `card_network`.
+    pub fn get_card_network(&self) -> Option<common_enums::CardNetwork> {
+        use std::str::FromStr;
+
+        match self {
+            Self::Card(additional_card_info) => additional_card_info.card_network.clone(),
+            Self::Wallet {
+                apple_pay,
+                google_pay,
+                samsung_pay,
+            } => apple_pay
+                .as_ref()
+                .map(|apple_pay| apple_pay.network.as_str())
+                .or_else(|| {
+                    google_pay
+                        .as_ref()
+                        .and_then(|google_pay| google_pay.card_network.as_deref())
+                })
+                .or_else(|| {
+                    samsung_pay
+                        .as_ref()
+                        .and_then(|samsung_pay| samsung_pay.card_network.as_deref())
+                })
+                .and_then(|network| common_enums::CardNetwork::from_str(network).ok()),
+            _ => None,
+        }
+    }
 }
 
 #[derive(Debug, Clone, Eq, PartialEq, serde::Deserialize, serde::Serialize)]
--- a/crates/common_enums/src/enums.rs
+++ b/crates/common_enums/src/enums.rs
@@ -3391,26 +3391,42 @@ pub enum MandateStatus {
 #[smithy(namespace = "com.hyperswitch.smithy.types")]
 pub enum CardNetwork {
     #[serde(alias = "VISA")]
+    #[strum(to_string = "Visa", serialize = "VISA")]
     Visa,
     #[serde(alias = "MASTERCARD")]
+    #[strum(
+        to_string = "Mastercard",
+        serialize = "MasterCard",
+        serialize = "MASTERCARD"
+    )]
     Mastercard,
     #[serde(alias = "AMERICANEXPRESS")]
     #[serde(alias = "AMEX")]
+    #[strum(
+        to_string = "AmericanExpress",
+        serialize = "AMEX",
+        serialize = "AmEx",
+        serialize = "AMERICAN EXPRESS"
+    )]
     AmericanExpress,
     JCB,
     #[serde(alias = "DINERSCLUB")]
     DinersClub,
     #[serde(alias = "DISCOVER")]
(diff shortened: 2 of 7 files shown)


crates/api_models/src/payments.rs
```diff
@@ -4314,6 +4314,36 @@ impl AdditionalPaymentData {
...
+                    google_pay
+                        .as_ref()
+                        .and_then(|google_pay| google_pay.card_network.as_deref())
+                })
+                .or_else(|| {
+                    samsung_pay
+                        .as_ref()
+                        .and_then(|samsung_pay| samsung_pay.card_network.as_deref())
+                })
+                .and_then(|network| common_enums::CardNetwork::from_str(network).ok()),
```
hrithikesh026: you can add a error log here

crates/common_enums/src/enums.rs
```diff
@@ -3391,26 +3391,42 @@ pub enum MandateStatus {
...
     Visa,
     #[serde(alias = "MASTERCARD")]
+    #[strum(
+        to_string = "Mastercard",
+        serialize = "MasterCard",
+        serialize = "MASTERCARD"
+    )]
     Mastercard,
     #[serde(alias = "AMERICANEXPRESS")]
     #[serde(alias = "AMEX")]
```
hrithikesh026, line 3448: check is this is needed.

hrithikesh026 (dismissed):

su-shivanshmathur (approved):

crates/common_enums/src/enums.rs
ayush22667: Yes, it is needed as strum is being for json so that list filters can use that.

hrithikesh026 (approved):

PR #13966, opened 2026-09-01
feat(alerts): scaffold the alerting crate, its config, and the API key guard

ThisIsMani:
## Type of Change

- [x] New feature

## Description

Stands up `crates/alerts` as a running, guarded, **empty** service — the socket the alert-delivery
work plugs into. No notifier logic yet; this is the scaffold, its configuration, and the internal
API key guard.

Targets the long-lived `observability-plane-i` branch, not `main`.

### `alerts` is the domain; notification is its first tenant

Not two names for one thing. Further alerting concerns are expected to live *beside* `notifier`,
not inside it — so `src/notifier.rs` exists now as an empty, documented seam rather than being
carved out later. Deciding what is alert-worthy stays out of this crate entirely.

### Shape

Lib + binary on the `drainer` model: `cargo run -p alerts` serves its own `HttpServer`, and the
crate exposes actix `Scope` factories so the router can mount the same route definitions in-process
later. Only the standalone path is wired up here — the router integration is deliberately deferred
so it gets designed with the router in hand.

### Authentication follows the router, not middleware

The obvious choice was actix middleware wrapping the `Scope`: a route added later cannot forget a
`.wrap()`. The router's own idiom is better. `server_wrap` takes `api_auth: &dyn
AuthenticateAndFetch` as a **required positional argument** — so a route cannot skip auth either
(omitting it is a compile error) *and* each route still chooses. Middleware has the first property
and not the second.

`alerts` mirrors this with `services::server_wrap(state, request, payload, handler, auth)` and an
`Authenticate` trait with `InternalApiKeyAuth` / `NoAuth` impls. Ours yields `()`; the router's
returns a merchant context the handler consumes, which is precisely why *it* cannot be middleware.

Health lives in its own **unguarded scope** rather than as a path exception inside the guard —
matching the router, whose `Health` scope does not go through `server_wrap`. A path-exception list
inside a guard is where auth bypasses come from; someone eventually adds `/healthz` or a trailing
slash. Two scopes makes "unauthenticated" structural and visible in the route tree.

This matters because when the crate is mounted in the router, the router serves a **single**
`HttpServer` on its public port — there is no internal-only listener, so anything mounted there is
publicly reachable.

### Errors, in three layers

Mirroring the router's own layering: `ConfigurationError` (boot, never rendered) → a semantic
`AlertsError` carried in an `error_stack::Report` → a wire `ApiErrorResponse` keyed to HTTP status,
bridged by `common_utils::errors::ErrorSwitch`. The full report reaches the log while the client
sees only the wire shape, so internal detail cannot leak into a response by accident. The router
keeps its wire layer in `api_models`; `alerts` has no equivalent crate, so it is
`alerts::errors::{types,actix}`, mirroring `api_models/src/errors/{types,actix}.rs` file-for-file.

### Configuration

`config/alerts.toml`, env prefix `ALERTS__` with `__` separating levels. A **dedicated** file
rather than `drainer`'s reliance on `Default` impls: that only works when the defaults are
genuinely runnable (localhost postgres, localhost redis), and ours cannot be — there is no sensible
default for an API key.

`Settings` is generic over `SecretState` from the start, with a `secrets_transformers.rs`, so KMS
handles resolve at boot in release builds and the channel tokens landing next add a section to an
existing pattern instead of restructuring it.

Boot fails loudly on a missing key. **There is deliberately no way to disable authentication** — an
empty `internal_api_key` is not "open", it is a refusal to start.

### On the `v1`/`v2` features

`alerts` has no version-flavoured type of its own and this crate has no opinion on API versioning.
The features exist purely to forward: the repo's secrets-management layer is reachable only through
`hyperswitch_interfaces` → `hyperswitch_domain_models` → `api_models` + `diesel_models`, and those
do not compile unless a flavour is selected. Defaulted to `v1`, matching the router, so
`cargo run -p alerts` works without anyone having to think about it.

## Additional Changes

Adds `config/alerts.toml`. No existing configuration is touched.

## Motivation and Context

Resolves juspay/hyperswitch-cloud#23114.

First build ticket of the observability-plane alert-delivery work. Everything downstream — the
Notifier interface, the chat and email notifiers, the router mount — needs a crate, a config
module, and a guard to exist first.

Scope deliberately pushed to later tickets rather than half-designed here: the router integration
(and with it the embedded-mode config split), and the email/chat config sections, which cannot be
shaped sensibly without the notifiers that consume them.

## How did you test it?

Manually, against the running binary rather than asserted:

| Case | Result |
|---|---|
| `GET /health`, no credentials | `200` |
| `GET /alerts/ping`, no key | `401` `{"error":{"type":"invalid_request","message":"API key not provided or invalid","code":"IR_01"}}` |
| `GET /alerts/ping`, wrong key | `401`, identical body — the response does not distinguish absent from wrong |
| `GET /alerts/ping`, valid key | `200` `{"message":"alerts is reachable"}` |
| Rejection logging | `WARN` carrying request id, path, peer address; the supplied key appears **zero** times in the log |
| Boot with empty `internal_api_key` | refuses to start |

```console
$ curl -s -w '\nstatus=%{http_code}\n' http://127.0.0.1:8085/alerts/ping
{"error":{"type":"invalid_request","message":"API key not provided or invalid","code":"IR_01"}}
status=401

$ curl -s -w '\nstatus=%{http_code}\n' -H 'X-Internal-Api-Key: alerts_dev_key' http://127.0.0.1:8085/alerts/ping
{"message":"alerts is reachable"}
status=200
```

`cargo clippy -p alerts --all-targets` is clean under the workspace lints.

No automated tests yet — there is no logic to test beyond the guard, and the guard is exercised
above end-to-end. Worth adding an actix integration test for the 401/200 pair when the first real
route lands.

## Checklist

- [x] `cargo clippy --all-features --all-targets` (clean for `alerts`)
- [ ] Added tests
- [ ] Added documentation to `docs/` — `crates/alerts/README.md` covers the crate

## Reviewer notes

Three things where I diverged from the original plan and would value a second opinion:

1. **`Scope` factories in the crate, not routes in `app.rs`.** `crates/subscriptions` is the repo's
   existing satellite-lib precedent and there the *router* declares the scope. Ours does not,
   because these routes need no `SessionState`, merchant auth, or DB — and one scope definition
   means standalone and embedded cannot drift. This is the most likely thing to draw pushback.
2. **Plain `!=` for the key comparison**, matching `AdminApiAuth`. The workspace contains no
   constant-time comparison crate at all; adding the first one to guard this felt like
   inconsistency in the direction of theatre. Happy to add `subtle` if you disagree.
3. **`GET /alerts/ping` is a temporary placeholder** and says so in the source. It deliberately does
   not squat on `/notify` — a stub returning `200` for anything is the sort of thing that gets
   deployed and quietly believed.


crates/alerts/src/errors.rs
```diff
@@ -0,0 +1,94 @@
...
+/// below, so a handler never has to think about HTTP.
+#[derive(Debug, Error)]
+pub enum AlertsError {
+    /// The internal API key was missing, malformed, or did not match.
+    #[error("Authentication failed")]
+    Unauthorized,
+
+    /// Something failed that the client can do nothing about.
+    #[error("Internal server error")]
+    InternalServerError,
```
ThisIsMani, line 67: Move this to the top.

crates/alerts/src/errors.rs
```diff
@@ -0,0 +1,94 @@
...
+            Self::Unauthorized => ApiErrorResponse::Unauthorized(ApiError::new(
+                "IR",
+                1,
+                "API key not provided or invalid",
+            )),
+            Self::InternalServerError => ApiErrorResponse::InternalServerError(ApiError::new(
+                "HE",
+                0,
+                "Something went wrong",
+            )),
```
ThisIsMani: Change order here as well.

crates/alerts/src/errors.rs
```diff
@@ -0,0 +1,94 @@
...
+#[derive(Debug, Error)]
+pub enum ConfigurationError {
+    /// A configuration value was present but unusable.
+    #[error("Error in parsing config: {0}")]
+    ConfigParsingError(String),
+
+    /// The configuration file could not be read or deserialized.
+    #[error("Application configuration error: {0}")]
+    ConfigurationError(config::ConfigError),
+
```
ThisIsMani, line 37: Are we using this variant anywhere? If not can we remove this?

crates/alerts/src/errors.rs
```diff
@@ -0,0 +1,94 @@
...
+    /// Binding the listener failed, or another I/O error occurred during startup.
+    #[error("I/O: {0}")]
+    IoError(std::io::Error),
```
ThisIsMani, line 40: Same with this, can we remove this if not being used.

crates/alerts/src/errors/actix.rs
```diff
@@ -0,0 +1,25 @@
+//! The one place the wire error becomes an HTTP response.
+//!
+//! Kept in its own module, mirroring `api_models::errors::actix`, so that the error *shape*
+//! ([`super::types`]) stays independent of the framework rendering it.
+
+use super::types::ApiErrorResponse;
+
+impl actix_web::ResponseError for ApiErrorResponse {
+    fn status_code(&self) -> reqwest::StatusCode {
+        use reqwest::StatusCode;
```
ThisIsMani: Move import to the top.

crates/alerts/src/errors/types.rs
```diff
@@ -0,0 +1,120 @@
...
+
+use serde::Serialize;
+
+/// The category of an error, as reported to the client.
+#[derive(Debug, Serialize)]
+pub enum ErrorType {
+    /// The request itself was at fault.
+    InvalidRequestError,
+    /// The service was at fault.
+    AlertsError,
```
ThisIsMani, line 19: When would this be used?

crates/alerts/src/health_check.rs
```diff
@@ -0,0 +1,32 @@
...
+
+/// The unauthenticated health scope.
+pub struct Health;
+
+impl Health {
+    /// Build the health scope.
+    pub fn server() -> Scope {
+        web::scope("health").service(web::resource("").route(web::get().to(health)))
+    }
+}
```
ThisIsMani: Can we put this in a file call app.rs or something so that I can see all the routes in the application in a single place.

crates/alerts/src/errors/types.rs
```diff
@@ -0,0 +1,120 @@
...
+    pub message: String,
+    /// The stable code, e.g. `IR_01`.
+    pub code: String,
+}
+
+impl From<&ApiErrorResponse> for ErrorResponse {
+    fn from(value: &ApiErrorResponse) -> Self {
+        let error_info = value.get_internal_error();
+        Self {
+            code: format!("{}_{:02}", error_info.sub_code, error_info.error_identifier),
```
ThisIsMani, line 103: Do we need :02?

crates/alerts/src/health_check.rs
ThisIsMani, line 1: Can we put this in routes/health_check.rs. Let's follow router file format. We might need core anyways.

crates/alerts/Cargo.toml
```diff
@@ -0,0 +1,50 @@
...
+# features exist only to tell shared dependencies which flavour to compile: the repo's secrets
+# management layer lives behind `hyperswitch_interfaces`, which reaches `api_models` and
+# `diesel_models`, and those do not build unless one of the two is selected.
+#
+# Defaulted to `v1` — matching the router — so that `cargo run -p alerts` works and nobody has to
+# think about a choice this crate does not care about. A workspace v2 build
+# (`--no-default-features --features v2`) still selects the other flavour.
+default = ["v1"]
+v1 = ["common_utils/v1", "external_services/v1", "hyperswitch_interfaces/v1"]
+v2 = ["common_utils/v2", "external_services/v2", "hyperswitch_interfaces/v2"]
```
ThisIsMani, line 21: Do we need this at all? And do we need to separate out v1 and v2 separately?

crates/alerts/src/secrets_transformers.rs
```diff
@@ -0,0 +1,55 @@
...
+        secret_management_client: &dyn SecretManagementInterface,
+    ) -> CustomResult<SecretStateContainer<Self, RawSecret>, SecretsManagementError> {
+        let secured_auth_config = value.get_inner();
+        let raw_internal_api_key = secret_management_client
+            .get_secret(secured_auth_config.internal_api_key.clone())
+            .await?;
+
+        Ok(value.transition_state(|_auth| Self {
+            internal_api_key: raw_internal_api_key,
+        }))
```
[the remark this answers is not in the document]
ThisIsMani: Confirmed and fixed in `9f544f1307` — this was a real auth bypass, thank you.

`validate()` is `impl Settings<SecuredSecret>`, so it checked the *handle*, not the key behind it. And it's worse than an empty key merely being accepted: `headers.get()` returns `Some("")` for a present-but-empty header, so `"" != ""` is false and **every** request sending a bare `X-Internal-Api-Key:` would have authenticated.

Now re-validated inside `fetch_raw_secrets` immediately after decryption, so the check runs on both sides of the `SecuredSecret` → `RawSecret` transition. Verified against the running service: an empty key header returns 401.

crates/alerts/src/lib.rs
```diff
@@ -0,0 +1,64 @@
...
+    let server = state.conf.server.clone();
+
+    let web_server = actix_web::HttpServer::new(move || {
+        actix_web::App::new()
+            .service(health_check::Health::server())
+            .service(routes::Alerts::server(state.clone()))
+            .wrap(router_env::RequestIdentifier::new(REQUEST_ID_HEADER))
+            .wrap(router_env::tracing_actix_web::TracingLogger::<
+                router_env::CustomRootSpanBuilder,
+            >::new())
```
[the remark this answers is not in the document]
ThisIsMani: Confirmed and fixed in `9f544f1307`.

actix runs the *last* registered `.wrap` first, so registering `RequestIdentifier` first made `TracingLogger` the outermost layer — `CustomRootSpanBuilder::on_request_start` reached `.get::<RequestId>()` before the identifier had inserted it and fell through to `unwrap_or_default()`, i.e. an empty string. Swapped to match the router at `router/src/lib.rs:574`.

Verified from the running service: the root span's `request_id` now matches the id in the rejection log line, where previously it would have been empty.

(Unrelated observation while checking: the root span carries *two* `request_id` fields with different values. That's inherent to `CustomRootSpanBuilder` — `root_span!` already emits its own — so the router has it too. The one sourced from extensions is the one that correlates.)

crates/alerts/src/errors.rs
ThisIsMani: Done in `9f544f1307` — `InternalServerError` is now first in `AlertsError`.

crates/alerts/src/errors.rs
ThisIsMani: Done in `9f544f1307`. Reordered the `switch()` match to match, and applied the same order to `ApiErrorResponse` and its `get_internal_error` / `error_type` matches in `errors/types.rs`, so all four orderings agree rather than just these two.

crates/alerts/src/errors.rs
ThisIsMani: It is used — kept it. `settings.rs` relies on the `From<config::ConfigError>` impl three times:

- `router_env::Config::builder(&environment.to_string())?` (line 148)
- `.build()?` (line 155)
- `errors::ConfigurationError::from(error.into_inner())` (line 162)

The first two convert via `?` so the call site never names the variant, which is why it reads as unused.

crates/alerts/src/errors.rs
ThisIsMani: Also used, same pattern — `main.rs:48`:

```rust
server.await.map_err(errors::ConfigurationError::from)?;
```

`Server::await` yields `io::Result<()>`, and that's the only thing turning it into our error type. Invisible at the call site because it goes through `From`.

crates/alerts/src/errors/actix.rs
ThisIsMani: Done in `9f544f1307` — both `use reqwest::StatusCode;` and `use actix_web::http::header;` are at module top now, and the return type is written as `StatusCode` rather than `reqwest::StatusCode`.

For context on why they were function-local: `api_models/src/errors/actix.rs` does exactly this, and I mirrored it file-for-file. Happy to have diverged — it's the odd one out in the repo.

crates/alerts/src/errors/types.rs
ThisIsMani: Both variants are used, in `error_type()` (`types.rs:81-82`) — `InvalidRequestError` for `Unauthorized`, `AlertsError` for `InternalServerError`. They become the `"type"` field in the response body.

That said, your instinct is right that it's ceremony: it's a two-variant enum whose only job is mapping to two string literals, and it exists because `api_models::errors::types::ErrorType` exists. Inlining the strings into `error_type()` and deleting the enum would lose nothing today. Say the word and I'll drop it — I left it for symmetry with `api_models`, not because it earns its place.

crates/alerts/src/health_check.rs
ThisIsMani: Done in `9f544f1307` — both scope factories now live in `routes/app.rs`, so the whole route tree is one file:

```rust
impl Alerts { pub fn server(state: AppState) -> Scope { web::scope("/alerts")... } }
impl Health { pub fn server() -> Scope { web::scope("health")... } }
```

Handlers moved out to `routes/alerts.rs` and `routes/health_check.rs`, so `app.rs` carries only the shape of the tree. "What does this service expose, and what guards it?" is now one file rather than a search.

crates/alerts/src/health_check.rs
ThisIsMani: Done in `9f544f1307`, and I took the `core` hint too:

```
routes.rs
routes/app.rs            — the whole route tree
routes/health_check.rs   — health handler
routes/alerts.rs         — ping handler
core.rs
core/notifier.rs         — was src/notifier.rs
```

`core` decides, `routes` exposes — the router's split. The notifier is domain logic rather than HTTP, so `core/notifier.rs` is a better home than the top level, and it means hyperswitch-cloud#23116 fills a seam that's already in the right place.

crates/alerts/src/errors/types.rs
ThisIsMani: It mirrors `api_models`, which formats codes the same way (`format!("{}_{:02}", ...)` → `IR_01`, `HE_00`). The padding keeps codes fixed-width so they sort and grep predictably once there are more than nine in a category.

With two errors it buys nothing concrete yet, so it's consistency-with-`api_models` rather than necessity. Happy to drop it to `{}_{}` if you'd rather not carry the convention — just say which.

crates/alerts/Cargo.toml
ThisIsMani: We need it, and they do need to be separate. I tested both halves rather than reasoning about it.

**Do we need it at all — yes.** Without a flavour selected, the build fails well before reaching our code: `api_models` and `diesel_models` don't compile. `alerts` uses no version-flavoured type itself, but the secrets-management layer reaches them transitively (`external_services` / `hyperswitch_interfaces` → `hyperswitch_domain_models` → `api_models` + `diesel_models`). The features forward and do nothing else.

**Do they need to be separate — yes, they're mutually exclusive.** `cargo check -p api_models --features v1,v2`:

```
error[E0428]: the name `Payment` is defined multiple times
  --> crates/common_utils/src/events.rs:22:5
error[E0428]: the name `Refund` is defined multiple times
error[E0428]: the name `PaymentMethod` is defined multiple times
error[E0428]: the name `Customer` is defined multiple times
```

So hardcoding `hyperswitch_interfaces = { features = ["v1"] }` with no crate-level feature would break any workspace v2 build, because Cargo would unify our forced `v1` with the workspace's `v2` and enable both. Two features with `default = ["v1"]` is what lets a v2 build turn ours off via `--no-default-features --features v2`.

`drainer` carries the same pair for the same reason.

crates/alerts/src/errors/types.rs
ThisIsMani: I wanted to understand where would we use AlertsError specifically. I understand why InvalidRequestError would come, but the other, I'm not clear.

crates/alerts/src/errors/types.rs
ThisIsMani: We assume then only 99 errors could exist at max. It's fine, don't change, it's a fair assumption imo.

crates/alerts/src/lib.rs
```diff
@@ -0,0 +1,71 @@
...
+use error_stack::ResultExt;
+use hyperswitch_interfaces::secrets_interface::secret_state::RawSecret;
+
+use crate::state::AppState;
+
+/// The configuration, after secrets have been resolved.
+pub type Settings = settings::Settings<RawSecret>;
+
+/// The header carrying the request id, threaded through every log line for a request.
+const REQUEST_ID_HEADER: &str = "x-request-id";
```
ThisIsMani: Our convention seems to be weird, internal auth has capital letters.

crates/alerts/src/errors/types.rs
ThisIsMani: Straight answer: **nowhere, today.** I checked, and `AlertsError::InternalServerError` is never constructed — it appears only in its own definition and in the two matches that map it onward. So `ErrorType::AlertsError` is reachable in theory and dead in practice. Your instinct was right and my earlier "both variants are used" was too generous; they're *referenced*, not *reached*.

The distinction it's meant to draw is fault: `InvalidRequestError` means the caller did something wrong and retrying unchanged won't help, `AlertsError` means we did and retrying might. Nothing in a scaffold can fail that way — the only error path is the guard, and that's the caller's fault.

What will construct it is the notifier doing real work: a Xyne POST returning non-2xx or `{"ok": false}` (hyperswitch-cloud#23128), SES rejecting a send (hyperswitch-cloud#23111). Those are squarely "we failed, not you".

So it's a placeholder, and the honest options are:

1. **Keep it** — six lines, and the next two tickets both need it.
2. **Delete `InternalServerError` and `ErrorType` entirely**, leaving `AlertsError` with one variant, and let #23128 add them back with a real failure in hand.

I'd keep it, but weakly — (2) is the more honest YAGNI position and I'll do it if you prefer. Worth noting (2) also makes the three-layer error split look thin, since a one-variant enum mapping to a one-variant enum is hard to justify on its own.

crates/alerts/src/errors/types.rs
ThisIsMani: Leaving it as-is. One small correction though, in case it changes your read: `{:02}` is a *minimum* width, not a cap — it pads to two digits but doesn't truncate, so error 100 formats as `HE_100`, not `HE_00`. Nothing breaks past 99; the codes just stop being fixed-width.

So the assumption is even softer than you're granting it.

crates/alerts/src/lib.rs
ThisIsMani: It is inconsistent, but the inconsistency is the repo's rather than mine — I matched each header to its own existing precedent:

- **Auth headers are Title-Case.** `X-Internal-Api-Key` is spelled that way in `router/src/lib.rs:145` and again in `subscriptions/src/helpers.rs:6`.
- **The request id header is lowercase.** `common_utils::consts::X_REQUEST_ID` is `"x-request-id"`, `config/development.toml:129` sets `header_name = "x-request-id"`, and `router_env`'s own docs and tests use that spelling throughout.

Functionally it makes no difference — HTTP header names are case-insensitive (RFC 7230 §3.2), actix's `HeaderMap` lookup is case-insensitive, and HTTP/2 lowercases everything on the wire anyway. So this is purely how it reads in source.

Your comment did surface a real improvement, which I've made in `1d0ea12fad`: I was carrying my own `"x-request-id"` literal instead of using `common_utils::consts::X_REQUEST_ID`. Now it uses the shared constant — one less duplicated string, and it stays aligned with the router's trace header if that value ever changes, so an id set by an upstream hop is the one we log rather than a second id for the same request.

I stopped short of two things, both easy to add if you want them:

- Declaring our own `X_INTERNAL_API_KEY` const in `common_utils::consts` next to `X_REQUEST_ID`, instead of a third copy of the literal — that would fix the actual duplication rather than the casing.
- Making the request id header configurable as the router does via `trace_header`. Felt like scope creep for a scaffold.

PR #13968, opened 2026-09-01
feat(external_services): add the chat service with Xyne and Slack clients

ThisIsMani:
## What

Adds `external_services::chat_service` — delivery of chat messages to a destination channel — with
two backends, `XyneClient` and `SlackClient`.

[Xyne](https://xynehq.com) is an open-source AI search and answer engine that also exposes a
**Slack-compatible messaging API** (`{base}/slack/chat.postMessage`, the `{ok, error, ts}`
envelope, `thread_ts` threading), so the two backends share their wire work through a private
`slack_compatible` module and differ only in the path namespace, the credential and a couple of
limits.

This is the client only. Nothing is wired to it yet.

## Why it looks like this

**Named `chat_service`, not `chat`.** `chat` already belongs to the AI assistant —
`router::core::chat`, `api_models::chat`, `conf.chat`. A second `chat` would collide the moment
anything mounts this in the router.

**A client is bound to one destination** — base URL, credential, channel — so a caller holding
several destinations builds one client each. Clients own no connection state: `http_client` caches
the `reqwest::Client` and its pool globally (`DEFAULT_CLIENT` / `PROXY_CLIENT_CACHE`), so building
one per message is a couple of `String`s. No config store and no lookup live here; deciding what to
say, and where, belongs to the callers.

**`ChatClient` is object-safe on purpose.** `EmailClient` next door has an associated type
(`RichText`), which is why it cannot be used as a trait object and why the erased `EmailService`
had to be invented alongside it. A caller resolving destinations at runtime needs
`Arc<dyn ChatClient>`, so this trait takes no associated type.

**`ok: false` cannot be read as a success.** This API reports failure twice over: once as a
non-2xx status, and once as **HTTP 200** carrying `{"ok": false, "error": "..."}`. A client that
only checks the status code silently reports the second as a success. The body parses into a
private `PostMessageResponse` that mirrors the wire, and the only way out of the module is
`TryFrom<PostMessageResponse> for MessageId`. `ok` is a required field, so a body carrying no
success marker fails to deserialize rather than reading as a success.

**The provider's error code is a type.** `SlackErrorCode` derives `Deserialize`, so the wire
spellings live in one place serde checks rather than in a hand-written `match`, and adding a code
means adding a variant. A code the provider introduces later still arrives verbatim through an
untagged `Unrecognised(String)`. It maps into a **backend-neutral** `ChatErrorReason`
(`ChannelNotFound`, `NotInChannel`, `InvalidAuth`, `TokenRevoked`, `MessageTooLong`, `RateLimited`,
`Other`) so callers stay backend-agnostic and a future backend maps into the same vocabulary
instead of inventing a second enum.

**`MessageId` is an enum, not a `String`.** Slack-compatible backends identify a message by a `ts`
string; others do not. Handing a foreign id back as a reply target is a typed
`IncompatibleReplyTarget` error rather than a malformed field on the wire. `#[non_exhaustive]`, so
adding a variant is additive.

**`ChatMessage` keeps its fields private**, behind a builder and accessors. It is the type most
likely to grow, and two growth directions are documented on it: a backend that is not
Slack-compatible (markup is not portable — Slack reads `*bold*` where Discord reads `**bold**`),
and files (a different endpoint with a different result, so a sibling method on `ChatClient` rather
than a field here).

**Base URLs are `Url`**, so a malformed one fails at deserialization rather than at construction.
`Url::join` is deliberately avoided — it resolves relatively, so a base of `/api/apps` with no
trailing slash would silently yield `/api/chat.postMessage`.

**Secrets.** The token is `Secret<String>`, the `Authorization` header is `Maskable`, and there is
a test asserting the token never reaches a `Debug` rendering.

## Two things Xyne does not do the way Slack does

Both were found by reading [`juspay/xyne-spaces`](https://github.com/juspay/xyne-spaces), the
public service behind this endpoint (`apps/backend/src/apps/platform-adapters/slack/`), and both
were wrong here before that:

- **`mrkdwn` has to be sent.** Slack enables markup by default. The Xyne adapter takes the markup
  path only on `mrkdwn === true`, so omitting the flag delivers `*bold*` and backticks as literal
  characters. It is now sent explicitly — a no-op against Slack, and the only spelling that behaves
  the same on both.
- **The message cap is per backend.** Xyne enforces **10,000** characters against Slack's 40,000, so
  the two cannot share a default: a message between the two would be truncated by nobody and then
  refused with `msg_too_long`. Note Xyne measures the message *after* rendering markup, and
  rendering expands it, so a body near the ceiling can still be refused — lower it in config rather
  than raising the default.

Truncation is otherwise a **character** cap, because that is a wire-format limit. Capping the
*number of items* rendered into a message belongs to whatever composes the message — this module
never sees items.

## Testing

28 unit tests, no credentials required. The client tests run against a real HTTP server on a
loopback port (`wiremock`, already in `Cargo.lock` via `router`), so headers, JSON body, status and
envelope all go over the wire — only the provider is fake. Both failure modes are covered, along
with each distinguished error code, `429` + `Retry-After`, threading, nested-`ts` fallback,
truncation (including a cap shorter than the truncation marker), a refusal naming no code, and a
body with no `ok` marker.

```
cargo test -p external_services --features v1 chat_service
```

Clippy is clean under the workspace lints; `cargo +nightly fmt` clean.

## Known behaviour worth reviewing

**`post_message` is not idempotent and the transport may retry.** `http_client::send_request`
clones a JSON request and resends it once on `ConnectionClosedIncompleteMessage`, and
`chat.postMessage` has no idempotency key, so a message can be delivered twice. Documented rather
than worked around: for alert delivery a duplicate costs far less than a drop, and every connector
call in this workspace already carries the same retry. Bypassing the shared transport would also
cost its proxy handling and metrics.

## Not in this PR

- **Any wiring.** No `AppState` field, no route, no toml key, no config schema.
- **File upload.** The adapter supports both the legacy `files.upload` and the current
  `files.getUploadURLExternal` / `files.completeUploadExternal` flow, so either is available when
  it is wanted — it is simply out of scope here.
- **Dry-run and an enabled/disabled switch.** Deferred rather than guessed at: with destinations
  supplied per client, "chat is off" is "the caller has no destination", and a real off-switch
  belongs with the wiring.
- **Message formatting.** No renderer, no escaping helpers, no templating. Whatever composes
  message text owns that — including escaping `&`, `<` and `>`, which markup would otherwise eat.

## Context

The wire contract was established from a working implementation and then verified against
[`juspay/xyne-spaces`](https://github.com/juspay/xyne-spaces). Xyne's published API reference
covers search and ingestion rather than this messaging surface, so the adapter source is the
authority for it.

Refs juspay/hyperswitch-cloud#23108, juspay/hyperswitch-cloud#23107

https://claude.ai/code/session_01JesWVUgg7SX72gvxjP1Mfi


crates/external_services/src/chat_service.rs
```diff
@@ -0,0 +1,203 @@
...
+
+    /// The message body.
+    pub fn text(&self) -> &str {
+        &self.text
+    }
+
+    /// The message this one replies to, if any.
+    pub fn reply_target(&self) -> Option<&MessageId> {
+        self.reply_to.as_ref()
+    }
```
ThisIsMani, line 122: Are we using these anywhere? If not let's remove them.

crates/external_services/src/chat_service.rs
```diff
@@ -0,0 +1,203 @@
...
+    /// The message was delivered but the provider named no id for it, so replies cannot be
+    /// threaded under it. Retrying would post a duplicate.
+    #[error("Chat provider accepted the message without returning a message id")]
+    MissingMessageId,
+
+    /// [`ChatMessage::reply_to`] carried an id this backend cannot thread against — typically an
+    /// id minted by a different backend.
+    #[error("The message id supplied cannot thread a reply on this chat provider")]
+    IncompatibleReplyTarget,
+}
```
ThisIsMani, line 165: Keep only the ones that are being used. Remove all the unused ones.

crates/external_services/src/chat_service.rs
```diff
@@ -0,0 +1,203 @@
...
+
+    /// The caller is posting too fast.
+    RateLimited {
+        /// How long the provider asked us to wait, when it said.
+        retry_after_seconds: Option<u64>,
+    },
+
+    /// Anything else, carrying the provider's own code verbatim.
+    Other(String),
+}
```
ThisIsMani, line 205: Same here.

crates/external_services/src/chat_service/slack.rs
```diff
@@ -0,0 +1,144 @@
...
+fn default_max_message_chars() -> usize {
+    DEFAULT_MAX_MESSAGE_CHARS
+}
+
+/// One Slack destination: a workspace and a channel within it.
+#[derive(Debug, Clone, Deserialize)]
+pub struct SlackConfig {
+    /// Root of the Slack API. Defaults to the public one.
+    #[serde(default = "default_base_url")]
+    pub base_url: String,
```
ThisIsMani: Can't it be a proper type instead of string?

crates/external_services/src/chat_service/slack_compatible.rs
```diff
@@ -0,0 +1,529 @@
...
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("Endpoint")
+            .field("base_url", &self.base_url)
+            .field("method_prefix", &self.method_prefix)
+            .field("channel", &self.channel)
+            .field("timeout_seconds", &self.timeout_seconds)
+            .field("max_message_chars", &self.max_message_chars)
+            .finish_non_exhaustive()
+    }
+}
```
ThisIsMani: Won't secret solve this already?

crates/external_services/src/chat_service/slack_compatible.rs
```diff
@@ -0,0 +1,529 @@
...
+        max_message_chars: usize,
+        proxy: Proxy,
+    ) -> ChatResult<Self> {
+        let base_url = base_url.trim().trim_end_matches('/').to_owned();
+        if base_url.is_empty() {
+            Err(ChatError::InvalidConfiguration(
+                "base url must not be empty",
+            ))?
+        }
+        url::Url::parse(&base_url).change_context(ChatError::InvalidConfiguration(
```
ThisIsMani: Should use Url in the type itself...

crates/external_services/src/chat_service/slack_compatible.rs
```diff
@@ -0,0 +1,529 @@
...
+        ok: OkTrue,
+        ts: Option<String>,
+        message: Option<NestedMessage>,
+    },
+    Failure {
+        #[allow(dead_code)]
+        ok: OkFalse,
+        error: Option<String>,
+    },
+}
```
ThisIsMani: I think this can be better written using ok as tag and True and False as variants. If we are having to write custom deserialiser, it's better to do it manually instead of forcing with serde.

ThisIsMani (commented):
Check if using any crate reduces our impl effort.

crates/external_services/src/chat_service/slack_compatible.rs
ThisIsMani: Right — removed, `Debug` is derived again. `Secret`'s `Debug` delegates to the masking strategy (`hyperswitch_masking/src/secret.rs:164-171`), so the manual impl was guarding nothing. Kept the test asserting the token never prints, since that is the property we actually care about.

crates/external_services/src/chat_service/slack_compatible.rs
ThisIsMani: Done — `base_url` is `url::Url` on both configs, so a malformed one fails at deserialization instead of at construction. `url` was already a dependency of this crate with the `serde` feature on.

One trap worth recording: `Url::join` is *not* used. It resolves relatively, so a base of `/api/apps` without a trailing slash would produce `/api/chat.postMessage` and silently drop a segment. There is a test pinning the built URL.

crates/external_services/src/chat_service/slack.rs
ThisIsMani: Same change — `SlackConfig::base_url` is `Url` now.

crates/external_services/src/chat_service/slack_compatible.rs
ThisIsMani: Tried the tag form first and it does not compile: serde's internal tagging resolves the tag as a *variant identifier*, so `{"ok": true}` fails with `invalid type: boolean `true`, expected variant identifier`. It only works when the tag is a string (`{"ok": "true"}`).

So I took your second suggestion instead, which is better than what I had: `PostMessageResponse` is faithful to the wire, and `TryFrom<PostMessageResponse> for MessageId` converts. `ok` stays a required field, so a body with no success marker still fails to deserialize rather than reading as a success — the property the untagged enum enforced structurally is now one branch in one private conversion, with real serde errors instead of "data did not match any variant". It also matches how connectors here already do it.

crates/external_services/src/chat_service/slack_compatible.rs
```diff
@@ -0,0 +1,529 @@
...
+/// is a wire-level limit and belongs here; capping the *number of items* rendered into a message
+/// is the formatter's business, since this module never sees items.
+fn truncate(text: &str, max_chars: usize) -> String {
+    if text.chars().count() <= max_chars {
+        return text.to_owned();
+    }
+
+    let keep = max_chars.saturating_sub(TRUNCATION_MARKER.chars().count());
+    let mut truncated: String = text.chars().take(keep).collect();
+    truncated.push_str(TRUNCATION_MARKER);
```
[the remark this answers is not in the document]
ThisIsMani: Fixed. The marker is now clamped to the cap as well, so the output can never exceed `max_message_chars`. Added a test that walks every cap from 1 up to the marker length.

crates/external_services/src/chat_service/slack_compatible.rs
```diff
@@ -0,0 +1,529 @@
...
+                ),
+                (
+                    http::header::CONTENT_TYPE.to_string(),
+                    "application/json".to_owned().into(),
+                ),
+            ])
+            .set_body(RequestContent::Json(Box::new(payload)))
+            .build();
+
+        let response = http_client::send_request(&self.proxy, request, Some(self.timeout_seconds))
```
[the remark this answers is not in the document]
ThisIsMani: Acknowledged and documented on `post_message` rather than worked around.

The retry is real: `send_request` clones a JSON request and resends once on `ConnectionClosedIncompleteMessage`, and `chat.postMessage` has no idempotency key. But the trade runs the other way for this caller — a duplicate alert costs far less than a dropped one, and every connector call in this workspace already carries the same retry. Bypassing the shared transport would also cost us its proxy handling and metrics.

crates/external_services/src/chat_service.rs
ThisIsMani: I went looking for genuinely dead code and found exactly one thing: **the whole `mrkdwn` module**, which was called nowhere but its own tests. Deleted.

Everything else here is constructed. Every `ChatError` and `ChatErrorReason` variant is produced by the client, and both `ChatMessage` accessors are used by `slack_compatible`. What is true is that nothing *matches* on them yet, because the Notifier that consumes them does not exist — and #23108 explicitly asks for `channel_not_found`, `not_in_channel`, `invalid_auth`, `token_revoked`, `msg_too_long` and `rate_limited` to be distinguished, so cutting to what a caller reads today would delete the taxonomy the ticket asked for. Say the word if you would rather they go and come back with the consumer.

On your extensibility worry, which I think is the more important half — the accessors *are* the answer. Private fields plus a builder mean the type can grow without breaking callers, whereas public fields would freeze the shape. I documented both directions on `ChatMessage`:

- **A backend that is not Slack-compatible.** `text` is markup and markup is not portable — Slack reads `*bold*`, Discord reads `**bold**`. Today every backend is Slack-compatible so a rendered string is honest. The first one that is not forces a choice between rendering at the call site and carrying structured content, and keeping `text` private is what keeps that choice open.
- **Files.** They should not be a field here. Uploading is a different endpoint with a different result — a file id, not a message id — so it earns a sibling method on `ChatClient`. Worth flagging that Slack retired `files.upload` on 12 Nov 2025 and now requires a three-call external flow, while Xyne still serves the old single-shot method, so this is the point where Xyne and Slack stop being one client.

crates/external_services/src/chat_service.rs
ThisIsMani: See the reply on the `ChatMessage` thread — short version: `mrkdwn` was the only dead code and is gone; every `ChatError` variant is constructed, and the six the ticket names are required by #23108. Happy to cut them back if you would rather they arrive with their consumer.

crates/external_services/src/chat_service.rs
ThisIsMani: Same — every `ChatErrorReason` variant is produced by `reason_from_code`. These six are the exact set #23108 asks to distinguish.

crates/external_services/src/chat_service.rs
```diff
@@ -0,0 +1,215 @@
...
+            Self::RateLimited {
+                retry_after_seconds: Some(seconds),
+            } => write!(f, "rate limited, retry after {seconds}s"),
+            Self::RateLimited {
+                retry_after_seconds: None,
+            } => f.write_str("rate limited"),
+            Self::Other(code) => write!(f, "{code}"),
+        }
+    }
+}
```
ThisIsMani: Where are we using these?

crates/external_services/src/chat_service/slack_compatible.rs
```diff
@@ -0,0 +1,531 @@
...
+use super::{ChatError, ChatErrorReason, ChatMessage, ChatResult, MessageId};
+use crate::http_client;
+
+/// The method that posts a message. The only one this crate calls; `files.upload` is out of v1.
+const CHAT_POST_MESSAGE: &str = "chat.postMessage";
+
+/// Slack's documented limit on the `text` field of `chat.postMessage`.
+pub(super) const DEFAULT_MAX_MESSAGE_CHARS: usize = 40_000;
+
+/// Matches the R alerts service, which has run this in production against Xyne.
```
ThisIsMani: Please don't have any references to the R service.

crates/external_services/src/chat_service/slack_compatible.rs
```diff
@@ -0,0 +1,531 @@
...
+        "not_in_channel" | "is_archived" => ChatErrorReason::NotInChannel,
+        "invalid_auth" | "not_authed" => ChatErrorReason::InvalidAuth,
+        "token_revoked" | "account_inactive" => ChatErrorReason::TokenRevoked,
+        "msg_too_long" => ChatErrorReason::MessageTooLong,
+        "rate_limited" | "ratelimited" => ChatErrorReason::RateLimited {
+            retry_after_seconds: None,
+        },
+        other => ChatErrorReason::Other(other.to_owned()),
+    }
+}
```
ThisIsMani: Can we have an enum for this instead. Let's not do this please.

crates/external_services/src/chat_service.rs
ThisIsMani: By `ChatError::Rejected`, which interpolates it: `#[error("Chat provider rejected the message: {reason}")]`. Deleting the impl fails to compile, since thiserror needs `ChatErrorReason: Display` to generate that message.

Fair question to ask of a hand-written impl, though, so I derived it instead — `ChatErrorReason` now uses `thiserror` with an `#[error("…")]` on each variant. Same output, twenty fewer lines, and each message sits on the variant it describes rather than in a separate impl that can silently drift from it.

crates/external_services/src/chat_service/slack_compatible.rs
ThisIsMani: Removed. The comment now gives the reason for the value instead of pointing at another service.

Also swept the rest of the PR: that was the only such reference in code, and I have edited two more out of the PR description. Worth knowing that the two existing commit messages still name it — I have left them alone because rewriting them means a force-push that would mark these review threads outdated. Say the word if you would rather I do it and I will.

crates/external_services/src/chat_service/slack_compatible.rs
ThisIsMani: Done — `SlackErrorCode` is a `Deserialize` enum now, and `From<SlackErrorCode> for ChatErrorReason` maps it, so `PostMessageResponse.error` is typed rather than a `String` matched by hand. The wire spellings live in one derive that serde checks, and adding a code means adding a variant instead of remembering to extend a `match`.

The one thing I did not want to lose is the raw code for an error the provider adds later. `#[serde(untagged)] Unrecognised(String)` keeps it verbatim — I checked it against serde 1.0.219 before relying on it:

```
channel_not_found -> Ok(ChannelNotFound)
     ratelimited  -> Ok(RateLimitedCompact)
  brand_new_code  -> Ok(Unrecognised("brand_new_code"))
```

The test now deserializes each code and maps it, rather than calling a mapping function directly, so the spellings are covered too — including `is_archived`, `not_authed` and `account_inactive`, which were previously mapped but never asserted.

crates/external_services/src/chat_service/slack_compatible.rs
ThisIsMani: Correction to my earlier reply on this thread — I had the description wrong, not just the reference.

Xyne is public and open source ([xynehq.com](https://xynehq.com), Apache-2.0, `github.com/xynehq/xyne`), and it is an AI search and answer engine rather than a chat service. Both my doc comments called it "an internally hosted chat service", which is wrong twice over. Fixed: it is now described as what it is, with a link, and the module says we use its Slack-compatible messaging API.

One thing worth knowing: its published API reference covers search and ingestion only — the `/api/apps/slack/` messaging surface is not in it. So the docs now state where the contract came from rather than implying it is documented upstream.

crates/external_services/src/chat_service.rs
ThisIsMani: Correcting something I said on this thread earlier, now that I have read [`juspay/xyne-spaces`](https://github.com/juspay/xyne-spaces) — the public service behind this endpoint.

I claimed files were the point where Xyne and Slack stop being one client, because Slack retired `files.upload`. **That was wrong.** The adapter implements both flows:

```
POST /files.upload                    ← legacy multipart
POST /files.getUploadURLExternal      ← Slack current flow
POST /files.completeUploadExternal
POST /_upload/:fileId                 ← raw binary, 50 MB
```

So they stay one client through uploads, and whichever flow we prefer is available. The design note on `ChatMessage` still holds — uploading returns a file id rather than a message id, so it earns a sibling method rather than a field — but not for the reason I gave.

Reading it also turned up two real bugs, fixed in `afe112c`: `mrkdwn` must be sent explicitly (their adapter takes the markup path only on `mrkdwn === true`, so every formatted message we sent would have rendered literally), and the message cap is per backend — Xyne enforces 10,000 characters against Slack's 40,000. PR description updated to match.

PR #13984, opened 2026-09-02
fix(email): route SMTP sends through the configured proxy

itsharshvb:
## Type of Change

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

The SMTP email client received a `proxy_url` parameter on every send but
discarded it (`_proxy_url`, never read) — SMTP traffic always connected
directly to the mail host regardless of proxy config. SES already honors
proxying correctly; SMTP silently didn't.

`lettre` (the SMTP crate used here) has no proxy support anywhere in its
API — this drops to lettre's lower-level async connection API and builds
proxying underneath it via `tokio-socks`:

- SMTP gets its own, self-contained proxy config — `[email.smtp.socks5]`
  (`host`, `port`, optional `username`/`password`) — independent of the
  generic `proxy.https_url`/`bypass_proxy_hosts` used by connector calls
  and SES. SOCKS5 rather than HTTP CONNECT: it's a binary protocol (no
  hand-rolled HTTP parsing), and critically it lets the **proxy** resolve
  the target SMTP hostname rather than the app — the property that matters
  for restricted-egress deployments (e.g. k8s clusters with no direct
  DNS/network path to the mail relay).
- `SmtpServer::create_client` dials either directly or through the SOCKS5
  proxy depending on whether `smtp_config.socks5` is configured. The proxy
  address is resolved via tokio's non-blocking resolver first
  (`tokio::net::lookup_host`) rather than handed to `tokio-socks` as a
  `(&str, u16)` tuple — that path resolves via blocking
  `std::net::ToSocketAddrs`, which would stall a runtime thread and escape
  `smtp_config.timeout` entirely.
- `SmtpServer::send_email` manually drives AUTH/SEND/QUIT over that
  connection, every network step wrapped in a timeout. QUIT
  failure/timeout after a successful SEND is best-effort (logged, not
  fatal) — the message is already accepted by the server at that point, so
  failing the whole call would risk a caller retrying and double-sending.
- `SmtpServerConfig::validate()` validates `socks5` when present (empty
  host, zero port, partial username/password) via its own
  `Socks5Config::validate()`, and fails fast at process startup rather
  than surfacing as a confusing failure on first send.
- New `Cargo.toml` dep: `tokio-socks = "0.5"`. 
- No automated tests (matching this file's existing no-test-coverage
  state); verified manually end-to-end against real services instead.

Scope: `crates/external_services/src/email/smtp.rs`,
`crates/external_services/Cargo.toml`, `Cargo.lock`, one call-site
simplification in `crates/router/src/routes/app.rs` (`SmtpServer::create`
no longer threads `proxy.bypass_proxy_hosts` through), and documentation
of the new optional `[email.smtp.socks5]` block in
`config/docker_compose.toml` and `config/deployments/env_specific.toml`.
No `EmailClient`/`EmailService` trait changes.

### Additional Changes

## Motivation and Context

Deployments running `active_email_client = "SMTP"` behind an egress proxy
had no way to route mail traffic through it — `proxy.https_url` was set but
silently ignored on this path, unlike every other outbound integration in
the codebase.

## How did you test it?

Manually, end-to-end, through the real running application — against real
services, not stubs.

**Setup:** `mailhog` (real SMTP server) and a real SOCKS5 proxy
(`serjs/go-socks5-proxy`, auth required) in Docker on the same Docker
network. Built the actual `router` binary
(`cargo build --bin router --features v1,email`) against this branch,
config from `config/development.toml` with `[email.smtp.socks5]` pointed
at the proxy container.

**Verification:** triggered `POST /user/signup_with_merchant_id` (creates
an org/merchant/user and sends a real "set your password" email through
`send_email`) → `{"is_email_sent":true}`, confirmed delivered via
mailhog's inbox API. The SMTP host in config (`mailhog`) is a hostname
this process cannot resolve on its own — it only exists on the Docker
network the proxy sits on — so a successful send is only possible if the
SOCKS5 path is genuinely taken, not incidental.

### Screenshot:

- **SMTP and proxy enabled:**
Screenshot 2026-09-17 at 7 40 05 PM

- **Curl request:**
Screenshot 2026-09-04 at 3 07 01 PM

- **Application Logging:**
Screenshot 2026-09-04 at 3 09 13 PM

- **Proxy Logging:**
Screenshot 2026-09-04 at 3 08 17 PM

- **Mailhog recieves the req:**
Screenshot 2026-09-04 at 3 06 55 PM

## Checklist


jagan-jaya (dismissed):

SanchithHegde (changes requested):
I don't think it's a good idea to be constructing the HTTP CONNECT request ourselves, or have handwritten response parsing logic. Requesting changes so as to prevent this from being merged, and so that we discuss this internally.

PR #13985, opened 2026-09-02
feat(payments): expose card subtype, segment type and funding source

### LINKED ISSUE #14019: [FEATURE] Expose card subtype, segment type and funding source on payment responses
## Feature Description

The `card_info` table holds `card_subtype`, `card_segment_type` and `funding_source` against a card BIN, and `get_additional_payment_data` already fetches that row while building `AdditionalCardInfo` for a payment. Those three columns are dropped on the way into the response, so `payment_method_data.card` exposes `card_type` (`credit`/`debit`) and nothing more.

That leaves no way to tell from a payment response:

- a commercial, business or government card from a consumer one — `card_segment_type`
- a deferred debit or charge card from a plain credit — `funding_source`, which `card_type` collapses
- the product the BIN is issued under, e.g. `CLASSIC` vs `ELECTRON` — `card_subtype`

Today a caller that needs any of this has to make a second `GET /cards/{bin}` call per payment, for data the router already read while processing it. The router itself consumes `card_segment_type` off the same row for blocklist evaluation (`crates/router/src/core/blocklist/utils.rs`), so the value is live in the payment path — just not visible to the caller.

A concrete case: Worldpay reports its own `cardSubBrand` (`ECMC_CREDIT`, `ECMC_COMMERCIAL_CREDIT`, `VISA_ELECTRON`, …) and downstream fraud screening is keyed on those tokens. Reconstructing them from a Hyperswitch payment response is currently only possible down to the consumer credit/debit split, because the commercial segment never reaches the caller.

## Possible Implementation

Add the three fields to `AdditionalCardInfo` and to `CardResponse` (the wire type behind `payment_method_data.card`), and populate them at the `get_additional_payment_data` branches that already call `db.get_card_info(card_isin)` — no additional lookups, so no added latency. `None` wherever no BIN record is in scope: locker details, mandate updates, revenue-recovery tokens, billing-connector webhooks.

`funding_source` is already a typed `FundingSource` enum on the row, so it can be carried through as-is; `card_subtype` and `card_segment_type` are free-form text in the table and would stay `Option<String>`, matching the existing `CardInfoResponse`.

All three are additive and nullable, so existing consumers are unaffected.

Longsueynin:
## Type of Change

- [x] New feature

## Description

Carries `card_subtype`, `card_segment_type` and `funding_source` from the `card_info` row through `AdditionalCardInfo` to `CardResponse`, the wire type behind `payment_method_data.card`.

- Populated at the four `get_additional_payment_data` branches that already call `db.get_card_info(card_isin)`, so no extra lookups and no added latency.
- `None` wherever no BIN record is in scope: locker details, mandate updates, revenue-recovery tokens, billing-connector webhooks.
- `funding_source` is carried as the existing `FundingSource` enum; `card_subtype` and `card_segment_type` stay `Option<String>`, matching `CardInfoResponse`.
- `FundingSource` registered in the v2 OpenAPI schema list, which `CardResponse` now references.

All three fields are additive and nullable, so existing consumers are unaffected.

Every caller of `get_additional_payment_data` is `#[cfg(feature = "v1")]`, so v1 is populated and v2 carries the schema without data for now. The one v2 path that reads the same row, `populate_bin_details_for_payment_method`, takes only `card_type` into a transient local to derive `payment_method_subtype` and never into a response. Happy to extend the v2 flow here or in a follow-up if that is preferred.

### Additional Changes

Three new nullable response fields on `CardResponse`; no request-side change. Regenerated `api-reference/v1/openapi_spec_v1.json` and `api-reference/v2/openapi_spec_v2.json`.

## Motivation and Context

Closes #14019

## How did you test it?

- `cargo check --workspace --all-targets` for v1 and for `--no-default-features --features "release,v2,redis-rs"`
- `cargo fmt --check`
- OpenAPI specs regenerated for v1 and v2; verified no dangling `$ref` in either

No behavioural change to test beyond the added fields: the BIN row was already being fetched and the values are passed through unmodified.

## Checklist


crates/api_models/src/payments.rs
```diff
@@ -4164,6 +4164,18 @@ pub struct AdditionalCardInfo {
     /// Card type, can be either `credit` or `debit`
     pub card_type: Option<String>,
 
+    /// The product the card is issued under, e.g. `CLASSIC` or `ELECTRON`.
+    /// Free form, as the value comes straight from the BIN record.
+    pub card_subtype: Option<String>,
+
+    /// The segment the card is issued to, e.g. `Consumer`, `Commercial`, `Business`
+    /// or `Government`. Free form, as the value comes straight from the BIN record.
+    pub card_segment_type: Option<String>,
```
ayush22667: Can we use an enum here, `CardSegmentType`(already in codebase), rather than a string?

crates/router/src/core/payments/helpers.rs
```diff
@@ -6017,6 +6017,9 @@ pub async fn get_additional_payment_data(
                         card_issuer: card_data.card_issuer.to_owned(),
                         card_network,
                         card_type: card_data.card_type.to_owned(),
+                        card_subtype: None,
+                        card_segment_type: None,
+                        funding_source: None,
```
ayush22667: Since we skip the BIN lookup in this branch, could we add card_subtype, card_segment_type and funding_source to domain card and carry them through apply_additional_card_info, so they're already on card_data here instead of being dropped to None?

crates/router/src/core/payments/helpers.rs
```diff
@@ -6155,6 +6164,9 @@ pub async fn get_additional_payment_data(
...
```
ayush22667: same here

crates/router/src/core/payments/helpers.rs
```diff
@@ -6623,6 +6650,9 @@ pub async fn get_additional_payment_data(
                         card_issuer: card_with_limited_details.card_issuer.to_owned(),
                         card_network: card_with_limited_details.card_network.clone(),
                         card_type: card_with_limited_details.card_type.to_owned(),
+                        card_subtype: None,
+                        card_segment_type: None,
+                        funding_source: None,
```
ayush22667: same here

hrithikesh026 (approved):

ayush22667 (approved):

AkshayaFoiger (approved):

ThisIsMani (approved):

Sakilmostak (approved):

PR #13987, opened 2026-09-02
feat(Generic): [FRM] Implementing FRM framework flows

### LINKED ISSUE #14110: [FEATURE] FRM framework
### Feature Description

[FEATURE] FRM framework

### Possible Implementation

[FEATURE] FRM framework

None

THE-CORE-BUG:
## Type of Change

- [ ] Bugfix
- [ x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

Adds a path for FRM providers whose risk evaluation runs on the Unified Connector Service instead of in-process, and onboards **nSure.ai** as the first such provider.

  Existing FRM providers (Signifyd, Riskified, CyberSource Decision Manager) each ship an in-process connector implementing five flow traits. A UCS-backed provider has none: the connector-service owns the provider-specific
  transformation, and Hyperswitch only assembles the risk facts and interprets the verdict. This PR builds that bridge on the same gateway abstraction payments and payouts use, without disturbing the native FRM path.

  **Routing — no FRM-specific gate**

  A UCS-backed FRM provider is listed in the existing `ucs_only_connectors`, which makes it a `ConnectorIntegrationType::UcsConnector`. `call_frm_service` asks the same `should_call_unified_connector_service` every other
  flow does; for a UCS-only connector that returns `UnifiedConnectorService` unconditionally and the kill switch is skipped (there is no direct integration to divert to). Only the pre-authorization risk evaluation is routed
  (`frm_transaction_type == PreFrm && last_step == Processing` — `pre_payment_frm_core` calls `call_frm_service` twice for a Pre flow, and only the first is an evaluation; gating on `last_step` ensures the provider is asked
  exactly once per payment). Notify-style flows have no UCS equivalent and always resolve to Direct.

  **Gateway (`crates/router/src/core/fraud_check/gateway.rs`, new)**

  - `impl PaymentGateway for frm_api::Checkout` — the UCS gateway. Fetches an access token if needed, builds the request from `RouterData`, and runs `frm_pre_risk_check` under `ucs_logging_wrapper_granular`, so the call gets
  the same audit event, timing, kill-switch reporting and `tonic::Status` error extraction as every other UCS call.
  - `impl FlowGateway for frm_api::Checkout` — selects `DirectGateway` or the UCS gateway from the execution path.
  - `FeatureFrm::decide_frm_flows` now takes a `RouterGatewayContext`; `checkout_flow.rs` dispatches through `execute_payment_gateway`.

  **Transport (`crates/external_services`)**

  - `UcsClient` gains `frm_service_client` and a `frm_pre_risk_check` RPC. FRM calls carry `x-frm-connector`, selected by `ConnectorType::PaymentVas`.
  - `frm_pre_risk_check` decodes gRPC failures via `from_grpc_error`, matching the payment RPCs.
  - **Also applied to `surcharge_calculate`**, which previously collapsed every failure to `SurchargeCalculateFailure`. This changes kill-switch reason attribution and error/metric shape for the surcharge flow; the caller is
  best-effort so behaviour is unaffected. Called out here since it is outside the FRM scope.

  **Transformers and handlers**

  - `impl ForeignTryFrom<&RouterData<Checkout, …>> for FrmServicePreRiskCheckRequest` in `transformers.rs`, alongside the other request builders. Reuses the shared `PaymentAddress`, `build_ucs_order_details`, 
  `build_unified_connector_service_payment_method` and `ConnectorState` converters. `frm_metadata` is forwarded verbatim as `connector_feature_data`.
  - `handle_unified_connector_service_response_for_frm_pre_risk_check` in the parent module: `error` → `Err(ErrorResponse)` (same shape as the sibling handlers); a success without a decision is a 
  `ResponseDeserializationFailed`. `frm_status_from_ucs_decision` maps `Approve/Reject/Review` and errors on `Unspecified/Error`.
  - `FraudCheckCheckoutData` gains `customer_name` and `payment_method_data_full` (the domain `PaymentMethodData`; the existing `AdditionalPaymentData` field stays for the in-process connectors), plus `Serialize` for the 
  gateway bound — with a doc comment noting `Secret` masks only under `masked_serialize`.

  **FRM-specific orchestration (`crates/router/src/core/unified_connector_service/frm.rs`)**

  - `get_frm_access_token` — Redis cache keyed by MCA, falling back to the shared access-token gateway (`execute_payment_gateway` on a router data re-typed to `AccessTokenAuth`). The gateway reads the merchant-declared
  `connector_type` from the MCA to pick the `PaymentVas` header namespace.
  - `FrmNotification` / `call_unified_connector_service_for_frm_notification` — lifecycle-event sender over `notify_connector`. **Not wired in this PR**; kept deliberately so the dispute-webhook follow-up only adds the call
  site.

  **nSure**

  - `Nsure` added to `Connector`, `FrmConnectors`, `euclid`, `connector_configs`; not routable as a payment connector.
  - Stub connector in `hyperswitch_connectors` so nSure is a first-class `Connector` for the gateway, like other UCS-only connectors. Its FRM flows override `build_request` to return `Err(NotImplemented)` rather than the
  `Ok(None)` default — on the Direct path (UCS down, or nsure missing from `ucs_only_connectors`) this routes through the existing fail-open/fail-closed handling instead of silently passing the payment with `Pending`.
  - `ConnectorSpecificConfig::Nsure { api_key, app_id: Option, api_version }` sent via `x-connector-config`; built from `BodyKey` (key + Application ID) or `HeaderKey` (key alone), matching what `connector_validation`
  accepts. `app_id` is optional on the connector-service side.
  - `nsure.base_url` added to the `[connectors]` config.

## How did you test it?
 Exisiting Kount Creds support was added to hyperswitch and then forwarded to already present Prism to test the end-to-end flow


crates/router/src/core/fraud_check.rs
```diff
@@ -129,6 +129,108 @@ where
...
+    //
+    // `pre_payment_frm_core` invokes `call_frm_service` twice for a Pre flow —
+    // once for Checkout and once for Transaction — and both carry
+    // `FraudCheckType::PreFrm`. Only the first is a risk *evaluation*; the
+    // second is a notification whose result is discarded by the caller. Gating
+    // on `last_step == Processing` (set at row insert, advanced to
+    // `CheckoutOrSale` after the first call) ensures nSure is asked exactly once
+    // per payment, so the provider is not billed twice and does not see a
+    // repeated `uniqueRequestId`.
+    #[cfg(feature = "v1")]
```
hrithikesh026: this function itself is under v1 flag. no need for this.
```suggestion
```

crates/router/src/core/unified_connector_service/connector_config.rs
```diff
@@ -1904,6 +1904,17 @@ pub fn build_connector_config_header(
     auth_type: &ConnectorAuthType,
     merchant_account_metadata: Option<&serde_json::Value>,
 ) -> RouterResult<Option<String>> {
+    // nSure is an FRM connector executed by the connector-service. The pinned
+    // `unified-connector-service-client` has no `nsure` case in the
+    // `ConnectorSpecificConfig` oneof, so there is nothing to build here.
+    // Suppressing the header makes UCS take the legacy auth-header path
+    // (`x-auth` / `x-api-key` / `x-key1`), which maps BodyKey onto
+    // `ConnectorSpecificConfig::Nsure` connector-side. Same treatment as
```
hrithikesh026: this can be combined with the Netcetera branch

config/deployments/integration_test.toml
```diff
@@ -1052,6 +1052,7 @@ connector_list = "worldpayvantiv"
 [grpc_client.unified_connector_service]
 ucs_only_connectors = "imerchantsolutions, paytm, phonepe, hyperpg, revolv3, fiservcommercehub, absa_sanlam, interpayments, payconex, dlocal, barclaycard, tsys_transit, jpmorgan, datatrans, givepayments, tesouro, citigate, ilixium, moneris, elavon, worldpayraft"    # Comma-separated list of connectors that use UCS only
 ucs_psync_disabled_connectors = "cashtocode, trustly"    # Comma-separated list of connectors to disable UCS PSync call
+ucs_frm_connectors = "nsure"    # Comma-separated FRM connectors whose risk evaluation runs on UCS
```
gopikrishna000: i think payouts connectors dont have ucs_only_payout_connectors
but all of it usually written in ucs_only_connectors?

crates/external_services/src/grpc_client/unified_connector_service.rs
```diff
@@ -82,6 +84,12 @@ pub struct UnifiedConnectorServiceClientConfig {
     /// Set of connectors for which psync is disabled in unified connector service
     #[serde(default, deserialize_with = "deserialize_hashset")]
     pub ucs_psync_disabled_connectors: HashSet<Connector>,
+
+    /// Set of FRM connectors whose risk evaluation is executed by the unified
+    /// connector service rather than by an in-process connector. Empty by
+    /// default, so no FRM traffic is routed to UCS unless explicitly configured.
+    #[serde(default, deserialize_with = "deserialize_hashset")]
+    pub ucs_frm_connectors: HashSet<Connector>,
```
gopikrishna000: here type is same as ucs_only_connectors ryt, so can we set in ucs_only_connectors only?

crates/router/src/core/unified_connector_service/connector_config.rs
```diff
@@ -1929,20 +1926,27 @@ pub fn build_connector_config_header(
...
+    //   case in its `ConnectorSpecificConfig` oneof yet.
+    // - Netcetera: connector-service's `ConnectorSpecificConfig` oneof has no
+    //   `netcetera` case at all.
+    // Sending the header for either makes UCS hard-error on deserialization
+    // instead of falling back to the legacy auth header, so both are suppressed
+    // here before a config is even built. UCS then takes the legacy auth-header
+    // path (`x-auth` / `x-api-key` / `x-key1`), which maps BodyKey onto
+    // `ConnectorSpecificConfig::Nsure` connector-side, and needs no config at
+    // all for Netcetera.
+    if matches!(connector, Connector::Nsure | Connector::Netcetera) {
```
gopikrishna000: why no connectorspecificconfig for Nsure

docs/design/frm-architecture.md
```diff
@@ -0,0 +1,435 @@
...
+- a merchant can express *their* risk policy without us shipping code for it,
+- a new FRM provider can be integrated by implementing two traits, not five,
+- FRM works identically in v1 and v2,
+- FRM failure modes (timeout, provider down, ambiguous verdict) are a merchant choice, not an accident of which internal call errored.
+
+This document maps what exists, states the specific defects with file references, and proposes a layered target with a phased, back-compatible migration.
+
+## 2. What exists today
+
+FRM is already substantially built. Nothing here is greenfield.
```
gopikrishna000: is this file intended to be pushed
but it seems not generic for oss reader

gopikrishna000:
should_call_unified_connector_service_for_frm(
lets reuse exisiting generic one (should_call_unified_connector_service)? or is there any blocker in that?

gopikrishna000:
build_payment_address
can we have this as  tryfrom, or is it alraedy there, we can reuse it

gopikrishna000:
> build_payment_address can we have this as tryfrom, or is it alraedy there, we can reuse it

same for build_address,build_order_detail,handle_unified_connector_service_response_for_frm_pre_risk_check,

usually domain to proto transformation or vice versa are present in the unified_connector_service.rs/transformer file

same for handle_unified_connector_service_response_for_frm_pre_risk_check

gopikrishna000:
can we have function for this mapping

    let status = match (status_code, decision) {
        (200..=299, Some(FrmDecision::Approve)) => diesel_models::enums::FraudCheckStatus::Legit,
        (200..=299, Some(FrmDecision::Reject)) => diesel_models::enums::FraudCheckStatus::Fraud,
        (200..=299, Some(FrmDecision::Review)) => {
            diesel_models::enums::FraudCheckStatus::ManualReview
        }
        // `Error`/`Unspecified`/absent verdict, or any non-2xx status: the
        // provider gave us nothing usable. Hold for review rather than approve.
        (200..=299, _) => diesel_models::enums::FraudCheckStatus::ManualReview,
        (code, _) => {
            router_env::logger::warn!(
                status_code = code,
                "FRM pre risk check returned a non-success status; treating as manual review"
            );
            diesel_models::enums::FraudCheckStatus::ManualReview
        }
    };

gopikrishna000:
can we please use ucs_logging_wrapper_granular whenever ucs grpc call is made
same way other endpoints does

gopikrishna000:
how pub async fn get_frm_access_token(
 is different from normal access token call?

deepanshu-iiitu (dismissed):
Connector changes look good

config/deployments/integration_test.toml
THE-CORE-BUG: Right, payouts share ucs_only_connectors — a payout connector is the same integration as the payment connector, so it's the same question and the same code path (determine_connector_integration_type). FRM doesn't go through that path at all ucs_frm_connectors is a plain yes/no read only by the FRM gate, with no Direct/Shadow/rollout/kill-switch behaviour. Adding nsure to ucs_only_connectors would mark it a UCS-only payment connector and change webhook routing for it.

crates/external_services/src/grpc_client/unified_connector_service.rs
THE-CORE-BUG: Even though type is same, consumers are different. Also FRM is a different flow, so tried to keep the code convention for it to be separate

docs/design/frm-architecture.md
THE-CORE-BUG: removed

THE-CORE-BUG:
> should_call_unified_connector_service_for_frm( lets reuse exisiting generic one (should_call_unified_connector_service)? or is there any blocker in that?

Yes FRM markdown for ucs is not present, plus the shadowMode mapping and other bypass feature buildup is not done. So created it separate
@gopikrishna000

crates/router/src/core/fraud_check.rs
```diff
@@ -129,6 +129,142 @@ where
...
+                crate::core::unified_connector_service::frm::call_unified_connector_service_for_frm_pre_risk_check(
+                    state,
+                    platform.get_processor(),
+                    merchant_connector_account.clone(),
+                    frm_data.connector_details.connector_name.clone(),
+                    &frm_data.connector_details.profile_id,
+                    context,
+                )
+                .await?;
+
```
Saptak88: Can we use the gateway abstraction to branch between the direct gateway (execute_connector_processing_steps) and the UCS gateway (as used in the payments and payout flows), and use grpc_logging_wrapper_granuler for the UCS call as It handles event publishing and error response extraction from tonic::Status.

crates/router/src/core/fraud_check.rs
```diff
@@ -129,6 +131,67 @@ where
...
+                execution_path,
+                execution_mode: common_enums::ExecutionMode::Primary,
+            };
+
+            return router_data
+                .decide_frm_flows_via_ucs(state, gateway_context)
+                .await;
+        }
+    }
+
```
Saptak88: Instead of returning early, pass gateway_context to decide_frm_flows, and call gateway::execute_payment_gateway() from decide_frm_flow() in checkout_flow.rs (don't need decide_frm_flows_via_ucs).

gopikrishna000:
Three things worth cleaning up before merge, all in the new FRM→UCS bridge.

## 1. `frm_pre_risk_check` drops error detail on the floor

`UnifiedConnectorServiceClient::frm_pre_risk_check` uses `.change_context(UnifiedConnectorServiceError::FrmPreRiskCheckFailure)`. `from_grpc_error(status, connector_name)` already exists, is flow-agnostic, and decodes `status.details()` into `ConnectorError`/`IntegrationError`/`TonicStatus` — used by `payment_authorize`/`payment_get`/`payment_refund`/`payout_*`. Swap:
```rust
.map_err(|e| error_stack::Report::new(UnifiedConnectorServiceError::from_grpc_error(&e, &connector_name)))
```
(`surcharge_calculate` has the same gap and is self-contained to itself, easy same-PR fix. `notify_connector` has it too but is shared with the existing non-FRM dispute-notify path — still low-risk since the richer variants are already handled everywhere downstream, but out of scope for this PR, so flagging separately rather than bundling it in.)

## 2. The request builder belongs in `transformers.rs`

Every flow's `RouterData → proto request` builder — including all of Payouts — lives in `unified_connector_service/transformers.rs`. `impl ForeignTryFrom<&RouterData<frm_api::Checkout,...>> for FrmServicePreRiskCheckRequest` is the first to live elsewhere (`unified_connector_service/frm.rs`). `frm.rs` for the FRM-specific orchestration (gate, access-token fetch, notification) is right and mirrors `kill_switch.rs`'s precedent — just not this impl.

Inside it, `access_token` is hand-built instead of reusing `transformers.rs`'s `impl ForeignFrom<&AccessToken> for ConnectorState` (`.map(ConnectorState::foreign_from)`, used by every other flow) — and skips the `.expose()` that impl has, worth a compile check.

`frm_metadata` → `connector_feature_data` being inlined is fine — every flow builds that field inline from its own `connector_meta`, no shared helper exists for it today.

## 3. `get_frm_access_token` bypasses the granular audit wrapper

Every other UCS call reachable from a flow goes through `ucs_logging_wrapper_granular` (comprehensive request/response/timing event) — including the *identical* `create_access_token` RPC for payments, via `access_token_gateway.rs`. `frm.rs::get_frm_access_token` instead calls `ucs_client.create_access_token(...)` directly, so FRM token fetches get only the bare transport-layer `logger::error!` on failure, not the audit event the payments path gets for the same call. Worth routing it through the wrapper too, or noting why it's intentionally excluded.

gopikrishna000 (approved):

gopikrishna000 (dismissed):

GauravRawat369 (dismissed):

crates/hyperswitch_domain_models/src/router_request_types/fraud_check.rs
```diff
@@ -37,6 +37,14 @@ pub struct FraudCheckCheckoutData {
...
     pub phone: Option<Secret<String>>,
     pub phone_country_code: Option<String>,
+    /// Buyer's full name from the intent's customer details. Risk providers
+    /// match buyer history on it alongside `email`/`phone`.
+    pub customer_name: Option<Secret<String>>,
+    /// The full instrument being scored (card BIN, last four, expiry). The
+    /// `payment_method_data` above is the thinner `AdditionalPaymentData`
+    /// shape the in-process connectors consume; UCS-backed providers need the
+    /// domain type.
+    pub payment_method_data_full: Option<PaymentMethodData>,
```
Saptak88, line 48: If we are adding payment_method_data_full. Is the masked one still needed?

crates/router/src/core/fraud_check/operation/fraud_check_post.rs
```diff
@@ -155,6 +155,8 @@ impl GetTracker<PaymentToFrmData> for FraudCheckPost {
                     order_details: payment_data.order_details,
                     refund: None,
                     frm_metadata: payment_data.frm_metadata,
+                    payment_method_data: payment_data.payment_method_data,
+                    payment_method_token: payment_data.payment_method_token,
                 };
```
Saptak88, line 160: why is payment_method_token needed

crates/router/src/core/fraud_check/gateway.rs
```diff
@@ -0,0 +1,210 @@
...
+            )
+            .await
+            .unwrap_or_else(|err| {
+                logger::warn!(
+                    error = ?err,
+                    "Failed to obtain an FRM access token; continuing without one"
+                );
+                None
+            });
+        }
```
Saptak88, line 107: if access token is required I think we should use the access_token gateway

crates/router/src/core/unified_connector_service.rs
```diff
@@ -3065,6 +3074,37 @@ pub fn handle_unified_connector_service_response_for_create_connector_customer(
...
+) -> CustomResult<FraudCheckResponseData, UnifiedConnectorServiceError> {
+    use payments_grpc::FrmDecision;
+
+    // Same status-code handling every payments UCS handler performs.
+    let status_code = transformers::convert_connector_service_status_code(response.status_code)?;
+
+    let decision = response
+        .frm_decision
+        .and_then(|decision| FrmDecision::try_from(decision).ok());
+
```
Saptak88: why are we using .ok? Also if it is none I think we should throw error

crates/router/src/core/unified_connector_service.rs
```diff
@@ -3065,6 +3074,37 @@ pub fn handle_unified_connector_service_response_for_create_connector_customer(
...
+    let status = transformers::frm_status_from_ucs_decision(status_code, decision);
```
Saptak88: I think we dont need frm_status_from_ucs_decision, it will reach here only in case of 2xx.

crates/hyperswitch_domain_models/src/router_request_types/fraud_check.rs
THE-CORE-BUG: Still needed. It's read by the three in-process FRM connectors — Signifyd ,Riskified, CyberSource DM — for BIN/last4/holder-name projections that AdditionalCardInfo carries directly.

crates/router/src/core/fraud_check/operation/fraud_check_post.rs
THE-CORE-BUG: It's the decrypted wallet token (Apple Pay/Google Pay), set before FRM runs. The shared build_unified_connector_service_payment_method needs it to encode a wallet instrument for UCS before sending request

crates/router/src/core/unified_connector_service.rs
THE-CORE-BUG: Updating with Serialization decode handling

crates/router/src/core/unified_connector_service/transformers.rs
```diff
@@ -9989,3 +10009,145 @@ impl transformers::ForeignTryFrom<payments_grpc::NotifyConnectorResponse>
         })
     }
 }
+
+use hyperswitch_domain_models::{
+    router_flow_types::fraud_check as frm_api,
+    router_request_types::fraud_check::FraudCheckCheckoutData,
+    router_response_types::fraud_check::FraudCheckResponseData,
+};
+
```
Saptak88: nit: We can import this at the top

Saptak88 (dismissed):

crates/router/src/core/unified_connector_service.rs
```diff
@@ -3072,6 +3076,69 @@ pub fn handle_unified_connector_service_response_for_create_connector_customer(
...
+    }
+
+    let decision = response
+        .frm_decision
+        .and_then(|decision| FrmDecision::try_from(decision).ok())
+        .ok_or(UnifiedConnectorServiceError::ResponseDeserializationFailed)
+        .attach_printable("UCS FRM pre risk check succeeded but returned no decision")?;
+
+    let status = transformers::frm_status_from_ucs_decision(decision)?;
+
```
Saptak88:
```suggestion
    let status = transformers::frm_status_from_ucs_decision(response.frm_decision())?;

```

Saptak88: nit: We can do like this as it already has pub fn frm_decision(&self) -> FrmDecision

Saptak88 (dismissed):

gopikrishna000 (approved):

gopikrishna000 (dismissed):

GauravRawat369 (dismissed):

deepanshu-iiitu (dismissed):

Saptak88 (dismissed):

Sakilmostak (dismissed):

GauravRawat369 (dismissed):

deepanshu-iiitu (dismissed):

gopikrishna000 (dismissed):

crates/router/src/core/unified_connector_service/frm.rs
```diff
@@ -0,0 +1,284 @@
...
+        AccessToken,
+    > = connector.connector.get_connector_integration();
+
+    let gateway_context = payments::gateway::context::RouterGatewayContext {
+        creds_identifier: None,
+        processor: processor.clone(),
+        header_payload: hyperswitch_domain_models::payments::HeaderPayload::default(),
+        lineage_ids,
+        merchant_connector_account: merchant_connector_account.clone(),
+        execution_path: common_enums::ExecutionPath::UnifiedConnectorService,
```
pixincreate: why ucs is hard coded here?

crates/router/src/core/unified_connector_service/frm.rs
```diff
@@ -0,0 +1,284 @@
...
+// generic sender below owns that shared work so each event is a thin caller.
+// Chargeback is wired here; the payment/refund events follow the same shape.
+// ─────────────────────────────────────────────────────────────────────────────
+
+/// Which lifecycle event to report, plus the detail payload it carries.
+pub enum FrmNotification {
+    /// A chargeback/dispute was opened against a previously-scored payment.
+    ChargebackReceived {
+        connector_dispute_id: Option<String>,
+        merchant_dispute_id: Option<String>,
```
pixincreate: shouldn't these be a secret?

crates/router/src/core/fraud_check.rs
```diff
@@ -130,6 +132,70 @@ where
...
+        frm_data.fraud_check.last_step,
+        FraudCheckLastStep::Processing
+    );
+
+    // Same routing decision every UCS flow makes. A UCS-only FRM provider is a
+    // `ConnectorIntegrationType::UcsConnector` (listed in `ucs_only_connectors`),
+    // for which `decide_execution_path` returns UCS unconditionally and the kill
+    // switch is skipped — there is no direct integration to divert to.
+    // Everything else resolves to Direct.
+    let execution_path = if is_pre_risk_evaluation {
```
pixincreate, line 190: if ucs is disabled or unavailable, `should_call_unified_connector_service` falls back to `ExecutionPath::Direct` with only a debug log. for a ucs-only provider like nsure, the direct path runs the stub connector, whose default `build_request` returns `Ok(None)`, so the whole flow is a no-op and the seeded `FraudCheckStatus::Pending` response is what lands in the tracker. `should_stop_payment` never stops on `Pending`, so the payment proceeds unchecked even when the merchant configured `PreFrmFailureMode::FailClosed`. the same silent skip happens if nsure is simply missing from `ucs_only_connectors`. can we map the skipped check to `TransactionFailure` (which fail_closed already honors), or at least log at warn with a metric, instead of silently approving the bypass?

crates/router/src/core/connector_validation.rs
```diff
@@ -650,6 +650,20 @@ impl ConnectorAuthTypeAndMetadataValidation<'_> {
                 zsl::transformers::ZslAuthType::try_from(self.auth_type)?;
                 Ok(())
             }
+            api_enums::Connector::Nsure => {
```
pixincreate, line 653: this accepts both `BodyKey` and `HeaderKey`, but `ConnectorSpecificConfig::foreign_try_from` in `unified_connector_service/connector_config.rs` only matches `BodyKey` and errors with "nsure requires BodyKey auth type". a merchant onboarded with `HeaderKey` passes mca validation here and then fails every risk check when the config header is built. either restrict this arm to `BodyKey`, or handle `HeaderKey` in the config builder (the ucs-side `NsureConfig` proto has `app_id` as optional, so a header-key-only mapping is representable).

crates/router/src/core/fraud_check.rs
```diff
@@ -959,3 +1028,135 @@ pub async fn make_fulfillment_api_call(
...
+///
+/// Self-contained entry point for the dispute webhook flow: resolves the
+/// `fraud_check` row for the payment, checks the provider is UCS-backed and
+/// configured, and sends `FRM_CHARGEBACK_RECEIVED`.
+///
+/// Best-effort by design — a chargeback has already happened, so failing to
+/// inform the provider must never fail the webhook. Errors are logged and
+/// swallowed, matching how the pre-connector FRM call is treated.
+#[cfg(all(feature = "v1", feature = "frm"))]
+pub async fn notify_frm_of_chargeback(
```
pixincreate, line 1287: this function has no call site in this pr, and neither do `FrmNotification` / `call_unified_connector_service_for_frm_notification` outside their definitions. can this move to the follow-up pr that wires the dispute webhook? as merged it is unreachable, so it gets no coverage from the end-to-end testing described in the pr and can drift before it is ever called.

crates/hyperswitch_domain_models/src/router_request_types/fraud_check.rs
```diff
@@ -24,7 +24,7 @@ pub struct FraudCheckSaleData {
     pub payment_method_data: Option<api_models::payments::AdditionalPaymentData>,
 }
 
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, Serialize)]
```
pixincreate, line 33: this `Serialize` derive is here to satisfy the `Req: serde::Serialize` bound on `execute_payment_gateway`, right? with `payment_method_data_full` this struct now carries the full pan, and plain `Serialize` on `Secret` / `CardNumber` exposes the raw value (only `masked_serialize` masks). the ucs logging wrapper does use `masked_serialize`, so the current sinks look safe, but a doc comment here warning against feeding this type to a raw serializer would be cheap insurance against a future log or event leaking card data.

crates/router/src/core/payments/gateway/access_token_gateway.rs
```diff
@@ -102,11 +102,14 @@ where
...
-        // A merchant-authentication (access-token) call can originate from either a
-        // payment or a payout. The connector type selects the UCS connector header
-        // namespace, while the ids below carry the payment/payout reference context.
+        // A merchant-authentication (access-token) call can originate from a
+        // payment, a payout, or an FRM pre-risk check. The connector type selects
+        // the UCS connector header namespace, while the ids below carry the
+        // payment/payout reference context.
         let connector_type = if router_data.payout_id.is_some() {
             ConnectorType::PayoutProcessor
+        } else if api_models::enums::FrmConnectors::from_str(&router_data.connector).is_ok() {
```
pixincreate: nit: inferring the connector type by string-parsing the name into `FrmConnectors` is fragile. any future connector whose name collides with an frm connector would get `PaymentVas` headers on its access token call. the caller already knows this request came from the frm path; carrying an explicit connector type (or flag) on `RouterGatewayContext` would state the intent instead of deriving it from the name.

crates/external_services/src/grpc_client/unified_connector_service.rs
```diff
@@ -1565,7 +1578,12 @@ impl UnifiedConnectorServiceClient {
             .clone()
             .calculate(request)
             .await
```
pixincreate, line 1580: this swaps the surcharge flow's blanket `SurchargeCalculateFailure` for `from_grpc_error` decoding, so connector 4xx/5xx now surface as `ConnectorError` / `TonicStatus` instead of being blamed on ucs. the change itself looks right (matches the payments flows, and the caller in `payments.rs` is best-effort so behavior is safe), but it also shifts kill-switch reason attribution and metric/log shape for a flow this pr does not otherwise touch, and the description does not mention it. can it be called out there, or split into its own pr?

crates/router/src/core/unified_connector_service/transformers.rs
```diff
@@ -9989,3 +10011,139 @@ impl transformers::ForeignTryFrom<payments_grpc::NotifyConnectorResponse>
...
+        let merchant_details = Some(payments_grpc::MerchantDetails {
+            merchant_id: Some(router_data.merchant_id.get_string_repr().to_owned()),
+            merchant_category_code: None,
+        });
+
+        let browser_info =
+            request
+                .browser_info
+                .as_ref()
+                .map(|info| payments_grpc::BrowserInformation {
```
pixincreate: there is already a full `ForeignTryFrom<router_request_types::BrowserInformation>` for `payments_grpc::BrowserInformation` in this file (15 fields), and `FraudCheckCheckoutData.browser_info` is the same domain type. this hand-rolled subset drops `screen_height`/`screen_width`, `time_zone_offset_minutes`, `os_type`/`os_version`, `device_model`, `java_script_enabled`, `accept_language` and `referer`, which are exactly the device-fingerprint signals a risk provider wants. any reason not to reuse the existing converter?

crates/router/src/core/unified_connector_service/frm.rs
```diff
@@ -0,0 +1,284 @@
...
+    pub profile_id: &'a id_type::ProfileId,
+}
+
+/// Send an FRM lifecycle notification to the connector-service.
+///
+/// Generic over the event: the caller supplies the [`FrmNotification`] variant
+/// and this handles auth metadata, `x-frm-connector` routing, the access token
+/// for bearer-authenticated providers, and the gRPC call.
+#[cfg(feature = "v1")]
+pub async fn call_unified_connector_service_for_frm_notification(
```
pixincreate, line 204: the doc says this handles "the access token for bearer-authenticated providers", but the function never fetches or attaches one: no `get_frm_access_token` call, and `NotifyConnectorRequest` is built with `..Default::default()` so no connector state is sent. for a bearer-authenticated provider like kount, the chargeback notification would reach ucs without a token and fail auth at the provider. either wire the token the way the pre-risk-check gateway does, or fix the doc so the follow-up events are not built on a wrong assumption.

crates/router/src/core/unified_connector_service/frm.rs
```diff
@@ -0,0 +1,284 @@
...
+        payments::{self, helpers::MerchantConnectorAccountType},
+    },
+    routes::SessionState,
+    services,
+    types::{api::FraudCheckConnectorData, transformers::ForeignTryFrom},
+};
+
+/// Fetch (and cache) the OAuth token a bearer-authenticated FRM provider needs.
+///
+/// Kount's risk API is bearer-authenticated: prism reads the token from
```
pixincreate: nit: "prism" here (and at the lifecycle-notifications block comment below) is the internal codename; in this repo the public name is connector-service / ucs. same concern that came up on the design doc in an earlier thread, just for code comments. worth a rename for oss readers.

loadtest/config/development.toml
```diff
@@ -216,6 +216,7 @@ noon.base_url = "https://api-test{{region}}.noonpayments.com/"
 nordea.base_url = "https://api.nordeaopenbanking.com"
 noon.key_mode = "Test"
 novalnet.base_url = "https://payport.novalnet.de/v2"
+nsure.base_url = "https://sandbox.nsureapi.com"
```
pixincreate, line 219: consider adding base urls for other deployment config envs - integ, sandbox and prod

crates/router/src/core/unified_connector_service/frm.rs
```diff
@@ -0,0 +1,284 @@
...
+// prism exposes four FRM notification events on a single `NotifyConnector` RPC:
+//
+//   FRM_PAYMENT_SUCCEEDED · FRM_PAYMENT_FAILURE
+//   FRM_REFUND_PROCESSED  · FRM_CHARGEBACK_RECEIVED
+//
+// They differ only in the event type and which `notification_type` variant is
+// populated; auth, headers, token and response handling are identical. The
+// generic sender below owns that shared work so each event is a thin caller.
+// Chargeback is wired here; the payment/refund events follow the same shape.
+// ─────────────────────────────────────────────────────────────────────────────
```
pixincreate: nit: you can also consider removing or trimming these verbose comments that adds little to no value.

pixincreate:
the pr is failing migration compatibility check. can you check that -- https://github.com/juspay/hyperswitch/actions/runs/34957628201/job/104343431066?pr=13987

crates/router/src/core/connector_validation.rs
hrithikesh026: @THE-CORE-BUG could you please address this.

crates/router/src/core/fraud_check.rs
THE-CORE-BUG: I will return an error message now

crates/router/src/core/unified_connector_service/frm.rs
THE-CORE-BUG: I have fetched now from gateway_context

crates/router/src/core/connector_validation.rs
THE-CORE-BUG: Sure, I will handle for Header-Key as well

crates/router/src/core/fraud_check.rs
THE-CORE-BUG: Keeping the implementation now was a desired decision, will hook all this up once the creds are available in upcoming PRs.

crates/external_services/src/grpc_client/unified_connector_service.rs
THE-CORE-BUG: Updated the description, the change was also requested by @gopikrishna000 and is verified

loadtest/config/development.toml
THE-CORE-BUG: Already present right

THE-CORE-BUG:
> the pr is failing migration compatibility check. can you check that -- https://github.com/juspay/hyperswitch/actions/runs/34957628201/job/104343431066?pr=13987

Already present on main, nothing on this PR, ran locally

pixincreate:
clippy checks are failing:

```
error: large future with a size of 16728 bytes
   --> crates/router/src/core/fraud_check/gateway.rs:128:38
    |
128 |                   let response = match client
    |  ______________________________________^
129 | |                     .frm_pre_risk_check(
130 | |                         pre_risk_check_request,
131 | |                         connector_auth_metadata,
132 | |                         grpc_headers,
133 | |                     )
    | |_____________________^
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/rust-1.98.0/index.html#large_futures
    = note: `-D clippy::large-futures` implied by `-D warnings`
    = help: to override `-D warnings` add `#[allow(clippy::large_futures)]`
help: consider `Box::pin` on it
    |
128 ~                 let response = match Box::pin(client
129 +                     .frm_pre_risk_check(
130 +                         pre_risk_check_request,
131 +                         connector_auth_metadata,
132 +                         grpc_headers,
133 +                     ))
    |

error: this function has too many arguments (9/7)
    --> crates/router/src/core/fraud_check.rs:1064:1
     |
1064 | / pub async fn notify_frm_of_chargeback(
1065 | |     state: &SessionState,
1066 | |     platform: &domain::Platform,
1067 | |     payment_id: &common_utils::id_type::PaymentId,
...    |
1073 | |     chargeback_reason: Option<String>,
1074 | | ) {
     | |_^
     |
     = help: for further information visit https://rust-lang.github.io/rust-clippy/rust-1.98.0/index.html#too_many_arguments
     = note: `-D clippy::too-many-arguments` implied by `-D warnings`
     = help: to override `-D warnings` add `#[allow(clippy::too_many_arguments)]`

error: could not compile `router` (lib) due to 2 previous errors
warning: build failed, waiting for other jobs to finish...
error: could not compile `router` (lib test) due to 2 previous errors
error: recipe `clippy` failed with exit code 101
Error: Process completed with exit code 101.
0s
0s
0s
```

pixincreate (dismissed):
prism changes, lgtm

GauravRawat369 (dismissed):

hrithikesh026 (dismissed):

Sakilmostak (dismissed):

deepanshu-iiitu (dismissed):

gopikrishna000 (dismissed):

hrithikesh026 (dismissed):

pixincreate (dismissed):

GauravRawat369 (dismissed):

pixincreate (approved):

hrithikesh026 (approved):

GauravRawat369 (approved):

gopikrishna000 (approved):

AkshayaFoiger (approved):

deepanshu-iiitu (approved):

PR #13999, opened 2026-09-02
feat(router): add browse offers API for offer engine

spritianeja03:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

Adds `POST /offer_engine/offers/list`, which returns the offers a merchant can currently use, for a merchant dashboard browse screen.

Internally it calls Offer Engine's `v1/offers/list` — the same upstream endpoint the payment eligibility flow uses — and filters the result to eligible offers only, so the caller sees just what is usable.

**Request**

```json
{
  "offer_payment_info": {
    "currency": "USD"
  }
}
```

`offer_payment_info` is optional; omit it to browse without an order context.

**Response**

```json
{
  "offers": [
    {
      "code": "TESTHS",
      "title": "Test offer HS",
      "display_title": "TSH",
      "description": "Testing offers",
      "currency": "USD",
      "valid_till": "2026-08-31T18:29:59.000Z"
    }
  ]
}
```

A few notes on the implementation:

- **Nothing existing is modified.** Browse gets its own request/response types, its own microservice flow, and its own core module. `OfferListRequest`, `OfferOrder`, `OfferListResponse`, `OfferListEntry`, `eligibility.rs` and the payment flow are all untouched. Additions to `types.rs` and `client.rs` are appends only.
- The new `OfferBrowseList` flow targets the same `v1/offers/list` path as the existing `OfferList` and `OfferConnectivity` flows, which already share it.
- Offer Engine credentials are resolved through `resolve_offer_engine_credential_source`, the same path `validate-token` uses, which doubles as the enablement gate: a merchant without Offer Engine enabled gets a `403`.
- Auth is API key only. This endpoint is for the merchant dashboard, not the Hyperswitch dashboard.
- `best_offer_combinations` is deliberately not modelled. It carries only offer ids and discounts for offers already present in `offers`, and this response reports no discount, so nothing is lost by ignoring it.
- Every optional field on the upstream response is defaulted, so the response parses whether or not Offer Engine returns `display_title`, `currency` or `valid_till`.

### Additional Changes

New endpoint only. No existing endpoint or response shape changes.

## Motivation and Context

The merchant dashboard needs to show which offers are available to a merchant. Offer Engine's `/list` returns both eligible and ineligible offers along with recommended combinations, none of which a browse screen should have to reason about — so this endpoint does the filtering and returns a flat list of usable offers.

Closes https://github.com/juspay/hyperswitch-cloud/issues/22768

## How did you test it?

Request
```
curl --location 'http://localhost:8080/offer_engine/offers/list' \
--header 'api-key: <api_key> \
--header 'content-type: application/json' \
--data '{"offer_payment_info": {"currency": "USD"}}'
```

Response
```json
{
    "offers": [
        {
            "code": "TESTHS1",
            "title": "Test offer HS",
            "display_title": null,
            "description": "Testing offers",
            "currency": null,
            "valid_till": null
        },
        {
            "code": "HSVELO1",
            "title": "HS Velocity Once Per Card",
            "display_title": null,
            "description": "Velocity test - once per card",
            "currency": null,
            "valid_till": null
        }
    ]
}
```

## Checklist


crates/router/src/routes/offer_engine.rs
```diff
@@ -24,3 +25,33 @@ pub async fn offer_engine_connectivity_check(
...
+            offer_engine::browse::browse_offers(state, auth.platform, profile_id, payload)
+        },
+        &auth::HeaderAuth(auth::ApiKeyAuth {
+            allow_connected_scope_operation: true,
+            allow_platform_self_operation: false,
+        }),
+        api_locking::LockAction::NotApplicable,
+    ))
+    .await
+}
```
ThisIsMani, line 53: Let's only enable this api in v1. Offers are only enabled in v1 afaik.

crates/router/src/core/offer_engine/browse.rs
```diff
@@ -0,0 +1,121 @@
...
+                .filter(|entry| entry.status == OfferStatus::Eligible)
+                .map(to_browse_offer)
+                .collect(),
+        },
+    ))
+}
+
+/// Resolves the Offer Engine credentials for the merchant behind the request. Anything that
+/// leaves us without a usable config means Offer Engine is not available to them.
+async fn resolve_offer_config(
```
ThisIsMani: I think we have a function for this already...

crates/router/src/core/offer_engine/browse.rs
```diff
@@ -0,0 +1,121 @@
...
+
+    BrowseOffer {
+        code: entry.offer_code,
+        title,
+        display_title: entry.display_title,
+        description,
+        currency: entry.currency,
+        valid_till: entry.valid_till,
+    }
+}
```
ThisIsMani: Can be moved to From impl, and let's not put this in core... Can be moved to types probably.

crates/router/src/lib.rs
```diff
@@ -380,7 +380,10 @@ pub fn mk_app(
 
     server_app = server_app.service(routes::Cache::server(state.clone()));
     server_app = server_app.service(routes::Health::server(state.clone()));
-    server_app = server_app.service(routes::OfferEngine::server(state.clone()));
+    #[cfg(feature = "v1")]
+    {
+        server_app = server_app.service(routes::OfferEngine::server(state.clone()));
+    }
```
ThisIsMani: We can move this top where we attach all the v1 services.

crates/router/src/core/offer_engine/browse.rs
ThisIsMani: We could move this to the place we declared other 2 similar functions and name this different from what if we already have.

apoorvdixit88: @spritianeja03 Can we do the resolve offer config in the same way we are doing at other places? First we get the context and then do the match to get the config. You can take reference from other places.

crates/router/src/core/offer_engine/browse.rs
```diff
@@ -0,0 +1,106 @@
...
+    };
+
+    let client = OfferEngineClient::new(config, &state.conf.trace_header.header_name);
+    let response = client
+        .browse_offers(&state, list_request)
+        .await
+        // No gateway-level variant exists on `ApiErrorResponse`, and `ConnectorError` would
+        // mislabel this as a payment connector failure, so an unreachable Offer Engine surfaces
+        // as a 500 — the same treatment the eligibility flow gives it.
+        .change_context(errors::ApiErrorResponse::InternalServerError)
```
apoorvdixit88: nit* please remove this comment

apoorvdixit88 (commented):
Can we please remove `resolve_offer_config`, and use the existing one. That we already have.

ThisIsMani (dismissed):

apoorvdixit88 (dismissed):

crates/router/src/routes/offer_engine.rs
```diff
@@ -24,3 +25,29 @@ pub async fn offer_engine_connectivity_check(
     ))
     .await
 }
+
+#[cfg(feature = "v1")]
+#[instrument(skip_all, fields(flow = ?Flow::OfferEngineBrowseOffers))]
+pub async fn offer_engine_browse_offers(
```
hrithikesh026, line 31: Should generate OpenApi spec for this new endpoint.

crates/router/src/core/offer_engine/types.rs
```diff
@@ -327,3 +327,50 @@ pub struct OfferNotifyRequest {
...
+/// A per-offer entry in a browse response.
+#[derive(Debug, Clone, serde::Deserialize)]
+pub struct BrowseOfferListEntry {
+    pub status: OfferStatus,
+    pub offer_code: String,
+    #[serde(default)]
+    pub offer_description: Option<OfferDescription>,
+    #[serde(default)]
+    pub display_title: Option<String>,
+    #[serde(default)]
```
hrithikesh026: no need to mention explicit serde(default) for Option fields

crates/router/src/core/offer_engine/types.rs
```diff
@@ -327,3 +327,50 @@ pub struct OfferNotifyRequest {
...
+    pub currency: Option<common_enums::Currency>,
+    #[serde(default, with = "common_utils::custom_serde::iso8601::option")]
```
hrithikesh026, line 354:
```suggestion
    #[serde(with = "common_utils::custom_serde::iso8601::option")]
```

hrithikesh026 (approved):

apoorvdixit88 (approved):

VenuMadhav2541 (approved):

PR #14000, opened 2026-09-02
refactor(deja): let the gRPC boundary hold an optional transport

### LINKED ISSUE #14027: deja gRPC boundary cannot hold an absent transport, so a UCS call cannot be substituted on replay
The deja gRPC boundary requires a live transport to be constructed. When the
unified connector service is unreachable, the client cannot be built at all, so
the boundary has nothing to attach to and a recorded UCS call cannot be
substituted from a tape.

That makes a whole class of recordings unscoreable: a replay in an environment
without UCS reaches the boundary, finds no transport, and fails rather than
answering from the recording. It also means the failure mode we most want to
verify — the shadow comparison path when UCS is down — is exactly the one a
replay cannot exercise.

The boundary should be able to hold an absent transport and decide, at call
time, whether to connect or to answer from the recording.

maverox:
Closes #14027

## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [x] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

**Base of a 2-PR stack. This PR changes no behaviour on its own.**

`DejaGrpcTransport` presents itself as a generic tower `Service` wrapper, but it carried a
hidden precondition: it required a successfully constructed inner service. Under replay it
never used one — substitution happens in `call`, before the inner service is reached — so the
transport beneath the boundary is dead weight there, demanded by the type and unused by the
behaviour.

That precondition is what forces a host to change *how it builds its transport* in order to be
observed. A host that connects eagerly and fallibly — a reasonable thing to do, and what
`Endpoint::connect` is for — cannot install the boundary at all when its connect fails, even
under replay, where the connect could not have succeeded and would not have mattered.

This PR removes the precondition:

- `DejaGrpcTransport` holds `inner: Option<S>`. `new()` behaves exactly as before.
- `substituted()` builds a boundary with nothing beneath it — valid only where every call is
  served from the recording.
- `connect_or_substitute` is the seam for a host with an eager connect. It is generic over the
  transport *and* over the host's own connect closure (no tonic or UCS types in its signature),
  and runs that closure verbatim in every mode except replay.
- `poll_ready` answers directly when there is no transport; the delegating case is unchanged.

Nothing constructs a transport-less boundary yet — the first consumer is the stacked PR — so
this is additive and behaviour-neutral.

### A separate fail-stop, and why

A boundary with no transport needs its own fail-stop rather than reusing the substitute-miss
one. Two reasons:

1. The existing message tells the reader that re-running is unsafe and to declare
   `replay_strategy = Execute`. With no transport there is nothing to re-run and `Execute` is
   not available, so it would hand a reader an instruction that cannot work.
2. The two states put the burden on different people. A miss with the transport **present** may
   be a genuine novel call whose author needs to look at it. A miss with the transport
   **absent** establishes nothing about the candidate at all — the boundary could not have
   served the call whatever the candidate did. Collapsing them makes an instrumentation-parity
   problem indistinguishable from a candidate defect.

The new message carries the same `FAIL_STOP_SENTINEL` (via the public export, not a copied
literal) so a censored run stays detectable from the diff artifact, and it names the **rpc path
and authority** — `BOUNDARY`/`COMPONENT`/`OPERATION` are module constants shared with the
dynamic routing, health check and recovery decider pool, so they cannot say *which* transport
was missing.

### Additional Changes

## Motivation and Context

Under a deja replay, a service that is deliberately unreachable could prevent its recording
boundary from being installed at all — which silently removes the recorded feature from the
candidate's reachable behaviour and reports it as a candidate divergence. Fixing that at the
host (by making it connect lazily) would change production behaviour and make a *recording*
diverge from the deployment it records. The accommodation belongs here, in the boundary, which
is already the deja-specific component.

A known fragility is documented rather than fixed: `poll_ready` delegates to the inner service
ungated by `is_active()`, which is safe only because tonic answers readiness from its buffer
rather than from a connection (`connect_lazy` → `Channel::new` → `Buffer::pair`; tonic 0.14.5).
An inner service answering readiness from a live connection would fail there, before `call` can
substitute. The no-transport case no longer delegates at all; the delegating case is left alone
with the reasoning recorded at the call site, and is tracked deja-side where the mode's
lifecycle lives.

## How did you test it?

Compile gates only — this PR adds no reachable behaviour. Run unpiped with
`RUSTFLAGS="-D warnings"` as CI sets it, exit code taken from the command itself:

| gate | toolchain | exit |
|---|---|---|
| `just clippy` | rustc 1.97.1 | 0 |
| `just clippy_v2` | rustc 1.97.1 | 0 |
| `cargo check --features "release"` (CI `check-msrv`) | rustc 1.85.0 | 0 |

Worth flagging for reviewers: **neither clippy gate isolates the feature-off arm.**
`justfile:47` removes `deja` from the v2 feature list, but `crates/router/Cargo.toml:62` lists
`deja` inside `release`, which the filter keeps — so `release` re-enables it transitively. Both
gates lint the `deja` arm; `cfg(not(feature = "deja"))` compiles but no gate exercises it alone.

## Checklist


crates/external_services/src/grpc_client/deja_transport.rs
```diff
@@ -40,18 +40,140 @@ const OPERATION: &str = "unary";
...
+/// unchanged.
+pub async fn connect_or_substitute<S, F, Fut>(connect: F) -> Option<DejaGrpcTransport<S>>
+where
+    F: FnOnce() -> Fut,
+    Fut: Future<Output = Option<S>>,
+{
+    if deja::__private::runtime_mode().is_replay() {
+        return Some(DejaGrpcTransport::substituted());
+    }
+    connect().await.map(DejaGrpcTransport::new)
```
maverox:
```suggestion
    if deja::__private::runtime_mode().is_replay() {
       Some(DejaGrpcTransport::substituted())
    } else {
        connect().await.map(DejaGrpcTransport::new)
    }
```

maverox: avoid return keyword for such cases unless absolutely necessary (short circuit behaviour is needed)

crates/external_services/src/grpc_client/deja_transport.rs
```diff
@@ -40,18 +40,140 @@ const OPERATION: &str = "unary";
...
+impl std::fmt::Display for AbsentTransportError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.write_str(
+            "deja: gRPC boundary has no transport (constructed for substitution only) \
+             and cannot issue this call live",
+        )
+    }
+}
+
+impl std::error::Error for AbsentTransportError {}
```
maverox, line 136: Don't you think this belongs in the deja layer abstracted away?

jarnura (dismissed):

hrithikesh026 (dismissed):

ThisIsMani (approved):

hrithikesh026 (approved):

jarnura (approved):

PR #14009, opened 2026-09-02
fix(ucs): say when a missing proxy override disables UCS shadow mode

### LINKED ISSUE #14031: A missing proxy override disables UCS shadow mode without saying so
When the proxy override is absent, shadow mode is turned off silently at debug level. Nothing in the ordinary log stream states that the comparison is no longer happening, so a deployment can run for a long time believing it is shadowing when it is not.

An absent configuration that changes behaviour should say so at a level someone reads, and should name the consequence rather than the condition.

maverox:
Closes #14031

## Type of Change

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

**Third in a stack — #14000 → #14001 → this. One log statement, deliberately its own commit so
it can be dropped without disturbing the other two.**

When a Shadow UCS execution path has no `proxy_override`, it is not merely un-proxied: it is
downgraded to `ExecutionPath::Direct` and **the shadow comparison never runs**
(`crates/router/src/core/unified_connector_service.rs:1055`). That happens at `debug!`, so at
default log levels a comparison that was configured and expected can be entirely absent with
nothing saying why.

This raises it to `info!` and makes the message name the consequence rather than the mechanism.
The old text —

> No proxy override available for Shadow UCS, Using the Original State and Sending Request Directly

— describes the state handling but never says that shadow mode is off for this request, so
someone grepping for why comparison traffic is missing would not match it. It now reads:

> No proxy override available for Shadow UCS; falling back to Direct, so no shadow comparison
> will run for this request

The level and the wording fix different problems: the level makes the event visible, the wording
makes it findable.

### Additional Changes

## Motivation and Context

The stack beneath this one fixes a case where UCS shadow-mode comparison traffic was absent from
every replay because the gRPC client was never constructed. With that fixed, a missing
`proxy_override` becomes the **most likely remaining reason** for shadow traffic to be missing —
and it is currently the only one that leaves no trace at default log levels.

This is the line that separates "UCS shadow ran and matched" from "UCS shadow was never asked to
run". That distinction is the first question worth asking whenever comparison traffic is missing,
and answering it should be a log read rather than an investigation.

Note this touches the **router** crate, whereas #14000 and #14001 are confined to
`external_services` — `crates/router/src/core/unified_connector_service.rs` and
`crates/external_services/src/grpc_client/unified_connector_service.rs` share a basename but are
different crates. That is the main reason it is kept as a separate, droppable commit.

## How did you test it?

Compile gates only; this changes a log statement's level and text, and no control flow. Run
unpiped with `RUSTFLAGS="-D warnings"` as CI sets it, exit code taken from the command itself,
and each run verified to have recompiled the `router` crate:

| gate | toolchain | exit |
|---|---|---|
| `just clippy` | rustc 1.97.1 | 0 |
| `just clippy_v2` | rustc 1.97.1 | 0 |
| `cargo check --features "release"` (CI `check-msrv`) | rustc 1.85.0 | 0 |

All three commits in the stack were gated independently, so each PR is green on its own rather
than only as part of the pair.

## Checklist

### DIFF
--- a/crates/external_services/src/grpc_client/deja_transport.rs
+++ b/crates/external_services/src/grpc_client/deja_transport.rs
@@ -40,17 +40,139 @@ const OPERATION: &str = "unary";
 /// The transport wrapper. Generic over the inner tower service so the same
 /// type serves the shared hyper pool (dynamic routing / health / recovery)
 /// and the UCS `tonic::transport::Channel`s.
+///
+/// The inner transport is optional because a host may build its transport
+/// eagerly and fallibly — `tonic::transport::Endpoint::connect` is a perfectly
+/// reasonable choice — and under replay that connect is ceremony: every unary
+/// rpc is substituted here and the transport beneath is never driven. Rather
+/// than ask such a host to become lazy (which would change what it does in
+/// production, and make a recording diverge from the thing it records), the
+/// boundary can stand in for a transport that was never built. See
+/// [`DejaGrpcTransport::substituted`] and [`connect_or_substitute`].
 #[derive(Debug, Clone)]
 pub struct DejaGrpcTransport<S> {
-    inner: S,
+    inner: Option<S>,
 }
 
 impl<S> DejaGrpcTransport<S> {
-    /// Wraps a transport. Call at the construction site, never per request.
+    /// Wraps a live transport. Call at the construction site, never per request.
     pub fn new(inner: S) -> Self {
-        Self { inner }
+        Self { inner: Some(inner) }
+    }
+
+    /// A boundary with no transport beneath it.
+    ///
+    /// Valid ONLY where every call is substituted, i.e. replay. Nothing here is
+    /// a fault: the host's connect legitimately did not succeed (the service is
+    /// unreachable from a replay namespace, and it does not need to be), and
+    /// replay never issues the call live, so there is nothing for a transport to
+    /// do. Readiness is answered directly and every call is served from the
+    /// recording; a call the recording cannot answer fail-stops via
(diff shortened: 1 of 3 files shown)


jarnura (dismissed):

maverox:
Closing without merging.

The change raises the log line that reports a Shadow UCS execution path falling back to Direct when no proxy override is configured, so that a silently disabled shadow comparison becomes visible. On review it cannot do the job it was added for, and it is not free.

It cannot help a replay comparison. The tracing layers that feed a recording implement `on_new_span` and no `on_event`, and a log statement emits an event rather than a span, so nothing a `logger::` macro produces reaches the recorded graph. Changing a level from `debug!` to `info!` is therefore invisible to the comparison in both directions: it can neither introduce a divergence nor remove one. The value it offers is to a person reading logs, not to the diff, and it was proposed on the understanding that it would narrow a divergence class.

It is not free in production. Shadow is the default execution mode when no rollout configuration key matches, so the branch this touches is reached by every eligible request that has no proxy override. At `info!` that is a per-request line in normal operation, which is a poor trade for a diagnostic that only matters when someone is already investigating.

The underlying gap is real and worth recording: a shadow comparison that was configured and expected can be skipped for a configuration reason with nothing at default log levels saying so. That wants a counter, or a signal emitted once per scope rather than once per request, and it should be raised on its own terms rather than as a level change.

The branch is left in place rather than deleted, and it has drifted well behind main in the meantime, so it should be rebuilt rather than rebased if any of this is revisited.

PR #14010, opened 2026-09-02
feat(alerts): define the notifier interface and the webhook API

ThisIsMani:
## Type of Change

- [x] New feature

Targets `observability-plane-i`, not `main`.

## Description

Defines the Notifier's interface and the webhook that receives alerts, on top of the scaffold from
#13966 and beside the chat client from #13968.

Resolves [hyperswitch-cloud#23116](https://github.com/juspay/hyperswitch-cloud/issues/23116).

### The API

Two routes, one per channel. The path names the channel; the body names a **destination id** and
nothing else, so channel ids, recipient addresses and credentials stay in configuration and none of
them travels on the wire.

```http
POST /alerts/chat/notify    { "destination": "sr_alerts", "text": "…", "reply_to": "1503435956.000247" }
                          → { "message_id": "1503435991.000318" }

POST /alerts/email/notify   { "destination": "oncall", "subject": "…", "body": "<pre>…</pre>" }
                          → { }
```

`reply_to` threads a reply under an earlier message, which the R service needs so a recovery notice
lands under the alert it clears. It exists on chat only, and sending it to `/email/notify` is a 400
rather than a dropped field: a recovery notice that quietly loses its link is a bug nobody notices.

**Nothing renders.** Text, subject and body are delivered exactly as they arrive. The R alerts
service already renders a summary in Slack `mrkdwn` for chat and a full per-alert list for email,
and those are not the same message, so the caller owns markup and escaping. `body` is HTML because
both email backends hardcode an HTML body (`email/ses.rs` uses `.html(...)`, `email/smtp.rs` sets
`ContentType::TEXT_HTML`) and there is no plain-text path to reach.

### Decisions worth reviewing

**One route per channel, rejecting `POST /notify/{id}` over a tagged body.** The destination id
already resolves the channel through configuration, so a tag in the body is a second authority on
the same fact and the two can disagree. It would also give every implementation a match arm for a
channel it can never serve, growing by one per channel added.

**Two traits rather than one `Notifier`.** `ChatNotifier` and `EmailNotifier` have separate request
types for the same reason. They exist at all — rather than `alerts` calling
`external_services::chat_service::ChatClient` straight — because that client knows nothing about
alerts and should not have to; these are this crate's own seam.

**Errors split by blame, not by layer.** The provider reports an oversized message and an unknown
channel through the same field of the same response, and they are not the same problem.
`msg_too_long` is a 400 the caller can act on; `channel_not_found` is a 500 about our own
configuration. Answering 500 to an oversized message would invite a retry that can never succeed.
The provider's own code rides in `reason` so a caller can match on it instead of parsing prose, and
a 429 carries `Retry-After`.

| | Status | Code |
|---|---|---|
| Body did not parse | 400 | `IR_04` |
| Unknown destination | 400 | `IR_02` |
| Provider refused, caller can fix it | 400 | `IR_03` |
| Missing or wrong API key | 401 | `IR_01` |
| Rate limited | 429 | `HE_04` |
| Provider refused because of our configuration | 500 | `HE_02` |
| Provider unreachable or unreadable | 502 | `HE_03` |

**Destinations resolve at boot into a registry.** A chat destination is tagged `xyne`, `slack` or
`log`; Xyne and Slack are one client differing in base URL and credential. Destination ids are
validated at boot against what the `config` crate can round-trip from the environment, since it
lowercases every key and splits levels on `__` — so an id like `SR_ALERTS` fails to start rather
than silently failing to match at lookup time.

**No destinations warns; an unbuildable destination fails the boot.** A first deployment has none
until credentials exist, and refusing to start would make the service undeployable before them. But
dropping a destination that is configured and broken would leave the service answering "unknown
destination" to something that is very much configured.

**`log` is a destination type, not a `dry_run` flag.** It accepts and delivers nothing, so the
whole path can be exercised before credentials arrive. A flag would put an `if` at the top of every
implementation; a type keeps the delivery path branchless. It logs sizes and never the message,
because it writes to the same stream as everything else and an alert body carries merchant ids and
payment volumes. The request types have hand-written `Debug` impls for the same reason.

### Deliberately not here

**Email delivery.** How `alerts` reaches `external_services::email` is
[hyperswitch-cloud#23111](https://github.com/juspay/hyperswitch-cloud/issues/23111)'s explicit
decision — `EmailService`'s only method is built for templated product emails and takes a
`base_url` this crate has no use for. Every email destination accepts and logs until then. Chat is
not in that position: `ChatClient::post_message` maps one to one, with nothing to decide.

**A live Xyne channel.** [hyperswitch-cloud#23128](https://github.com/juspay/hyperswitch-cloud/issues/23128)
verifies against real credentials: a message landing in the test channel, the threading round trip,
and `{"ok": false}` at HTTP 200 not surfacing as success.

**Multi-recipient and plain-text email.** `EmailClient::send_email` takes one `pii::Email` and both
backends hardcode HTML, so one destination is one address and the body must be HTML. Filed as
[hyperswitch-cloud#23160](https://github.com/juspay/hyperswitch-cloud/issues/23160). When it lands,
a destination widens to a list without any caller changing, since a request only ever names an id.

### Known behaviour, asserted rather than fixed

actix runs the body extractor before the handler, and the guard runs inside the handler via
`server_wrap`, so a request whose JSON does not parse is answered 400 without its key being
checked. Restoring "guard strictly first" means extracting raw bytes and deserializing by hand,
trading typed extraction for concealment of a schema that is documented anyway.
`tests/notify.rs::an_unparseable_body_is_rejected_before_the_key_is_checked` pins it.

## How did you test it?

35 tests, all new: 25 unit, 10 through the real route tree in `crates/alerts/tests/notify.rs`.
The route tests go through actix rather than calling handlers, because the guard, the extractor's
rejection shape and the status mapping all live between the handler and the caller.

```
cargo test -p alerts --features v1     # 35 passed
cargo clippy -p alerts --features v1 --all-targets    # clean
```

Also run against a real listener with the shipped `config/alerts.toml`, both destinations set to
`log`:

```
chat -> log destination                 200  {"message_id":"log.000001"}
chat threaded reply under log.000001    200  {"message_id":"log.000002"}
email -> log destination                200  {}
email + reply_to (threading rejected)   400  {"error":{…,"code":"IR_04"}}
chat -> unknown destination             400  {"error":{…,"code":"IR_02","destination":"typo"}}
chat with no api key                    401  {"error":{…,"code":"IR_01"}}
```

The service logs from that run carry sizes only, no message content:

```
INFO alerts::core::notifier::chat: not delivered: this destination is configured as `log`,
     tag: "chat_notify_skipped", destination: smoke, chars: 9, threaded: true
```

## Checklist

- [x] `cargo clippy --all-features --all-targets -- -D warnings` (for `alerts`)
- [x] Added tests — 35
- [x] Added documentation — `crates/alerts/README.md` gains the API, error table and destination
      sections


crates/alerts/src/core/notifier/chat.rs
```diff
@@ -0,0 +1,325 @@
...
+    }
+}
+
+#[async_trait::async_trait]
+impl ChatNotifier for ChatClientNotifier {
+    async fn notify(&self, notification: ChatNotification) -> AlertsApiResult<ChatReceipt> {
+        let mut message = ChatMessage::new(notification.text);
+        if let Some(reply_to) = notification.reply_to {
+            message = message.reply_to(MessageId::ts(reply_to));
+        }
```
ThisIsMani: Let's not use mut please. We can directly construct if needed, we can remove that function which takes mut as well.

ThisIsMani:
Pushed `9bcd8b933a` — a defect found by re-reading the live output, not by a test.

`reason` is documented as a field a caller branches on, but four of the seven chat refusals reached the wire as `ChatErrorReason`'s `Display`, which is English prose:

```diff
-{"code":"HE_02","destination":"xyne_bad_channel","reason":"channel not found"}
+{"code":"HE_02","destination":"xyne_bad_channel","reason":"channel_not_found"}
```

The same field was already returning `msg_too_long` and `invalid_arguments` correctly for other refusals, so it was a mix of wire codes and sentences. Anything matching on it would have broken on `channel_not_found`, `not_in_channel`, `invalid_auth` and `token_revoked` — the four cases most worth handling.

`reason_code` is now deliberately separate from `Display`: one is a log line for a human, the other is a wire contract, and deriving one from the other means rewording an error message silently changes the API.

One honest limit, documented rather than papered over: the code is **not** always the exact bytes the provider sent, and cannot be, because `external_services` folds synonyms on the way in — `is_archived` arrives as `NotInChannel`, `account_inactive` as `TokenRevoked`. The guarantee is one condition, one code. The four places that described this as the provider's code "verbatim" are corrected.

Two new tests: every reason produces snake_case rather than prose, and the codes match the provider's spelling. 37 tests total, clippy clean, and re-verified against the live Xyne channel.

crates/alerts/src/core/notifier/chat.rs
ThisIsMani, line 1: Should the trait definition be in core... Core is for api logics... Let's move the trait definition and other stuff like conversions to somewhere else.

crates/alerts/src/types.rs
```diff
@@ -0,0 +1,169 @@
...
+    /// the request never carries a channel id or a credential of its own.
+    pub destination: String,
+
+    /// The message, in the markup the destination reads. Delivered unchanged.
+    pub text: String,
+
+    /// Post this as a reply in the thread of an earlier message, identified by the `message_id`
+    /// that message's [`ChatNotifyResponse`] returned.
+    ///
+    /// The R alerts service uses this to put a recovery notice under the alert it clears, so a
```
ThisIsMani: Please don't mention this anywhere. Remove from here as well.

crates/alerts/src/errors/types.rs
```diff
@@ -54,32 +82,56 @@ impl ApiError {
...
+    /// 400 — the request was wrong in a way its sender can fix.
+    ///
+    /// Covers a provider refusal that blames the request, not just a malformed body: a message
+    /// over the provider's size limit is the caller's problem, and answering 500 would tell them
+    /// to retry something that will never succeed.
+    BadRequest(ApiError),
     /// 401 — authentication failed.
     Unauthorized(ApiError),
+    /// 429 — the provider is rate limiting us.
+    TooManyRequests(ApiError),
```
ThisIsMani: Should we throw 429 if provider is rate limiting?

crates/alerts/src/errors/types.rs
```diff
@@ -54,32 +82,56 @@ impl ApiError {
...
+    /// rejected. A bad channel id or a revoked token is not something the caller can fix.
+    InternalServerError(ApiError),
+    /// 502 — the provider could not be reached, or answered something we could not read.
+    BadGateway(ApiError),
 }
 
 impl ApiErrorResponse {
     /// The payload of whichever variant this is.
-    fn get_internal_error(&self) -> &ApiError {
+    pub(crate) fn get_internal_error(&self) -> &ApiError {
```
ThisIsMani: When you say internal_error, it's sounds like 5xx, let's name this something else.

crates/alerts/src/errors/types.rs
```diff
@@ -54,32 +82,56 @@ impl ApiError {
...
         match self {
-            Self::InternalServerError(_) => ErrorType::AlertsError.as_str(),
-            Self::Unauthorized(_) => ErrorType::InvalidRequestError.as_str(),
+            Self::BadRequest(_) | Self::Unauthorized(_) => ErrorType::InvalidRequestError.as_str(),
+            Self::TooManyRequests(_) | Self::InternalServerError(_) | Self::BadGateway(_) => {
+                ErrorType::AlertsError.as_str()
+            }
         }
     }
 }
```
ThisIsMani, line 110: This is there in router?

crates/alerts/src/errors/types.rs
```diff
@@ -41,10 +67,12 @@ pub struct ApiError {
     pub error_identifier: u16,
     /// Human-readable description. Must never contain a secret or the contents of a request.
     pub error_message: String,
+    /// Machine-readable detail, flattened into the body.
+    pub extra: Option<Extra>,
```
ThisIsMani: I think there is a bit of confusion in error handling... Maybe it's better not to send error details if something is not our fault but the providers, if our fault is there, then message should cover it I guess. Let's discuss about this.

crates/alerts/src/routes/app.rs
```diff
@@ -24,13 +26,50 @@ impl Alerts {
...
+    /// `/email/notify` takes an email one. See [`crate::types`] for why the alternative — one
+    /// `/notify/{id}` over a tagged body — was rejected.
     pub fn server(state: AppState) -> Scope {
         web::scope("/alerts")
             .app_data(web::Data::new(state))
-            .service(web::resource("/ping").route(web::get().to(alerts::ping)))
+            .app_data(json_config())
+            .service(
+                web::scope("/chat")
+                    .service(web::resource("/notify").route(web::post().to(notify::chat))),
```
ThisIsMani: We are not taking id in the path?

crates/alerts/src/core/notifier/chat.rs
```diff
@@ -0,0 +1,325 @@
...
+            reason: "incompatible_reply_target".to_owned(),
+        },
+
+        // Nothing is known about delivery in any of these. `MissingMessageId` in particular means
+        // the message *was* delivered and retrying would post it twice, which is why it does not
+        // land anywhere a caller would read as retryable.
+        ChatError::RequestFailed
+        | ChatError::HttpStatus { .. }
+        | ChatError::UnreadableResponse
+        | ChatError::MissingMessageId => AlertsError::ProviderUnavailable { destination },
```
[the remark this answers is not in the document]
ThisIsMani: Fixed in `c74621ad88`, and the arm was worse than flagged — my own comment on it claimed it "does not land anywhere a caller would read as retryable", while 502 is exactly that.

It is now a delivery rather than any kind of error: `{"status": "delivered", "message_id": null}`. The alert went out and only the ability to thread under it was lost, so reporting a failure would have been the thing that caused duplicates. This falls out of the wider change in that commit, where a provider's answer is an outcome in the body rather than an HTTP status.

crates/alerts/src/core/notifier/chat.rs
ThisIsMani: Done in `c74621ad88`. `ChatMessage::reply(text, message_id)` sits beside `ChatMessage::new(text)`, and the `mut self` builder is gone, so the call site is a `match` with no mutation and no clone. Fields stay private.

The two existing call sites in `external_services` tests moved with it.

crates/alerts/src/core/notifier/chat.rs
ThisIsMani: Done in `c74621ad88`. `domain/notifier/` holds the traits, the notification and outcome types, and the conversions; `core/notifier.rs` keeps the per-request logic a handler calls.

```
routes/   deserialize, call core, serialize
core/     what one request does: resolve a destination, hand the message over
domain/   what delivering an alert is: traits and the types they exchange
```

`domain` holds no HTTP and `core` holds no traits.

crates/alerts/src/types.rs
ThisIsMani: Removed in `c74621ad88`, and I swept the rest of the crate rather than just this line — I had internal tracker ids in doc comments across `settings.rs`, `state.rs`, `config/alerts.toml` and the README, which is a private tracker being named in a public repo.

Two notes. The merged scaffold has some of the same references (`core/notifier.rs`, the old `/ping` handler), which I have not touched — say the word if you want those cleaned too. And the field itself is gone from this file: the destination moved to the path.

crates/alerts/src/errors/types.rs
ThisIsMani: Removed in `c74621ad88`. You were right, and it broke the rule I had written two files earlier: the caller is not at fault for the provider rate limiting us, so it was never a 4xx.

It does not become a 5xx either. It is now a refusal in a 200 — `{"status": "refused", "error_code": "rate_limited", "retry_after_seconds": 30}` — along with every other answer the provider gives. `Retry-After` goes with it, since that header has nowhere to sit on a 200.

crates/alerts/src/errors/types.rs
ThisIsMani: Renamed to `payload()` in `c74621ad88`.

Worth knowing it was copied verbatim from `api_models::errors::types:120`, so this diverges from the repo on purpose. It is `pub(crate)`, so the cost is cosmetic.

crates/alerts/src/errors/types.rs
ThisIsMani: Yes — `api_models::errors::types::ApiErrorResponse::error_type`, at `crates/api_models/src/errors/types.rs:156`. Same name and same shape, mirrored deliberately rather than invented. `get_internal_error` came from the same file at :120, which is why it carried a name that reads as a 5xx.

Kept as-is.

crates/alerts/src/errors/types.rs
ThisIsMani: Settled by deleting the mechanism rather than writing a rule for it. `Extra` and `reason` are gone in `c74621ad88`.

Your instinct was the right one and went further than I first took it. Once every answer the provider gives is a 200 outcome, there is no error left that has provider detail to withhold: errors carry a code and a message, and all provider vocabulary lives in the success body where a caller can act on it. `destination` is gone too — it is in the path now, so there is nothing to echo.

What is left is short enough to read at a glance: 400 body did not parse, 401 bad key, 404 unknown destination, 502 provider unreachable, 500 we failed.

crates/alerts/src/routes/app.rs
ThisIsMani: Changed in `c74621ad88` — `POST /alerts/chat/notify/{destination}`, and the body is content only.

Better than what I had, for a reason I had not weighed: the destination now reaches access logs, metrics labels and tracing spans without anyone parsing a body, so "which destination is failing" is answerable from the ops view. Unknown destination becomes a 404, which reads correctly for a path segment.

No conflict with rejecting `/notify/{id}` earlier — the channel is still in the path too, so there is exactly one authority on both facts.

ThisIsMani:
Pushed `c74621ad88`. All nine threads answered inline; the two design ones changed the shape of the PR, so here is the summary.

## The status code and the body now answer different questions

**HTTP says whether the notifier worked. `status` says whether the message arrived.**

A provider that refuses — for any reason, including `channel_not_found` and `token_revoked` — is a `200` carrying `status: "refused"`. It was reached, it answered, and this service did its job.

```diff
-500 {"error":{"code":"HE_02","destination":"xyne_bad_channel","reason":"channel_not_found"}}
+200 {"status":"refused","error_code":"channel_not_found"}
```

The decisive argument was the second one in review: fault is the wrong axis. Whether a bad channel id is our mistake or a merchant's depends on who owns the destination, and that moves from a config file to a database row without a status code being able to move with it. Encoding fault in HTTP would have baked in an answer that needs a breaking change to undo.

So the line is drawn by **what the provider said**, which is the line payments already draws between a connector declining a transaction and a connector being unreachable. `5xx` now means this service is broken and nothing else, which is what makes alerting on it worth anything — a config typo no longer pages someone.

Three things fall out:

- **`Extra` and `reason` are gone.** All provider vocabulary lives in the 200 body where a caller can act on it. That settles the error-detail thread by deleting the mechanism rather than writing a rule for it.
- **No 429.** A caller is not at fault for the provider rate limiting us, so it was never a 4xx. It is a refusal carrying `retry_after_seconds`, and `Retry-After` goes with it since the header has nowhere to sit on a 200.
- **`MissingMessageId` is a delivery.** `{"status":"delivered","message_id":null}` — the alert went out and only threading was lost. As a retryable 502 it would have posted duplicates. Raised by Codex, and my own comment on that arm asserted the opposite of what the code did.

`status` is required on every success response, so a caller cannot deserialize one without confronting whether the message arrived. Same trick `external_services` uses on the provider's own `ok` field, and for the same reason: the failure mode of this shape is a caller that sees `200` and stops looking.

## Destination in the path

`POST /alerts/chat/notify/{destination}`, body is content only. It reaches access logs, metrics labels and tracing spans without anyone parsing a body. Unknown destination is now `404`.

## Layout

```
routes/   deserialize, call core, serialize
core/     what one request does: resolve a destination, hand the message over
domain/   what delivering an alert is: traits and the types they exchange
```

## Verified

40 tests, clippy clean, and re-run end to end against the live Xyne test channel:

```
post                     200  {"status":"delivered","message_id":"cmtkasg8407u912l4703nfxrb"}
threaded reply           200  {"status":"delivered","message_id":"cmtkasgs507vx12l41h1c5ie9"}
bad channel (was 500)    200  {"status":"refused","error_code":"channel_not_found"}
unknown destination      404  {"error":{...,"code":"IR_02"}}
email                    200  {"status":"delivered"}
no api key               401  {"error":{...,"code":"IR_01"}}
```

One unrelated thing found while running the suite: `external_services`' `utils::tests::test_deserialize_hashset_inner_empty_string` fails on `observability-plane-i` before any of my changes. Not touched here.

## Two things left for you

**Silent truncation.** `slack_compatible.rs` truncates unconditionally at `max_message_chars`, so 300 characters into a destination capped at 100 returns `200 delivered` and what lands is 100 characters ending `…(truncated)`. I demonstrated it against the live channel. That means "too many characters" almost never produces `msg_too_long` at all — the provider only sees it in the narrow window where the text is under our cap but over theirs after markup rendering. Nothing in the current design can even detect it, since `post_message` returns only a `MessageId`. Worth its own ticket rather than a fix here.

**The scaffold's internal references.** I swept my own doc comments, but the merged scaffold has internal tracker ids in `core/notifier.rs` and the old `/ping` handler. Left alone — tell me if you want those cleaned too.

crates/alerts/src/types.rs
```diff
@@ -0,0 +1,287 @@
...
+
+impl fmt::Debug for EmailNotifyRequest {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("EmailNotifyRequest")
+            .field("subject_chars", &self.subject.chars().count())
+            .field("body_chars", &self.body.chars().count())
+            .finish()
+    }
+}
+
```
ThisIsMani: Can we put them as secret<string> and let debug impl do the job instead of manually doing this?

ThisIsMani: Done in `ae8749ab33`. `text`, `subject` and `body` are `Secret<String>` and both request types derive `Debug` again; the two hand-written impls are gone.

Better than what I had, and not only shorter: a hand-written `Debug` holds until someone adds a field and forgets, whereas the type cannot forget. I carried it through the domain types too, so a log line added anywhere between the handler and the transport redacts by default rather than by review.

Nothing diagnostic is lost — the sizes those impls printed are already logged where they are useful, `slack_compatible.rs` emitting `chars` per request.

PR #14011, opened 2026-09-02
feat(alerts): deliver email over the external_services backends

ThisIsMani:
## Type of Change

- [x] New feature

**Stacked on [#14010](https://github.com/juspay/hyperswitch/pull/14010)** — base is `23116-notifier-interface`, not `observability-plane-i`. Review that one first; this diff is only the email transport.

## Description

`/alerts/email/notify/{destination}` now sends. #14010 left it wired to a log stand-in because reaching `external_services::email` needed a decision; this makes it.

`external_services::email` is behind a feature flag `alerts` did not enable (`external_services/Cargo.toml:12`, `lib.rs:11`), so `EmailService` was not even in scope before this. That is the first change.

### Reaching that module

Its object-safe trait exposes one method, and it is built for templated product emails:

```rust
compose_and_send_email(base_url: &str, email_data: Box<dyn EmailData>, proxy_url: Option<&String>)
```

`EmailData` is the composition hook, where an implementation renders a template against a base URL. This crate has a subject and a body already in hand, no template, and no meaningful base URL. Three ways across, and the choice is less free than it looks:

| | Cost |
|---|---|
| **Implement `EmailData`, ignore `base_url`** | A dummy argument. Stays inside this crate. |
| Add a plain-send method to `EmailService` | Widens a trait `router` depends on, to suit one caller |
| Depend on the concrete backends | Loses the SES / SMTP / no-email selection, and with it the config-driven off switch |

Only the first keeps its cost local, so that is what `AlertEmail` does. The smell is one ignored parameter in one function.

### One client, many addresses

Unlike chat — where a destination *is* an endpoint with its own credential — email has a single transport. The client is built once and shared across destinations, rather than a connection pool per recipient. `[email]` reuses `external_services`' own `EmailSettings` wholesale rather than mirroring it, so the backend selection and its validation cannot drift from the router's.

### `NO_EMAIL_CLIENT` is the off switch

It accepts and logs, it is the default, and it is what a deployment runs with before SES credentials exist. So there is no separate "email enabled" flag, and `LogEmailNotifier` is **deleted** rather than duplicating a backend that already does the job.

### Validation follows use

The transport is validated at boot only when destinations exist, so a service with none does not need a verified sender to start. A destination with no address *does* fail the boot, because it would otherwise accept alerts and send them nowhere.

Both matter more than they look: `create_email_client` logs and falls back rather than returning an error, so without a boot check a bad SES configuration becomes a client that silently never sends — the worst possible failure mode for an alerting pipe.

### One thing worth knowing

**`EmailError` has no refusal vocabulary.** A rejected recipient, a throttle and an unverified SES sender all arrive as `EmailSendingFailure`. So this channel only ever reports `delivered` or fails — `status: "refused"` is reachable for chat and not yet for email. The response shape stays uniform; email simply does not use half of it until that enum grows.

## How did you test it?

45 tests, clippy clean. The integration tests now build the **real** notifier over `NoEmailClient` rather than a stand-in, so the `EmailData` composition path is exercised without credentials.

```
cargo test -p alerts --features v1                    # 45 passed
cargo clippy -p alerts --features v1 --all-targets    # clean
```

Against a running service with `active_email_client = "NO_EMAIL_CLIENT"`:

```
email -> configured destination   200  {"status":"delivered"}
email -> unknown destination      404  {"error":{...,"code":"IR_02"}}
```

No subject or body reached the logs. The tracing span carries the destination, which is the payoff of moving it into the path in #14010:

```
http.route:  /alerts/email/notify/{destination}
http.target: /alerts/email/notify/oncall
```

## Checklist

- [x] `cargo clippy --all-features --all-targets -- -D warnings` (for `alerts`)
- [x] Added tests
- [x] Added documentation — `crates/alerts/README.md` and `config/alerts.toml`

### DIFF
--- a/crates/alerts/src/domain/notifier/email.rs
+++ b/crates/alerts/src/domain/notifier/email.rs
@@ -63,7 +63,7 @@ pub type EmailOutcome = Outcome<()>;
 /// both backends build a single-recipient message. When that is lifted, a destination widens to a
 /// recipient list without the wire contract changing, since a request only ever names an id.
 #[async_trait::async_trait]
-pub trait EmailNotifier: Send + Sync + std::fmt::Debug {
+pub trait EmailNotifier: Send + Sync {
     /// Attempt delivery.
     ///
     /// A provider that refuses returns `Ok(Outcome::Refused)`, not an error. Email cannot reach
@@ -97,17 +97,6 @@ impl EmailServiceNotifier {
             recipient,
             proxy_url,
         }
-    }
-}
-
-// Written out rather than derived, because `Box<dyn EmailService>` is not `Debug`. It names only
-// the destination: the recipient masks itself, but a field nobody prints cannot be peeked into a
-// log line by accident either.
-impl std::fmt::Debug for EmailServiceNotifier {
-    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        f.debug_struct("EmailServiceNotifier")
-            .field("destination", &self.destination)
-            .finish_non_exhaustive()
     }
 }
 
@@ -210,16 +199,39 @@ mod tests {
         assert_eq!(outcome, Outcome::Delivered(()));
     }
 
-    /// A subject carries merchant ids and a body carries volumes. Neither belongs in a log line,
-    /// and nor does the recipient.
-    #[tokio::test]
-    async fn debug_leaks_neither_the_content_nor_the_recipient() {
-        let rendered = format!("{:?} {:?}", no_email_notifier().await, notification());
+    /// A subject carries merchant ids and a body carries volumes, and neither belongs in a log
+    /// line. `Secret` is what enforces that, rather than a hand-written `Debug` somebody has to
+    /// remember to update.
+    #[test]
+    fn debug_leaks_no_content() {
+        let rendered = format!("{:?}", notification());
 
-        assert!(rendered.contains("oncall"));
-        assert!(!rendered.contains("example.com"));
         assert!(!rendered.contains("merchant_1234"));
         assert!(!rendered.contains("4,201"));
+    }
+
+    /// `compose_and_send_email` uses `base_url` for exactly one thing — passing it to
+    /// `get_email_data` — so the empty string we hand it reaches nothing but this impl, which
+    /// drops it. Pinned rather than argued: if a future `AlertEmail` ever starts reading it, this
+    /// fails instead of quietly composing a different email.
+    #[tokio::test]
+    async fn base_url_cannot_affect_the_composed_email() {
+        let compose = |base_url: &'static str| async move {
+            AlertEmail {
+                subject: "subject".to_owned().into(),
+                body: "<pre>body</pre>".to_owned().into(),
+                recipient: recipient(),
+            }
+            .get_email_data(base_url)
+            .await
+            .unwrap()
+        };
+
+        let empty = compose("").await;
+        let populated = compose("https://example.com").await;
+
+        assert_eq!(empty.subject, populated.subject);
+        assert_eq!(empty.body.into_inner(), populated.body.into_inner());
     }
 
     #[tokio::test]
--- a/crates/alerts/src/domain/notifier/email.rs
+++ b/crates/alerts/src/domain/notifier/email.rs
@@ -1,20 +1,42 @@
 //! Delivering an alert to an email destination.
 //!
-//! **Only [`LogEmailNotifier`] exists today, and that is deliberate.** A real implementation means
-//! choosing how to reach `external_services::email`, whose object-safe trait exposes one method
-//! built for templated product emails: `compose_and_send_email(base_url, Box<dyn EmailData>,
-//! proxy_url)`. This crate has a subject and a body already in hand, no template and no meaningful
-//! base URL. That choice belongs to the ticket that owns it, which asks for the reasoning in its
-//! pull request. This one fixes the contract; that one fills it in.
+//! ## Reaching `external_services::email`
 //!
-//! Chat is not in the same position, which is why it has a real implementation:
-//! `ChatClient::post_message` maps onto [`ChatNotifier`](super::chat::ChatNotifier) one to one,
-//! with nothing to decide.
+//! That module's object-safe trait exposes exactly one method, and it is built for templated
+//! product emails: `compose_and_send_email(base_url, Box<dyn EmailData>, proxy_url)`. `EmailData`
+//! is the composition hook, where an implementation renders a template against a base URL. This
+//! crate has a subject and a body already in hand, no template, and no meaningful base URL.
+//!
+//! Three ways to bridge that were open, and the choice is less free than it looks:
+//!
+//! - **Implement `EmailData` and ignore `base_url`.** What [`AlertEmail`] does. Nothing outside
+//!   this crate changes.
+//! - **Add a plain-send method to `EmailService`.** Honest, and smaller here — but it widens a
+//!   trait `router` depends on, to suit one caller.
+//! - **Depend on the concrete backends.** Loses the SES / SMTP / no-email selection that
+//!   `EmailClientConfigs` gives for free, and with it the ability to turn email off by config.
+//!
+//! The first is the only one whose cost stays inside this crate, so a dummy `base_url` is the
+//! price. It is one ignored parameter in one function, and it buys not touching a shared trait.
+//!
+//! ## Email cannot refuse
+//!
+//! `EmailError` has no refusal vocabulary: a rejected recipient, a throttle and an unverified SES
+//! sender all arrive as `EmailSendingFailure`. So this channel only ever produces
+//! [`Outcome::Delivered`] or an error, and `status: "refused"` is unreachable for it until that
+//! enum grows. The response shape stays uniform across channels; email simply never uses half of
+//! it yet.
 
+use std::sync::Arc;
+
+use common_utils::{errors::CustomResult, pii};
+use external_services::email::{
+    EmailContents, EmailData, EmailError, EmailService, IntermediateString,
+};
 use hyperswitch_masking::{PeekInterface, Secret};
 
 use super::Outcome;
-use crate::{errors::AlertsApiResult, logger};
+use crate::errors::{AlertsApiResult, AlertsError};
 
 /// A message to send to one email destination.
 #[derive(Debug, Clone)]
@@ -22,8 +44,9 @@ pub struct EmailNotification {
     /// The subject line, delivered unchanged.
     pub subject: Secret<String>,
 
-    /// The body, as HTML. Both backends in `external_services::email` hardcode an HTML body, so
-    /// there is nothing else to send until that module grows a plain-text path.
+    /// The body, as HTML. Both backends hardcode an HTML body — `email/ses.rs` builds it with
+    /// `.html(...)`, `email/smtp.rs` sets `ContentType::TEXT_HTML` — so there is nothing else to
+    /// send until that module grows a plain-text path.
     pub body: Secret<String>,
 }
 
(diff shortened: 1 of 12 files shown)


crates/alerts/src/domain/notifier/email.rs
```diff
@@ -43,62 +66,182 @@ pub type EmailOutcome = Outcome<()>;
...
-    pub fn new(destination: String) -> Self {
-        Self { destination }
+// Written out rather than derived, because `Box<dyn EmailService>` is not `Debug`. It names only
+// the destination: the recipient masks itself, but a field nobody prints cannot be peeked into a
+// log line by accident either.
+impl std::fmt::Debug for EmailServiceNotifier {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("EmailServiceNotifier")
+            .field("destination", &self.destination)
+            .finish_non_exhaustive()
```
ThisIsMani: It's fine I guess. We can remove this.

crates/alerts/src/domain/notifier/email.rs
```diff
@@ -43,62 +66,182 @@ pub type EmailOutcome = Outcome<()>;
...
-            destination = %self.destination,
-            subject_chars = notification.subject.peek().chars().count(),
-            body_chars = notification.body.peek().chars().count(),
-            "not delivered: no email transport is wired yet"
-        );
+        self.client
+            .compose_and_send_email(
+                // No template renders against this, so there is no URL to give. `EmailData` takes
+                // one because the router's product emails build links with it.
+                "",
```
ThisIsMani: This won't break anything? I mean passing empty string?

crates/alerts/src/domain/notifier/email.rs
ThisIsMani: Removed in `d3f4962883`, along with the reason it existed: the `Debug` supertrait is off both `ChatNotifier` and `EmailNotifier` now.

The impl was only there because `EmailNotifier: Debug` forced it and `Box<dyn EmailService>` is not `Debug`. Nothing formats a notifier or a registry, so a bound that cannot be met for free and is never used was not earning its place. Dropped from chat too rather than leaving the two traits asymmetric.

One knock-on: `Registry` is no longer `Debug` either, since its derive could never apply to the only two instantiations that exist.

crates/alerts/src/domain/notifier/email.rs
ThisIsMani: No, and it is safe by construction rather than by luck.

`compose_and_send_email` uses `base_url` for exactly one thing (`email.rs:68`):

```rust
let email_data = email_data.get_email_data(base_url);
```

That is its only mention. `email_data` here is our own `AlertEmail`, whose `get_email_data` ignores the argument, so the empty string reaches nothing but a function that drops it. It never touches the transport, the message, the headers or the recipient.

The parameter exists because the router's product emails — `VerifyEmail`, `ResetPassword`, `InviteUser` — build links against it. There are no links in an alert.

I have pinned it rather than leaving it as an argument (`d3f4962883`): a test composes the same alert with `""` and with `"https://example.com"` and asserts the results are identical. If a future `AlertEmail` ever starts reading `base_url`, that fails instead of quietly composing a different email.

crates/alerts/src/domain/notifier/email.rs
```diff
@@ -43,62 +66,182 @@ pub type EmailOutcome = Outcome<()>;
...
     async fn notify(&self, notification: EmailNotification) -> AlertsApiResult<EmailOutcome> {
-        logger::info!(
-            tag = "email_notify_skipped",
-            destination = %self.destination,
-            subject_chars = notification.subject.peek().chars().count(),
-            body_chars = notification.body.peek().chars().count(),
-            "not delivered: no email transport is wired yet"
-        );
+        self.client
+            .compose_and_send_email(
```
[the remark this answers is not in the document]
ThisIsMani: Confirmed and fixed in `ccc04fa697`, in its own commit since it changes shared code.

`lettre::SmtpTransport` is the blocking transport and `email_client.send(&email)` has no `.await`, so it parks the worker until the relay answers or the timeout expires. Latent in the router, which runs many workers and does not send mail on a hot path — not latent here, where `server.workers = 1` ships by default and a slow relay would take the health check down with it.

Moved to `spawn_blocking`. No signature change; `tokio` gains the `rt` feature. Verified `cargo check -p router --features v1,olap,oltp,email` still passes.

crates/alerts/src/settings.rs
```diff
@@ -146,9 +156,36 @@ impl ChatSettings {
...
-        validate_destination_ids(&self.destinations, "email")
+        validate_destination_ids(&self.destinations, "email")?;
+
+        if self.destinations.is_empty() {
+            return Ok(());
+        }
+
+        self.client
+            .validate()
+            .map_err(|error| errors::ConfigurationError::ConfigParsingError(error.to_owned()))?;
```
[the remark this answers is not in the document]
ThisIsMani: Confirmed and fixed in `d3f4962883`.

`EmailSettings::validate` only calls the backend-specific check — `SESConfig::validate` covers `email_role_arn` and `sts_role_session_name`, `SmtpServerConfig::validate` covers the host — and never touches the shared sender. It defaults to an empty `pii::Email`, so SMTP fails building the `From` mailbox and SES submits an empty sender, exactly as described.

Now rejected at boot whenever a real transport is configured, reusing the same emptiness check as recipients. Left unchecked under `NO_EMAIL_CLIENT`, which sends nothing and needs no sender.

crates/alerts/src/state.rs
```diff
@@ -129,21 +135,62 @@ fn build_chat_registry(
...
+/// that crate. Copied rather than shared because lifting it into `external_services` would mean
+/// moving `Proxy` handling with it, and the function is a three-arm match.
+async fn create_email_client(
+    settings: &EmailClientSettings,
+    proxy: &Proxy,
+) -> Box<dyn EmailService> {
+    match &settings.client_config {
+        EmailClientConfigs::Ses { aws_ses } => {
+            Box::new(AwsSes::create(settings, aws_ses, proxy.https_url.clone()).await)
+        }
```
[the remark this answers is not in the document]
ThisIsMani: Confirmed and fixed in `d3f4962883`. This one is worse than it looks, because my own doc comment claimed the boot check caught it.

`AwsSes::create` does `Self::create_client(..).await.map_err(|e| logger::error!(..)).ok();` and then returns `Self` built from config — the client it just made is discarded entirely. So a wrong role ARN logs once at boot and every alert 502s forever, and `EmailSettings::validate` cannot help: it is a static check, and the ARN is well-formed but wrong.

`create_email_client` now probes the fallible `create_client` path first and fails startup. Two consequences named in the code rather than hidden: one extra `AssumeRole` at boot, and a transient AWS outage now prevents startup. I took that trade deliberately — for a service whose only job is delivery, refusing to start with a clear error beats running while silently dropping every alert — but it is a trade, so flagging it.

crates/alerts/src/domain/notifier/email.rs
ThisIsMani: Reworked in `cb4fd29d0b` + `c32d897dba` — the empty string is gone rather than justified.

You were right to push back, and my earlier framing of the alternative was wrong. I had said adding a method to `EmailService` "widens a trait `router` depends on", implying a cost. There is none: the trait has exactly one implementor, the blanket `impl<T> EmailService for T where T: EmailClient`, so SES, SMTP and `no_email` pick it up for free and all fourteen `compose_and_send_email` call sites in `router` are untouched. Confirmed with `cargo check -p router --features v1,olap,oltp,email`.

`EmailService` now has `send_contents(EmailContents, proxy_url)`, and composition is defined in terms of it, so the two paths cannot drift:

```rust
async fn compose_and_send_email(&self, base_url, email_data, proxy_url) -> EmailResult<()> {
    let email_data = email_data.get_email_data(base_url).await?;
    self.send_contents(email_data, proxy_url).await
}
```

The caller here becomes a direct `send_contents` call. `AlertEmail`, its `EmailData` impl, the `""`, and the two tests that existed only to prove the dummy argument was harmless are all deleted.

The honest reading is the one your comment pointed at: the trait was missing an operation, not a cleverer caller.

crates/alerts/src/domain/notifier/email.rs
ThisIsMani: Understood, and reverted in `4c15932444` + `c72eb490b7` — `Debug` is back on both notifier traits, derived.

The reason it could not be derived before is the one that needed fixing: `EmailServiceNotifier` holds `Arc<Box<dyn EmailService>>`, and a derive requires every field to be `Debug`. So `EmailService` now requires it, matching `chat_service::ChatClient` which already does. That costs nothing — `AwsSes`, `SmtpServer` and `NoEmailClient` all already derive `Debug`, and the trait's only impl is the blanket one — and `cargo check -p router --features v1,olap,oltp,email` still passes.

One behavioural difference worth knowing, since the derive prints a field my impl skipped: the recipient now renders as `******@example.com`. `pii::Email` masks its own local part and keeps the domain, which is what every other email in this workspace logs as, so it is consistent rather than a new exposure. The test asserts that shape rather than asserting the field is absent.

`Registry` gets its derive back too.

crates/external_services/src/email.rs
```diff
@@ -42,21 +42,41 @@ pub trait EmailClient: Sync + Send + dyn_clone::DynClone {
 }
 
 /// A super trait which is automatically implemented for all EmailClients
+///
+/// `Debug` so that a caller holding one as a field can derive `Debug` rather than hand-write an
+/// impl to skip it. Matches [`crate::chat_service::ChatClient`], and every implementor already
+/// derives it.
 #[async_trait::async_trait]
-pub trait EmailService: Sync + Send + dyn_clone::DynClone {
+pub trait EmailService: Sync + Send + std::fmt::Debug + dyn_clone::DynClone {
```
ThisIsMani: Can we remove Debug from here.

crates/external_services/src/email.rs
```diff
@@ -42,21 +42,41 @@ pub trait EmailClient: Sync + Send + dyn_clone::DynClone {
...
+
+    /// Send contents that are already composed.
+    ///
+    /// For a caller that holds a subject, a body and a recipient and has no template to render.
+    /// [`EmailService::compose_and_send_email`] exists to *run* an [`EmailData`], and its
+    /// `base_url` is there because the product emails build links against it; a caller with
+    /// nothing to compose would otherwise have to implement [`EmailData`] purely to hand back what
+    /// it was given, and invent a `base_url` for that impl to ignore.
+    ///
+    /// Composition is defined in terms of this rather than the other way round, so the two paths
```
ThisIsMani: Reduce this comment.

ThisIsMani: Trimmed in `d93a6275cf`:

```rust
/// Send contents that are already composed, for a caller with no template to render.
///
/// [`EmailService::compose_and_send_email`] is defined in terms of this, so the two cannot drift.
```

The reasoning about `base_url` moved to the `alerts` side, where it is context for a decision rather than API documentation.

crates/external_services/src/email.rs
ThisIsMani: Removed in `d93a6275cf`, and `EmailNotifier` loses its `Debug` bound rather than gaining a hand-written impl.

Worth saying plainly that the three asks across this thread turned out to be mutually exclusive, so something had to give:

1. remove the hand-written `Debug` impl
2. derive `Debug` instead
3. do not put `Debug` on `EmailService`

`EmailServiceNotifier` holds `Arc<Box<dyn EmailService>>`, and a derive requires **every** field to be `Debug`. So (2) is only possible via (3), which leaves three real options: derive and widen the shared trait, hand-write the impl, or drop the bound.

I took the third, since it is the only one that asks nothing of `external_services`. It costs nothing either — nothing formats a notifier or a registry, so the bound was paying for itself with an impl and buying nothing.

`ChatNotifier` keeps `Debug`, because every chat implementation derives it for free and no shared trait has to change for that to be true. The asymmetry is not an oversight: it is exactly the difference between a trait whose implementations happen to be `Debug` and one whose implementations hold something that is not.

Net footprint on `external_services` is now the `send_contents` method, plus the SMTP `spawn_blocking` fix which you may well want out of this PR anyway.

PR #14017, opened 2026-09-03
fix(payments): durable retry for save-payment-method fire-and-forget failures

### LINKED ISSUE #12904: [BUG] Payment method tokenization can leave vault and payment_attempt in inconsistent state due to fire-and-forget spawn
## Summary

In `crates/router/src/core/payments/operations/payment_response.rs:689`, the save-payment-method flow is dispatched via `tokio::spawn` with the handle dropped. While the async dispatch is correct for latency reasons, there are **two independent silent-failure modes** that can each produce a permanent inconsistency with no retry or detection mechanism.

**Failure mode 1 — outer locker save (line 712):**
```rust
if let Err(err) = result {
    logger::error!("Asynchronously saving card in locker failed : {:?}", err);
}
```
If `save_payment_method` itself fails (locker down, network timeout, transient error), the user's card is never vaulted, but the payment flow has already returned success to the caller. No metric is emitted, no retry is attempted, no user notification is sent.

**Failure mode 2 — DB update after successful locker save (line 753–755):**
```rust
if let Err(err) = respond {
    logger::error!("Error updating payment attempt: {:?}", err);
}
```
If `update_payment_attempt_with_attempt_id` fails after `save_payment_method` succeeds, the vault now holds a saved payment method, but `payment_attempt.payment_method_id` is `NULL`. This is a permanent vault↔DB split with no detection or recovery path.

Both failures are the same class of bug — fire-and-forget with log-only error handling — and deserve to be fixed together.

## Reproduction sketch

1. Trigger a successful authorize flow that hits the "Save card" branch (line 676)
2. **For mode 1:** Inject a locker-level failure (network drop, locker restart) during `save_payment_method`
3. **For mode 2:** Inject a DB-side failure in `update_payment_attempt_with_attempt_id` after the locker save succeeds (e.g., connection drop)
4. Observe: log line emitted at line 712 or 754, but no retry; state is permanently inconsistent

## Proposed fix

Move the work into a `ProcessTracker`-managed workflow following the existing pattern (e.g., `DeleteTokenizeDataWorkflow` in `crates/router/src/workflows/tokenized_data.rs`):

- New workflow: `crates/router/src/workflows/save_payment_method.rs` implementing `ProcessTrackerWorkflow<SessionState>`
- Replace `tokio::spawn` at line 689 with a `process_tracker` enqueue
- Drainer picks up the workflow; its `error_handler` routes failures to `business_status = REVIEW` for operator visibility
- On retry, the locker call is naturally idempotent: `StoreCardReq` already carries `requestor_card_reference`, and `add_card_hs` returns a `DataDuplicationCheck::Duplicated` response when the card already exists — so retry #2 after a partial failure will detect the existing vault entry rather than creating a duplicate

This preserves the latency-isolation design while adding durability and observability.

## Implementation notes

- **Serialization scope**: `VaultOperation` derives `serde::Serialize` but not `serde::Deserialize` (see `crates/hyperswitch_domain_models/src/payments.rs:1667`). Since the `ProcessTracker` workflow body must be round-trippable from a DB row, either `Deserialize` needs to be derived on the captured types, or the workflow should store only IDs and re-fetch the full objects inside the workflow runner. This may widen the fix scope modestly.
- **Idempotency at locker layer**: As noted above, `requestor_card_reference` already provides the hook. The workflow should pass the attempt ID (or a stable composite key) as `card_reference` to ensure the locker deduplicates on retry.

Happy to implement if the approach is acceptable — would value maintainer input on whether `SavePaymentMethodWorkflow` should live alongside `tokenized_data.rs` or as a sibling file, and on the preferred strategy for the serialization gap.

## Acceptance criteria

- [ ] New `SavePaymentMethodWorkflow` implementing `ProcessTrackerWorkflow<SessionState>`
- [ ] `tokio::spawn` at line 689 replaced with `insert_process` call (covers both failure modes)
- [ ] Locker call passes stable idempotency key via `requestor_card_reference`
- [ ] Integration test: simulate `update_payment_attempt` failure → verify `process_tracker` retries → vault and DB consistent
- [ ] `cargo +nightly fmt --all` + `cargo test -p router` passing

unspecifiedcoder:
## Description

Fixes both silent-failure modes described in #12904 in the async save-payment-method flow (`payment_response.rs`'s `save_pm_and_mandate`, "Save card" branch):

1. **Locker save failure** - a transient locker error (network blip, locker restart) was logged once and dropped, permanently skipping the vault save with no retry. Now retried up to 3 times in-process before giving up (bounded, in-memory only, so no new persistence/PCI surface).
2. **Post-save DB update failure** - if `payment_attempt.payment_method_id` failed to persist after the locker save had already succeeded, the vault and DB were left permanently inconsistent with only a log line. Now enqueues a new `SavePaymentMethodAttemptUpdateWorkflow` via `ProcessTracker` for a durable, retried write, following the same pattern as `PaymentMethodStatusUpdateWorkflow`.

The new workflow's tracking data holds only IDs (attempt/payment/merchant id, the already-resolved `payment_method_id`, `updated_by`) - by the time it runs the vault entry already exists, so nothing sensitive needs to be persisted to `process_tracker`.

Also adds three metrics to make both failure paths observable:
- `SAVE_PAYMENT_METHOD_LOCKER_SAVE_FAILURE`
- `SAVE_PAYMENT_METHOD_ATTEMPT_UPDATE_ENQUEUED`
- `SAVE_PAYMENT_METHOD_ATTEMPT_UPDATE_ENQUEUE_FAILURE`

## Additional Changes

- `SavePaymentMethodData` now derives `Clone` (needed to retry the locker call with the same request data).

## Motivation and Context

Closes #12904.

## How did you test it?

- `cargo check -p router --features v1` (locally resource-constrained, re-verifying in CI)
- Unit tests added in `workflows/save_payment_method_attempt_update.rs` covering tracking-data (de)serialization round-trip and the retry-mapping defaults.

## Checklist

- [ ] I added tests for my changes
- [ ] I updated the documentation
- [ ] I added a changelog entry if applicable


XyneSpaces:
[blocking] `SavePaymentMethodAttemptUpdateWorkflow::execute_workflow` for the `v2` feature contains a bare `todo!()` at `crates/router/src/workflows/save_payment_method_attempt_update.rs:127`. It will panic at runtime when v2 is enabled and a save-payment-method attempt-update task is picked up.

Return `Err(ProcessTrackerError::UnexpectedFlow)` (or implement the v2 retry path) instead of `todo!()`.

XyneSpaces:
[blocking] `crates/router/src/core/payments/operations/payment_response.rs:93-96` defines `SAVE_PAYMENT_METHOD_LOCKER_MAX_ATTEMPTS` and `SAVE_PAYMENT_METHOD_LOCKER_RETRY_DELAY` under `#[cfg(any(feature = "v1", all(test, feature = "deja")))]`, but the retry loop at `:~957` uses them unconditionally. A build with only `feature = "v2"` enabled will fail with undeclared constants. Either gate the loop behind the same `cfg`, or move the constants outside the conditional block.

unspecifiedcoder:
@XyneSpaces Addressed in 96b4bcd:

- `SavePaymentMethodAttemptUpdateWorkflow::execute_workflow` (v2) now returns `Err(ProcessTrackerError::UnexpectedFlow)` instead of `todo!()`. The flow that enqueues this task (`save_pm_and_mandate`) is v1-only, so a v2 scheduler should never pick one up; failing explicitly is the right behaviour rather than a panic.
- `SAVE_PAYMENT_METHOD_LOCKER_MAX_ATTEMPTS` / `SAVE_PAYMENT_METHOD_LOCKER_RETRY_DELAY` are now gated on `#[cfg(feature = "v1")]` only. Note the retry loop lives inside the `#[cfg(feature = "v1")] impl PostUpdateTracker<_, _, PaymentsAuthorizeData>` block, so a v2-only build did not actually reference them, but the `all(test, feature = "deja")` gate was wrong (unused constants under that config), and matching the enclosing impl's gate is clearer.

Two small follow-ups in the same commit:
- The ProcessTracker retry is only enqueued when the locker actually returned a `payment_method_id`; with `None` there is nothing to reconcile.
- Dropped an unneeded `#[allow(clippy::too_many_arguments)]` (the helper takes 6 args).

unspecifiedcoder:
@claude please review this PR. Focus on: idempotency of retrying `save_payment_method` on partial failure (locker written, DB insert failed), the `SavePaymentMethodAttemptUpdateWorkflow` retry mapping and early-exit logic, and v1/v2 feature gating.

unspecifiedcoder:
Pushed a follow-up hardening commit (`cbd3e15`) after a closer pass over the retry and ProcessTracker paths.

### Locker retry

* Internal locker: retries up to 3 times, since duplicate saves are handled by the locker.
* External vault: single attempt only, since retrying can create duplicate vault tokens.
* Retry failures are now logged once after the final attempt instead of dumping the full error on every attempt.

### ProcessTracker

* Process IDs now include `merchant_id` via `pt_utils::get_process_tracker_id`.
* Tasks use the configured `application_source`.
* Retry tasks are only enqueued when the locker returned a `payment_method_id`.

### Workflow hardening

* Key-store, merchant and payment-attempt lookups are now inside the retried path.
* Missing attempts are treated as a terminal `ATTEMPT_NOT_FOUND` case rather than retried.
* The lookup now uses the unique payment-attempt index.
* Escaped errors go through `consumer_error_handler` instead of leaving tasks stuck.
* Added metrics for retries exhausted and attempt-not-found cases.

Also removed the dead v2 workflow path and replaced the retry-mapping test with a test of the actual retry schedule.

Verified with:

* `cargo check -p router`
* `cargo check -p router --no-default-features --features "release,v2,redis-rs"`

unspecifiedcoder:
@XyneSpaces

PR #14026, opened 2026-09-03
feat(router): generate Apple Pay PPC keypair internally, keep merchant out of PCI scope

ShankarSinghC:
## Summary
- Adds `POST /account/{merchant_id}/connectors/{merchant_connector_id}/applepay/certificate/generate`, which generates the Apple Pay payment-processing-certificate (PPC) EC keypair server-side and returns only the CSR (as a raw downloadable file) for the merchant to submit to Apple — the private key never leaves Hyperswitch.
- The generated key is stored in a new sibling field, `system_generated_payment_processing_certificate_key`, on `PaymentProcessingDetails`. A masked placeholder is written into the existing `payment_processing_certificate_key` field once, at generation time, so the dashboard form still shows a non-empty value — the raw key is never returned by any API response.
- `decrypt_apple_pay_wallet_data` prefers the internally-generated key when present, falling back to the merchant-supplied key otherwise.
- Merchant connector account update no longer silently drops the internally-generated key when a merchant updates other fields without resending `metadata` (fixed during testing).
- Fully additive: merchants who never call the new endpoint see zero change in stored data, API responses, or decrypt behavior.

## Why
Today, using Hyperswitch-side Apple Pay decryption requires the merchant to generate their own EC private key and paste it into the dashboard — putting the merchant in PCI scope for that key. This lets Hyperswitch generate and retain the key internally instead, while the merchant only ever handles a CSR and the certificate Apple issues back.

## Test plan
- [x] `cargo check` clean on v1 (default), v2/olap, and full workspace — no new warnings.
- [x] Manually verified end-to-end against a local stack: `generate` returns a valid CSR (`openssl req -text -noout` parses it cleanly); the CSR was submitted through Apple's real PPC issuance flow and the signed certificate was uploaded back onto the MCA.
- [x] Confirmed via direct DB query that the raw private key is stored (required for decryption) and confirmed via the live API that it never appears in any `GET`/update response — only the masked placeholder does.
- [x] Confirmed updating the MCA without resending `metadata` leaves the internally-generated key and all other metadata fields intact (regression check for a bug found during testing).
- [x] Confirmed a full real-world round trip: generated a keypair, got Apple's certificate issued and uploaded, and successfully decrypted a freshly-generated Apple Pay token end-to-end.

https://claude.ai/code/session_01CCMh1qVfR2NNMBbrGpUqi5

### DIFF
--- a/api-reference/v1/openapi_spec_v1.json
+++ b/api-reference/v1/openapi_spec_v1.json
@@ -32863,6 +32863,10 @@
           },
           "payment_processing_certificate_key": {
             "type": "string"
+          },
+          "system_generated_payment_processing_certificate_key": {
+            "type": "string",
+            "nullable": true
           }
         }
       },
--- a/api-reference/v2/openapi_spec_v2.json
+++ b/api-reference/v2/openapi_spec_v2.json
@@ -24647,6 +24647,10 @@
           },
           "payment_processing_certificate_key": {
             "type": "string"
+          },
+          "system_generated_payment_processing_certificate_key": {
+            "type": "string",
+            "nullable": true
           }
         }
       },
--- a/crates/api_models/src/payments.rs
+++ b/crates/api_models/src/payments.rs
@@ -10509,6 +10509,10 @@ pub struct PaymentProcessingDetails {
     #[schema(value_type = String)]
     #[smithy(value_type = "String")]
     pub payment_processing_certificate_key: Secret<String>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    #[schema(value_type = Option<String>)]
+    #[smithy(value_type = "Option<String>")]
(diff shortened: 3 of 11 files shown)


XyneSpaces:
[should-fix] `crates/router/src/core/admin.rs:~3177` persists the generated Apple Pay metadata with `MerchantConnectorAccountUpdate::ConnectorWebhookRegisterationUpdate { connector_webhook_registration_details: None, connector_webhook_details: None, metadata: Some(...) }`. Because this update variant overwrites the webhook fields, an MCA that already has webhook registration details will have them cleared when the certificate is generated. Use a metadata-only update variant or preserve the existing webhook values.

ShankarSinghC:
Closing this pr as the these changes are taken in https://github.com/juspay/hyperswitch/pull/14140

PR #14039, opened 2026-09-03
refactor(observability): rename the `alerts` crate to `observability`

ThisIsMani:
Renames the alert-delivery crate so that the deployed artifact, the crate and the domain language
agree. The crate is `observability`, with the Notifier as one component inside it rather than the
whole of it. Targets `observability-plane-i`.

Closes juspay/hyperswitch-cloud#23219.

## Why the crate is `observability` and the deployment is `hyperswitch-observability-plane`

The two names differ on purpose. **"Plane" is a deployment-topology word** — a tier deployed and
scaled apart from the data path. That is true of the standalone binary and false of the embedded
library the router mounts in-process, where there is no separate plane at all. The suffix therefore
belongs to the artifact that is one, and not to the crate that is both.

The `router` precedent (crate `router` → artifact `hyperswitch-router`) does not apply, because
`router` has one shape and so its two names cannot disagree. Ours can, and the difference carries
information.

The rejected alternatives were `observability` everywhere with artifact
`hyperswitch-observability`, `observability_plane` everywhere, and a narrower name such as
`notifier`. The crate README carries this reasoning under "Why the deployed name differs", so that
someone reading the Helm chart and grepping for `observability-plane` finds the connection.

`hyperswitch-alerts` is untouched: it belongs to the R alerts service, which owns that name in ECR,
`.jenkins/` and `helm-charts/charts/`.

## What changed

| | from | to |
|---|---|---|
| directory | `crates/alerts` | `crates/observability` |
| package + binary | `alerts` | `observability` |
| config file | `config/alerts.toml` | `config/observability.toml` |
| env prefix | `ALERTS__` | `OBSERVABILITY__` |
| error types | `AlertsError`, `AlertsApiResult`, `AlertsResult` | `ObservabilityError`, `ObservabilityApiResult`, `ObservabilityResult` |
| wire error type | `"alerts_error"` | `"observability_error"` |

The workspace globs `crates/*`, so there is no member edit. The Dockerfile is already generic
(`ARG BINARY`), so it needs no change either — Jenkins passes `BINARY=observability`
(juspay/hyperswitch-cloud#23220).

## Routes (unchanged by this PR)

The whole tree is in `crates/observability/src/routes/app.rs`. Nothing here moved — listed so the
rename can be read against the surface it does *not* touch.

| Method | Path | Auth |
|---|---|---|
| `POST` | `/alerts/chat/notify/{destination}` | `X-Internal-Api-Key` |
| `POST` | `/alerts/email/notify/{destination}` | `X-Internal-Api-Key` |
| `GET` | `/health` | none — liveness probe |

The two notify routes hang off `routes::Alerts`, the guarded scope; `/health` hangs off
`routes::Health`, a separate unguarded scope so that "unauthenticated" is a structural property of
the route tree rather than an exception inside a guard. The path carries the channel and the
destination; the body is content only, and a provider *refusing* a message comes back `200` rather
than as an error.

## What deliberately did not change

- **The HTTP scope stays `/alerts`**, along with `routes::Alerts`. The path names the resource being
  posted, not the service, and stays correct as the crate widens past delivery — the same reasoning
  that makes the Notifier one component inside the crate rather than the whole of it.
- **Every domain use of "alert" stays.** The `sr_alerts` destination id, `alerts@example.com`, and
  prose about what an alert is and when one fires. Only crate identity moved.

The wire string `"alerts_error"` is the one API-visible change. It names the *service* at fault, so
it does track the rename; nothing consumes it yet (juspay/hyperswitch-cloud#23119 is not built), so
it is free now and a breaking change later.

## Why now

Three branches are building on `crates/alerts` — `23111-email-notifier`, `23116-notifier-interface`,
`23119-r-alerts-integration` — so each rebase costs a `git mv` and a find-and-replace. Waiting costs
those same rebases plus a second pass over the Jenkinsfile, the Helm chart, the values files and the
secret keys, all of which bake in the binary name, the config path and the environment prefix.

## Testing

All run locally in the nix devshell:

- `cargo check -p observability --features v1` — clean.
- `cargo check --no-default-features --features "release,v2,redis-rs"` — clean. This is the CI v2
  gate, and it is the check the issue's "the workspace builds" wording asks for.
- `cargo test -p observability --features v1` — 12 passed, 0 failed (`tests/notify.rs`), plus the
  lib and doc test targets.
- `cargo clippy -p observability --features v1 --all-targets` — no findings in `crates/observability`.
- `cargo +nightly fmt -p observability --check` — clean.

Note for reviewers: `cargo check -p observability --no-default-features --features v2` fails **in
isolation**, in `hyperswitch_domain_models`, which imports `crate::revenue_recovery` under plain
`cfg(feature = "v2")` while the module is gated on `all(v2, revenue_recovery)`. This is pre-existing
— the identical check against the old crate name on the base commit fails the same way — and CI
never hits it because it checks the whole workspace, where feature unification supplies
`revenue_recovery`. Nothing in `hyperswitch_domain_models` is touched by this PR.

https://claude.ai/code/session_01RupqpSYGSzxRwjUZkZcbpk


docs/adr/0001-observability-crate-and-plane-naming.md
ThisIsMani, line 1: Can we remove this.

ThisIsMani: Removed, along with `docs/adr/` — this PR no longer creates that directory.

The reasoning isn't lost: it lives in `crates/observability/README.md` under "Why the deployed name
differs", which is the place someone actually looks when they read `hyperswitch-observability-plane`
in the Helm chart and wonder where the matching crate is. The rejected alternatives are in the PR
description.

PR #14042, opened 2026-09-03
fix(connector): fall back to BIN-detected card network for ACI paymentBrand

### LINKED ISSUE #9059: [BUG] : [ACI] add paymentBrand to the card payment request
### Bug Description

ACI has made the `paymentBrand` field in their payments API mandatory. Currently this value is not sent by hyperswitch

### Expected Behavior

Depending on the card brand send the `paymentBrand` to ACI

### Actual Behavior

Currently `paymentBrand` is not sent to the connector for card payments leading to a `Field is paymentBrand and the message is may not be null` error from connector

### Steps To Reproduce

Provide an unambiguous set of steps to reproduce this bug. Include code or configuration to reproduce, if relevant.

### Context For The Bug

_No response_

### Environment

Are you using hyperswitch hosted version? Yes/No
If yes, please provide the value of the `x-request-id` response header to help us debug your issue.

1. Operating System or Linux distribution:
2. Rust version (output of `rustc --version`): ``
3. App version (output of `cargo r --features vergen -- --version`): ``

None

KeEbEe123:
## Description

Fixes #9059 — silent-failure mode in the ACI connector's card payment brand resolution (`aci/transformers.rs`):

- **Authorize/general card path** (`PaymentDetails::TryFrom<(Card, ...)>`) - if `card_network` was not explicitly specified, `.ok()` swallowed the failure in `get_aci_payment_brand(...)`, producing `None`. The field `paymentBrand` is required in ACI's Payments API, so it was silently skipped in the request and ACI rejected the payment outright with `"paymentBrand may not be null"`. Now detects the network via the BIN code first (`CardIssuer::from_isin`) before failing; works like 14 other connectors do (adyen, cybersource, payone, wellsfargo, ...).

- **SetupMandate path** - the exact same `.ok()` swallowing existed one more time here, unseen until now. If the result of the resolution was `None`, it was silently accepted, so the mandate setup could go through without `paymentBrand` the same way. Now propagates the detection logic and errors out instead.

Added `card_issuer_to_payment_brand` to map `CardIssuer` into ACI's `PaymentBrand` enumeration; it returns `NotSupported` on ACI's unsupported issuers (Carte Blanche, Cartes Bancaires) as expected.

Uses `CardIssuer::from_isin` (an explicitly ordered BIN range table, already used by the checkout connector) rather than the unordered `HashMap`-backed `CardData::get_card_issuer`, which has no defined precedence for overlapping BIN ranges (e.g. Discover vs UnionPay's `622126`-`625925`) and could otherwise misclassify a real card.

## Testing

Added 12 unit tests covering: explicit network, BIN fallback detection for Visa/MC/Amex/JCB/DinersClub/Discover/UnionPay/Maestro, the Discover/UnionPay overlap range resolving deterministically, an unrecognized BIN, an unsupported issuer, and the Carte Blanche BIN case erroring instead of being misclassified as Diners Club.

Verified via `cargo check --tests -p hyperswitch_connectors` (full feature set: `v1,dummy_connector,frm,payouts`) — 0 errors.

## Related Issue

Fixes #9059

### DIFF
--- a/crates/hyperswitch_connectors/src/connectors/aci/transformers.rs
+++ b/crates/hyperswitch_connectors/src/connectors/aci/transformers.rs
@@ -35,8 +35,8 @@ use super::aci_result_codes::{FAILURE_CODES, PENDING_CODES, SUCCESSFUL_CODES};
 use crate::{
     types::{RefundsResponseRouterData, ResponseRouterData},
     utils::{
-        self, CardData, NetworkTokenData as NetworkTokenDataTrait, PaymentsAuthorizeRequestData,
-        PhoneDetailsData, RouterData as _,
+        self, CardData, CardIssuer, NetworkTokenData as NetworkTokenDataTrait,
+        PaymentsAuthorizeRequestData, PhoneDetailsData, RouterData as _,
     },
 };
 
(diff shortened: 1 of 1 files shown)


XyneSpaces:

**Summary:** Solid connector fix with good test coverage; only minor performance and consistency nits.

- **[nit]** `crates/hyperswitch_connectors/src/connectors/aci/transformers.rs:599` — `card_data.card_network.clone()` clones the `Option<CardNetwork>` unnecessarily. Use `card_data.card_network.as_ref()` and dereference inside the `Some` arm (if `CardNetwork` is `Copy`) to avoid the extra allocation.

- **[nit]** `crates/hyperswitch_connectors/src/connectors/aci/transformers.rs:584` — The `NotSupported` error hardcodes the connector name `"ACI"`. Prefer the connector identifier constant or the `Self::id()` pattern used elsewhere so the name stays consistent if the connector is renamed.

PR #14050, opened 2026-09-04
feat(payments): add update intent api support to have update intent, session tokens, combined pml flow based on integration header

prasunna09:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

Adds a server-integration response shape to the existing update-intent endpoint,
selected by a request header rather than a new route.

`POST /payments/{payment_id}` now reads `X-Integration-Type`:

- `client`, an unrecognised value, or **no header at all** → the response is
  byte-for-byte what it is today. This is the default.
- `server` → the same payment response, additionally carrying two fields:
  - `payment_method_list` — the combined payment-method list, identical in shape
    to `GET /payments/{id}/payment-methods/client`
  - `session_tokens` — the full `PaymentsSessionResponse`, identical to
    `POST /payments/session_tokens`, including `vault_details` (the internal
    vault SDK authorization)

A server-to-server merchant integrating a checkout needs all three of these
before it can render anything. Today that is three sequential round trips, with
the client secret threaded through two of them. This collapses it to one call
authenticated by the merchant API key.

No new route is introduced, so there is one update-intent contract to maintain
rather than a client variant and a server variant that drift apart.

### Additional Changes

## How did you test it?

1. Create payment (amount 6540, confirm false)

curl -X POST 'http://localhost:8081/payments' -H 'api-key: <API_KEY>' -H 'Content-Type: application/json' \
  -d '{"amount":6540,"currency":"USD","confirm":false,"customer_id":"cus_mod","email":"m@example.com"}'

HTTP 200

{
  "payment_id": "pay_U36zXVvdEce3OpToswsJ",
  "merchant_id": "mod_1788556424",
  "status": "requires_payment_method",
  "amount": 6540,
  "net_amount": 6540,
  "shipping_cost": null,
  "amount_capturable": 0,
  "amount_received": null,
  "processor_merchant_id": "mod_1788556424",
  "initiator": null,
  "sdk_authorization": "cHJvZmlsZV9pZD1wcm9fMVFIdG9wMzZPMUlKTHhHUVBzaHAscHVibGlzaGFibGVfa2V5PXBrX2Rldl8zZjc0ZDFhZDBiMzU0MjIyYWZjZmI0NzY5MzdhOWNjOCxjbGllbnRfc2VjcmV0PXBheV9VMzZ6WFZ2ZEVjZTNPcFRvc3dzSl9zZWNyZXRfOEc0alNyV3lMQm5hbU5kT1JFV2osY3VzdG9tZXJfaWQ9Y3VzX21vZCxjbGllbnRfc2Vzc2lvbl9pZD1jbGllbnRfc2Vzc192T21qbjhMY29nTkNSTjBxbUJmYSxwYXltZW50X2lkPXBheV9VMzZ6WFZ2ZEVjZTNPcFRvc3dzSg==",
  "connector": null,
  "state_metadata": null,
  "client_secret": "pay_U36zXVvdEce3OpToswsJ_secret_8G4jSrWyLBnamNdOREWj",
  "created": "2026-09-04T21:13:45.115Z",
  "modified_at": "2026-09-04T21:13:45.219Z",
  "connector_customer_id": null,
  "currency": "USD",
  "customer_id": "cus_mod",
  "customer": {
    "id": "cus_mod",
    "name": null,
    "email": "m@example.com",
    "phone": null,
    "phone_country_code": null,
    "customer_document_details": null
  },
  "description": null,
  "refunds": null,
  "disputes": null,
  "mandate_id": null,
  "mandate_data": null,
  "setup_future_usage": null,
  "off_session": null,
  "capture_on": null,
  "capture_method": "automatic",
  "payment_method": null,
  "payment_method_data": null,
  "payment_token": null,
  "shipping": null,
  "billing": null,
  "order_details": null,
  "email": "m@example.com",
  "name": null,
  "phone": null,
  "return_url": null,
  "authentication_type": null,
  "statement_descriptor_name": null,
  "statement_descriptor_suffix": null,
  "next_action": null,
  "cancellation_reason": null,
  "error_code": null,
  "error_message": null,
  "unified_code": null,
  "unified_message": null,
  "error_details": null,
  "payment_experience": null,
  "payment_method_type": null,
  "connector_label": null,
  "business_country": null,
  "business_label": "default",
  "business_sub_label": null,
  "allowed_payment_method_types": null,
  "manual_retry_allowed": null,
  "connector_transaction_id": null,
  "frm_message": null,
  "metadata": null,
  "connector_metadata": null,
  "connector_response_metadata": null,
  "feature_metadata": null,
  "reference_id": null,
  "payment_link": null,
  "profile_id": "pro_1QHtop36O1IJLxGQPshp",
  "surcharge_details": null,
  "applied_offer": null,
  "attempt_count": 1,
  "merchant_decision": null,
  "merchant_connector_id": null,
  "incremental_authorization_allowed": null,
  "authorization_count": null,
  "incremental_authorizations": null,
  "external_authentication_details": null,
  "external_3ds_authentication_attempted": false,
  "expires_on": "2026-09-04T21:28:45.115Z",
  "fingerprint": null,
  "fingerprint_type": null,
  "browser_info": null,
  "payment_channel": null,
  "payment_method_id": null,
  "network_transaction_id": null,
  "network_transaction_link_id": null,
  "payment_method_status": null,
  "updated": "2026-09-04T21:13:45.219Z",
  "split_payments": null,
  "frm_metadata": null,
  "extended_authorization_applied": null,
  "extended_authorization_last_applied_at": null,
  "request_extended_authorization": null,
  "capture_before": null,
  "merchant_order_reference_id": null,
  "order_tax_amount": null,
  "connector_mandate_id": null,
  "card_discovery": null,
  "force_3ds_challenge": false,
  "force_3ds_challenge_trigger": false,
  "issuer_error_code": null,
  "issuer_error_message": null,
  "whole_connector_response": null,
  "enable_partial_authorization": null,
  "enable_overcapture": null,
  "is_overcapture_enabled": null,
  "network_details": null,
  "is_stored_credential": null,
  "mit_category": null,
  "billing_descriptor": null,
  "tokenization": null,
  "partner_merchant_identifier_details": null,
  "payment_method_tokenization_details": null,
  "installment_options": null,
  "installment_data": null,
  "sender_payment_instrument_id": null
}

---

2. Update WITHOUT X-Integration-Type

curl -X POST 'http://localhost:8081/payments/pay_vfee37UMGlVH1x6wpsv8' -H 'api-key: <API_KEY>' \
  -H 'Content-Type: application/json' -d '{"amount":12999}'

HTTP 200

{
  "payment_id": "pay_vfee37UMGlVH1x6wpsv8",
  "merchant_id": "mod_1788556424",
  "status": "requires_payment_method",
  "amount": 12999,
  "net_amount": 12999,
  "shipping_cost": null,
  "amount_capturable": 0,
  "amount_received": null,
  "processor_merchant_id": "mod_1788556424",
  "initiator": null,
  "sdk_authorization": "cHJvZmlsZV9pZD1wcm9fMVFIdG9wMzZPMUlKTHhHUVBzaHAscHVibGlzaGFibGVfa2V5PXBrX2Rldl8zZjc0ZDFhZDBiMzU0MjIyYWZjZmI0NzY5MzdhOWNjOCxjbGllbnRfc2VjcmV0PXBheV92ZmVlMzdVTUdsVkgxeDZ3cHN2OF9zZWNyZXRfaDQyemZmZ1R2UW1Cb3FmekdzbG0sY3VzdG9tZXJfaWQ9Y3VzX21vZCxjbGllbnRfc2Vzc2lvbl9pZD1jbGllbnRfc2Vzc19xbnR0OHFyVlkza05OSDNEV1ZzSyxwYXltZW50X2lkPXBheV92ZmVlMzdVTUdsVkgxeDZ3cHN2OA==",
  "connector": null,
  "state_metadata": null,
  "client_secret": "pay_vfee37UMGlVH1x6wpsv8_secret_h42zffgTvQmBoqfzGslm",
  "created": "2026-09-04T21:13:45.307Z",
  "modified_at": "2026-09-04T21:13:45.399Z",
  "connector_customer_id": null,
  "currency": "USD",
  "customer_id": "cus_mod",
  "customer": {
    "id": "cus_mod",
    "name": null,
    "email": "m@example.com",
    "phone": null,
    "phone_country_code": null,
    "customer_document_details": null
  },
  "description": null,
  "refunds": null,
  "disputes": null,
  "mandate_id": null,
  "mandate_data": null,
  "setup_future_usage": null,
  "off_session": null,
  "capture_on": null,
  "capture_method": "automatic",
  "payment_method": null,
  "payment_method_data": null,
  "payment_token": null,
  "shipping": null,
  "billing": null,
  "order_details": null,
  "email": "m@example.com",
  "name": null,
  "phone": null,
  "return_url": null,
  "authentication_type": null,
  "statement_descriptor_name": null,
  "statement_descriptor_suffix": null,
  "next_action": null,
  "cancellation_reason": null,
  "error_code": null,
  "error_message": null,
  "unified_code": null,
  "unified_message": null,
  "error_details": null,
  "payment_experience": null,
  "payment_method_type": null,
  "connector_label": null,
  "business_country": null,
  "business_label": "default",
  "business_sub_label": null,
  "allowed_payment_method_types": null,
  "manual_retry_allowed": null,
  "connector_transaction_id": null,
  "frm_message": null,
  "metadata": null,
  "connector_metadata": null,
  "connector_response_metadata": null,
  "feature_metadata": null,
  "reference_id": null,
  "payment_link": null,
  "profile_id": "pro_1QHtop36O1IJLxGQPshp",
  "surcharge_details": null,
  "applied_offer": null,
  "attempt_count": 1,
  "merchant_decision": null,
  "merchant_connector_id": null,
  "incremental_authorization_allowed": null,
  "authorization_count": null,
  "incremental_authorizations": null,
  "external_authentication_details": null,
  "external_3ds_authentication_attempted": false,
  "expires_on": "2026-09-04T21:28:45.307Z",
  "fingerprint": null,
  "fingerprint_type": null,
  "browser_info": null,
  "payment_channel": null,
  "payment_method_id": null,
  "network_transaction_id": null,
  "network_transaction_link_id": null,
  "payment_method_status": null,
  "updated": "2026-09-04T21:13:45.399Z",
  "split_payments": null,
  "frm_metadata": null,
  "extended_authorization_applied": null,
  "extended_authorization_last_applied_at": null,
  "request_extended_authorization": null,
  "capture_before": null,
  "merchant_order_reference_id": null,
  "order_tax_amount": null,
  "connector_mandate_id": null,
  "card_discovery": null,
  "force_3ds_challenge": false,
  "force_3ds_challenge_trigger": false,
  "issuer_error_code": null,
  "issuer_error_message": null,
  "whole_connector_response": null,
  "enable_partial_authorization": null,
  "enable_overcapture": null,
  "is_overcapture_enabled": null,
  "network_details": null,
  "is_stored_credential": null,
  "mit_category": null,
  "billing_descriptor": null,
  "tokenization": null,
  "partner_merchant_identifier_details": null,
  "payment_method_tokenization_details": null,
  "installment_options": null,
  "installment_data": null,
  "sender_payment_instrument_id": null
}

No payment_method_list, no session_tokens. Byte-for-byte the old shape.

---

3. Update WITH X-Integration-Type: server

curl -X POST 'http://localhost:8081/payments/pay_7bJ9w3PJICNJmXNAx2w4' -H 'api-key: <API_KEY>' \
  -H 'X-Integration-Type: server' \
  -H 'Content-Type: application/json' -d '{"amount":12999}'

HTTP 200

{
  "payment_id": "pay_7bJ9w3PJICNJmXNAx2w4",
  "merchant_id": "mod_1788556424",
  "status": "requires_payment_method",
  "amount": 12999,
  "net_amount": 12999,
  "shipping_cost": null,
  "amount_capturable": 0,
  "amount_received": null,
  "processor_merchant_id": "mod_1788556424",
  "initiator": null,
  "sdk_authorization": "cHJvZmlsZV9pZD1wcm9fMVFIdG9wMzZPMUlKTHhHUVBzaHAscHVibGlzaGFibGVfa2V5PXBrX2Rldl8zZjc0ZDFhZDBiMzU0MjIyYWZjZmI0NzY5MzdhOWNjOCxjbGllbnRfc2VjcmV0PXBheV83Yko5dzNQSklDTkptWE5BeDJ3NF9zZWNyZXRfMXNIMG43UUJSSmdQWnV6VE9KYTgsY3VzdG9tZXJfaWQ9Y3VzX21vZCxjbGllbnRfc2Vzc2lvbl9pZD1jbGllbnRfc2Vzc19oalJRcTU2cjNBSjhhRGlLakxhYixwYXltZW50X2lkPXBheV83Yko5dzNQSklDTkptWE5BeDJ3NA==",
  "payment_method_list": {
    "payment_methods_enabled": [
      {
        "payment_method": "card",
        "card_networks": [
          "Mastercard",
          "Visa"
        ],
        "payment_experience": null,
        "collect_shipping_details_from_wallets": null,
        "collect_billing_details_from_wallets": null,
        "customer_acceptance_support": "supported"
      },
      {
        "payment_method": "card",
        "payment_method_type": "debit",
        "card_networks": [
          "Mastercard",
          "Visa"
        ],
        "payment_experience": null,
        "collect_shipping_details_from_wallets": null,
        "collect_billing_details_from_wallets": null,
        "customer_acceptance_support": "supported"
      }
    ],
    "customer_payment_methods": [],
    "sdk_next_action": {
      "next_action": "confirm",
      "should_block_confirm": false
    },
    "intent_data": {
      "payment_id": "pay_7bJ9w3PJICNJmXNAx2w4",
      "status": "requires_payment_method",
      "amount": 12999,
      "currency": "USD",
      "client_secret": "pay_7bJ9w3PJICNJmXNAx2w4_secret_1sH0n7QBRJgPZuzTOJa8",
      "description": null,
      "customer_id": "cus_mod",
      "return_url": null,
      "setup_future_usage": null,
      "billing": {
        "address": null,
        "phone": null,
        "email": "m@example.com"
      },
      "shipping": null,
      "email": "m@example.com",
      "metadata": null,
      "order_details": null,
      "created": "2026-09-04 21:13:45.451822",
      "expires_on": "2026-09-04 21:28:45.45181",
      "profile_id": "pro_1QHtop36O1IJLxGQPshp",
      "merchant_order_reference_id": null,
      "attempt_count": 1,
      "installment_options": null,
      "capture_method": "automatic",
      "merchant_name": "Modular Demo",
      "mandate_payment": null,
      "payment_type": "normal",
      "request_external_three_ds_authentication": false,
      "is_tax_calculation_enabled": false,
      "is_guest_customer": false
    }
  },
  "session_tokens": {
    "payment_id": "pay_7bJ9w3PJICNJmXNAx2w4",
    "client_secret": "pay_7bJ9w3PJICNJmXNAx2w4_secret_1sH0n7QBRJgPZuzTOJa8",
    "session_token": [],
    "vault_details": null
  },
  "connector": null,
  "state_metadata": null,
  "client_secret": "pay_7bJ9w3PJICNJmXNAx2w4_secret_1sH0n7QBRJgPZuzTOJa8",
  "created": "2026-09-04T21:13:45.451Z",
  "modified_at": "2026-09-04T21:13:45.513Z",
  "connector_customer_id": null,
  "currency": "USD",
  "customer_id": "cus_mod",
  "customer": {
    "id": "cus_mod",
    "name": null,
    "email": "m@example.com",
    "phone": null,
    "phone_country_code": null,
    "customer_document_details": null
  },
  "description": null,
  "refunds": null,
  "disputes": null,
  "mandate_id": null,
  "mandate_data": null,
  "setup_future_usage": null,
  "off_session": null,
  "capture_on": null,
  "capture_method": "automatic",
  "payment_method": null,
  "payment_method_data": null,
  "payment_token": null,
  "shipping": null,
  "billing": null,
  "order_details": null,
  "email": "m@example.com",
  "name": null,
  "phone": null,
  "return_url": null,
  "authentication_type": null,
  "statement_descriptor_name": null,
  "statement_descriptor_suffix": null,
  "next_action": null,
  "cancellation_reason": null,
  "error_code": null,
  "error_message": null,
  "unified_code": null,
  "unified_message": null,
  "error_details": null,
  "payment_experience": null,
  "payment_method_type": null,
  "connector_label": null,
  "business_country": null,
  "business_label": "default",
  "business_sub_label": null,
  "allowed_payment_method_types": null,
  "manual_retry_allowed": null,
  "connector_transaction_id": null,
  "frm_message": null,
  "metadata": null,
  "connector_metadata": null,
  "connector_response_metadata": null,
  "feature_metadata": null,
  "reference_id": null,
  "payment_link": null,
  "profile_id": "pro_1QHtop36O1IJLxGQPshp",
  "surcharge_details": null,
  "applied_offer": null,
  "attempt_count": 1,
  "merchant_decision": null,
  "merchant_connector_id": null,
  "incremental_authorization_allowed": null,
  "authorization_count": null,
  "incremental_authorizations": null,
  "external_authentication_details": null,
  "external_3ds_authentication_attempted": false,
  "expires_on": "2026-09-04T21:28:45.451Z",
  "fingerprint": null,
  "fingerprint_type": null,
  "browser_info": null,
  "payment_channel": null,
  "payment_method_id": null,
  "network_transaction_id": null,
  "network_transaction_link_id": null,
  "payment_method_status": null,
  "updated": "2026-09-04T21:13:45.513Z",
  "split_payments": null,
  "frm_metadata": null,
  "extended_authorization_applied": null,
  "extended_authorization_last_applied_at": null,
  "request_extended_authorization": null,
  "capture_before": null,
  "merchant_order_reference_id": null,
  "order_tax_amount": null,
  "connector_mandate_id": null,
  "card_discovery": null,
  "force_3ds_challenge": false,
  "force_3ds_challenge_trigger": false,
  "issuer_error_code": null,
  "issuer_error_message": null,
  "whole_connector_response": null,
  "enable_partial_authorization": null,
  "enable_overcapture": null,
  "is_overcapture_enabled": null,
  "network_details": null,
  "is_stored_credential": null,
  "mit_category": null,
  "billing_descriptor": null,
  "tokenization": null,
  "partner_merchant_identifier_details": null,
  "payment_method_tokenization_details": null,
  "installment_options": null,
  "installment_data": null,
  "sender_payment_instrument_id": null
}

---

4. Standalone POST /payments/session_tokens (baseline)

curl -X POST 'http://localhost:8081/payments/session_tokens' -H 'api-key: <PUBLISHABLE_KEY>' \
  -H 'Content-Type: application/json' \
  -d '{"payment_id":"pay_7bJ9w3PJICNJmXNAx2w4","client_secret":"<CS>","wallets":["google_pay"]}'

HTTP 200

{
  "payment_id": "pay_7bJ9w3PJICNJmXNAx2w4",
  "client_secret": "pay_7bJ9w3PJICNJmXNAx2w4_secret_1sH0n7QBRJgPZuzTOJa8",
  "session_token": [],
  "vault_details": null
}

---

Assertions

no header    amount=12999  pml=false [-]                session_tokens=false [-]
server       amount=12999  pml=true  [ok(2 methods)]    session_tokens=true  [ok(0 tokens, vault=false)]
standalone                                              0 tokens, vault=false

## Checklist


kashif-m (approved):

crates/router/src/core/payments/update_context.rs
```diff
@@ -0,0 +1,270 @@
...
+    // ── 1 · Update — hard gate ───────────────────────────────────────────────
+    // Same core and operation the standalone update route uses. A failure here fails the whole
+    // request: nothing downstream would be meaningful, and no state has changed yet.
+    let update_response = Box::pin(payments::payments_core::<
+        api_types::UpdatePostConfirm,
+        payment_types::PaymentsResponse,
+        _,
+        _,
+        _,
+        payments::PaymentData<api_types::UpdatePostConfirm>,
```
kashif-m: Should we consider adding a separate operation for payment update?

kashif-m (dismissed):

XyneSpaces:

**Summary:** Well-structured composite endpoint that updates the intent and refreshes client context; main fixes needed are leaking debug errors, discarding detected auth flow, and a silent Superposition call in the new test script.

- **[should-fix]** `crates/router/src/core/payments/update_context.rs:104` — `degrade_on_error` embeds `format!("{error:?}")` in the merchant-facing `UpdateContextWarning.message`, which can leak internal debug details. Use a stable, sanitized message in the response and keep the detailed error in the existing `logger::warn!` call.

- **[should-fix]** `crates/router/src/routes/payments.rs:1566` — The handler destructures `(auth_type, _auth_flow)` from `check_internal_api_key_auth_no_client_secret` but hard-codes `AuthFlow::Merchant` inside the core closure. Propagate the detected `auth_flow` instead so internal-service callers behave identically to the standalone update route.

- **[should-fix]** `scripts/test_platform_modular_nt.sh:73` — `set_modular` calls Superposition with `-o /dev/null` and never checks the HTTP status, so a failed override silently leaves stale feature-flag state. Capture the status and exit early if the context update is unsuccessful.

- **[nit]** `crates/router/src/core/payments/update_context.rs:117` — `json_body` collapses any non-JSON `ApplicationResponse` variant into `InternalServerError` without logging which variant was seen. Add a log line or include the variant name before returning to ease debugging if a composite section unexpectedly returns a redirect or file response.

XyneSpaces:
[should-fix] `crates/router/src/core/payments/update_context.rs` (~line 95)

`degrade_on_error` stores `format!("{error:?}")` in `UpdateContextWarning.message`, which exposes the internal `Debug` representation (potentially stack traces and connector internals) to API consumers. Replace it with a stable, human-readable message or an error code.

```suggestion
            warnings.push(payment_types::UpdateContextWarning {
                section,
                code: code.to_string(),
                wallet: None,
                message: "Section failed after the update committed".to_string(),
            });
```

If detailed diagnostics are needed, log the full `?error` on the server side and return only a code or sanitized summary in the response.

XyneSpaces:
Automated review findings:

[blocking] `crates/router/src/core/payments/update_context.rs:476` — The second `fp_utils::when` call validating amount > 0 is missing `?`. The `RouterResult` is silently dropped, so requests with zero/negative amounts bypass validation. Append `?` to propagate the error.

[should-fix] `crates/router/src/core/payments/update_context.rs:517` — `UpdateContextWarning.message` is built with `format!("{error:?}")`, exposing raw debug internals to API consumers. Use a fixed, caller-safe message and log the detailed error server-side.

[should-fix] `crates/router/src/core/payments/update_context.rs:142` — The new `payments_update_context` core flow has no unit or integration tests for validation, successful update, or degraded section handling. Add tests covering the composite endpoint paths.

crates/router/src/core/payments/update_context.rs
prasunna09: yeah will add this in separate pr

XyneSpaces:
[blocking] `UpdateContextWarning.message` is populated with `format!("{error:?}")` at `crates/router/src/core/payments/update_context.rs:97`. This serializes the internal debug representation of the error and returns it to API consumers, potentially exposing stack traces, internal identifiers, or connector payloads. Use a static, user-facing message here and keep the diagnostic detail in the server-side log that already exists on lines 88–92.

XyneSpaces:
[blocking] The server-integration response should not expose the full `PaymentsSessionResponse` because that shape includes `vault_details` with internal vault SDK authorization. Please return only the wallet session-token array (or a redacted DTO) for `SessionTokensResult::Success` so API-key callers cannot receive internal vault authorization material.

kashif-m (approved):

crates/router/src/core/payments/update_context.rs
```diff
@@ -0,0 +1,333 @@
...
+            errors::ApiErrorResponse::InternalServerError
+        ))
+        .attach_printable_lazy(|| {
+            format!("server-integration: {section} core returned a non-JSON response")
+        }),
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
kashif-m, line 262: We don't need these tests

crates/router/src/core/payments/update_context.rs
```diff
@@ -0,0 +1,333 @@
...
+/// caller is entitled to. Each section is bounded independently so a slow one cannot starve the
+/// other.
+const SECTION_TIMEOUT: std::time::Duration = std::time::Duration::from_secs(10);
+
+/// Which integration the caller is building, taken from `X-Integration-Type`.
+///
+/// Defaults to `Client` when the header is absent or unrecognised, so an existing integration
+/// that has never heard of the header keeps its current response shape.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum IntegrationType {
```
kashif-m, line 42: Move to API models?

crates/router/src/core/payments/update_context.rs
```diff
@@ -0,0 +1,333 @@
...
+                unrecognised => {
+                    logger::warn!(
+                        header = consts::X_INTEGRATION_TYPE,
+                        value = unrecognised,
+                        "unrecognised integration type, defaulting to client"
+                    );
+                    Self::Client
+                }
+            })
+            .unwrap_or(Self::Client)
```
kashif-m, line 66: We can avoid checking for client since it's the default behavior by design

```suggestion
.map_or(Self::Client, |value| {
    let value = value.trim();

    if value.eq_ignore_ascii_case("server") {
        Self::Server
    } else {
        logger::warn!(
            header = consts::X_INTEGRATION_TYPE,
            value,
            "unrecognised integration type, defaulting to client"
        );
        Self::Client
    }
})
```

crates/router/src/routes/payments.rs
```diff
@@ -927,33 +929,88 @@ pub async fn payments_update(
...
+                        "server integration type requested on a client-authenticated request; \
+                         returning the client response shape"
+                    );
+                }
+
+                // Payments responses come back as `JsonWithHeaders`; `Json` is handled too so the
+                // enrichment does not silently skip if that ever changes.
+                let enrich_payment = |mut payment: payment_types::PaymentsResponse| async {
+                    let id = payment.payment_id.clone();
+                    Box::pin(update_context::attach_server_context(
```
kashif-m, line 988: Internally, this calls session and PML which has a timeout of 10s

Can we perform this outside `server_wrap`? Or else this step blocks the payment operations due to API locking

crates/router/src/routes/payments.rs
```diff
@@ -927,33 +929,88 @@ pub async fn payments_update(
...
+                    _,
+                    _,
+                    _,
+                    payments::PaymentData<api_types::UpdatePostConfirm>,
+                >(
+                    state.clone(),
+                    req_state.clone(),
+                    auth.platform,
+                    profile_id.clone(),
+                    payments::PaymentUpdate,
```
kashif-m, line 957: Can we avoid these clones for non server path? We can check if enrichment is needed - if not, there's no point cloning these

crates/router/src/core/payments/update_context.rs
```diff
@@ -0,0 +1,333 @@
...
+                header_payload,
+            )),
+        ),
+        tokio::time::timeout(
+            SECTION_TIMEOUT,
+            pm_client::list_payment_methods_client(
+                state.clone(),
+                platform.clone(),
+                payment_id.clone(),
+                // Merchant API key authenticated; there is no client secret to validate.
```
kashif-m, line 146: Can we avoid these clones?

crates/api_models/src/payments.rs
```diff
@@ -10442,11 +10480,40 @@ pub enum SessionToken {
...
     #[smithy(value_type = "AmazonPaySessionTokenResponse")]
     AmazonPay(Box<AmazonPaySessionTokenResponse>),
+    /// A wallet whose session token could not be minted.
+    ///
+    /// The connector call failed, or returned no usable token. Previously these were logged and
+    /// the entry dropped, which left the caller unable to tell "this wallet is not configured"
+    /// apart from "this wallet failed just now".
+    #[smithy(value_type = "SessionTokenErrorResponse")]
+    Failed(Box<SessionTokenErrorResponse>),
     /// Whenever there is no session token response or an error in session response
```
kashif-m, line 10490: Can we not have failed variant in this enum? We can handle this as a Result<> wherever we need to use this?

crates/api_models/src/payment_methods.rs
```diff
@@ -3160,8 +3160,45 @@ pub struct ClientPaymentMethodsListResponse {
...
+    /// Machine-readable error code, e.g. `HE_00`.
+    #[schema(example = "HE_00")]
+    pub code: String,
+}
+
+/// Wrapper carrying a section's error payload under an `error` key.
+#[cfg(feature = "v1")]
+#[derive(Debug, Clone, PartialEq, serde::Serialize, ToSchema)]
+pub struct SectionError {
+    pub error: SectionErrorDetail,
```
kashif-m, line 3186: Reuse ErrorResponse?

kashif-m (changes requested):

crates/router/src/core/payments/update_context.rs
prasunna09: Removed — the unit tests are gone.

crates/router/src/core/payments/update_context.rs
prasunna09: Moved. `IntegrationType` now lives in `api_models::payments`, with a pure `from_header_value(Option<&str>)`. The actix `HeaderMap` extraction and the warn stay in the router as a thin `integration_type_from_headers` wrapper — `api_models` has no `router_env` dep and only an optional `actix-web`, so keeping the parse pure avoids adding either.

crates/router/src/core/payments/update_context.rs
prasunna09: Applied — using `map_or` with `Client` as the default, no explicit client arm.

crates/api_models/src/payments.rs
prasunna09: Removed. Good catch — this was also the cause of the API Compatibility Check failure: adding the variant produced `response-property-one-of-added` x2 on the v2 schema. `SessionToken.oneOf` is now identical to base (10 == 10) on both v1 and v2. Per-wallet failures go back to being logged and dropped; the section-level `SessionTokensResult` still reports a failed session call.

crates/api_models/src/payment_methods.rs
prasunna09: Done — `SectionErrorDetail` is gone and `SectionError` now wraps `api_models::errors::types::ErrorResponse` directly, so an inline section error is byte-for-byte the standalone endpoint body. Needed `Clone + PartialEq` on `ErrorResponse`/`Extra` since `PaymentsResponse` derives `PartialEq`; both are derive-only, no wire change.

crates/router/src/routes/payments.rs
prasunna09: Fixed. `enrich` depends only on the header and `auth_flow`, both known before `server_wrap`, so it is computed once up front and the clones are now behind `enrich.then(|| ...)`. The client path clones nothing.

crates/router/src/core/payments/update_context.rs
prasunna09: Reduced. `attach_server_context` and `session_tokens` now take owned `SessionState` / `Platform` / `HeaderPayload` instead of refs, so the two concurrent futures share a single clone of each rather than cloning twice.

crates/router/src/routes/payments.rs
prasunna09: You are right, and the lock is real: `PaymentsRequest::get_locking_input` returns `Hold`/`HoldMultiple` on the intent (and customer) id, and `server_wrap_util` frees it only after the closure returns — so the enrichment currently holds it for up to `SECTION_TIMEOUT`.

I have not moved it out yet, because the obvious routes each have a cost I did not want to pick unilaterally:

1. `server_wrap` returns an already-serialised `HttpResponse`, so enriching after it returns means re-parsing the body. `server_wrap_util` does return the typed `ApplicationResponse` with the lock already freed, but reusing it means duplicating ~100 lines of `server_wrap`s response/metrics tail, or refactoring that tail into a shared fn.
2. Pass `LockAction::NotApplicable` to `server_wrap` and call `perform_locking_action` / `free_lock_action` manually around just the `payments_core` call. Smallest diff and it does exactly what you asked, but it puts hand-rolled distributed lock management in a payment route — any error path that misses the free leaks the lock.

(2) is what I would do if you are happy with it; I would rather not hand-roll locking here without your sign-off. Option (3) is bounding the exposure by dropping `SECTION_TIMEOUT` well below 10s, which shrinks the window without removing it.

Which would you prefer?

prasunna09: Following up on my earlier reply — after digging into the lock keys I think the enrichment is fine where it is, and moving it out would actually be the riskier change. Reasoning, in case you disagree:

The lock key is `API_LOCK_{merchant_id}_{api_identifier}_{unique_locking_key}` (`core/api_locking.rs:44`). `Flow::PaymentsUpdate` and `Flow::PaymentsSessionToken` both map to `ApiIdentifier::Payments` (`routes/lock_utils.rs`), and `PaymentsSessionRequest::get_locking_input` is `Hold { unique_locking_key: payment_id }`. So the session-token endpoint takes the *same* key this request already holds.

Two things follow:

1. The enrichment cannot go through `server_wrap` at all — nested, it would ask for the key the outer request holds and spin until `ResourceBusy`. It calls `payments_core` directly, so it takes no lock of its own. I have added a comment at the call site saying so, since it is not obvious and someone will be tempted to wrap it later.

2. On contention: a client doing `update` then `POST /payments/session_tokens` acquires that identical key twice anyway, so folding both into one hold adds nothing on the `payment_id` dimension.

The one place it is genuinely worse is that `PaymentsRequest::get_locking_input` is `HoldMultiple` — `payment_id` *and* `customer_id` — while session-tokens locks only `payment_id`. So the enrichment does extend the customer-level lock, which can block other payments for that customer by up to `SECTION_TIMEOUT`.

I did build the alternative (update under the lock, then a second `server_wrap` with `NotApplicable` doing retrieve + enrich) and backed it out: it emits two API events per request and rebuilds the body from `PaymentStatus` instead of `PaymentUpdate`, which is a lot of new surface for that one dimension. Happy to bring it back, or to drop `SECTION_TIMEOUT` from 10s to bound the customer-lock window, if you would rather not carry it.

prasunna09: even the call session needs a lock, so doing it inside the sever wrap.

crates/api_models/src/payment_methods.rs
```diff
@@ -3160,8 +3160,28 @@ pub struct ClientPaymentMethodsListResponse {
...
+#[cfg(all(feature = "v1", feature = "errors"))]
+#[derive(Debug, Clone, PartialEq, serde::Serialize, ToSchema)]
+#[serde(untagged)]
+pub enum PaymentMethodListResult {
+    Success(Box<ClientPaymentMethodsListResponse>),
+    /// Serializes as `{ "error": { ... } }` — the same envelope the HTTP layer puts around
+    /// `ErrorResponse`, so this reads identically to the standalone endpoint's error body.
+    Failed {
+        #[schema(value_type = GenericErrorResponseOpenApi)]
+        error: Box<crate::errors::types::ErrorResponse>,
```
kashif-m: avoid full imports


```suggestion
        error: Box<errors::types::ErrorResponse>,
```

crates/api_models/src/payments.rs
```diff
@@ -7353,6 +7353,31 @@ pub struct PaymentsResponse {
     #[smithy(value_type = "Option<String>")]
     pub sdk_authorization: Option<String>,
 
+    /// The combined payment-method list for this payment.
+    ///
+    /// Returned only when the request carries `X-Integration-Type: server`; absent from the
+    /// response for client integrations. On success this is the body
+    /// `GET /payments/{payment_id}/payment-methods/client` would have returned; on failure it is
+    /// `{ "error": { ... } }`.
+    #[cfg(feature = "errors")]
```
kashif-m: nit: we can write a more concise description for this - the variants are already understood by looking at the data type

crates/api_models/src/payments.rs
```diff
@@ -10410,6 +10435,56 @@ pub struct GooglePayTokenizationParameters {
...
+#[serde(untagged)]
+pub enum SessionTokensResult {
+    /// The session response exactly as `POST /payments/session_tokens` returns it, including
+    /// `vault_details` — which carries the internal vault SDK authorization.
+    Success(Box<PaymentsSessionResponse>),
+    /// Serializes as `{ "error": { ... } }` — the same envelope the HTTP layer puts around
+    /// `ErrorResponse`, so this reads identically to the standalone endpoint's error body.
+    Failed {
+        #[schema(value_type = GenericErrorResponseOpenApi)]
+        error: Box<crate::errors::types::ErrorResponse>,
```
kashif-m:
```suggestion
        error: Box<errors::types::ErrorResponse>,
```

crates/router/src/core/payments/update_context.rs
```diff
@@ -0,0 +1,224 @@
...
+};
+
+/// How long a single section may take before it is reported as degraded.
+///
+/// The session core runs with `CallConnectorAction::Trigger`, so it can make outbound connector
+/// calls, and the payment-method list is served by the modular service over HTTP. The payment has
+/// already committed by the time either runs, so a section that hangs would hold a response the
+/// caller is entitled to. Each section is bounded independently so a slow one cannot starve the
+/// other.
+const SECTION_TIMEOUT: std::time::Duration = std::time::Duration::from_secs(10);
```
kashif-m: nit: move to consts

crates/router/src/routes/payments.rs
```diff
@@ -927,33 +929,111 @@ pub async fn payments_update(
...
+    let integration_type = update_context::integration_type_from_headers(req.headers());
+
+    // Both inputs are known before the operation runs, so the decision — and therefore whether
+    // the enrichment inputs need cloning at all — is made once, here.
+    //
+    // Gated on merchant auth as well as the header: this route also accepts publishable-key +
+    // client-secret, and the enrichment runs the session core as `AuthFlow::Merchant` and skips
+    // client-secret validation on the list, so a client-authenticated caller must not be able to
+    // opt in with a header alone.
+    let enrich = integration_type.is_server() && auth_flow == api::AuthFlow::Merchant;
```
kashif-m: Make this concise

crates/router/src/routes/payments.rs
```diff
@@ -927,33 +929,111 @@ pub async fn payments_update(
...
+
+    if integration_type.is_server() && !enrich {
+        logger::warn!(
+            "server integration type requested on a client-authenticated request; \
+             returning the client response shape"
+        );
+    }
+
```
kashif-m: Not needed

crates/router/src/routes/payments.rs
```diff
@@ -927,33 +929,111 @@ pub async fn payments_update(
...
+                // The two enrichment cores are invoked directly rather than through `server_wrap`,
+                // so they take no lock of their own. That matters: `Flow::PaymentsUpdate` and
+                // `Flow::PaymentsSessionToken` both map to `ApiIdentifier::Payments`, so a nested
+                // `server_wrap` would ask for the very key this request already holds and
+                // deadlock until it gave up with `ResourceBusy`.
+                //
+                // Payments responses come back as `JsonWithHeaders`; `Json` is handled too so the
+                // enrichment does not silently skip if that ever changes.
+                let enrich_payment = |mut payment: payment_types::PaymentsResponse| async {
```
kashif-m: Make it concise

crates/router/src/core/payments/update_context.rs
```diff
@@ -0,0 +1,224 @@
...
+/// Wallets to mint session tokens for when the caller asks for the server shape.
+///
+/// Empty on purpose. The session core reads an empty list as "every eligible type": it keeps a
+/// payment method type when `requested_payment_method_types.contains(..) || ..is_empty()`, so an
+/// empty list yields a token for every `InvokeSdkClient`-capable method the merchant has enabled.
+/// Naming wallets explicitly here would silently exclude anything not on the list — Klarna SDK
+/// sessions today, and every type added later.
+fn requested_wallets() -> Vec<api_models::enums::PaymentMethodType> {
+    Vec::new()
+}
```
ShankarSinghC: The current session request allows us to pass specific wallets, which act as a filter for the wallets displayed in the session response.

If someone has integrated this feature, updating the payment would result in different behavior.

We can also evaluate removing wallets from the session request, since Payment Intent creation already has an allowed payment methods filter. Ideally, the session should make use of that filter instead.

crates/router/src/core/payments/update_context.rs
ShankarSinghC: There are already a few outgoing calls made during the session call, such as the Apple Pay session call, which currently has a timeout configured. I think it should use the same timeout that we use for external connector calls. Please verify this and use the same timeout here as well.

crates/router/src/core/payments/update_context.rs
prasunna09: yeah, currently there is no fallback if we remove wallets in request body, we do no use `allowed_payment_method_types` from intent body. we create all the wallets session tokens.

will create an issue for this, and raise this in seperate pr

ShankarSinghC (dismissed):

kashif-m (dismissed):

ShankarSinghC (dismissed):

prasunna09:
Screenshot 2026-09-07 at 9 08 09 PM
v2 is compiled locally!!

ShankarSinghC (approved):

ShankarSinghC (approved):

PR #14052, opened 2026-09-04
feat(payments): add new s2s route for session tokens

prasunna09:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

### Additional Changes

## Checklist


kashif-m (approved):

XyneSpaces:

**Summary:** Adds a straightforward server-to-server session-token route; only concern is weaker analytics tagging because `payment_id` lives on the path.

- **[should-fix]** `crates/api_models/src/events/payment.rs:438` — `PaymentsSessionServerRequest::get_api_event_type` returns `Miscellaneous` because the body lacks `payment_id`, making API-event correlation harder for a server-to-server call. Consider threading the path `payment_id` into the event type (e.g., via handler-level enrichment) so session-token analytics stay tied to the payment.

XyneSpaces:
Automated review findings:

[blocking] `crates/router/src/routes/app.rs:1038` — The new `/payments/{payment_id}/session_tokens` route is registered unconditionally, but `payments_connector_session_server` is annotated with `#[cfg(feature = "v1")]`. Building with `feature = "v2"` will fail to resolve the handler function. Wrap the new `.service(...)` block in `#[cfg(feature = "v1")]` to match the handler's feature gate.

[blocking] `crates/openapi/src/openapi.rs:82` — `routes::payments::payments_connector_session_server` is referenced in the `paths(...)` macro and `PaymentsSessionServerRequest` in the `components(schemas(...))` macro without v2 gating, while both symbols are compiled only for v1. Gate the new route/schema entries with `#[cfg(feature = "v1")]` or keep them out of the v2 OpenAPI bundle.

[blocking] `crates/router/src/types/api/payments.rs:33` — `PaymentsSessionServerRequest` is re-exported unconditionally, but the type itself in `api_models::payments` is gated behind `#[cfg(feature = "v1")]`. v2 builds will fail to resolve the import. Wrap the re-export in the same `#[cfg(feature = "v1")]` gate.

[should-fix] `crates/api_models/src/events/payment.rs:438` — `ApiEventMetric` for `PaymentsSessionServerRequest` returns `ApiEventsType::Miscellaneous` because the payment id arrives as a path parameter. Record the path-parameter payment id in the event type (e.g., by passing it into the metric context from the route handler) so session-token calls are associated with the correct payment.

XyneSpaces:
[blocking] Please add route-level coverage for the new server-to-server session-token endpoint. This PR adds `POST /payments/{payment_id}/session_tokens` with API-key auth and a new request body, but no router/integration test changes are included to verify auth, ownership, and the rejected-client-secret shape.

prasunna09:
since we have given out the wrapper for s2s merchant. this is no longer required

PR #14057, opened 2026-09-04
feat(observability): upload alert reports through the notifier

ThisIsMani:
Gives the notifier a route for uploading a file to a chat destination, so the R alerts service can
hand over its PNG and PDF reports instead of talking to the chat provider itself.

Resolves [#23300](https://github.com/juspay/hyperswitch-cloud/issues/23300). This merges **before**
[#23301](https://github.com/juspay/hyperswitch-cloud/issues/23301), the R port — that ordering is the
point of the ticket. Uploading is the only reason R would keep a chat credential after the port, so
with this in place `XYNE_APP_JWT`, `XYNE_CHANNEL_ID` and the squid proxy config all leave R with the
old sinks rather than surviving just to upload a PNG.

## The route

`POST /alerts/chat/upload/{destination}`, `multipart/form-data`, one file per call.

| Field | | |
|---|---|---|
| `file` | required | the bytes; filename from its `Content-Disposition` |
| `filename` | optional | overrides the part's own name |
| `title` | optional | display title |
| `comment` | optional | text posted with the file, in the destination's markup |
| `reply_to` | optional | the `message_id` from `/chat/notify`, to land inside that thread |

Answers `{"status": "delivered", "file_id": "..."}`, matching the notify routes: the status code says
whether the notifier worked, the body says whether the file arrived.

Multipart rather than base64 in the existing JSON body, which costs a third more on the wire and
pushes the whole file through `serde_json` as a `String`. It is also what R already speaks, so the R
side of this is a rename rather than new code.

## The protocol, and why not the simpler one

Three calls — `files.getUploadURLExternal`, a raw POST to the URL it names, then
`files.completeUploadExternal` — rather than the one-call `files.upload`. Both work on Xyne. Slack
retired `files.upload`, so building on it would leave the Slack backend's upload dead on arrival for
the sake of saving two round trips on a 50 KB PDF.

**Verified live against Xyne**, end to end through the running service: an alert posted to
`hyperswitch-alert-test`, the report uploaded into its thread, and `conversations.replies` confirming
the file message carries the right `thread_ts`. Unknown destination, missing key, no file part,
unknown field and an oversized body were all exercised against the running service too.

Three things the provider's source does not tell you, each of which shapes the code:

- **The raw leg answers `OK` in plain text**, not the `{ok}` envelope, so it needs its own response
  handling. A body that *does* parse as the envelope is a refusal; anything else at 2xx is accepted.
- **The `file_id` from step one is not the id you get back.** Xyne keys its pending-upload state on a
  UUID and returns a store-assigned id from step three, so the id is read from the response rather
  than reused.
- **`files.completeUploadExternal` returns no `ts`.** Sharing a file creates a message and the
  protocol does not name it. So a caller can thread an upload *under* something and can never thread
  anything under the upload. That is why `FileId` is a separate type from `MessageId`, and why the
  response has no `message_id` field rather than one that is always `null`.

## The credential on step two

Step two sends the bearer token to a URL that came out of a response body. It is attached only when
that URL is on the configured origin, and withheld otherwise.

Refusing off-origin URLs was the first shape of this and it was wrong: Slack's pre-signed URLs are
served from `files.slack.com` rather than its API root, so refusing would have broken the backend by
construction — and they need no credential anyway, which is exactly why withholding it costs nothing.

## Limits

The multipart body is read only **after** the API key is checked, so an unauthenticated caller cannot
make this process buffer megabytes. The cap (`chat.max_upload_bytes`, 25 MiB, one value for every
destination because it protects this process rather than any provider) is enforced **while
streaming**, not after — buffering the whole body and then measuring it would spend the memory the
limit exists to protect.

Oversized, empty, unnamed, duplicated and unknown-field bodies are all `400`s rather than refusals:
nothing was attempted, so there is no provider answer to report.

## Notes for review

- `ChatFile`'s `Debug` is hand-written, the only one in that module that is. `error-stack` prints
  `Debug` for everything attached to a report, and an alert report is a rendered picture of merchant
  ids and payment volumes.
- `post_message`'s send-and-status handling moved into a shared `execute`, so the three new calls do
  not each re-derive the 429 and non-2xx rules.
- Uploads are not retried by the transport and cannot be: `RequestBuilder::try_clone` returns `None`
  for a body it cannot replay. That is the property we want for a non-idempotent call, and it falls
  out for free.
- `SlackClient::upload_file` shares the implementation but is **unverified** — no Slack destination
  is configured anywhere yet.

Tests: 42 in `observability` (9 new, covering the multipart reader), plus the upload conversions in
`external_services`. The one failing test there,
`utils::tests::test_deserialize_hashset_inner_empty_string`, fails identically on a clean worktree
and is unrelated.

https://claude.ai/code/session_017dhCVLEf2oP4zxTbiRux6L

### DIFF
--- a/crates/external_services/src/chat_service.rs
+++ b/crates/external_services/src/chat_service.rs
@@ -35,6 +35,14 @@ pub type ChatResult<T> = CustomResult<T, ChatError>;
 pub trait ChatClient: Send + Sync + std::fmt::Debug {
     /// Post a message, returning the id of the message that was created.
     async fn post_message(&self, message: ChatMessage) -> ChatResult<MessageId>;
+
+    /// Upload a file, returning the id the backend stored it under.
+    ///
+    /// A sibling of [`ChatClient::post_message`] rather than a field on [`ChatMessage`], because
+    /// it is a different endpoint with a different result. It returns a [`FileId`] and **not** a
+    /// [`MessageId`]: the upload creates a message, but the backends do not agree that you may
+    /// know its id, so nothing can be threaded under an upload.
+    async fn upload_file(&self, file: ChatFile) -> ChatResult<FileId>;
 }
 
 /// Identifies a message that a backend has accepted.
@@ -128,6 +136,145 @@ impl ChatMessage {
     }
 }
 
+/// Identifies a file a backend has stored.
+///
+/// Deliberately *not* a [`MessageId`]. Uploading produces a file, and the message carrying it is a
+/// side effect the backend does not always name — Xyne's `files.completeUploadExternal` returns no
+/// `ts` at all. Conflating the two would promise threading that the upload path cannot deliver.
+///
+/// Opaque, and not the id the upload was started with: a backend is free to remap it between
+/// reserving an upload and completing it, and Xyne does.
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+pub struct FileId(String);
+
+impl FileId {
+    /// Wrap a backend's file id.
+    pub fn new(value: impl Into<String>) -> Self {
+        Self(value.into())
+    }
+
+    /// The id as the backend spelled it.
+    pub fn as_str(&self) -> &str {
+        &self.0
+    }
+}
+
+/// A file to upload, and how it should appear when it lands.
+///
+/// Fields are private and reached through a constructor and chained setters, matching
+/// [`ChatMessage`]. Unlike a message, most of what can be said about a file is optional, so this
+/// one takes the setters: four optional fields would otherwise be sixteen constructors.
+#[derive(Clone)]
+pub struct ChatFile {
+    filename: String,
+    content_type: Option<String>,
+    bytes: Vec<u8>,
+    title: Option<String>,
+    comment: Option<String>,
+    reply_to: Option<MessageId>,
+}
+
+impl ChatFile {
+    /// A file to upload, named and with its contents.
+    pub fn new(filename: impl Into<String>, bytes: Vec<u8>) -> Self {
+        Self {
+            filename: filename.into(),
+            content_type: None,
+            bytes,
+            title: None,
+            comment: None,
+            reply_to: None,
+        }
+    }
+
+    /// Declare the file's media type. Without one the backend infers from the filename.
+    #[must_use]
+    pub fn with_content_type(mut self, content_type: impl Into<String>) -> Self {
+        self.content_type = Some(content_type.into());
+        self
+    }
+
+    /// A display title. Xyne echoes it and stores the filename regardless; Slack honours it.
+    #[must_use]
+    pub fn with_title(mut self, title: impl Into<String>) -> Self {
+        self.title = Some(title.into());
+        self
+    }
+
+    /// Text posted alongside the file, in the destination's markup.
+    #[must_use]
+    pub fn with_comment(mut self, comment: impl Into<String>) -> Self {
+        self.comment = Some(comment.into());
+        self
+    }
+
+    /// Land the file inside the thread of an existing message.
+    #[must_use]
+    pub fn reply_under(mut self, message_id: MessageId) -> Self {
+        self.reply_to = Some(message_id);
+        self
+    }
+
+    /// The filename the backend should store.
+    pub fn filename(&self) -> &str {
+        &self.filename
+    }
+
+    /// The declared media type, if any.
+    pub fn content_type(&self) -> Option<&str> {
+        self.content_type.as_deref()
+    }
+
+    /// The file's contents.
+    pub fn bytes(&self) -> &[u8] {
+        &self.bytes
+    }
+
+    /// How many bytes there are. Backends want this before they will accept the upload.
+    pub fn len(&self) -> usize {
+        self.bytes.len()
+    }
+
+    /// Whether there is nothing to upload. Backends refuse an empty file.
+    pub fn is_empty(&self) -> bool {
+        self.bytes.is_empty()
+    }
+
+    /// The display title, if one was set.
+    pub fn title(&self) -> Option<&str> {
+        self.title.as_deref()
+    }
+
+    /// The accompanying text, if any.
+    pub fn comment(&self) -> Option<&str> {
+        self.comment.as_deref()
+    }
+
+    /// The message this file lands under, if any.
+    pub fn reply_target(&self) -> Option<&MessageId> {
+        self.reply_to.as_ref()
+    }
+}
+
+/// Hand-written so the contents cannot reach a log line.
+///
+/// The one place in this module where `Debug` is not derived. An alert report is a rendered
+/// picture of merchant ids and payment volumes, and `error-stack` prints `Debug` for every value it
+/// attaches, so a derive here would put the whole file into the log stream the first time an upload
+/// failed. Sizes are what diagnosis needs and they are all that appears.
+impl std::fmt::Debug for ChatFile {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("ChatFile")
+            .field("filename", &self.filename)
+            .field("content_type", &self.content_type)
+            .field("bytes", &format_args!("{} bytes", self.bytes.len()))
+            .field("title", &self.title)
+            .field("comment", &self.comment.as_ref().map(|_| "<set>"))
+            .field("reply_to", &self.reply_to)
+            .finish()
+    }
+}
+
 /// Errors raised when posting a chat message.
 #[derive(Debug, thiserror::Error)]
 pub enum ChatError {
@@ -168,6 +315,24 @@ pub enum ChatError {
     /// id minted by a different backend.
     #[error("The message id supplied cannot thread a reply on this chat provider")]
     IncompatibleReplyTarget,
+
+    /// The upload completed but the provider named no id for the stored file.
+    ///
+    /// The file is up. Retrying would store it twice, so this is reported apart from the failures.
+    #[error("Chat provider stored the file without returning a file id")]
+    MissingFileId,
+
+    /// The provider's upload URL pointed somewhere other than the provider.
+    ///
+    /// The multi-step upload takes a URL out of a response body and then sends the credential to
+    /// it. A URL on another origin is refused rather than followed: a compromised or spoofed
+    /// response would otherwise be handed a working bot token.
+    #[error("Chat provider returned an upload URL on an unexpected origin")]
+    UntrustedUploadUrl,
+
+    /// The provider rejected the bytes at the upload URL, or never took them.
+    #[error("Chat provider did not accept the file contents")]
+    UploadRejected,
 }
 
 /// Why a provider refused a message, in vocabulary no single backend owns.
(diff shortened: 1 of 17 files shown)


crates/external_services/src/chat_service.rs
```diff
@@ -128,6 +136,145 @@ impl ChatMessage {
...
+/// Identifies a file a backend has stored.
+///
+/// Deliberately *not* a [`MessageId`]. Uploading produces a file, and the message carrying it is a
+/// side effect the backend does not always name — Xyne's `files.completeUploadExternal` returns no
+/// `ts` at all. Conflating the two would promise threading that the upload path cannot deliver.
+///
+/// Opaque, and not the id the upload was started with: a backend is free to remap it between
+/// reserving an upload and completing it, and Xyne does.
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+pub struct FileId(String);
```
ThisIsMani: Are we assuming that we will get FileId in all the chat clients? Should be a enum like in MessageId imo.

crates/external_services/src/chat_service/slack.rs
```diff
@@ -94,6 +94,18 @@ impl ChatClient for SlackClient {
...
     }
+
+    /// Uploads use `files.getUploadURLExternal` and `files.completeUploadExternal` rather than the
+    /// retired one-call `files.upload`, so this leg is the same protocol Xyne serves.
+    ///
+    /// **Verified against Xyne only.** No Slack destination is configured anywhere yet, so the
+    /// shared implementation is proven on one of the two backends it serves. The difference to
+    /// expect is that Slack's pre-signed upload URL is on `files.slack.com` rather than its API
+    /// root, so the bytes go up without the bot token — which is what Slack documents, and what
+    /// the origin rule in `Endpoint::send_bytes` produces on its own.
```
ThisIsMani: Can we remove this.

crates/external_services/src/chat_service/slack_compatible.rs
```diff
@@ -10,20 +10,35 @@
...
+const FILES_GET_UPLOAD_URL: &str = "files.getUploadURLExternal";
+
+/// Step three of an upload: share the stored file into the channel.
+///
+/// Step two has no method name — it posts to a URL step one returns.
+const FILES_COMPLETE_UPLOAD: &str = "files.completeUploadExternal";
+
+/// Sent when a file declares no media type. Both backends sniff the filename when they get this.
+const DEFAULT_UPLOAD_CONTENT_TYPE: &str = "application/octet-stream";
+
```
ThisIsMani: We should be having this some where no? Mime headers or something?

crates/external_services/src/chat_service/slack_compatible.rs
```diff
@@ -190,39 +219,228 @@ impl Endpoint {
...
+    /// **the token is attached only when the URL is on the origin we were configured to talk to**,
+    /// and withheld otherwise. Refusing off-origin URLs outright was the first shape of this and it
+    /// was wrong: Slack's pre-signed URLs are served from `files.slack.com` rather than its API
+    /// root, so refusing them would break the backend by construction. They also need no
+    /// credential, which is exactly why withholding it costs nothing and forwarding it would be
+    /// the only real loss.
+    ///
+    /// This leg does not answer with the `{ok}` envelope on success — Xyne replies with the two
+    /// characters `OK` — so a body that *does* parse as a refusal is the only failure it can
+    /// report at HTTP 200.
```
ThisIsMani: Are we saying xyne is diverging from slack impl?

crates/external_services/src/chat_service.rs
ThisIsMani: Done in `0b03ab6cb4` — non-exhaustive enum with `FileId::slack_compatible` / `as_slack_compatible`, mirroring `MessageId`.

Worth saying that the argument is weaker here than for `MessageId`, and the doc comment now says so rather than implying they are the same case. `MessageId` has to be an enum because you hand it *back* to thread a reply, so a mismatched id is a round trip the type system can catch. Nothing hands a file id back to anything — it goes to the caller and stops.

What it does buy is the other half: the first backend whose file id is not a bare string, or is not interchangeable with a Slack-compatible one, adds a variant instead of forcing the type open. Cheap now, so worth taking.

crates/external_services/src/chat_service/slack.rs
ThisIsMani: Removed in `0b03ab6cb4`.

One line of it was load-bearing, so I moved it to the module docs rather than dropping it: nothing configures a Slack destination, so the shared upload path is proven against Xyne alone. That matters more now than when I wrote it — see the Codex thread about capability URLs reaching the logs, which is a Slack-only problem.

crates/external_services/src/chat_service/slack_compatible.rs
ThisIsMani: Yes. `mime::APPLICATION_OCTET_STREAM` in `0b03ab6cb4`; `mime` was already in the workspace and is now a dependency of `external_services`.

crates/external_services/src/chat_service/slack_compatible.rs
ThisIsMani: No, and the comment was badly worded — fixed in `0b03ab6cb4`.

Neither backend returns the `{ok}` envelope on that leg, because it posts to a storage endpoint rather than an API method. Slack documents a plain 200 from its file store; Xyne answers with the two characters `OK`. I named only Xyne because that is what I verified, which made it read as a divergence.

Xyne *does* diverge from Slack, just elsewhere, and the comment now names those where they bite:

- it requires the credential on the upload leg, where Slack's pre-signed URL does not
- `files.completeUploadExternal` returns `url_private` as a path relative to its storage root, where Slack returns an absolute URL
- the `title` you send is echoed in the response but not persisted; `conversations.replies` shows the filename

crates/external_services/src/chat_service/slack_compatible.rs
```diff
@@ -190,39 +219,228 @@ impl Endpoint {
...
+        if file.is_empty() {
+            Err(ChatError::Rejected {
+                reason: ChatErrorReason::Other(NO_FILE_DATA.to_owned()),
+            })
+            .attach_printable("refusing to upload an empty file")?
+        }
+
+        let reservation = self.reserve_upload(&file).await?;
+        self.send_bytes(&reservation.upload_url, &file).await?;
+        self.complete_upload(&file, &reservation.file_id).await
```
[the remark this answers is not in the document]
ThisIsMani: Correct, and my comment asserted the opposite. Fixed in `0b03ab6cb4`; tracked as [#23318](https://github.com/juspay/hyperswitch-cloud/issues/23318).

I had reasoned from the comment at `http_client.rs:172` — "we cannot clone the request type, because it has Form trait which is not cloneable" — which is true of `FormData` and not of this. `RequestContent::RawBytes` is a `Vec<u8>`, becomes `client.body(payload)`, and reqwest clones a buffered body happily. All three legs retry.

Leg by leg, when the connection drops after the provider acted:

- **reserve** — orphaned reservation, expires on the provider's TTL. Harmless.
- **raw upload** — refused `invalid_arguments`; the reservation is no longer `pending`.
- **complete** — refused for the same reason; the provider drops its upload state once the file is shared.

So the outcome is a success **reported as a refusal** — the file is in the channel and the caller is told it is not — rather than a file posted twice. That distinction is why I have not blocked the PR on it: the duplicate is the failure that would matter, and R does not retry on a refusal.

Not fixed here because opting out means changing `send_request`, which every connector in the workspace shares. That decision belongs outside a ticket about an upload route. #23318 also asks whether any other caller is relying on the same false premise about `RawBytes`.

crates/external_services/src/chat_service/slack_compatible.rs
```diff
@@ -190,39 +219,228 @@ impl Endpoint {
...
+                tag = "chat_upload_offsite",
+                origin = %parsed.origin().ascii_serialization(),
+                "sending file contents without the credential: the upload URL is not on the \
+                 configured origin"
+            );
+        }
+
+        let request = RequestBuilder::new()
+            .method(Method::Post)
+            .url(upload_url)
```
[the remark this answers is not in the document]
ThisIsMani: Correct. Not fixed here; tracked as [#23319](https://github.com/juspay/hyperswitch-cloud/issues/23319), with a warning added to `slack`'s module docs in `0b03ab6cb4`.

Confirmed the mechanism: `send_request` logs `?request`, and `Request::url` is a plain `String` under a derived `Debug`. Headers are `Maskable` and the body is redacted; the URL is not.

On severity, the two backends differ and the split is worth recording:

- **Xyne**, the only configured destination, leaks nothing usable. Its `_upload/:fileId` endpoint authenticates (`slackRawBodyAuthenticateApp`), so the URL alone opens nothing — which is also why the client sends the bearer token with it.
- **Slack** is where you are right. Its pre-signed URL *is* the capability, and it is off-origin so the token is deliberately withheld — meaning the URL is the only thing protecting that upload, and it is in the logs.

So this is latent rather than live, and it goes live silently the day someone configures a Slack destination. Masking it means changing a shared transport, so same reasoning as the retry thread. #23319 asks for a sweep for other credential-bearing URLs rather than a fix for this one call site.

crates/external_services/src/chat_service.rs
```diff
@@ -128,6 +136,158 @@ impl ChatMessage {
...
+/// forcing this type open. Non-exhaustive, so match with a wildcard arm.
+///
+/// The value is also not the id the upload was started with: a backend is free to remap it between
+/// reserving an upload and completing it, and Xyne does.
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+#[non_exhaustive]
+pub enum FileId {
+    /// An id minted by a Slack-compatible backend: opaque, and only meaningful to the backend that
+    /// issued it. Slack spells these `F...`; Xyne returns its own store id.
+    SlackCompatible(String),
```
ThisIsMani: Don't name this please. Name it Id(String) or something. This is interface agnostic code.

crates/external_services/src/chat_service.rs
```diff
@@ -128,6 +136,158 @@ impl ChatMessage {
...
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("ChatFile")
+            .field("filename", &self.filename)
+            .field("content_type", &self.content_type)
+            .field("bytes", &format_args!("{} bytes", self.bytes.len()))
+            .field("title", &self.title)
+            .field("comment", &self.comment.as_ref().map(|_| "<set>"))
+            .field("reply_to", &self.reply_to)
+            .finish()
+    }
```
ThisIsMani: Add Secret<String> instead of custom impl. Should use derive.

XyneSpaces:

**Summary:** The chat file-upload path is generally well designed (streaming size cap, double authentication, clear refusal semantics). Two cleanup items should be addressed before merge.

- **[should-fix]** `crates/observability/src/routes/notify.rs:59` — The new async route handler `chat_upload` and the helpers it calls (`read_attachment`, `upload_chat_file`, `ChatClientNotifier::upload`) lack `#[instrument(skip_all)]`. Add `#[instrument(skip_all)]` to `chat_upload`, `read_attachment`, and `upload_chat_file` so request tracing matches the rest of the observability plane.

- **[should-fix]** `crates/external_services/src/chat_service/slack_compatible.rs:678` — `complete_upload` uses `response.files.unwrap_or_default()` on the provider response, which silently treats a missing `files` array as empty. Replace `unwrap_or_default()` with an explicit chain such as `response.files.and_then(...).map(...).ok_or(ChatError::MissingFileId)?` so a `None` `files` field still produces `MissingFileId` without the unwrap.

XyneSpaces:
Automated review findings:

[should-fix] `crates/observability/src/routes/notify.rs:190` — The resolved `filename` (from the part header or the explicit `filename` field) is not sanitized for path separators, `..` sequences or control characters before being used. Add a filename-safety check in `read_attachment` that rejects `/`, `\`, NUL and `..` so a malicious multipart payload cannot carry a traversal-like name into the chat backend.

[should-fix] `crates/observability/src/routes/notify.rs:100` — `read_attachment` buffers the whole accepted file into a `Vec<u8>`. The config comment correctly notes that the upload is buffered in-process, but the route doc-comment says the cap is enforced "while streaming" which can be read as implying the body is not fully buffered. Align documentation with reality, and consider an upper-bound sanity check on `max_upload_bytes` during `ChatSettings::validate` to avoid accidental multi-gigabyte caps.

Pre-existing (not introduced by this PR): The surrounding implementation already uses `Secret<Vec<u8>>` and origin-aware credential withholding.

PR #14067, opened 2026-09-04
feat(observability): upload alert reports through notifier

ThisIsMani:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [x] Documentation
- [ ] CI/CD

## Description

Adds `POST /alerts/chat/upload/{destination}`, accepting one file per multipart request with optional `filename`, `title`, `comment`, and `reply_to` fields.

The shared Slack-compatible chat client now performs the current external upload flow:

1. `files.getUploadURLExternal`
2. raw upload to the returned URL
3. `files.completeUploadExternal`

Uploads can be threaded beneath a message returned by the existing chat notification route. The response distinguishes file identifiers from message identifiers and follows the existing notifier outcome contract:

```json
{ "status": "delivered", "file_id": "..." }
```

The implementation also:

- limits multipart bodies through `chat.max_upload_bytes` (25 MiB by default)
- rejects missing, empty, duplicate, and unknown-field multipart requests
- redacts file contents and metadata from derived debug output
- delegates API and raw-upload headers to each concrete backend, so Slack does not send its bot token to pre-signed storage while Xyne sends the app JWT required by its upload endpoint
- treats an accepted upload without a returned file ID as delivered, preventing duplicate retries

This is a fresh, isolated implementation for comparison with the existing implementation in [#14057](https://github.com/juspay/hyperswitch/pull/14057). The two alternatives should not be merged together.

### Additional Changes

Configuration changes:

- [`config/observability.toml`](https://github.com/juspay/hyperswitch/blob/23300-notifier-file-upload-fresh/config/observability.toml)
- [`crates/observability/src/settings.rs`](https://github.com/juspay/hyperswitch/blob/23300-notifier-file-upload-fresh/crates/observability/src/settings.rs)

## Motivation and Context

The R alerts service renders PNG charts and PDF reports but currently uploads them directly to the chat provider. Moving upload delivery behind the Notifier lets R hand off those artifacts without retaining chat credentials or duplicating destination configuration.

Resolves [Upload alert reports through the Notifier](https://github.com/juspay/hyperswitch-cloud/issues/23300).

## How did you test it?

Automated validation:

- `nix shell nixpkgs#perl --command cargo test -p observability --lib --tests`
- `nix shell nixpkgs#perl --command cargo test -p external_services --features v1 chat_service`
- `nix shell nixpkgs#perl --command cargo clippy -p observability --tests --no-deps -- -D warnings`
- `nix shell nixpkgs#perl --command cargo clippy -p external_services --features v1 --tests --no-deps -- -D warnings`

Live verification against Xyne through a locally running Observability service:

- posted a notification to `hyperswitch-alert-test`
- uploaded a PNG beneath the notification thread
- uploaded a PDF beneath the same thread
- all three requests returned `status: delivered` with provider identifiers

## Checklist


crates/external_services/src/chat_service.rs
```diff
@@ -35,6 +35,89 @@ pub type ChatResult<T> = CustomResult<T, ChatError>;
...
     /// Post a message, returning the id of the message that was created.
     async fn post_message(&self, message: ChatMessage) -> ChatResult<MessageId>;
+
+    /// Upload a file and optionally share it under an existing message.
+    async fn upload_file(&self, file: ChatFile) -> ChatResult<FileId>;
+}
+
+/// Identifies a file that a backend has accepted.
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+pub struct FileId(String);
```
ThisIsMani: Let's make this an enum similar to MessageId, so that we we can easily add new implemenations which aren't slack compatible later.

crates/external_services/src/chat_service.rs
```diff
@@ -35,6 +35,89 @@ pub type ChatResult<T> = CustomResult<T, ChatError>;
...
+    pub fn with_comment(mut self, comment: Option<String>) -> Self {
+        self.comment = comment;
+        self
+    }
+
+    /// Share the file under an existing message.
+    pub fn with_reply_to(mut self, reply_to: Option<MessageId>) -> Self {
+        self.reply_to = reply_to;
+        self
+    }
```
ThisIsMani: Let's not use builder pattern and introduce muts please.

crates/external_services/src/chat_service/slack_compatible.rs
```diff
@@ -203,6 +205,158 @@ impl Endpoint {
...
+
+        let prepare_url = self.method_url(FILES_GET_UPLOAD_URL);
+        let prepare_body = self
+            .send(
+                &prepare_url,
+                RequestContent::Json(Box::new(GetUploadUrlPayload {
+                    filename: file.filename().to_owned(),
+                    length: file.bytes().len(),
+                })),
+                "application/json",
```
ThisIsMani: Let's use existing headers already, let's not define this manually.

crates/external_services/src/chat_service/slack_compatible.rs
```diff
@@ -223,6 +377,66 @@ impl Endpoint {
     }
 }
 
+fn reject_if_needed(ok: bool, error: Option<SlackErrorCode>) -> ChatResult<()> {
+    if ok {
+        return Ok(());
```
ThisIsMani: Let's not use early returns. Either you aren't decomposing the function correctly or you could use ?.

crates/external_services/src/chat_service/slack_compatible.rs
```diff
@@ -203,6 +205,158 @@ impl Endpoint {
...
+            .attach_printable("files.getUploadURLExternal returned no upload_url")?;
+        let pending_file_id = prepared
+            .file_id
+            .filter(|id| !id.is_empty())
+            .ok_or(ChatError::UnreadableResponse)
+            .attach_printable("files.getUploadURLExternal returned no file_id")?;
+
+        // Xyne returns an on-origin URL that requires its app JWT. Slack returns a pre-signed URL
+        // on files.slack.com; sending the bot credential there would leak it and is unnecessary.
+        let authorize_upload = upload_url.origin() == self.base_url.origin();
```
ThisIsMani: So, xyne and slack behaviour is a bit different here?

crates/external_services/src/chat_service/slack_compatible.rs
```diff
@@ -203,6 +205,158 @@ impl Endpoint {
...
+        let upload_body = self
+            .send(
+                upload_url.as_str(),
+                RequestContent::RawBytes(file.bytes().to_vec()),
+                "application/octet-stream",
```
ThisIsMani: Same here.

crates/external_services/src/chat_service/slack_compatible.rs
ThisIsMani, line 1: Now that this file is growing, it's better to organize this code properly. We can split it into smaller files probably. That can be a separate PR though.

crates/external_services/src/chat_service/slack_compatible.rs
```diff
@@ -203,6 +205,158 @@ impl Endpoint {
...
+            .files
+            .unwrap_or_default()
+            .into_iter()
+            .find_map(|file| file.id.filter(|id| !id.is_empty()))
+            .map(FileId::new)
+            .ok_or(ChatError::MissingFileId)
+            .attach_printable("the upload completed but the response named no resulting file")
+    }
+
+    async fn send(
```
ThisIsMani, line 263: Could this be used for send message also?

crates/observability/src/domain/notifier/chat.rs
```diff
@@ -35,6 +35,35 @@ pub struct ChatNotification {
...
+        formatter
+            .debug_struct("ChatFileUpload")
+            .field("bytes", &format_args!("<{} bytes>", self.bytes.len()))
+            .field("filename", &self.filename)
+            .field("title", &self.title)
+            .field("comment", &self.comment)
+            .field("reply_to", &self.reply_to)
+            .finish()
+    }
+}
```
ThisIsMani: Use secret instead if you think something shouldn't be logged. Let's not write custom impls.

crates/observability/src/routes/notify.rs
```diff
@@ -35,6 +45,135 @@ pub async fn chat(
     .await
 }
 
+/// `POST /alerts/chat/upload/{destination}`.
+pub async fn chat_upload(
+    state: web::Data<AppState>,
+    request: HttpRequest,
+    destination: web::Path<String>,
+    payload: Multipart,
```
ThisIsMani: Could you check other Multipart impls in code, check theme upload API, let's clean up this impl.

crates/observability/src/routes/notify.rs
```diff
@@ -35,6 +45,135 @@ pub async fn chat(
...
+) -> HttpResponse {
+    // Multipart is a stream. Authenticate before consuming it so an unauthenticated caller cannot
+    // make this process buffer a large file. `server_wrap` checks again after parsing to preserve
+    // the required-auth argument on every core invocation.
+    if let Err(error) = auth::InternalApiKeyAuth.authenticate(request.headers(), state.get_ref()) {
+        return ErrorSwitch::<ApiErrorResponse>::switch(error.current_context()).error_response();
+    }
```
ThisIsMani: Let's not do this. It's fine, let wrap() handle this.

XyneSpaces:
Automated review findings:

[blocking] `crates/external_services/src/chat_service.rs:68` — `ChatFile` derives `Debug`, so error-stack attachments or any `{:?}` logging will print raw file bytes, filename, title and comment. Alert reports can contain merchant/payment data; replace the derived `Debug` with a hand-written impl that prints only the byte length and masks the comment/title.

[should-fix] `crates/observability/src/routes/notify.rs:115` — The multipart `filename` is trimmed and checked for emptiness, but not validated for path separators, `..` sequences or control characters. Reject filenames containing `/`, `\`, NUL, or `..` before passing them to the notifier.

[should-fix] `crates/observability/src/routes/notify.rs:170` — `parse_upload` buffers the entire accepted multipart body into `Vec<u8>` while the README claims the cap is enforced "while the multipart stream is consumed". Either stream the body directly to the provider or update the documentation/comment to match the actual buffering behavior.

[should-fix] `crates/observability/Cargo.toml:32` — The full `futures = "0.3.31"` crate is added only to use `StreamExt`. Depend on `futures-util` instead, which is already common in the workspace and sufficient for `StreamExt`.

Pre-existing (not introduced by this PR): This is a fresh implementation in the observability service; no shared production flows are touched.

crates/observability/src/domain/notifier/chat.rs
```diff
@@ -174,6 +229,10 @@ fn classify(error: &ChatError) -> Verdict {
         ChatError::InvalidConfiguration(_) => {
             Verdict::Failed(|_destination| ObservabilityError::InternalServerError)
         }
+
+        ChatError::MissingFileId => {
+            Verdict::Failed(|destination| ObservabilityError::ProviderUnavailable { destination })
+        }
```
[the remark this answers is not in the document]
ThisIsMani: Fixed in 816913eba9. `MissingFileId` now follows `MissingMessageId`: the upload is reported as delivered with no `file_id`, so callers are not invited to retry an already accepted upload. Added classifier coverage.

crates/external_services/src/chat_service.rs
ThisIsMani: Fixed in 816913eba9. `FileId` is now a non-exhaustive enum with a `SlackCompatible` variant and a family-specific accessor, matching the extensibility shape of `MessageId`.

crates/external_services/src/chat_service.rs
ThisIsMani: Fixed in 816913eba9. `ChatFile::new` now takes the complete upload state in one call; the builders and mutable assembly are gone.

crates/external_services/src/chat_service/slack_compatible.rs
ThisIsMani: Fixed in 816913eba9. Content types now use `mime::APPLICATION_JSON` / `mime::APPLICATION_OCTET_STREAM`, and request execution is shared with message sending.

crates/external_services/src/chat_service/slack_compatible.rs
ThisIsMani: Fixed in 816913eba9. The early return is gone; the success/refusal split is expressed as a match.

crates/external_services/src/chat_service/slack_compatible.rs
ThisIsMani: The wire protocol is shared, but the URL returned by step one differs in practice: Xyne returns an on-origin URL requiring its app JWT, while Slack returns an off-origin pre-signed URL that must not receive the bot token. The behavior is therefore based on the returned URL origin rather than a backend branch, with tests for both cases.

crates/external_services/src/chat_service/slack_compatible.rs
ThisIsMani: Fixed in 816913eba9. The raw leg uses the existing `mime::APPLICATION_OCTET_STREAM` constant.

crates/external_services/src/chat_service/slack_compatible.rs
ThisIsMani: Agreed. I kept the file split out of this ticket as suggested; this revision reduces the added request-execution duplication in place.

crates/external_services/src/chat_service/slack_compatible.rs
ThisIsMani: Fixed in 816913eba9. `post_message` now uses the same `send` path as all three upload calls, including the shared 429/non-2xx handling.

crates/observability/src/domain/notifier/chat.rs
ThisIsMani: Fixed in 816913eba9. File bytes, filename, title and comment are now `Secret` fields with derived `Debug`; both custom debug implementations were removed. Added a regression test proving content and metadata do not render.

crates/observability/src/routes/notify.rs
ThisIsMani: Fixed in 816913eba9. The route now uses a derived `MultipartForm` model like theme upload, with deny-unknown/deny-duplicate attributes and a configured total/memory cap.

crates/observability/src/routes/notify.rs
ThisIsMani: Fixed in 816913eba9. The manual pre-authentication and double check are gone; the route delegates authentication to `server_wrap` like the other handlers.

ThisIsMani:
Addressed the review feedback in `816913eba9`:

- protected file bytes and metadata with `Secret` + derived `Debug`
- rejected path-like/control-character filenames
- clarified that accepted multipart data is buffered while limits are accounted per chunk
- replaced the manual multipart loop with the standard derived `MultipartForm` extractor (so no direct futures dependency remains)
- preserved accepted uploads that return no final file id as `delivered`
- also incorporated the human review suggestions around `FileId`, constructor shape, shared send handling, MIME constants, and auth handling

Live verification also passed against Xyne through the running service: one notification plus threaded PNG and PDF uploads all returned `delivered` with provider ids.

crates/external_services/src/chat_service.rs
```diff
@@ -35,6 +36,111 @@ pub type ChatResult<T> = CustomResult<T, ChatError>;
...
+            Some("merchant-1234 report".to_owned()),
+            Some("payment volume 4201".to_owned()),
+            None,
+        );
+        let rendered = format!("{file:?}");
+
+        for sensitive in ["merchant-1234", "payment volume", "4201"] {
+            assert!(!rendered.contains(sensitive));
+        }
+    }
```
ThisIsMani: Let's remove this.

crates/external_services/src/chat_service/slack_compatible.rs
ThisIsMani: Can we offload the header creation to xyne / slack implementation then?

ThisIsMani:
Follow-up to the `FileId` review discussion: updated in `84d6d7e1ee`. `FileId` remains an extensible enum, but its current variant is representation-oriented (`Identifier`) rather than backend-oriented (`SlackCompatible`), mirroring `MessageId::Ts`. The shared interface no longer names a provider implementation.

crates/observability/src/core/notifier.rs
```diff
@@ -39,6 +40,46 @@ pub async fn notify_chat(
...
+        .await
+}
+
+fn is_usable_filename(filename: &str) -> bool {
+    !filename.trim().is_empty()
+        && !filename.contains("..")
+        && !filename.contains('/')
+        && !filename.contains('\\')
+        && !filename.chars().any(char::is_control)
+}
```
ThisIsMani: Should we validate this? I think the external service will anyways throw error and we can't have a general impl for this and assume every provider will have same restrictions.

crates/external_services/src/chat_service.rs
ThisIsMani: Removed the additional `ChatFile` debug regression test in 8e4a7a2767.

crates/external_services/src/chat_service/slack_compatible.rs
ThisIsMani: Done in 8e4a7a2767. Xyne and Slack now construct their own API/raw-upload header sets and pass them into the shared endpoint. Xyne attaches its app JWT to both legs; Slack attaches its bot token only to API calls and sends no credential to storage. Added provider-level tests for both behaviors.

crates/observability/src/core/notifier.rs
ThisIsMani: Agreed. Removed the shared path/control-character filename policy in 8e4a7a2767 and left provider-specific restrictions to the external service. The API only retains the general contract checks that a file and non-empty filename are present.

PR #14071, opened 2026-09-04
fix(common_utils): fix `PhoneNumberStrategy` masking underflow and missing asterisks

### LINKED ISSUE #14200: [BUG] PhoneNumberStrategy masking underflows on short values and never prints the asterisks
### Bug Description

`PhoneNumberStrategy::fmt` (`crates/common_utils/src/pii.rs`), the `Debug` / `Display` masking for `Secret<_, PhoneNumberStrategy>`, has two problems:

1. `val_str.get(val_str.len() - 4..)` underflows for values shorter than four bytes.
2. The `if let Some(val_str) = ...` binding shadows the full value, so `"*".repeat(val_str.len() - 4)` is computed on the four-character suffix and always yields zero asterisks.

### Expected Behavior

`+15555555555` is rendered as `********5555`, as the comment (and `EmailStrategy` next to it) intend; a value shorter than four characters is rendered without panicking.

### Actual Behavior

`+15555555555` is rendered as `5555` (the masked digits are simply dropped from the output, nothing hides them). For short values, debug builds panic with `attempt to subtract with overflow` while formatting; release builds fall through to the generic `WithType` output. `Secret<String, PhoneNumberStrategy>` is also used for `payment_processor_token` in the revenue-recovery backfill models and workflow, where a short token is plausible.

### Steps To Reproduce

```rust
use common_utils::pii::PhoneNumberStrategy;
use masking::Secret;
let s: Secret<String, PhoneNumberStrategy> = Secret::new("+15555555555".to_string());
assert_eq!(format!("{s:?}"), "********5555"); // fails: "5555"
let short: Secret<String, PhoneNumberStrategy> = Secret::new("12".to_string());
let _ = format!("{short:?}"); // debug build: panic
```

### Context For The Bug

Found while running a small property-style harness over the pure helpers in `common_utils`.

### Environment

`main` at 9e515b8; independent of OS or deployment.

Yes, I am willing to submit a PR! (#14071)

edubraqd:
## Type of Change
- [x] Bugfix

## Description
`PhoneNumberStrategy::fmt` (the `Debug`/`Display` masking used for `Secret<_, PhoneNumberStrategy>`)
has two problems:

1. `val_str.get(val_str.len() - 4..)` underflows for values shorter than four bytes. Debug builds
   panic with `attempt to subtract with overflow` while formatting; release builds silently fall
   through to the generic `WithType` output. `Secret<String, PhoneNumberStrategy>` is also used for
   `payment_processor_token` in the revenue-recovery backfill models and in the revenue-recovery
   workflow, where a short token is plausible.
2. The `if let Some(val_str) = ...` binding shadows the full value, so `"*".repeat(val_str.len() - 4)`
   is computed on the four-character suffix and always yields zero asterisks: `+15555555555` was
   rendered as `5555` instead of `********5555` as the comment (and the `EmailStrategy` next to it)
   intend.

This PR uses `checked_sub` for the suffix index, binds the suffix under a separate name so the
asterisk count is computed on the full value, and adds a unit test for the regular, exact-length,
too-short and empty cases.

### Additional Changes

## Motivation and Context
Found while running a small property-style harness over the pure helpers in `common_utils`.
Reproduction on `main` (debug build):

```rust
use common_utils::pii::PhoneNumberStrategy;
let secret = hyperswitch_masking::Secret::<String, PhoneNumberStrategy>::new("12".to_string());
let _ = format!("{secret:?}");
// thread panicked at 'attempt to subtract with overflow', crates/common_utils/src/pii.rs:47
```

## How did you test it?
- `cargo test -p common_utils --lib pii` — the new `test_phone_number_strategy_masking` fails on
  `main` (`5555` instead of `********5555`, and a panic for the short value) and passes with this
  change.
- `cargo +nightly fmt --all` and `cargo clippy -p common_utils --lib --tests` are clean.

## Checklist

Closes #14200

### DIFF
--- a/crates/common_utils/src/pii.rs
+++ b/crates/common_utils/src/pii.rs
@@ -44,9 +44,13 @@ where
     fn fmt(val: &T, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         let val_str: &str = val.as_ref();
 
-        if let Some(val_str) = val_str.get(val_str.len() - 4..) {
+        if let Some(last_four) = val_str
+            .len()
+            .checked_sub(4)
+            .and_then(|index| val_str.get(index..))
+        {
             // masks everything but the last 4 digits
-            write!(f, "{}{}", "*".repeat(val_str.len() - 4), val_str)
+            write!(f, "{}{}", "*".repeat(val_str.len() - 4), last_four)
         } else {
             #[cfg(feature = "logs")]
             logger::error!("Invalid phone number: {val_str}");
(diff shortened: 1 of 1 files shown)


hrithikesh026 (approved):

edubraqd:
@hrithikesh026 thanks for the approval on 09-07. Since then the only thing that changed is the required "linked issues" check, which was failing because the PR had no `Closes` line: it now closes #14200 and all checks are green (last run 09-13, after merging `main` in).

The PR has no status label, so I am not sure it is in anyone's queue. Is there something else needed from my side for it to move on, e.g. a code-owner approval from @juspay/hyperswitch-maintainers, or a fresh rebase on `main`? Happy to do either.

PR #14077, opened 2026-09-05
fix(webhooks): handle duplicate dispute webhooks idempotently

### LINKED ISSUE #14003: [BUG] Duplicate dispute webhook from connector returns 500 (DuplicateValue on dispute insert)
## Bug Description

When a connector delivers the **same dispute webhook twice at nearly the same instant**, one of the two requests fails with a `500`.

Both requests independently check whether the dispute already exists, both see "not present", and both attempt an insert. One insert wins; the other violates the unique constraint on the dispute record and surfaces as `DuplicateValue` → `500 Internal Server Error` back to the connector.

There is **no data corruption** — the dispute is persisted correctly, exactly once. The problem is purely that the losing request returns a 5xx instead of a successful/idempotent response, which shows up in our 5xx metrics and can trigger connector-side webhook retries.

## Root Cause

The dispute webhook flow has:

1. **No distributed lock** — unlike the payments flow, which locks on `payment_id` to serialize concurrent deliveries, the dispute flow does not lock on `connector_dispute_id`.
2. **No duplicate-insert handling** — a `DuplicateValue` from the DB is propagated as an internal error instead of being treated as "already processed".

This makes the check-then-insert sequence a classic race: the gap between the existence check and the insert is unguarded.

## Expected Behavior

Concurrent duplicate deliveries of the same dispute webhook should be handled idempotently — both requests should return `200`, with the dispute created once and subsequent deliveries treated as a no-op/update.

## Actual Behavior

- Request A: creates the dispute → `200`
- Request B (same instant, same `connector_dispute_id`): existence check passes → insert → unique constraint violation → `DuplicateValue` → `500`

## Impact

- 5xx on the webhook endpoint (alerting noise)
- Possible connector-side webhook retries for a dispute that was already recorded successfully
- No functional/data impact on the dispute record itself

## References

- Slack thread: https://juspay.slack.com/archives/C0516PQ9T9P/p1778324782272049

devaniketh:
## Type of Change

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description
When payment connectors deliver the same dispute webhook concurrently, both requests pass the initial lookup (`find_by_processor_merchant_id_payment_id_connector_dispute_id`) and attempt an insert into the database. One request succeeds while the second fails with a database unique constraint violation (`StorageError::DuplicateValue`). Previously, `.to_not_found_response(...)` mapped this `DuplicateValue` error to `ApiErrorResponse::InternalServerError` (HTTP 500).

This PR makes incoming dispute webhook processing strictly idempotent:
1. In `crates/router/src/core/webhooks/incoming.rs`:
   - Extracted `update_dispute_object` to encapsulate status/stage validation and database updates.
   - Intercepts `is_db_unique_violation()` upon `state.store.insert_dispute(...)`.
   - On detecting a duplicate due to concurrent insertion, logs the event, fetches the persisted dispute record from the store, and forwards it to `update_dispute_object` to return an idempotent **HTTP 200 OK**.
   - Increments `INCOMING_DISPUTE_WEBHOOK_NEW_RECORD_METRIC` only on successful insertions.
2. In `crates/router/src/db/dispute.rs`:
   - Enforces compound uniqueness across `(merchant_id, payment_id, connector_dispute_id)` in `MockDb::insert_dispute`, returning `StorageError::DuplicateValue` instead of `StorageError::MockDbError`.
   - Adds unit test `test_insert_duplicate_dispute_returns_duplicate_value`.

### Additional Changes

## Motivation and Context
Fixes #14003

Connectors aggressively retry webhook deliveries when receiving 5xx errors, causing unnecessary load and alerting noise for disputes that were already persisted successfully. Handling duplicate inserts gracefully ensures webhook idempotency.

## How did you test it?
- Added unit test `test_insert_duplicate_dispute_returns_duplicate_value` in `crates/router/src/db/dispute.rs` to verify that duplicate dispute insertions on matching `(merchant_id, payment_id, connector_dispute_id)` return `StorageError::DuplicateValue` and satisfy `is_db_unique_violation()`.

## Checklist

### DIFF
--- a/crates/router/src/core/webhooks/incoming.rs
+++ b/crates/router/src/core/webhooks/incoming.rs
@@ -1768,6 +1768,43 @@ pub async fn get_payment_attempt_from_object_reference_id(
 }
 
 #[allow(clippy::too_many_arguments)]
+async fn update_dispute_object(
+    state: &SessionState,
+    dispute: diesel_models::dispute::Dispute,
+    dispute_details: api::disputes::DisputePayload,
+    dispute_status: common_enums::enums::DisputeStatus,
+    platform: &domain::Platform,
+) -> CustomResult<diesel_models::dispute::Dispute, errors::ApiErrorResponse> {
+    let db = &*state.store;
+    logger::info!("Dispute Already exists, Updating the dispute details");
+    metrics::INCOMING_DISPUTE_WEBHOOK_UPDATE_RECORD_METRIC.add(1, &[]);
+    core_utils::validate_dispute_stage_and_dispute_status(
+        dispute.dispute_stage,
+        dispute.dispute_status,
+        dispute_details.dispute_stage,
+        dispute_status,
+    )
+    .change_context(errors::ApiErrorResponse::WebhookBadRequest)
+    .attach_printable("dispute stage and status validation failed")?;
+    let update_dispute = diesel_models::dispute::DisputeUpdate::Update {
+        dispute_stage: dispute_details.dispute_stage,
+        dispute_status,
+        connector_status: dispute_details.connector_status,
+        connector_reason: dispute_details.connector_reason,
+        connector_reason_code: dispute_details.connector_reason_code,
+        challenge_required_by: dispute_details.challenge_required_by,
+        connector_updated_at: dispute_details.updated_at,
+    };
+    db.update_dispute(
+        dispute,
+        update_dispute,
+        platform.get_processor().get_account().storage_scheme,
+    )
+    .await
+    .to_not_found_response(errors::ApiErrorResponse::WebhookResourceNotFound)
+}
+
+#[allow(clippy::too_many_arguments)]
 pub async fn get_or_update_dispute_object(
     state: SessionState,
     option_dispute: Option<diesel_models::dispute::Dispute>,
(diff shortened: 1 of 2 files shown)


XyneSpaces:
[blocking] The duplicate-insert recovery fetches the dispute with the processor merchant id, but the insert above stores `merchant_id` from `platform.get_provider()`. For concurrent duplicate dispute webhooks on a platform payment where processor and provider merchant ids differ, the unique-violation branch will fail to find the row that was just inserted and return `WebhookResourceNotFound` instead of updating idempotently. Use the same merchant id used by `DisputeNew::merchant_id` for this lookup, or store the processor id consistently in the insert path.

devaniketh:
@XyneSpaces  Good catch! Thank you for pointing that out.

I've updated the duplicate-recovery lookup to consistently use the provider merchant ID (`platform.get_provider().get_account().get_id()`) matching `DisputeNew::merchant_id`:

1. Bound `dispute_merchant_id` from `platform.get_provider().get_account().get_id()` and used it for both `DisputeNew::merchant_id` and the duplicate-recovery `find_by_processor_merchant_id_payment_id_connector_dispute_id` lookup, ensuring the row is found on platform payments where processor and provider differ.
2. Updated `MockDb::find_by_processor_merchant_id_payment_id_connector_dispute_id` to match either `processor_merchant_id` or `merchant_id`, aligning with the PostgreSQL store's query fallback behavior.
3. Added a unit test `test_find_by_processor_merchant_id_when_processor_and_provider_differ` covering the case where provider and processor merchant IDs differ.

Pushed in commit `2fe026c4d`. Please let me know if any other adjustments are needed!

PR #14078, opened 2026-09-06
feat(blocklist): export a merchant's blocklist data as a CSV

ayush22667:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

Adds a merchant-facing CSV export of the blocklist.

`POST /blocklist/export` inserts a job row and enqueues a `process_tracker` task, then returns
`202`. Nothing is scanned in the request path. The scheduler workflow keyset-pages the `blocklist`
table on `fingerprint_id` and streams each page into an S3 multipart upload, flushing a part once
the buffer crosses 5 MB, so peak memory is one page plus one part regardless of blocklist size.

The result is surfaced through the two batch job endpoints that already exist rather than new ones:

- `GET /blocklist/batch` now returns exports alongside uploads, with `job_type`, `file_name` and
  `downloadable` on each row, and an optional `?job_type=upload|export` filter. It is scoped by
  `X-Profile-Id` the same way `GET /blocklist` is.
- `GET /blocklist/batch/{job_id}` returns a short-lived presigned `download_url` for a completed
  export. The link is minted per request and never stored, so a page of exports does not issue a
  credential for every row.

Details worth calling out:

- The scan is bounded by the job's creation time, so an export that takes minutes still describes
  one moment, and a retry reproduces the same file rather than a later one. A bulk upload running
  concurrently contributes nothing to the file.
- Owner matching is split into two queries (`processor_merchant_id`, then the legacy
  `processor_merchant_id IS NULL` rows) rather than OR'd into one predicate. No index covers the
  OR'd form, so the planner falls back to a scan plus a sort of the whole matching set on every
  page — measured at 13x the wall clock and 40x the CPU over a 100k-row export.
- The scan leases from the read replica explicitly, since `pg_connection_read` resolves to master
  in the default build.
- The CSV header and encoders are shared with `POST /blocklist/batch`, so an export can be edited
  and uploaded back.
- `FileStorageInterface` gains multipart and presign methods. The `file_system` backend implements
  multipart against a staging file; it cannot presign, so blocklist export requires `aws_s3`.

### Additional Changes

- API: one new endpoint (`POST /blocklist/export`). `GET /blocklist/batch` and
  `GET /blocklist/batch/{job_id}` gain response fields and, for the listing, a query parameter and
  profile scoping. Existing fields are unchanged.
- Schema: `migrations/2026-09-04-100000_add_export_columns_to_batch_blocklist_jobs` adds
  `job_type`, `file_name`, `file_key`, `error_message` and `expires_at` to `batch_blocklist_jobs`,
  all nullable, with a one-off backfill setting `job_type = 'upload'` on existing rows.
- Config: `config/config.example.toml` documents that blocklist export requires the `aws_s3`
  file storage backend.

## Motivation and Context

Merchants can bulk-upload a blocklist but have no way to get it back out. `GET /blocklist` takes a
`u16` limit and pages by offset over a non-unique `created_at` ordering, so extracting a large
blocklist means hundreds of stitched-together calls, and rows shift underneath a paging client
while the bulk importer inserts.

Merchants want the list out to review what is currently blocked, to hand to a risk or compliance
team, to edit and re-upload in bulk, and to keep a copy before a large change.

## Linked Issues

- Closes https://github.com/juspay/hyperswitch-cloud/issues/23156

## How did you test it?

<details>
<summary>1. Blocklist Export - Start</summary>

Request
```
curl --location 'http://localhost:8080/blocklist/export' \
--header 'X-Profile-Id: pro_F8AQSrd8epJuReBK2N7m' \
--header 'api-key: [REDACTED]
```

Response
```
{
    "export_id": "blkexp_C07MenUYx02IcIHfxih7",
    "status": "initiated",
    "file_name": "blocklist_export_20260906_075650.csv"
}
```
</details>

<details>
<summary>2. Bulk Blocklist - Get Job Status</summary>

Request
```
curl --location 'http://localhost:8080/blocklist/batch/blkexp_C07MenUYx02IcIHfxih7' \
--header 'api-key: [REDACTED]
```

Response
```
{
    "job_id": "blkexp_C07MenUYx02IcIHfxih7",
    "merchant_id": "merchant_test_1788271010",
    "job_type": "export",
    "file_name": "blocklist_export_20260906_075650.csv",
    "status": "completed",
    "total_rows": 99988,
    "succeeded_rows": 99988,
    "failed_rows": 0,
    "created_at": "2026-09-06T07:56:50.012Z",
    "updated_at": "2026-09-06T07:56:53.834Z",
    "expires_at": null,
    "downloadable": true,
    "error_message": null,
    "download_url": null,
    "download_url_expires_at": null
}
```
</details>

## Checklist


XyneSpaces:
[blocking] This PR adds a new blocklist export API/workflow, storage changes, and file-storage integration without any test coverage in the changed files.

Please add coverage for export initiation plus worker success/failure paths so regressions in CSV generation, job status transitions, and download metadata are caught before merge.

XyneSpaces:
[blocking] This PR updates `batch_blocklist_jobs` models/schema for new export columns but only adds `down.sql`; please add the matching `up.sql` migration for `job_type`, `file_name`, `file_key`, `error_message`, and `expires_at`, including the upload backfill/defaults. Without the forward migration, a deployed DB will not have the columns that the new Diesel schema and runtime queries expect.

XyneSpaces:
[should-fix] This adds a new blocklist CSV export route plus scheduler workflow, but the PR does not add tests for export initiation, worker completion/failure, or download URL generation. Add coverage for the new `/blocklist/export` flow so the DB job state transitions and CSV/file-storage behavior are exercised.

XyneSpaces:
[blocking] This PR introduces the new `/blocklist/export` flow and scheduler workflow, but there are no success/failure/edge-case tests covering export creation, worker completion, or download URL/status behavior. Please add coverage under the existing router/blocklist test surface so regressions in the async export path are caught before release.

XyneSpaces:
[blocking] This PR updates Diesel schema for new `batch_blocklist_jobs` columns and adds a `down.sql`, but the matching `up.sql` migration is missing from the changed files. Deploying this code against an existing database will leave `job_type`, `file_name`, `file_key`, `error_message`, and `expires_at` absent, so the new reads/writes will fail at runtime.

Please add the corresponding `up.sql` that creates these columns, including any intended backfill/default for legacy upload rows.

XyneSpaces:
[blocking] Please add the missing `up.sql` for `migrations/2026-09-04-100000_add_export_columns_to_batch_blocklist_jobs`. This PR adds the down migration and updates Diesel schema/model code for `job_type`, `file_name`, `file_key`, `expires_at`, and `error_message`, but without a matching up migration an existing database will not get those columns before the new queries/inserts run.

XyneSpaces:
[blocking] Add the matching `up.sql` migration for the new `batch_blocklist_jobs` columns and backfill existing rows to `job_type = 'upload'` (or make the filters include `NULL` as upload). This PR updates Diesel schemas and adds queries filtering on `job_type`, but the changed files only include the down migration; a deployed app will either hit missing columns or hide pre-existing upload jobs whose `job_type` is `NULL`.

crates/diesel_models/src/query/batch_blocklist_job.rs
```diff
@@ -50,10 +83,51 @@ impl BatchBlocklistJob {
...
+                dsl::profile_id
+                    .eq(profile_id.to_owned())
+                    .or(dsl::profile_id.is_null()),
+            ),
+            Some(limit),
+            Some(offset),
+            Some(dsl::created_at.desc()),
+        )
+        .await
+    }
```
su-shivanshmathur, line 116: I don't see any index for this

crates/api_models/src/blocklist.rs
```diff
@@ -175,6 +180,24 @@ pub struct BatchBlocklistJobStatusResponse {
...
+    pub expires_at: Option<time::PrimitiveDateTime>,
+    /// Whether this job's file can be fetched right now.
+    pub downloadable: bool,
+    pub error_message: Option<String>,
+}
+
+/// Response for `GET /blocklist/batch/{job_id}`.
+#[derive(Debug, Clone, serde::Serialize, serde::Deserialize, ToSchema)]
+pub struct BatchBlocklistJobDetailResponse {
+    #[serde(flatten)]
```
hrithikesh026: no need for serde flatten

crates/diesel_models/src/query/blocklist.rs
```diff
@@ -320,6 +320,59 @@ impl Blocklist {
         .await
     }
 
+    pub async fn list_after_fingerprint_by_processor_merchant_id_profile_id(
+        conn: &DatabaseConnectionWithContext<'_>,
+        processor_merchant_id: &common_utils::id_type::MerchantId,
+        profile_id: &common_utils::id_type::ProfileId,
+        after_fingerprint_id: String,
+        snapshot_at: time::PrimitiveDateTime,
+        limit: i64,
```
hrithikesh026, line 363: lets have a hard limit on limit.

hrithikesh026: api should have a hard limit on the limit value.

crates/router/src/core/blocklist/export.rs
```diff
@@ -0,0 +1,224 @@
...
+        .attach_printable("Failed to write blocklist export CSV header")?;
+    writer
+        .into_inner()
+        .map_err(|error| error.into_error())
+        .change_context(errors::ApiErrorResponse::InternalServerError)
+        .attach_printable("Failed to finalize blocklist export CSV header")
+}
+
+/// Serializes one page of blocklist entries into CSV.
+pub(crate) fn rows_to_csv_bytes(rows: &[storage::Blocklist]) -> RouterResult<Vec<u8>> {
```
hrithikesh026, line 65: csv serializationa dn deserialization can be done through serde with csv crate

crates/diesel_models/src/query/batch_blocklist_job.rs
ayush22667: We have index on merchant_id, so this query is using it is only filtering based on profile id not doing seq scan
image

hrithikesh026 (approved):

su-shivanshmathur (approved):

PR #14083, opened 2026-09-07
ci(workflow): add automated draft-release generation for stable versions

itsharshvb:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [x] CI/CD

## Description

Adds `publish-stable-release.yml`, which auto-drafts the GitHub release for a
stable version once it's cut, instead of a maintainer assembling release
notes by hand.

It runs after `release-stable-version.yml` (`Release a stable version`)
completes successfully, or manually via `workflow_dispatch` (for
re-generating notes for a past tag). It:

- resolves the new/previous tag pair, cross-checking the version parsed from
  the triggering run's `release-notes.md` artifact against the newest tag at
  that commit
- builds a migration SQL diff, config diffs (`sandbox.toml`,
  `env_specific.toml`), static Docker-image/compatibility sections, and a
  filtered PR changelog via `git-cliff`
- assembles the full release body and, only if it exceeds GitHub's
  release-notes size limit, hard-truncates it at a UTF-8-safe boundary and
  appends a note asking the reviewer to download the untrimmed
  `full-release-notes.md` (attached to the release) and curate it by hand —
  this is expected to be a rare, once-in-a-while manual step rather than
  something worth automating further
- creates the release as a **draft** via `gh release create`, refusing to
  proceed if a release already exists for that tag, so a re-run can't
  silently create a duplicate

It authenticates as the same `hyperswitch-bot[bot]` GitHub App used by the
other release workflows (`release-stable-version.yml`,
`create-hotfix-tag.yml`), so the draft release is attributed to that bot
identity rather than the default `github-actions[bot]`.

### Additional Changes

## Motivation and Context

Stable-release notes were previously compiled by hand from a checklist
(migration/config diffs, changelog cleanup, Docker image links). This
automates that so every stable release gets a consistent, complete draft to
review and publish, without manual copy-paste.

## How did you test it?

`gh`/GitHub Actions can't be exercised locally without minting a real bot
token and creating a real draft release against the live repo, so instead the
non-GitHub-auth steps were extracted into standalone scripts and run against
real tags in this checkout (`v1.125.0` → `v1.126.0`):

- tag resolution, migration/config diffs, and `git-cliff` changelog
  generation + filtering all ran cleanly and produced the expected sections
- the assemble step was exercised both under the real 125,000-character
  GitHub limit (body passed through untouched) and with an artificially
  lowered limit to force the truncation path (cut cleanly at a UTF-8
  boundary, stayed under the limit, and appended the manual-curation note)

`yamllint` passes on the workflow file, and `bash -n` passes on every
extracted `run:` script block.

## Checklist


.github/workflows/release-stable-version.yml
```diff
@@ -145,22 +148,207 @@ jobs:
...
+      - name: Generate migration diff
+        id: migration-diff
         shell: bash
         run: |
-          # Override git-cliff tag pattern to only consider SemVer tags
-          export GIT_CLIFF__GIT__TAG_PATTERN='v[0-9]*'
+          set -euo pipefail
+          if ! DIFF_OUTPUT=$(git diff "$PREVIOUS_TAG" HEAD -- 'migrations/*/up.sql'); then
+            echo "::error::git diff failed for migrations/*/up.sql between $PREVIOUS_TAG and $NEXT_TAG"
+            exit 1
```
pixincreate: should we stop the workflow when there exist no migration diffs?

.github/workflows/release-stable-version.yml
```diff
@@ -145,22 +148,207 @@ jobs:
...
             --raw-field 'sha=${{ github.sha }}'
 
-      - name: Generate changelog
+      - name: Generate migration diff
+        id: migration-diff
         shell: bash
         run: |
-          # Override git-cliff tag pattern to only consider SemVer tags
-          export GIT_CLIFF__GIT__TAG_PATTERN='v[0-9]*'
+          set -euo pipefail
```
pixincreate, line 155: you can instead create a new file in ./scripts dir and dump it all there. then call the functions from here?

.github/workflows/release-stable-version.yml
```diff
@@ -145,22 +148,207 @@ jobs:
...
+        env:
+          GH_TOKEN: ${{ steps.generate_app_token.outputs.token }}
+        run: |
+          set -euo pipefail
+          REPO="${{ github.repository }}"
+
+          if gh release view "$NEXT_TAG" --repo "$REPO" >/dev/null 2>&1; then
+            echo "::error::A release already exists for tag $NEXT_TAG -- delete or publish it first, or this is an unintended re-run"
+            exit 1
+          fi
```
pixincreate, line 334: on a re-run this mints the next version instead of retrying the same one. cog get-version already returns the version that was just released, so next_tag moves up by one. the tag creation step would fail with a 422 first, but if anything ever gets past it, this check passes and you get a second tag and a draft release for the same commit with an empty changelog.

retrying the old version is not possible either, the calver check at the top rejects anything that is not today's date.

maybe accept previous and next tag as inputs, or make the tag creation idempotent and derive the target from github.sha.

.github/workflows/release-stable-version.yml
```diff
@@ -145,22 +148,207 @@ jobs:
...
+              }
+            }
+            {
+              if (drop_section) next
+              line = $0
+              lower_line = tolower(line)
+              if (lower_line ~ /\*\*deps:\*\*|\*\*dependencies:\*\*|\*\*dependabot:\*\*/) next
+              if (lower_line ~ /multitenancy|unified auth|click-to-pay|click to pay/) next
+              print
+            }
```
pixincreate, line 268: this only drops the parent line, not the children under it. git-cliff prints multi commit scopes as a '- **deps:**' bullet followed by indented '  - ' children, so dependabot and multitenancy entries still ship as parentless bullets.

drop the indented lines too.

.github/workflows/release-stable-version.yml
```diff
@@ -145,22 +148,207 @@ jobs:
...
+            echo "::error::git-cliff failed generating PR list for ${PREVIOUS_TAG}..${NEXT_TAG}"
+            exit 1
+          fi
+
+          awk '
+            BEGIN { drop_section = 0 }
+            /^### / {
+              heading = $0
+              gsub(/^### /, "", heading)
+              if (heading == "Testing" || heading == "Documentation") {
```
pixincreate, line 254: drop_section stays set once it is set. it only resets when another '### ' heading comes along, so when testing or documentation is the last group, the 'full changelog' footer gets dropped too.

reset it on the footer, or stop filtering once the footer is reached.

.github/workflows/release-stable-version.yml
```diff
@@ -145,22 +148,207 @@ jobs:
...
+              } else {
+                drop_section = 0
+              }
+            }
+            {
+              if (drop_section) next
+              line = $0
+              lower_line = tolower(line)
+              if (lower_line ~ /\*\*deps:\*\*|\*\*dependencies:\*\*|\*\*dependabot:\*\*/) next
+              if (lower_line ~ /multitenancy|unified auth|click-to-pay|click to pay/) next
```
pixincreate, line 266: this matches anywhere in the line, not just the start. a commit like '- **router:** handle timeout in click-to-pay flow' gets dropped even though it is a real router change.

match at the start of the line, or match on the scope prefix.

.github/workflows/release-stable-version.yml
```diff
@@ -145,22 +148,207 @@ jobs:
...
+
+          if [ "$FINAL_SIZE" -lt "$HARD_LIMIT" ]; then
+            cp /tmp/full-release-notes.md /tmp/assembled-body.md
+            echo "trimmed=false" >> "$GITHUB_OUTPUT"
+          else
+            echo "Body exceeds GitHub's release notes size limit -- truncating."
+            MARKER=$'\n\n---\n**Truncated: this release exceeds the GitHub release notes size limit. Download full-release-notes.md attached to this release and curate the notes manually.**'
+            MARKER_LEN=$(LC_ALL=C; echo -n "$MARKER" | wc -c)
+            KEEP=$((HARD_LIMIT - MARKER_LEN - 100))
+            head -c "$KEEP" /tmp/full-release-notes.md > /tmp/assembled-body-truncated.md
```
pixincreate, line 303: head -c can cut inside a fenced block, so the marker can end up inside the code fence and the rest of the fence never closes.

truncate on a line boundary, or close the fence before the marker.

.github/workflows/release-stable-version.yml
```diff
@@ -145,22 +148,207 @@ jobs:
...
+          set -euo pipefail
+          generate_config_diff_section() {
+            local file_path="$1"
+            local heading="$2"
+            local diff_output
 
-      - name: Upload changelog as build artifact
-        uses: actions/upload-artifact@v4
-        with:
-          name: release-notes.md
```
pixincreate, line 164: this removes the release-notes.md artifact upload. i did not find anything else in the repo reading it, but any external consumer of that artifact will stop getting it.

worth a check before this lands.

.github/workflows/release-stable-version.yml
pixincreate: Pending review for `.github/workflows/release-stable-version.yml` at `11f54c2`.

Verdict: assembly and truncation are sound. I ran the truncation path with a body over the 125,000-character limit; the output stayed under it and decoded as valid UTF-8. The changelog filter corrupts output, and a failed run cannot be retried, so I would hold the merge on the two Critical comments.

Description mismatch: the PR body describes a new `publish-stable-release.yml` with a `workflow_run` trigger and a version cross-check against the `release-notes.md` artifact. Commit `c251abba` folded that into this file and removed the artifact, so the description no longer matches the code. Update it before merge.

Nits, covered here instead of inline:
- Unused step ids: `migration-diff`, `config-diffs`, `static-sections`, `pr-list` (lines 152, 175, 208, 231).
- `CREATE_ERR` (line 347) holds gh output echoed on success (the release URL); the name misleads.
- `--ignore-space-change --ignore-space-at-eol` (line 184): the second is mostly subsumed by the first.
- `LC_ALL=C; echo -n ... | wc -c` (line 301): `wc -c` is locale-independent, and the assignment leaks to later commands in the step.
- No `concurrency` group, so two dispatches can race on the `latest` tag.
- Hardcoded registry/image names and SES text (lines 213-218) go stale silently.

What works well:
- `set -euo pipefail` and explicit `::error::` handling on every new step.
- The git-cliff config is edited in `/tmp`; the old step mutated the tracked `.github/git-cliff-changelog.toml`.
- Exclusive `PREVIOUS_TAG..HEAD` range replaces generate-then-delete.
- Least-privilege `permissions: contents: read`, with writes through the bot app token.

This review is pending. Edit, delete, or add comments in the Files changed tab, then submit when ready.

.github/workflows/release-stable-version.yml
```diff
@@ -145,22 +148,207 @@ jobs:
...
+              echo '```diff'
+              echo "$diff_output"
+              echo '```'
+            fi
+            echo ""
+          }
+
+          {
+            generate_config_diff_section "config/deployments/sandbox.toml" "Config Diff — sandbox.toml"
+            generate_config_diff_section "config/deployments/env_specific.toml" "Config Diff — env_specific.toml"
```
SanchithHegde, line 204: Can we have a single heading itself, as we have all the configuration changes mentioned under a single heading today?

.github/workflows/release-stable-version.yml
```diff
@@ -145,22 +148,207 @@ jobs:
...
+            echo "${NEXT_TAG} (with AWS SES support): docker pull docker.juspay.io/juspaydotin/hyperswitch-router:${NEXT_TAG}"
+            echo ""
+            echo "${NEXT_TAG}-standalone (without AWS SES support): docker pull docker.juspay.io/juspaydotin/hyperswitch-router:${NEXT_TAG}-standalone"
+            echo ""
+          } > /tmp/section-docker-images.md
+
+          {
+            echo "## Compatibility"
+            echo ""
+            echo "Comparing \`${PREVIOUS_TAG}\` → \`${NEXT_TAG}\`."
```
SanchithHegde, line 224: Why do we have the "comparing" line under the compatibility section?

.github/workflows/release-stable-version.yml
SanchithHegde, line 254: Can't we update the git-cliff config file to drop certain headings instead?

.github/workflows/release-stable-version.yml
SanchithHegde, line 266: Can we revisit these topics/phrases and decide whether we still have to drop changes related to these from the changelog, or should we keep them?

.github/workflows/release-stable-version.yml
```diff
@@ -145,22 +148,207 @@ jobs:
...
+        shell: bash
+        run: |
+          set -euo pipefail
+
+          HARD_LIMIT=125000
+
+          cat /tmp/section-docker-images.md /tmp/section-compatibility.md \
+              /tmp/section-migration-diff.md /tmp/section-config-diffs.md \
+              /tmp/filtered-pr-list.md \
+              > /tmp/full-release-notes.md
```
SanchithHegde, line 290: This seems to be reordering sections compared to what we have in the existing release notes?

.github/workflows/release-stable-version.yml
```diff
@@ -145,22 +148,207 @@ jobs:
...
+
+          FINAL_SIZE=$(wc -c < /tmp/full-release-notes.md)
+          echo "Assembled body size: $FINAL_SIZE"
+
+          if [ "$FINAL_SIZE" -lt "$HARD_LIMIT" ]; then
+            cp /tmp/full-release-notes.md /tmp/assembled-body.md
+            echo "trimmed=false" >> "$GITHUB_OUTPUT"
+          else
+            echo "Body exceeds GitHub's release notes size limit -- truncating."
+            MARKER=$'\n\n---\n**Truncated: this release exceeds the GitHub release notes size limit. Download full-release-notes.md attached to this release and curate the notes manually.**'
```
SanchithHegde, line 300: What is this even about?

PR #14088, opened 2026-09-07
ci: add permissive license compliance check on pull requests

### LINKED ISSUE #14072: [FEATURE] Add permissive license compliance check on pull requests
### Feature Description

Add a GitHub Actions workflow that inspects the dependency licenses introduced by a pull request and blocks the merge if any strong-copyleft license enters the tree. Weak-copyleft and unknown licenses shall be surfaced as a non-blocking PR comment.

## Policy to implement
 
| Tier | Licenses | PR behaviour |
|---|---|---|
| **A — Allow** | MIT, MIT-0, Apache-2.0, Apache-2.0 WITH LLVM-exception, BSD-1/2/3-Clause, ISC, 0BSD, Zlib, BSL-1.0, Unlicense, CC0-1.0, Unicode-3.0, Unicode-DFS-2016, CDLA-Permissive-2.0, OpenSSL, PostgreSQL | pass, silent |
| **B — Warn** | MPL-2.0, CDDL-1.0/1.1, EPL-1.0/2.0, CPL-1.0, Ms-PL, OSL-3.0, Artistic-2.0, CC-BY-3.0/4.0 | pass + sticky comment |
| **C — Deny** | GPL-1.0/2.0/3.0 (`-only`/`-or-later`), AGPL-1.0/3.0, LGPL-2.0/2.1/3.0, SSPL-1.0, BUSL-1.1, Elastic-2.0, CC-BY-NC-\*, CC-BY-SA-\*, CC-BY-ND-\*, JSON, WTFPL, Beerware, Commons-Clause, Sleepycat, EUPL-1.2, RPL-1.5, SISSL | **fail, blocks merge** |
| **D — Unknown** | no declared license, `non-standard`, `NOASSERTION`, below confidence threshold | pass + sticky comment |

**In scope**
- PRs evaluate the **dependency delta only** — packages added or version-bumped relative to the merge base. Authors get told about the dependency *they* introduced
- The implementation shall be replicable to other repos: juspay/prism, juspay/hyperswitch-web, juspay/hyperswitch-control-center, juspay/hyperswitch-sdk-android, juspay/hyperswitch-sdk-ios, juspay/hyperswitch-suite, juspay/hyperswitch-helm, juspay/decision-engine 
- Runtime and dev/build dependencies both evaluated, reported separately.

**Out of scope**
- Source-file license header scanning (needs ScanCode-class tooling and has a long false-positive tail).
- Container base images / `Dockerfile`.
- Any change to existing dependencies. This PR must be green on `main` as-is.

None

Utkal059:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [x] CI/CD

## Description

Adds a pull request check that inspects the dependency licenses a change
introduces and blocks the merge when a strong-copyleft or source-available
license enters the tree. Weak-copyleft and undeclared licenses are surfaced in
a sticky comment without blocking.

| File | Role |
| --- | --- |
| `.github/license-policy.toml` | The policy: tiers A–D, the lockfiles to scan, per-package exceptions |
| `.github/scripts/license_check.py` | Lockfile diffing, license resolution, classification, reporting |
| `.github/scripts/tests/test_license_check.py` | 60 unit tests, run by the workflow |
| `.github/workflows/license-compliance.yml` | Runs the check on `pull_request` and `merge_group` |
| `.github/workflows/license-compliance-comment.yml` | Posts the sticky comment via `workflow_run` |
| `docs/license_compliance.md` | Design, operations and the porting guide |

### Delta only

Only packages that a pull request **adds or version-bumps relative to the merge
base** are evaluated. Existing dependencies are never re-evaluated, so authors
are told only about dependencies they actually introduced, and tightening the
policy cannot retroactively break unrelated pull requests. Runtime and
dev/build dependencies are both evaluated and reported in separate columns.

### No build, no toolchain

Licenses come from crates.io and npm registry metadata, and lockfiles are read
straight out of the git object database. Nothing from the pull request is
installed, built or executed, and no language toolchain is provisioned — the
whole check takes a few seconds. Source-file header scanning stays out of
scope, as the issue specifies.

### SPDX expressions are evaluated, not string-matched

This is where a naive implementation gets both false positives and false
negatives. Verified against real registry data:

| Declared license | Result | Why |
| --- | --- | --- |
| `(MIT OR GPL-3.0-or-later)` — how `jszip` ships | Allow | Either license may be taken |
| `MIT AND GPL-3.0-only` | **Deny** | Both sets of terms must be honoured |
| `Apache-2.0 WITH LLVM-exception` | Allow | Listed in tier A as a whole |
| `GPL-2.0-or-later WITH Classpath-exception-2.0` | **Deny** | An unlisted exception cannot soften its license |
| `GPL-2.0+`, `MIT/Apache-2.0` | Deny, Allow | Deprecated spellings normalised |

`OR` resolves to its least restrictive operand, `AND` to its most restrictive
one, and where an identifier matches more than one tier the most restrictive
tier wins.

### Runtime vs dev/build for Cargo

`Cargo.lock` records no dependency kinds, so the runtime set is derived by
walking the lockfile graph from the `[dependencies]` of every workspace
manifest. Two things had to be handled explicitly, both verified against this
workspace:

- A workspace crate's lockfile edges merge its normal, dev and build
  dependencies into one list. Walking them pulled `criterion`, `proptest` and
  `test-case` into the runtime set, so the walk uses the crate's manifest
  instead.
- A workspace crate that only ever appears as another crate's dev dependency —
  `test_utils` — must not contribute its dependencies to the runtime set,
  otherwise `serial_test`, `thirtyfour` and `fantoccini` look like runtime.

The result on `main` today is 1042 runtime and 43 dev/build Cargo packages, the
latter being exactly the `criterion`, `proptest`, `wiremock`, `serial_test` and
WebDriver chains. The npm trees here are test harnesses in their entirety, so
the policy pins them to dev/build with `scope = "dev/build"`.

### Security model

The check runs on `pull_request`, which for a fork gets a read-only token and
no secrets, and uploads its report as an artifact. A separate `workflow_run`
workflow posts the comment with a write-capable token. The alternative —
`pull_request_target` with a checkout of the pull request's head — would run
fork-authored code with a privileged token, which `pr-convention-checks.yml`
already warns against. In the commenting workflow the artifact is treated
strictly as data: the body is passed as a single quoted argument, and the pull
request to comment on is resolved from the head commit rather than read out of
the report. Only `gh` and its built-in `--jq` are used, so no standalone `jq`
is required on the runner.

### Escape hatch

A misclassification does not need a policy change. `[[exceptions]]` pins a
single package to a tier, is matched before any registry lookup, and requires a
written reason that shows up in the report.

### Additional Changes

## Motivation and Context

Closes #14072.

The policy tiers, the delta-only scope, the separate runtime and dev/build
reporting and the list of repositories to replicate to are taken from that
issue. `.github/license-policy.toml` is the only repository-specific file:
porting the check to `juspay/prism`, `hyperswitch-web`,
`hyperswitch-control-center`, `hyperswitch-suite`, `hyperswitch-helm`,
`decision-engine` or the SDK repositories means copying five files and editing
the `[[ecosystems]]` list. `docs/license_compliance.md` documents that, and the
`COLLECTORS` registry is where Gradle and Swift Package Manager support would
be added for the Android and iOS SDKs.

## How did you test it?

**Unit tests** — 60 tests covering the SPDX tokenizer, parser precedence, tier
evaluation, the policy this repository ships, both Cargo and npm collectors,
delta computation, exceptions and report rendering. They need no network and no
dependencies beyond the standard library, and the workflow runs them:

```console
$ python -m unittest discover --start-directory .github/scripts/tests
Ran 60 tests in 0.043s

OK
```

**End to end against a scratch commit** that added real copyleft dependencies —
`sequoia-openpgp` to `crates/router` `[dependencies]`, `option-ext` to its
`[dev-dependencies]`, and `jszip` plus `spdx-exceptions` to
`cypress-tests/package-lock.json`:

```console
$ python .github/scripts/license_check.py --base-ref HEAD~2 --head-ref HEAD
1775 dependencies at 903d3944e895, 1771 at 4deee415e790: 4 added or updated.
::error title=Denied license::The cargo package sequoia-openpgp 1.21.1 is licensed under LGPL-2.0-or-later, which the license policy denies.

Denied       sequoia-openpgp  1.21.1  LGPL-2.0-or-later          [runtime, cargo]
Needs review option-ext       0.2.0   MPL-2.0                    [dev/build, cargo]
Needs review spdx-exceptions  2.5.0   CC-BY-3.0                  [dev/build, npm]
Allowed      jszip            3.10.1  (MIT OR GPL-3.0-or-later)  [dev/build, npm]

1 denied license(s) introduced by this change.
$ echo $?
1
```

Note `jszip`: the dual-licensed package passes on its MIT option, where a
string match against the deny list would have blocked it. Moving
`sequoia-openpgp` between `[dependencies]` and `[build-dependencies]` flips its
scope between `runtime` and `dev/build` as expected, and bumping `jszip` to
`3.9.1` reports it as `` `3.9.1` (was `3.10.1`) ``, which covers the
version-bump path.

**Green on `main` as-is**, as the issue requires — this pull request changes no
dependencies, so the delta is empty and no comment is posted:

```console
$ python .github/scripts/license_check.py --base-ref "$(git merge-base upstream/main HEAD)"
1771 dependencies at 4deee415e790, 1771 at 3aeafab0da64: 0 added or updated.
No new or updated dependencies to check.
$ echo $?
0
```

**Workflow logic** — both workflows parse as YAML, every `run` block passes
`bash -n` under `set -euo pipefail`, and the revision-determination and check
steps were executed locally with the `pull_request` event variables set,
producing the expected `GITHUB_OUTPUT` values, the three artifact files and the
job summary.

**Not verifiable before merge:** GitHub only runs `workflow_run` workflows from
the default branch, so the sticky comment itself cannot be exercised on this
pull request. The report body it posts is what the block above renders. Once
merged, **Check licenses of newly introduced dependencies** is the job to mark
as a required status check.

## Checklist


Utkal059:
@SanchithHegde @knutties @manojradhakrishnan — this is ready for a look whenever you have time.

Two things that might save you some review effort.

**It follows the spec literally.** I ran all 64 licenses from the tier table in #14072 through the shipped policy and every one lands in the tier the issue assigns it, with the behaviours matching too (A silent, B and D comment-only, C blocks the merge). The PR itself changes no dependencies, so the delta is empty and it's green on main as-is.

**Replication status.** I went and looked at what each repo in the issue list actually has committed today:

- `hyperswitch-prism`, `hyperswitch-web`, `hyperswitch-control-center`, `decision-engine` — work as-is. Copy the five files, edit the `[[ecosystems]]` list, done.
- `hyperswitch-suite` — has `.terraform.lock.hcl`, so it needs a Terraform collector.
- `hyperswitch-sdk-android`, `hyperswitch-sdk-ios`, `hyperswitch-helm` — no `gradle.lockfile` / `Podfile.lock` / `Chart.lock` committed. Worth flagging that this isn't about collector coverage: a delta-only check needs a lockfile to diff against, so enabling dependency locking there is a prerequisite for any implementation of this, not just mine.

On the obvious alternative — I did look at `actions/dependency-review-action` first, but `allow-licenses` and `deny-licenses` are mutually exclusive and `warn-only` is all-or-nothing, so "block on C, comment on B and D, stay quiet on A" isn't expressible in a single run, and it doesn't split runtime from dev/build. `cargo-deny` (the commented-out job in `CI-push.yml`) is Rust-only and scans the whole tree rather than the delta, so it would trip on dependencies already on main.

Three things I need from your side:

1. Approval for the workflow runs, so the two new jobs actually execute here.
2. After merge, marking **Check licenses of newly introduced dependencies** as a required status check.
3. Heads up that the sticky comment can't be demonstrated on this PR — GitHub only runs `workflow_run` workflows from the default branch, so it starts working once this lands on main.

Everything in the policy is data — tiers, `fail_on`/`comment_on`, and the lockfiles scanned are all editable in `.github/license-policy.toml` with no code change. Happy to drop `docs/license_compliance.md` or the `workflow_dispatch` trigger if you'd rather not carry them.

PR #14091, opened 2026-09-07
feat(deja): route every clock and entropy read in the router through one seam

### LINKED ISSUE #14112: [FEATURE] Route every clock and entropy read in the router through one instrumented seam
### Feature Description

A recorded request can only be replayed against a candidate build if everything
the original code read that was not part of the request is reproducible. The
clock and the process's entropy source are the two things nearly every
non-trivial request touches that are not inputs, and today nothing stops a new
call site from reading either directly.

`common_utils` already has instrumented `date_time` functions and a
`generate_id` family, but they are a convention rather than a rule. Connectors
and core code read the clock and the RNG raw in many places, and several of
those values reach the wire: a timestamp signed into an outbound request, a
UUID used as an idempotency key, a `nanoid` payment reference, a nonce or salt
that is signed over.

The concrete failure this causes: a connector signs its outbound request over
`OffsetDateTime::now_utc()` read at the call site. That value is invisible to a
replay harness — there is nowhere to substitute it — so the replayed signature
never matches the recorded one, the request fails before reaching the assertion
that mattered, and the divergence is reported against the candidate rather than
against the harness. The same shape recurs for `Uuid::new_v4()`/`Uuid::now_v7()`
identifiers, `thread_rng()` nonces, and raw `nanoid!` references.

The gap is not that these sites are wrong today. It is that nothing prevents the
next one being added, and that when one is added the resulting failure looks
like a product bug rather than a missing seam.

### Possible Implementation

Route every read of the clock or of entropy in the router through one
instrumented place in `common_utils`, and make that a property the build
enforces rather than a convention:

- Extend the `common_utils` seams to cover the entropy side — UUID v4 and v7,
  `nanoid` with its own alphabet, random alphanumeric/numeric strings, random
  bytes, a random index, a uniform `f64`, a random permutation — plus the two
  ambient values that differ between a recording host and a replay host
  (process id, hostname).
- Add `disallowed-methods` entries to `.clippy.toml` naming the raw method each
  seam replaces, and a `disallowed-macros` entry for `nanoid!`, which a
  `disallowed-methods` rule can never see because it expands at the call site
  rather than naming a function.
- Allow the raw read only in a seam body, or where routing it would be a genuine
  dependency cycle, with the reason recorded at the site.

Recording mode must return the real value in every case, so the outbound bytes
are unchanged; only replay substitutes a recorded one. The substitution is of an
*input*, not of a computed outcome — the candidate still computes its own
signature over the recorded timestamp and its own retry schedule from the
recorded jitter, so a candidate that changed that logic still diverges.

maverox:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [x] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

Closes #14112

This PR routes every read of the clock or of entropy in the router through
one place in `common_utils`, so that a recorded run can be replayed against a
candidate byte-for-byte. The seams themselves do not change what the router
sends: in record mode every one still calls the real system clock or the real
RNG and returns the real value, and only replay substitutes a recorded one. Two
behaviour changes do ride along, both bug fixes surfaced by the seaming and both
described below — a double clock read in revenue recovery, and the ordering that
OIDC signing-key selection needed. Nothing else in this diff changes what runs. A candidate that
changed its own signing algorithm, retry schedule, or connector-selection
logic still diverges, because the candidate still computes over the
substituted input itself — this replaces a raw clock or RNG read, not a
computed outcome.

The concrete case that motivated this: a connector signs its outbound request
over `OffsetDateTime::now_utc()` read directly at the call site. That value is
invisible to a replay harness — there is nowhere to substitute it — so the
replayed signature never matches the one that was recorded, and the request
fails before it reaches the assertion that mattered. The same shape recurs
for `Uuid::new_v4()`/`Uuid::now_v7()` idempotency keys and request references,
`thread_rng()` nonces and salts several connectors sign into their requests,
and raw `nanoid!` payment references.

`common_utils` already had five instrumented `date_time` functions and a
`generate_id` family; this adds the entropy side (UUID v4/v7, nanoid with
nanoid's own alphabet, random alphanumeric/numeric strings, random bytes, a
random index, a uniform `f64`, a random permutation) plus two ambient-state
seams (`process_id`, `hostname`) for the two values that differ between a
recording host and a replay host. `.clippy.toml` gains a `disallowed-methods`
list naming the raw method each seam replaces and a `disallowed-macros` entry
for `nanoid!`, which a `disallowed-methods` rule can never see because it
expands at the call site rather than naming a function. Every seam body is
the one place still allowed to read the raw source, marked with an
`#[allow(clippy::disallowed_methods)]` carrying the reason; the same applies
to a small number of sites where routing through the seam is a genuine
dependency cycle (`router_env` cannot depend on `common_utils` — the
dependency runs the other way — so its own clock, pid, and hostname reads,
and its hand-written UUID v7 record/replay boundary, stay where they are,
documented as such) or where the read happens once at process boot before any
request exists.

One addition in `date_time` is worth separating out, because the diff places
it next to a removal and so reads as a rename. `now_unix_timestamp_millis` is
new, and it is for the connectors: around ten of them were open-coding a
millisecond timestamp at the call site, as
`now_utc().unix_timestamp_nanos() / 1_000_000` or
`SystemTime::now().duration_since(UNIX_EPOCH)?.as_millis()`, several signing
that value into the outbound request. It has nothing to do with
`now_unix_timestamp_nanos`, which is still present. That function's two
telemetry callers now reach the clock through the existing `now()` seam and
convert, and the function itself was rewritten to do the same rather than
read the clock directly — so it is reproducible on replay for the first time,
where before it was one of the raw reads this change exists to remove. No
timestamp in this diff changes unit: every nanosecond site stayed
nanoseconds, and every site that moved onto the millisecond seam was already
computing milliseconds beforehand.

Two real bugs surfaced and got fixed as a side effect of seaming, not as
separate work: revenue recovery was reading the clock twice to build one
timestamp (`OffsetDateTime::now_utc().date()` and `.time()` as two separate
calls), so a row written in the last instant of a day could get tomorrow's
date with today's time. And the OIDC signing-key selection drew a random
index into a `HashMap`'s iteration order, which is reseeded per process — the
draw alone being reproducible wasn't enough, because the same recorded index
could resolve to a different key depending on process-local hash seeding. The
map is now sorted before the draw, so the index has a stable meaning.

One case is intentionally left open: a connector signs over an RSA-PSS salt,
and the RNG that produces it is unreachable from outside `ring` — the trait
that would let it be substituted is sealed. The right fix is to move that one
signing implementation onto a crate whose API takes the RNG as an argument,
which is a live cryptographic change that deserves its own review rather than
riding inside a determinism refactor. It's called out with the relevant code
citations in `crypto.rs` rather than worked around.

The ingress recorder in `router_env` now stamps `deja::ROLE_INGRESS` on the
event it raises for `http_incoming`, so a recording says what its correlation
root is rather than leaving a consumer to recognise that boundary by name.
Every reader of the field treats it as a classification input, OR'd with the
boundary name this code already sets, and it is never compared between a
recorded and an observed event — so no admission, scoring or divergence
outcome changes. The visible effect is the role appearing beside the boundary
in the sealed correlation index.

On recorded volume, the number worth stating plainly: routing the database
and Redis telemetry through the `now()` seam is the expensive part of this
change. Measured on an existing tape rather than derived, it adds roughly
twenty events per request, taking a correlation from about forty-four
recorded events to about sixty-four — near a 45% increase. The connector-side
seams are cheap beside that, at most three per payment request. If the volume
turns out to matter, the fix belongs inside the seam — a boundary declared
non-scoring, recorded and replayed as normal but excluded from divergence
comparison and from the event budget — rather than in raw reads scattered
back across the codebase.

This branch is rebased onto current `main`, and two commits were dropped in
the process rather than carried. One of them re-applied a form that had
already been reviewed and changed on `main`: it attached `.in_current_span()`
to the outer `tokio::spawn` argument, which reindents every line of the
spawned block, where `main` binds the future first and spawns it in one line.
It also reintroduced `crates/router_env/tests/deja_tail_capture.rs`, a file
removed on review whose name asserts the opposite of what its body asserts.
Neither was visible to `git cherry`, because #14000 was squash-merged and no
patch-id matches; it was found by reading `main`, and the two one-line fixups
that existed only to patch that commit were dropped with it.

Every `Cargo.lock` change against `main` has been accounted for line by line,
which on this branch is not ceremony: a rebase whose lock predates `main`'s
silently reverts everything `main` has moved since, and it compiles, and the
lint gates pass. The lock is compared as per-package dependency sets rather
than as a textual diff, because a diff cannot say which `[[package]]` block a
moved line belongs to. It differs from `main` by exactly one edge.

This branch also moves the `deja` pin to `de0a42a`, which is `deja`'s `main`,
across all nine declarations — and that corrects a regression rather than only
moving forward. Until this change the branch pinned `9809180`, which is an
*ancestor* of the `1838093` that hyperswitch `main` pins, so the diff showed
nine declaration lines walking the pin sixteen commits backwards across seven
crates (`router` and `router_env` carry two each). Merging in that state would
have taken `deja` with it. `1838093` is an ancestor of `de0a42a`, so the pin
now moves forward of `main` instead of behind it.

All nine move together by necessity: two revisions of the same crate in one
build means two copies of the recording statics, and a recorder writing into
the copy nobody reads fails silently rather than erroring. So the check that
matters is that the old revision has zero occurrences afterwards, `Cargo.lock`
included, rather than that the build still succeeds.

`Cargo.lock` ends up differing from `main` by exactly one edge — `test_utils`
gaining `common_utils` from the test-harness seaming — and that cleanliness is
load-bearing rather than incidental. An earlier pin bump on this branch had
quietly moved two unrelated transitive edges, `libloading`'s `windows-targets`
and `prost-build`'s `itertools`, and nothing catches that: both resolutions
compile, both pass every lint gate, and a textual diff of the lock does not
even say which package a moved line belongs to. Both manifests ask for wide
ranges (`>=0.48, <0.54` and `>=0.10, <=0.14`), so cargo is not misbehaving —
it is exercising latitude the requirements grant it, and it will do so again
whenever the lock is regenerated. They are restored to `main`'s resolutions
here, which is why they no longer appear in the diff at all.

The gRPC transport's hand-rolled `fail_stop_absent_transport` is replaced by
`deja::__private::fail_stop_absent_executor`, which `deja` now provides, so the
router stops carrying its own copy of a convention it does not own. The two
copies had already drifted — `deja`'s message wraps the target in backticks and
the local one wrapped it again — so keeping both would have put doubled
backticks in operator-facing fail-stop text. `AbsentTransportError` stays: it is
the error a boundary with no transport returns if the run thunk is ever
invoked, which replay never does.

One change here is deliberately outside that subject, and it is better to say
so than to have a reviewer discover it. The `http_outgoing` codec recorded a
response's headers one value per name, so a response carrying three
`set-cookie` headers recorded one and lost the other two. It is an insert that
should have been an append: capture now accumulates values under their name,
and reconstruct appends each one. That accounts for 12 of the divergent rows
on the reference sweep — not at this boundary, which matches on both sides,
but downstream at `km`, where a payload built from the live response on record
is compared against one built from the reconstructed response on replay.

Header order is deliberately left alone rather than fixed alongside it, since
the comparison already sorts arrays before comparing and absorbs a reordering
as non-blocking. A second, unrelated defect found while testing this is
filed as #14115 and deliberately not fixed here.

**It is not a clock read, not an entropy read, and not the gate that catches
one.** It is in this PR because the same person owns both changes and wanted
them in one image and one review, not because it meets the bar the rest of
this diff is held to. Stating that rather than filing it under "deja work" is
the point: the PR should not appear to claim a scope it does not have.

Aliasing `HashMap`/`HashSet` construction behind a seam, so that a hash map's
per-process random seed is reproducible on replay, is **deferred rather than
out of scope, and the distinction matters**: the hash-seed collections work has
been deprioritised, and the per-process seed is being tolerated with a
non-blocking warning instead of being closed. That is a known nondeterminism
source knowingly left open, and it should be read as a decision someone made
rather than a gap nobody noticed. **The image still needs a fresh build before
this can be tested on the sandbox pod.**
### Changes by area

The diff is 132 files, and all but a handful are one-line call-site moves. It
reads much faster grouped by what each path is doing.

**The seams themselves — `crates/common_utils/`**

`src/lib.rs` (+312) is the substance of this PR: the entropy and identifier
seams (UUID v4 and v7, nanoid over nanoid's own alphabet, random alphanumeric
and numeric strings, random bytes, a random index, a uniform `f64`, a random
permutation) alongside the `date_time` functions that already existed, plus the
two ambient-state seams `process_id` and `hostname` for the values that differ
between a recording host and a replay host. `src/crypto.rs` (+76) carries the
salt and nonce generation that several connectors sign over. `src/types.rs`,
`src/keymanager.rs` and `src/tokenization.rs` are call sites in the same crate.

**Clock call sites**

`crates/hyperswitch_connectors/src/connectors/` — 27 files, plus `src/utils.rs`
and `src/connector_relay/fiservcommercehub.rs`. These are the ones that matter:
a connector signing over a directly-read timestamp is the case that motivated
the PR.

`crates/router/src/` — five under `core/`, three under `services/`, three under
`events/`, and one each under `workflows/`, `utils/` and `types/`.
`crates/hyperswitch_interfaces/src/events/` (both API-log files),
`crates/redis_interface/src/metrics.rs`,
`crates/diesel_models/src/query/generics.rs` and
`crates/router_env/src/logger/formatter.rs` are telemetry timestamps rather than
anything a connector signs.

**Entropy and identifier call sites**

`crates/hyperswitch_connectors/src/connectors/` — 36 files: idempotency keys,
request references, nonces and salts.

`crates/router/src/` — 16 under `core/`, plus `types/domain/user.rs` and
`utils/user/password.rs`. `crates/router/src/configs/settings.rs` (+69) is worth
looking at directly: OIDC signing-key selection now sorts the key ids before
indexing, so the choice no longer depends on which `HashMap` instance holds the
set, and the index itself comes from the seam.

`crates/scheduler/src/{consumer,producer,utils}.rs` and the test files under
`crates/router/tests/` and `crates/test_utils/tests/connectors/` are the
remainder.

**Boundary payload and header construction — `crates/external_services/`**

`src/http_client/boundary.rs` (+96/-18) is the `http_outgoing` codec, and the
change is an insert becoming an append. Response headers were recorded one value
per name, so a response carrying three `set-cookie` headers recorded one; capture
now accumulates values under their name and reconstruct appends each one. A
recording written before this had already lost every repeat when it was written,
so reconstruct refuses it rather than replaying a tape that still carries the
defect — deja turns that into a fail-stop naming the cause.

`src/http_client.rs` (+2/-1) and a comment in the same codec correct a claim
`main` still carries: that a recorded connector error reconstructs to `None`,
which the boundary answers by executing live against the real endpoint. It does
not. `None` maps to `Reconstructed::Failed` and fail-stops the request; there is
no silent live fallback.

Where that mattered is worth stating, because it is not at this boundary.
`http_outgoing` captures identically on both sides and matches. The webhook path
builds a key-manager payload from the LIVE response when recording and from the
RECONSTRUCTED one when replaying, so only the replay side reads through the
lossy capture — fifteen entries became thirteen, and the two payloads diverged
downstream at `km`. That accounts for twelve of the fifty-seven value
divergences on the reference sweep.

Header ORDER is deliberately not addressed. The comparison canonicalises arrays
before comparing them (`sort_as_bag` is a sort with no dedup), so the same
entries in a different order are already absorbed as non-blocking, while
thirteen entries against fifteen cannot be made equal by any sort. The loss was
the whole cause; ordering was never going to be charged.

`src/grpc_client/deja_transport.rs` (+15/-39) deletes a locally-written
fail-stop in favour of the one the runtime already provides, so there is one
definition of that behaviour rather than two that can drift.

**The gate and its declared exceptions**

`.clippy.toml` (+18) names the raw method each seam replaces under
`disallowed-methods`, and `nanoid!` under `disallowed-macros` — a macro expands
at the call site and so can never be caught by a method rule.

The exceptions are declared at the site with the reason inline rather than
collected in a list: `crates/router/src/utils.rs`,
`crates/router/src/services/jwt.rs` and `crates/router/src/deja_boot.rs` (reads
that happen once at boot, before any request exists, where seaming would be
circular), and `crates/router_env/src/request_id.rs` (`router_env` cannot depend
on `common_utils` — the dependency runs the other way).

**Manifests**

Eight `Cargo.toml` files and `Cargo.lock` move the `deja` pin off a topic branch
and onto `main`. All of them move together by necessity: two revisions in one
build means two copies of the recording statics and a recorder that silently
does nothing.

### Additional Changes

## Motivation and Context

A recorded request can only be replayed against a candidate if everything the
original code read that wasn't part of the request is reproducible. The
clock and the process's entropy source are the two things every non-trivial
request touches that aren't inputs, and until now nothing stopped a new call
site from reading either directly. This makes that a property the build
enforces: `disallowed-methods`/`disallowed-macros` fail the lint gate on any
raw read outside a seam body, so the next person who needs a timestamp or a
UUID is pointed at the seam instead of reintroducing the same class of
divergence.

## How did you test it?

`just clippy` and `just clippy_v2` both pass clean (0 lint hits) on this
branch, including a third run with the `deja` feature explicitly disabled
(neither of the two default gate configurations actually exercises that
build arm, since `release` transitively re-enables `deja` even where the
recipe excludes it by name). Mutation-verified: reintroducing a single raw
clock read in a connector makes `just clippy_v2` fail and name the exact
line. `cargo +1.85.0 check -p router --features release` passes, for the
MSRV floor.

Ran the full workspace test suite before and after this branch's changes,
against the commit this branch is based on, in an isolated worktree: the same
21 pre-existing failures appear on both, unrelated to this branch (missing
local fixtures and credentials this environment doesn't have — a test
private key, live database/KMS access) — zero tests newly failing, zero newly
passing.

Worth stating plainly, because it changes what a green tick means here: **CI
runs no `cargo test`.** The job named "Run tests on stable toolchain" runs
`just clippy`, `just clippy fred`, a `Cargo.lock` check and `cargo check`;
the only `cargo test` invocations in the repository are under
`.github/workflows/archive/`, and the `cargo-nextest` step is commented out.
So the tests this PR adds will never execute in CI, and the workspace run
above was done by hand rather than by a pipeline. Separately and not caused
by this branch, `external_services`'
`utils::tests::test_deserialize_hashset_inner_empty_string` already fails on
unmodified `main`; it is noted here so that it is not attributed to this
change if anyone does start running tests on the branch.

Added two unit tests for properties this refactor specifically needs to keep
that a shape-only test wouldn't catch: that the recovery codes a user is
issued are pairwise distinct (not just individually well-formed), and that
choosing an OIDC signing key resolves to the same candidate order regardless
of which `HashMap` instance holds the key set (the property the sort exists
for, tested directly rather than assumed).

## Checklist

### DIFF
--- a/crates/common_utils/src/keymanager.rs
+++ b/crates/common_utils/src/keymanager.rs
@@ -11,7 +11,6 @@ use once_cell::sync::OnceCell;
 use router_env::{
     global_meter, histogram_metric_f64, instrument, logger, metric_attributes, tracing,
 };
-use time::OffsetDateTime;
 
 #[cfg(feature = "ext_services_latency")]
 use crate::consts::EXTERNAL_CALL_TAG;
@@ -141,7 +140,7 @@ where
         ))?;
 
     let latency_ms = elapsed.as_millis();
-    let created_at_timestamp = OffsetDateTime::now_utc().unix_timestamp_nanos();
+    let created_at_timestamp = crate::date_time::now().assume_utc().unix_timestamp_nanos();
     #[cfg(feature = "ext_services_latency")]
     {
         let downstream_request_id = response
--- a/crates/common_utils/src/lib.rs
+++ b/crates/common_utils/src/lib.rs
@@ -106,14 +106,10 @@ pub mod date_time {
 
     /// Return the UNIX timestamp in nanoseconds of the current date and time in UTC
     ///
-    /// Reads the clock through `now()` rather than calling `now_utc()` again.
-    /// `now()` is already an instrumented seam, so this carries no seam and no
-    /// gate exception of its own, and it costs no event a caller of `now()`
-    /// would not already have paid. The value is unchanged: `now()` keeps the
-    /// full nanosecond, so this returns the same `i128` the direct read did.
-    ///
-    /// `track_caller` so that a future caller's own location, not this body,
-    /// is what `now()` records.
+    /// Reads through `now()`, which is already a seam, so this needs no seam and
+    /// no gate exception of its own and costs no extra event. `now()` keeps the
+    /// full nanosecond, so the value is unchanged. `track_caller` so a caller's
+    /// own location is what `now()` records.
     #[cfg_attr(feature = "deja", track_caller)]
     pub fn now_unix_timestamp_nanos() -> i128 {
         now().assume_utc().unix_timestamp_nanos()
@@ -121,9 +117,8 @@ pub mod date_time {
 
     /// Return the UNIX timestamp in milliseconds of the current date and time in UTC.
     ///
-    /// Several connectors sign a millisecond timestamp into an outbound request.
-    /// They open-coded `now_utc().unix_timestamp_nanos() / 1_000_000`, which
-    /// reads the clock outside any seam and so cannot be reproduced on replay.
+    /// The seam for the millisecond timestamp several connectors sign into an
+    /// outbound request, previously open-coded as `now_utc()…nanos() / 1_000_000`.
     #[cfg_attr(feature = "deja", track_caller)]
     #[cfg_attr(
         feature = "deja",
@@ -301,15 +296,11 @@ pub mod date_time {
 
 /// Generate a version 4 (random) UUID.
 ///
-/// A v4 UUID is drawn entirely from the operating system's entropy, so nothing in
-/// the request determines it and a replay cannot reproduce one that was read
-/// raw. Connectors put this value in an idempotency key, a request reference and
-/// — for deutschebank, payeezy, authipay, fiserv, fiservemea and
-/// fiservcommercehub — inside the string they sign, so an unseamed read changes
-/// the outbound bytes and the substitute boundary misses.
-///
-/// Returns the `Uuid` rather than a formatted string so that callers keep their
-/// own `to_string`, `simple` or `hyphenated` rendering unchanged.
+/// Nothing in the request determines a v4 UUID, so an unseamed read cannot be
+/// replayed. Connectors put it in idempotency keys, request references, and — for
+/// deutschebank, payeezy, authipay and the fiserv family — inside the signed
+/// string, so a raw read moves the outbound bytes. Returns the `Uuid` rather than
+/// a string so callers keep their own rendering.
 #[inline]
 #[cfg_attr(feature = "deja", track_caller)]
 #[cfg_attr(
@@ -326,12 +317,10 @@ pub fn generate_uuid_v4() -> uuid::Uuid {
 /// This is the seam for the nonce-and-salt shape that connectors open-coded as
 /// `Alphanumeric.sample_string(&mut rand::thread_rng(), n)`.
 ///
-/// It deliberately keeps the thread-local RNG rather than routing through
-/// [`crypto::generate_cryptographically_secure_random_string`], which draws from
-/// `OsRng`: several of these values are signed into an outbound request, and
-/// swapping the generator underneath them would be a security change riding on a
-/// determinism refactor. Use the `crypto` one for anything that must be
-/// unpredictable to an attacker.
+/// Keeps the thread-local RNG deliberately: several of these are signed into
+/// outbound requests, and swapping the generator would be a security change riding
+/// on a determinism refactor. Use the `crypto` variant where unpredictability
+/// matters.
 #[inline]
 #[cfg_attr(feature = "deja", track_caller)]
 #[cfg_attr(
--- a/crates/common_utils/src/lib.rs
+++ b/crates/common_utils/src/lib.rs
@@ -79,6 +79,7 @@ pub mod date_time {
         deja::time(component = "common_utils", operation = "date_time::now", codec = SerdeCodec,)
     )]
     pub fn now() -> PrimitiveDateTime {
+        #[allow(clippy::disallowed_methods, reason = "this function IS the seam")]
         let utc_date_time = OffsetDateTime::now_utc();
         PrimitiveDateTime::new(utc_date_time.date(), utc_date_time.time())
     }
@@ -99,10 +100,47 @@ pub mod date_time {
         )
     )]
     pub fn now_unix_timestamp() -> i64 {
+        #[allow(clippy::disallowed_methods, reason = "this function IS the seam")]
         OffsetDateTime::now_utc().unix_timestamp()
     }
 
-    /// Return the UNIX timestamp in nanoseconds of the current date and time in UTC
+    /// Return the UNIX timestamp in milliseconds of the current date and time in UTC.
+    ///
+    /// Several connectors sign a millisecond timestamp into an outbound request.
+    /// They open-coded `now_utc().unix_timestamp_nanos() / 1_000_000`, which
+    /// reads the clock outside any seam and so cannot be reproduced on replay.
+    #[cfg_attr(feature = "deja", track_caller)]
+    #[cfg_attr(
+        feature = "deja",
+        deja::time(
+            component = "common_utils",
+            operation = "date_time::now_unix_timestamp_millis",
+            codec = SerdeCodec,
+        )
+    )]
+    pub fn now_unix_timestamp_millis() -> i128 {
+        // Read the clock directly rather than via `now_unix_timestamp_nanos()`:
+        // that one is not a seam, and routing through it would put an unseamed
+        // read on a seamed path.
+        #[allow(
+            clippy::disallowed_methods,
+            reason = "this IS the seam for a millisecond timestamp"
+        )]
+        let now = OffsetDateTime::now_utc();
+        now.unix_timestamp_nanos() / 1_000_000
+    }
+
+    /// Return the UNIX timestamp in nanoseconds of the current date and time in UTC.
+    ///
+    /// NOT a seam, deliberately: its only callers stamp telemetry events, one per
+    /// database call and one per redis call, and recording those would add roughly
+    /// nineteen clock events per request for a value that never reaches the wire
+    /// or a compared result. It is on `disallowed-methods` so that a new caller on
+    /// a request path has to say why; use `now_unix_timestamp_millis` there.
+    #[allow(
+        clippy::disallowed_methods,
+        reason = "telemetry-only; see the doc comment"
+    )]
     pub fn now_unix_timestamp_nanos() -> i128 {
         OffsetDateTime::now_utc().unix_timestamp_nanos()
     }
@@ -143,6 +181,7 @@ pub mod date_time {
             .encode();
         // Read the clock directly rather than via `now()`: both are instrumented seams, and
         // nesting them would record a redundant inner event for every outer call.
+        #[allow(clippy::disallowed_methods, reason = "this function IS the seam")]
         convert_to_pdt(OffsetDateTime::now_utc())
             .assume_utc()
             .format(&Iso8601::<ISO_CONFIG>)
@@ -158,6 +197,7 @@ pub mod date_time {
         )
     )]
     pub fn now_rfc7231_http_date() -> Result<String, time::error::Format> {
+        #[allow(clippy::disallowed_methods, reason = "this function IS the seam")]
         let now_utc = OffsetDateTime::now_utc();
         // Desired format: ddd, DD MMM YYYY HH:mm:ss GMT
         // Example: Fri, 23 May 2025 06:19:35 GMT
--- a/crates/common_utils/src/lib.rs
+++ b/crates/common_utils/src/lib.rs
@@ -104,6 +104,21 @@ pub mod date_time {
         OffsetDateTime::now_utc().unix_timestamp()
     }
 
+    /// Return the UNIX timestamp in nanoseconds of the current date and time in UTC
+    ///
+    /// Reads the clock through `now()` rather than calling `now_utc()` again.
+    /// `now()` is already an instrumented seam, so this carries no seam and no
+    /// gate exception of its own, and it costs no event a caller of `now()`
+    /// would not already have paid. The value is unchanged: `now()` keeps the
+    /// full nanosecond, so this returns the same `i128` the direct read did.
+    ///
+    /// `track_caller` so that a future caller's own location, not this body,
+    /// is what `now()` records.
+    #[cfg_attr(feature = "deja", track_caller)]
+    pub fn now_unix_timestamp_nanos() -> i128 {
+        now().assume_utc().unix_timestamp_nanos()
+    }
+
     /// Return the UNIX timestamp in milliseconds of the current date and time in UTC.
     ///
     /// Several connectors sign a millisecond timestamp into an outbound request.
--- a/crates/common_utils/src/lib.rs
+++ b/crates/common_utils/src/lib.rs
@@ -119,30 +119,12 @@ pub mod date_time {
         )
     )]
     pub fn now_unix_timestamp_millis() -> i128 {
-        // Read the clock directly rather than via `now_unix_timestamp_nanos()`:
-        // that one is not a seam, and routing through it would put an unseamed
-        // read on a seamed path.
         #[allow(
             clippy::disallowed_methods,
             reason = "this IS the seam for a millisecond timestamp"
         )]
         let now = OffsetDateTime::now_utc();
         now.unix_timestamp_nanos() / 1_000_000
-    }
-
-    /// Return the UNIX timestamp in nanoseconds of the current date and time in UTC.
-    ///
-    /// NOT a seam, deliberately: its only callers stamp telemetry events, one per
-    /// database call and one per redis call, and recording those would add roughly
-    /// nineteen clock events per request for a value that never reaches the wire
-    /// or a compared result. It is on `disallowed-methods` so that a new caller on
-    /// a request path has to say why; use `now_unix_timestamp_millis` there.
-    #[allow(
-        clippy::disallowed_methods,
-        reason = "telemetry-only; see the doc comment"
-    )]
-    pub fn now_unix_timestamp_nanos() -> i128 {
-        OffsetDateTime::now_utc().unix_timestamp_nanos()
     }
 
     /// Calculate execution time for a async block in milliseconds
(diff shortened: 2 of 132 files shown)


crates/common_utils/src/lib.rs
```diff
@@ -99,12 +100,31 @@ pub mod date_time {
         )
     )]
     pub fn now_unix_timestamp() -> i64 {
+        #[allow(clippy::disallowed_methods, reason = "this function IS the seam")]
         OffsetDateTime::now_utc().unix_timestamp()
     }
 
-    /// Return the UNIX timestamp in nanoseconds of the current date and time in UTC
```
maverox: why did you remove this and added a milli sec variant?

maverox, line 105: Two independent changes that share a hunk, not a rename. The diff puts the
deletion and the addition side by side, so it reads as `now_unix_timestamp_nanos`
becoming `now_unix_timestamp_millis`, and that isn't what happened.

**The millisecond variant is for the connectors.** Around ten of them were
open-coding a millisecond timestamp at the call site — `authipay`, `boku`,
`fiserv`, `fiservemea`, `fiservcommercehub`, `globepay` and `tokenio` as
`OffsetDateTime::now_utc().unix_timestamp_nanos() / 1_000_000`, and `paybox`
and `payeezy` as `SystemTime::now().duration_since(UNIX_EPOCH)?.as_millis()`.
Several sign that value into the outbound request. Every one was a raw clock
read outside any seam, so a replay had nowhere to substitute it and the
signature never matched what was recorded. `now_unix_timestamp_millis()` is the
seam for that shape, and it would have been added whether or not
`now_unix_timestamp_nanos` existed.

**The nanos function lost its callers separately.** On `main` it had exactly
two, both setting `created_at_timestamp` on an `ExternalServiceCall` event:
`diesel_models/src/query/generics.rs:89` and `redis_interface/src/metrics.rs:147`.
Both now go through the `date_time::now()` seam that already existed and
convert at the call site:

```rust
created_at_timestamp: common_utils::date_time::now()
    .assume_utc()
    .unix_timestamp_nanos(),
```

which left the function with zero callers.

**Units are unchanged everywhere, and that is the part worth checking.** Both
functions return `i128`, so a nanosecond field quietly becoming a millisecond
one would compile, pass every test, and land as telemetry off by a factor of a
million. I checked it mechanically rather than by eye — classifying every hunk
in the diff by the unit it produces — and exactly one code hunk shows a unit
change: this one, which is the artifact of two unrelated changes sharing a
hunk rather than a conversion. Every nanosecond site stayed nanoseconds, and
every site that moved onto the millisecond seam was already computing
milliseconds beforehand, the four telemetry ones included.

`now().assume_utc().unix_timestamp_nanos()` is also the same value as the
direct read rather than an approximation: `now()` builds its `PrimitiveDateTime`
from the same `OffsetDateTime`'s `date()` and `time()`, which carry the whole
nanosecond, and `time`'s serde format keeps every subsecond digit, so a replay
gets the recorded nanosecond back rather than a rounded second.

**On the deletion: you were right to stop on it, and it is back.** Restored in
`cd6d93b159` as a thin wrapper over the seam rather than as the raw read it
used to be:

```rust
#[cfg_attr(feature = "deja", track_caller)]
pub fn now_unix_timestamp_nanos() -> i128 {
    now().assume_utc().unix_timestamp_nanos()
}
```

The argument for deleting it was that a raw wrapper nobody calls cannot be
argued as a gate exception once the rule is that every clock read is
instrumented. That holds for a wrapper which reads the clock itself; it does
not hold for one that delegates. This reads no raw clock, so it needs neither
an `#[allow]` nor a `.clippy.toml` entry, and it is not itself a seam, so it
records nothing of its own — a call costs the single `date_time::now` event the
inline expression already costs. `track_caller` so a caller's own location,
rather than this body, is what `now()` records.

Two reasons that beat taking the break. On `main` this was an unseamed clock
read, which is the thing this PR exists to remove — deleting it made it
unavailable, whereas wrapping it makes it reproducible for the first time. And
zero callers in this tree is not zero callers: anything unmerged that adds one
would have met a compile error instead of a working seam.

It also puts the function back above `now_unix_timestamp_millis` rather than in
place of it, so the hunk stops reading as a rename. That was the real defect
here — not the deletion itself, but a diff that asserted something untrue about
two unrelated changes, which is what your question was reacting to and what the
next reviewer would have asked too.

swangi-kumari (dismissed):
connector changes looks good to me

crates/router/src/services/oidc_provider.rs
```diff
@@ -192,10 +192,8 @@ async fn generate_id_token(
     state: &SessionState,
     auth_code_data: &AuthCodeData,
 ) -> OidcResult<String> {
-    let now = SystemTime::now()
-        .duration_since(UNIX_EPOCH)
-        .change_context(OidcErrors::ServerError)?
-        .as_secs();
+    let now = u64::try_from(common_utils::date_time::now_unix_timestamp())
+        .map_err(|_| report!(OidcErrors::ServerError))?;
```
jarnura: Don't discard errors, do change_context

crates/router/src/services/oidc_provider.rs
```diff
@@ -226,10 +224,8 @@ async fn generate_access_token(
...
```
jarnura: Don't discard errors, do change_context

jarnura (changes requested):

hrithikesh026 (approved):

crates/common_utils/src/keymanager.rs
```diff
@@ -141,7 +140,7 @@ where
         ))?;
 
     let latency_ms = elapsed.as_millis();
-    let created_at_timestamp = OffsetDateTime::now_utc().unix_timestamp_nanos();
+    let created_at_timestamp = crate::date_time::now().assume_utc().unix_timestamp_nanos();
```
kashif-m, line 143: nit: avoid specifying full crate path for imports

```suggestion
    let created_at_timestamp = date_time::now().assume_utc().unix_timestamp_nanos();
```

crates/common_utils/src/tokenization.rs
```diff
@@ -11,6 +11,5 @@ use crate::consts::TOKEN_LENGTH;
 /// # Returns
 /// A randomly generated token string of length `TOKEN_LENGTH`
 pub fn generate_token() -> String {
-    use nanoid::nanoid;
-    nanoid!(TOKEN_LENGTH)
+    crate::generate_nanoid_with_default_alphabet(TOKEN_LENGTH)
```
kashif-m, line 14: nit: avoid using full crate path

crates/common_utils/src/types.rs
```diff
@@ -1539,7 +1539,8 @@ pub struct PublishableKey(LengthString<PUBLISHABLE_KEY_LENGTH, PUBLISHABLE_KEY_L
 impl PublishableKey {
     /// Create a new PublishableKey Domain type without any length check from a static str
     pub fn generate(env_prefix: &'static str) -> Self {
-        let publishable_key_string = format!("pk_{env_prefix}_{}", uuid::Uuid::now_v7().simple());
+        let publishable_key_string =
+            format!("pk_{env_prefix}_{}", crate::generate_uuid_v7().simple());
```
kashif-m, line 1543: nit: same

crates/external_services/src/http_client/boundary.rs
```diff
@@ -174,12 +181,15 @@ impl deja::codec::ReplayCodec for HttpResponseCodec {
...
+/// Returning `None` does not fall through to a live call. deja maps it to
+/// `Reconstructed::Failed`, which fail-stops the request with a named reason —
+/// so a recorded value this build cannot read halts the replay rather than
+/// quietly reaching the real endpoint.
+///
+/// Headers are read as an array of values per name and no other shape is
+/// accepted. A recording written before that capture stored a single string per
+/// name and had already lost every repeat; reading one would replay a tape that
+/// still carries the defect this fixes, so it fail-stops instead.
 pub(super) fn replay_response(recorded: &serde_json::Value) -> Option<reqwest::Response> {
```
kashif-m, line 193: nit: this entire description can be concise - what it does and how it does it. One sentence each

crates/external_services/src/http_client/boundary.rs
```diff
@@ -193,8 +203,13 @@ pub(super) fn replay_response(recorded: &serde_json::Value) -> Option<reqwest::R
 
     let mut builder = http::Response::builder().status(status);
     if let Some(headers) = recorded.get("response_headers").and_then(|h| h.as_object()) {
-        for (name, value) in headers {
-            if let Some(value) = value.as_str() {
+        for (name, values) in headers {
+            // `?` rather than a skip: a name whose values are not an array is a
+            // pre-fix recording, and replaying it with its headers dropped would
+            // be worse than refusing it.
+            for value in values.as_array()?.iter().filter_map(|value| value.as_str()) {
```
kashif-m, line 210: Not needed?

jarnura (approved):

kashif-m (approved):

jagan-jaya (approved):

VenuMadhav2541 (approved):

ThisIsMani (approved):

PR #14093, opened 2026-09-07
feat(core): consume connector webhooks carrying associated data for settled payments

Vani-1107:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

Currently, we do not consume webhooks that are received after a payment has already moved to the Success state.
For the Trustly connector, two independent webhooks are sent:
- One webhook contains the payment status update.
- The other contains payment-related information.

If the payment has already moved to Success when the payment-information webhook is received, we currently ignore it.
This PR handles this case by introducing a new event to process payment-associated data. This event only updates the information associated with the payment and does not modify the payment status.

Webhook changes
- New AssociatedDataUpdate flow, peer of dispute/mandate, for events carrying supplementary data rather than a status change
- Fixes Trustly's account notification being dropped — arrived post-settlement, classified as a payment event, PSync skipped it over a Succeeded intent
- Writes straight to the attempt: no status change, no PSync, no outgoing webhook; should_call_connector untouched
- Sources per path — trait hook on Direct, decoded EventContent on UCS
- Grouped by target record, so payment methods later is a field addition (later will be extended to PaymentMethods also)

Payout changes
- Adds Trustly bank transfer support to the UCS payout conversions
- Passthrough check on the Create Recipient flow: supports_create_recipient takes is_passthrough, and Trustly skips it for passthrough — because passthrough carries a psp_token/psp_customer_id, so the recipient already exists at the PSP and registering it again is redundant

### Additional Changes

## How did you test it?

TestCases in Prism PR - https://github.com/juspay/hyperswitch-prism/pull/2255

## Checklist


Mahnoor-Zaffar (commented):
Reviewed:`crates/router/src/core/webhooks/incoming.rs`. 


CI: multiple test failures, including on stable toolchain.

Overall, the changes look promising, but addressing the CI failures and adding tests for edge cases will strengthen the PR. Looking forward to seeing the final version!

crates/router/Cargo.toml
```diff
@@ -243,8 +243,8 @@ rust_decimal = { version = "1.37.1", features = [
 ] }
 rust-i18n = { git = "https://github.com/kashif-m/rust-i18n", rev = "f2d8096aaaff7a87a847c35a5394c269f75e077a" }
 
-unified-connector-service-client = { git = "https://github.com/juspay/connector-service", tag = "2026.09.03.0", package = "rust-grpc-client" }
-unified-connector-service-cards = { git = "https://github.com/juspay/connector-service", tag = "2026.09.03.0", package = "ucs_cards" }
+unified-connector-service-client = { git = "https://github.com/juspay/connector-service", branch = "trustly-payment-additional-data", package = "rust-grpc-client" }
+unified-connector-service-cards = { git = "https://github.com/juspay/connector-service", branch = "trustly-payment-additional-data", package = "ucs_cards" }
```
hrithikesh026, line 247: Please change this to point to a tag

crates/router/src/core/unified_connector_service/transformers.rs
```diff
@@ -9019,11 +9023,50 @@ impl_ucs_payout_response_transformation!(
     payments_grpc::PayoutServiceStageResponse,
     merchant_payout_id
 );
+
 #[cfg(feature = "payouts")]
-impl_ucs_payout_response_transformation!(
-    payments_grpc::PayoutServiceCreateRecipientResponse,
-    merchant_payout_id
-);
+impl
```
hrithikesh026: can't we just use the macro itself??

Vani-1107: The macro only covers the five fields common to all payout responses(merchant_payout_id, payout_status, connector_payout_id, error, status_code) - everything else, including payout_connector_metadata and
connector_eligibility_reference_id, is hardcoded to None in its body.
This response has an extra `payout_connector_metadata` field carrying the recipient details returned by the connector, which we need to map into payout_connector_metadata since the subsequent transfer call reads it back. Same reason the eligibility flow has its own impl - it needs payout_eligible and connector_eligibility_reference_id on top of the common fields.
We could write a macro specific to this flow, but it would have exactly one call site, so it wouldn't really help.

crates/router/Cargo.toml
Vani-1107: will change this before getting this merged (once the corresponding UCS PR is merged). This has some proto changes and would need proto bump.

crates/router/src/core/unified_connector_service/transformers.rs
hrithikesh026: can't we update the macro to include those extra fields?

crates/router/src/core/payouts.rs
```diff
@@ -1338,6 +1338,11 @@ pub async fn complete_create_recipient(
     connector_data: &api::ConnectorData,
     payout_data: &mut PayoutData,
 ) -> RouterResult<()> {
+    let is_passthrough = matches!(
+        payout_data.payout_method_data.as_ref(),
+        Some(payouts::PayoutMethodData::Passthrough(_))
+    );
```
Sakilmostak: make it a impl function over payoutmethoddata

crates/common_enums/src/connector_enums.rs
```diff
@@ -238,11 +238,15 @@ impl Connector {
...
-        )
+    pub fn supports_create_recipient(
+        self,
+        payout_method: Option<PayoutType>,
+        is_passthrough: bool,
+    ) -> bool {
+        match self {
+            Self::Trustly => !is_passthrough,
+            _ => matches!(payout_method, Some(PayoutType::Bank)),
+        }
```
Sakilmostak: flatten this match case

crates/hyperswitch_domain_models/src/router_flow_types/webhooks.rs
```diff
@@ -19,3 +19,17 @@ impl ConnectorNetworkTxnId {
         &self.0
     }
 }
+
+/// Associated data about a payment shared by a connector over a webhook.
+/// Fields are grouped by the record they are written to, so that support for further records
+/// (payment methods, for instance) can be added without changing the flow's signature.
+#[derive(Debug, Clone, Serialize)]
+pub struct WebhookAssociatedData {
+    pub payment_attempt: Option<PaymentAttemptAssociatedData>,
```
Sakilmostak: can we make this non optional given empty object can deserialize this object silently

crates/router/src/core/webhooks/incoming.rs
```diff
@@ -2303,6 +2324,196 @@ async fn mandates_incoming_webhook_flow(
...
+    connector: &ConnectorEnum,
+    request_details: &IncomingWebhookRequestDetails<'_>,
+    content: &super::gateway::WebhookContent,
+) -> CustomResult<WebhookResponseTracker, errors::ApiErrorResponse> {
+    if !source_verified {
+        logger::error!("Webhook source verification failed for associated data webhook flow");
+        return Err(report!(
+            errors::ApiErrorResponse::WebhookAuthenticationFailed
+        ));
+    }
```
Sakilmostak: nit: introduce match case instead of early return

crates/router/src/core/webhooks/incoming.rs
```diff
@@ -2303,6 +2324,196 @@ async fn mandates_incoming_webhook_flow(
...
+        super::gateway::WebhookContent::UnifiedConnectorService(bytes) => {
+            associated_data_from_ucs_event_content(bytes)?
+        }
+    };
+
+    // Acknowledge webhooks that carry nothing to write so the connector does not retry them.
+    let Some(associated_data) = associated_data else {
+        logger::info!("No associated data in incoming webhook body, skipping update");
+        return Ok(WebhookResponseTracker::NoEffect);
+    };
```
Sakilmostak: same here

crates/router/src/core/webhooks/incoming.rs
```diff
@@ -2303,6 +2324,196 @@ async fn mandates_incoming_webhook_flow(
...
+        .clone()
+        .perform_locking_action(&state, merchant_id.clone())
+        .await?;
+
+    let payment_attempt = get_payment_attempt_from_object_reference_id(
+        &state,
+        webhook_details.object_reference_id.clone(),
+        processor,
+    )
+    .await?;
```
Sakilmostak: redundant call, you could have persisted the payment_attempt from the above call

crates/router/src/core/webhooks/incoming.rs
```diff
@@ -2303,6 +2324,196 @@ async fn mandates_incoming_webhook_flow(
...
+            "Failed to read the payments response from unified connector service event content",
+        )?;
+
+    let sender_payment_instrument_id = payments_response
+        .sender_payment_instrument_id
+        .map(Secret::new);
+
+    if sender_payment_instrument_id.is_none() {
+        return Ok(None);
+    }
```
Sakilmostak: nit: no early return, use match case

crates/router/src/core/webhooks/incoming.rs
```diff
@@ -2303,6 +2324,196 @@ async fn mandates_incoming_webhook_flow(
...
+            .as_ref()
+            .map(|id| id.peek().as_str()),
+        "sender_payment_instrument_id",
+        &payment_attempt.payment_id,
+    );
+
+    if sender_payment_instrument_id.is_none() {
+        logger::info!("No new associated data to write to the payment attempt, skipping update");
+        return Ok(());
+    }
```
Sakilmostak: same here

crates/router/src/core/webhooks/incoming.rs
Vani-1107: The first call is only there to get the payment_id - we need it as the locking key, so it has to happen before the lock is acquired.
The second call re-reads the attempt after the lock is held. That matters because resolve_write_once_field compares the incoming value against what's already stored, so it needs current state. Reusing the first result would mean comparing against a row read before the lock.

Sakilmostak (approved):

deepanshu-iiitu (approved):

hrithikesh026 (approved):

PR #14094, opened 2026-09-07
ci(deja): auto-apply deja-art label and re-trigger replay on S-test-ready

iemyashasvi:
Adds `.github/workflows/deja-art-labeler.yml`, which automates the `deja-art` label so every eligible PR gets a Deja ART replay without anyone having to remember to add the label.

## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [x] CI/CD

## Description

Two jobs, both driven by `pull_request_target` (no checkout, so the trigger is safe):

- **Add `deja-art` on PR creation.** When a PR is opened, reopened, or marked ready for review, the `deja-art` label is applied. Draft PRs are skipped and get the label once they are marked ready for review. Adding a label that is already present is a no-op, so reopen does not re-trigger anything.
- **Re-trigger the replay when `S-test-ready` is added.** When the `S-test-ready` label lands, the workflow removes `deja-art` (if present), waits 10 seconds so webhook consumers see a clean `unlabeled` → `labeled` transition, and re-adds it. This re-runs the Deja ART replay against the test-ready code.

Notes:

- Labels are applied with the hyperswitch-bot GitHub App token rather than `GITHUB_TOKEN`. This is required so the resulting `labeled` events trigger other Actions workflows (for example the gate in #13614); label changes made with `GITHUB_TOKEN` never do.
- Both jobs are restricted to same-repo branches. The `deja-art` label kicks off a Jenkins pipeline that builds and replays the PR's code, which we do not want to run for forks. This mirrors how `S-test-ready` is handled in `cypress-tests-runner.yml`.
- This lives in its own workflow file rather than in `pr-convention-checks.yml`, so the convention checks are not re-run on every label change.

### Additional Changes

## Motivation and Context

We want Deja ART replays to be mandatory for PRs rather than opt-in. Auto-applying the label on creation removes the manual step, and re-triggering on `S-test-ready` ensures the replay reflects the code that is actually going through the cypress test gate.

## How did you test it?

- YAML parsed and validated with `actionlint` (including shellcheck) with no findings.
- Job conditions and token usage follow the existing `pr_labeler` job in `pr-convention-checks.yml`.

## Checklist

https://claude.ai/code/session_01Es6Dybc1fsYKRzgpi8F6Km

### DIFF
--- a/.github/workflows/deja-art-labeler.yml
+++ b/.github/workflows/deja-art-labeler.yml
@@ -7,7 +7,7 @@ on:
   # pull request whenever this trigger is used.
   # Since we do not have a checkout step in this workflow, this is an
   # acceptable use of this trigger.
-  pull_request:  # TEMP: smoke test only, revert to pull_request_target before merge
+  pull_request_target:
     types:
       - opened
       - reopened
--- a/.github/workflows/deja-art-labeler.yml
+++ b/.github/workflows/deja-art-labeler.yml
@@ -1,0 +1,98 @@
+name: Deja ART Labeler
+
+on:
+  # This is a dangerous event trigger as it causes the workflow to run in the
+  # context of the target repository.
+  # Avoid checking out the head of the pull request or building code from the
+  # pull request whenever this trigger is used.
+  # Since we do not have a checkout step in this workflow, this is an
+  # acceptable use of this trigger.
+  pull_request_target:
+    types:
+      - opened
+      - reopened
+      - ready_for_review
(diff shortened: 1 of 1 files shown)


iemyashasvi:
Closing: the auto-labeling approach was rejected. Superseded by a plain required check that verifies the `deja-art` label is present.

PR #14098, opened 2026-09-08
ci(deja): add required check verifying the deja-art label is present

### LINKED ISSUE #14099: [FEATURE] Make the deja-art label a required check on pull requests
### Feature Description

Every pull request should go through a Deja ART replay before it can be merged. The replay pipeline is triggered by adding the `deja-art` label to a PR, but today adding that label is optional and easy to forget, so PRs can be merged without ever having been replayed.

Add a required status check that fails until the `deja-art` label is present on the PR and passes once it is. This makes triggering the replay a precondition for merging, without adding any tokens, API polling, or long-running runners to CI.

### Possible Implementation

- Add a small GitHub Actions workflow, triggered on `pull_request` (`opened`, `reopened`, `synchronize`, `ready_for_review`, `labeled`, `unlabeled`), with a single job named `Verify PR has the deja-art label` that reads the labels from the event payload and exits non-zero if `deja-art` is missing.
- The job always runs and reports pass/fail explicitly, rather than being skipped with a job-level `if`, because a skipped job counts as passing for required checks.
- Run on `synchronize` so a result is reported for every head commit, since required status checks are evaluated per commit.
- Pass unconditionally on `merge_group` so the check does not stall the merge queue, matching the other required checks.
- After the workflow is merged, a repository admin adds `Verify PR has the deja-art label` to the required status checks in the "Default Branch" ruleset.

Related: #13614 proposes a fuller gate that also waits for the replay verdict. This issue covers only the label-presence check.

iemyashasvi:
Adds `.github/workflows/deja-art-label-check.yml`: a single, fast check that fails until the `deja-art` label is on the PR and passes once it is.

## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [x] CI/CD

## Description

One job, **`Verify PR has the deja-art label`**, that reads the labels from the `pull_request` event payload and exits non-zero if `deja-art` is missing. No tokens, no API calls, no polling; the job takes a few seconds.

- Triggers on `labeled` / `unlabeled` so the verdict updates as soon as the label is added or removed, and on `synchronize` so a result is reported for every new head commit (required status checks are evaluated per commit).
- Handles `merge_group` by passing unconditionally, matching how the other required checks behave in the merge queue.
- The job always runs and reports pass/fail explicitly rather than being skipped via a job-level `if`, because a skipped job is treated as passing for required checks.

Once merged, the check name `Verify PR has the deja-art label` needs to be added to the "Default Branch" ruleset's required status checks by a repo admin.

### Additional Changes

## Motivation and Context

We want every PR to go through a Deja ART replay. The `deja-art` label is what triggers the replay pipeline, so making the label's presence a required check ensures nobody can merge without having triggered one.

## How did you test it?

- Validated with `actionlint` (including shellcheck) with no findings.
- Since this uses `pull_request`, the workflow runs from the PR head, so this PR itself exercises it: the check fails until `deja-art` is added here.

## Checklist

https://claude.ai/code/session_01Es6Dybc1fsYKRzgpi8F6Km

### DIFF
--- a/.github/workflows/deja-art-label-check.yml
+++ b/.github/workflows/deja-art-label-check.yml
@@ -1,23 +1,20 @@
 name: Deja ART Label Check
 
 on:
+  # Runs on `opened`, `synchronize` and `reopened` (the default types).
+  # `synchronize` is required so that a result is reported for every head
+  # commit, since required status checks are evaluated per commit.
   pull_request:
-    types:
-      - opened
-      - reopened
-      # Required status checks are evaluated per head commit, so the check
-      # must run on every push to report a result for the new commit.
-      - synchronize
-      - ready_for_review
-      - labeled
-      - unlabeled
 
+  # Required status checks must report a result in the merge queue as well,
+  # otherwise the queue entry stalls waiting for this check.
   merge_group:
     types:
       - checks_requested
 
 permissions:
   contents: read
+  pull-requests: read
 
 jobs:
   deja_art_label_check:
@@ -25,20 +22,28 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-      - name: Skip check for merge queue
-        if: ${{ github.event_name == 'merge_group' }}
-        shell: bash
-        run: echo "Skipping deja-art label check for merge queue"
-
+      # Labels are read via the API rather than from the event payload so
+      # that re-running this job after adding the label sees the current
+      # labels. A re-run reuses the original event payload, which would
+      # otherwise still report the label as missing.
       - name: Verify PR has the deja-art label
         if: ${{ github.event_name == 'pull_request' }}
         shell: bash
         env:
-          LABELS: ${{ toJSON(github.event.pull_request.labels.*.name) }}
+          GH_TOKEN: ${{ github.token }}
+          REPOSITORY: ${{ github.repository }}
+          PR_NUMBER: ${{ github.event.pull_request.number }}
         run: |
-          if jq --exit-status 'index("deja-art")' <<< "${LABELS}" > /dev/null; then
+          labels="$(
+            gh pr view "${PR_NUMBER}" \
+              --repo "${REPOSITORY}" \
+              --json labels \
+              --jq '.labels[].name'
+          )"
+
+          if echo "${labels}" | grep --quiet --line-regexp --fixed-strings 'deja-art'; then
             echo "'deja-art' label is present."
           else
-            echo "::error::Add the 'deja-art' label to this pull request to run the Deja ART replay."
+            echo "::error::Add the 'deja-art' label to this pull request to run the Deja ART replay, then re-run this check."
             exit 1
           fi
--- a/.github/workflows/deja-art-label-check.yml
+++ b/.github/workflows/deja-art-label-check.yml
@@ -1,0 +1,44 @@
+name: Deja ART Label Check
+
+on:
+  pull_request:
+    types:
+      - opened
+      - reopened
+      # Required status checks are evaluated per head commit, so the check
+      # must run on every push to report a result for the new commit.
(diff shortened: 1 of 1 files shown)


.github/workflows/deja-art-label-check.yml
```diff
@@ -0,0 +1,44 @@
...
+jobs:
+  deja_art_label_check:
+    name: Verify PR has the deja-art label
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Skip check for merge queue
+        if: ${{ github.event_name == 'merge_group' }}
+        shell: bash
+        run: echo "Skipping deja-art label check for merge queue"
```
pixincreate: imo, you don't even need this step...

.github/workflows/deja-art-label-check.yml
```diff
@@ -0,0 +1,44 @@
...
+      # Required status checks are evaluated per head commit, so the check
+      # must run on every push to report a result for the new commit.
+      - synchronize
+      - ready_for_review
+      - labeled
+      - unlabeled
+
+  merge_group:
+    types:
+      - checks_requested
```
pixincreate, line 13: also, this is unnecessary.

.github/workflows/deja-art-label-check.yml
```diff
@@ -0,0 +1,44 @@
...
+  pull_request:
+    types:
+      - opened
+      - reopened
+      # Required status checks are evaluated per head commit, so the check
+      # must run on every push to report a result for the new commit.
+      - synchronize
+      - ready_for_review
+      - labeled
+      - unlabeled
```
pixincreate: we really need all these?

.github/workflows/deja-art-label-check.yml
```diff
@@ -0,0 +1,49 @@
+name: Deja ART Label Check
+
+on:
+  # Runs on `opened`, `synchronize` and `reopened` (the default types).
+  # `synchronize` is required so that a result is reported for every head
+  # commit, since required status checks are evaluated per commit.
```
pixincreate, line 6: nit: you need not put these comments.

.github/workflows/deja-art-label-check.yml
```diff
@@ -0,0 +1,49 @@
...
+  pull_request:
+
+  # Required status checks must report a result in the merge queue as well,
+  # otherwise the queue entry stalls waiting for this check.
```
pixincreate, line 10: why do we want to run it in merge queue? this ci just checks the deja label, right?

maverox:
Closing this in favour of https://github.com/juspay/hyperswitch/pull/14181, which adds the same check in the same file.

Three pull requests were open for this gate — this one, https://github.com/juspay/hyperswitch/pull/13614 and https://github.com/juspay/hyperswitch/pull/14181 — and this one and 14181 share the head branch name `ci/deja-art-label-check` under two different authors, which is how they went unnoticed beside each other.

The choice between the two is close, and worth writing down, because this pull request is the better of the two in one respect.

What decided it is the trigger. 14181 fires on `labeled` and `unlabeled`, so adding the label produces a fresh event and the check re-runs on its own. This one uses the default trigger types, so adding the label fires nothing and the contributor has to re-run the check by hand every time. The common path should not need a manual step.

The respect in which this one is better is the label read. Reading through the API rather than from the event payload is correct, and the comment explaining why is right: a re-run replays the original payload, so a payload read still reports the label missing after it has been added. 14181 reads the payload, which is current on a `labeled` event but stale for anyone who clicks re-run instead of toggling the label. That idea is not lost with this pull request — it is left as a review comment on 14181, pointing back here.

PR #14106, opened 2026-09-08
fix(payments): populate business details in payment intent from merchant connector account and merchant account

### LINKED ISSUE #14162: [BUG] connector_label is never generated unless business_country and business_label are passed in the payment request
## Bug Description

`connector_label` is absent from payments responses for merchants who have their business details configured on the merchant account or the merchant connector account, rather than passing them on every payment request.

`connector_label` is derived from `business_country` and `business_label` on the payment intent, via `get_connector_label`, which zips the two together — so a missing value on either side yields no label at all.

Those two fields are only ever populated from the payment request:

- `PaymentCreate::make_payment_intent` copies `request.business_country` and `request.business_label` straight into the intent.
- The merchant account's `primary_business_details` is read only for validation (`validate_business_details`) and profile resolution (`get_profile_id_from_business_details`), never to populate the intent.
- The merchant connector account's `business_country` / `business_label` columns are never read anywhere in the payments flow.

The impact is not limited to a cosmetic response field: the same label is used as the key for storing `connector_customer` identifiers in `call_create_connector_customer_if_required`.

This matters because `business_country` and `business_label` are deprecated on the request in favour of `profile_id`, which is precisely why merchants stop sending them per request.

## Expected Behavior

When the business details are not supplied in the payment request, they should be sourced from stored configuration, so that `connector_label` is still generated. A sensible order of precedence is:

**request > merchant connector account > merchant account**

## Actual Behavior

The label is generated only when the values are passed directly in the payment request. Otherwise `business_country`, `business_label` and `connector_label` are all `null`.

## Steps To Reproduce

1. Create a merchant account with `"primary_business_details": [{"country": "US", "business": "default"}]`.
2. Create a connector account under it (e.g. the `phonypay` dummy connector) without `business_country` / `business_label`.
3. Create and confirm a payment **without** `business_country` or `business_label` in the request.
4. Observe `connector_label` in the response.

Measured against a local router on the current `main`:

```json
{
  "business_country": null,
  "business_label": null,
  "connector_label": null,
  "connector": "phonypay",
  "status": "succeeded"
}
```

```
# persisted payment_intent row
business_country=NULL  business_label=NULL
```

The label appears only when the request itself carries the pair. Repeating the same flow with `"business_country": "US", "business_label": "default"` in the request yields `connector_label: "phonypay_US_default"`.

## Context For The Bug

Reproduced in sandbox and observed in production.

## Environment

Self-hosted, built from source on `main`.

Sakilmostak:
## Type of Change

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

The `connector_label` returned in payments responses, and the key used for storing `connector_customer` identifiers, are both derived from `business_country` and `business_label` on the payment intent. `get_connector_label` zips the two together, so a missing value on either side yields no label at all.

These fields were only ever populated from the payment request:

- `PaymentCreate::make_payment_intent` copied `request.business_country` and `request.business_label` straight into the intent.
- The merchant account's `primary_business_details` was read only for validation (`validate_business_details`) and profile resolution (`get_profile_id_from_business_details`), never to populate the intent.
- The merchant connector account's `business_country` / `business_label` columns were never read anywhere in the payments flow.

As a result, merchants with these values configured on their merchant account or merchant connector account still received no `connector_label` unless they passed the values explicitly in every payment request.

Business details are now resolved in order of precedence: **request > merchant connector account > merchant account**.

- `MerchantAccount::get_business_details_for_profile` (in `hyperswitch_domain_models`) resolves the pair from `primary_business_details` for a given business profile, picking the entry whose `{country}_{business}` matches the profile name (profiles created from `primary_business_details` are named this way), or the only entry when exactly one is configured. Ambiguous configurations resolve to `None` rather than guessing.
- `MerchantConnectorAccountType::get_business_details` exposes the pair configured on the connector account.
- `PaymentIntent::resolve_business_details` applies the precedence for a set of request values, and `PaymentIntent::get_business_details_to_populate` / `PaymentIntent::populate_business_details` do the same for an existing intent.
- Payments create resolves the pair before constructing the intent, and update populates the intent, both falling back to the merchant account. The connector call path (`call_connector_service_prerequisites`) falls back to the merchant connector account first and then the merchant account, immediately after the connector account is fetched, so the existing intent update in confirm persists the values and the connector customer label picks them up.
- The pair is always resolved together, so a `connector_label` is never generated from a business country and business label that do not belong to each other. A request that sends only one of the two fields keeps the existing behaviour.

`PaymentUpdate::populate_payment_intent_with_request` also no longer clears the business details already on the intent when an update request omits them, keeping the in-memory intent consistent with what is stored. This is defensive rather than a bug fix: `PaymentIntentUpdateInternal` is a diesel `AsChangeset` without `treat_none_as_null`, so a `None` field leaves the column unchanged and the stored pair was never at risk. Verified on `main`: creating with `US/default` and then updating only the amount leaves the row, the update response and a subsequent retrieve all showing `US/default`.

### Additional Changes

## Motivation and Context

Closes #14162

`connector_label` is silently absent from payments responses for merchants who have configured their business details on the merchant account or merchant connector account rather than passing them on every request. The behaviour was reproduced in sandbox and confirmed in production. Since the label also keys `connector_customer` identifiers, the gap is not limited to a cosmetic response field.

`business_country` and `business_label` are deprecated in favour of `profile_id` on the request, which is precisely why merchants no longer pass them per request. Sourcing them from stored configuration keeps the derived label working for those merchants.

## How did you test it?

Verified against a locally running router using the `phonypay` dummy connector. Every payment below reached `succeeded`, so the labels come from completed payment flows rather than validation-only paths.

### Setup

```bash
# merchant whose business details are configured on the account
curl -X POST $BASE/accounts -H "api-key: $ADMIN_KEY" -d '{
  "merchant_id": "'$MID'",
  "merchant_name": "Demo",
  "primary_business_details": [{"country": "US", "business": "default"}],
  "return_url": "https://duck.com"
}'

# connector account carrying no business details of its own
curl -X POST $BASE/account/$MID/connectors -H "api-key: $API_KEY" -d '{
  "connector_name": "phonypay",
  "connector_account_details": {"auth_type": "HeaderKey", "api_key": "test_key"},
  "test_mode": true,
  "disabled": false,
  "payment_methods_enabled": [ ... card ... ]
}'
```

### The reported case: business details configured on the account, none sent in the request

```bash
curl -X POST $BASE/payments -H "api-key: $API_KEY" -d '{
  "amount": 6540,
  "currency": "USD",
  "confirm": true,
  "payment_method": "card",
  "payment_method_data": {"card": {"card_number": "4242424242424242", "card_exp_month": "10", "card_exp_year": "2030", "card_holder_name": "John Doe", "card_cvc": "123"}}
}'
```

```json
{
  "business_country": "US",
  "business_label": "default",
  "connector_label": "phonypay_US_default",
  "connector": "phonypay",
  "status": "succeeded"
}
```

```
# persisted payment_intent row
business_country=US  business_label=default
```

Running the identical setup and request against a binary built from `main` gives:

```json
{
  "business_country": null,
  "business_label": null,
  "connector_label": null,
  "connector": "phonypay",
  "status": "succeeded"
}
```

```
# persisted payment_intent row on main
business_country=NULL  business_label=NULL
```

The whole matrix was run against both binaries. On `main` every row above yields `connector_label: null` except the one where the request itself carries the pair, which matches the originally reported behaviour: the label only appears when the values are passed directly in the payment request.

### Full matrix

Three merchant setups were used: **A** with a single `primary_business_details` entry and a connector account carrying nothing; **B** ambiguous (two entries) with a profile whose name matches no entry, so only its connector account carrying `GB/shop` can supply the pair; and **C** ambiguous with a connector account carrying nothing, so no source can. The last column is the persisted `payment_intent` row.

| Scenario | `connector_label` | `payment_intent` |
| --- | --- | --- |
| A: create, no business details in request | – | `US / default` |
| A: confirm | `phonypay_US_default` | `US / default` |
| A: retrieve | `phonypay_US_default` | `US / default` |
| A: create + confirm in one call | `phonypay_US_default` | `US / default` |
| B: create, profile matches no entry | – | `NULL / NULL` |
| B: confirm, connector account supplies the pair | `phonypay_GB_shop` | `GB / shop` |
| B: retrieve | `phonypay_GB_shop` | `GB / shop` |
| B: create + confirm in one call | `phonypay_GB_shop` | `GB / shop` |
| B: request `US/food` vs connector `GB/shop` | `phonypay_US_food` | `US / food` |
| C: no source has the details | – | `NULL / NULL` |
| A: update amount only | – | `US / default` (retained) |

This exercises each position of the precedence chain: the request wins over the connector account (row 9), the connector account wins over the merchant account (rows 6 to 8), and the merchant account fills the gap (rows 1 to 4). Row 10 confirms the behaviour is unchanged when no source has the details, and the last row confirms an amount-only update leaves the resolved pair in place. `main` behaves the same way on that last row, so it is a guard rather than a fix.

`cargo +nightly fmt --all` and `cargo clippy -p router --all-targets -- -D warnings` both pass.

Worth noting for reviewers:

- On a retry to a different connector, the values already stored on the intent are kept, so the label combines the retried connector name with the original business details.
- Intents created before this change are populated on their next connector call, and only confirm persists that, so a plain retrieve of such an intent shows nothing until it is confirmed.

## Checklist


crates/router/src/core/payments/operations/payment_create.rs
```diff
@@ -2028,7 +2028,18 @@ impl PaymentCreate {
...
+        };
+
+        // If the business details are not passed in the request, populate them from the merchant
+        // account so that the connector label can be generated for the payment
+        helpers::populate_business_details_in_payment_intent(
+            &mut payment_intent,
+            None,
+            platform.get_processor().get_account(),
+            business_profile,
+        )?;
```
Sakilmostak: refrain from populating the data after payment_intent struct is created

crates/router/src/core/payments/helpers.rs
```diff
@@ -4888,6 +4888,103 @@ pub fn validate_business_details(
...
+        .clone()
+        .parse_value::<Vec<api_models::admin::PrimaryBusinessDetails>>("PrimaryBusinessDetails")
+        .change_context(errors::ApiErrorResponse::InternalServerError)
+        .attach_printable("failed to parse primary business details")?;
+
+    Ok(get_business_details_from_primary_business_details(
+        &primary_business_details,
+        &business_profile.profile_name,
+    ))
+}
```
Sakilmostak: let this be a impl function over merchant account

Sakilmostak: naming is also wrong, since it also takes the data from business profile

crates/router/src/core/payments/helpers.rs
```diff
@@ -4888,6 +4888,103 @@ pub fn validate_business_details(
...
+/// were not passed in the payment request. Refer to [`get_business_details_for_payment_intent`]
+/// for the order of precedence
+pub fn populate_business_details_in_payment_intent(
+    payment_intent: &mut PaymentIntent,
+    merchant_connector_account: Option<&MerchantConnectorAccountType>,
+    merchant_account: &domain::MerchantAccount,
+    business_profile: &domain::Profile,
+) -> RouterResult<()> {
+    if let Some((business_country, business_label)) = get_business_details_for_payment_intent(
+        payment_intent,
```
Sakilmostak: let them be impl over payment intent

crates/hyperswitch_domain_models/src/payments.rs
```diff
@@ -209,6 +209,104 @@ impl PaymentIntent {
...
+            return Ok((business_country, business_label));
+        }
+
+        let business_details = Self::get_fallback_business_details(
+            connector_business_details,
+            merchant_account,
+            business_profile,
+        )?;
+
+        Ok(business_details.unzip())
```
Sakilmostak: use match case instead of if else

crates/hyperswitch_domain_models/src/payments.rs
```diff
@@ -209,6 +209,104 @@ impl PaymentIntent {
...
+    > {
+        if self.business_country.is_some() || self.business_label.is_some() {
+            return Ok(None);
+        }
+
+        Self::get_fallback_business_details(
+            connector_business_details,
+            merchant_account,
+            business_profile,
+        )
```
Sakilmostak: use match cases

crates/hyperswitch_domain_models/src/payments.rs
```diff
@@ -209,6 +209,104 @@ impl PaymentIntent {
...
+            connector_business_details,
+            merchant_account,
+            business_profile,
+        )? {
+            self.business_country = Some(business_country);
+            self.business_label = Some(business_label);
+        }
+
+        Ok(())
+    }
```
Sakilmostak: rename it to set_business_details

crates/router/src/core/payments.rs
```diff
@@ -6154,6 +6154,22 @@ where
...
+            merchant_connector_account.get_business_details(),
+            platform.get_processor().get_account(),
+            business_profile,
+        )
+        .change_context(errors::ApiErrorResponse::InternalServerError)
+        .attach_printable("Failed to resolve business details for the payment")?
+    {
+        let mut payment_intent = payment_data.get_payment_intent().clone();
+        payment_intent.business_country = Some(business_country);
+        payment_intent.business_label = Some(business_label);
```
Sakilmostak: create setter over payment_intent for these

PR #14121, opened 2026-09-09
feat(diesel_models): observability schema and its own migration lineage

VenuMadhav2541:
## Type of Change

- [x] New feature

## Description

The alert manager writes its own operational state — alert configuration and alert history — into **the application's ClickHouse**, the same store it queries for payments analytics. This adds the schema that state moves to: Postgres, in a database the observability plane owns rather than `hyperswitch_db`.

Part of [#23153](https://github.com/juspay/hyperswitch-cloud/issues/23153); implements [#23398](https://github.com/juspay/hyperswitch-cloud/issues/23398).

Schema only. Nothing reads or writes these tables yet — the connection pool, the API routes and the alert manager's client are separate changes. Purely additive, and inert on merge.

### Models here, lineage under the crate

The models live in `diesel_models`, where models live. Nothing is paid for that: the `observability` crate already reaches this one through `hyperswitch_interfaces`, so the dependency and its feature matrix are already compiled in.

The **lineage** stays separate — `crates/observability/migrations` with `diesel_observability.toml` naming its own generated schema. The two lineages share no tables and no version table, and they target different databases: putting these tables in `hyperswitch_db` would repeat, in Postgres, exactly the boundary this move exists to end.

### Two properties worth knowing before changing a column

- **`json` and `jsonb` are not interchangeable here.** The columns holding values the dashboard sends already serialised are `json`, which preserves the bytes it was given. `jsonb` re-encodes and would hand back something different from what was saved.
- **`alerts_main` is separate from `alerts_intermediate`.** One records each announcement that was sent, the other holds current lifecycle state. Keeping them apart is what makes it possible to say whether an alert was delivered — see [#23304](https://github.com/juspay/hyperswitch-cloud/issues/23304).

### Two deliberate calls, both commented in the migration

- **`hyperswitch_alerts_app_user` is not carried across.** It holds local login accounts, both environments run with `localUsers: false`, and nothing reads it.
- **`merchants_alert_external_config` gets `PRIMARY KEY (name, product)`**, without which two rows can disagree about whether the same alert is enabled — and diesel cannot model a table with no key at all.

Timestamps are naive `TIMESTAMP` and identifiers are sized `VARCHAR`, matching the rest of the repository.

## Testing

Verified against Postgres 16:

- applies to an empty database; re-running is a no-op
- `redo` round-trips; `revert` leaves only the version table
- foreign keys cascade — deleting an announcement clears its lifecycle and instance rows
- the partial unique index on `alerts_dicts` rejects a second **enabled** row for the same `(name, key_)` while allowing a disabled one
- a full alert inserted and read back joined across definition, announcement, lifecycle and instance
- `diesel migration redo --locked-schema` passes, and was confirmed non-vacuous by deliberately editing the generated schema and watching it fail

`cargo check -p diesel_models` is clean under **both** `v1` and `v2` — the new module and its UUID columns compile on both sides of the feature matrix. `cargo +nightly fmt --all --check` is clean.

### Additional Changes

### DIFF
--- a/crates/diesel_models/src/lib.rs
+++ b/crates/diesel_models/src/lib.rs
@@ -37,6 +37,7 @@ pub mod mandate;
 pub mod merchant_account;
 pub mod merchant_connector_account;
 pub mod merchant_key_store;
+pub mod observability;
 pub mod organization;
 pub mod payment_attempt;
 pub mod payment_intent;
--- a/crates/diesel_models/src/observability.rs
+++ b/crates/diesel_models/src/observability.rs
@@ -1,0 +1,7 @@
+//! The observability plane's own state: alert configuration, lifecycle, and what was announced.
+//!
+//! These tables are not part of `hyperswitch_db`. They live in a database the observability plane
+//! owns, with its own migration lineage under `crates/observability/migrations` and its own diesel
+//! configuration in `diesel_observability.toml`.
+
+pub mod schema;
--- a/crates/diesel_models/src/observability/schema.rs
+++ b/crates/diesel_models/src/observability/schema.rs
@@ -1,0 +1,296 @@
+// @generated automatically by Diesel CLI.
+
+diesel::table! {
+    alerts_dicts (id) {
+        id -> Uuid,
+        #[max_length = 64]
+        name -> Varchar,
+        #[max_length = 255]
+        key_ -> Varchar,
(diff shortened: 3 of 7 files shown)


ThisIsMani (commented):
Please use diesel models instead.

ThisIsMani (approved):

PR #14125, opened 2026-09-09
fix(ucs): populate connector_order_reference_id and access_token on PostAuthenticate requests

### LINKED ISSUE #14152: fix(ucs): populate connector_order_reference_id and access_token on PostAuthenticate requests- #14125
fix(ucs): populate connector_order_reference_id and access_token on PostAuthenticate requests- #14125

shuklatushar226:
## Type of Change
- [x] Bugfix

## Description

Both `PostAuthenticate` UCS request builders hardcoded `connector_order_reference_id: None`. PayPal's `PostAuthenticate` addresses an order the connector already created (`GET v2/checkout/orders/{order_id}`) and reads that id from `PaymentFlowData.reference_id`, which UCS populates **solely** from that request field. With it unset, the call failed inside UCS before any connector call was made:

```
grpc_code: 3 (InvalidArgument)
error_code: MISSING_REQUIRED_FIELD
error_message: "Missing required field: reference_id (order_id)"
latency_ms: 0        <- PayPal never contacted
```

The order id was never actually missing — the sibling Authorize leg on the *same* request id sends the same value as `connector_order_id` and PayPal accepts it. Only this leg failed to send it.

This also explains why CI stayed green: the `PaymentMethodAuthenticationService_PostAuthenticate` fixture hardcodes `connector_order_reference_id`, supplying exactly the field the real caller never set.

## Root cause

`PaymentFlowData.reference_id` for this flow has one source: the request's `connector_order_reference_id`. The only in-repo producer of that field copies it from the **Authenticate** response, and PayPal declares `Authenticate`/`PreAuthenticate` as `not_implemented` — so it could never be populated for PayPal via that path.

## Changes

- Add `order_id` to `CompleteAuthorizeData` and `PaymentsPostAuthenticateData`, carried across the `TryFrom`.
- Populate it in `payments/transformers.rs`, gated on `requires_order_creation_before_payment` so connectors without a CreateOrder leg aren't handed an id that doesn't address an order.
- Prefer `order_id`, falling back to `connector_transaction_id`, in the UCS builders — for `connector_order_reference_id` on **both** PostAuthenticate builders (plain and external-vault-proxy) and for `connector_order_id` on CompleteAuthorize.
- Standalone 3DS authentication sets `order_id: None`; its `connector_transaction_id` is a 3DS server transaction id, not an order id.

## Blast radius before the fix

Began 2026-09-08. 100% of PayPal `PostAuthenticate` calls failed (42 events/24h). PayPal only; **shadow mode only, so no customer impact** — but the flow would fail outright once promoted out of shadow.

Regression risk is nil: across all connectors, only `paypal.rs` reads `resource_common_data.reference_id`, so populating the field is inert everywhere else.

## Testing

Verified end to end against **PayPal sandbox** (`api-m.sandbox.paypal.com`) with a locally built router talking to a live UCS instance.

Two real payments were driven through hyperswitch -> UCS -> PayPal. The outgoing PostAuthenticate requests carried:

```
"connector_order_reference_id":"1H999124A68377838"
"connector_order_reference_id":"0PC148328T096283L"
```

Each matches that payment's `payment_attempt.connector_transaction_id` (the PayPal-created order id). `MISSING_REQUIRED_FIELD` / `reference_id (order_id)` occurrences in the router log after the fix: **0**.

Controlled A/B against the same UCS instance:

| Request | Result |
|---|---|
| without `connector_order_reference_id` | `InvalidArgument MISSING_REQUIRED_FIELD "Missing required field: reference_id (order_id)"` |
| with the real order id | PayPal sandbox **HTTP 200** -> `{"payment_source":{"card":{"last_digits":"1111"}}}` |

`last_digits` matches the test card, confirming the id is real and usable. `cargo check --bin router` passes.

## Follow-up (not in this PR)

The same builders set `state: None`, so no `access_token` reaches UCS while PayPal's `get_headers` requires one — the next blocker for a live (non-shadow) PayPal PostAuthenticate.

## Checklist
- [x] `cargo +nightly fmt --all` / builds clean
- [x] Manually tested against connector sandbox


shuklatushar226:
Added a second commit fixing the **next blocker in the same two builders**.

### `state: None` → no access token reaches UCS

Both PostAuthenticate builders also hardcoded `state: None`, so no OAuth access token was sent. PayPal's PostAuthenticate `get_headers` requires `resource_common_data.access_token` and fails with `FAILED_TO_OBTAIN_AUTH_TYPE` without it.

This defect was **invisible in production** because UCS calls `get_url` before `get_headers` — every request died at the missing `connector_order_reference_id` first and never reached header construction. Both defects were present on the same requests; only the first was ever logged.

The token was already in scope: `add_access_token` runs before `decide_flows`, and `router_data_type_conversion` (`helpers.rs:5402`) carries `access_token` across the CompleteAuthorize → PostAuthenticate conversion. `state` is now populated exactly as the Authorize builder does it.

### End-to-end verification, both fixes together

Fresh payment driven through hyperswitch → UCS → **PayPal sandbox** on a locally built router:

```
payment_id                    pay_94hreiidhWzC0muBTpNs
PayPal order id               0011174513256425A
connector_order_reference_id  0011174513256425A     <- fix 1
PostAuthenticate response     AUTHENTICATION_SUCCESSFUL
final attempt status          authorized
```

Failure-signature scan of the router log for that run — all zero:

| signature | count |
|---|---|
| `MISSING_REQUIRED_FIELD` | 0 |
| `reference_id (order_id)` | 0 |
| `FAILED_TO_OBTAIN_AUTH_TYPE` | 0 |
| `InvalidArgument` | 0 |
| `Unauthenticated` | 0 |

Before these two commits the same flow failed at `MISSING_REQUIRED_FIELD`; with only the first commit it advanced to `FAILED_TO_OBTAIN_AUTH_TYPE`; with both it completes and the payment authorizes.

`cargo check --bin router` passes.

JeevaRamu0104 (approved):

JeevaRamu0104 (dismissed):

crates/router/src/core/payments/transformers.rs
```diff
@@ -7374,6 +7374,25 @@ impl<F: Clone> TryFrom<PaymentAdditionalData<'_, F>> for types::CompleteAuthoriz
             .get_connector_metadata_from_intent()
             .change_context(errors::ApiErrorResponse::InternalServerError)
             .attach_printable("Failed to parse connector metadata")?;
+        // The connector's order id survives the redirect in `connector_response_reference_id` only
+        // for connectors whose Authorize leg is preceded by a CreateOrder. Every other connector
+        // puts its own reference there (PayPal, for one, echoes back our
+        // `connector_request_reference_id`), so leave `order_id` unset for them rather than hand
+        // the connector an id that does not address an order.
+        let connector_creates_order = api_models::enums::Connector::from_str(connector_name)
+            .ok()
```
sai-harsha-vardhan: we may have to propagate the error

sai-harsha-vardhan (dismissed):

JeevaRamu0104 (approved):

JeevaRamu0104 (dismissed):

JeevaRamu0104 (dismissed):

JeevaRamu0104 (approved):

sai-harsha-vardhan (approved):

PR #14129, opened 2026-09-09
feat(observability): database configuration and connection pool

VenuMadhav2541:
## Type of Change

- [x] New feature

## Description

[#14121](https://github.com/juspay/hyperswitch/pull/14121) added the alert state schema, but the service had nothing to connect with. This adds the database section, the pool, and a readiness probe.

Implements [#23400](https://github.com/juspay/hyperswitch-cloud/issues/23400); part of [#23153](https://github.com/juspay/hyperswitch-cloud/issues/23153). Targets `observability-plane-i`. No routes that read or write yet: those are the API PRs.

### Configuration, as drainer has it

`[database]` defaults `host`/`port` to `localhost:5432`, `pool_size` to 5 and `connection_timeout` to 10 seconds. `dbname`, `username` and `password` must be set: validation rejects them empty with drainer's messages, and rejects a zero port, pool size or timeout (the last two would panic the pool builder).

### The password rides the existing secrets path

A `SecretStateContainer` alongside the auth key, resolved by the same transformer and validated again after decryption, so a well-formed secret handle that resolves to an empty string fails the boot rather than later as an authentication error.

### The connection string goes through `DbConnectionParams`

`DatabaseSettings` implements only the `common_utils::DbConnectionParams` getters, like drainer's and the router's database configs, and the pool uses the trait's `get_database_url` with the `public` schema (so `application_name` and `search_path` are `public`, as for router and drainer). Credentials are not percent-encoded, the same as those services.

### The pool is built without connecting

bb8 over `async-bb8-diesel` with `DejaPgConnection`, the same dependencies and versions as drainer. `build_unchecked` with no `min_idle`, so a database blip at boot never becomes a failed start or a crash loop; an instance that cannot reach its database reports unready instead.

### Liveness and readiness

`/health` is liveness and touches nothing. `/health/ready` answers `200` when a pooled connection is available within 900 ms and `503` ("Observability database is unreachable") otherwise; the connection failure goes to the log.

## Testing

- `cargo +nightly fmt --all --check`; `cargo clippy -p observability --all-targets -- -D warnings` and `-p diesel_models --features v1`; `cargo check -p diesel_models --features v1,deja`; the v2 check with the workspace's v2 feature set; `cargo test -p observability` (the tests on the base branch pass).
- Against Postgres 16, service started for each case (42 checks):
  - no `[database]` section, or an empty dbname / username / password → boot exits non-zero with the drainer message and nothing listens
  - zero pool size, empty host, zero port or zero timeout → boot fails with the matching message
  - only credentials set → `/health` `200`, `/health/ready` `200`, delivery routes unaffected
  - unreachable database → `/health` `200`, `/health/ready` `503` within about 940 ms
  - reachable database → `/health/ready` `200`, and `pg_stat_activity` shows `application_name=public`

### Additional Changes

### DIFF
--- a/crates/observability/src/routes/health_check.rs
+++ b/crates/observability/src/routes/health_check.rs
@@ -1,15 +1,67 @@
 //! Health handlers. The route tree that mounts them is in [`crate::routes::app`].
 //!
-//! Liveness only. A readiness check that dials the chat provider and the mail backend sounds
-//! thorough and turns every third-party blip into a restart loop; the point of this service is to
-//! be up when its dependencies are flaky.
+//! Two probes, answering different questions.
+//!
+//! [`health`] is **liveness**: is the process running. It touches nothing. A liveness check that
+//! dialled the chat provider and the mail backend would sound thorough and turn every third-party
+//! blip into a restart loop; the point of this service is to be up when its dependencies are
+//! flaky.
+//!
+//! [`deep_health_check`] is **readiness**: can this instance do its job. It checks the database and
+//! only the database — chat and email are destinations this service delivers *to*, and a failing
+//! one is already reported per request, but without its own state store there is nothing to serve.
+//!
+//! The split is what lets the pool be built without connecting: a database briefly away makes an
+//! instance unready, which stops traffic, rather than dead, which restarts it. Mirrors the router's
+//! `/health` and `/health/ready` pair.
 
+use std::time::Duration;
+
+use actix_web::web;
 use router_env::{instrument, tracing};
 
-use crate::logger;
+use crate::{logger, state::AppState};
+
+/// How long the probe waits for a connection before answering unready.
+///
+/// Deliberately shorter than the pool's own connection timeout. `Pool::get` waits that full budget
+/// before giving up, so a probe inheriting it answers long after a `timeoutSeconds: 1` readiness
+/// check has already given up and recorded nothing — the 503 would never be seen.
+const PROBE_BUDGET: Duration = Duration::from_millis(900);
 
 #[instrument(skip_all)]
 pub async fn health() -> impl actix_web::Responder {
     logger::info!("Observability health was called");
     actix_web::HttpResponse::Ok().body("Observability health is good")
+}
+
+/// Report whether this instance can reach the observability database.
+///
+/// Answers `503` rather than `500` when it cannot: the service itself is fine, and the condition is
+/// expected to clear without intervention. The reason reaches the log and not the body — a probe
+/// response is read by anyone who can reach the port, and a connection error names the host, the
+/// database and the role.
+#[instrument(skip_all)]
+pub async fn deep_health_check(state: web::Data<AppState>) -> impl actix_web::Responder {
+    match tokio::time::timeout(PROBE_BUDGET, state.database.get()).await {
+        Err(_elapsed) => {
+            logger::error!("Observability deep health check timed out taking a connection");
+            return actix_web::HttpResponse::ServiceUnavailable()
+                .body("Observability database is unreachable");
+        }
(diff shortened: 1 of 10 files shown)


VenuMadhav2541:
Force-pushed: `cargo +nightly fmt --all` had not been run, and CI's `fmt --all --check` would have failed on 9 diffs. Formatting only — no behaviour change, 55 tests still green.

VenuMadhav2541:
Rebased onto `observability-plane-i` now that #14121 has merged, and brought in line with it: `hyperswitch_alerts` renamed to `observability` throughout, including the default `dbname`, and the doc comments cut back.

Re-verified after the rename: 55 tests green, `cargo +nightly fmt --all --check` clean, booted against an unreachable database (stays up, `/health` 200, `/health/ready` 503) and against a real `observability` database (both 200, `application_name=observability` visible in pg_stat_activity).

VenuMadhav2541:
Pushed fixes for the review findings:

**Credentials are percent-encoded.** A generated password containing `/`, `@`, `?` or `#` previously produced a connection string where libpq read the host, database and query parameters from the wrong side of the character. Verified against a real role with the password `Rk9/pa@ss?x` — `/health/ready` returns 200.

**`connection_timeout = 0` is rejected at boot.** It used to reach `bb8::Builder::connection_timeout`, whose `assert!` names neither the setting nor the file. Now: `ConfigParsingError("database connection_timeout must be greater than zero")`.

**`Default` is hand-written** so it agrees with the `#[serde(default = ...)]` attributes, the same reason `ChatSettings` does. Deriving it produced `pool_size: 0` / `connection_timeout: 0` — values the serde defaults exist to prevent and which bb8 asserts against.

**The pool is typed on `DejaPgConnection`**, not `diesel::PgConnection`. Without the `deja` feature they are the same type; with it they are not, and every `diesel_models` query helper takes the alias — so as it stood, the pool could not have been used with the schema it exists to serve.

55 tests green, `cargo +nightly fmt --all --check` clean.

Still open from the review and not addressed here: the readiness probe cannot report *why* the database is unreachable (bb8 0.8.6 discards connect errors through a `NopErrorSink` and only ever returns `TimedOut`), and it burns the full connection timeout before answering, so a probe at `timeoutSeconds: 1` never sees the 503. That needs `Builder::error_sink` plus a bounded probe timeout — I would rather do it as its own change than widen this one.

crates/observability/src/routes/health_check.rs
```diff
@@ -1,15 +1,35 @@
...
+            logger::info!("Observability deep health check passed");
+            actix_web::HttpResponse::Ok().body("Observability deep health is good")
+        }
+        Ok(Err(_)) | Err(_) => {
+            logger::error!("Observability deep health check failed: database unreachable");
+            actix_web::HttpResponse::ServiceUnavailable()
+                .body("Observability database is unreachable")
+        }
+    }
+}
```
ThisIsMani, line 35: Can we keep this similar to router.

crates/observability/src/state.rs
```diff
@@ -171,6 +181,31 @@ async fn build_email_registry(
...
+    let manager = async_bb8_diesel::ConnectionManager::<DejaPgConnection>::new(
+        database.get_database_url(DATABASE_SCHEMA),
+    );
+
+    bb8::Pool::builder()
+        .max_size(database.pool_size)
+        .connection_timeout(std::time::Duration::from_secs(database.connection_timeout))
+        .error_sink(Box::new(LogConnectionErrors))
+        .build_unchecked(manager)
+}
```
ThisIsMani, line 207: This as well.

ThisIsMani (commented):
Take reference from router.

PR #14131, opened 2026-09-09
refactor(storage): find_config_by_key -> find_config_by_key_optional

### LINKED ISSUE #14229: [REFACTOR] find_config_by_key should not bypass cache on not-found
### Problem

`ConfigInterface::find_config_by_key` returns `CustomResult<Config, Error>`, erroring with `ValueNotFound` whenever the key is absent. Because the caching helper (`cache::get_or_populate_in_memory`) only writes an entry back to Redis/in-memory cache on `Ok`, a lookup for a config key that legitimately doesn't exist in the DB **never gets cached** — every single call re-hits Postgres, even when the same missing key is queried repeatedly (e.g. per-connector feature-flag or rollout-config checks that run on the hot payments path).

This also forces every call site to distinguish "real DB error" from "not found" by pattern-matching on `err.current_context().is_db_not_found()`, which is easy to get wrong and duplicated across ~20 call sites.

Separately, the trait carries some dead surface:
- `find_config_by_key_from_db` duplicates `find_config_by_key` but skips the cache entirely, with no clear reason left in the code for why some callers bypass it.
- `update_config_in_database` is a standalone trait method used by exactly one caller (`update_config_by_key`).
- `find_config_by_key_unwrap_or`'s `default_config: Option<String>` is always called with `Some(..)` — the `None` branch is unreachable dead code.

### Proposed change

- Rename `find_config_by_key` → `find_config_by_key_optional`, returning `Option<Config>` instead of erroring on not-found.
- Make the DB-not-found case map to `Ok(None)` before it reaches `cache::get_or_populate_in_memory`, so absent keys get cached like present ones.
- Fold `find_config_by_key_from_db` into `find_config_by_key_optional` so there's a single, always-cached read path.
- Inline `update_config_in_database` into `update_config_by_key`.
- Change `find_config_by_key_unwrap_or`'s `default_config` param from `Option<String>` to `String`.
- Update all call sites to match on `Option<Config>` instead of `Result`, preserving existing error-handling/logging behavior.

### Additional context

Tracking issue for #14131.

hrithikesh026:
Fixes #14229

## Summary
- Renames `ConfigInterface::find_config_by_key` to `find_config_by_key_optional`, returning `Option<Config>` instead of erroring with `ValueNotFound` when a key is absent.
- The `RouterStore` implementation now converts a DB not-found into `Ok(None)` before handing it to `cache::get_or_populate_in_memory`. Since `Option<T>` already satisfies the `Cacheable` bound, the "not found" result is now cached in Redis + the in-memory cache too — previously every lookup of a missing key re-hit Postgres because only `Ok` values were ever written back to the cache.
- Removes `find_config_by_key_from_db`, folding its logic into `find_config_by_key_optional` so there is a single, always-cached read path (the few call sites that bypassed the cache now go through it too).
- Removes `update_config_in_database` as a standalone trait method, inlining it directly into `update_config_by_key`'s `publish_and_redact` closure — it had no other caller.
- Changes `find_config_by_key_unwrap_or`'s `default_config` parameter from `Option<String>` to `String`: every call site already passed `Some(..)`, so the `None` branch (bubble the not-found error instead of defaulting) was dead code.
- Adds a doc comment on `ConfigInterface` documenting that every method is expected to go through the caching layer, and that direct DB access is intentionally not exposed on the trait.
- Updates all call sites across `router` and `subscriptions` (~20 sites) to handle `Option<Config>`, preserving each site's existing error-handling and logging behavior (not-found vs. real DB error) exactly as before — `Err(e) if e.current_context().is_db_not_found()` guards become `Ok(None)` match arms.

## Test plan
- [x] `cargo check --workspace` (default `v1` feature set) — compiles clean
- [x] `cargo check -p router --no-default-features --features "v2,common_default,olap,revenue_recovery,redis-rs"` — compiles clean
- [ ] `cargo test -p storage_impl configs`
- [ ] `cargo test -p router configs`
- [ ] Manual smoke test: call `read_config` for a missing key twice, confirm the second call is served from cache instead of hitting Postgres again

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>

https://claude.ai/code/session_015xsTJfcXQFUnfmpcCbWbVw


jarnura (dismissed):

crates/hyperswitch_domain_models/src/configs.rs
```diff
@@ -9,7 +9,10 @@ pub trait ConfigInterface {
         config: storage::ConfigNew,
     ) -> CustomResult<storage::Config, Self::Error>;
 
-    async fn find_config_by_key(&self, key: &str) -> CustomResult<storage::Config, Self::Error>;
+    async fn find_config_by_key_optional(
```
kashif-m, line 12: nit:
```suggestion
    async fn find_optional_config_by_key(
```

kashif-m (dismissed):

ThisIsMani (dismissed):

ThisIsMani (dismissed):

prajjwalkumar17 (dismissed):

jarnura (dismissed):

AmitsinghTanwar007 (dismissed):

jarnura (approved):

ThisIsMani (approved):

AmitsinghTanwar007 (approved):

prajjwalkumar17 (approved):

prasunna09 (approved):

PR #14140, opened 2026-09-09
feat(resources): org-scoped, reusable Apple Pay certificate management

ShankarSinghC:
## Summary
- Adds a generic, org-scoped `resources` table so a merchant can generate an Apple Pay Payment Processing Certificate once and link it across multiple merchant connector accounts, business profiles, or the merchant account itself — fallback resolution walks MCA -> profile -> merchant_account.
- New endpoints (all org-scoped via the authenticated API key/JWT, no `merchant_id` in the path): `POST /resources` (generate), `PUT /resources/{resource_id}` (upload signed certificate — accepts either JSON `{"certificate": "<base64 DER>"}` or the raw `.cer` file bytes directly), `POST /resources/list` (list + resolve effective link for a given entity), `POST /resources/{resource_id}/link` (link to an entity).
- Private key is encrypted at rest under a per-organization key, reusing the existing merchant-key-store machinery (the org id is re-wrapped as a `MerchantId` for this purpose — zero keymanager-service changes needed).
- A decrypt-time cache (`apple_pay_certificates` / `apple_pay_certificates_encrypted`) is written onto `merchant_account`/`business_profile`/`merchant_connector_account` at link time, so the actual payment-decrypt path resolves the linked certificate without a `resources`-table lookup.
- Wires the resolved, active certificate into the existing Apple Pay decrypt-at-payment path (`decrypt_apple_pay_wallet_data`); falls through unchanged to the existing manual MCA-metadata path when no managed certificate is linked or not yet active — no behavior change for merchants who don't use this feature.
- Scope: v1 only. The v2 build's resource-linking storage layer compiles cleanly but returns a clean "not supported" error at runtime, since v1's and v2's underlying entity schemas (column names, nullability) differ enough that porting the fallback-chain logic to v2 is a separate task.

## Test plan
- [x] `cargo check`/`just clippy` (v1 default features) — clean
- [x] `just clippy_v2` — clean
- [x] Live-tested end to end against a local Postgres/Redis stack: generate -> upload (both JSON and raw-binary-body upload) -> link -> list, including cross-org rejection (403), not-found (404), and cross-merchant-within-org sharing
- [x] Live-tested the full decrypt-at-payment path with a real Apple-issued certificate and a real Apple Pay payment token — confirmed the payment response contains genuinely decrypted card data via the managed-certificate path
- [ ] Reviewer smoke test against a fresh dev DB (new migrations: `resources` table, cache columns on the three entity tables)


XyneSpaces:
[blocking] Add Diesel migrations for the new `resources` table and the `apple_pay_certificates` / `apple_pay_certificates_encrypted` columns on merchant accounts, business profiles, and connector accounts. This PR updates the Diesel schemas and queries but includes no migration SQL, so a fresh deployment will fail on the first resource create/link request with missing-table or missing-column errors.

crates/router/src/core/payments/transformers.rs
```diff
@@ -2067,11 +2067,33 @@ where
         None
     };
 
+    let business_profile = match payment_data.payment_intent.profile_id.as_ref() {
+        Some(profile_id) => state
```
jarnura: No database read when already the data present in parent scope

crates/router/src/core/payments/transformers.rs
```diff
@@ -2394,11 +2416,33 @@ pub async fn construct_payment_router_data_for_update_metadata<'a>(
...
+            .store
```
jarnura: No database read when already the data present in parent scope

crates/router/src/core/payments.rs
```diff
@@ -9190,24 +9213,109 @@ async fn get_feature_data(
...
+    #[cfg(feature = "v1")]
+    if let Some(payment_processing_details) =
+        resolve_managed_apple_pay_certificate(&ApplePayCertificateAccounts {
+            processor: _processor,
+            business_profile: _business_profile,
+            merchant_connector_account: mca,
+        })
+        .await
+    {
+        return Some(domain::ApplePayFlow::DecryptAtApplication(
```
jarnura: No early return with data

migrations/2026-09-09-100000_create_resources_table/up.sql
```diff
@@ -0,0 +1,13 @@
+CREATE TABLE resources (
+    id VARCHAR(64) NOT NULL PRIMARY KEY,
+    resource_type VARCHAR(64) NOT NULL,
+    scope VARCHAR(32) NOT NULL,
+    scope_id VARCHAR(64) NOT NULL,
+    data JSONB NOT NULL DEFAULT '{}'::JSONB,
+    encrypted_data BYTEA,
+    created_by TEXT NOT NULL,
+    created_at TIMESTAMP NOT NULL DEFAULT now(),
+    modified_at TIMESTAMP NOT NULL DEFAULT now()
```
jarnura: application should always pass this data

migrations/2026-09-09-100000_create_resources_table/up.sql
```diff
@@ -0,0 +1,13 @@
+CREATE TABLE resources (
+    id VARCHAR(64) NOT NULL PRIMARY KEY,
+    resource_type VARCHAR(64) NOT NULL,
+    scope VARCHAR(32) NOT NULL,
+    scope_id VARCHAR(64) NOT NULL,
+    data JSONB NOT NULL DEFAULT '{}'::JSONB,
```
jarnura: lets have null as null only instead of empty.

jarnura (changes requested):

VenuMadhav2541 (dismissed):
Analytics LGTM

jarnura (dismissed):

ThisIsMani (dismissed):
Patch file changes looks fine.

jarnura (dismissed):

VenuMadhav2541 (dismissed):

Sakilmostak (dismissed):

VenuMadhav2541 (approved):

Sakilmostak (approved):

jarnura (approved):

PR #14146, opened 2026-09-09
feat(observability): alert definition and enablement configuration API

VenuMadhav2541:
Implements [#23401](https://github.com/juspay/hyperswitch-cloud/issues/23401) and [#23402](https://github.com/juspay/hyperswitch-cloud/issues/23402); part of [#23153](https://github.com/juspay/hyperswitch-cloud/issues/23153).

**Stacked on #14129** — review that first; this targets its branch.

## Routes

All under the existing `/alerts` scope, behind the internal API key (`401 IR_01` without it), mounted through `routes::AlertsConfig::server()`. Handlers, core and wire types sit next to the delivery ones: `routes/config.rs`, `core/config.rs`, `types/config.rs`.

| Method | Path | Answer |
|---|---|---|
| `GET` | `/alerts/config/definitions` | `{count, definitions}` |
| `POST` | `/alerts/config/definitions` | the created definition |
| `GET` / `POST` | `/alerts/config/definitions/{id}` | the definition / the updated definition (partial) |
| `GET` | `/alerts/config/enablement` | `{count, enablements}` |
| `GET` / `POST` | `/alerts/config/enablement/{name}/{product}` | the row / the upserted row |
| `GET` | `/alerts/config/merchant-thresholds` | `{count, merchant_thresholds}`, optional exact-match filters `name`, `product`, `merchant_id`, `is_enabled`, `author` |
| `POST` | `/alerts/config/merchant-thresholds` | the upserted row |
| `GET` / `POST` / `DELETE` | `/alerts/config/merchant-thresholds/{id}` | the row / the updated row (partial) / `{id, deleted: true}` |

`POST` rather than `PUT` for updates, because bodies are partial by design. Definitions have no delete route — `is_enabled` is how an alert is turned off.

## Definitions store r-apps' own documents

`blacklist`, `thresholds` and `snooze` are `json` columns carried as raw JSON and returned byte for byte, each holding exactly the document r-apps reads:

- `blacklist` — a list or object of groups, each mapping dimensions to a value or list of values (r-apps `flow_utils.R` blacklist clause builders; `config_backup.R`).
- `thresholds` — one object per definition (`thresholds.min_volume`, `tolerance`, …).
- `snooze` — an object keyed `snooze_entry_<time>` / `custom_snooze_entry_<time>`, each entry with the dimension values it covers and `snooze_end_time` (optionally `snooze_start_time`) as `YYYY-MM-DD HH:MM:SS` IST.

Each type checks its own shape before a connection is taken (a wrong shape is `400 IR_06` naming the field). An empty list, empty object or blank string is stored as `{}`, as r-apps' `createAlertInfo` does, and the same applies to the free-form `metadata`; `null` stores NULL.

An update names only the fields it changes: absent is left alone, `null` clears, a value sets (`Option<Option<T>>` via `serde_with`'s `double_option`). The request converts into `AlertsInfoUpdate` → `AlertsInfoUpdateInternal`, which sets `last_updated_at`, the shape `RefundUpdate` uses. `name`, `product` and `author` are required and must not be blank; `name` and `product` cannot change. Ids come from `common_utils::generate_uuid_v7()`.

## Two switches

As in r-apps, the definition's `is_enabled` decides whether a detector runs, and the enablement row decides whether its alerts reach merchants; a missing row delivers nothing. Every enablement response carries `effective_is_enabled = definition.is_enabled AND coalesce(enablement.is_enabled, false)`, read with a select of `name, product, is_enabled` only. The enablement upsert is one `INSERT … ON CONFLICT (name, product) DO UPDATE`; on an existing row `category` and `metadata` change only when mentioned. An enablement for a name/product with no definition, or for the reserved `all` definition, is refused.

## Merchant thresholds

Per-merchant thresholds live in r-apps' own `merchant_thresholds` table, added in a new migration directory (`2026-09-14-000001_create_merchant_thresholds_table`) with the columns of r-apps `alerts/api/config.R`: `name`, `product`, `merchant_id`, ten nullable threshold numbers, `metadata`, `author`, `is_enabled`, `last_updated_at`. The unique index is r-apps' `addMerchantThresholds` conflict key `(name, product, merchant_id, is_enabled, author)`, and the upsert targets it; an absent or `null` threshold keeps the stored value. Updates and deletes address a row by id (r-apps' key is five columns, so name/product/merchant_id can match several rows); `metadata` must be an object, and an update merges it into the stored object. r-apps' `metadata || patch` leaves a NULL metadata NULL and drops the patch; here the patch is stored.

## Errors

| | Status | Code |
|---|---|---|
| Body does not parse (shared `json_config` on the base) | 400 | `IR_04` |
| A stored document in the wrong shape; merchant-threshold `metadata` not an object (upsert or update); filter query does not parse | 400 | `IR_06` |
| A value wider than its column, or a blank `name`, `product` or `author` (and `merchant_id` for merchant thresholds) | 400 | `IR_07` |
| Definition exists for this name and product; merchant-threshold update collides with another row | 400 | `HE_01` |
| Unknown definition, enablement or merchant-threshold | 404 | `HE_02` |
| Enablement for a name/product with no definition, or for `all` | 400 | `HE_03` |
| A query failed | 500 | `HE_00` |
| No database connection could be taken | 503 | `HE_00` |

Widths, blanks and shapes are checked before a connection is taken. A path segment that does not parse (a malformed id) gets the same empty `404` as a route that does not exist, through a `web::PathConfig` handler on `/alerts`.

## Schema

The `(name, product)` unique index is created in the lineage's create migration, which has not run anywhere yet; the new table gets its own migration.

## Testing

- `cargo +nightly fmt --all --check`; `cargo clippy -p observability --all-targets` and `-p diesel_models --features v1 --all-targets` with `-D warnings`; `cargo check -p diesel_models --features v1,deja`; the v2 check with the workspace's v2 feature set; `cargo test -p observability` (the tests on the base pass).
- Fresh database: both migrations run, `redo --all` clean, `diesel print-schema` matches `schema.rs`.
- Against the running service, 120 checks on definitions, enablement and merchant thresholds: r-apps documents stored and returned byte for byte, empty documents stored as `{}`, wrong shapes `IR_06`, widths and blanks `IR_07` without touching the database, duplicate `HE_01`, unknown ids `HE_02`, malformed ids an empty `404`, partial updates and `null` clearing, the switch precedence, merchant-threshold upsert on the five-column key, filters, metadata merge and delete; and 22 checks on snooze and the two switches.

### Additional Changes


crates/diesel_models/src/query/observability/alerts_info.rs
```diff
@@ -0,0 +1,102 @@
...
+            query.get_results_async(conn.raw_connection()),
+        )
+        .await
+        .attach_printable("Failed to list the alert definitions")
+        .switch()
+    }
+
+    pub async fn list_is_enabled(
+        conn: &DatabaseConnectionWithContext<'_>,
+    ) -> StorageResult<Vec<(String, String, Option<bool>)>> {
```
ThisIsMani: Take is_enabled as filter in above query.

crates/diesel_models/src/query/observability/alerts_info.rs
```diff
@@ -0,0 +1,102 @@
...
+            generics::db_metrics::DatabaseOperation::FindOne,
+            query.get_result_async(conn.raw_connection()),
+        )
+        .await
+        .optional()
+        .attach_printable("Failed to find whether the alert definition is enabled")
+        .switch()
+    }
+
+    pub async fn list(conn: &DatabaseConnectionWithContext<'_>) -> StorageResult<Vec<Self>> {
```
ThisIsMani: Is pagination required for this? How large can this table grow?

PR #14150, opened 2026-09-10
feat(external_services): add a metrics provider and its CloudWatch backend

ThisIsMani:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

Adds `metrics_service` to `external_services`: a `MetricsProvider` trait for reading metric
datapoints, and `aws_cloudwatch`, a `GetMetricData` backend for it. It fetches datapoints and
nothing else — no thresholds, no severities, no alarm state. The evaluator that reads alarms out of
these datapoints is a separate change.

Same move [#13968](https://github.com/juspay/hyperswitch/pull/13968) made for the chat client: the
transport lands on its own with its own tests, and the consumer is built against it afterwards.

### The interface is not CloudWatch's

Amazon calls the key/value selectors *dimensions*, its aggregations *statistics*, and groups
metrics into *namespaces*; Google and Prometheus both say *labels*, and Prometheus has no namespace
at all. The trait takes `Labels`, `Aggregation` and an optional namespace, so a second provider is
a translation rather than a rewrite. Two deliberate consequences:

- **Metric math stays off the trait.** `ANOMALY_DETECTION_BAND` and expression syntax are
  CloudWatch's query language. A vendor-neutral type that accepts a CloudWatch formula string is
  not vendor-neutral, it only looks it. The three anomaly alarms that need it are
  [#23452](https://github.com/juspay/hyperswitch-cloud/issues/23452)'s problem.
- **A trait is the intersection of its implementations.** One time range covers the whole request
  because that is what `GetMetricData` accepts, even where another provider would allow more.
  Recorded in the module docs as a known limitation rather than left as a surprise.

### Gaps are not zeroes

The load-bearing property of the return type. CloudWatch reports a period with no data by
*omitting its timestamp*, so a client that flattened the response into a `Vec<f64>` would turn a
metric that stopped reporting into a run of zeroes — and a metric that stopped reporting is exactly
what an infrastructure alarm exists to catch. `treat_missing_data` is set on 47 severities in the
existing Terragrunt catalogue, so the distinction has to survive to the consumer.

Datapoints keep their timestamps, nothing is defaulted or filled, and
`MetricSeries::window(end, period, count)` lays them back onto a period grid as `Option<f64>` —
`None` for a gap, `Some(0.0)` for a real zero. That helper lives here rather than in each caller
because grid alignment is precisely where an `unwrap_or(0.0)` gets typed by accident.

### Query ids are ours, not the caller's

`GetMetricData` requires ids matching `[a-z][a-zA-Z0-9_]*` and rejects the **whole batch** if one is
malformed — so a single bad id would blind all 76 metrics for that tick. That is CloudWatch's
constraint rather than a property of reading metrics, so it stays out of the trait: ids are minted
from each query's position and series come back keyed by that position.

### Pagination is the caller's

`fetch` returns one page plus an opaque `Cursor`; the caller feeds it back for the next. Nothing
loops on the caller's behalf, so nothing accumulates an unbounded response in memory without it
having asked page by page. A single series can span pages, so a caller that does paginate must
concatenate datapoints under the same index — documented, and in practice one page always
suffices, since what fills a page is datapoint count rather than elapsed time.

### Credentials and egress

The default provider chain, exactly as `aws_kms` and `file_storage::aws_s3` use it — in the cluster
that resolves to the pod's IRSA role. Deliberately *not* `email::ses`'s explicit assume-role on
every call, which [#23199](https://github.com/juspay/hyperswitch-cloud/issues/23199) records as a
problem rather than a pattern to copy.

No proxy. The sandbox VPC's `vpc-network` terragrunt sets `enable_vpc_endpoints = true` with
`monitoring` among its interface endpoints, so CloudWatch is reached privately inside the VPC;
squid exists for public-internet destinations such as Xyne.
[#23442](https://github.com/juspay/hyperswitch-cloud/issues/23442) therefore only needs to add the
IAM grant, not open egress.

### Additional Changes

No config file changes: `CloudWatchConfig` is a type the consumer will wire up in
[#23445](https://github.com/juspay/hyperswitch-cloud/issues/23445). Nothing in `config/`,
`crates/router/src/configs` or `loadtest/config` is touched.

**One dependency change worth a reviewer's attention.** The ticket asked for this to be
feature-gated like `aws_kms`/`aws_s3`; it is **not**, by decision during design. That required
`aws-config` to become a non-optional dependency of `external_services`, so `aws_kms`, `email` and
`aws_s3` no longer name it (`email` is now an empty feature that still gates its module), and
`router`/`drainer`/`scheduler` now compile `aws-config` and `aws-sdk-cloudwatch` where they
previously compiled neither. Reversible in one line by gating the `aws_cloudwatch` submodule if the
build cost proves annoying.

## Motivation and Context

Infrastructure alerting for Hyperswitch is moving into the observability plane: it queries
CloudWatch metrics, evaluates thresholds it owns, and delivers through the Notifier. This PR is the
transport that makes that possible — the evaluator has something to call.

Tracked by [hyperswitch-cloud#23443](https://github.com/juspay/hyperswitch-cloud/issues/23443),
under the map at
[hyperswitch-cloud#23440](https://github.com/juspay/hyperswitch-cloud/issues/23440). Closing the
ticket is manual: `Resolves` does not work across repositories.

## How did you test it?

21 unit tests over the translation in both directions, built on the SDK's own types
(`cargo test -p external_services --features v1 metrics_service`). The gap-versus-zero distinction
is asserted three separate ways; a forbidden or errored series is asserted to arrive as *data*
rather than as a failed call, so one mis-scoped IAM permission cannot blind the other 75 metrics.
Also checked that `email`, `aws_kms` and `aws_s3` still compile after the `dep:aws-config` removals.

Separately, and **not part of this diff**, the request was driven over a real HTTP round trip
locally against a loopback server to confirm the SDK serialises what we intend. The body it puts on
the wire:

```
Action=GetMetricData&Version=2010-08-01
&MetricDataQueries.member.1.Id=m0
&MetricDataQueries.member.1.MetricStat.Metric.Namespace=AWS%2FRDS
&MetricDataQueries.member.1.MetricStat.Metric.MetricName=Widgets
&MetricDataQueries.member.1.MetricStat.Metric.Dimensions.member.1.Name=Host
&MetricDataQueries.member.1.MetricStat.Metric.Dimensions.member.1.Value=host-a
&MetricDataQueries.member.1.MetricStat.Period=60
&MetricDataQueries.member.1.MetricStat.Stat=Maximum
&StartTime=2026-09-09T12%3A00%3A00Z&EndTime=2026-09-09T12%3A04%3A00Z
&ScanBy=TimestampAscending
```

A canned CloudWatch XML response parsed back with the gap intact
(`[Some(4.0), Some(0.0), None, Some(7.0)]`). That test needed an endpoint override, which is
deliberately kept out of the shipped code so no test-only knob appears in the public surface — so
it was run locally and reverted rather than committed.

**This has still not been run against real CloudWatch.** No sandbox credentials are available, so
what remains unproven is whether AWS *accepts* the request semantically — permissions, quotas, and
anything the service validates beyond its wire format. The sandbox deploy is the first such
exercise.

## Checklist

`cargo clippy -p external_services --features v1 --all-targets` reports nothing from this change;
the one warning in the crate's tree is pre-existing in `common_utils/src/keymanager.rs`.

https://claude.ai/code/session_01FFFd9wBmVQWbcAfCURRVYG


crates/external_services/src/metrics_service.rs
```diff
@@ -0,0 +1,675 @@
...
+    pub const ONE_MINUTE: Self = Self(60);
+    /// Five minutes.
+    pub const FIVE_MINUTES: Self = Self(300);
+    /// One hour.
+    pub const ONE_HOUR: Self = Self(3600);
+
+    /// Build a period from a number of seconds.
+    ///
+    /// Sub-minute periods are accepted only at the resolutions high-resolution metrics support
+    /// (1, 5, 10, 20 and 30 seconds); anything else must be a positive multiple of 60.
```
ThisIsMani: Will this be a restriction on all providers?

crates/external_services/src/metrics_service.rs
```diff
@@ -0,0 +1,675 @@
...
+    }
+
+    /// Place the metric in a namespace.
+    ///
+    /// Required by providers that group metrics that way, ignored by those that do not.
+    #[must_use]
+    pub fn in_namespace(mut self, namespace: impl Into<String>) -> Self {
+        self.namespace = Some(namespace.into());
+        self
+    }
```
ThisIsMani: Let's not create constructors using mut. Let the caller construct these as required manually. Remove this builder pattern.

crates/external_services/src/metrics_service.rs
ThisIsMani, line 1: Make the comments concise.

crates/external_services/src/metrics_service.rs
```diff
@@ -0,0 +1,675 @@
...
+    /// which is how a state *transition* is detected without remembering anything between ticks.
+    pub fn ending_at(end: time::OffsetDateTime, period: Period, count: u32) -> MetricsResult<Self> {
+        let span = period
+            .seconds_i64()
+            .checked_mul(i64::from(count))
+            .ok_or_else(|| report!(MetricsError::InvalidRequest))
+            .attach_printable("The requested window is too wide to represent")?;
+
+        Self::new(end - time::Duration::seconds(span), end)
+    }
```
ThisIsMani: What is happening here?

crates/external_services/src/metrics_service.rs
```diff
@@ -0,0 +1,675 @@
...
+
+    /// The series laid out on a period grid, oldest first, `None` where there is no datapoint.
+    ///
+    /// This lives here rather than in each caller because laying sparse datapoints back onto their
+    /// periods is precisely the step where an `unwrap_or(0.0)` gets typed by accident, and a
+    /// missing datapoint is emphatically not a zero.
+    ///
+    /// Buckets cover `[end - count * period, end)` and a datapoint falls in the bucket containing
+    /// its timestamp, so exact grid alignment on the provider's side is not assumed. If two
+    /// datapoints share a bucket — which means `period` is wider than the one the query asked for
```
ThisIsMani: Isn't this cloudwatch specific? Will this be case of every provider?

crates/external_services/src/metrics_service.rs
ThisIsMani, line 1: This file has become too large, try to organize this into separate files.

crates/external_services/src/metrics_service.rs
ThisIsMani: Yes, and it should not have been. `1, 5, 10, 20, 30, or a multiple of 60` is purely CloudWatch's rule — GCP Monitoring accepts any duration from 60s upward, and Prometheus has no equivalent constraint at all. Baking it into `Period` would have made the neutral type reject a period a future provider happily accepts.

Moved: `Period::from_seconds` is now infallible and just carries the number. `aws_cloudwatch::validate` rejects what CloudWatch cannot take, along with empty batches, batches over 500, and backwards ranges — all before the round trip, since a rejected batch takes every metric in it down with it.

crates/external_services/src/metrics_service.rs
ThisIsMani: Done. `in_namespace`/`with_labels` are gone; `MetricQuery`, `MetricRequest` and `TimeRange` are plain structs with public fields, constructed directly.

`TimeRange::ending_at` survives as a plain constructor (not a `mut self` builder) — it computes `end - count * period`, which is the shape alarm evaluation always asks for. Say if you'd rather that went too.

crates/external_services/src/metrics_service.rs
ThisIsMani: Trimmed throughout. The module header went from ~44 lines to 15, and the per-item docs now say what a thing is and why it is not the obvious alternative, without the essay.

crates/external_services/src/metrics_service.rs
ThisIsMani: Overflow guard on `period × count` before subtracting it from `end` — and fair question, because the ceremony was doing more harm than the bug it prevented. It needed `count` in the order of 10^17 to trigger.

Now `saturating_mul`, and `ending_at` returns `Self` rather than a `Result`, so the whole error path disappeared.

crates/external_services/src/metrics_service.rs
ThisIsMani: Partly — and the part that was CloudWatch-specific was the justification, not the code.

Sparse-series-with-gaps is universal: GCP Monitoring and Prometheus both omit empty intervals rather than sending zeroes, so every provider needs the same "lay these back on a grid without inventing zeroes" step, and it should not be retyped per caller. What was CloudWatch-specific was me explaining it via `treat_missing_data`. That reference is gone.

Also fixed a real bug while in there: a non-positive period divided by zero. It now returns all-gaps, with a test.

crates/external_services/src/metrics_service.rs
ThisIsMani: Split. `metrics_service.rs` is now 47 lines — module docs, the trait, re-exports — with `query.rs` (184), `response.rs` (255) and `error.rs` (32) beside `aws_cloudwatch.rs`.

crates/external_services/src/metrics_service/aws_cloudwatch.rs
```diff
@@ -0,0 +1,433 @@
...
+    let dimensions = query
+        .labels()
+        .iter()
+        .map(|(name, value)| Dimension::builder().name(name).value(value).build())
+        .collect::<Vec<_>>();
+
+    let metric = Metric::builder()
+        .set_namespace(query.namespace().map(ToOwned::to_owned))
+        .metric_name(query.name())
+        .set_dimensions((!dimensions.is_empty()).then_some(dimensions))
```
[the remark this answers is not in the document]
ThisIsMani: Correct, and this was a real bug in the documentation rather than a nitpick — thank you.

`Labels::none()` claimed the metric would be "read across every stream reporting it". That is false: a dimension combination *identifies* a CloudWatch metric, so an unlabelled `MetricStat` query selects the stream published with no dimensions at all, and a metric published per DB instance or target group comes back as an empty `Complete` series. That is exactly the silent-empty-result failure this design exists to prevent, arriving through a doc comment.

Taken the "restrict the contract" option: the neutral type now says an empty set is not a request to aggregate and that what it selects is provider business, and `aws_cloudwatch` states the CloudWatch meaning outright. No query expansion — callers name the stream they mean, and Terraform publishes the dimension values for that purpose (hyperswitch-cloud#23444).

crates/external_services/src/metrics_service/aws_cloudwatch.rs
```diff
@@ -0,0 +1,538 @@
...
+}
+
+/// Reject a request CloudWatch would refuse, before spending a round trip on it.
+///
+/// A rejected batch takes every metric in it down, so one bad period would blind a whole tick.
+fn validate(request: &MetricRequest) -> MetricsResult<()> {
+    if request.queries.is_empty() {
+        return Err(report!(MetricsError::InvalidRequest))
+            .attach_printable("A GetMetricData call needs at least one query");
+    }
```
ThisIsMani: Don't use early return, in this case, you can use ? instead.

ThisIsMani: Done, and applied across the module rather than just here.

`validate` was four guard clauses in a row — which is really four validations sharing a body. Split into `validate_batch_size`, `validate_range` and `validate_period`, composed with `?` and `try_for_each`, using `common_utils::fp_utils::when` since `aws_kms` and `aws_s3` already reach for it.

Two more went the same way. The placement loop had a `continue` that was hiding a question worth naming — *which slot does this timestamp fall in?* — now `slot_of` returning `Option<usize>`, so "outside the window" is an absent slot rather than a jump. `slot_count` and `validate_pairing` lost their guards too.

No `return`, `continue` or `break` left in the module.

crates/external_services/src/metrics_service/aws_cloudwatch.rs
```diff
@@ -0,0 +1,655 @@
...
+        .into_iter()
+        .filter_map(|(timestamp, value)| {
+            slot_of(timestamp, request.range.start, query.period).map(|slot| (slot, value))
+        })
+        .for_each(|(slot, value)| {
+            // A later datapoint sharing a slot wins, since they arrive oldest-first.
+            if let Some(bucket) = slots.get_mut(slot) {
+                *bucket = Some(value);
+            }
+        });
```
ThisIsMani: Can we use fold instead?

ThisIsMani: Yes — and it removed more than the `mut`.

The previous version had to `collect::<MetricsResult<Vec<_>>>()?` first, because `from_aws_timestamp` is fallible and `for_each` has nowhere to put an error. `try_fold` carries both the accumulator and the failure, so that intermediate `Vec` is gone too: one pass, one allocation.

```rust
let slots = timestamps.iter().zip(values).try_fold(
    vec![None; slot_count(request.range, query.period)],
    |mut slots, (timestamp, value)| -> MetricsResult<_> {
        let slot = slot_of(from_aws_timestamp(timestamp)?, request.range.start, query.period);

        // A later datapoint sharing a slot wins, since they arrive oldest-first.
        if let Some(bucket) = slot.and_then(|slot| slots.get_mut(slot)) {
            *bucket = Some(*value);
        }

        Ok(slots)
    },
)?;
```

The closure needs the `-> MetricsResult<_>` annotation, since nothing else constrains the error type.

Took three more while I was in there: `query_for` collapses the two-step id lookup into one (an unparseable id and an out-of-range one both mean "we did not ask for this"), `validate_batch_size` is one range check rather than two `when`s, and `slot_of`/`slot_count` use `bool::then` instead of `Some(x).filter(..).map(..)`.

PR #14154, opened 2026-09-10
fix(core): external 3ds retry failover and ucs failed attempts

pixincreate:
## Type of Change

- [x] Bugfix
- [x] Enhancement

## Description

- Skip the connector that just ran when expanding the external 3DS retry list. An unpinned plan entry (`merchant_connector_id: None`) did not match the resolved connector (`Some(id)`), so it was appended again and retried on the same connector instead of failing over.
- Record a failed attempt when UCS rejects a request during transformation on the primary UCS path. The error now produces a failed attempt instead of leaving the payment intent in `processing`, matching the connector-error branch.
- Add `rapyd` to `ucs_only_connectors` in all shipped configs.

### Additional Changes

`config/config.example.toml`, `config/development.toml`, `config/deployments/env_specific.toml`, `config/deployments/integration_test.toml`, `config/deployments/production.toml`, `config/deployments/sandbox.toml`

## Motivation and Context

External 3DS failover can authorize twice on the same connector, and UCS request failures leave the payment intent stuck in `processing`.

## How did you test it?

`cargo check -p router`, `cargo clippy -p router --no-deps`, and `cargo +nightly fmt --all -- --check` pass.

## Checklist


Shubhodip900 (dismissed):

sahkal (dismissed):

crates/router/src/core/payments/routing.rs
```diff
@@ -1472,9 +1472,13 @@ where
 
     let mut ordered = vec![predetermined.clone()];
     for choice in candidates {
+        // An unpinned plan entry (`None`) refers to the connector that just ran; skip it so
+        // it is not appended again and retried on the same connector.
         let is_current_connector = choice.connector.to_string()
             == predetermined.connector_data.connector_name.to_string()
-            && choice.merchant_connector_id == predetermined.connector_data.merchant_connector_id;
+            && (choice.merchant_connector_id.is_none()
```
jagan-jaya, line 1479: Can you share an example scenario where the choice connector has a name but not the mca_id? 
How is this possible in our system currently?

pixincreate: sure, a plan entry can name only the connector. that is a supported form.

example: a priority plan ["adyen", "cybersource"] with no accounts pinned.

- the plan is stored as-is, so both entries have `merchant_connector_id: null`.
- When adyen runs, its account id is resolved separately into `payment_attempt.merchant_connector_id` -> `Some(mca_...)`.

so at retry time the connector that ran has `Some(id)`, while the plan entry has `None`. the old check compared them directly, never matched, and re-added adyen to its own retry list.

name-only entries are normal:

- `RoutableConnectorChoice` accepts a bare connector name (only-connector form) -> id stays null.
- stored routing rules can list connector names only: in [crates/router/src/core/routing/helpers.rs](https://github.com/juspay/hyperswitch/blob/9e7d13e78b9d09e4f2651840c422ef2c13c94335/crates/router/src/core/routing/helpers.rs) (v2, v1), a choice with `Some(mca)` must match a connector + MCA pair, while `None` only checks the connector exists for the profile.

  v2:
  https://github.com/juspay/hyperswitch/blob/9e7d13e78b9d09e4f2651840c422ef2c13c94335/crates/router/src/core/routing/helpers.rs#L378-L404
  v1:
  https://github.com/juspay/hyperswitch/blob/9e7d13e78b9d09e4f2651840c422ef2c13c94335/crates/router/src/core/routing/helpers.rs#L555-L579

- decision engine results can omit the gateway id.

the fix treats `None` as "this is the current connector". pinned entries for a different account of the same connector still stay as valid failover targets.

jagan-jaya (dismissed):

crates/router/src/core/unified_connector_service.rs
```diff
@@ -4067,11 +4067,40 @@ where
                     "error": error.to_string(),
                     "error_type": "ucs_call_failed"
                 });
-                (
-                    error.current_context().http_status(),
-                    Some(error_body),
-                    Err(error),
-                )
+                match error.current_context() {
```
gopikrishna000: but why sperical case logic for priamry mode vs shadeo mode then we will get false difff

crates/router/src/core/unified_connector_service.rs
```diff
@@ -4067,11 +4067,40 @@ where
...
+                    // Record deterministic integration errors as a failed attempt on the
+                    // primary UCS path; otherwise the intent stays in `processing`.
+                    UnifiedConnectorServiceError::NotImplemented(message)
+                        if execution_mode == ExecutionMode::Primary =>
```
gopikrishna000: and also why error mappign logic for this sepcific error alone in wropae\r
[5:45 PM]other errors it has seprate error mappign transform ryt

sahkal (dismissed):

Shubhodip900 (dismissed):

jagan-jaya (dismissed):

gopikrishna000 (approved):

jagan-jaya (approved):

Shubhodip900 (approved):

sahkal (approved):

PR #14167, opened 2026-09-10
refactor(routing): remove the DE-vs-HS shadow diff and its kill switch

### LINKED ISSUE #14175: [BUG] Decision Engine shadow routing logs false mismatches from comparison-stage artifacts
### Bug Description

Decision Engine shadow-mode routing reports a steady stream of DE-vs-Hyperswitch mismatches for payments where both engines actually agree. One merchant profile showed a 39 identical / 101 mismatching split, with no routing difference behind any of the 101.

Two independent comparison-stage causes.

**1. Hyperswitch diffs a pre-fallback list in two flows.**

When a rule matches nothing and `default_selection` is `{ priority: [] }`, Hyperswitch's DSL yields `[]`. The Decision Engine answers the same rule by substituting the `fallback_output` the request carried, returning the merchant's fallback connectors under `status: default_selection`. Hyperswitch falls back too — but in two of the four comparison sites it does so *after* the diff is taken:

| site | fallback applied before the diff? |
|---|---|
| `perform_static_routing_locally` → shadow `RoutingFlow::Payment` | yes |
| `SessionRoutingStage` (`session_token_routing`) | yes |
| `perform_session_routing_for_pm_type` (`payment_method_list_pre_routing`) | no |
| `perform_static_routing_v1` (payouts / MIT / v2) | no |

`perform_static_routing_v1` is the one with real consequences: unlike the shadow paths it feeds the diff kill switch, so a profile with an empty `default_selection` on a payout or MIT rule counts a `LengthMismatch` per payment toward the threshold-100 revert to Hyperswitch routing.

**2. The diff compares a pre-filter list against a post-filter one.**

The Decision Engine narrows `evaluated_output` by its `pm_filter` graph and leaves `output` as the rule produced it. `transform_de_output_for_router` builds the router's list as `evaluated_output` first, then whatever remains in `output`, so a connector the `pm_filter` dropped is demoted to the tail rather than removed. Hyperswitch's own side of the diff is the raw rule order — its kgraph filter runs after the comparison. Same set, different order.

### Expected Behavior

The shadow diff compares rule-engine output against rule-engine output, so a logged mismatch means the two engines genuinely disagree.

### Actual Behavior

Mismatches are logged for payments where both engines selected the same connectors. Two shapes:

- `[]` on the Hyperswitch side vs the full fallback list on the Decision Engine side, for every non-matching payment on a profile with an empty `default_selection`.
- Same eight connectors in a different order, e.g. `paypal` at index 2 on the Hyperswitch side and last on the Decision Engine side, with `is_equal_length=true`, `is_volume=false`, `is_de_result_empty=false`.

### Steps To Reproduce

1. Configure a profile with a routing rule that does not match the payment being sent (e.g. `payment_method == reward AND reward == classic_reward`) and `default_selection: { priority: [] }`.
2. Enable `static_routing_enabled` and `shadow_routing_enabled` for the profile without cutting it over.
3. Send a wallet payment (Apple Pay / Google Pay) and fetch the payment method list.
4. Read the `decision_engine_euclid` diff log: `payment_method_list_pre_routing` logs `hs_response=[]` against a `de_response` holding the merchant's fallback connectors, while `session_token_routing` for the same payment logs a match.

For the second cause, send a payment whose payment method type the Decision Engine's `pm_filter` graph rejects for one connector in the rule, and compare the `evaluate_routing` diff ordering.

### Context For The Bug

Shadow-mode diff output is the signal driving the Decision Engine cutover. A constant background of false mismatches means a genuine ordering regression inside the rule engine would be invisible, and the `perform_static_routing_v1` half can silently revert a profile through the kill switch.

Because the Decision Engine's `pm_filter` and Hyperswitch's kgraph are not equivalent, the ordering divergence will not settle on its own.

### Environment

Self-hosted, `main`.

---

### Resolution (2026-09-16)

The report above stands as written — both causes were confirmed against the code of both engines, and the mismatches were comparison-stage artifacts, not routing disagreements.

The resolution changed, however. #14167 was first implemented as an *alignment* fix (compare post-fallback lists, and rule-decision against rule-decision via a dedicated diff shape). With the investigation establishing that the engines agree on every payment, the shadow diff was judged to have served its purpose, and the PR was rewritten to **remove the shadow diff and the kill switch built on it entirely**, rather than patch them:

- All `compare_and_log_result` sites removed (payment, session-token, payment-method-list, and the `list_routing` / `list_active_routing` management diffs).
- All shadow evaluation machinery removed (`shadow_decision_engine_routing` single + batch, `ShadowBatchEntry`, the spawned off-path evaluations, `open_router.shadow_routing_enabled`).
- The kill switch removed end to end: Redis diff counter, threshold overlay in `get_routing_result_source`, `[open_router.diff_kill_switch]` config, both metrics, and the `DELETE /routing/decision-engine/{profile_id}/diff-counter` reset endpoint.
- `perform_static_routing_v1` no longer makes a synchronous Decision Engine call for non-cutover profiles — that call only ever existed to feed the diff, so cause 1's kill-switch consequence (a profile permanently barred from cutover by comparison artifacts on a lifetime counter) is gone by construction.

Cutover behavior is unchanged: cut-over profiles are still served by the Decision Engine, with the fallback to the Hyperswitch result on an empty engine response.

Resolved by #14167.

prajjwalkumar17:
## Type of Change

- [x] Refactoring

## Description

> Scope change: this PR originally aligned the shadow-diff comparison stages (see the earlier revision and #14175). The shadow-mode investigation it came out of concluded that every reported mismatch was a comparison-stage artifact — the two engines agreed on every payment. With that established, the diff has served its purpose, so this PR now removes the DE-vs-HS diff and the kill switch built on it, instead of patching them (net **−798 lines**). **Cutover is untouched.**
>
> The rewrite force-replaced the branch head, superseding `b5ea647a7f` ("skip the hybrid shadow diff when there is no static half") — every line of that commit modified diff machinery this PR deletes, so nothing of it survives semantically. Details in [this comment](https://github.com/juspay/hyperswitch/pull/14167#issuecomment-5695749544).

Removed:

- **Shadow diff logging** — every `compare_and_log_result` site: `evaluate_routing`, `session_token_routing`, `payment_method_list_pre_routing`, and the management-read diffs (`list_routing`, `list_active_routing`).
- **Shadow evaluation** — `shadow_decision_engine_routing` (single + batch), `ShadowBatchEntry`, the spawned off-request-path evaluations in the payment, session-token and payment-method-list flows, and the `open_router.shadow_routing_enabled` flag. These existed only to produce diffs.
- **The diff kill switch** — the Redis diff counter, `record_de_diff_and_maybe_trip_kill_switch`, the threshold overlay inside `get_routing_result_source` (now a plain source read), the `[open_router.diff_kill_switch]` config, both metrics (`DECISION_ENGINE_ROUTING_DIFF`, `DECISION_ENGINE_KILL_SWITCH_TRIGGERED`), and the reset API `DELETE /routing/decision-engine/{profile_id}/diff-counter` with its `Flow` variant.
- **`is_volume_split` plumbing** — it existed solely to annotate diffs as expected-noise (volume splits are randomized draws).

Also as a consequence: `perform_static_routing_v1` (payouts / MIT / v2 payments) previously made a **synchronous Decision Engine call on every payment** whenever the global `static_routing_enabled` flag was on, even for profiles not cut over — purely to feed the diff. The DE call is now gated on effective cutover, where its result is actually load-bearing. Non-cutover profiles no longer pay that round trip.

Preserved unchanged:

- `is_decision_engine_routing_effective` / `get_routing_result_source` (cutover resolution)
- `select_routing_result` — cut-over profiles are served the DE result, with the warn-logged fallback to the Hyperswitch result when the DE result is empty
- `decision_engine_routing` single/batch/batch-with-fallback and DE-only rule writes for cut-over profiles
- The hybrid routing stage for cut-over payments
- `RoutingApproach` derivation and persistence (analytics-visible values do not move)

### Additional Changes

API: removes the internal admin endpoint `DELETE /routing/decision-engine/{profile_id}/diff-counter` (kill-switch counter reset; meaningless without the counter). It is not in the OpenAPI spec.
Config: removes `open_router.shadow_routing_enabled` and the `[open_router.diff_kill_switch]` section from `config.example.toml` / `development.toml` and the settings structs. Deployments still setting these keys are unaffected — unknown config keys are ignored at deserialization.
Ops note: existing `routing_decision_engine_{profile}_diff_count` Redis keys become inert (nothing reads or writes them); no TTL was set, so they can be deleted manually if desired.

## Motivation and Context

Closes #14175.

The shadow-diff investigation (#14175) proved the reported mismatches were artifacts of comparing lists taken at different pipeline stages, not routing disagreements. Beyond being noise, the machinery was actively hazardous:

- The kill-switch counter was **lifetime with no TTL** and was fed from `perform_static_routing_v1` before cutover — so comparison artifacts (e.g. an empty `default_selection` on a payout rule) could permanently bar a profile from cutting over until a manual reset.
- Under cutover, rule writes are DE-only, so the Hyperswitch baseline is stale by design and a diff carries no signal there either — counting was already disabled post-cutover, meaning the switch could never provide the post-cutover protection its name implies.
- Every non-cutover payment through `perform_static_routing_v1` paid a synchronous DE round trip only to produce the diff log.

## How did you test it?

- `RUSTFLAGS="-D warnings" just clippy` and `just clippy_v2` (the CI lint gates) pass on the exact pushed commit; `cargo +nightly fmt --all -- --check` is clean.
- `cargo test -p router` routing suites: **11/11 pass** (`transform_de_output_tests` + `de_program_round_trip_tests` — both untouched; they cover the preserved cutover transformation).
- Repo-wide sweep for every removed identifier (code, configs, TOMLs, docs, cypress, api-reference, .github) returns zero hits.
- A three-lens adversarial review (cutover behavior / leftovers / operational surface) ran over the final diff; its one confirmed finding — `RoutingFlow::as_str` left as dead pub API with stale docs, invisible to clippy because the whole chain is `pub` — is fixed in this commit, along with the similarly-orphaned `RoutingApproach::is_de_static_result`.

## Checklist

- [ ] I added unit tests for my changes where possible (removal-only; existing suites cover the preserved paths)


AnkitKmrGupta (dismissed):

prajjwalkumar17:
**Scope pivot, and a force-push notice.**

Per the shadow-mode investigation's conclusion (every reported DE-vs-HS mismatch was a comparison-stage artifact — the engines agreed on every payment), this PR was rewritten from *aligning* the shadow comparisons to *removing* the shadow diff and the kill switch entirely. Cutover behavior is untouched — see the updated description.

@AnkitKmrGupta this force-push replaced the previous head, including your `b5ea647a7f` ("skip the hybrid shadow diff when there is no static half"). That commit is fully subsumed by the rewrite: the hybrid-stage `compare_and_log_result` call, the `diff_connectors` field, and the diff-shape wiring in `normalize_hybrid_routing_response` it modified are all deleted here, so there is no longer a hybrid diff to skip. Your kill-switch question is also answered by this PR the blunt way — `record_de_diff_and_maybe_trip_kill_switch`, the counter, and the `get_routing_result_source` overlay are gone; `!de_routing_effective` no longer exists anywhere.

jagan-jaya (approved):

AnkitKmrGupta (approved):

jarnura (approved):

PR #14168, opened 2026-09-10
feat(observability): evaluate CloudWatch metric alarms and announce transitions

ThisIsMani:
## Type of Change

- [x] New feature

## Description

Stacked on #14150, which is this branch's base — review that one first; this PR's diff is the evaluator alone.

The observability plane now decides as well as delivers. `POST /alerts/cloudwatch/evaluate`, behind the existing `X-Internal-Api-Key` guard, reads the twenty `rds-alerts` definitions from configuration, evaluates them against CloudWatch through the metrics client added in #14150, and hands each state transition to the in-process notifier. Closes [#23445](https://github.com/juspay/hyperswitch-cloud/issues/23445).

### Reproducing CloudWatch

Three things decide whether a port changes when an alarm fires, and all three are easy to get wrong:

**Evaluations are a minute apart, not a period apart.** AWS evaluates every metric alarm every minute for any period of a minute or longer, and the sliding window advances by a minute — *"at the end of minute 5 the alarm evaluates on minutes 1–5. Then at the end of minute 6, on minutes 2–6."* So two consecutive evaluations of a 300-second metric are two overlapping 300-second windows, not two adjacent datapoints. The twelve 60-second definitions get both windows out of a single `range + 1` request, one slot apart; the eight 300-second ones need two differently aligned requests, because a returned datapoint lands on the grid the request's own start time defines. A run is three `GetMetricData` calls.

**The evaluation range is wider than the window.** A gap in a recent period is covered by a real reading from further back before the missing-data policy is consulted at all. AWS documents that the size of this widening *"depends on the length of the alarm period and whether it is based on a metric with standard resolution or high resolution"* and publishes no formula, so `missing_data_lookback` is configuration defaulting to 2 — the only figure its worked examples put to a standard-resolution metric — rather than a constant presented as parity.

**A failed query is not a metric that stopped reporting.** A series that came back `Failed` or `Partial`, or did not come back, is not evaluated. It never becomes a window of gaps for a `breaching` policy to turn into an alarm about a database that is perfectly healthy. Everything else in the run evaluates normally, and the affected definitions get one operational summary.

### Two divergences, named rather than approximated

`treat_missing_data = "ignore"` keeps the alarm's previous state — a memory that stateless window reconstruction does not have. A catalogue asking for it is **refused at boot** with that reason rather than quietly evaluated as something else.

AWS's *premature alarm state* rule is not implemented. Its two published tables contain rows that satisfy the stated condition and disagree on the outcome, and it only ever changes the result under `missing` or `ignore`. All twenty cells of both tables are encoded as tests, including the two cells this diverges on, so the gap is visible rather than assumed away.

Neither policy appears in `rds-alerts`. Worth flagging against the ticket text: the Terraform module defaults `treat_missing_data` to `notBreaching`, **not** AWS's `missing`, so across the 20 definitions it is `breaching` on 2 severities and `notBreaching` on the other 51 — which means `INSUFFICIENT_DATA` is unreachable for this catalogue, and the ticket's "a transition to INSUFFICIENT_DATA on a `breaching` alarm is worth announcing" cannot occur (`breaching` + all-missing is ALARM). The state still exists for `missing`. Likewise `datapoints_to_alarm` is used 7 times in the full 76-definition catalogue and **zero** times in `rds-alerts`; M-out-of-N is implemented and tested, but nothing shipped exercises it, and a test asserts that so the day a rendered catalogue starts using it, something notices.

### Nothing is remembered

Both windows are reconstructed from CloudWatch on every request. No `HashMap`, no Redis, no table — and no mute, acknowledgement, history or delivery state, because that is alert *management*, it is source-agnostic, and building it here would produce a second mechanism covering infrastructure alerts and nothing else. The accepted costs, stated plainly: repeated requests announce the same transition twice, and a transition hidden by a failed query or a late-arriving datapoint is not announced at all. Delivery retries are [#23485](https://github.com/juspay/hyperswitch-cloud/issues/23485).

### The catalogue belongs to Terraform

The infrastructure repository knows the resource identifiers and already expands the templated classifications, so it renders the catalogue and Helm delivers it; definitions arrive with **dimensions already resolved** and this service validates and evaluates them. No Terraform parsing, no resource discovery, no group expansion in Rust. `config/observability.toml` carries a hand-checked snapshot until that rendering lands ([#23444](https://github.com/juspay/hyperswitch-cloud/issues/23444)) and says so — `[cloudwatch.source]` records origin and revision and sets `temporary`, which makes the service log a warning at **every** boot. A comment would have done the same job right up until someone deleted it.

One thing worth a look in review: **dimensions are name/value pairs, not a map.** The `config` crate lowercases every key it reads, so `DBInstanceIdentifier` written as a map key arrives as `dbinstanceidentifier` — a case CloudWatch does not know. It would name a metric that reports nothing, so under this catalogue's `notBreaching` default all twenty alarms would evaluate cleanly and never fire. This was a real bug in an earlier revision of this branch, caught by the catalogue test; carrying the name as a *value* is what preserves its case.

## Additional Changes

- [x] Configuration file changes

`MetricSeries::new`, `MetricPage::new` and `Cursor::new` in `external_services` become public. They were crate-private, which meant `MetricsProvider` — object-safe and provider-neutral on purpose — could not be implemented outside `external_services`: a trait no other crate can produce a value for is one no other crate can implement, so "provider-neutral" quietly meant "the two providers in this module". It is also what lets this crate stand a stub provider up instead of reaching for a live account.

`crates/observability/src/lib.rs` documented the crate as delivery-only — *"Deciding what is alert-worthy happens elsewhere"* — which is no longer true, so it now says what the crate actually does and where the line between generation and management falls.

## Motivation and Context

The keystone of [the CloudWatch-sourced infrastructure alerts map](https://github.com/juspay/hyperswitch-cloud/issues/23440). `rds-alerts` ships first because its dimensions are stable names rather than generated ids, so it does not wait on the Terraform work, and because at 20 definitions it is the largest classification — the config shape gets stressed rather than validated against a toy.

## How did you test it?

`cargo test -p observability --features v1` — **116 passing**, plus `clippy` and `fmt` clean.

Worth knowing what the tests actually pin down:

- **The run itself** (`tests/alarm_evaluation.rs`), against a stub provider. This is the one that matters: it covers the ticket's acceptance criterion — a condition that stays breaching across two consecutive evaluations announces **exactly once** — plus recovery, dry run, partial failure, page merging, and delivery refusals. The stub is a minute-resolution metric history that aggregates on demand rather than a canned response, which is what makes the five-minute test meaningful: it fails if the two windows a minute apart are ever collapsed into one. Both key properties were checked by mutation, not assumed — making `Transition::between` always fire failed exactly the announce-once test, and treating a `Failed` series as data failed exactly the two tests about query failure not being missing data.
- **Both AWS missing-data tables**, all twenty cells, as the parity contract.
- **The shipped catalogue itself** (`tests/alarm_catalogue.rs`) loaded exactly as the service loads it, through `Settings::with_config_path` and full `validate()`. This is the guard that matters most, because every way a hand-checked snapshot goes wrong is silent: a threshold that lost a digit still evaluates, a definition that lost its dimensions queries a metric that reports nothing, and a severity pointing at an unconfigured destination evaluates perfectly and delivers nowhere. It asserts 20 definitions / 53 severities / 20 queries, the 12–8 period split, the resolved dimension values against the infrastructure they were copied from, and the missing-data policies in use.
- **Window planning**: that a 60-second metric needs one request and a 300-second metric needs two, one minute apart.

Exercised end to end against a running server, with CloudWatch deliberately failing (see below):

```
evaluated_at   2026-09-10T18:13:00.000Z
compared_with  2026-09-10T18:12:00.000Z     ← one minute, not one period
definitions 20 · readings 20 · severities 53
```

- all 53 severities came back `state: null` with an `error` — not one fabricated breach
- one total-failure summary naming the affected definitions, truncated at 12 with `…and 8 more`
- `"delivery":{"status":"skipped_dry_run"}` — the dry run skipped the failure notification too, not just the announcements
- `401` unauthenticated; `400` on `{"dryrun": true}`, the misspelling that would otherwise have silently sent real messages

### Not yet verified — flagging honestly

- **No live CloudWatch data.** The IAM identity available to me is denied `cloudwatch:GetMetricData` on the sandbox account, which is what produced the failure path above. The ticket's "all 20 definitions evaluate against real sandbox CloudWatch data" needs that grant, and so does empirically confirming the 300-second window alignment — that is documented from the AWS API reference and the client's own grid placement, not measured.
- **No synthetic breach delivered to a real channel.** Everything is pointed at the `log` destination; a sandbox destination id is needed for that check. Per the handoff, that must be the sandbox destination and never a production channel.

I would keep the build ticket open until both are done.

https://claude.ai/code/session_011FoL6nSZvJnajbpScxLrEP

### DIFF
--- a/crates/observability/src/core/alarm.rs
+++ b/crates/observability/src/core/alarm.rs
@@ -1,0 +1,668 @@
+//! One evaluation run: read the metrics, decide, announce, report what happened.
+//!
+//! This is the only module in the crate that has a clock, a metrics provider and a notifier at
+//! once. Everything it decides is decided by [`crate::domain::alarm`]; what lives here is the
+//! order of operations and the answer to "what should happen when part of this fails".
+//!
+//! ## The clock
+//!
+//! A run evaluates through the **latest completed minute** and compares against the evaluation
+//! ending one minute earlier — not one period earlier. CloudWatch evaluates every minute for any
+//! period of a minute or longer, so two consecutive evaluations of a five-minute metric are two
+//! overlapping five-minute windows, and the trigger that eventually drives this should run every
+//! minute regardless of the periods in the catalogue.
+//!
+//! ## Nothing is remembered
+//!
+//! Both windows are reconstructed from CloudWatch on every request, and no state is written
+//! anywhere. Two consequences are accepted rather than worked around:
+//!
+//! * Two requests in the same minute announce the same transition twice. There is no
+//!   exactly-once delivery guarantee here, and adding one would mean the alert *management* store
+//!   this ticket is explicitly not building.
+//! * A transition that happens while a query is failing, or that is only visible in a datapoint
+//!   that arrives late, is not announced at all — the next run compares two windows that both
+//!   already contain it. See the delivery-hardening ticket.
+//!
+//! ## Failure is not missing data
+//!
+//! The distinction this module exists to protect: a metric that stopped reporting and a CloudWatch
+//! call that did not work look identical if you flatten them, and one of them is an alarm while
+//! the other is an outage. A reading whose series came back `Failed` or `Partial`, or that did not
+//! come back at all, is **not evaluated** — it does not become a window of gaps for the
+//! missing-data policy to turn into a breach. It is reported, and every other reading in the run
+//! is evaluated normally.
+
+use std::{collections::BTreeMap, sync::Arc};
+
+use common_utils::date_time;
+use error_stack::report;
+use external_services::metrics_service::{Cursor, MetricRequest, MetricsProvider, SeriesStatus};
+use hyperswitch_masking::Secret;
+
+use crate::{
+    domain::{
+        alarm::{
+            self,
+            catalogue::{AlarmTarget, Catalogue, FetchPlan, Reading},
+            message, AlarmState, Evaluation, Transition,
+        },
+        notifier::{chat::ChatNotification, Outcome},
+    },
+    errors::{ObservabilityApiResult, ObservabilityError},
+    logger,
+    state::AppState,
+};
+
+/// How many pages one request is followed for before the run gives up on it.
+///
+/// A page holds up to 100,800 datapoints and a run asks for tens, so the first page is always the
+/// last. The cap is here because a provider that kept handing back a cursor would otherwise be an
+/// unbounded loop inside a request handler.
+const MAX_PAGES: usize = 8;
+
+/// What a caller asked of one run.
+#[derive(Debug, Clone, Copy, Default)]
+pub struct EvaluationOptions {
+    /// Evaluate and render, but send nothing at all — not even a failure summary.
+    ///
+    /// A dry run is the same evaluation, so its proposed messages are the messages a real run
+    /// would have sent rather than a separately built preview of them.
+    pub dry_run: bool,
+}
+
+/// Evaluate every configured definition and announce what changed.
+///
+/// Never returns an error for a metric that could not be read or a message that could not be sent:
+/// both are reported in the run's own result, because a run that read nineteen of twenty metrics
+/// did nineteen twentieths of its job and saying so is more useful than a 502.
+pub async fn evaluate(
+    state: AppState,
+    options: EvaluationOptions,
+) -> ObservabilityApiResult<RunReport> {
+    let catalogue = Arc::clone(&state.alarms);
+    let evaluated_at = evaluation_instant(date_time::now(), catalogue.evaluation_delay_seconds);
+
+    let mut report = RunReport {
+        evaluated_at,
+        previous_evaluated_at: evaluated_at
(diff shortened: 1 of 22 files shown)


crates/observability/src/core/alarm.rs
```diff
@@ -0,0 +1,668 @@
...
+    let mut report = RunReport {
+        evaluated_at,
+        previous_evaluated_at: evaluated_at
+            - time::Duration::seconds(alarm::catalogue::EVALUATION_INTERVAL_SECONDS),
+        dry_run: options.dry_run,
+        definitions: catalogue.definitions(),
+        readings: catalogue.readings.len(),
+        targets: Vec::with_capacity(catalogue.targets.len()),
+        failure: None,
+    };
```
ThisIsMani: Let's not use muts.

PR #14171, opened 2026-09-11
feat(observability): read the infrastructure alarm catalogue from configuration

ThisIsMani:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

Adds the `cloudwatch` configuration section: the infrastructure alarm catalogue this service will
evaluate, read from configuration like everything else the deployment sets, validated at boot, and
logged once so a deployment states how much of the catalogue actually reached it.

```
{"level":"INFO","message":"Observability started on 127.0.0.1:8085"}
{"level":"INFO","message":"Loaded 20 cloudwatch alarm definitions carrying 53 severity rules"}
```

No evaluation yet — that is the next change. This is the input it will read, landing on its own so
the shape can be reviewed without an algorithm sitting on top of it.

### The environment dictates the shape, and two parts of it are load-bearing

The deployment renders `.Values.configs` into `OBSERVABILITY__`-prefixed environment variables, so
the catalogue has to survive that trip. Two consequences are worth the review time:

**Dimensions are a single `Name=value,Name=value` string, not a nested table.** `config` lowercases
every key it reads from the environment, and CloudWatch dimension names are case-sensitive. A
nested table would arrive as `dbinstanceidentifier`, which CloudWatch answers with *no data* rather
than with an error — the alarm would never fire and nothing would say why. As a value the casing
survives untouched. There is a test pinning exactly this, because it is the kind of thing a later
refactor "tidies up" into a map. A table is still accepted, since a file can express one.

**Severities nest under their definition.** Metric, namespace and dimensions are properties of the
thing being watched rather than of how alarming it is, so they are set once and two severities of
one definition cannot end up watching different metrics. Ids are lowercase and `_`-separated
because an environment variable name cannot contain `-`; the name each definition has upstream
travels as a `name` value rather than being reconstructed from the id by guesswork.

### No defaults for the rule fields

The upstream catalogue leaves `period`, `statistic`, `comparison_operator`, `evaluation_periods`
and `treat_missing_data` unset on most entries, and the Terraform module fills them in *before the
AWS alarms are created*. Repeating those defaults here would be a second copy of somebody else's
decision, free to drift from the module that creates the alarms — and the drift would look like a
working alarm evaluating a different rule. So whatever renders the configuration resolves them, and
a rule that does not say what it evaluates is refused at boot. `datapoints_to_alarm` is the one
exception: absent genuinely means "all of them" in CloudWatch, so it is an `Option`.

### Validation rejects what would otherwise be a wrong answer

Not a taste checklist — each of these boots into a service that silently evaluates something other
than what the catalogue says: an id that cannot be set from the environment; a definition with no
severities (watches a metric, can never say anything); a period of zero or beyond CloudWatch's
maximum; M-out-of-N with M greater than N, a window that can never fill; a NaN threshold, which
compares false against every reading and so is indistinguishable from an alarm that is simply never
breached; and an empty description, which is the message that gets delivered, not documentation.

### Additional Changes

- `config/observability.toml` — the new section, documented with a worked example. Empty by
  default: a deployment that has not been given a catalogue is valid, not broken.
No catalogue ships in this repo. The definitions are deployment configuration and are written
directly in hyperswitch-infra as chart values; the sandbox set is recorded on
[#23444](https://github.com/juspay/hyperswitch-cloud/issues/23444#issuecomment-5631724945).

## Motivation and Context

[Evaluate CloudWatch metric alarms in the observability plane](https://github.com/juspay/hyperswitch-cloud/issues/23445)
cannot evaluate anything until the definitions are in the process, and
[Publish the IaC CloudWatch alarm catalogue through Helm](https://github.com/juspay/hyperswitch-cloud/issues/23444)
owns generating them from Terraform. Rather than wait on a second repo, an immutable module tag and
a deployment, this agrees the consumer contract now and populates it by hand — the ownership
decision is unchanged, only the order of work.

Supersedes #14170, which proposed shipping the definitions as a JSON file plus generator scripts.
How they are produced is still open, but they do not belong in this repo either way.

## How did you test it?

Seven new unit tests, hermetic: `config::Environment::source` takes a supplied map rather than the
process environment, so an environment test says nothing about whichever test runs beside it and
still parses through exactly the production path — same prefix, same `__` separator, same parsing.

- A definition set entirely from the environment arrives as typed rules, with `90` a number,
  `breaching` a policy, and an absent `datapoints_to_alarm` a `None` rather than a zero.
- Dimension names keep their case through the environment — the test that fails if the string form
  is ever replaced by a table.
- Dimensions parse spaced and multi-valued, round-trip through `Display`, and reject a non-pair, an
  empty name, an empty value and a duplicate.
- Nine invalid catalogues are each rejected, named by what they would have done.
- An empty catalogue is accepted.

End to end, through the actual delivery path — `helm template` over the observability-plane chart
with the sandbox catalogue as `configs.cloudwatch.alarms` values, then the rendered ConfigMap
(405 `OBSERVABILITY__CLOUDWATCH__*` entries, 48 KB) exported into the service:

```
extracted 405 vars from the rendered ConfigMap
{"level":"INFO",...,"message":"Observability started on 127.0.0.1:8085"}
{"level":"INFO",...,"message":"Loaded 20 cloudwatch alarm definitions carrying 53 severity rules"}
```

Validation runs before anything binds, so those 405 variables are also 53 rules that passed every
check above.

**Not tested against CloudWatch.** Nothing here queries anything — the definitions are parsed and
validated, not used. Whether AWS returns what these select is the evaluator's sandbox run.

## Checklist

`cargo clippy -p observability --all-targets` is clean, and `cargo test -p observability --lib`
passes 42 tests.

https://claude.ai/code/session_01Ej2mkfp9QVkgox9AMubMDi


crates/observability/src/settings/cloudwatch.rs
ThisIsMani, line 1: Remove comments in this file. This has more comments than the code itself.

config/cloudwatch_alarms.sandbox.env
ThisIsMani, line 1: Can't this be a part of toml instead?

crates/observability/src/settings/cloudwatch.rs
```diff
@@ -0,0 +1,688 @@
...
+            ("namespace", &self.namespace),
+        ] {
+            if value.is_default_or_empty() {
+                return reject(format!("{field} must not be empty"));
+            }
+        }
+
+        // A period of zero divides the window into nothing; one above CloudWatch's maximum is a
+        // request the provider will reject at query time, which is a worse place to find out.
+        if self.period == 0 || self.period > MAX_PERIOD_SECONDS {
```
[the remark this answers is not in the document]
ThisIsMani: Fixed in 0123fe4ed5. Validation now mirrors the backend's own rule — 1, 5, 10, 20, 30 or a multiple of 60, capped at a day — rather than checking a range. The batch-poisoning is the part that made it worth catching at boot: `metrics_service` validates every query before sending, so one alarm at 61 would have failed the whole `MetricRequest` and blinded the other 52 rules. Tested both directions, including that every period the catalogue actually uses (60 and 300) passes.

crates/observability/src/settings/cloudwatch.rs
```diff
@@ -0,0 +1,688 @@
...
+            }
+
+            fn visit_str<E: de::Error>(self, text: &str) -> Result<Self::Value, E> {
+                Dimensions::parse(text).map_err(de::Error::custom)
+            }
+
+            fn visit_map<M: de::MapAccess<'de>>(self, mut map: M) -> Result<Self::Value, M::Error> {
+                let mut dimensions = BTreeMap::new();
+                while let Some((name, value)) = map.next_entry::<String, String>()? {
+                    dimensions.insert(name, value);
```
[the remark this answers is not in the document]
ThisIsMani: Fixed in 0123fe4ed5. Both forms now go through one `insert`, so the table path gets the same non-empty name and value checks and the same duplicate rejection. Test added covering `{"Role":""}`, an empty key, and a whitespace-only value.

config/cloudwatch_alarms.sandbox.env
```diff
@@ -0,0 +1,477 @@
...
+OBSERVABILITY__CLOUDWATCH__ALARMS__RDS_FAILOVER_CPU__NAME=rds-failover-cpu
+OBSERVABILITY__CLOUDWATCH__ALARMS__RDS_FAILOVER_CPU__CLASSIFICATION=rds-alerts
+OBSERVABILITY__CLOUDWATCH__ALARMS__RDS_FAILOVER_CPU__METRIC_NAME=CPUUtilization
+OBSERVABILITY__CLOUDWATCH__ALARMS__RDS_FAILOVER_CPU__NAMESPACE=AWS/RDS
+OBSERVABILITY__CLOUDWATCH__ALARMS__RDS_FAILOVER_CPU__DIMENSIONS=DBInstanceIdentifier=failover-replica-1
+OBSERVABILITY__CLOUDWATCH__ALARMS__RDS_FAILOVER_CPU__PERIOD=60
+OBSERVABILITY__CLOUDWATCH__ALARMS__RDS_FAILOVER_CPU__STATISTIC=Average
+OBSERVABILITY__CLOUDWATCH__ALARMS__RDS_FAILOVER_CPU__SEVERITIES__SEV1__THRESHOLD=90
+OBSERVABILITY__CLOUDWATCH__ALARMS__RDS_FAILOVER_CPU__SEVERITIES__SEV1__COMPARISON_OPERATOR=GreaterThanOrEqualToThreshold
+OBSERVABILITY__CLOUDWATCH__ALARMS__RDS_FAILOVER_CPU__SEVERITIES__SEV1__EVALUATION_PERIODS=1
```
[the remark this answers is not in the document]
ThisIsMani: Not changing this one, and deliberately: those descriptions are transcribed verbatim from the source catalogue, and the mismatch exists there — `rds-primary-cpu` sev1 really is `period = 60`, `evaluation_periods = 1` with text saying "for 5 minutes". Editing the sentence here would make the file disagree with the rules that create the live AWS alarms, which is the one property this transcription has to keep. Worth a fix upstream in hyperswitch-infra, where it would correct both the AWS alarm and this copy; noting it there rather than patching the copy.

crates/observability/src/settings/cloudwatch.rs
ThisIsMani: Done in 0123fe4ed5 — the file went from 688 lines to 575 while gaining two tests, so the prose is most of what left. What stayed is the handful of things the code cannot say: why dimensions are a string rather than a table (the lowercasing), why the period rule is what it is (the provider rejects the whole batch), and what a NaN threshold does. Test names carry the rest.

config/cloudwatch_alarms.sandbox.env
ThisIsMani: Yes — TOML parses unchanged. `Dimensions` already accepts a table as well as the string, so `[cloudwatch.alarms.rds_primary_cpu]` with a `[...severities.sev1]` sub-table deserializes into exactly the same types, and there is a test covering the table form.

Two things stopped me making this file TOML:

1. **The chart only delivers env.** `hyperswitch-observability-plane` renders `.Values.configs` through `configToEnvVars` — maps recurse, slices join on commas, leaves stringify — so a TOML file in the repo never reaches the pod. Only `config/observability.toml` baked into the image does, plus env.
2. **These values are environment-specific.** The dimension identifiers here are sandbox's. The TOML in the image is shared by every environment, so putting them there is wrong the moment a second environment deploys this.

So env is the form that actually deploys today. If we are willing to add a ConfigMap mount to the chart, a TOML or JSON file becomes the better artifact — far more readable than 405 variables, and closer to what #23444 will render — and I would happily switch. Say which you prefer and I will make it the single form; I would rather not carry both, since two copies of twenty definitions will drift.

ThisIsMani: Removed in b466ea392f — the file is gone from the repo, and the answer to "TOML or env" turned out to be neither: it goes straight into the chart values, written as a `configs.cloudwatch.alarms` tree that `configToEnvVars` renders into the same 405 variables. Verified by rendering the real chart, extracting the ConfigMap and booting the service from it.

The sandbox values are recorded on https://github.com/juspay/hyperswitch-cloud/issues/23444#issuecomment-5631724945 with their provenance, so this PR is only the code that reads them.

crates/observability/src/settings/cloudwatch.rs
```diff
@@ -0,0 +1,584 @@
...
+                        .insert(name.trim(), value.trim())
+                        .map_err(de::Error::custom)?;
+                }
+                Ok(dimensions)
+            }
+        }
+
+        deserializer.deserialize_any(DimensionsVisitor)
+    }
+}
```
ThisIsMani: Can we follow the deserialiser patterns we use for these kind of envs in router... we should have some functions already lying around...

crates/observability/src/settings/cloudwatch.rs
```diff
@@ -0,0 +1,584 @@
...
+
+        for pair in text.split(',').filter(|pair| !pair.trim().is_empty()) {
+            let (name, value) = pair
+                .split_once('=')
+                .ok_or_else(|| format!("`{}` is not a `Name=value` pair", pair.trim()))?;
+            dimensions.insert(name.trim(), value.trim())?;
+        }
+
+        Ok(dimensions)
+    }
```
ThisIsMani: These should also be solved by the deserialisers written in router.

crates/observability/src/settings/cloudwatch.rs
ThisIsMani, line 1: Do you think the validations should stop the application from booting? We can talk about this once.

crates/observability/src/settings/cloudwatch.rs
```diff
@@ -0,0 +1,597 @@
...
+            value: value.to_owned(),
+        })
+    }
+}
+
+/// A comma-separated value, as a set.
+///
+/// `router` carries the same helper for its own configuration. Copied rather than shared, so a
+/// change made for one of its settings cannot quietly change how alarms are read here.
+fn deserialize_hashset<'de, D, T>(deserializer: D) -> Result<HashSet<T>, D::Error>
```
ThisIsMani: Can we move this to common settings utils file instead?

ThisIsMani: Moved in 229d33a0ec to `crates/observability/src/settings/utils.rs`, together with `validate_config_ids`, which chat, email and the catalogue all already shared from `settings.rs`. So the section files hold configuration and the helpers sit in one place.

crates/observability/src/settings/cloudwatch.rs
ThisIsMani: Done in abdde303ab. It now follows the router pattern — `FromStr` on the element, and a deserializer that splits the value — instead of a hand-written `Deserialize` for the collection:

```rust
pub struct Dimension { pub name: String, pub value: String }
impl FromStr for Dimension { /* splits on the first `=` */ }

pub dimensions: HashSet<Dimension>,
```

Per your other note the helper is a copy of the router's rather than a shared extraction, with a comment saying so, so a change made for one of the router's settings cannot alter how alarms are read here.

crates/observability/src/settings/cloudwatch.rs
ThisIsMani: Same commit. The comma splitting, the per-element error accumulation and the `D::Error::custom` wrapping all come from that helper now; what stayed local is the pair grammar itself, which the router has no equivalent for — `deserialize_hashset` splits a comma list and `deserialize_hashmap` takes a table of comma lists, neither parses `Name=value`. That part is ~12 lines of `FromStr`.

One behavioural consequence worth flagging: a set cannot reject the same dimension name twice, so that check moved into `validate` beside the other rules, with a test for it.

crates/observability/src/settings/cloudwatch.rs
ThisIsMani: Happy to talk it through. My view, with the reasoning rather than a verdict:

Boot-fail follows what the crate already does — an empty `internal_api_key`, an unusable email transport or a destination id that cannot round-trip through the environment all refuse to start today, on the principle that a misconfiguration should surface while somebody is still watching the deploy rather than on the first alert.

The argument against is real though, and it grows as the catalogue does. This is bulk data from another system, not a handful of fields someone typed. Refusing to boot on one bad rule takes down the notify routes too, which have nothing to do with CloudWatch — and once #23444 renders the catalogue from Terraform, a bad entry can appear without anyone touching this repo.

So I would draw it by who wrote the value: the sections a human configures (auth, destinations) keep failing loudly, while the catalogue could skip the definitions it cannot parse, log them and report them — evaluating the other 52 rules rather than none. The cost is that a silently-skipped rule is a rule nobody is watching, so it would need to be visible: counted in the boot log and returned in the evaluate response, not just logged once.

What I would not do is make it a config flag; that just moves the decision to whoever is deploying at the time.

Left as is for now — every validation in this PR is on hand-written values, so there is nothing to skip yet. Worth settling before the Terraform-rendered catalogue lands.

PR #14172, opened 2026-09-11
feat(payments): Add support for create intent to have upated intent + combined pml + session tokens flow

prasunna09:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

Extends the server-integration response shape introduced for update intent in #14050
to the create-intent endpoint, and moves the shared enrichment logic into one module
so both flows use the same code path.

`POST /payments` now reads `X-Integration-Type`, the same way `POST /payments/{payment_id}` does:

- `client`, an unrecognised value, or **no header at all** → the response is
  byte-for-byte what it is today. This is the default.
- `server` → the same payment response, additionally carrying:
  - `payment_method_list` — the combined payment-method list, identical in shape
    to `GET /payments/{id}/payment-methods/client`
  - `session_tokens` — the full `PaymentsSessionResponse`, identical to
    `POST /payments/session_tokens`

Differences from the update route, both deliberate:

- `POST /payments` only ever authenticates the merchant (API key, or a dashboard JWT
  outside production), so there is no auth-flow gate; the header alone opts in.
- A create-and-confirm request (`confirm: true`) keeps the plain response shape even
  with the header. The enrichment serves a checkout that has yet to confirm, and the
  session core rejects the statuses a confirmed intent lands in.

Both sections are best-effort: the intent is already created by the time they run, so a
failing or timed-out section reports its own error inline and the response still succeeds.
Each section is bounded by `SERVER_INTEGRATION_SECTION_TIMEOUT`, now derived from the
outbound HTTP client timeout (`REQUEST_TIME_OUT`, 30s) rather than a local 10s constant,
as requested in the review of #14050.

Structure:

- `core/payments/server_integration.rs` (new) — header parsing, the concurrent fetch of
  both sections, timeouts, and error mapping. Moved verbatim from the old update module.
- `core/payments/create_intent.rs` (new) and `core/payments/update_intent.rs`
  (renamed from `update_context.rs`) — thin per-flow wrappers that attach the fetched
  sections to their response.
- `routes/payments.rs` — `payments_create` gains the same opt-in and enrichment path as
  `payments_update`. Enrichment inputs are only cloned when the server shape is requested.
  Comments on both routes trimmed per review feedback on #14050.

No new route is introduced.

### Additional Changes

- [x] This PR modifies the API contract
- [ ] This PR modifies the database schema
[...]

PR #14173, opened 2026-09-11
feat(merchant_account): add integration_type and validate X-Integration-Type against it

prasunna09:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

Adds a merchant-level `integration_type` (`client`, `server`, `client_and_server`) to the merchant account: a new nullable column, the diesel and domain models, the v1 create/update/response API models, and the OpenAPI spec.

The update-intent route now rejects a request whose `X-Integration-Type` header does not match the account: a `client` account may not send `server`, a `server` account must send `server` (an absent header reads as `client`), and a `client_and_server` account accepts either. Accounts that never set the value read as `client_and_server`, so existing behaviour is unchanged until a merchant opts into a stricter setting.

### Additional Changes

## How did you test it?

Merchant config  │ no header │ client │ server │
├───────────────────┼───────────┼────────┼────────┤
│ no config         │ 200       │ 200    │ 422    │
├───────────────────┼───────────┼────────┼────────┤
│ client            │ 200       │ 200    │ 422    │
├───────────────────┼───────────┼────────┼────────┤
│ server            │ 422       │ 422    │ 200    │
├───────────────────┼───────────┼────────┼────────┤
│ client_and_server │ 200       │ 200    │ 200    │

all the test cases are added in the attached issue

## Checklist

### DIFF
--- a/config/superposition_seed.toml
+++ b/config/superposition_seed.toml
@@ -2672,6 +2672,11 @@ pt_mapping_outgoing_connector_webhooks = { value = { default_mapping = { frequen
 "vaulting.save_wallet_decrypted_data" = { value = false, schema = { type = "boolean" }, description = "Whether to save wallet decrypted data", change_reason = "Initial setup" }
 "vaulting.fingerprint_secret" = { value = "", schema = { type = "string" }, description = "Fingerprint secret used as a migration fallback for merchant accounts", change_reason = "Retain during merchant account migration" }
 "payments.should_call_gsm" = { value = false, schema = { type = "boolean" }, description = "Whether to call GSM (Global Status Mapping) for auto-retries on payment failure", change_reason = "Initial setup" }
+"payments.integration_type" = { value = "client_and_server", schema = { type = "string", enum = [
+    "client",
+    "server",
+    "client_and_server",
+] }, description = "Which integration the merchant builds its checkout with. The X-Integration-Type header on payment requests must match it: 'client' rejects 'server', 'server' requires 'server' (an absent header reads as 'client'), 'client_and_server' accepts either. Override per processor_merchant_id.", change_reason = "Merchant-level gate for the server-integration response shape" }
 "system.should_call_pm_modular_service" = { value = false, schema = { type = "boolean" }, description = "Whether to call the PM modular service for a provider merchant, with organization fallback", change_reason = "Initial setup" }
 "vaulting.should_disable_vault_tokenization" = { value = false, schema = { type = "boolean" }, description = "Whether to disable vault tokenization for a merchant", change_reason = "Initial setup" }
 "payments.should_enable_mit_with_limited_card_data" = { value = false, schema = { type = "boolean" }, description = "Whether to allow merchant-initiated transactions (MIT) with limited card data", change_reason = "Initial setup" }
--- a/crates/common_enums/src/enums/accounts.rs
+++ b/crates/common_enums/src/enums/accounts.rs
@@ -52,8 +52,9 @@ pub enum MerchantAccountType {
 /// Which integration a merchant builds its checkout with, and therefore which
 /// `X-Integration-Type` header values its payment requests may carry.
 ///
-/// `client_and_server` is the default: an account that never set this keeps accepting either
-/// header value, exactly as it did before the setting existed.
+/// Resolved per merchant from Superposition (`payments.integration_type`). `client_and_server`
+/// is the default: a merchant with no override keeps accepting either header value, exactly as
+/// it did before the setting existed.
 #[derive(
     Clone,
     Copy,
@@ -65,9 +66,7 @@ pub enum MerchantAccountType {
     serde::Serialize,
     strum::Display,
     strum::EnumString,
-    ToSchema,
 )]
-#[router_derive::diesel_enum(storage_type = "text")]
 #[strum(serialize_all = "snake_case")]
 #[serde(rename_all = "snake_case")]
 pub enum MerchantIntegrationType {
--- a/crates/common_enums/src/enums/accounts.rs
+++ b/crates/common_enums/src/enums/accounts.rs
@@ -49,6 +49,37 @@ pub enum MerchantAccountType {
     Connected,
 }
 
+/// Which integration a merchant builds its checkout with, and therefore which
+/// `X-Integration-Type` header values its payment requests may carry.
+///
+/// `client_and_server` is the default: an account that never set this keeps accepting either
+/// header value, exactly as it did before the setting existed.
+#[derive(
+    Clone,
+    Copy,
+    Debug,
+    Default,
+    Eq,
+    PartialEq,
+    serde::Deserialize,
+    serde::Serialize,
+    strum::Display,
+    strum::EnumString,
+    ToSchema,
+)]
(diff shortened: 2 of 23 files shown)


config/superposition_seed.toml
```diff
@@ -2672,6 +2672,11 @@ pt_mapping_outgoing_connector_webhooks = { value = { default_mapping = { frequen
 "vaulting.save_wallet_decrypted_data" = { value = false, schema = { type = "boolean" }, description = "Whether to save wallet decrypted data", change_reason = "Initial setup" }
 "vaulting.fingerprint_secret" = { value = "", schema = { type = "string" }, description = "Fingerprint secret used as a migration fallback for merchant accounts", change_reason = "Retain during merchant account migration" }
 "payments.should_call_gsm" = { value = false, schema = { type = "boolean" }, description = "Whether to call GSM (Global Status Mapping) for auto-retries on payment failure", change_reason = "Initial setup" }
+"payments.integration_type" = { value = "client_and_server", schema = { type = "string", enum = [
```
kashif-m, line 2675: Name it `system.payment_integration_type`

crates/router/src/core/configs/dimension_config.rs
```diff
@@ -917,6 +917,25 @@ impl DatabaseBackedConfig for OfferEngineCredentialSource {
     const KEY: &'static str = "offer_engine.credential_source";
 }
 
+config! {
+    superposition_key = MERCHANT_INTEGRATION_TYPE,
+    output = common_enums::MerchantIntegrationType,
+    default = common_enums::MerchantIntegrationType::ClientAndServer,
+    string_enum = true,
+    requires = dimension_state::DimensionsWithProcessorAndProviderMerchantId,
```
kashif-m, line 925: Dimension state is ProcessorAndProviderMerchantId, but we're creating only with `.get_processor_merchant_id()`?

kashif-m (dismissed):

crates/router/src/routes/payments.rs
```diff
@@ -943,6 +943,13 @@ pub async fn payments_update(
         move |state, auth: auth::AuthenticationData, req, req_state| {
             let header_payload = header_payload.clone();
             async move {
+                let merchant_integration_type =
+                    update_context::merchant_integration_type(&state, &auth.platform).await;
```
kashif-m, line 947: Do we not need to add this for https://github.com/juspay/hyperswitch/pull/14172?

config/superposition_seed.toml
prasunna09: Renamed to `system.payment_integration_type` — seed config, the `MERCHANT_INTEGRATION_TYPE` const and the `DatabaseBackedConfig::KEY`.

Heads up: main has since added `system.payment_method_integration_type` (vault_then_pay / pay_then_vault) and the two now sit next to each other in the seed. Different settings, similar names — happy to pick something more distinct if you would rather avoid the collision.

crates/router/src/core/configs/dimension_config.rs
prasunna09: Deliberate, and it matches the existing configs here. The `requires` clause drives the Superposition context, so both the processor and provider merchant ids are sent and an override can target either. `db_key` is only the `configs`-table fallback, and every sibling keys that on the processor merchant alone — `requires_cvv`, `client_session_validation_enabled` and `save_wallet_decrypted_data` all pair the same dimension state with a processor-only db key.

Added a comment on `db_key` spelling that out, since it reads as a mismatch otherwise.

crates/router/src/routes/payments.rs
prasunna09: Good catch — added. `payments_create` now parses the header and runs the same `validate_integration_type` check, so this PR stands on its own regardless of which of the two lands first.

Worth noting it is a real gate on create even without #14172: the create route ignored the header entirely before, so a `client`-only merchant sending `server` now gets a 400 instead of silence. Default stays `client_and_server`, so no existing merchant changes behaviour.

prasunna09: Correction to my note above: the rejection is **HTTP 422** (`InvalidRequestData`), not 400.

Verified locally across the full matrix (Superposition serving the key from the seed fallback, one merchant per value):

| merchant | header `server` | header `client` | no header |
|---|---|---|---|
| `client_and_server` | 200 | 200 | 200 |
| `client` | 422 | 200 | 200 |
| `server` | 200 | 422 | 422 |

Same result on both routes — create and update each reject independently.

One thing worth a look: for a `server` merchant with no header at all, the message reads ``header value `client` does not match…``, because an absent header resolves to `client` before the check. It is accurate but reads oddly to a caller who sent nothing. Happy to special-case the absent-header wording if you think it is worth it.

crates/common_enums/src/enums/accounts.rs
```diff
@@ -49,6 +49,36 @@ pub enum MerchantAccountType {
     Connected,
 }
 
+/// Which integration a merchant builds its checkout with, and therefore which
+/// `X-Integration-Type` header values its payment requests may carry.
+///
+/// Resolved per merchant from Superposition (`system.payment_integration_type`). `client_and_server`
+/// is the default: a merchant with no override keeps accepting either header value, exactly as
+/// it did before the setting existed.
```
kashif-m: nit: we can make it concise

kashif-m (dismissed):

crates/router/src/core/payments/update_context.rs
```diff
@@ -68,6 +68,53 @@ pub fn integration_type_from_headers(
...
+        .await
+}
+
+/// Rejects a request whose `X-Integration-Type` header does not match the integration the
+/// merchant is configured for.
+///
+/// A `client_and_server` merchant may send either value. A `client` or `server` merchant must
+/// send its own; an absent header reads as `client`, so a `server` merchant has to send the
+/// header on every request this check guards.
+pub fn validate_integration_type(
```
kashif-m, line 89: nit: we can add a TC for
1. Server <> client auth
2. Server <> API auth
3. Client <> client auth
4. Client <> API auth

crates/common_enums/src/enums/accounts.rs
prasunna09: Trimmed — the enum doc is down to two lines, and I gave the same treatment to the `merchant_integration_type` / `validate_integration_type` docs and the `db_key` comment. Net 26 lines of comment down to 15.

Also flipping the default in the same push: `client` instead of `client_and_server`, in the macro default, the `#[default]` variant and the seeded value. Before this header existed every caller was a client integration, so an unconfigured merchant should keep that behaviour and opt in explicitly before it can request the server shape.

ShankarSinghC (dismissed):

kashif-m (approved):

ShankarSinghC (approved):

crates/router/src/core/payments/update_context.rs
prasunna09: added in the pr issue

prasunna09: client auth is not supported for both the create api and update api.

PR #14174, opened 2026-09-11
feat(payments): keep the vault session and combined PML stable per payment

prasunna09:
## Type of Change

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

The server-integration enrichment of an intent update, the standalone session-token call and the SDK's own list call each minted a fresh PM vault session and a fresh combined payment-method list, so a server integration and its SDK saw different SDK authorizations and different saved-method tokens for the same payment.

Both are now cached in Redis per payment:

- The PM vault session is cached for as long as the PM service says the session lives, keyed on the customer and storage type it was created under, so an intent update that changes either mints a new one.
- The combined payment-method list is cached for the profile's order fulfillment time (the lifetime of the payment tokens inside it), keyed on the intent's `modified_at`, so an intent update rebuilds it. The attempt's `modified_at` is deliberately not part of the key: building the list writes routing results to the attempt.

A cache miss or a cache write failure is never fatal; the worst case is a fresh session or list, which is what every call used to get. The Redis read/write helpers are shared between the two.

### Additional Changes

## How did you test it?

test cases are in the attached issue

## Checklist

### DIFF
--- a/crates/common_utils/src/id_type/payment.rs
+++ b/crates/common_utils/src/id_type/payment.rs
@@ -63,6 +63,24 @@ impl PaymentId {
         )
     }
 
+    /// Get the Redis key under which the payment-method vault session for this payment is cached
+    pub fn get_pm_vault_session_redis_key(&self, merchant_id: &super::MerchantId) -> String {
+        format!(
+            "{}_{}_pm_vault_session",
+            merchant_id.get_string_repr(),
+            self.get_string_repr()
+        )
+    }
+
+    /// Get the Redis key under which the combined payment-method list for this payment is cached
+    pub fn get_combined_pm_list_redis_key(&self, merchant_id: &super::MerchantId) -> String {
+        format!(
+            "{}_{}_combined_pm_list",
+            merchant_id.get_string_repr(),
+            self.get_string_repr()
+        )
+    }
+
     /// Generate a test payment id with prefix test_
     pub fn generate_test_payment_id_for_sample_data() -> Self {
         let id = generate_id_with_default_len("test");
--- a/crates/router/src/core/payment_methods/client.rs
+++ b/crates/router/src/core/payment_methods/client.rs
@@ -27,6 +27,7 @@ use crate::{
             cards, transformers::list_customer_payment_methods_from_modular_service,
         },
         payments::helpers,
+        utils as core_utils,
     },
     pii::PeekInterface,
     routes::{self, payment_methods::ParentPaymentMethodToken},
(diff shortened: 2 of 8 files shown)


crates/router/src/core/payments/vault_session.rs
```diff
@@ -108,6 +110,131 @@ where
     Ok(())
 }
 
+/// What the internal PM service handed back for a freshly created session.
+#[cfg(feature = "v1")]
+struct CreatedPmVaultSession {
```
kashif-m, line 115: nit: move these structs to types module

crates/router/src/core/payment_methods/client.rs
```diff
@@ -716,12 +733,100 @@ pub async fn list_payment_methods_client(
...
+    })
+}
+
+/// Returns the combined list for this payment, building it on first use.
+///
+/// The cache lives no longer than the payment tokens inside the list: those expire after the
+/// profile's order fulfillment time from the moment they were minted, so the entry gets that
+/// window less however long the build took. A cache miss or a cache write failure is never
+/// fatal; the worst case is a fresh list, which is what every call used to get.
+async fn get_or_build_client_payment_methods_list(
```
kashif-m, line 750: We need to make this function a single writer

How are concurrent requests handled in this? Since there is no lock for GET PML

crates/router/src/core/payments/vault_session.rs
```diff
@@ -108,6 +110,131 @@ where
...
+#[cfg(feature = "v1")]
+pub async fn get_or_create_pm_vault_session(
+    state: &SessionState,
+    external_vault_profile: &domain::Profile,
+    merchant_id: &id_type::MerchantId,
+    payment_id: &id_type::PaymentId,
+    customer_id: Option<&id_type::CustomerId>,
+    storage_type: common_enums::StorageType,
+) -> RouterResult<Option<api::VaultDetails>> {
+    let redis_key = payment_id.get_pm_vault_session_redis_key(merchant_id);
```
kashif-m, line 164: Check if this is the correct merchant ID? In context of platform? Provider vs processor merchant ID?

crates/router/src/core/payment_methods/client.rs
```diff
@@ -716,12 +733,100 @@ pub async fn list_payment_methods_client(
...
+        .payment_id
+        .get_combined_pm_list_redis_key(&payment_intent.merchant_id);
+
+    let cached = core_utils::read_cached_value::<CachedClientPaymentMethodsList>(
+        state,
+        &redis_key,
+        CachedClientPaymentMethodsList::TYPE_NAME,
+    )
+    .await
+    .filter(|cached| cached.is_for(payment_intent_context));
```
kashif-m, line 766: There can be multiple ways of filtering the cached value; what if MCA is disabled / SPMs are deleted? Merchant disables PMs? We need to consider those invalidation cases as well or else it will always give out a stale response

crates/router/src/core/payments/vault_session.rs
prasunna09: Moved to `core/payments/types.rs` in 25e4c94d2b.

crates/router/src/core/payments/vault_session.rs
prasunna09: Good catch — it was wrong. A payment is unique under its *processor* merchant (that is the pair every lookup of it uses), but the intent's own `merchant_id` is the *provider* merchant, which in a platform flow is shared by every connected merchant under it. Two connected merchants choosing the same `payment_id` would have collided on one cached session. Now keyed on `platform.get_processor().get_account().get_id()` in 25e4c94d2b.

crates/router/src/core/payment_methods/client.rs
prasunna09: You were right that there was no single writer, and rather than add a lock I dropped the response cache — see the reply on the invalidation comment for why.

What is left to keep steady is the payment token, which is the one part of the listing that is newly generated on every call. It is now pinned per `(payment, payment_method_id)` with SETNX: the first call to store its token wins, and a call racing it reads that token back instead of its own, so concurrent callers answer with the same token without a lock. Both tokens resolve to the same saved payment method, so the one that loses is simply unused and expires.

Verified with ten concurrent `GET /payments/{id}/client` calls: one distinct token, one distinct response body across all ten.

crates/router/src/core/payment_methods/client.rs
prasunna09: Agreed, and this is why I stopped caching the response. There is no set of invalidation hooks that covers MCA disabled, saved methods deleted, payment methods disabled, blocklist entries added and the config lookups behind the list, so the cache would have gone stale in ways we could not enumerate.

The listing is now built on every call from live data — nothing to invalidate. Only the payment token is pinned per `(payment, payment_method_id)`, and the responses are still identical across calls because the rest of the build is deterministic: the enabled methods are also sorted now, since they are consolidated through `HashMap`s whose iteration order varies run to run.

Verified: disabling the connector drops `payment_methods_enabled` to 0 on the very next call, re-enabling restores it, and the response then matches the original byte for byte with the same pinned token.

crates/router/src/core/payments/types.rs
```diff
@@ -477,3 +477,21 @@ impl ForeignTryFrom<&api_models::payments::ExternalThreeDsData> for Authenticati
         })
     }
 }
+
+/// What the internal PM service handed back for a freshly created payment-method vault session.
+#[cfg(feature = "v1")]
+pub struct CreatedPmVaultSession {
+    pub vault_details: Option<crate::types::api::VaultDetails>,
```
kashif-m: nit: 
```suggestion
    pub vault_details: Option<types::api::VaultDetails>,
```

crates/router/src/core/payments/types.rs
```diff
@@ -477,3 +477,21 @@ impl ForeignTryFrom<&api_models::payments::ExternalThreeDsData> for Authenticati
...
+
+/// The vault session cached per payment, so every call for that payment hands the SDK the same
+/// authorization. `customer_id` and `storage_type` travel with it so an intent update that
+/// changes either mints a fresh session instead of reusing one created under different terms.
+#[cfg(feature = "v1")]
+#[derive(Debug, serde::Serialize, serde::Deserialize)]
+pub struct CachedPmVaultSession {
+    pub customer_id: Option<common_utils::id_type::CustomerId>,
+    pub storage_type: common_enums::StorageType,
+    pub vault_details: crate::types::api::VaultDetails,
```
kashif-m:
```suggestion
    pub vault_details: types::api::VaultDetails,
```

kashif-m (dismissed):

crates/router/src/core/payments/types.rs
prasunna09: Applied in 49c86c9977: `types` is now imported in the module, so both fields read `types::api::VaultDetails`.

crates/router/src/core/payments/types.rs
prasunna09: Applied in 49c86c9977: `types` is now imported in the module, so both fields read `types::api::VaultDetails`.

crates/common_utils/src/id_type/payment.rs
```diff
@@ -63,6 +63,29 @@ impl PaymentId {
...
+    pub fn get_pm_vault_session_redis_key(&self, merchant_id: &super::MerchantId) -> String {
+        format!(
+            "{}_{}_pm_vault_session",
+            merchant_id.get_string_repr(),
+            self.get_string_repr()
+        )
+    }
+
+    /// Get the Redis key under which this payment's token for one saved payment method is pinned
+    pub fn get_pm_token_redis_key(
```
ShankarSinghC, line 76: verify if the tagging the payment token to a payment id causes a problem in case of manual retry

ShankarSinghC (approved):

kashif-m (approved):

PR #14181, opened 2026-09-11
ci: add deja-art label required check on pull requests

### LINKED ISSUE #14099: [FEATURE] Make the deja-art label a required check on pull requests

Sayandevxyz:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [x] CI/CD

## Description
Every pull request in Hyperswitch should undergo a Deja ART replay prior to merging. Today, adding the `deja-art` label is optional and can easily be overlooked.

This PR adds a GitHub Actions workflow (`.github/workflows/deja-art-label-check.yml`) that:
- Triggers on `pull_request` (`opened`, `reopened`, `synchronize`, `ready_for_review`, `labeled`, `unlabeled`) and `merge_group` (`checks_requested`).
- Exposes a single job named `Verify PR has the deja-art label`.
- Inspects the event payload (`${GITHUB_EVENT_PATH}`) for the `deja-art` label without needing code checkouts, bot tokens, or long-running runners.
- Fails with an annotation (`::error::PR is missing the required 'deja-art' label`) if the label is missing, and succeeds once it is added.
- Always runs explicitly rather than using a job-level `if` skip, ensuring required status checks evaluate the actual presence of the label.
- Always passes on `merge_group` to prevent stalling the merge queue.

After merging, repo admins can add `Verify PR has the deja-art label` as a required status check in the "Default Branch" ruleset.

### Additional Changes

## Motivation and Context
Closes #14099

## How did you test it?
- Validated YAML syntax against GitHub Actions schema.
- Verified exact label extraction and match semantics against mock event payloads (empty labels, unrelated labels, substring matches, and matching label).
- Verified `merge_group` bypass behavior.

## Checklist


abhisheksharma2411 (commented):
The goal is sound — "should undergo a replay before merging" enforced by convention alone is enforced by nobody — and the mechanics are careful: `grep -Fx` is an exact match so a future `deja-art-skip` can't satisfy it, and giving `merge_group` its own passing step means the check still reports into the queue instead of hanging it. Both are the details this kind of workflow usually gets wrong.

Three things I'd want settled before it's made required.

## 1. This PR does not satisfy its own rule

```
$ gh pr view 14181 --json labels
(no labels)
```

So once merged, the rule's own PR would have been blocked by it. Not fatal — but it makes the point that a CI-only change now needs a Deja ART replay label, and there is no opt-out path in the workflow. Worth deciding deliberately rather than discovering it on the first docs typo fix.

## 2. Nothing in the repository consumes this label yet

```
$ git grep -rn "deja-art" origin/main -- .github/ docs/
(nothing)
$ git ls-tree --name-only origin/main .github/workflows/ | grep -i deja
(nothing)
```

The `Deja ART Replay` check I've seen on my own PRs is therefore an external app rather than a workflow in-tree. That's fine, but it means this gate asserts *"a human remembered to add a label"*, not *"a replay ran"* — and those come apart exactly when it matters, e.g. a label added after the replay already failed, or added to a PR that is then force-pushed.

If the app exposes a check run, gating on **its** conclusion would be strictly stronger and wouldn't need anyone to remember anything. If it doesn't, the label is the only handle available and this is the right thing to do — worth a sentence in the PR saying which.

## 3. There is already a labeler

`Attach suitable labels to PR` runs and succeeds on this very PR. If that job can be taught to add `deja-art`, the gate becomes self-satisfying for the common case and only needs human attention when someone deliberately removes it. That inverts the failure mode from "everyone forgets, everyone gets a red check" to "the exception is explicit", which is usually the difference between a gate that sticks and one that gets made non-required six weeks later.

## Small

`labeled` / `unlabeled` are in the trigger list, which is right — without them the check never re-runs after someone adds the label, and the PR stays red until an unrelated push. Easy to leave out; good that it's there.

Sayandevxyz:
Thanks @abhisheksharma2411  for the detailed review and sharp feedback!

Non-code / CI PRs: Great point — docs, CI, and typo fixes shouldn't be blocked waiting for a Deja ART replay. We can add path-filtering so the check only runs/fails if core application directories (like crates/) are touched, or allow an explicit bypass label like skip-deja-art / M-documentation.

Label check vs Check run: The Deja ART replay runner is triggered externally when the deja-art label is added to the PR (especially for fork PRs where external runner triggering is gated). The label check is intended as an enforcement gate to ensure the replay trigger is attached prior to merge. I'll document this workflow in the PR description.

Auto-labeling: Teaching the existing pr_labeler in pr-convention-checks.yml (Attach suitable labels to PR) to auto-attach deja-art for code changes is a great idea. That makes it self-satisfying for standard PRs while keeping manual intervention only for explicit exceptions.

Would you like me to update this PR to include path-filtering / opt-out directly, or should we also update pr-convention-checks.yml to automate the label?

maverox:
Keeping this as the deja-art label check, and closing the two others that were open for the same gate: https://github.com/juspay/hyperswitch/pull/13614 and https://github.com/juspay/hyperswitch/pull/14098. Worth knowing that 14098 shares this head branch name under a different author, which is why the two have been reading as a single entry in listings.

This one wins on the trigger. Firing on `labeled` and `unlabeled` means adding the label re-runs the check by itself, where the alternative fired nothing and needed a manual re-run every time.

One thing worth carrying over from https://github.com/juspay/hyperswitch/pull/14098 before this lands, which is the one respect in which it was better. It reads the labels through the API rather than from the event payload:

```yaml
      - name: Verify PR has the deja-art label
        if: ${{ github.event_name == 'pull_request' }}
        shell: bash
        env:
          GH_TOKEN: ${{ github.token }}
          REPOSITORY: ${{ github.repository }}
          PR_NUMBER: ${{ github.event.pull_request.number }}
        run: |
          labels="$(gh pr view "${PR_NUMBER}" --repo "${REPOSITORY}" --json labels --jq '.labels[].name')"
```

with `permissions: pull-requests: read` on the workflow. The reason is that a re-run replays the original event payload, so `GITHUB_EVENT_PATH` still shows the label as missing after somebody has added it.

Your `labeled` trigger covers the normal path, where the payload is fresh. The gap it leaves is the contributor who reacts to a red check by clicking re-run rather than by toggling the label, and gets a second identical failure with the label already applied. Reading through the API makes both paths work, and costs one step rewrite.

Separately, a point raised on 13614 that applies here too: `labeled` and `unlabeled` fire on every label change, not just on `deja-art`, so the job runs on unrelated label edits. A job-level `if` on the label name would keep it quiet.

PR #14182, opened 2026-09-11
feat(observability): evaluate a rule the way CloudWatch evaluates an alarm

ThisIsMani:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [x] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

Decides what state a rule is in. Given a rule and the readings around its window, return `Ok`,
`Alarm` or `InsufficientData` — a pure function, so it can be held to AWS's published tables with
no metrics client anywhere near it. Fetching and the route follow in their own PRs.

Stacked on #14171, which reads the catalogue this evaluates. Review that first; the diff here is
the last two commits.

### The algorithm is not the obvious one

The obvious implementation fills a window's gaps straight from `treat_missing_data`. It is wrong,
and expensively so. AWS's own table says a `breaching` rule holding one good reading and four gaps
(`0 - - - -`) is **OK**; the obvious version says ALARM.

Two rules in the RDS catalogue use `breaching` — `rds-primary-cpu-sev1` and
`rds-failover-cpu-sev1` — and they exist precisely to fire when the database stops reporting CPU.
So the obvious version pages whoever is on call every time metric ingestion runs a minute late.

What CloudWatch actually does, from
[Configuring how CloudWatch alarms treat missing data](https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/alarms-and-missing-data.html):

1. Fetch wider than the window — the *evaluation range*.
2. Fill each gap in the window with the nearest unused real reading from further back.
3. Only what is still empty takes the policy value: "CloudWatch uses missing data points only as
   few times as possible."
4. Then a guard against calling an alarm on a metric that has only just stopped reporting:
   `- - - - X` waits for the next reading, `- - X - -` does not.

### The tests are AWS's tables

Both tables from that page, in its own notation, one assertion per cell:

```rust
Row { datapoints: "0 - - - -", missing: Ok, ignore: Some(Ok), breaching: Ok,    not_breaching: Ok },
Row { datapoints: "- - - - -", missing: InsufficientData, ignore: None, breaching: Alarm, not_breaching: Ok },
Row { datapoints: "- - X - -", missing: Alarm, ignore: None, breaching: Alarm,  not_breaching: Ok },
```

Both mechanisms are load-bearing, checked by mutation: removing the real-reading backfill fails
three tests, removing the premature-alarm guard fails three others.

### Two places we knowingly differ from CloudWatch

**`ignore` falls through to `missing`.** It asks for the rule's previous state, which a stateless
evaluator does not have. Rather than assert this is fine, the test proves it: on every documented
row where `ignore` names a state at all, that state is the one `missing` gives. The rows where it
says "Retain current state" are marked `None`. No rule in the catalogue uses `ignore` or `missing`,
so `InsufficientData` is currently unreachable for RDS.

**The evaluation range is `evaluation_periods + 2`.** AWS does not publish the number — it "depends
on the length of the alarm period and whether it is based on a metric with standard resolution or
high resolution" — and both worked examples pair a three-period window with a range of five. It is
one named constant with the reasoning beside it, rather than a number chosen silently, and it only
matters when readings are missing.

### The rename, in the first commit

The catalogue holds two things and the types called both of them alarms: the metric to watch, and
the threshold judgements over it. `AlarmDefinition` → `Definition` (what to measure, no opinion
about what is bad), `SeverityRule` → `Rule` (what is bad; the unit that has a state, one per live
AWS alarm).

"Alarm" now means only the AWS resource we run alongside, which is what lets *"our rule says ALARM
but the alarm says OK"* be a sentence. The settings module doc records it next to the types.

`lib.rs` claimed "deciding what is alert-worthy happens elsewhere". That is no longer true and is
corrected here rather than left to mislead.

### Additional Changes

**`cloudwatch.alarms` becomes `cloudwatch.definitions`**, and the environment variables with it:
`OBSERVABILITY__CLOUDWATCH__ALARMS__<ID>__*` → `OBSERVABILITY__CLOUDWATCH__DEFINITIONS__<ID>__*`.
Free right now — #14171 has not merged and nothing renders these yet — and expensive once Helm
does.

## Motivation and Context

Closes nothing on its own. Part of [Evaluate CloudWatch metric alarms in the observability
plane](https://github.com/juspay/hyperswitch-cloud/issues/23445), which was narrowed to
current-state-only; transitions and delivery are [its
successor](https://github.com/juspay/hyperswitch-cloud/issues/23496).

## How did you test it?

`cargo test -p observability --features v1` — 68 tests, 9 of them here. The evaluation tests need
no network and no fixtures beyond the doc's own patterns.

Not yet tested against live CloudWatch: that arrives with the fetch PR, and needs credentials for
the account the sandbox RDS lives in.

## Checklist

https://claude.ai/code/session_019j3grZ93qqCAU85i29kKPM


crates/observability/src/domain/cloudwatch.rs
ThisIsMani, line 1: Remove the comments keep only the ones necessary.

crates/observability/CONTEXT.md
ThisIsMani, line 1: Why do we need this file?

crates/observability/src/domain/cloudwatch.rs
ThisIsMani: Trimmed. Gone: every per-variant docstring on `State` and `Reading`, the module doc's framing paragraph, the restatement on `evaluate`, and the "only reachable when..." note on the premature-alarm guard.

Kept four, each because a reader would otherwise undo it:

- **Module doc, 3 lines** — `treat_missing_data` is a last resort. This is the thing someone "simplifies" into a straight gap-fill, which flips `0 - - - -` from OK to ALARM.
- **`RANGE_MARGIN = 2`** — the number is otherwise arbitrary. AWS does not publish it.
- **`policy_reading`** — one line on why `Ignore` shares a branch with `Missing`, which reads as a bug without it. Points at the test that proves it safe.
- **`window`** — one line on the older readings being spare parts.

In the tests the comments are mostly gone too; names carry it now. I split the `ignore` assertion out of the table helper into `an_ignored_gap_lands_where_missing_does`, so the claim is a test name rather than a comment inside a loop.

crates/observability/CONTEXT.md
ThisIsMani: Dropped it.

It came from a skill convention, not from this repo — no `CONTEXT.md` exists anywhere in the tree. And the vocabulary it held is already in two other places: the ticket, and the module docs on `settings/cloudwatch.rs`. Three copies of the same glossary is two too many, and the one nothing compiles against is the one that goes stale.

The part worth keeping — that a `Definition` is what to measure and a `Rule` is what is bad about it, and that "alarm" stays reserved for the AWS resource — is in the settings module doc, next to the types it describes.

PR #14183, opened 2026-09-11
fix(core): validate refund reverse cancellation_reason length before db update

### LINKED ISSUE #13619: PR #13610: refund reverse cancellation_reason length not validated before hitting varchar(255) column
## Manual review: `POST /refunds/{id}/reverse` — `cancellation_reason` length is not enforced before it reaches Postgres

**PR:** #13610 (feat(core): support for post refund void)
**Files:** `crates/api_models/src/refunds.rs`, `crates/router/src/core/refunds.rs`, `crates/router/src/routes/refunds.rs`

### Summary

`RefundReverseRequest.cancellation_reason` is documented with `#[schema(max_length = 255, ...)]`, matching the `refund.refund_reason` column, which is `character varying(255)` in Postgres. That `max_length` is a Utoipa/OpenAPI annotation only — it is not enforced by serde deserialization or by any validator on the request path. I could not find a `validate(length = ...)` (or equivalent) check on this field anywhere in `refunds.rs`.

`refund_reverse_core` (crates/router/src/core/refunds.rs) passes `req.cancellation_reason` straight through to `db.update_refund(..., RefundUpdate::MetadataAndReasonUpdate { reason: req.cancellation_reason, ... })` with no length check in between. Once the Unified Connector Service integration path is live (i.e. once this reaches the DB update instead of short-circuiting on `NotImplemented`), a `cancellation_reason` longer than 255 characters will hit the `character varying(255)` column and fail as a raw database error, surfacing to the caller as an opaque `500 Internal Server Error` (`IR_00`-style) instead of a clean `400` validation error.

### Repro (local, manual)

Built PR branch (`pull/13610/head`) from a fresh `cargo build --bin router`, ran against local Postgres/Redis with a test merchant configured against the `fauxpay` test connector, and exercised the full flow: create payment → capture → refund → reverse.

1. Create payment (dummy test card, `fauxpay` connector) → `succeeded`.
2. Refund it in full → `succeeded`.
3. Call the new endpoint with a normal reason — correctly short-circuits since UCS isn't configured in this environment:
   ```
   POST /refunds/{refund_id}/reverse
   {"cancellation_reason": "POST_AUTH_USER_DECLINE"}

   HTTP/1.1 501
   {"error":{"type":"invalid_request","message":"refund reverse is supported only through Unified Connector Service is not implemented","code":"IR_00"}}
   ```
4. Same call with a 400-character `cancellation_reason` — accepted by JSON/schema validation, still only short-circuited by the "UCS not configured" check, **not** by any length check:
   ```
   POST /refunds/{refund_id}/reverse
   {"cancellation_reason": "aaaa...(400 chars)..."}

   HTTP/1.1 501
   {"error":{"type":"invalid_request","message":"refund reverse is supported only through Unified Connector Service is not implemented","code":"IR_00"}}
   ```
   Because the 501 happens before the `db.update_refund` call, this doesn't crash today. But it confirms nothing between the HTTP boundary and the DB call rejects an oversized `cancellation_reason` — once UCS is wired up and the happy path is reached, this will hit the `varchar(255)` column directly.

Other cases verified as working correctly:
- Reversing a non-existent refund → clean `404` (`"Refund does not exist in our records."`).
- Unknown JSON field in the request body → clean `400` (`deny_unknown_fields` working as intended).
- Reversing a refund not in `Success` state — code path present (`PaymentUnexpectedState`) but not separately exercised here since I didn't have a non-`Success` refund on hand; recommend a unit/integration test for this branch specifically.

### Suggested fix

Add an explicit length check (or a `validator`-crate `#[validate(length(max = 255))]` + a call to `.validate()` in the route/core handler, matching whatever pattern the rest of `refunds.rs`/`payments.rs` uses for bounded string fields) on `cancellation_reason` before it's used to construct the DB update, so an oversized reason fails fast with a `400` instead of surfacing as a `500` from a DB constraint violation.

### Environment
- Local build from PR branch, `cargo build --bin router` (dev profile), Postgres 17/Redis 7 via Docker, `fauxpay` test connector, no real credentials or PII used (dummy Visa test card `4242 4242 4242 4242`, dev-only API keys, all local-only).

---
Filed by HS-QUALIBOT as part of an automated manual PR review pass.

Sayandevxyz:
## Type of Change

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description
In `POST /refunds/{id}/reverse`, `req.cancellation_reason` was passed straight through to `db.update_refund(..., RefundUpdate::MetadataAndReasonUpdate { reason: req.cancellation_reason, ... })` without prior length validation.

The underlying database column in Postgres (`refund.refund_reason`) is `character varying(255)`. If a merchant provides a `cancellation_reason` exceeding 255 characters, it bypasses HTTP-layer validation and fails on database update with a constraint error, resulting in an internal 500 (`IR_00`-style) error instead of an actionable 400 Bad Request error.

This PR adds:
1. `validate_cancellation_reason` helper in `crates/router/src/core/utils/refunds_validator.rs` with `MAX_CANCELLATION_REASON_LENGTH = 255`. If exceeded, it returns `errors::ApiErrorResponse::InvalidDataFormat` (`IR_05`).
2. Invokes `validator::validate_cancellation_reason(req.cancellation_reason.as_ref())?;` at the entry point of `refund_reverse_core` in `crates/router/src/core/refunds.rs`.
3. Unit tests for `validate_cancellation_reason` covering `None`, valid strings, exactly 255 characters, and oversized (> 255) strings.

### Additional Changes

## Motivation and Context
Closes #13619

## How did you test it?
- Added unit tests in `crates/router/src/core/utils/refunds_validator.rs`:
  - `test_validate_cancellation_reason_valid`: tests valid reason, None, and maximum boundary (255 chars).
  - `test_validate_cancellation_reason_exceeds_max_length`: tests 256 chars and verifies error response.

## Checklist

- [x] I added unit tests for new functionality


abhisheksharma2411 (commented):
Validating before the write rather than letting Postgres reject it is the right shape — a `value too long for type character varying(255)` surfacing as a 500 from the DB layer is a worse answer than a 400 naming the field. And 255 is the correct number: `MetadataAndReasonUpdate` lands `reason` in `refund_reason` (`diesel_models/src/refund.rs:470`), which is `VARCHAR(255)` per `migrations/2022-12-21-071825_add_refund_reason/up.sql`.

The boundary tests are the right three — exactly 255 accepted, 256 rejected, `None` accepted.

## One thing: `len()` is bytes, the column is characters

```rust
utils::when(reason.len() > MAX_CANCELLATION_REASON_LENGTH, || { … })
```

`String::len()` returns **bytes**. Postgres `VARCHAR(255)` counts **characters**. For ASCII those agree, which is why the tests pass — `"a".repeat(255)` is 255 bytes *and* 255 characters.

They diverge on any non-ASCII reason:

```
255 Japanese characters  →  765 bytes, 255 chars
```

Postgres accepts that row; this validator rejects it with "length should be less than or equal to 255 characters", which is the one thing it definitely isn't measuring. A merchant sending a localised cancellation reason gets a 400 for a value the column would have taken — and the error message tells them to shorten something that is already inside the stated limit.

The direction is the safe one (stricter, never lets an over-long value through), so this is a usability bug rather than a correctness hole. `reason.chars().count()` fixes it, and the existing tests keep passing unchanged since they're ASCII. Worth one more case:

```rust
let max_len_multibyte = Some("あ".repeat(MAX_CANCELLATION_REASON_LENGTH));
assert!(validate_cancellation_reason(max_len_multibyte.as_ref()).is_ok());
```

which fails today and passes after the change — so it pins the distinction rather than just asserting it.

## Smaller

`Option<&String>` works, but `Option<&str>` is the more usual signature and callers can pass `req.cancellation_reason.as_deref()`. No behavioural difference; it just avoids pinning the parameter to `String` specifically if a `Cow` or `&str` field ever appears.

Nice catch on the gap — `refund_reverse_core` was the only refund path passing the field through unvalidated.

Sayandevxyz:
Thanks for the great feedback and catching the UTF-8 bytes vs character
count distinction!

I have pushed an update with the requested changes:

   1. Changed reason.len() to reason.chars().count() so character length
   matches Postgres VARCHAR(255).
   2. Changed parameter type to Option<&str> in
validate_cancellation_reason and
   used .as_deref() at the call site.
   3. Added unit tests for multibyte strings: accepting 255 multibyte
   characters and rejecting 256 multibyte characters.

PR #14185, opened 2026-09-11
feat(observability): read the catalogue's metrics and answer with every rule's state

ThisIsMani:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

The endpoint. `GET /alerts/cloudwatch/evaluate` behind the existing internal-key guard reads every
configured definition from CloudWatch, evaluates all of its rules, and returns what each one
currently says.

It delivers nothing and remembers nothing — two calls a second apart may disagree only because the
metrics did. Transition detection and chat delivery are
[a separate ticket](https://github.com/juspay/hyperswitch-cloud/issues/23496).

Stacked on #14182, which is the evaluation algorithm this calls. Review that first; the diff here
is the last commit.

```
$ curl -sH "X-Internal-Api-Key: $KEY" localhost:8085/alerts/cloudwatch/evaluate
{
  "definitions": [
    {
      "id": "rds_primary_cpu",
      "name": "rds-primary-cpu",
      "classification": "rds-alerts",
      "metric_name": "CPUUtilization",
      "period_seconds": 60,
      "outcome": "evaluated",
      "readings": [12.4, 11.9, 87.2],
      "rules": [
        { "severity": "sev1", "state": "OK",    "threshold": 90.0, "description": "SEV1: ..." },
        { "severity": "sev2", "state": "ALARM", "threshold": 85.0, "description": "SEV2: ..." },
        { "severity": "sev3", "state": "ALARM", "threshold": 65.0, "description": "SEV3: ..." }
      ]
    },
    {
      "id": "rds_replica_lag",
      "outcome": "unread",
      "reason": "query_failed"
    }
  ]
}
```

`readings` rides along per definition because "why did this say ALARM" is the question the route
exists to answer, and for a while the only caller is a human with curl.

### One query per definition, one request per period

Rules under a definition share metric, dimensions, period and statistic and differ only in
threshold, so reading per rule would be the same call three times over. A `MetricRequest` carries
one range for every query it holds, so mixing periods would size each series to the widest one — a
request per period keeps each as narrow as its own rules need.

For the RDS catalogue that is **two calls covering 20 queries, feeding 53 rules**: a 60-second batch
seven periods wide (its widest rule is `evaluation_periods = 5`, plus the margin) and a 300-second
batch five wide. Each rule then slices its own tail from its definition's series.

### The window slides, and that needed checking

It ends at the latest completed minute and moves with it, which is what CloudWatch does by default —
its evaluation boundaries are explicitly *not* aligned to the wall clock. That only works if
`GetMetricData` will return 300-second buckets off the five-minute grid, which the docs do not say.
Checked against a live account:

```
start 10:48 → buckets 10:48, 10:53, 10:58, 11:03
start 10:49 → buckets 10:49, 10:54, 10:59, 11:04
```

Buckets anchor to the request's start. So a 300-second window ending at 12:07 covers 12:02–12:07
rather than snapping back to 12:00, and the same fact is what will make the successor ticket's
"compare with one minute earlier" possible at all.

### An unread definition carries no rule states

Three ways a definition goes unread, each named in the response: the call failed
(`query_failed`), the series came back `Partial`/`Failed` (`series_incomplete`), or the provider
said nothing about that query (`series_missing`).

None of them is a missing *reading*, so none of them reaches `treat_missing_data`. A definition we
could not read reporting `"state": "OK"` would be the worst lie this service could tell. Batches are
read independently, so a failed 60-second call still leaves the 300-second one answering.

`series_incomplete` is the one judgement call: a `Partial` series might still carry every reading its
window needs, and this drops it anyway. Evaluating it would silently judge a shortened window as
though the periods it lost had reported nothing, so strict seemed the right default — say if you
disagree.

### Structure, and one gap in the tests

`plan`, `absorb` and `interpret` are pure and hold every decision — batching and ranges, page
concatenation and the `Partial` rule, slicing and failure mapping. `fetch` is just the loop around
the provider.

That split is deliberate: `MetricPage`, `MetricSeries` and `Cursor` have `pub(crate)` constructors in
`external_services`, so a stub `MetricsProvider` cannot be built from this crate and the loop itself
is the one part covered by the live run rather than by a test. Making those three constructors `pub`
would close the gap — a public trait arguably cannot be implemented without them — but that is
another crate's decision, so I left it. Happy to follow up if you want it.

### Additional Changes

**New `cloudwatch.client.region`** (`OBSERVABILITY__CLOUDWATCH__CLIENT__REGION`). Boot refuses a
catalogue with definitions and no region, which would otherwise start cleanly and evaluate nothing
at all. An empty catalogue still boots without one, so a deployment that has not been given a
catalogue yet is valid rather than broken.

The client is built once at boot, matching the chat and email registries, so a region the SDK cannot
resolve stops the service starting rather than turning every evaluation into a failed batch.

## Motivation and Context

[Evaluate CloudWatch metric alarms in the observability plane](https://github.com/juspay/hyperswitch-cloud/issues/23445),
narrowed to current-state-only. This completes it.

## How did you test it?

`cargo test -p observability --features v1` — 80 tests, 12 of them new.

Not yet run against live CloudWatch. The account reachable from this machine has no RDS in it
(`list-metrics --namespace AWS/RDS` is empty), so the 53-rule run needs credentials for the sandbox
account. The grid-alignment check above did run live, against `AWS/Usage` — that behaviour is the
API's, not the metric's.

## Checklist

https://claude.ai/code/session_019j3grZ93qqCAU85i29kKPM


crates/observability/src/core/cloudwatch.rs
ThisIsMani, line 1: Reduce comments from this file.

crates/observability/src/core/cloudwatch.rs
```diff
@@ -0,0 +1,567 @@
...
+};
+use time::OffsetDateTime;
+
+use crate::{
+    domain::cloudwatch::{evaluate, evaluation_range, State},
+    logger,
+    settings::cloudwatch::{AlarmDefinition, CloudWatchSettings},
+};
+
+/// Every definition's outcome, by id.
```
ThisIsMani: Let's not define types in core... we can move this to domain.

crates/observability/src/routes/cloudwatch.rs
```diff
@@ -0,0 +1,38 @@
...
+        |state, ()| async move {
+            let catalogue = match state.metrics.as_deref() {
+                Some(provider) => {
+                    core::cloudwatch::evaluate_catalogue(
+                        provider,
+                        &state.conf.cloudwatch,
+                        date_time::now().assume_utc(),
+                    )
+                    .await
+                }
```
ThisIsMani: Let's not have this logic in routes... Move this to core. And have a internal function if required.

crates/observability/src/core/cloudwatch.rs
```diff
@@ -0,0 +1,567 @@
...
+        definitions.extend(interpret(&batch, readings));
+    }
+
+    definitions.sort_by(|one, other| one.id.cmp(&other.id));
+    Catalogue { definitions }
+}
+
+/// The requests that cover `settings`: one per period, each as wide as its own widest rule.
+fn plan(settings: &CloudWatchSettings, end: OffsetDateTime) -> Vec<Batch<'_>> {
+    let mut by_period: BTreeMap<Period, Vec<(&String, &AlarmDefinition)>> = BTreeMap::new();
```
ThisIsMani: Let's try to get rid of muts please.

crates/observability/src/core/cloudwatch.rs
```diff
@@ -0,0 +1,567 @@
...
+    }
+}
+
+/// `TimeRange` is half-open, so this as the end makes the last slot the minute before it — one
+/// CloudWatch has had a chance to receive readings for.
+fn latest_completed_minute(now: OffsetDateTime) -> OffsetDateTime {
+    now.replace_second(0)
+        .and_then(|minute| minute.replace_nanosecond(0))
+        .unwrap_or(now)
+}
```
ThisIsMani: Move to utils.

crates/observability/src/core/cloudwatch.rs
ThisIsMani: Down to four, each one a thing the code cannot say: why pages merge instead of concatenating, why the query limit is chunked, why an unread definition carries no rule states, and what the window slice is for. Everything else went, including the module doc.

crates/observability/src/core/cloudwatch.rs
ThisIsMani: Moved — `Catalogue`, `Evaluation`, `Outcome`, `Unread` and `RuleState` now live in `domain::cloudwatch` beside `State`, which they are all shaped around. `core` keeps only `Batch` and `Readings`, which are about how one request is carried out rather than what an evaluation is. `types.rs` now converts from `domain` rather than `core`.

crates/observability/src/routes/cloudwatch.rs
ThisIsMani: Gone. `core::cloudwatch::evaluate_catalogue(&state, now)` takes the `AppState` and resolves the provider itself, with a private `read(provider, settings, now)` underneath it — which is also what the tests drive, so they never build an `AppState`. The handler is now deserialize, call in, serialise.

One thing fell out of this: moving the clock read into the handler tripped `.clippy.toml`, which disallows `time::OffsetDateTime::now_utc` because a raw clock read cannot be reproduced on replay. It now goes through `common_utils::date_time::now().assume_utc()`.

crates/observability/src/core/cloudwatch.rs
ThisIsMani: Both gone. The period grouping is a `fold` into a `BTreeMap`, matching the idiom `settings::utils::deserialize_hashset_inner` already uses; `absorb` folds a page into the readings and returns them rather than taking `&mut`.

Two `mut` remain and I left them deliberately: the `for` loop in `read` that awaits each batch in turn, and the `Vec<RuleState>` sorted immediately after collection. Say the word if you want those restructured too.

crates/observability/src/core/cloudwatch.rs
ThisIsMani: Moved to a new `crate::utils` as `latest_completed_period(now, period)`, with its own tests.

It also grew while moving. Codex was right that flooring to the minute is wrong for high-resolution periods — a ten-second rule lost up to four completed periods. The boundary is now per period, capped at a minute so a 300-second window still advances a minute at a time rather than snapping to the five-minute grid.

crates/observability/src/core/cloudwatch.rs
```diff
@@ -0,0 +1,567 @@
...
+///
+/// `now` is a parameter so the window is a fact of the call rather than of the clock.
+pub async fn evaluate_catalogue(
+    provider: &dyn MetricsProvider,
+    settings: &CloudWatchSettings,
+    now: OffsetDateTime,
+) -> Catalogue {
+    let mut definitions = Vec::with_capacity(settings.definitions.len());
+
+    for batch in plan(settings, latest_completed_minute(now)) {
```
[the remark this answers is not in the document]
ThisIsMani: Correct, fixed. The boundary is now derived per period in `utils::latest_completed_period`, capped at a minute — below a minute the cadence is the period itself, at or above it stays the minute, so a 300-second window still advances a minute at a time rather than snapping to the five-minute grid.

crates/observability/src/core/cloudwatch.rs
```diff
@@ -0,0 +1,567 @@
...
+                    queries: definitions
+                        .iter()
+                        .map(|(_, definition)| MetricQuery {
+                            namespace: Some(definition.namespace.clone()),
+                            name: definition.metric_name.clone(),
+                            labels: definition.labels(),
+                            period,
+                            aggregation: definition.statistic.into(),
+                        })
+                        .collect(),
```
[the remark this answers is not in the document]
ThisIsMani: Correct, fixed. Each period group is chunked at 500 queries against a named constant that points back at the provider. Covered by a test that builds 510 definitions on one period and asserts the requests come out 500 + 10.

crates/observability/src/core/cloudwatch.rs
```diff
@@ -0,0 +1,567 @@
...
+/// A series can span pages, so values accumulate under their index. Incomplete retrieval is not a
+/// run of missing readings — a shortened window would be judged as though the periods it lost had
+/// reported nothing — so it poisons the index instead.
+fn absorb(readings: &mut Readings, index: usize, status: SeriesStatus, values: &[Option<f64>]) {
+    let entry = readings.entry(index).or_insert_with(|| Some(vec![]));
+
+    match status {
+        SeriesStatus::Complete => {
+            if let Some(collected) = entry {
+                collected.extend_from_slice(values);
```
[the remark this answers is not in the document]
ThisIsMani: Correct, and this was the serious one — thank you. Pages now merge slot by slot rather than concatenating.

Confirmed by mutation: putting the concatenation back produces `[Some(99.0), Some(99.0), None, None, None, None, None, Some(99.0), None, None]` where five slots belong, and the breach reported on the first page falls outside the window entirely, flipping the rule from ALARM to OK. That is exactly the state change you described. It is now pinned by `a_series_spanning_pages_merges_slot_by_slot`, which only became testable because `MetricPage`/`MetricSeries`/`Cursor` constructors were made `pub` in `external_services` in this same change.

PR #14187, opened 2026-09-11
fix(ucs): record request rejections as payment failures

pixincreate:
## Type of Change

- [x] Bugfix

## Description

When UCS rejected a request before calling the connector, Hyperswitch returned the error to the caller and recorded nothing on the payment. The attempt kept its previous status and the intent stayed in `processing`.

This PR records that rejection as a payment outcome and then returns the mapped error response, so the payment state moves and the API contract stays the error response:

- On the connector service path, the mapped API error is written to the attempt through the normal `PaymentResponse` post-update tracker, then the API error is returned. No hardcoded error codes.
- **Only a request-phase rejection is recorded.** A response-phase failure (for example the connector answered but UCS could not deserialize the response) leaves the outcome unknown — the payment may have succeeded at the connector — so it keeps its existing behavior and is left for sync/reconcile. The predicate lives on `ConnectorError::is_request_phase_rejection`, the enum both paths already share: native connectors raise it from their transformers, and UCS errors are switched into it. No UCS-specific downcast.
- `NotSupported` is split from `NotImplemented`. Not-supported codes (`NOT_SUPPORTED`, `FLOW_NOT_SUPPORTED`, `CAPTURE_METHOD_NOT_SUPPORTED`, `CURRENCY_NOT_SUPPORTED`) map to `IR_19` / HTTP 400; `NOT_IMPLEMENTED` maps to `IR_00` / 501. Either way the recorded attempt is marked `Failure` explicitly on the `ErrorResponse` rather than derived from the status code, so the payment does not sit in flight and the recorded state does not drift if that derivation changes.
- Other decoded classes carry their documented codes: missing required field `IR_04`, missing required fields `IR_21`, invalid data format `IR_05`.
- The rejection reason no longer names the connector twice. UCS already sends a complete, attributed message (`... is not supported by worldpayxml`), so the router surfaces it verbatim instead of re-appending `is not supported by unified_connector_service`, which `ConnectorError::NotSupported`'s `&'static str` connector field could never populate correctly.

The direct path is unchanged and needs no equivalent: it builds its connector request *before* the trackers move the payment to `processing` and aborts there, so it commits nothing and never reaches this call site. The recording is positional — anything failing after the commit point is classified by the shared predicate.

No UCS-side change is required for this fix. The router classifies on the decoded `IntegrationError` proto `error_code`, not on the gRPC status code (`decode_integration_error` never reads `status.code()`), so the wire code is only consulted in the `TonicStatus` fallback.

### Additional Changes

- [x] This PR modifies the API contract - a request rejection over UCS now records the failed attempt and returns the mapped error response (HTTP 400 for not-supported codes, HTTP 501 for not-implemented) instead of leaving the payment untouched.

## Motivation and Context

A merchant running worldpayxml over UCS reported that a request rejection leaves the payment intent in `processing` with no attempt recorded while the router answers `501 IR_00`. They reconcile on the intent state, so the payment looked in flight and their fallback never ran. This applies to any UCS connector that refuses during request construction, not only worldpayxml.

## How did you test it?

Local stack, router built from this branch, UCS `2026.09.11.1-2-g94c1ad158c75` on `:8000`, worldpayxml and stripe both routed through UCS. The direct runs were produced by flipping `rollout_percent` 1 → 0 on `*_card_Authorize` and restoring afterwards (config diffed byte-identical to the backup); every direct payment was verified absent from the UCS log.

### Matrix — both services

| case | http | code | intent | attempt | service |
|---|---|---|---|---|---|
| worldpayxml no-CVC (the reported case) | 400 | `IR_19` | `failed` | `failure` | ucs |
| worldpayxml no-CVC | 501 | - | `requires_payment_method` | `payment_method_awaited` | direct |
| stripe no-CVC | 501 | `IR_00` | `failed` | `failure` | ucs |
| stripe no-CVC | 501 | - | `requires_payment_method` | `payment_method_awaited` | direct |
| worldpayxml card (happy path) | 200 | - | `succeeded` | `charged` | ucs |
| worldpayxml card (happy path) | 200 | - | `succeeded` | `charged` | direct |
| stripe decline (connector 4xx) | 200 | `card_declined` | `failed` | `failure` | ucs |
| stripe decline (connector 4xx) | 200 | `card_declined` | `failed` | `failure` | direct |

The first row is the fix: the reported case now ends terminal instead of stranding in `processing` with nothing recorded. Rows 5–8 are the regression guard — the happy path and connector declines are untouched on both paths, and declines still return a payment object with the connector's own error code rather than an error envelope.

### What UCS sent, and what Hyperswitch made of it

| case | UCS sent | Hyperswitch rendered |
|---|---|---|
| worldpayxml no-CVC | gRPC `3 InvalidArgument`, `error_code: NOT_SUPPORTED`, `http_status_code: null`, `"Selected payment method is not supported by worldpayxml"` | `IR_19` / 400, reason = message verbatim, attempt recorded `failure` |
| stripe no-CVC | gRPC `12 Unimplemented`, `error_code: NOT_IMPLEMENTED`, `http_status_code: null` | `IR_00` / 501, attempt recorded `failure` |
| worldpayxml happy path | status 200, no gRPC error | 200 payment object, `charged` |
| stripe decline | gRPC `3 InvalidArgument`, `error_code: CONNECTOR_ERROR_RESPONSE`, `http_status_code: 402`, `"Your card was declined."` | 200 payment object, `card_declined`, attempt `failure` |

`http_status_code` `null` vs `402` is the phase marker: `null` means the connector was never called, `402` means it answered. Both arrive as gRPC `3`, which is why the classification is driven by the proto `error_code` rather than the gRPC status.

<details><summary><b>worldpayxml no-CVC over UCS — before and after</b></summary>

```sh
curl -X POST 'http://localhost:8080/payments' -H 'Content-Type: application/json' -H 'api-key: ****' \
  -d '{"amount":1000,"currency":"USD","confirm":true,"capture_method":"automatic","authentication_type":"no_three_ds","connector":["worldpayxml"],"customer_id":"...","payment_method":"card","payment_method_type":"credit","payment_method_data":{"card_with_no_cvc":{"card_number":"4242424242424242","card_exp_month":"12","card_exp_year":"30","card_holder_name":"John Doe"}}}'
```

Before — HTTP 501, nothing recorded, intent left in flight:

```json
{"error":{"type":"invalid_request","message":"Selected payment method is not supported by worldpayxml is not implemented","code":"IR_00"}}
```

```
payment_attempt: status = pending, error_code = NULL
payment_intent:  status = processing
```

After — HTTP 400, recorded, terminal, and the connector named once:

```json
{
  "error": {
    "type": "invalid_request",
    "message": "Payment method type not supported",
    "code": "IR_19",
    "reason": "Selected payment method is not supported by worldpayxml"
  }
}
```

```
payment_attempt: status = failure, error_code = IR_19, error_message = Selected payment method is not supported by worldpayxml
payment_intent:  status = failed
```

</details>

<details><summary><b>Regression guard — happy path and connector decline (UCS)</b></summary>

Happy path, worldpayxml card with CVC:

```json
{"status":"succeeded","payment_id":"pay_zoLmEJelKAaIbtz3PU5q"}
```
```
payment_attempt: charged   payment_intent: succeeded
```

Connector decline, stripe `4000000000000002` — still a payment object with the connector's code, not an error envelope:

```json
{"status":"failed","payment_id":"pay_5L3k7OpHYzomvy0AjbJr","error_code":"card_declined","error_message":"message - Your card was declined., decline_code - generic_decline"}
```
```
payment_attempt: failure / card_declined   payment_intent: failed
```

</details>

### Checks

- `cargo check -p router` passes.
- `cargo +nightly fmt --all -- --check` passes.

## Checklist

### DIFF
--- a/crates/router/src/core/payments.rs
+++ b/crates/router/src/core/payments.rs
@@ -1093,7 +1093,10 @@ where
                         )
                         .await?;
 
-                    let (router_data, mca) = Box::pin(complete_connector_service(
+                    let failed_attempt_router_data =
+                        call_connector_service_response.router_data.clone();
+
+                    let (router_data, mca) = match Box::pin(complete_connector_service(
                         &updated_state,
                         platform.get_processor(),
                         &operation,
@@ -1112,7 +1115,48 @@ where
                         call_connector_service_response,
                         &dimensions.without_profile_id(),
                     ))
-                    .await?;
+                    .await
+                    {
+                        Ok(result) => result,
+                        Err(api_error) => {
+                            // Record the rejected attempt, then return the mapped error so the
+                            // API returns the error response instead of a payment object.
+                            let status_code = {
+                                use actix_web::ResponseError;
+                                api_error.current_context().status_code().as_u16()
+                            };
+                            let mut error_response: hyperswitch_domain_models::router_data::ErrorResponse =
+                                api_error.current_context().clone().into();
+                            error_response.status_code = status_code;
+                            let mut failed_router_data = failed_attempt_router_data;
+                            failed_router_data.response = Err(error_response);
+                            failed_router_data.connector_http_status_code = Some(status_code);
+                            // Record through the same tracker the normal path uses.
+                            let operation = Box::new(PaymentResponse);
+                            if let Err(tracker_error) = operation
+                                .to_post_update_tracker()?
+                                .update_tracker(
+                                    state,
+                                    platform.get_processor(),
+                                    payment_data,
+                                    failed_router_data,
+                                    &locale,
+                                    #[cfg(all(feature = "dynamic_routing", feature = "v1"))]
+                                    routable_connectors,
+                                    #[cfg(all(feature = "dynamic_routing", feature = "v1"))]
+                                    &business_profile,
+                                    &dimensions.without_profile_id(),
+                                )
+                                .await
+                            {
+                                logger::error!(
+                                    ?tracker_error,
+                                    "failed to record the rejected attempt"
+                                );
+                            }
+                            return Err(api_error);
+                        }
+                    };
 
                     let op_ref = &operation;
                     let should_trigger_post_processing_flows = is_operation_confirm(&operation);
@@ -1281,7 +1325,10 @@ where
                         )
                         .await?;
 
-                    let (router_data, mca) = Box::pin(complete_connector_service(
+                    let failed_attempt_router_data =
+                        call_connector_service_response.router_data.clone();
+
+                    let (router_data, mca) = match Box::pin(complete_connector_service(
                         &updated_state,
                         platform.get_processor(),
                         &operation,
@@ -1300,7 +1347,48 @@ where
                         call_connector_service_response,
                         &dimensions.without_profile_id(),
                     ))
-                    .await?;
+                    .await
+                    {
+                        Ok(result) => result,
+                        Err(api_error) => {
+                            // Record the rejected attempt, then return the mapped error so the
+                            // API returns the error response instead of a payment object.
+                            let status_code = {
+                                use actix_web::ResponseError;
+                                api_error.current_context().status_code().as_u16()
+                            };
+                            let mut error_response: hyperswitch_domain_models::router_data::ErrorResponse =
+                                api_error.current_context().clone().into();
+                            error_response.status_code = status_code;
+                            let mut failed_router_data = failed_attempt_router_data;
+                            failed_router_data.response = Err(error_response);
+                            failed_router_data.connector_http_status_code = Some(status_code);
+                            // Record through the same tracker the normal path uses.
+                            let operation = Box::new(PaymentResponse);
+                            if let Err(tracker_error) = operation
+                                .to_post_update_tracker()?
+                                .update_tracker(
+                                    state,
+                                    platform.get_processor(),
+                                    payment_data,
+                                    failed_router_data,
+                                    &locale,
+                                    #[cfg(all(feature = "dynamic_routing", feature = "v1"))]
+                                    routable_connectors,
+                                    #[cfg(all(feature = "dynamic_routing", feature = "v1"))]
+                                    &business_profile,
+                                    &dimensions.without_profile_id(),
+                                )
+                                .await
+                            {
+                                logger::error!(
+                                    ?tracker_error,
+                                    "failed to record the rejected attempt"
+                                );
+                            }
+                            return Err(api_error);
+                        }
+                    };
 
                     #[cfg(all(feature = "retry", feature = "v1"))]
                     let mut router_data = router_data;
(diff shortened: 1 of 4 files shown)


crates/router/src/core/payments.rs
```diff
@@ -645,6 +645,67 @@ where
...
+    failed_router_data.connector_http_status_code = Some(status_code);
+
+    let operation = Box::new(PaymentResponse);
+    if let Err(tracker_error) = operation
+        .to_post_update_tracker()?
+        .update_tracker(
+            state,
+            processor,
+            payment_data,
+            failed_router_data,
```
hrithikesh026: I don't understand how this `failed_router_data` would have failed status.

pixincreate: the name was wrong and i've fixed it.

crates/router/src/core/payments.rs
```diff
@@ -645,6 +645,67 @@ where
...
+    PaymentResponse: Operation<F, FData, Data = D>,
+    FData: Send + Sync + Clone + router_types::Capturable + 'static + serde::Serialize,
+{
+    let status_code = {
+        use actix_web::ResponseError;
+        api_error.current_context().status_code().as_u16()
+    };
+    let mut error_response: hyperswitch_domain_models::router_data::ErrorResponse =
+        api_error.current_context().clone().into();
+    error_response.status_code = status_code;
```
gopikrishna000: this right side status code is merchant api response status code , is it same intended in this db?
this case no connector call happened

crates/router/src/core/payments.rs
```diff
@@ -645,6 +645,67 @@ where
...
+    let mut failed_router_data = failed_attempt_router_data;
+    failed_router_data.response = Err(error_response);
+    failed_router_data.connector_http_status_code = Some(status_code);
```
gopikrishna000: how this is connector_http_status_code
when no connector call was made

hrithikesh026 (dismissed):

crates/router/src/core/payments.rs
```diff
@@ -645,6 +645,83 @@ where
...
+    };
+    // Only the initial authorization can be failed outright here. For any later flow the
+    // connector has already set a status this rejection must not overwrite, so leave it unset
+    // and let the tracker's flow-aware derivation keep the prior status.
+    if matches!(
+        core_utils::get_flow_name::<F>()?.as_str(),
+        "Authorize" | "SetupMandate"
+    ) {
+        error_response.attempt_status = Some(enums::AttemptStatus::Failure);
+    }
```
hrithikesh026: Any request construction failure must set status to failure no???

hrithikesh026: Have a trait implementation on F.

gopikrishna000 (approved):

hrithikesh026 (approved):

PR #14188, opened 2026-09-12
perf(merchant_connector_account): cache MCA list queries via scope supersets

### LINKED ISSUE #14335: [FEAT] Caching: Add IMC and Redis caching support for MCA list
Add IMC and Redis caching support for MCA list because List Call for MCAs during OLTP Affect Latency and DB load.

hrithikesh026:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [x] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

All seven merchant connector account list queries hit Postgres on every call, including hot paths in routing and payments. This PR caches them as two whole-scope supersets — every account of a merchant, and every account of a merchant's profile — which each list query then projects from in memory.

Keying on scope rather than on each query's arguments makes invalidation complete by construction: any insert, update or delete redacts exactly two keys, both derivable from the account alone, so a single MCA change drops every list entry that could contain it, and a list query added later is covered by adding a projection rather than by extending a write path.

- Add `MCA_LIST_CACHE` and `CacheKind::MerchantConnectorAccountList`, wired into the pub/sub invalidation handler, the `CacheKind::All` fan-out, and the background metrics collector
- Collapse the merchant- and profile-scoped queries into `list_by_merchant_id` and `list_by_merchant_id_profile_id`, removing the seven list builders they replace, leaving no way to read a list of MCAs that skips the cache
- Decrypt list results concurrently; with the encryption service enabled each conversion is its own keymanager round-trip
- Add `MerchantConnectorAccount::is_enabled()`, mirroring the SQL `disabled = false` check that does not match `NULL`
- Fix: `insert` previously performed no cache invalidation at all, so a newly created connector would never have appeared in a cached list

### Additional Changes

## Motivation and Context

Merchant connector account list queries are on hot paths (routing, payments) and were uncached, causing repeated Postgres round-trips for data that changes infrequently relative to how often it's read.

## How did you test it?

Verified with `cargo +nightly fmt --all`, `cargo clippy` (v1) and `check_v2` across all targets, plus unit tests. Not yet exercised against a live Postgres/Redis.

## Checklist

### DIFF
--- a/crates/storage_impl/src/merchant_connector_account.rs
+++ b/crates/storage_impl/src/merchant_connector_account.rs
@@ -17,6 +17,166 @@ use crate::{
     CustomResult, DatabaseStore, MockDb, RouterStore, StorageError,
 };
 
+/// Cache keys and invalidation for the merchant connector account list caches.
+///
+/// Every list query in this module is a row-subset of one of two supersets — all accounts
+/// of a merchant, or all accounts of a profile — so those are the only two things ever
+/// cached. A write touching a single account therefore invalidates every list entry that
+/// could contain it by redacting exactly two keys, both derivable from the account alone.
+///
+/// This is the only place list cache keys are constructed. Adding a new list query means
+/// adding a projection over one of the supersets; no write path needs to change for it.
+#[cfg(feature = "accounts_cache")]
+mod list_cache {
+    use common_utils::id_type;
+
+    use crate::redis::cache::CacheKind;
+
+    /// Key of the superset holding every account of a merchant, disabled included.
+    pub(super) fn merchant_scope_key(merchant_id: &id_type::MerchantId) -> String {
+        format!("mca_list_m_{}", merchant_id.get_string_repr())
+    }
+
+    /// Key of the superset holding every account of a merchant's profile, disabled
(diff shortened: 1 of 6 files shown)


crates/storage_impl/src/merchant_connector_account.rs
```diff
@@ -378,31 +538,13 @@ impl<T: DatabaseStore> MerchantConnectorAccountInterface for RouterStore<T> {
...
+        let accounts = list_all_by_merchant_id(self, merchant_id)
+            .await?
+            .into_iter()
+            .filter(|account| account.connector_name == connector_name)
+            .collect();
+
+        decrypt_all(self, accounts, key_store).await
     }
 
     #[instrument(skip_all)]
```
hrithikesh026, line 550: Should find one call also try to fetch from the superset list?? By application level filtering

jarnura: yes

jarnura (approved):

sai-harsha-vardhan (approved):

PR #14194, opened 2026-09-13
fix(connector): [BANKOFAMERICA] forward network transaction id from payment and setup mandate responses

### LINKED ISSUE #14193: [BUG] Bank of America connector drops the network transaction id from payment and setup mandate responses
### Bug Description

The `bankofamerica` connector deserializes the card network's transaction id but never forwards it, so `network_transaction_id` is never stored for payments or saved cards processed through Bank of America.

`processorInformation.networkTransactionId` is parsed into `ClientProcessorInformation.network_transaction_id` ([L740](https://github.com/juspay/hyperswitch/blob/2c2e58976363eaa5ed0414c6fa2d355c048cb161/crates/hyperswitch_connectors/src/connectors/bankofamerica/transformers.rs#L740)), but both places that build a successful `TransactionResponse` hardcode `network_txn_id: None`:

- SetupMandate response ([L437](https://github.com/juspay/hyperswitch/blob/2c2e58976363eaa5ed0414c6fa2d355c048cb161/crates/hyperswitch_connectors/src/connectors/bankofamerica/transformers.rs#L437))
- `get_payment_response`, shared by Authorize, Capture and Void ([L1666](https://github.com/juspay/hyperswitch/blob/2c2e58976363eaa5ed0414c6fa2d355c048cb161/crates/hyperswitch_connectors/src/connectors/bankofamerica/transformers.rs#L1666))

The field has no readers anywhere in the connector.

The connectors built on the same Cybersource-style response already forward it from the same field: `cybersource` ([L3549](https://github.com/juspay/hyperswitch/blob/2c2e58976363eaa5ed0414c6fa2d355c048cb161/crates/hyperswitch_connectors/src/connectors/cybersource/transformers.rs#L3549), [L4796](https://github.com/juspay/hyperswitch/blob/2c2e58976363eaa5ed0414c6fa2d355c048cb161/crates/hyperswitch_connectors/src/connectors/cybersource/transformers.rs#L4796)) and `wellsfargo` ([L1823](https://github.com/juspay/hyperswitch/blob/2c2e58976363eaa5ed0414c6fa2d355c048cb161/crates/hyperswitch_connectors/src/connectors/wellsfargo/transformers.rs#L1823), [L2013](https://github.com/juspay/hyperswitch/blob/2c2e58976363eaa5ed0414c6fa2d355c048cb161/crates/hyperswitch_connectors/src/connectors/wellsfargo/transformers.rs#L2013)).

### Expected Behavior

When Bank of America returns `processorInformation.networkTransactionId`, it should be propagated as `network_txn_id`, as `cybersource` and `wellsfargo` do. For an off-session (`setup_future_usage: off_session`) card save, it should then be stored on the payment method.

### Actual Behavior

`network_txn_id` is always `None` for Bank of America. `tokenization.rs` therefore logs `Skip storing network transaction id` ([L242](https://github.com/juspay/hyperswitch/blob/2c2e58976363eaa5ed0414c6fa2d355c048cb161/crates/router/src/core/payments/tokenization.rs#L242)), and the saved payment method has no `network_transaction_id`.

### Steps To Reproduce

1. Configure a `bankofamerica` merchant connector account and enable `is_connector_agnostic_mit_enabled` on the profile.
2. Create and confirm a card payment with `setup_future_usage: off_session` and `customer_acceptance`. This sends Bank of America's stored-credential CIT action ([`get_boa_mandate_action_details`](https://github.com/juspay/hyperswitch/blob/2c2e58976363eaa5ed0414c6fa2d355c048cb161/crates/hyperswitch_connectors/src/connectors/bankofamerica/transformers.rs#L2699)).
3. Inspect the stored payment method: its `network_transaction_id` is `NULL`, even though the Bank of America response carried `processorInformation.networkTransactionId`.

### Context For The Bug

The network transaction id is what makes connector-agnostic MIT possible: `get_mca_status` treats a saved card as usable for agnostic MIT only when `network_transaction_id.is_some()` ([cards.rs L6550](https://github.com/juspay/hyperswitch/blob/2c2e58976363eaa5ed0414c6fa2d355c048cb161/crates/router/src/core/payment_methods/cards.rs#L6550)). Cards first saved through Bank of America can therefore never be used for a network-transaction-id MIT through another connector, even though Bank of America declares mandate support for cards and returns the id.

No schema, migration or API change is needed. `network_txn_id` already exists on `PaymentsResponseData::TransactionResponse`, so this is connector-side wiring only.

### Environment

Identified by code inspection of `main` at commit `2c2e58976363eaa5ed0414c6fa2d355c048cb161`; all line references above are permalinks to that commit.

Utkal059:
## Type of Change

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

`bankofamerica` parses `processorInformation.networkTransactionId` into
`ClientProcessorInformation.network_transaction_id` but never reads it. Both places that build a
successful `TransactionResponse` hardcoded `network_txn_id: None`:

- the SetupMandate response, and
- `get_payment_response`, which the Authorize, Capture and Void responses share.

This PR adds a small `get_network_transaction_id` helper and uses it at both sites. That matches
what `cybersource` and `wellsfargo` already do with the same field. Bank of America keeps the id as
`Secret<String>`, so the helper exposes it at the boundary into `network_txn_id: Option<String>`.

The PSync paths still return `None`, as they do in `cybersource` and `wellsfargo`. This change is
limited to the reported defect.

### Additional Changes

## Motivation and Context

Closes #14193.

Because `network_txn_id` was always `None`, `tokenization.rs` never stored a
`network_transaction_id` on cards saved off-session through Bank of America. `get_mca_status` only
treats a saved card as usable for connector-agnostic MIT when that id is present, so these cards
could never be charged through another connector with a network-transaction-id MIT, even though
Bank of America supports card mandates and returns the id.

## How did you test it?

Unit tests added in `crates/hyperswitch_connectors/src/connectors/bankofamerica/transformers.rs`:

- `payment_response_forwards_network_transaction_id` covers the regression: a response that
  carries `processorInformation.networkTransactionId` now yields that value as `network_txn_id`.
- `payment_response_without_network_transaction_id_is_none` checks that a response without the
  field still parses and yields `None`, so no id is invented.

The helper separates "response did not parse" from "parsed with no id", so neither test can pass
because of a deserialization failure.

Locally verified with:

```
cargo +nightly fmt --all --check
cargo clippy -p hyperswitch_connectors --all-targets --features v1,payouts,frm,revenue_recovery
cargo test -p hyperswitch_connectors --features v1,payouts,frm,revenue_recovery --lib -- bankofamerica::transformers::tests
```

No Bank of America sandbox credentials were available, so this was verified at the transformer
level rather than against a live authorization.

## Checklist


XyneSpaces:
[should-fix] Please forward `processor_information.network_transaction_id` from the payment-sync `BankOfAmericaTransactionResponse` paths as well. Both sync `TransactionResponse` constructors still set `network_txn_id: None`, so a sync response containing an NTI drops it and downstream MIT flows cannot reuse it; use `get_network_transaction_id(&item.response.processor_information)` in both branches and add a sync-path test.

Issue #14034, opened 2026-09-03
ci(cypress): follow-up perf work — retries/timeout tuning, Diesel CLI install, dedupe job setup


HS-QUALIBOT:
## Context

Follow-up to #11018, which added spec-level timing diagnostics (`specTimings.js`) and made a first pass at CI perf improvements (npm caching, faster server-readiness polling, timing report surfaced in the GitHub Actions step summary).

A few more optimization ideas came up during that work that need more validation before landing, since they change test behavior rather than being pure CI-time savings:

## Follow-up items

- [ ] **Retries tuning** — `cypress.config.js` sets `retries: 2` for non-mock runs. Each retry re-runs the full test including all `before`/`beforeEach` hooks and API calls. Worth auditing whether retries are masking genuine flakiness that should be fixed at the source instead.
- [ ] **`taskTimeout` tuning** — currently `120000` (2 min) for all custom tasks (`readFileOrNull`, `cli_log`, `computeHmac`, `setGlobalState`/`getGlobalState`). These are simple/fast operations; a lower default with per-task overrides would fail fast on a stuck task instead of hanging.
- [ ] **Diesel CLI install speedup** — `cargo-install` for `diesel_cli` runs from source in every Cypress job (~8 jobs), which is slow. Consider a pinned prebuilt binary release, or bundling the binary into the `hyperswitch-server` build artifact so it's built once.
- [ ] **Dedupe repeated job setup** — the Rust toolchain / sccache / Diesel CLI / Just / Node.js install steps are duplicated near-identically across ~8 jobs in `.github/workflows/cypress-tests-runner.yml`. Pulling this into a reusable composite action would reduce maintenance surface (though since jobs run in parallel, this is more about job-count/maintainability than wall-clock savings).

## Related

- #11018 — landed: npm caching, 2s readiness polling (down from 10s), timing report in `$GITHUB_STEP_SUMMARY`, video recording disabled for Cypress runs.

PR #14196, opened 2026-09-13
ci: restore Rust build caching for check, clippy and MSRV jobs

### LINKED ISSUE #14195: ci: Rust jobs rebuild from scratch — no build cache has been saved since 2026-09-09
## Summary

`cargo check`, `clippy` and the MSRV job rebuild the full dependency graph on every
run. Caching is configured but inert. **No Rust cache has been created in this
repository since 2026-09-09.**

## Evidence

Every surviving `v0-rust-*` cache predates that date, and all four are `check-v2`:

```
created:2026-09-08  1.09GiB  v0-rust-check-v2-Linux-x64-d8a1f82f-aa5131c1  refs/heads/main
created:2026-09-09  1.01GiB  v0-rust-check-v2-Linux-x64-5daddb8d-aa5131c1  refs/heads/main
created:2026-09-09  1.01GiB  v0-rust-check-v2-Linux-x64-5daddb8d-aa5131c1  refs/heads/gh-readonly-queue/...
created:2026-09-09  1.01GiB  v0-rust-check-v2-Linux-x64-5daddb8d-aa5131c1  refs/heads/gh-readonly-queue/...
```

There is no cache for `check-msrv` or `test` at all.

## Causes

**1. rust-cache was swapped for an inert sccache.** `d8f9262a96` moved `test` and
`check-v2` to self-hosted runners, removed `Swatinem/rust-cache` and enabled
`RUSTC_WRAPPER: sccache`. No sccache backend is configured, so it falls back to
local disk — but the runners are ephemeral ARC pods, a distinct one per job:

```
run 34773927305  hyperswitch-runners-dz97x-runner-zcd2z  <- test
run 34768674075  hyperswitch-runners-dz97x-runner-kz8kz  <- test
run 34766241446  hyperswitch-runners-dz97x-runner-nncnc  <- test
```

The cache directory is destroyed with the pod, so sccache costs time and returns
nothing.

**2. `check-msrv` keeps rust-cache but never saves.** The action is
`post-if: success()` and the job fails on `Cargo hack`, so the save step is skipped:

```
JOB: Check compilation on MSRV toolchain   [25 min]
   success   Install rust cache
   failure   Cargo hack
   skipped   Post Install rust cache
```

**3. PR jobs are restore-only** (`save-if: false`) and depend on main to populate the
cache, so 1 and 2 together leave them nothing to restore.

## Why sccache could never work here, and rust-cache can

Both are build caches, but they differ in **where the cached bytes live between jobs**.
On ephemeral runners that difference is the whole story.

| | sccache (as configured) | `Swatinem/rust-cache` |
|---|---|---|
| Unit of caching | individual `rustc` invocations, keyed by compiler, flags and input hashes | `~/.cargo` registry/git/bin plus `target/` dependency artifacts, as one archive |
| Where bytes are stored | **local disk on the runner** (`~/.cache/sccache`) — the default when no backend is configured | GitHub's **Actions cache service**, over HTTPS |
| Survives pod destruction | **No** | **Yes** |
| Granularity | per compilation unit; can hit across branches, feature flags and target dirs | per toolchain + lockfile key; coarse, effectively all-or-nothing |

sccache's storage is pluggable — S3, GCS, Redis, memcached, Azure, or the GitHub Actions
cache. **No backend was configured**, so it fell back to local disk.

That is fatal on ARC runners, because every job gets a brand-new pod:

```
run 34773927305  hyperswitch-runners-dz97x-runner-zcd2z  <- test
run 34768674075  hyperswitch-runners-dz97x-runner-kz8kz  <- test
run 34766241446  hyperswitch-runners-dz97x-runner-nncnc  <- test
run 34780717792  hyperswitch-runners-dz97x-runner-5sz9x  <- test
```

The pod's filesystem, `~/.cache/sccache` included, is created when the job starts and
destroyed when it ends. So sccache began every job with an empty cache and finished by
discarding everything it had just written. This is not a low hit rate to be tuned: with
local storage on an ephemeral runner the hit rate is **structurally zero** and cannot be
anything else.

It was not free either. Every job paid an `Install sccache` step plus a wrapper process
around each `rustc` invocation, in exchange for a cache that could never hit.

`rust-cache` is unaffected because it never relies on the pod's disk — it uploads to and
downloads from GitHub's cache service over the network, so pod lifetime is irrelevant.
Confirmed on this repository's own self-hosted runners in #14196: `Install rust cache`
succeeded on `hyperswitch-runners-dz97x-runner-92z95` and `-5sz9x`, and a 1041.8 MiB
cache was uploaded in 23 s.

### Could sccache be made to work?

Yes — by giving it a shared backend (`SCCACHE_GHA_ENABLED`, or S3/Redis). Its
per-compilation-unit granularity is genuinely more capable than rust-cache's
all-or-nothing archive: it can still hit when feature flags change, or across branches
where the lockfile has moved.

It was not chosen here for two reasons:

1. The GitHub Actions cache backend draws on the same 10 GB budget described below, and
   sccache writes many entries into it. It would compete with the caches it is meant to
   complement.
2. An S3 backend requires repository secrets, and GitHub never exposes secrets to pull
   requests from forks. External contributors would silently get no caching at all, with
   no signal that anything was wrong.

Worth revisiting once the budget is healthy — ideally with a cluster-local backend, so
self-hosted runners stop paying a round trip to GitHub for every restore.

## Contributing factor: the cache budget is full

9.56 GiB of the 10 GB cap across 411 entries, LRU-evicted:

| Key prefix | Count | Size |
|---|---|---|
| `codeql-overlay-base` | 73 | 4.20 GiB |
| `v0-rust-check` | 4 | 4.13 GiB |
| `cargo-install-diesel_cli` | 344 | 0.95 GiB |
| `codeql-trap-1` | 5 | 0.28 GiB |

The 344 `diesel_cli` entries come from only 2 distinct keys, duplicated per branch.
2.02 GiB of the Rust caches sit on disposable `gh-readonly-queue/*` refs that are
deleted after the merge queue completes.

## Out of scope here

CodeQL's 4.20 GiB is the single largest consumer, but it runs via GitHub's *default
setup* — there is no workflow file, so it needs a repository admin rather than a code
change. Worth tracking separately.

## Baselines

`test` 47 min, `check-msrv` 25 min.

hrithikesh026:
Closes #14195
Ticks the "Diesel CLI install speedup" item in #14034.

## Problem

Rust CI jobs rebuild the full dependency graph every run. Caching is configured but
inert — **no Rust cache has been created in this repository since 2026-09-09**. Full
evidence in #14195; in short, three independent causes:

1. `d8f9262a96` swapped `Swatinem/rust-cache` for `RUSTC_WRAPPER: sccache` on `test`
   and `check-v2` when moving them to self-hosted runners. No sccache backend is
   configured, so it caches to local disk — but those runners are ephemeral ARC pods
   (a distinct one per job), so the cache dies with the pod.
2. `check-msrv` kept rust-cache but never saves: the action is `post-if: success()`
   and the job fails on `Cargo hack`, so `Post Install rust cache` is **skipped**.
3. PR jobs are restore-only and depend on main to populate, so 1 and 2 leave them
   nothing to restore.

## Changes

- **`cache-on-failure: true`** on CI-push, so main saves despite the failing
  `Cargo hack` step. This is load-bearing: without it every other change here is inert.
- **Restore `Swatinem/rust-cache`** to `test` and `check-v2` in both workflows. It is
  network-backed, so runner ephemerality does not matter.
- **Remove `RUSTC_WRAPPER: sccache`** and its install step from the four self-hosted
  jobs, where it costs time and caches nothing.
- **Explicit `save-if` on `check-v2`**, which previously had no `with:` block and so
  defaulted to saving on `merge_group` too — 2.02 GiB is currently stranded on
  disposable `gh-readonly-queue/*` refs.
- **Matching `shared-key`** across both workflows (`msrv` / `stable-v1` /
  `stable-v2`) so saver and restorer stay paired if a job is renamed.
- **`test`'s `CARGO_BUILD_JOBS` moved from job-level to step-level env** in
  `CI-push.yml`. `Swatinem/rust-cache` hashes any env var starting with
  `CARGO`/`CC`/`CFLAGS`/`CXX`/`CMAKE`/`RUST` into the cache key, and it reads the
  environment as of its own step — which runs *before* any step-scoped env in either
  workflow. `CI-push.yml` had `CARGO_BUILD_JOBS: 2` at job level (visible to the cache
  step), while `CI-pr.yml` only sets it (a different value, 3) inside its clippy steps
  (after the cache step). Different env-hash inputs → different key → PRs could never
  restore what `main` had just written for the `test` job, even with a matching
  `shared-key` and identical `Cargo.lock`. Moved to step-level on just the four
  cargo-invoking steps in `CI-push.yml`, mirroring `CI-pr.yml`'s existing pattern —
  same runtime parallelism everywhere, just invisible to the cache-key step now.
- **diesel_cli from a prebuilt binary** instead of `cargo-install` from source, across
  all 8 install sites in the repo. See **Why `taiki-e/install-action`** below for the
  reasoning; the first pass covered 6 of them (5 in `cypress-tests-runner.yml`, 1 in
  `postman-collection-runner.yml`) and missed two — `migration-check.yaml` and the
  inert `archive/connector-ui-sanity-tests.yml` — closed in a follow-up commit on this
  branch. No `baptiste0928/cargo-install` reference remains anywhere in the repo.

The rust-cache step sits before the tool installs because it restores
`$CARGO_HOME/bin` and would otherwise clobber `cargo-hack`/`just`.

### Why `taiki-e/install-action` instead of `baptiste0928/cargo-install`

`baptiste0928/cargo-install` compiles the crate from source (`cargo install diesel_cli
--no-default-features --features postgres`) and caches the resulting binary via
`actions/cache`, keyed as `cargo-install-<crate>-<version>-<sha256(os+arch+osVersion+args)[:24]>`
— e.g. `cargo-install-diesel_cli-2.3.13-f53f772646f363701f3f99bd`. That key text never
varies for a given crate/version/args/runner, but **cache storage is scoped per ref**
(a workflow run can only restore a cache from its own branch, its base branch, or the
default branch — never a sibling PR or an unrelated ref). Since the key is identical
everywhere but the storage is isolated per PR, every open PR and every merge-queue
attempt independently misses, recompiles, and writes its own byte-identical copy —
the "344 entries from only 2 distinct keys" diagnosed in #14195.

This isn't fixable by config on that action: `restoreCache`/`saveCache` are two halves
of one atomic step (miss → compile → save, unconditionally) — there's no
`save-if`-equivalent input to make it read-mostly, unlike `Swatinem/rust-cache`'s
separate main/post lifecycle. And `migration-check.yaml` specifically has no
`push`-to-`main` trigger at all (only `pull_request`/`merge_group`), so there was
never even a code path that could seed a default-branch-scoped entry for other refs to
restore from — every single run was guaranteed to rebuild from source.

`taiki-e/install-action` sidesteps the whole problem: it downloads diesel's own
published prebuilt binary from GitHub Releases and never touches `actions/cache` at
all (confirmed against its source — no `restoreCache`/`saveCache` calls). Nothing to
compile, nothing to write, nothing to duplicate per branch.

## Verification

`actionlint` reports **55 findings before and after, identical sets** — nothing
introduced. All are pre-existing shellcheck info and self-hosted runner-label warnings.

The prebuilt diesel_cli binary was inspected directly rather than assumed — downloaded
the actual `diesel_cli-aarch64-unknown-linux-gnu.tar.xz` release asset (v2.3.13, the
same one `taiki-e/install-action` installs) and checked its symbol table and dynamic
dependencies:

```
postgres:  PQconnectdb, PQexec, PQconnectPoll, PQerrorMessage, ...  (present)
mysql:     mysql_close, mysql_client_plugin_init, ...               (present)
sqlite:    sqlite3_bind_blob, sqlite3_backup_step, ...              (present)
dynamic deps: libc.so.6, libgcc_s.so.1, libm.so.6, libstdc++.so.6   (no libpq/libmysqlclient/libsqlite3)
```

Postgres support is confirmed present and statically linked — no external database
client is needed at runtime. Note this binary is actually a superset of the old
`--no-default-features --features postgres` build: the published release bundles
**all three backends** (postgres, mysql, sqlite) in one binary, all statically linked.
Functionally a strict superset for this repo's postgres-only usage, just a larger
download with unused backend code along for the ride — not a correctness concern.

The temporary cache-saving commit has been **reverted** (8ecaf3ecf1); `save-if: false`
is back in force on all three PR jobs, so this PR no longer writes caches. Measured
results below.

## Measured result

Verified on this PR rather than asserted. The temporary commit (since reverted in
8ecaf3ecf1) let this branch populate a cache on one run, so the next run could restore
it — demonstrating the whole path without waiting for a merge to `main`.

- **Run 1, cold** — [`34780717792`](https://github.com/juspay/hyperswitch/actions/runs/34780717792) — nothing to restore; saved a cache at the end.
- **Run 2, warm** — [`34783920786`](https://github.com/juspay/hyperswitch/actions/runs/34783920786) — restored it.

| Job | Run 1 (cold) | Run 2 (warm) | Saved | |
|---|---|---|---|---|
| [`check-msrv`](https://github.com/juspay/hyperswitch/actions/runs/34783920786/job/103795820710) (GitHub-hosted, x64) | [13 m 57 s](https://github.com/juspay/hyperswitch/actions/runs/34780717792/job/103787114548) | **6 m 57 s** | 7 m 00 s | **−50%** |
| [`test`](https://github.com/juspay/hyperswitch/actions/runs/34783920786/job/103795820726) (self-hosted, arm64) | [51 m 00 s](https://github.com/juspay/hyperswitch/actions/runs/34780717792/job/103787114565) | **34 m 10 s** | 16 m 50 s | **−33%** |
| [`check-v2`](https://github.com/juspay/hyperswitch/actions/runs/34783920786/job/103795820701) (self-hosted, arm64) | [59 m 15 s](https://github.com/juspay/hyperswitch/actions/runs/34780717792/job/103787114543) | **38 m 34 s** | 20 m 41 s | **−35%** |

The three run in parallel, so the figure a contributor actually experiences is the
longest job: **PR feedback time drops from 59 min to 39 min.** Total compute saved is
44 m 31 s per run.

The clearest single step is `check-msrv`:

```
Run 1:  Install rust cache     3 s  (miss)    Run cargo check  13 m 05 s
Run 2:  Install rust cache    17 s  (hit)     Run cargo check   6 m 16 s   (-52%)
```

### Caches produced

```
1041.8 MiB  v0-rust-msrv-Linux-x64-7d0fad79-916e7071
1635.7 MiB  v0-rust-stable-v1-Linux-arm64-e73c3637-916e7071
2248.4 MiB  v0-rust-stable-v2-Linux-arm64-642ce0d1-916e7071
```

Note `Linux-arm64` versus `Linux-x64`: the self-hosted runners are **arm64** while
`ubuntu-latest` is x64. This is an additional, independent reason nothing was being
restored — the surviving `v0-rust-check-v2-Linux-x64-*` entries were built before
`d8f9262a96` moved that job to arm64 runners, so they could never have matched it again
even if the cache step had still been present.

### Restore cost, hosted vs self-hosted

| | Cache | Restore | Throughput |
|---|---|---|---|
| `check-msrv` (GitHub-hosted) | 1041.8 MiB | 17 s | **64 MB/s** |
| `test` (self-hosted) | 1635.7 MiB | 3 m 33 s | **8.1 MB/s** |
| `check-v2` (self-hosted) | 2248.4 MiB | 5 m 04 s | **7.8 MB/s** |

The self-hosted runners reach the Actions cache service about **8× slower** than
GitHub-hosted ones (upload measured the same, ~7×). Restore therefore consumes 10–13%
of each self-hosted job, against 0.4% on the hosted runner — roughly 7½ minutes across
the two. Still strongly net positive, but it is the obvious next target: a
cluster-local cache backend would recover most of it.

### Where the remaining time goes

After dependencies are restored, the bulk of the self-hosted jobs is clippy over the
workspace crates (10 m 05 s + 10 m 14 s in `test`; 8 m 41 s + 8 m 47 s in `check-v2`).
`Swatinem/rust-cache` does not cache workspace crates by default, so this is the
portion it structurally cannot reach.

A separate experiment layering `sccache` (GHA cache backend) on top, scoped to
`check-msrv`, is tracked in #14197: a modest ~7% warm-run gain came bundled with cache
write-errors and thrash symptoms, and `test`/`check-v2` turned out not to be viable at
all — sccache cannot cache `cargo clippy` on workspace crates due to a still-open
upstream limitation ([mozilla/sccache#539](https://github.com/mozilla/sccache/issues/539)).
Not part of this PR.

## Not fixed by this PR

- **The jobs are still red.** `check-msrv` and `test` fail on `Cargo hack` on main for
  unrelated reasons; `cache-on-failure` makes caching survive that, it does not fix it.
- **CodeQL's 4.20 GiB** (73 overlay databases) is now the largest cache consumer, but it
  runs via GitHub *default setup* with no workflow file, so it needs a repo admin.

Baselines to compare against: `test` 47 min, `check-msrv` 25 min.


XyneSpaces:
[should-fix] Align the job-level `CARGO_BUILD_JOBS` environment between the push and PR `stable-v1` jobs, or remove it before `Swatinem/rust-cache` computes the key. The push job includes `CARGO_BUILD_JOBS=2` in the cache key while the PR job only sets it at command-step scope, so the PR cannot restore the cache this change is intended to share.

.github/workflows/CI-pr.yml
```diff
@@ -177,11 +176,11 @@ jobs:
         with:
           repo-token: ${{ secrets.GITHUB_TOKEN }}
 
-      - name: Install sccache
-        uses: taiki-e/install-action@v2
+      - name: Install rust cache
+        uses: Swatinem/rust-cache@v2.9.1
```
pixincreate, line 180: why was this reverted? i had intentionally moved it to using sccache since we're using hs-runners in https://github.com/juspay/hyperswitch/pull/14081

my intent was to use sccache for compiler artifacts rather than relying on the GitHub Actions target-directory cache.

cc: @SanchithHegde

.github/workflows/CI-pr.yml
```diff
@@ -278,11 +273,11 @@ jobs:
...
```
pixincreate, line 277: same here.

.github/workflows/CI-push.yml
```diff
@@ -174,11 +175,12 @@ jobs:
...
-      #   uses: taiki-e/install-action@v2
-      #   with:
-      #     tool: sccache
-      #     checksum: true
+      - name: Install rust cache
+        uses: Swatinem/rust-cache@v2.9.1
+        with:
+          shared-key: stable-v1
+          save-if: ${{ github.event_name == 'push' }}
+          cache-on-failure: true
```
pixincreate, line 183: same here.

.github/workflows/CI-push.yml
```diff
@@ -286,11 +286,12 @@ jobs:
         with:
           repo-token: ${{ secrets.GITHUB_TOKEN }}
 
-      - name: Install sccache
-        uses: taiki-e/install-action@v2
+      - name: Install rust cache
+        uses: Swatinem/rust-cache@v2.9.1
         with:
```
pixincreate, line 291: same here

.github/workflows/CI-pr.yml
hrithikesh026: How was this pr verified?? https://github.com/juspay/hyperswitch/pull/14081
Did you find any reduction in time taken by CI jobs???

hrithikesh026: sccache needs to save the artifacts cache at the end of the job to some storage, So that it can be restored in the next run. By default it is the local storage of the runner. Which will be destroyed at the end of the run(which is what was happening so far). 
It can be configured to be stored in github actions cache or S3. Which was also missing. So caching was not at all happening. 

Also sccache only works with cargo build, and check. There is no caching support for clippy yet. 

Rust cache supports caching all artifacts. And the cached artifacts can be restored in self-hosted runners as well.

hrithikesh026:
this PR was created for experimentation.

PR #14202, opened 2026-09-14
feat(deja): move the pin to deja main and port the seams onto the miss arm

maverox:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [x] Refactoring
- [x] Dependency updates

## Description

**Stacked on #14201.** That PR moves the deja pin to the addressing fix and
nothing else, so it can land on its own timeline. This one carries everything
needed to exercise the remainder of the work against a real recording, and is
expected to be tested before it is merged rather than the other way round.

The pin moves from `ade8fd16c1` (deja #147) to `d587865` (deja main), across all
nine declarations and `Cargo.lock` together. All nine move by necessity: two
revisions of deja in one build means two copies of the recording statics, and a
recorder writing into the copy nobody reads fails silently rather than erroring.
The check that matters is that the old revision has zero occurrences afterwards,
lock included, not that the build still succeeds.

That brings in three deja changes: on-miss synthesis, the `deja::synth`
deterministic-value helpers, and the hash-seed primitive.

### What actually changed in the seams

A boundary used to **declare in advance** what a miss would do — a `MissPolicy`
beside a separate `dispatch_async_or_miss` seam and a thunk. It now **returns
what it did**: one `reconstruct` closure answers both a hit and a miss, and the
observation records the outcome rather than the declaration. The two used to be
the same thing, because a declared miss value always produced one. They came
apart once a miss arm could decline.

Three call sites move onto that shape.

**Superposition** returns `Synthesized(Err(NotFound))` on a miss — the error the
caller already knows how to survive, so the DB-to-default fallback runs and the
correlation continues. Previously `MissPolicy::Absorb`. The message is now built
from the miss itself rather than from a captured variable, which is the property
the closure signature exists to enforce: the value must be a function of the
query alone, or two replays of one candidate disagree with each other and "the
candidate changed" cannot be separated from "the fabrication changed".

**The gRPC transport** keeps its two arms distinct, which is the part that would
have been easy to lose. With a transport present a miss may be a genuine novel
call, so that arm returns `NoValue` and keeps the default substitute-miss
fail-stop. With the transport deliberately absent there is nothing to run and a
miss establishes nothing about the candidate, so that arm still calls
`fail_stop_absent_executor` and keeps its own reason. Collapsing both into
`NoValue` would have compiled, and would have silently replaced a specific
operator-facing message with a generic one.

**`router_env`'s id-generation lookup** simply loses the declared field.
Behaviour is unchanged: no miss value is offered, so the caller falls through to
minting a fresh uuid — precisely a value the recording never held.

### Additional Changes

## Motivation and Context

#14201 fixes replay addressing and is worth landing on its own. Everything
after it — answering a miss deterministically instead of fail-stopping, the
helpers that make a synthesized value non-colliding with a recorded one, and
recording a collection's hash seed — needs a build to be tested at all. This is
that build.

## How did you test it?

By compiling rather than by reading: `external_services` and `router_env` at
`--features v1,deja,superposition`, and `router` at `--features v1,deja,release`.
All clean.

**The first run of that check was worthless, and it is worth recording why.**
`superposition.rs` sits behind `#[cfg(feature = "superposition")]`, so a check
run without that feature never compiled the file being ported and reported
success anyway. What caught it was inserting a `compile_error!` into the file and
confirming the check failed: with the feature absent, the probe did not fire.
A check that cannot fail is not evidence, and the feature set is the thing that
decides whether this one can.

Five fixture branches are cut from this branch's tip, each injecting exactly one
class of cross-build divergence with a predicted replay outcome, so the replay
classifier can be checked rather than trusted: `fixture/novel-id-call`,
`fixture/args-perturbed`, `fixture/new-seam-site`, `fixture/value-diverged`,
`fixture/omitted-call`. None of them may ever merge; every hunk says so.

## Checklist


maverox:
Superseded by #14311, which is the same work cut fresh against `main`.

Closing rather than rebasing. This branch sits 309 files behind, and every rebase conflict turned out to be one comment in two phrasings — the shortened versions had already landed via #14201, so resolving to `main` made a *later* commit on this branch conflict with its own effect. That is the signal that the history no longer reflects what is on main.

One correction worth leaving on the record, because it was stated loosely elsewhere: the pin here (`d587865`) was **not** wrong or on a divergent lineage. `ade8fd16c1` is an ancestor of it, so this PR would have moved the build six commits forward on the same line. It is simply 41 commits stale now — deja main has moved a long way since.

Everything of substance here is carried into #14311: the pin, and all four seam ports.

PR #14204, opened 2026-09-14
feat(routing): move dynamic routing decisions to the Decision Engine

AnkitKmrGupta:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

Hyperswitch decided *what to ask the Decision Engine* by reading its own `business_profile` row first, so a profile whose routing lives on the engine could still be answered from stale local config. This PR removes those reads from the payment path and leaves the decisions with the engine.

**1. The dynamic half of `/routing/hybrid` is built on the config flag alone**

`HybridRoutingStage::build_dynamic_routing_request` used to build the dynamic sub-request only when `profile.dynamic_routing_algorithm` was present, parsed into a `DynamicRoutingAlgorithmRef`, had success-rate or elimination enabled, **and** a locally drawn volume split landed on `Dynamic`. The static sub-request beside it was already profile-free, so the two halves of the same request were built on completely different conditions.

It is now built whenever `open_router.dynamic_routing_enabled` is on. Whether dynamic routing actually applies is the engine's per-merchant `sr_routing_enabled` flag — it answers `dynamic_routing: { status: "skipped" }` when off — and the volume split and ranking are the engine's as well.

**2. `elimination_enabled` is removed from `OpenRouterDecideGatewayRequest`**

The engine only ever ORs this flag with its own per-merchant check (`payment_flow_enabled || elimination_enabled == Some(true)`), so omitting the field means "the engine decides". That is the same contract `enableMultiObjective` already has — a field Hyperswitch has never sent, whose absence the engine resolves from the merchant's own feature flag.

**3. Effective cut-over accepts either capability flag**

`is_decision_engine_routing_effective` now requires `(static_routing_enabled || dynamic_routing_enabled)` rather than `static_routing_enabled` alone, so a profile the engine routes dynamically is cut over even with static routing off.

**4. The gateway-score report drops the same profile read**

It previously fired only when `profile.dynamic_routing_algorithm` was present *and* the attempt's `routing_approach` was one of the two success-rate variants — which silently excluded every elimination-routed attempt, since the engine's elimination approach maps to `DefaultFallback`.

It now fires for every terminal attempt on a cut-over profile, whether the engine answered statically or dynamically, so the engine accumulates outcome history before dynamic routing is ever switched on and decides for itself which window an outcome belongs to. The cut-over check is the only gate, and it is resolved inside the existing detached task — off the response path — using the `dimensions` that `PostUpdateTracker::update_tracker` already carried and every implementation ignored.

Expect a higher volume of `update-gateway-score` calls per cut-over profile than before.

### Additional Changes

`elimination_enabled` (`eliminationEnabled` on the wire) is removed from `OpenRouterDecideGatewayRequest`, which is the request body of `POST /routing/evaluate`, and `api-reference/v1/openapi_spec_v1.json` is regenerated accordingly. The field was optional and never in `required`, and the struct has no `deny_unknown_fields`, so a caller still sending it gets a normal `200` — the value is ignored rather than forwarded. Callers who need to force elimination should call the Decision Engine's `/decide-gateway` directly, which still accepts it.

No new or renamed configuration keys. Note though that the *meaning* of the existing `open_router.static_routing_enabled` and `open_router.dynamic_routing_enabled` flags changes, per point 3 above.

## Motivation and Context

The Decision Engine is being made the modular owner of routing. As long as Hyperswitch interprets `business_profile.dynamic_routing_algorithm` to decide whether to send a dynamic request — or whether to report an outcome — a cut-over profile is still half-routed by Hyperswitch's reading of config the engine is supposed to own. Under DE-only writes that local config is frozen at the moment of cut-over, so it can only get staler.

Concretely, before this PR a cut-over profile with no `dynamic_routing_algorithm` row got **no** dynamic request and **no** score feedback at all, no matter what the engine was configured to do for it.

## How did you test it?

Manually, end to end, by running the router against a stubbed Decision Engine and driving real payments through `POST /payments` with the dummy connector. No automated tests were added — the change is about which outbound requests are produced, which the existing suites do not assert on.

**Setup**

- Router built with `v1,dynamic_routing,olap,oltp,redis-rs,dummy_connector,...` and run against local Postgres/Redis.
- `open_router.url` pointed at a stub on `127.0.0.1:8099` that logs every request body and answers `/routing/hybrid` with `{}`, so Hyperswitch falls back to its own static result and the payment completes — leaving the **outgoing request** as the thing under test.
- Test profile chosen so the old code would have skipped every path: **`routing_algorithm` and `dynamic_routing_algorithm` both NULL**, 3 active MCAs, cut over via `routing.routing_result_source = decision_engine`.
- Card `4242424242424242` → `succeeded`; `4000000000000002` → `failed`. Both are terminal.

**1. Dynamic request is sent although the profile has no dynamic config, and carries no `eliminationEnabled`**

```jsonc
// POST /routing/hybrid  — keys: ['dynamic_routing_request', 'static_routing_request']
{
  "paymentInfo": { "paymentId": "pay_z2A7ly2U5faBPvluzeEi", "amount": 6540, "currency": "USD", ... },
  "merchantId": "pro_Sw1sLkdR6utFknZCsABP",
  "eligibleGatewayList": [
    "paypal_test:mca_GzYYYnmdTH9ewpkoh5xb",
    "paypal_test:mca_uXJ9KmIkJ6VARhmSMefW",
    "paypal_test:mca_dzb7fQrCNWaEmCHw2KzH"
  ],
  "rankingAlgorithm": "SR_BASED_ROUTING"
}
```

No `eliminationEnabled` key, and `eligibleGatewayList` is the Hyperswitch static output as intended. Profile columns confirmed still NULL afterwards.

**2. Score update fires on an approach the old gate rejected**

The attempt recorded `routing_approach = default_fallback` (the engine returned nothing, so Hyperswitch's own fallback served the payment), and the report still went out:

```json
{"merchantId":"pro_Sw1sLkdR6utFknZCsABP","gateway":"paypal_test:mca_GzYYYnmdTH9ewpkoh5xb",
 "status":"CHARGED","paymentId":"pay_z2A7ly2U5faBPvluzeEi"}
```

Both old gates would have dropped this: the profile's `dynamic_routing_algorithm` is NULL, and `default_fallback` is not a success-rate approach.

**3. Flag permutations, all on the cut-over profile**

| `static_routing_enabled` | `dynamic_routing_enabled` | `/routing/hybrid` | static half | dynamic half | `/update-gateway-score` |
|---|---|---|---|---|---|
| ON | ON | called | ✓ | ✓ | CHARGED |
| ON | OFF | called | ✓ | — | CHARGED |
| OFF | ON | called | — | ✓ | CHARGED |

The third row is only reachable because of the widened cut-over predicate; previously the engine could not be called with static routing off.

**Build checks:** `cargo clippy --all-targets` clean with `v1,dynamic_routing,...` and again with `dynamic_routing` off (the `#[cfg(not(...))]` stub and the cfg-gated `dimensions` parameter), `just check_v2` passing, `cargo fmt --check` clean.

## Checklist


jagan-jaya (dismissed):

crates/router/src/core/payments/routing.rs
```diff
@@ -1679,86 +1679,27 @@ impl HybridRoutingStage {
...
-                        );
-                        (None, Some(dynamic_routing_volume_split))
-                    } else {
-                        match perform_dynamic_routing_volume_split(
-                            vec![
-                                api_models::routing::RoutingVolumeSplit {
-                                    routing_type: api_models::routing::RoutingType::Dynamic,
-                                    split: dynamic_routing_volume_split,
-                                },
-                                api_models::routing::RoutingVolumeSplit {
```
prajjwalkumar17, line 1715: @jagan-jaya are we intented to remove the volume split between the static and dynamic routing as well, this was added in order to split the payments in-between rules and SR, I think this can be taken care by AB in DE?

crates/router/src/core/payments/routing/utils.rs
```diff
@@ -2673,13 +2673,16 @@ pub async fn get_routing_result_source(
...
+/// global routing flags is on. The flags always win, for APIs and payment paths alike. Either
+/// capability is enough, so a profile routed dynamically by the engine is cut over even with
+/// static routing off, and vice versa.
 pub async fn is_decision_engine_routing_effective(
     state: &SessionState,
     dimensions: &dimension_state::DimensionsWithProcessorAndProviderMerchantIdAndProfileId,
 ) -> bool {
-    state.conf.open_router.static_routing_enabled
+    (state.conf.open_router.static_routing_enabled
+        || state.conf.open_router.dynamic_routing_enabled)
```
prajjwalkumar17: use `is_decision_engine_routing_effective` here, instead of manually checking, refactor this everywhere manual checking is performed.

crates/router/src/core/payments/routing/utils.rs
```diff
@@ -2673,13 +2673,16 @@ pub async fn get_routing_result_source(
     }
 }
 
-/// Effective cutover: routing_result_source is DecisionEngine AND the global
-/// static_routing_enabled flag is on — the flag always wins, for APIs and payment paths alike.
+/// Effective cutover means routing_result_source is DecisionEngine and at least one of the
+/// global routing flags is on. The flags always win, for APIs and payment paths alike. Either
+/// capability is enough, so a profile routed dynamically by the engine is cut over even with
+/// static routing off, and vice versa.
```
prajjwalkumar17: one-liners are preferred everywhere.

crates/router/src/core/routing/helpers.rs
```diff
@@ -843,44 +843,37 @@ pub async fn update_gateway_score_helper_with_open_router(
...
-            &payment_attempt.merchant_id,
-            &payment_attempt.payment_id,
-            payment_attempt.status,
-        )
-        .await
-        .change_context(errors::ApiErrorResponse::InternalServerError)
-        .attach_printable("failed to update gateway score in open_router service")?;
-    }
+    logger::debug!(
+        "performing update-gateway-score for gateway with id {} in open_router for profile: {}",
```
prajjwalkumar17: add a decision_engine prefix for better visibility, during debug sessions.

crates/router/src/core/payments/routing.rs
AnkitKmrGupta: We are not calling SR directly form hyperswitch anywhere, so we dont't need volume split

prajjwalkumar17 (approved):

sai-harsha-vardhan (approved):

jagan-jaya (approved):

PR #14215, opened 2026-09-15
fix: install rustls CryptoProvider before GCP KMS client construction

itsharshvb:
## Type of Change

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

Consumer and Producer are both the `scheduler` binary, differentiated only by `SCHEDULER_FLOW` at runtime. Unlike `router.rs` (see #13834), its `main()` never installed a process-level rustls `CryptoProvider`, so `GcpKmsClient::new()` has no default to pick between the `ring` and `aws-lc-rs` backends linked into the binary when `gcp_kms` is enabled.

The same gap exists in the `drainer` binary — it's a separate crate/binary from `router`/`scheduler`, but its `AppState::new()` goes through the identical `secrets_management.get_secret_management_client()` → `fetch_raw_secrets()` → `GcpKmsClient::new()` path, and its own `release` feature also links both crypto backends. It never installed a default provider either.

This PR ports the same fix `router.rs` already has (#13834) to both:
- `crates/router/src/bin/scheduler.rs` — added the `install_default()` call to `main()`.
- `crates/drainer/src/main.rs` — added the same call to `main()`. This also required adding `rustls` as a direct dependency to `crates/drainer/Cargo.toml` (drainer previously only reached rustls transitively via `external_services`, and Rust requires a direct dependency to reference a crate's items directly), and exposing `gcp_kms` as drainer's own feature name rather than only forwarding it to `external_services/gcp_kms` — without that, the `#[cfg(feature = "gcp_kms")]` guard would silently never compile in, which `cargo check` confirmed via an `unexpected cfg condition value` warning before the fix.

### Additional Changes

## How did you test it?

- Built the actual `scheduler` binary locally with the same feature set Docker uses (`--features release,v1,redis-rs`) and ran it against a `secrets_management.secrets_manager = "gcp_kms"` config with `SCHEDULER_FLOW=producer`/`consumer`: reproduced the startup panic without the fix, confirmed it's gone and the process proceeds past KMS client construction (through Redis stream setup and consumer polling) with the fix applied.
- Built `drainer` locally with `--no-default-features --features release,v1,redis-rs` (`cargo check -p drainer`): confirmed the `unexpected cfg condition value: gcp_kms` warning present before the `Cargo.toml` feature fix, and gone after.

## Checklist


abhisheksharma2411 (commented):
Correct fix, and the right shape for it — this is byte-identical to the #13834 precedent in `router.rs`: same comment, same `#[cfg(feature = "gcp_kms")]` gate, same `#[allow(clippy::expect_used)]`, same position as the first statement in `main()`. Two binaries that differ only by `SCHEDULER_FLOW` should not differ here either.

I checked the reasoning rather than taking it on trust, and it holds: `scheduler.rs` is the only file under `crates/router/src/bin/` without an `install_default` call, and `router.rs` is the only one with it.

### One more binary looks like it has the same gap

`drainer` isn't covered by this PR, and from the feature graph it appears to need the same line:

```
crates/drainer/Cargo.toml:12
  release = ["vergen", "external_services/aws_kms", "external_services/gcp_kms"]
```

So a **release build of drainer links the GCP KMS path**, which is what pulls in tonic → `ring` alongside the AWS SDK's `aws-lc-rs` — the exact two-backend ambiguity this PR resolves. And `crates/drainer/src/main.rs` has no `install_default` (0 occurrences; I grepped every `main.rs` in the workspace).

The path is there too, not just the feature flag: `crates/drainer/src/settings.rs:5` imports `SecretsManagementConfig` from `external_services::managers::secrets_management`, and that module carries a `#[cfg(feature = "gcp_kms")]` `GcpKms` variant which constructs the client.

**Stated as a question rather than a finding**, because I traced the feature graph rather than reproducing a failure: does drainer actually reach `GcpKmsClient::new()` in a GCP-KMS-configured release deployment? It doesn't construct one directly — `grep GcpKmsClient crates/drainer/src/` is empty — so it depends on whether the secrets manager is instantiated there. If it is, drainer panics or misbehaves the same way the scheduler does today, and it'd be worth the same three lines in the same PR while the context is loaded.

If drainer never instantiates it, ignore this — and it might be worth a word in the commit message about *why* the two service binaries need it and the third doesn't, since the next person will ask.

### Small note on `.expect()`

`install_default()` returns `Err` when a provider is **already** installed, so this panics on a double install rather than no-opping. That's fine here — one call, first statement, one process — and it matches `router.rs`, so I wouldn't change it. Flagging only because if a future dependency starts installing a provider lazily, this becomes a startup panic rather than a silent win, and the failure message (`Failed to install default rustls CryptoProvider`) reads like "couldn't install" rather than "someone else already did". Not worth churn now.

(Reviewer note — not a maintainer, no approval implied.)

crates/drainer/Cargo.toml
```diff
@@ -27,6 +28,11 @@ diesel = { version = "2.2.10", features = ["postgres"] }
 error-stack = "0.4.1"
 mime = "0.3.17"
 reqwest = { version = "0.11.27" }
+rustls = { version = "0.23", default-features = false, features = [
```
SanchithHegde, line 31: Is `rustls` a dependency that's required always (within this crate), or can it be moved behind a feature flag?

SanchithHegde (commented):

SanchithHegde (approved):

sai-harsha-vardhan (approved):

PR #14222, opened 2026-09-15
refactor(ucs): move rollout proxy http/https URLs to comparison service env config

### LINKED ISSUE #14251: comparison_service: enabled flag is dead code; Shadow UCS silently falls back to Direct on missing/invalid proxy config
## Context

Follow-up from review discussion on #14222 (`refactor(ucs): move rollout proxy http/https URLs to comparison service env config`).

## Problems

1. **`ComparisonServiceConfig.enabled` is never read.** It's deserialized (`crates/hyperswitch_interfaces/src/types.rs`) but no code path checks it — `get_comparison_service_config()` (`crates/router/src/routes/app.rs:311-314`) just returns `Some(...)` whenever the `comparison_service` block is present, regardless of `enabled`'s value. Toggling `enabled = false` currently has no effect: diff-reporting to the comparison service and the Shadow UCS proxy override both stay active.

2. **Shadow UCS silently downgrades to `Direct` when the proxy override is missing/invalid.** In `crates/router/src/core/unified_connector_service.rs:1064-1090`, if `rollout_result.proxy_override` is `None` (e.g. `http_url`/`https_url` fail validation, see `crates/router/src/core/payments/helpers.rs` `validate_proxy_url`/`create_proxy_override`), the code falls back to `ExecutionPath::Direct` and logs only at `info!` level:
   > "No proxy override available for Shadow UCS; falling back to Direct, so no shadow comparison will run for this request"

   This means an operator can have a rollout config with `execution_mode: Shadow` fully enabled and get zero shadow traffic/comparison data, with no error and no alertable log line — only discoverable by reading `info` logs closely.

## Suggested fix

- Wire `enabled` into `get_comparison_service_config()` (or equivalent) so it actually gates both diff-reporting and the proxy-override path.
- Elevate the Shadow-UCS-fallback log to `warn!` (or emit a metric) so silent shadow-mode outages are observable in dashboards/alerts, not just debug-level log archaeology.

## References
- PR: #14222
- `crates/hyperswitch_interfaces/src/types.rs` (`ComparisonServiceConfig`)
- `crates/router/src/routes/app.rs:311-314` (`get_comparison_service_config`)
- `crates/router/src/core/unified_connector_service.rs:1064-1090`
- `crates/router/src/core/payments/helpers.rs` (`validate_proxy_url`, `create_proxy_override`, `build_rollout_proxy_override`)

AmitsinghTanwar007:
## Summary

Moves the UCS rollout proxy `http_url`/`https_url` fields off the DB-backed `RolloutConfig` JSON and onto the env-configured `ComparisonServiceConfig` (`crates/hyperswitch_interfaces/src/types.rs`), alongside the existing comparison service block.

`RolloutExecutionResult::from(RolloutConfig)` no longer builds the `ProxyOverride` from the rollout payload; it is now built from `SessionState.conf.comparison_service` at the call sites (`should_execute_based_on_rollout_with_precedence`) only when the rollout decides to execute.

No behavior change to rollout percentage/execution-mode/kill-switch handling — only the source of the proxy override URLs changes, from per-key DB config to static env config.

**Not backward compatible**: any existing `ucs_rollout_config_*` DB rows that set `http_url`/`https_url` will have those fields silently ignored after this change — the proxy override becomes a single, global, env-configured value instead of one configurable per precedence key. Set `ROUTER__COMPARISON_SERVICE__HTTP_URL`/`HTTPS_URL` before/at deploy to avoid Shadow UCS silently falling back to Direct.

## Test plan
- [x] `cargo check -p router -p hyperswitch_interfaces`
- [x] `cargo clippy -p router --no-deps`
- [ ] Manual verification that Shadow UCS proxy override still applies correctly when `ROUTER__COMPARISON_SERVICE__HTTP_URL` / `ROUTER__COMPARISON_SERVICE__HTTPS_URL` are set

## Follow-up

`http_url`/`https_url` on `ComparisonServiceConfig` are now required (previously `Option<String>`) per review feedback — a misconfigured `comparison_service` block now fails at config-load time instead of Shadow UCS silently falling back to Direct at request time.

Related, but out of scope for this PR: #14251 (dead `enabled` flag on `ComparisonServiceConfig`, and the Shadow-UCS-fallback log should be `warn!`-level, not `info!`).


JeevaRamu0104 (dismissed):

crates/hyperswitch_interfaces/src/types.rs
```diff
@@ -601,4 +601,8 @@ pub struct ComparisonServiceConfig {
     pub enabled: bool,
     /// Timeout in seconds for comparison service requests
     pub timeout_secs: Option<u64>,
+    /// Override HTTP proxy URL used for UCS rollout proxy switching
+    pub http_url: Option<String>,
+    /// Override HTTPS proxy URL used for UCS rollout proxy switching
+    pub https_url: Option<String>,
```
hrithikesh026: shouldn't this be required fields??

hrithikesh026 (approved):

shuklatushar226 (approved):

JeevaRamu0104 (approved):

PR #14210, opened 2026-09-15
feat(connector): [Shift4] enable card and wallet mandates with connector customer creation for UCS recurring

shuklatushar226:
## Type of Change

- [x] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

> The core `payment_confirm.rs` fix was briefly split into #14223 and is now folded back into this PR, so the three commits here are exactly the combination that was tested end to end. #14223 is closed.

This enables Shift4 recurring payments when Shift4 is routed through the Unified Connector Service (UCS): CIT, then a stored payment method, then MIT. It covers **cards** and **Apple Pay / Google Pay wallets**. The UCS side is juspay/hyperswitch-prism#2279.

Shift4 charges a stored card or a stored wallet payment method only when the charge carries the `customerId` that owns it. UCS Shift4 therefore:
- requires `connector_customer_id` for SetupMandate and for an off-session CIT Authorize, for both card and wallet;
- returns the connector mandate id:
  - cards: the Shift4 card id (`card_...`);
  - Apple Pay / Google Pay: the Shift4 payment method id (`pm_...`) created from the encrypted wallet token;
- on RepeatPayment:
  - cards: sends `card=<card_id>` plus `customerId`;
  - wallets: sends `paymentMethod=<pm_id>` plus `customerId`, with no `card`.

Hyperswitch never created a Shift4 connector customer, and it downgraded `off_session` to `on_session` for Shift4, so no Shift4 mandate could be created.

### Changes

1. **`crates/hyperswitch_connectors/src/connectors/shift4.rs`** (v1): override `should_call_connector_customer`. This follows the Mollie pattern.
   - It returns `CallConnectorCustomer` for off-session payments, including zero-auth mandates, when the payment method is:
     - `card`, or
     - `wallet` with `payment_method_type` `apple_pay` or `google_pay`.
   - Otherwise it returns `NoAction`. Other wallets are excluded because UCS Shift4 does not support them.
2. **Add `shift4` to the connector lists** in `[mandates.supported_payment_methods]` and `[zero_mandates.supported_payment_methods]`, for `card.credit`, `card.debit`, `wallet.apple_pay` and `wallet.google_pay`. These lists do **not** decide whether a UCS mandate works:
[...]

PR #14223, opened 2026-09-15
fix(core): take setup_future_usage from the intent for attempts confirmed without one

shuklatushar226:
## Type of Change

- [x] Bugfix

## Description

An attempt created with `confirm=false` and no `setup_future_usage`, then confirmed with `setup_future_usage=off_session` and `customer_acceptance`, kept `payment_attempt.setup_future_usage_applied = None`. The confirm tracker only updated `payment_intent.setup_future_usage`.

Connector requests already use `setup_future_usage_applied.or(intent)`, so the Authorize request carried `off_session` correctly. Decisions made **before** the connector call read the attempt alone, though, so they saw `None`.

The visible effect is on connectors that override `should_call_connector_customer` (Shift4, and likewise Mollie, Airwallex, Paysafe and Payload): no connector customer was created for a create-then-confirm off-session payment. For Shift4 routed through UCS, the off-session CIT was then rejected with `IR_04` (missing `customer.connector_customer_id`) and the payment was left `processing`.

## Fix

In the confirm `get_trackers`, fill `setup_future_usage_applied` from the intent when the attempt has none — the same rule `retry.rs:865` already applies to new attempts. Existing precedence is kept: the attempt value wins when set. The persisted value is unchanged, because `payment_response` writes `setup_future_usage_applied` from the connector request after the call.

8 lines in `crates/router/src/core/payments/operations/payment_confirm.rs`.

## Motivation and Context

Found while testing Shift4 recurring payments end-to-end through the Unified Connector Service. Split out of https://github.com/juspay/hyperswitch/pull/14210 because it changes the confirm path for every connector and deserves separate review.

## How did you test it?

Manually, end-to-end (Hyperswitch -> UCS -> Shift4 sandbox), with a debug router built from this change:

- **Before:** create `confirm=false` + confirm with `off_session` -> no connector customer created, payment failed with `IR_04` and stayed `processing`.
- **After:** the same flow succeeds, stores a mandate, and a follow-up merchant-initiated charge succeeds.
- **Regression:** create+confirm in one call (future usage set at create), zero-amount setup, non-zero off-session CIT and one-off payments all still behave as before, for card, Apple Pay and Google Pay.
- `cargo clippy -p router --features v1` is clean.

## Checklist

- [x] `cargo clippy -p router --features v1`
- [ ] Added unit/integration tests — manual E2E only; the behaviour needs a connector with a `should_call_connector_customer` override plus a live processor


shuklatushar226:
Closing: this fix is folded back into #14210 as commit `723cee868f`, byte-identical to the commit here.

Keeping it separate would have split a 3-commit change that was tested together end to end, and #14210 could not merge without it for create-then-confirm off-session payments. Reviewing it in #14210 keeps the fix and its Shift4 test evidence in one place.

PR #14235, opened 2026-09-15
feat(connector): [Braintree] route through UCS and enable the three 3DS legs

shuklatushar226:
## What

Makes the Unified Connector Service's Braintree 3D Secure implementation reachable from Hyperswitch, and routes Braintree through UCS.

Companion UCS PR: juspay/hyperswitch-prism#2298

## Why this is needed

Braintree 3DS is implemented in UCS as three legs:

| Leg | Braintree call |
|---|---|
| `PreAuthenticate` | device data collection — `createClientToken` + `tokenizeCreditCard` |
| `Authenticate` | `performThreeDSecureLookup` |
| `PostAuthenticate` | `node(id:)` readback of the authenticated payment method |

The router decides whether to call each leg from three `ConnectorSpecifications` methods that default to `false` in `hyperswitch_interfaces::api` (`:463`, `:467`, `:471`). **Braintree overrode none of them**, so `pre_authentication_step` (`core/payments/flows/authorize_flow.rs:465`) short-circuited and Authorize ran straight through with no authentication. The UCS 3DS implementation was unreachable from Hyperswitch regardless of configuration.

Eleven connectors already override these — `barclaycard`, `cybersource`, `moneris`, `nexixpay`, `nmi`, `nuvei`, `paysafe`, `redsys`, `shift4`, `worldpayxml`, `ilixium`. Braintree was not among them.

> Note this covers the **granular** path only. UCS also runs its own composite authentication loop via `next_authentication_step`, but the router calls each leg itself rather than using the composite RPC, so that override does not substitute for these gates.

## Changes

**`crates/hyperswitch_connectors/src/connectors/braintree.rs`** — the three gates, following `barclaycard.rs:1640`:

- pre-authentication when `auth_type == ThreeDs && request_data.is_card()`
- `Authenticate` on `CompleteAuthorize` carrying **non-empty** redirect params — the cardholder came back from the ACS and the result still has to be submitted
- `PostAuthenticate` on the exact complement: no params means the ACS posted straight to Braintree's own `termUrl`, so the outcome is already recorded on the payment method and only needs reading back

**`config/development.toml`** — `braintree` added to `ucs_only_connectors`.

## What is deliberately *not* here

No proto change and no `connector_config.rs` change. `BraintreeMetadata` and the `Connector::Braintree` auth arm (`unified_connector_service/connector_config.rs:48`, `:832`) already match the UCS `BraintreeConfig` message field for field. `[tokenization]` (`config/development.toml:1193`) already carries `braintree`, so the router already calls UCS `payment_method_tokenize` before `payment_authorize` — which Braintree requires, since `chargeCreditCard` takes a `paymentMethodId`, not a raw card.

## Testing

Verified against the **live Braintree sandbox through UCS** — real gRPC calls making real HTTPS calls, no mocks:

| Flow | Result |
|---|---|
| Tokenize → Authorize | `CHARGED` |
| Manual capture | `AUTHORIZED` → `CHARGED` |
| Partial capture | passes (see caveat) |
| Void | `VOIDED` |
| Refund / partial refund (after `sandboxSettleTransaction`) | `REFUND_SUCCESS` |
| PSync | `CHARGED` |
| RSync | `REFUND_SUCCESS` |
| Zero-amount SetupRecurring | mandate reference returned |
| MIT with network transaction id | `CHARGED`, NTID returned |
| Payment / refund / dispute webhooks | correct event types, `sourceVerified: true` |
| 3DS PreAuthenticate | `DEVICE_DATA_COLLECTION_PENDING` |

A live decline confirms AVS, CVV, processor and network advice codes all arrive populated, which is what GSM smart retry keys off:

```json
{ "avs_street_address_response_code": "M", "cvv_response_code": "M",
  "processor_response_code": "1000", "processor_response_text": "Approved" }
"issuerDetails": { "networkDetails": { "adviceCode": "01", "declineCode": "XX" } }
```

## Known limitation — read before enabling 3DS in production

**The 3DS challenge and frictionless paths could not be exercised.** `performThreeDSecureLookup` returns `authenticationStatus: LOOKUP_ERROR` on the available sandbox account for every card tried, including Braintree's own 3DS test cards, and including a raw `curl` with no UCS in the path. 3D Secure is simply not provisioned on that gateway.

These gates are correct by construction and mirror a shipped reference, but **they should be re-proven on a 3DS-enabled Braintree account before Braintree 3DS is enabled in production.**


shuklatushar226:
Closing unmerged and reverting at the author's request, together with the companion UCS PR juspay/hyperswitch-prism#2298. Nothing here was merged, so `main` is unaffected.

This PR only existed to make the UCS Braintree 3DS legs reachable from the router. With the UCS side reverted it has no counterpart — and merging it alone would be actively wrong, since it adds `braintree` to `ucs_only_connectors` and would route Braintree traffic down the UCS path with nothing implemented on the other end.

The branch `feat/braintree-ucs-3ds` is being deleted. The commit remains recoverable from this PR:

    git fetch origin pull/14235/head:braintree-hs-recovered

One finding worth keeping regardless of this work: `crates/hyperswitch_connectors/src/connectors/braintree.rs` overrides none of `is_pre_authentication_flow_required` / `is_authentication_flow_required` / `is_post_authentication_flow_required`, all of which default to `false`. So any future Braintree 3DS integration — UCS or direct — needs these gates, or `pre_authentication_step` short-circuits and Authorize runs through with no authentication.

PR #14243, opened 2026-09-15
fix(config): enable mandate support for paypal wallet on paypal connector

### LINKED ISSUE #13765: PayPal wallet: setup_future_usage off_session downgraded to on_session, vault never requested
## Summary

For a PayPal wallet payment created with `setup_future_usage: off_session`, the value reaching the connector is downgraded to `on_session`. The PayPal connector only requests vaulting when it sees `OffSession`, so `payment_source.paypal.attributes` is serialised as `null`, PayPal never creates a payment token, and the payment method cannot be reused for MIT / renewals.

The payment itself succeeds. Only the vaulting is silently skipped.

The intent keeps the requested value; the attempt does not:

| field | value |
|---|---|
| `payment_intent.setup_future_usage` | `off_session` |
| `payment_attempt.setup_future_usage_applied` | `on_session` |

`setup_future_usage_applied` is what `construct_payment_router_data` passes to the connector, so the connector sees `on_session`.

## Expected Behavior

`setup_future_usage: off_session` on a PayPal wallet payment should reach the connector as `OffSession`, so that `paypal/transformers.rs` emits:

```json
"attributes": { "vault": { "store_in_vault": "ON_SUCCESS", "usage_type": "MERCHANT", "permit_multiple_payment_tokens": true } }
```

and PayPal returns a reusable payment token.

## Actual Behavior

`attributes` is sent as `null`. Verified two ways.

Captured from a logging proxy placed between the router and `api-m.sandbox.paypal.com`:

```
POST /v2/checkout/orders
  payment_source.paypal keys: [ 'experience_context', 'attributes' ]
  attributes: null
```

And confirmed independently from PayPal's own API, fetching the resulting order:

```
GET /v2/checkout/orders/{id}
  status: COMPLETED
  payment_source.paypal keys: ['account_id', 'account_status', 'address', 'email_address', 'name']
  attributes: absent
```

Subsequent MIT attempts against the stored payment method fail with `IR_39 No eligible connector was found for the current payment method configuration`. The saved payment method has `recurring_enabled: false` and no `connector_mandate_details`.

## Steps To Reproduce

1. Configure a PayPal connector (`BodyKey`, sandbox credentials) with `wallet` / `paypal` enabled and `recurring_enabled: true`.
2. Create a payment:

```json
POST /payments
{
  "amount": 199,
  "currency": "USD",
  "customer_id": "<customer>",
  "confirm": true,
  "setup_future_usage": "off_session",
  "return_url": "https://example.com/complete",
  "payment_method": "wallet",
  "payment_method_type": "paypal",
  "payment_method_data": { "wallet": { "paypal_redirect": {} } }
}
```

3. Read the response: `setup_future_usage` comes back as `on_session`.
4. Inspect the outgoing PayPal order: `payment_source.paypal.attributes` is `null`.

Splitting create and confirm makes the boundary obvious:

- `POST /payments` with `confirm: false` stores `setup_future_usage: off_session`
- `POST /payments/{id}/confirm` results in `setup_future_usage_applied: on_session`

So the downgrade happens during confirm, not create.

## What I ruled out

`override_setup_future_usage_to_on_session` in `crates/router/src/core/payments/helpers.rs` is the documented downgrade path, gated on the `skip_saving_wallet_at_connector_{merchant_id}` config. It does not appear to be the cause:

- the config key does not exist (`GET /configs/{key}` returns `HE_02`), so it falls back to the `"[]"` default
- reproduced on a new local database whose `configs` table holds only `routing_default_*` rows
- its `logger::debug!("Override setup_future_usage ...")` line never appears at `log.console.level = "DEBUG"`

Adding `customer_acceptance` and `mandate_data` with `multi_use` changes nothing.

## Context

- version: `v1.125.0` (`-standalone` image), also reproduced on a local `v1.125.0` router with a fresh database
- `RUN_ENV=sandbox`
- connector: PayPal, `test_mode: true`
- PayPal app has Save payment methods (Vault) and Subscriptions enabled

Card payments through `recurring_details.processor_payment_token` work correctly for MIT on the same router, so this appears specific to the wallet / PayPal path.

lorenzozanee:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

PayPal wallet payments with setup_future_usage off_session were downgraded to on_session before reaching the connector so no vault token was created. Align wallet.paypal mandate configuration in docker_compose and config.example with sandbox and development.

## Additional Changes

## Motivation and Context

Fixes #13765

## How did you test it?

Verified the mandates configuration for wallet.paypal now includes paypal, globalpay and nexinets and matches sandbox; confirmed the PayPal transformer vault path is preserved for off_session. Local regression check passes.

## Checklist

Supersedes #13922（原PR分支已删，按原提交重建）


abhisheksharma2411 (commented):
Verified the alignment claim rather than eyeballing it, and it's exactly right. After this change all six config files carry the identical string:

| file | `wallet.paypal` before | after |
|---|---|---|
| `config.example.toml` | `adyen,novalnet` | ✅ aligned |
| `docker_compose.toml` | `adyen,novalnet,authorizedotnet` | ✅ aligned |
| `deployments/sandbox.toml` | `adyen,globalpay,nexinets,novalnet,paypal,authorizedotnet` | unchanged |
| `deployments/production.toml` | same | unchanged |
| `deployments/integration_test.toml` | same | unchanged |
| `development.toml` | same | unchanged |

So the two edited files were the only two out of step, and they now match the other four character-for-character. Good fix for #13765.

### The drift is much wider than `wallet.paypal`

I diffed the whole `[mandates.supported_payment_methods]` section between `sandbox.toml` and the two files this PR touches. **25 of 26 keys drift**, and 10 are absent from both entirely:

```
DIFFERS   card.credit, card.debit, wallet.apple_pay, wallet.google_pay,
          wallet.samsung_pay, wallet.paypal, bank_debit.{ach,bacs,becs,sepa},
          bank_redirect.{giropay,ideal,sofort}, pay_later.klarna*
MISSING   wallet.{dana,gcash,go_pay,kakao_pay,momo,twint,vipps},
          bank_redirect.{bancontact_card,eps,open_banking_uk,trustly}
```

\* `pay_later.klarna` matches in `docker_compose` but not `config.example`.

Only `bank_debit.*`… actually only **one** key of 26 is consistent across all three.

That matters more for `docker_compose.toml` than for the example: it's what a self-hosted Docker deployment actually runs, so those users get a different mandate matrix from sandbox and production. The failure mode reported in #13765 — `setup_future_usage: off_session` silently downgraded to `on_session`, no vault token created, no error — is the *generic* symptom of a connector missing from one of these lists. Every one of those 25 keys is the same latent bug for a different payment method.

**Not asking you to fix 25 keys in a PayPal PR** — this one is correctly scoped to the reported issue and shouldn't grow. Raising it because it looks like a systemic drift worth its own issue, and because the fix that actually holds is probably a check rather than another manual sync: nothing today fails when these six files disagree, which is why they've drifted 25 keys apart.

I hit the adjacent version of this recently — `production.toml` had a malformed stanza that disabled `get_connector_config` for *every* connector, and `crates/connector_configs` has no tests at all, so it shipped green (#13874 / #14208). Same shape: config files that nothing compares.

Happy to open the drift issue with the full per-key table if that's useful.

(Reviewer note — not a maintainer, no approval implied.)

PR #728, opened 2023-03-09
chore: address Rust 1.68 clippy lints

SanchithHegde:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates

## Description

This PR addresses a clippy lint stabilized in Rust 1.68.

## Motivation and Context

N/A

## How did you test it?

N/A

## Checklist

- [x] I formatted the code `cargo +nightly fmt --all`
- [x] I addressed lints thrown by `cargo clippy`
- [x] I reviewed submitted code
- [ ] I added unit tests for my changes where possible
- [ ] I added a [CHANGELOG](/CHANGELOG.md) entry if applicable

PR #14253, opened 2026-09-16
perf(ci): bulk-restore sccache cache to speed up CI builds

### LINKED ISSUE #14195: ci: Rust jobs rebuild from scratch — no build cache has been saved since 2026-09-09

hrithikesh026:
## Summary

**Self-hosted `hyperswitch-runners` pool (`test`, `check-v2`, `build-hyperswitch`, `runner_v2`)**
- Switched sccache from its native S3 backend (or no backend at all) to sccache on **local disk**, bulk-restored/saved as a single compressed tarball streamed directly to/from S3 (`tar czf - | aws s3 cp -`, no intermediate file). Native S3's per-object request round trips through a shared CI egress proxy showed a real `Average cache read hit` latency spike under load; a single bulk transfer sidesteps that entirely.
- S3 keys are scoped by `${{ runner.os }}-${{ runner.arch }}` (the pool is a heterogeneous ARM64/AMD64 mix) and by job name (`test-`, `check-v2-`, `build-hyperswitch-`, `runner-v2-`).
- `CI-pr.yml` is permanently restore-only for all four jobs (PRs never write the shared cache); `CI-push.yml`/`cypress-tests-runner.yml` populate on `push`/`merge_group` respectively.
- `build-hyperswitch`/`runner_v2` also drop their `--jobs 2` / `CARGO_BUILD_JOBS: 6` caps entirely (now uncapped) — safe there since they're nearly-100%-cacheable plain builds. `test`/`check-v2` instead get a **`CARGO_BUILD_JOBS: 6`** cap (was `3` in `CI-pr.yml`, `2` in `CI-push.yml`) after a cold-cache run with no cap genuinely OOM-evicted a runner pod (node had ~4.5MB free; the runner container hit ~27.7GiB against its 28G memory request) — 6 is the last production-proven value for this pod type.
- The compiled `router` binary moves from GitHub artifacts (`upload-artifact`/`download-artifact`) to `aws s3 cp` against the same bucket, keyed by `github.run_id` — reuses the pod's existing IAM-role auth, no new secrets.

**GitHub-hosted `ubuntu-latest` (`check-msrv`, `wasm-ci`)**
- `check-msrv` switches from `Swatinem/rust-cache` to sccache on local disk, bulk-restored/saved via `actions/cache` — deliberately not sccache's native GHA backend (`SCCACHE_GHA_ENABLED`), which this repo's own CI history already found slower than no caching at all for this exact job.
- `wasm-ci` (`wasm-bulild-check.yml`) gets the same treatment; it had zero build caching before.
- `actions/cache` keys are immutable (unlike S3 objects, which overwrite in place), so the save side does a delete-then-save dance (`gh cache delete`, needs `permissions: actions: write`) instead.

**`cypress-tests-runner.yml`'s other 5 sccache jobs** (`mandatory-cypress-tests`, `mitm-cassette-replay-cypress-tests`, `extended-cypress-tests`, `runner-optional-connectors`, `runner_alpha`)
- Left on their existing native-S3 sccache design, deliberately out of scope — their only "compile" step (installing `diesel_cli`) is fast regardless of caching, so the added complexity of a bulk-sync design isn't justified there.

**Shared scripts** (`.github/scripts/`)
- Extracted the S3 mask/resolve/restore/save logic — previously duplicated near-identically across 9–11 job instances — into `sccache-mask-s3-env.sh`, `sccache-resolve-s3-env.sh`, `sccache-restore-s3.sh <cache-name>`, `sccache-save-s3.sh <cache-name>` (`$RUNNER_OS`/`$RUNNER_ARCH` come from GitHub Actions' own env vars, no need to pass them explicitly). Also extracted the duplicated "download the `router` binary from S3" step (5 instances) into `download-hyperswitch-binary-s3.sh`.
- The mask step now runs right after checkout in every job (not before, like the original inline version) — it needs the checked-out script to exist on disk, and the earliest S3-touching step in any job is never checkout itself.

**`diesel_cli` install** (`cypress-tests-runner.yml`, `migration-check.yaml`, `postman-collection-runner.yml`)
- Switched all 8 install sites from `baptiste0928/cargo-install` (builds from source, cached in GHA's cache) to `taiki-e/install-action` (downloads `diesel_cli`'s own `cargo-dist`-published prebuilt binary — covers both `aarch64-unknown-linux-gnu` and `x86_64-unknown-linux-gnu`, and ships with all three `*-bundled` backends statically linked, so no `libpq-dev` etc. needed on the runner either). Not a caching-architecture change, just removes an unnecessary source build for a small CLI tool.

## Measured impact

Compared each changed job's latest run on this branch (warm cache) against the latest run of [#14188](https://github.com/juspay/hyperswitch/pull/14188), an unrelated PR on unmodified `main` CI, using each job's own start/completion timestamps:

| Job | Baseline ([#14188](https://github.com/juspay/hyperswitch/pull/14188)) | This PR (warm) | Δ |
|---|---|---|---|
| `test` | [39m21s](https://github.com/juspay/hyperswitch/actions/runs/35357998799/job/105641993309) | [34m25s](https://github.com/juspay/hyperswitch/actions/runs/35468604362/job/105965396872) | **−12.5%** |
| `check-v2` | [40m27s](https://github.com/juspay/hyperswitch/actions/runs/35357998799/job/105641993163) | [36m50s](https://github.com/juspay/hyperswitch/actions/runs/35468604362/job/105965396945) | **−8.9%** |
| `check-msrv` | [10m26s](https://github.com/juspay/hyperswitch/actions/runs/35357998799/job/105641992776) | [9m18s](https://github.com/juspay/hyperswitch/actions/runs/35468604362/job/105965396959) | **−10.9%** |
| `Build hyperswitch server` (`build-hyperswitch`) | [37m08s](https://github.com/juspay/hyperswitch/actions/runs/35357998822/job/105642019573) | [5m19s](https://github.com/juspay/hyperswitch/actions/runs/35468604361/job/105967434622) | **−85.7%** |
| `Run Cypress tests on v2 and generate coverage report` (`runner_v2`) | [23m21s](https://github.com/juspay/hyperswitch/actions/runs/35357998822/job/105642019690) | [9m13s](https://github.com/juspay/hyperswitch/actions/runs/35468604361/job/105967434569) | **−60.5%** |
| `Check wasm build` (`wasm-ci`) | [13m41s](https://github.com/juspay/hyperswitch/actions/runs/35357999316/job/105641998163) | [2m58s](https://github.com/juspay/hyperswitch/actions/runs/35468604562/job/105965359542) | **−78.3%** |

The three plain-`cargo build`-shaped jobs (`build-hyperswitch`, `runner_v2`, `wasm-ci`) see the largest wins — bulk local-disk-tarball sccache (via S3 for the self-hosted pool, `actions/cache` for the GitHub-hosted `wasm-ci`/`check-msrv` runners) restores as one archive instead of one request per object, which is what these nearly-100%-cacheable jobs bottleneck on. `test`/`check-v2` move more modestly since they're clippy-dominated and clippy can't reuse sccache's cache for workspace-crate compiles (see "Known gap" below) — their gains here come from the bulk-restore avoiding S3 request-round-trip contention under load, not from clippy caching more.

Methodology note: `test`/`check-v2`/`check-msrv`/`wasm-ci` are permanently restore-only on PR branches (only `CI-push.yml`'s `push`/`merge_group` runs populate the shared cache) — getting a warm comparison without waiting for a merge to `main` required a branch-scoped temporary save step, reverted once each cache was confirmed populated and restore-verified. `build-hyperswitch`/`runner_v2` populate on `merge_group` already, so no temporary step was needed for those two.

### Known gap: `RUSTC_WORKSPACE_WRAPPER` doesn't help clippy caching here

Also tried adding `RUSTC_WORKSPACE_WRAPPER: sccache` to `test`/`check-v2`, since `RUSTC_WRAPPER` alone only wraps `rustc` for dependency crates once `cargo clippy` is involved — clippy sets `RUSTC_WORKSPACE_WRAPPER=clippy-driver` for workspace crates, which takes precedence. sccache has recognized `clippy-driver` as a cacheable compiler since 2020 ([mozilla/sccache#728](https://github.com/mozilla/sccache/pull/728)), so this seemed promising.

Verified against the same warm cache: it does activate (workspace-crate clippy calls now reach sccache, confirmed via `--show-stats`'s `Compile requests` count going up), but every one of those calls is rejected as non-cacheable under `multiple input files` — a structural argument-shape mismatch, not the "Unrecognized option" crash `#728` fixed. Net effect on both jobs: no new cache hits, `crate-type`-reason count unchanged (confirming unrelated pre-existing rejections weren't touched), just wasted round-trips. **Reverted** — not included in this PR's final diff.

## Test plan

- [x] Confirmed the bulk local-disk+tarball design fixes native S3's latency spike under load (`build-hyperswitch`/`runner_v2`, validated via real populate-then-warm-restore runs: 99.92% hit rate, 0.000s avg read).
- [x] Confirmed `SCCACHE_IDLE_TIMEOUT: "0"` fixes the mid-build server death (root-caused via a real failed run, `router`'s compile outlasting sccache's 600s default, [mozilla/sccache#204](https://github.com/mozilla/sccache/issues/204)).
- [x] Confirmed `RUSTC_WORKSPACE_WRAPPER` activates for clippy but yields no new cache hits (see "Known gap" below) — reverted, not part of this diff.
- [x] Root-caused a real `test`/`check-v2` runner-pod OOM eviction to cold cache + uncapped `CARGO_BUILD_JOBS`; fixed by re-capping at 6 (last production-proven value for `hyperswitch-runners`) — verified via a clean post-fix run (warm, 96.5% Rust hit rate, no crash).
- [x] Confirmed `check-msrv`/`wasm-ci`'s `actions/cache`-backed sccache populates and warm-restores correctly (verified via a real populate-then-restore round on this branch).
- [x] Confirmed the S3 mask/resolve/restore/save script extraction didn't change behavior — caught and fixed one regression (the mask step calling a script that didn't exist yet pre-checkout) before merge.
- [x] Confirmed `taiki-e/install-action`'s `diesel_cli` fallback resolves to `cargo-dist`'s prebuilt binary rather than a source build, and works on both ARM64 and AMD64.
- [ ] Watch the first real `push`-to-`main` run of `check-msrv`/`test`/`check-v2` populate the shared cache correctly — so far only validated via temporary branch-scoped populate commits, since this design hasn't gone through an actual `push` event yet.


XyneSpaces:
:warning: The workflow uses the floating tag `taiki-e/install-action@v2` for installing sccache and cargo-hack. Floating major-version tags can silently pick up a new, compromised release and bypass pinned-version review.

Pin these to an immutable SHA (or at minimum a specific semver tag such as `v2.47.0`) and update the dependency inventory accordingly.

XyneSpaces:
[blocking] The `test` and `check-v2` jobs now run `aws s3 sync` against the sccache bucket on self-hosted runners that also execute fork-PR code. This gives fork PRs ambient S3 read/write access to the build cache bucket with no approval gate.

Gate the S3 restore/save steps so they do not run for `pull_request` events from forks, e.g. `if: github.event.pull_request.head.repo.full_name == github.repository`, or move these jobs to the GitHub Actions cache backend which already blocks fork cache writes.

.github/workflows/CI-pr.yml
```diff
@@ -135,11 +136,30 @@ jobs:
...
+      SCCACHE_S3_RW_MODE: READ_ONLY
 
     steps:
+      - name: Mask S3 cache location in job logs
+        shell: bash
+        run: |
+          # Not secrets (auth is via IAM role), but no reason to publish the
+          # bucket layout in a public repo's PR logs — including logs from
+          # fork PRs. sccache's own `--show-stats` output prints these back
+          # verbatim, so mask before anything can echo them.
```
pixincreate, line 158: i don't find a reason to have these comments.

also, the buckets name needs to be put in github secrets. and putting it there shouldn't require us to mask the values explicitly afaik.

.github/workflows/CI-pr.yml
```diff
@@ -217,6 +233,11 @@ jobs:
         shell: bash
         run: cargo check --features "release"
 
+      - name: Show sccache stats
+        if: always()
+        shell: bash
+        run: sccache --show-stats
+
```
pixincreate, line 240: do we need this?

.github/workflows/CI-pr.yml
```diff
@@ -191,14 +211,10 @@ jobs:
 
       - name: Run clippy (redis-rs backend)
         shell: bash
-        env:
-          CARGO_BUILD_JOBS: 3
         run: just clippy
 
       - name: Run clippy (fred backend)
         shell: bash
-        env:
```
pixincreate, line 200: why removed?

.github/workflows/cypress-tests-runner.yml
```diff
@@ -302,12 +333,15 @@ jobs:
...
         if: ${{ env.RUN_TESTS == 'true' }}
-        uses: actions/download-artifact@v8
-        with:
-          name: hyperswitch-server
-          path: target/debug
+        shell: bash
+        run: |
+          mkdir -p target/debug
+          aws s3 cp \
+            "s3://${SCCACHE_BUCKET}/${SCCACHE_S3_KEY_PREFIX}build-artifacts/${{ github.run_id }}/router" \
```
pixincreate, line 342: put the values in secrets. put it in env in steps. use in run.

.github/workflows/CI-pr.yml
hrithikesh026: Yes. Helps with debugging the cache hit rates for sccache.

.github/workflows/CI-pr.yml
hrithikesh026: It was only slowing down the compile time. It was introduced when github runners were getting OOM killed. Since we are now using self-hosted runners, it was a dead weight.

.github/workflows/cypress-tests-runner.yml
hrithikesh026: Will take that up in a separate refactor PR. Currently we already have these envs set at the hyperswitch-runners. If needed can move it to secrets later.

.github/workflows/CI-pr.yml
hrithikesh026: I feel it explains the need for masking the envs.

.github/workflows/CI-pr.yml
```diff
@@ -191,14 +224,15 @@ jobs:
 
       - name: Run clippy (redis-rs backend)
         shell: bash
+        # Capped: uncapped + cold cache OOM'd a runner pod (git log).
         env:
-          CARGO_BUILD_JOBS: 3
+          CARGO_BUILD_JOBS: 6
```
SanchithHegde: We can probably go to higher numbers for `CARGO_BUILD_JOBS` I feel.

.github/workflows/wasm-bulild-check.yml
```diff
@@ -36,3 +62,23 @@ jobs:
...
+      - name: Delete stale sccache cache entry
+        if: ${{ always() && github.event_name == 'push' }}
+        continue-on-error: true
+        shell: bash
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: gh cache delete "sccache-local-wasm-${{ runner.os }}-${{ runner.arch }}" --repo "${{ github.repository }}"
+
+      - name: Save sccache cache
+        if: ${{ always() && github.event_name == 'push' }}
```
SanchithHegde: Won't this cause the step to run even if the workflow is cancelled?

https://docs.github.com/en/actions/reference/workflows-and-actions/expressions#always

.github/workflows/cypress-tests-runner.yml
```diff
@@ -549,6 +588,7 @@ jobs:
     env:
       # Use `sccache` for caching compilation artifacts
       RUSTC_WRAPPER: sccache
+      SCCACHE_IDLE_TIMEOUT: "0"
```
SanchithHegde: Do we need this env var in this job though?

.github/workflows/cypress-tests-runner.yml
```diff
@@ -629,11 +671,14 @@ jobs:
...
         with:
-          crate: diesel_cli
-          features: postgres
-          args: --no-default-features
+          tool: diesel_cli
+
+      - name: Show sccache stats
+        if: ${{ always() && env.RUN_TESTS == 'true' }}
+        shell: bash
+        run: sccache --show-stats
```
SanchithHegde: Nit: Do we need this step here?

hrithikesh026: Yeah. Helpful in debugging later.

SanchithHegde: But we are not doing any compilation in these jobs which run Cypress tests, no?

.github/workflows/wasm-bulild-check.yml
```diff
@@ -23,6 +37,18 @@ jobs:
...
           toolchain: stable 2 weeks ago
 
+      - name: Install sccache
+        uses: taiki-e/install-action@v2
+        with:
+          tool: sccache
+          checksum: true
+
+      - name: Restore sccache cache
+        uses: actions/cache/restore@v4
```
pixincreate: v6 is the latest one.

https://github.com/actions/cache

.github/workflows/wasm-bulild-check.yml
```diff
@@ -36,3 +62,24 @@ jobs:
...
+        if: ${{ !cancelled() && github.event_name == 'push' }}
+        continue-on-error: true
+        shell: bash
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: gh cache delete "sccache-local-wasm-${{ runner.os }}-${{ runner.arch }}" --repo "${{ github.repository }}"
+
+      - name: Save sccache cache
+        if: ${{ !cancelled() && github.event_name == 'push' }}
+        uses: actions/cache/save@v4
```
pixincreate: v5 is the latest one.

https://github.com/actions/cache/tree/main/save

.github/workflows/CI-pr.yml
```diff
@@ -121,22 +127,38 @@ jobs:
...
-        uses: Swatinem/rust-cache@v2.9.1
+      - name: Install sccache
+        uses: taiki-e/install-action@v2
         with:
-          save-if: false
+          tool: sccache
+          checksum: true
+
+      - name: Restore sccache cache
+        uses: actions/cache/restore@v4
```
pixincreate: v5 is latest one?

https://github.com/actions/cache/tree/main/restore

.github/workflows/CI-pr.yml
pixincreate: before merging to main, i think it is better we remove it?

.github/scripts/download-hyperswitch-binary-s3.sh
```diff
@@ -0,0 +1,16 @@
...
+if [[ "${CI:-false}" != "true" && "${GITHUB_ACTIONS:-false}" != "true" ]]; then
+  echo "This script is to be run in a GitHub Actions runner only. Exiting."
+  exit 1
+fi
+
+# Keyed by run id: build-hyperswitch uploads its compiled router binary
+# under this same path for every job in this run to download.
+mkdir -p target/debug
+aws s3 cp \
+  "s3://${CACHE_S3_BUCKET}/${CACHE_S3_KEY_PREFIX}build-artifacts/${GITHUB_RUN_ID}/router" \
```
pixincreate, line 14: missed the `/` i guess?

```suggestion
  "s3://${CACHE_S3_BUCKET}/${CACHE_S3_KEY_PREFIX}/build-artifacts/${GITHUB_RUN_ID}/router" \
```

.github/scripts/sccache-restore-s3.sh
```diff
@@ -0,0 +1,47 @@
...
+  echo "::warning::S3 cache bucket not configured on this runner; sccache starts cold"
+  exit 0
+fi
+
+# Streamed, not written to disk first — avoids doubling disk usage.
+restore() {
+  local key="$1"
+  echo "Restoring sccache cache, key: ${key}"
+  aws s3 cp \
+    "s3://${CACHE_S3_BUCKET}/${CACHE_S3_KEY_PREFIX}${key}" \
```
pixincreate, line 31: same here.


```suggestion
    "s3://${CACHE_S3_BUCKET}/${CACHE_S3_KEY_PREFIX}/${key}" \
```

.github/scripts/sccache-save-s3.sh
```diff
@@ -0,0 +1,26 @@
...
+
+# PR-scoped when a PR number is given, so concurrent PRs don't clobber each
+# other's — or the shared merge_group/main — cache.
+key="sccache-cache/${cache_name}-${RUNNER_OS}-${RUNNER_ARCH}${pr_number:+-pr${pr_number}}.tar.gz"
+echo "Saving sccache cache, key: ${key}"
+
+# Streamed, not written to disk first — avoids doubling disk usage.
+tar czf - -C "$SCCACHE_DIR" . \
+  | aws s3 cp - \
+    "s3://${CACHE_S3_BUCKET}/${CACHE_S3_KEY_PREFIX}${key}" \
```
pixincreate, line 25:
```suggestion
    "s3://${CACHE_S3_BUCKET}/${CACHE_S3_KEY_PREFIX}/${key}" \
```

unless `/` is added alongside the prefix...

.github/workflows/CI-push.yml
```diff
@@ -196,36 +233,60 @@ jobs:
...
 
       - name: Cargo build release
         if: ${{ github.event_name == 'merge_group' }}
         shell: bash
+        env:
+          CARGO_BUILD_JOBS: 9
         run: cargo check --features "release"
 
+      - name: Show sccache stats
+        if: always()
```
pixincreate, line 279: this conditional is not needed?

.github/scripts/download-hyperswitch-binary-s3.sh
hrithikesh026: No. this is intentional.

.github/scripts/sccache-restore-s3.sh
hrithikesh026: This is intentional. Prefix should be strictly prepended to the key

.github/workflows/CI-push.yml
hrithikesh026: I prefer this to be run all the time.

pixincreate (approved):

.github/workflows/CI-push.yml
SanchithHegde: What about when the workflow is cancelled?

hrithikesh026: wouldn't hurt right?

SanchithHegde (approved):

.github/scripts/sccache-save-s3.sh
hrithikesh026: This is also intentional

PR #14241, opened 2026-09-15
feat(observability): add list, create, retrieve and delete apis for alerts_dicts

VenuMadhav2541:
## Type of Change
- [x] New feature

## Description
Adds the mappers dictionary (`alerts_dicts`) APIs to the observability service, following the
`alerts_info` layering: list (`GET /alerts/alerts_manager/dicts`, enabled entries unless
`is_enabled` is given), save a new version (`POST`, one transaction that disables the live entry,
inserts the new one and keeps the two newest), and retrieve / delete by id. Deleting never
re-enables an older version. Shared pieces: `DuplicateResource` (409 `HE_01`),
`StorageErrorExt::to_duplicate_response`, and diesel's `uuid` feature.

### Additional Changes
- [x] This PR modifies the API contract

## Motivation and Context
Lets the alert manager, its dashboard backend and the alert jobs read and write the dictionary
through the observability service's Postgres database instead of the R alerts API.

## How did you test it?
`cargo check -p observability --all-targets` pass. `cargo clippy -p observability --all-targets`
pass (only 2 pre-existing warnings in unrelated files, untouched by this change). `cargo test -p
observability`: 99 passed, 0 failed (93 unit + 6 route). `cargo +nightly fmt --all` clean.

Manual verification against a local Postgres instance (real request/response pairs, actual bytes
returned):

<details>
<summary>Full manual test matrix (request + actual response, incl. a 20-way concurrency race)</summary>

### Save a new version — `POST /alerts/alerts_manager/dicts`
**Positive — first save gets defaults:**
```
curl -H 'X-Internal-Api-Key: <key>' -H 'Content-Type: application/json' -d '{"name":"t","key_":"k","values_":["v1"]}' .../dicts
→ 200 {"id":"2d76c38d-...","name":"t","key_":"k","product":[],"values_":["v1"],"ts_created":"2026-09-15T18:10:13.000Z","is_enabled":true,"username":"reliability_team","metadata":[]}
```
**Positive — a second save disables the first and becomes live; a third save prunes the first to keep only the newest two** — verified against `psql`:
```
id                  | key_ | values_ | is_enabled | ts_created
08c9e39f... | k | ["v2"] | f | 2026-09-15 18:10:14
59749ef4... | k | ["v3"] | t | 2026-09-15 18:10:15
```
**Negative — no API key:** `→ 401 IR_01`. **`values_` missing:** `→ 400 IR_04`. **caller sets
[...]

PR #14255, opened 2026-09-16
refactor(observability): group alert info modules

VenuMadhav2541:
## Summary
- Move alert-info API, Diesel, query, core, DB, domain-model, and route modules under alert_manager/alert_info.
- Preserve existing endpoints, behavior, and database schema.
- Keep this change separate from the alerts-dicts implementation.

## Merge order
Merge after PR #14241 (alerts-dicts).

## Validation
- cargo fmt --all
- cargo check -p observability


XyneSpaces:
[blocking] `crates/observability/migrations/2026-09-09-000001_create_alert_state_tables/up.sql` is being edited in-place (renaming `key_` → `key`, `values_` → `values` and rebuilding indexes on those renamed columns).

Once this migration has run anywhere, modifying the same file leaves existing deployments with the old column names while new deployments get the new names. Add a follow-up migration that renames the columns/indexes instead of changing the historical one.

XyneSpaces:
[blocking] Possible breaking API change — public request field renamed from `values_` to `values`.

The diff changes `AlertsDictsCreateRequest.values_` to `values` in `crates/api_models/src/observability/alert_manager/alert_dicts.rs`. If the `/alerts/alerts_manager/dicts` endpoint is already live, existing clients sending `values_` will fail to deserialize. A refactor should preserve the wire contract.

Either keep the old field name or support both during a transition:
```rust
#[serde(alias = "values_")]
pub values: Value,
```

XyneSpaces:
[should-fix] The new observability HTTP handlers are plain `pub async fn`s without `#[instrument]` spans. Other services in the codebase instrument route handlers for tracing and metrics; leaving these handlers uninstrumented makes failures and latency here invisible.

Add `#[instrument(skip(state, request, ...), fields(flow = "..."))]` to each new handler, matching the router crate convention.

PR #14259, opened 2026-09-16
feat(offers): send up to 9-digit card BIN to Offer Engine

apoorvdixit88:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

Hyperswitch sent only the **6-digit ISIN** as `card_bin` in the Offer Engine payload, on both `/v1/offers/list` (eligibility) and `/offers/apply` (confirm). Offer Engine accepts BINs of up to **9 digits**, and its `CARD_BIN` filter derives every prefix from 6 digits up to the length it receives and matches any of them. So a merchant could configure an offer on a 7, 8 or 9-digit BIN and it could never match — the extra digits never left Hyperswitch.

This sends the longest BIN available, capped at 9 digits, from both offer call sites.

- `cards`: new `MAX_OFFER_CARD_BIN_LENGTH = 9` constant next to the existing BIN bounds, and a `get_offer_card_bin()` accessor on `CardNumber` and on `CardBin`. Both take **no length argument**, so no caller outside the crate can widen the result towards the full card number; the existing private `get_bin_prefix(len)` helper stays private and still backs `get_card_isin` (6) and `get_extended_card_bin` (8).
- `hyperswitch_domain_models`: `EligibilityPaymentMethodData::get_offer_card_bin()` dispatches over the `Card` (full PAN) and `CardBin` (BIN-only) variants — the BIN-only variant carries no PAN to slice, hence the accessor on both types.
- `router`: both offer call sites call that one accessor, so eligibility and apply cannot drift apart.

Because the engine expands prefixes downward, sending a longer BIN is a superset: offers configured on 6 digits keep matching exactly as before. `get_card_iin()` is deliberately untouched — `cards_info` lookups, routing and debit routing are keyed on the 6-digit ISIN.

Sending fewer than 9 digits is still correct where fewer exist: for the BIN-only eligibility payload we forward exactly what the client sent (`cards::CardBin` accepts 6–10 digits, trimmed to 9), rather than truncating it to 6.

The first commit exposed `get_bin_prefix(len)` publicly; that was reworked in the follow-up commit, since a public length parameter on a PAN-bearing type let a caller read back the whole card number.

### Additional Changes

## Motivation and Context

Part of Offer Engine Milestone 4. Merchants receive BIN lists from issuers at finer granularity than 6 digits (card-program level), upload them as `CARD_BIN` filter values, and those offers were silently unmatchable.

Evidence from the Offer Engine source for the 9-digit ceiling and the prefix expansion:

- `Offer/Utils/Validation.hs` — `cardBin` accepts a BIN of length <= 9, enforced at `/offers/apply`.
- `Offer/Domain/Validator/FilterValidator.hs` — the `CARD_BIN` filter expands the received BIN into `[6..(T.length value)]` prefixes.
- `Storage/Repository/Filters.hs` — the existence query matches a stored filter row against any of those prefixes (`f.value inArray_ filterValue`).

Issue: https://github.com/juspay/hyperswitch-cloud/issues/23538

## How did you test it?

Unit tests added in `crates/cards/src/validate.rs`:

```
cargo test -p cards --lib offer_card_bin
  test validate::tests::offer_card_bin_from_card_bin ... ok
  test validate::tests::offer_card_bin_from_card_number ... ok
```

They cover the 9 digits taken from a PAN, a `CardBin` holding fewer digits returning only what it holds, and a 10-digit `CardBin` being trimmed to 9.

Manually tested end to end against the Offer Engine sandbox (merchant `offershs`), with the local router's Offer Engine calls routed through a logging proxy to capture the exact `card_bin` on the wire.

**1. Offer Engine matching — offers created with `CARD_BIN` whitelists at 6, 7, 8 and 9 digits, then eligibility requested with BINs of each length:**

| `card_bin` sent | digits | 6-digit offer | 7-digit offer | 8-digit offer | 9-digit offer |
| --- | --- | --- | --- | --- | --- |
| `424242` | 6 | ELIGIBLE | INELIGIBLE | INELIGIBLE | INELIGIBLE |
| `4242424` | 7 | ELIGIBLE | ELIGIBLE | INELIGIBLE | INELIGIBLE |
| `42424242` | 8 | ELIGIBLE | ELIGIBLE | ELIGIBLE | INELIGIBLE |
| `424242424` | 9 | ELIGIBLE | ELIGIBLE | ELIGIBLE | ELIGIBLE |
| `424242999` | 9 | ELIGIBLE | INELIGIBLE | INELIGIBLE | INELIGIBLE |

The first row is the behaviour before this change: three correctly configured offers unmatchable. The last row confirms matching is exact-prefix, not loose.

**2. Full payment flows** (`POST /payments` -> `POST /payments/{id}/eligibility` -> `POST /payments/{id}/confirm` with `offer_details.offer_quote_ids`), merchant-level Offer Engine credentials, dummy connector:

| Card | `card_bin` on the wire | Offer applied (configured BIN length) | Result |
| --- | --- | --- | --- |
| `4242424242424242` | `424242424` (9) | 6-digit, 15% | succeeded, 10000 -> 8500 |
| `4242424242424242` | `424242424` (9) | 9-digit, 9% | succeeded, 10000 -> 9100 |
| `4242424242424242` | `424242424` (9) | 8-digit, 25% | succeeded, 10000 -> 7500 |
| `5555555555554444` | `555555555` (9) | 6-digit, 15% | succeeded, 10000 -> 8500 |
| `4111111111111111` | `411111111` (9) | none (BIN not whitelisted) | succeeded, no discount |

`card_bin` carried 9 digits on both `/v1/offers/list` and `/offers/apply` in every run, so the quote validated at eligibility is applied against the same BIN at confirm. The last row is the negative control: a longer BIN did not make matching looser.

**3. BIN-only eligibility** (`"payment_method_data": {"card_bin": {"card_bin": "42424242"}}`, no PAN): forwarded `card_bin=42424242` — all 8 digits, previously truncated to 6 — and the 8- and 7-digit offers matched while the 9-digit one correctly did not.

Also verified: sending both `card` and `card_bin` is rejected at deserialization, so exactly one BIN source exists per request; and non-card payment methods omit the field entirely.

## Checklist


apoorvdixit88:
Addressed in `649de546d2`: the public `CardNumber::get_bin_prefix(len)` from the first commit let a caller pass any length and read back the full card number, which is not something a PAN-bearing type should expose.

That helper is private again, and the only public route is a fixed-length accessor:

```rust
/// Card BIN length sent to Offer Engine — the longest BIN it accepts.
pub const MAX_OFFER_CARD_BIN_LENGTH: usize = 9;

impl CardNumber {
    fn get_bin_prefix(&self, len: usize) -> String { ... }   // private

    /// Card BIN of [`MAX_OFFER_CARD_BIN_LENGTH`] digits. Takes no length, so no caller can
    /// widen it towards the full card number.
    pub fn get_offer_card_bin(&self) -> String {
        self.get_bin_prefix(MAX_OFFER_CARD_BIN_LENGTH)
    }
}
```

`CardBin` gets the same no-argument accessor (it holds 6-10 digits and is not a PAN, but keeping the names symmetric lets the two match arms read identically), and both offer call sites now call `get_offer_card_bin()` with no length. `get_card_isin()` (6) and `get_extended_card_bin()` (8) keep using the private helper internally, unchanged.

Since the length now lives in `cards` next to `MIN_CARD_BIN_LENGTH`/`MAX_CARD_BIN_LENGTH`, the `offer_engine::MAX_CARD_BIN_LENGTH` constant it replaced is removed rather than left as dead code.

Tests updated accordingly — `cargo test -p cards --lib offer_card_bin` covers the 9 digits taken from a PAN, a shorter `CardBin` returning only what it holds, and a 10-digit `CardBin` trimmed to 9.

ayush22667 (approved):

XyneSpaces:
[should-fix] Wildcard match on payment-method enum

`crates/hyperswitch_domain_models/src/payment_method_data.rs:454` implements `EligibilityPaymentMethodData::get_offer_card_bin` with a `_ => None` wildcard arm. The baseline requires explicit matches on domain enums so adding a new variant forces a conscious decision. Enumerate the remaining variants explicitly and remove the catch-all.

abhisheksharma2411 (commented):
The API design here is the part I'd call out as right: `get_offer_card_bin()` takes **no length argument**, and the doc comment says why — "no caller can widen it towards the full card number." That's the correct shape for a function whose job is to let card digits leave the system, and it's easy to get wrong by exposing `get_bin_prefix(n)` directly instead. The `CardBin` overload capping at the digits actually provided, with a test for both the 8- and 10-digit cases, is a good touch.

I checked the thing that worried me first and it came back clean: `OfferPaymentMethodInfo` carries **only** `card_bin` — no `last4`, no PAN, no token. So the obvious reconstruction concern (9 leading digits alongside a stored last-4 leaves ~3 unknown digits, and Luhn kills most of those) doesn't apply to this payload. Worth stating in the PR description, because it's the first question a reviewer will have about moving 6 → 9.

Two things worth a second look:

**1. `card_bin` is an unmasked `String` on a `Debug` struct, one field away from the masking pattern.** In `core/offer_engine/types.rs`:

```rust
#[derive(Debug, Clone)]
pub struct ... { pub api_key: Secret<String>, ... }      // L28 — masked

#[derive(Debug, Clone, serde::Serialize)]
pub struct OfferPaymentMethodInfo {
    pub card_bin: Option<String>,                         // L73 — plain
```

The file already imports `hyperswitch_masking::Secret` and uses it for the credential, so the convention is present and `card_bin` sits outside it. I grepped `offer_engine/{eligibility,apply}.rs` and found **no logging of this struct**, so today this is latent rather than live — but `Debug` is derived, and this PR raises what any future `{:?}` on it would print from 6 digits to 9. Given the whole design premise is "no caller can widen this," `Secret<String>` on the field would make the type enforce what the getter already does.

**2. Where does `9` come from?** `MAX_OFFER_CARD_BIN_LENGTH = 9` is described as "the longest BIN it accepts," which is a contract with an external system, and the constant is the only thing bounding how much PAN leaves Hyperswitch. A link to the Offer Engine spec or field definition in the doc comment would mean the next person can verify it rather than trust it — and would catch it if Offer Engine ever widens the field, since the natural follow-up ("they accept 11 now") is a one-character change to a security-relevant bound.

Neither blocks the change. The eligibility/confirm symmetry looks right, and the motivation — 7/8/9-digit BIN filters silently never matching — is a real bug worth fixing.

ShankarSinghC (approved):

ShankarSinghC (approved):

ayush22667 (approved):

apoorvdixit88:
@abhisheksharma2411 Hey, correct, better to wrap the bin in a secret. `card_bin` is `Secret<String>` now on `OfferPaymentMethodInfo` and both flow contexts.

Kept `get_offer_card_bin()` returning `String` though. Same as `get_card_isin` and `get_extended_card_bin`, and the isin already goes out as a plain `String` in the payments response, so masking at the accessor wouldn't really buy anything. Masking at the field instead, like `debit_routing` does with the extended bin. `Secret`'s `Serialize` is transparent so the payload to OE is unchanged.

On where the `9` comes from, it's OE's own limit. `Validation.hs` rejects anything longer with "CardBin more than length 9", and the `CARD_BIN` filter builds prefixes `[6..length]` in `FilterValidator.hs`, which is what the comment already describes.

PR #14265, opened 2026-09-16
feat(observability): add threshold override APIs

ThisIsMani:
## Scope

Adds the observability-plane side of the success-rate threshold override vertical slice only. It introduces authenticated threshold list/upsert/delete APIs and PostgreSQL persistence for `hyperswitch_alerts_sr_thresholds` state. It does not migrate lifecycle events, blacklist, dictionaries, metadata, toggles, users, or any analytical/business ClickHouse reads.

## Schema and API contract

- Adds `success_rate_threshold_overrides` with primary key `(name, product, merchant_id, profile_id)`, nullable threshold values, actor/update time, and tombstones.
- Adds authenticated `GET`, `POST`, and `DELETE /alerts/thresholds` routes using `InternalApiKeyAuth` through `server_wrap`.
- Upserts are full replacements; deletes are attributed idempotent tombstones.
- Active rows are ordered by merchant, profile, name, and product.
- Enforces a constant 5000-row global active-rule cap with simple Diesel existence/count/upsert queries and returns the explicit `429` error envelope on exhaustion.
- Preserves trimming, finite-number validation, nullable fields, all-null active rows, and all threshold names/products.

## Tests run

- `diesel migration redo --all --config-file crates/observability/diesel.toml --migration-dir crates/observability/migrations --database-url "$OBSERVABILITY_TEST_DATABASE_URL" --locked-schema`
- `cargo test -p observability --test thresholds` — 9 passed; PostgreSQL repository test intentionally ignored without `OBSERVABILITY_TEST_DATABASE_URL`
- `nix shell nixpkgs#perl -c cargo test -p observability --test notify` — 16 passed
- `cargo check -p diesel_models -p observability`
- `git diff --check`

Analytical ClickHouse reads are unchanged.


config/observability.toml
```diff
@@ -53,6 +53,10 @@ pool_size = 5
 min_idle_pool_size = 1
 connection_timeout = 10
 
+# Global cap across all active threshold names and products. Set to 0 or less to disable.
+[limits]
+max_active_threshold_rules = 5000
```
ThisIsMani: Let's make this a const.

crates/api_models/src/observability/alert_manager/thresholds.rs
```diff
@@ -0,0 +1,64 @@
...
+}
+
+/// Replacement body for one threshold override.
+#[derive(Clone, Debug, Deserialize)]
+#[serde(deny_unknown_fields)]
+pub struct ThresholdUpsertRequest {
+    pub name: String,
+    pub product: String,
+    pub merchant_id: String,
+    #[serde(default)]
```
ThisIsMani, line 27: What will be the default value here? Empty string?

crates/api_models/src/observability.rs
```diff
@@ -1,3 +1,4 @@
 //! Request and response types for the standalone `observability` service.
 
 pub mod alerts_info;
+pub mod thresholds;
```
ThisIsMani: Let's move this to alerts_manager module.

ThisIsMani: Same with diesel models and queries.

scripts/backfill_threshold_overrides.sh
ThisIsMani, line 1: Remove this.

docs/threshold-overrides-cutover.md
ThisIsMani, line 1: This as well.

crates/observability/src/routes/thresholds.rs
```diff
@@ -0,0 +1,64 @@
...
+        core::thresholds::list,
+        &auth::InternalApiKeyAuth,
+    )
+    .await
+}
+
+pub async fn upsert(
+    state: web::Data<AppState>,
+    request: HttpRequest,
+    payload: web::Bytes,
```
ThisIsMani: Why are we not using web::Json? Check other APIs or APIs in router.

crates/observability/src/domain_models/thresholds.rs
```diff
@@ -0,0 +1,242 @@
...
+        ("product", product),
+        ("merchant_id", merchant_id),
+        ("updated_by", updated_by),
+    ] {
+        if value.trim().is_empty() {
+            return Err(report!(ObservabilityError::InvalidRequest))
+                .attach_printable(format!("{field} must not be empty"));
+        }
+    }
+    Ok(())
```
ThisIsMani: Let's do this separately. Don't use for.

config/observability.toml
ThisIsMani: Moved the default to `MAX_ACTIVE_THRESHOLD_RULES` and removed the value from `config/observability.toml`.

crates/api_models/src/observability/alert_manager/thresholds.rs
ThisIsMani: Yes, the serde default is an empty string, meaning all profiles for the merchant. Added this explicitly to the field documentation.

crates/api_models/src/observability.rs
ThisIsMani: Moved `alerts_info` and `thresholds` under the new `observability::alert_manager` module.

ThisIsMani: Applied the same `alert_manager` grouping to Diesel models and their query modules.

scripts/backfill_threshold_overrides.sh
ThisIsMani: Removed the backfill script from this PR.

docs/threshold-overrides-cutover.md
ThisIsMani: Removed the cutover document from this PR.

crates/observability/src/routes/thresholds.rs
ThisIsMani: Changed POST and DELETE handlers to use `web::Json`, matching the other APIs.

crates/observability/src/domain_models/thresholds.rs
ThisIsMani: Replaced the loop with separate `validate_non_empty` calls for each field.

crates/observability/src/settings.rs
```diff
@@ -72,6 +72,25 @@ pub struct Settings<S: SecretState> {
     pub cloudwatch: CloudWatchSettings,
     /// The observability database, which holds alert state. Separate from `hyperswitch_db`.
     pub database: SecretStateContainer<Database, S>,
+    /// Caps on mutable observability state.
+    pub limits: Limits,
+}
```
ThisIsMani: Let's remove this from configs.

ThisIsMani: Removed the limit from `Settings` and secret transformation entirely. The 5000-row cap is now a constant in the threshold core.

PR #14270, opened 2026-09-16
feat(observability): add alert dictionary APIs

ThisIsMani:
## Summary

This is the stacked Alert dictionary port for mutable alert state. It ports dictionary list and upsert behavior keyed by `(name, key_)` while preserving the portal's JSON-encoded string representation and replacement defaults.

Depends on base PR #14269 (`feat/observability-alert-rule-toggles`).

## APIs

- `GET /alerts/dictionary`
- `PUT /alerts/dictionary`
- Both routes use `InternalApiKeyAuth` / `X-Internal-Api-Key`, typed models, and the existing `server_wrap` pattern.

## Schema and persistence

- Adds the reversible `alert_dictionary` PostgreSQL migration.
- Adds alert-manager API, domain, Diesel model, and Diesel query modules.
- Uses Diesel DSL for ordered list and `(name, key_)` upsert operations.

## Tests

- `cargo fmt --all`
- `cargo check -p observability`
- `cargo test -p observability --test dictionary` (5 passed, 1 PostgreSQL test ignored in the normal run)
- `cargo test -p observability` (121 passed, 4 PostgreSQL tests ignored)
- `diesel migration run --locked-schema` against a temporary local PostgreSQL database
- `diesel migration redo --locked-schema --all` against a temporary local PostgreSQL database
- `OBSERVABILITY_TEST_DATABASE_URL=... cargo test -p observability --test dictionary postgres_repository -- --ignored` (passed)
- `git diff --check`

Analytical/business ClickHouse reads and users are unchanged.


crates/api_models/src/observability/alert_manager/dictionary.rs
```diff
@@ -0,0 +1,56 @@
...
+#[derive(Clone, Debug, Serialize)]
+pub struct DictionaryListResponse {
+    pub entries: Vec<DictionaryEntryResponse>,
+}
+
+#[derive(Clone, Debug, Serialize)]
+pub struct DictionaryUpsertResponse {
+    pub ok: bool,
+    pub name: String,
+    pub key_: String,
```
ThisIsMani: Let's not have these underscores in here. Let's normalise everything to non-underscores... Applies to all PRs.

crates/diesel_models/src/observability/alert_manager/dictionary.rs
```diff
@@ -0,0 +1,33 @@
...
+
+use crate::observability::schema::alert_dictionary;
+
+#[derive(Clone, Debug, Insertable)]
+#[diesel(table_name = alert_dictionary)]
+pub struct DictionaryEntryNew {
+    pub name: String,
+    pub key_: String,
+    pub product: String,
+    pub values_: String,
```
ThisIsMani: Same here. We can update the migrations if needed.

crates/api_models/src/observability/alert_manager/dictionary.rs
```diff
@@ -0,0 +1,56 @@
...
+#[serde(deny_unknown_fields)]
+pub struct DictionaryUpsertRequest {
+    pub name: String,
+    pub key_: String,
+    #[serde(default = "default_product")]
+    pub product: String,
+    #[serde(default = "default_values")]
+    pub values_: String,
+    #[serde(default = "default_metadata")]
+    pub metadata: String,
```
ThisIsMani: Should we use vec and json instead?

crates/api_models/src/observability/alert_manager/dictionary.rs
ThisIsMani: Fixed: the dictionary API fields are now `key` and `values`.

crates/diesel_models/src/observability/alert_manager/dictionary.rs
ThisIsMani: Fixed across the new table migration, Diesel schema/models/query, domain model, and tests: `key` and `values` no longer have trailing underscores.

crates/api_models/src/observability/alert_manager/dictionary.rs
ThisIsMani: Keeping these as JSON-encoded strings intentionally. The established portal contract preserves the original encoded bytes; typed vectors/JSON values would deserialize and re-encode them.

crates/observability/migrations/2026-09-15-000005_create_alert_dictionary/up.sql
```diff
@@ -0,0 +1,12 @@
+CREATE TABLE alert_dictionary (
```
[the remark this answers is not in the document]
ThisIsMani: Valid migration concern, but legacy backfill/cutover is intentionally outside this PR contract. This stack keeps the new replacement-state table and does not add rollout scripts.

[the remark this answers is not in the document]
ThisIsMani: Declined: this PR introduces the dedicated mutable-state dictionary contract; migrating legacy `alerts_dicts` data is outside this stack’s scope, and backfill scripts are explicitly excluded from this rollout.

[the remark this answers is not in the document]
ThisIsMani: Declined for this PR: backfilling or coupling this API to the legacy ClickHouse-backed dictionary is outside the mutable-state port scope. This change preserves the established `key_`/`values_` wire contract; migration/backfill rollout work is intentionally not added here.

ThisIsMani:
Restored the established dictionary wire contract: requests and responses now use `key_` and `values_` while retaining idiomatic internal Rust/storage names. Route coverage verifies the exact fields and rejects the incompatible `key`/`values` spellings.

PR #14273, opened 2026-09-16
fix(ucs): shared keepalive gRPC channel, fatal build errors, real health check, transport failure class

gopikrishna000:
Closes juspay/hyperswitch-cloud#23600

## Problem

Payments confirmed through UCS intermittently fail with `UCS error: Unknown - transport error` (`error_type: ucs_call_failed`) in 0-2 ms, and the merchant receives HTTP 500 on PaymentsConfirm. UCS has no trace of the request; it never left the router.

Production evidence (us2, 2026-09-10 to 2026-09-16): 7 failures, all with this signature, all first RPC on a channel that had been idle for hours (AccessTokenAuth 4 of 6 calls failed; RSync after 2.4 days idle; PSync twice). On 2026-09-14 the same merchant failed at 07:44:49 and 07:44:59 and succeeded at 07:45:01 with nothing else changing. On 2026-09-16 12:35 the failing channel had last been used before the UCS rollout at 10:05.

## Root cause

The router opens 11 per-service tonic channels once at startup with no HTTP/2 or TCP keepalive. When a channel's connection dies while idle (peer rollout, or node connection state expiring), nothing on the client notices until the next RPC writes into it. tonic's reconnect layer only replaces a connection that reports itself closed at `poll_ready`; a failure on the write itself is returned to that RPC. Rarely used channels (merchant authentication, refund) are therefore almost guaranteed to fail their first call after any connection death.

A second, related defect: if any of the 11 eager connects failed or timed out at boot, the whole client became `None` and the pod silently served every payment via the direct connector path for its lifetime, which is not a valid fallback for `ucs_only_connectors`.

## Changes

- One shared `Channel` for all UCS service clients, built with `connect_lazy`, `http2_keep_alive_interval` (30s), `keep_alive_timeout` (20s, the gRPC-Go default), `keep_alive_while_idle(true)`, `tcp_keepalive` (60s), `connect_timeout` and the existing per-RPC `timeout`. A dead connection is detected by PING and rebuilt in the background before any payment RPC uses it; no client-specific channel can sit idle. Flow-control windows are left at hyper's defaults of 5 MiB per connection and 2 MiB per stream, which is the same connection credit each of the eleven channels had, so sharing one connection is not a throughput regression. `http2_adaptive_window` is deliberately not enabled: it lowers the initial windows to 65,535 bytes and grows them back from BDP samples, which helps large transfers on high-latency links and hurts same-region RPCs of a few KiB.
- Keepalive and TCP settings are wrapper types with `Default` impls (`UcsKeepAliveIntervalInSeconds`, `UcsKeepAliveTimeoutInSeconds`, `UcsTcpKeepAliveInSeconds`) and are documented in the example, development and env_specific configs.
- `UnifiedConnectorServiceClient::build_connections` returns `Result<Option<Self>, UcsClientBuildError>`. Configured but unbuildable UCS now fails startup instead of degrading silently.
- Real gRPC `grpc.health.v1/Check` against UCS in `health_check_unified_connector_service` (previously only checked that the client existed). Intended for use by a startup probe, not readiness or liveness: a UCS outage at runtime must not remove router pods from service.
- Diagnostics: `UnifiedConnectorServiceError::TonicStatus` carries an optional `UcsTransportFailure` when the status was produced by the local transport. It records the `std::error::Error::source()` chain verbatim, with no classification or text matching, so a failure mode we have not seen yet still describes itself. The root cause alone goes on the connector event `error` JSON as `transport_error`; the full chain goes to the router log as `transport_error_chain`, since an inner layer may render the UCS authority or a peer address and the event is merchant-visible.
- Keepalive values are configurable via `grpc_client.unified_connector_service.{keep_alive_interval_secs,keep_alive_timeout_secs,tcp_keepalive_secs}` with defaults; no config change is required.

## Not in this PR

- No retry of any kind. A transparent retry for the "request never left the client" class is a possible follow-up once the new diagnostics confirm the failure class distribution in production.
- Server side (UCS): `http2_keepalive_interval` and `http2_keepalive_timeout` in juspay/hyperswitch-prism#2313, plus a `preStop` drain in the Helm chart. `max_connection_age` was considered and rejected: tonic retires connections on an unjittered timer, which would synchronise every router pod onto the same boundary and add thousands of daily GOAWAY events to buy load spreading that is not needed at current replica counts.
- Infra: point the router startup probe at the deep health endpoint so a new pod that cannot reach UCS never starts.

## Verification

Built and run locally alongside UCS; CI is the authority. Expected observable effect: `ucs_call_failed` rows with `error` containing `transport error` drop to zero for idle channels; any remaining occurrence carries `transport_error` on the event and a logged `transport_error_chain` naming the layer that failed.


XyneSpaces:
[should-fix] `GrpcClientSettings::get_grpc_client_interface` now `.expect`s `UnifiedConnectorServiceClient::build_connections`, making a UCS DNS/URI/config failure fatal at router startup (`crates/external_services/src/grpc_client.rs:138-140`).

If a deployment does not run UCS or has a transient misconfiguration, the router pod will crash-loop instead of degrading to direct connectors. The `unified_connector_service_client` field is still `Option<UnifiedConnectorServiceClient>`, so either keep the optional behavior and log the error, or remove the `Option` and document UCS as mandatory. Do not leave a type that allows `None` while the constructor panics before it can ever be `None`.

```rust
// Option A: preserve optional degradation
let unified_connector_service_client =
    UnifiedConnectorServiceClient::build_connections(self)
        .await
        .inspect_err(|error| logger::error!(?error, "failed to build UCS client; UCS flows will be unavailable"))
        .ok();

// Option B: make it mandatory and change GrpcClients field to non-Option
```

XyneSpaces:
[blocking] `.expect("Failed to build the Unified Connector Service client from configuration")` at `crates/external_services/src/grpc_client.rs:138` will panic at startup for any config/build failure.

Replace the `.expect()` with typed error propagation (e.g., `?`) so connection failures are surfaced instead of crashing the process.

```suggestion
        let unified_connector_service_client =
            UnifiedConnectorServiceClient::build_connections(self)
                .await?;
```

XyneSpaces:
[blocking] `UnifiedConnectorServiceClient::build_connections` now returns a structured `Result`, but the caller in `crates/external_services/src/grpc_client.rs` unwraps it with `.expect("Failed to build the Unified Connector Service client from configuration")`. Panicking at startup masks the underlying `UcsClientBuildError` and breaks baseline rules against `.expect` in production code.

Propagate the error up to the startup path and exit cleanly with the error attached, or convert it to a startup failure log/event instead of panicking.

JeevaRamu0104 (dismissed):

XyneSpaces:
[blocking] Startup code introduces `.expect()` in a non-test path

`crates/external_services/src/grpc_client.rs:139` adds `.expect("Failed to build the Unified Connector Service client from configuration")` during service initialization. The baseline prohibits panicking `.expect()` outside tests; startup misconfiguration should surface as a structured error so the process exits with diagnostics rather than a thread panic. Replace the `.expect()` with `CustomResult`-style propagation or a graceful fatal log-and-exit.

[should-fix] Transport-failure classification relies on substring heuristics

`crates/external_services/src/grpc_client/unified_connector_service.rs:302` classifies `UcsTransportFailureClass` by ASCII-lowercasing the entire error-chain string and matching substrings. This is fragile and ignores the actual `io::ErrorKind` / hyper error enums. Classify from typed error values where possible.

[should-fix] New error variant may break serialization

`crates/hyperswitch_interfaces/src/unified_connector_service/transformers.rs:1938` adds `transport: Option<Box<UcsTransportFailure>>` to a public error enum variant. Verify the variant still derives/serializes correctly for events and OpenAPI, since boxed optional fields can require derive adjustments.

crates/external_services/src/grpc_client/unified_connector_service.rs
```diff
@@ -75,6 +80,19 @@ pub struct UnifiedConnectorServiceClientConfig {
...
+
+    /// Time (seconds) to wait for a keepalive PING acknowledgement before the connection is
+    /// treated as dead and re-established.
+    #[serde(default = "default_ucs_keep_alive_timeout_secs")]
+    pub keep_alive_timeout_secs: u64,
+
+    /// TCP keepalive idle time (seconds) on the shared channel.
+    #[serde(default = "default_ucs_tcp_keepalive_secs")]
+    pub tcp_keepalive_secs: u64,
+
```
hrithikesh026, line 95: Lets use wrapper types instead of having hanging default functions. Those wrapper types can have Default trait implementation on them.

hrithikesh026:
wouldn't a single connection throttle the bandwidth which 11 separate connections previously provided??

crates/external_services/src/grpc_client/unified_connector_service.rs
hrithikesh026, line 95: Also add this in example config file.

gopikrishna000:
Good question, and checking it properly turned up a bug in my own change, so thank you. Short answer: no, and I have removed the `http2_adaptive_window(true)` line that my original comment used to justify itself, because that line was the only part of this that *could* have throttled us.

**Why one connection is not a bandwidth cap**

HTTP/2 flow-control windows are receive-side credit, not a rate limit. They only bound how much unacknowledged data may be in flight before the receiver must send `WINDOW_UPDATE`. Throughput is capped only when the window is smaller than the bandwidth-delay product of the link.

My original code comment claimed each of the eleven channels had "its own default 64 KiB connection window". That was wrong, and I have corrected it. 65,535 bytes is the HTTP/2 *spec* default, but neither tonic nor hyper leaves it there. In tonic 0.14.6 `init_stream_window_size`, `init_connection_window_size` and `http2_adaptive_window` are all `None` on the endpoint, so hyper's own defaults apply (`hyper-1.8.1/src/proto/h2/client.rs`):

```rust
const DEFAULT_CONN_WINDOW: u32 = 1024 * 1024 * 5; // 5mb
const DEFAULT_STREAM_WINDOW: u32 = 1024 * 1024 * 2; // 2mb
```

So the real comparison is 5 MiB of connection credit before, and 5 MiB after. It is the same number. We were never getting 11x anything: each channel held its own 5 MiB window, but a single UCS call only ever used one of them, because a given RPC belongs to exactly one service client. Eleven windows did not add up for any individual request.

Against our payloads this is not close. UCS gRPC messages are a few KiB. The 2 MiB per-stream window alone is roughly two orders of magnitude more credit than a payment request or response consumes, so a request is never window-bound in the first place, on one connection or eleven.

**The part that was genuinely wrong**

Enabling adaptive window does not raise the ceiling, it lowers the floor. From `hyper-1.8.1/src/client/conn/http2.rs`:

```rust
pub fn adaptive_window(&mut self, enabled: bool) -> &mut Self {
    self.h2_builder.adaptive_window = enabled;
    if enabled {
        self.h2_builder.initial_conn_window_size = SPEC_WINDOW_SIZE;   // 65_535
        self.h2_builder.initial_stream_window_size = SPEC_WINDOW_SIZE; // 65_535
    }
    self
}
```

It drops both windows from 5 MiB and 2 MiB to 64 KiB and then relies on BDP sampling from PING RTT to grow them back, up to a 16 MiB `BDP_LIMIT`. That is a real win for large transfers over high-latency links. For same-region RPCs of a few KiB it is strictly worse: it starts us 32x lower than the default and the samples that would grow it back never arrive, because the payloads are too small to measure bandwidth with. So I have dropped it and left the windows at hyper's defaults.

**What actually would bind, if anything**

Concurrency, not bandwidth, and it does not bind either. The limit is `SETTINGS_MAX_CONCURRENT_STREAMS`, advertised by the UCS server. UCS does not set `max_concurrent_streams`, so the h2 default applies and it is not a limit we are near: each HS pod carries a modest number of in-flight UCS calls, and each is one short-lived stream. If we ever did approach it, the answer is an explicit limit plus more than one connection, not eleven connections split by service, which distributed streams by service name rather than by load.

There is a genuine tradeoff in sharing, but it is availability, not bandwidth: one connection means one shared fate, so a connection-level error affects every UCS service at once instead of one. That is the trade this PR intentionally makes, because the failure we are fixing came from the eleven-channel arrangement itself. Idle per-service channels died unnoticed and each cost a payment on the next call. A single connection that keepalive keeps warm and continuously exercised is far less likely to be dead when a payment needs it.

crates/external_services/src/grpc_client/unified_connector_service.rs
gopikrishna000: Done in 20bc35e. The three `u64` fields are now wrapper types with `Default` impls, matching `UcsRequestTimeoutInSeconds` already in this file, and the three hanging `default_ucs_*` functions are gone.

```rust
pub keep_alive_interval: UcsKeepAliveIntervalInSeconds,
pub keep_alive_timeout: UcsKeepAliveTimeoutInSeconds,
pub tcp_keepalive: UcsTcpKeepAliveInSeconds,
```

Each exposes `as_duration()`, so the channel builder reads `config.keep_alive_interval.as_duration()` and no longer constructs `Duration::from_secs` at the call site. The config keys lose their `_secs` suffix as a result, which also matches `connection_timeout` and `request_timeout`.

gopikrishna000: Added in 20bc35e to `config/config.example.toml`, `config/development.toml` and `config/deployments/env_specific.toml`, alongside the existing `connection_timeout` and `request_timeout`:

```toml
keep_alive_interval = 30                # HTTP/2 PING keepalive interval in Seconds on the shared channel
keep_alive_timeout = 20                 # Time in Seconds to wait for a keepalive PING acknowledgement before the connection is treated as dead
tcp_keepalive = 60                      # TCP keepalive idle time in Seconds on the shared channel
```

I left `production.toml`, `sandbox.toml` and `integration_test.toml` alone, since their `[grpc_client.unified_connector_service]` blocks carry only the connector lists and already inherit `base_url`, `connection_timeout` and `request_timeout` from env_specific. The `Default` impls mean those deployments pick up the same values without needing the keys.

gopikrishna000:
> wouldn't a single connection throttle the bandwidth which 11 separate connections previously provided??

No. The unit of concurrency in HTTP/2 is the stream, not the connection. Every call site clones the client before the RPC, so two concurrent requests on the same payment_service_client open two independent streams and dispatch in parallel. The tower Buffer in front of the connection hands the response future straight back to the caller instead of awaiting it, so nothing queues behind an in-flight request.

hrithikesh026 (approved):

JeevaRamu0104 (approved):

JeevaRamu0104 (approved):

hrithikesh026 (approved):

PR #14289, opened 2026-09-17
fix(refunds): reject refund once max_attempts refunds already exist

### LINKED ISSUE #14264: [BUG] [Refunds]: refund.max_attempts is off by one; default of 10 permits 11 refunds
### Bug Description

`refund.max_attempts` is enforced with a strict `>` against the count of refunds that already exist, and the check runs before the new refund is inserted. The effect is an off-by-one: with the default `max_attempts: 10`, a payment attempt can end up with 11 refunds.

The check, in [`refunds_validator.rs#L93-L100`](https://github.com/juspay/hyperswitch/blob/184ffd4c015fd3fea2f3868549f1a86ffa5f40da/crates/router/src/core/utils/refunds_validator.rs#L93-L100):

```rust
pub fn validate_maximum_refund_against_payment_attempt(
    all_refunds: &[diesel_refund::Refund],
    refund_max_attempts: usize,
) -> CustomResult<(), RefundValidationError> {
    utils::when(all_refunds.len() > refund_max_attempts, || {
        Err(report!(RefundValidationError::MaxRefundCountReached))
    })
}
```

In `validate_and_create_refund` ([`refunds.rs`](https://github.com/juspay/hyperswitch/blob/184ffd4c015fd3fea2f3868549f1a86ffa5f40da/crates/router/src/core/refunds.rs#L1599)), `all_refunds` is fetched at L1644 for the payment attempt's `connector_transaction_id`, the count check runs at L1676, and the new refund is built and inserted afterwards at L1705 and L1743. So when 10 refunds already exist, `10 > 10` is false, the check passes, and an 11th refund is created. The 12th is rejected.

The default is `max_attempts: 10` ([`defaults.rs#L92-L97`](https://github.com/juspay/hyperswitch/blob/184ffd4c015fd3fea2f3868549f1a86ffa5f40da/crates/router/src/configs/defaults.rs#L92-L97)).

### Expected Behavior

`max_attempts: 10` allows at most 10 refunds per payment attempt. The comparison should be `all_refunds.len() >= refund_max_attempts`, since the check runs before the insert.

### Actual Behavior

11 refunds can be created on one payment attempt with the default setting: the configured value, plus one.

### Steps To Reproduce

1. Create and capture a payment with an amount large enough to split into many partial refunds (for example 1200 in the minor unit).
2. Create refunds of 100 each against that payment, one at a time.
3. The 11th refund is accepted.
4. The 12th fails with `MaximumRefundCount`.

Note the amount check (`validate_refund_amount`, L1671) usually binds first, so the captured amount must be large enough to leave room for 11 refunds. Failed refunds also count toward the attempt total even though they do not consume the refundable amount.

### Context For The Bug

Found while documenting refund semantics for the docs (juspay/hyperswitch-docs#270). The docs page currently has to describe the actual behavior ("the default of 10 allows up to 11 refunds"), which reads as a bug rather than an intended limit.

This is code inspection, not an executed test.

### Environment

Code inspection at `main` (`184ffd4c015fd3fea2f3868549f1a86ffa5f40da`); also present at `02ba5ea0b4cc0ad8debea4cefd28a7de5916d9db`.

errmakov:
## Type of Change

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

`refund.max_attempts` let one more refund through than configured. `validate_maximum_refund_against_payment_attempt` (`crates/router/src/core/utils/refunds_validator.rs`) receives the refunds that already exist for the payment attempt, fetched before the new refund is inserted, and rejected only when `all_refunds.len() > refund_max_attempts`. With 10 refunds already present, `10 > 10` is false, so an 11th refund was created with the default `max_attempts = 10`.

This PR:

- Changes the check to `existing_refund_count >= refund_max_attempts`, so at most `max_attempts` refunds exist per payment attempt. The function is shared by the v1 (`core/refunds.rs`) and v2 (`core/refunds_v2.rs`) refund flows, so both are fixed.
- Takes the count (`existing_refund_count: usize`) instead of the refund slice, since only the length was ever used. Both call sites pass `all_refunds.len()`, and the boundary cases can be unit tested without building full refund records.
- Adds a doc comment stating that the refunds are counted before the new one is inserted, and that the count is read outside of any lock.

### Behaviour change: `max_attempts = 0`

Under the old `>` comparison, `max_attempts = 0` never rejected anything (`0 > 0` is false), so it effectively meant *unlimited refunds*. Under `>=` it means *no refunds at all*, which is the literal reading of the setting. The shipped default is `10` (`crates/router/src/configs/defaults.rs`, `config/config.example.toml`), and a config that omits `[refund] max_attempts` still gets `10` rather than `0`, so this only affects a deployment that explicitly set `max_attempts = 0`.

### Not addressed here

The count is read before the new refund is inserted and nothing serialises the two, so concurrent refund requests on the same payment attempt can each observe the same count and together exceed `max_attempts`. That race predates this PR and is left as is; the doc comment now says so.

### Additional Changes

The config value and its default are unchanged; it is now enforced as documented (`max_attempts = 10 # Number of refund attempts allowed` in `config/config.example.toml`). A payment attempt that previously accepted `max_attempts + 1` refunds now gets `400 IR_12` ("Reached maximum refund attempts") on that last one.

## Motivation and Context

Fixes #14264

## How did you test it?

**Unit tests** (`core::utils::refunds_validator::tests`, 5 new), with `max_attempts = 10` unless stated:

| Test | Existing refunds | Expected | Before fix | After fix |
|---|---|---|---|---|
| `allows_first_refund` | 0 | allowed | pass | pass |
| `allows_last_refund_within_limit` | 9 | allowed | pass | pass |
| `rejects_refund_when_limit_already_reached` | 10 | `MaxRefundCountReached` | **fail** | pass |
| `rejects_refund_when_limit_exceeded` | 11 | `MaxRefundCountReached` | pass | pass |
| `rejects_every_refund_when_max_attempts_is_zero` | 0, `max_attempts = 0` | `MaxRefundCountReached` | **fail** | pass |

```
cargo test -p router --lib core::utils::refunds_validator::tests
test result: ok. 5 passed; 0 failed; 0 ignored; 0 measured; 68 filtered out
```

**End to end**, against a locally running router (`config/development.toml`, default `refund.max_attempts = 10`, Postgres 17 + Redis 7, all migrations applied), using the built-in `stripe_test` dummy connector. The same script was run against a build of `main` and a build of this branch: create a merchant, API key and connector, make one card payment of 1200, then call `POST /refunds` with an amount of 100, twelve times.

| Refund | `main` | This PR |
|---|---|---|
| 1 – 10 | 200 `succeeded` | 200 `succeeded` |
| 11 | **200 `succeeded`** | **400 `IR_12`** Reached maximum refund attempts |
| 12 | 400 `IR_12` | 400 `IR_12` |
| Refund rows for the payment (`select count(*) from refund where payment_id = …`) | **11** | **10** |

I did not build the v2 feature set locally. The v2 call site is the same one-line change as v1 (`all_refunds.len()`), so no further change was needed there; CI covers the v2 build.

## Checklist


abhisheksharma2411 (commented):
The off-by-one is real and the fix is the right one. `all_refunds` is fetched before the new refund is inserted on **both** paths — `refunds.rs:1644` and `refunds_v2.rs:1180`, both validating ahead of the insert and both reading `state.conf.refund.max_attempts` — so `>=` is correct and consistent across v1 and v2. Nine existing plus the new one lands exactly on the configured ten. The tests pin both boundaries, which is more than most fixes of this size get.

I went looking for a regression and didn't find the one I expected, so for the record: `Refund` carries `#[serde(default)]` without a derived `Default`, which had me worried that a config omitting `[refund] max_attempts` would deserialize to `0` and — under the new `>=` — reject every refund in the deployment. It doesn't. `configs/defaults.rs:92` hand-implements `Default` with `max_attempts: 10`, so a missing key still means ten. No issue there.

That does leave one thing worth stating deliberately rather than leaving in a test:

**`max_attempts = 0` changes meaning.** Before, `0 > 0` was false, so an explicit `0` meant *unlimited refunds*. Now it means *no refunds ever*. Your `rejects_every_refund_when_max_attempts_is_zero` test says you thought about it and chose the literal reading, which I agree is the better semantic — "zero attempts allowed" should mean zero. But it's a silent behaviour flip for anyone who set `0` intending "off", and it belongs in the PR description and the changelog rather than only in a test name. `config.example.toml` calls it "Number of refund attempts allowed", which supports your reading.

**One thing I'd push back on: the `<T>` generic.**

```rust
pub fn validate_maximum_refund_against_payment_attempt<T>(
    all_refunds: &[T],
```

This loosens a signature purely for test convenience — it exists so the tests can pass `[(); MAX_ATTEMPTS]`. The cost is that the function now accepts a slice of *anything*, so a future call site that passes the wrong collection (payment attempts, captures, disputes) compiles silently where `&[diesel_refund::Refund]` would have caught it. On a money path I'd rather keep that.

Since the body only ever reads `.len()`, taking the count directly keeps both the type safety and the testability:

```rust
pub fn validate_maximum_refund_against_payment_attempt(
    existing_refund_count: usize,
    refund_max_attempts: usize,
) -> CustomResult<(), RefundValidationError> {
    utils::when(existing_refund_count >= refund_max_attempts, || {
        Err(report!(RefundValidationError::MaxRefundCountReached))
    })
}
```

Both call sites become `validate_maximum_refund_against_payment_attempt(all_refunds.len(), ...)`, the tests get simpler still (no placeholder arrays at all), and the parameter name states the precondition your new doc comment currently has to explain. Your call — if you'd rather keep the slice, the generic is a reasonable trade and I won't hold the PR on it.

**Out of scope, but worth knowing:** this check is advisory. Two concurrent refund requests both read nine existing rows, both pass, and eleven end up stored — there's no unique constraint or row lock making the count authoritative. Your change doesn't make that worse and fixing it isn't this PR's job, but "at most `max_attempts` refunds exist per payment attempt" is true under serial traffic rather than in general, and the doc comment currently states it unconditionally. Softening that sentence would keep the next reader from trusting it too far.

Nothing blocking from me beyond the `<T>` question.

errmakov:
Thanks for the careful read — I took all three suggestions.

**The generic.** You're right, and I've dropped it. The function now takes `existing_refund_count: usize` and both call sites pass `all_refunds.len()`. Your point about a future call site passing the wrong slice is the one that convinced me: the generic bought test convenience at the cost of the type check, whereas the count keeps the check *and* makes the tests simpler. The parameter name also says what the doc comment previously had to explain.

**`max_attempts = 0`.** Agreed it shouldn't live only in a test name. The PR description now has a "Behaviour change" section spelling out that `0` used to mean unlimited (`0 > 0` false) and now means none, that the default is `10`, and that an omitted key still deserializes to `10` — so only a deployment that explicitly set `0` is affected. The commit message carries the same note.

**The advisory check.** Good catch, and out of scope here as you say. I softened the doc comment to state that the count is read outside of any lock, so concurrent refund requests may each observe the same count and exceed the limit, and added a "Not addressed here" section to the description pointing at the same race. Serialising it needs a lock or a DB-level constraint, which is a bigger change than this off-by-one.

Also thanks for checking the missing-`[refund]`-section case — I had verified `configs/defaults.rs` gives `max_attempts: 10`, but hadn't thought to frame it as a `>=`-specific regression risk, and it's the right thing to have ruled out.

Re-ran the unit tests after the change: 5 passed, 0 failed.

PR #14292, opened 2026-09-17
fix(connector): [COINBASE] stop declaring manual capture and refunds webhook flow

### LINKED ISSUE #14277: [BUG] [Coinbase]: declares manual capture and a refunds webhook flow that the connector does not implement
### Bug Description

The Coinbase connector declares capabilities its flows do not implement. The declarations are what `GET /feature_matrix` reports, so the generated capability tables (including the connector docs page) advertise capture methods and a webhook flow that cannot work.

Two mismatches, both at [`crates/hyperswitch_connectors/src/connectors/coinbase.rs`](https://github.com/juspay/hyperswitch/blob/5fb7e5598eadd8ed5fa42822427107f271a1e112/crates/hyperswitch_connectors/src/connectors/coinbase.rs):

**1. Capture methods.** The supported payment methods declare three capture methods (L475-478):

```rust
let supported_capture_methods = vec![
    enums::CaptureMethod::Automatic,
    enums::CaptureMethod::Manual,
    enums::CaptureMethod::SequentialAutomatic,
];
```

But the capture flow is a stub (L349-361):

```rust
impl ConnectorIntegration<Capture, PaymentsCaptureData, PaymentsResponseData> for Coinbase {
    fn build_request(...) -> ... {
        Err(errors::ConnectorError::FlowNotSupported {
            flow: "Capture".to_string(),
            connector: "Coinbase".to_string(),
        }.into())
    }
}
```

A payment created with `capture_method: manual` or `sequential_automatic` fails at capture time with `FlowNotSupported`. Only automatic capture works, since it needs no capture call.

**2. Refund webhook flow.** The declared webhook flows include refunds (L496-497):

```rust
static ref COINBASE_SUPPORTED_WEBHOOK_FLOWS: Vec<enums::EventClass> =
    vec![enums::EventClass::Payments, enums::EventClass::Refunds,];
```

There are no refunds on this connector to report on. `ConnectorIntegration<Execute>` returns `FlowNotSupported` (L365-377), and the payment method declaration itself sets `refunds: FeatureStatus::NotSupported` (L487). `get_webhook_event_type` (L427-448) recognizes only `charge:confirmed`, `charge:resolved`, `charge:failed`, `charge:pending` and `charge:created`; no refund event is mapped.

### Expected Behavior

Declarations match the implementation:

1. `supported_capture_methods` for Coinbase lists only `Automatic`.
2. `COINBASE_SUPPORTED_WEBHOOK_FLOWS` lists only `EventClass::Payments`.

Or, if manual capture and refunds are intended for Coinbase, the flows are implemented and the declarations stay.

### Actual Behavior

`/feature_matrix` reports manual and sequential automatic capture, and a refunds webhook flow, for a connector that implements neither. Generated capability tables repeat both.

### Steps To Reproduce

1. Call `GET /feature_matrix` and read the `COINBASE` entry: card capture methods include `manual` and `sequential_automatic`, and the webhook flows include refunds.
2. Create a Coinbase payment with `capture_method: manual`, authorize it, then call capture. It fails with `FlowNotSupported`.
3. Note that no refund can exist on the connector to produce a refund webhook: refunds are declared `NotSupported` and the refund execute flow is a stub.

### Context For The Bug

Found while regenerating connector capability blocks for the docs (juspay/hyperswitch-docs#274). The docs page carries two sentences correcting the generated table, which is the wrong end to fix it: the page should be able to state what the connector declares without contradicting it.

Same shape as juspay/hyperswitch#14242 (Bamboraapac declares `SequentialAutomatic` and rejects it).

This is code inspection, not an executed test.

### Environment

Code inspection at `main` (`5fb7e5598eadd8ed5fa42822427107f271a1e112`); also present at `184ffd4c015fd3fea2f3868549f1a86ffa5f40da`.

errmakov:
## Type of Change

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

The Coinbase connector specification declared capabilities that the connector does not implement, so `GET /feature_matrix` (and the connector docs generated from it) advertised them:

| Declaration | Before | After | Why |
|---|---|---|---|
| `supported_capture_methods` | `Automatic`, `Manual`, `SequentialAutomatic` | `Automatic`, `SequentialAutomatic` | `ConnectorIntegration<Capture>` returns `FlowNotSupported`, so a manual capture can never be completed |
| `COINBASE_SUPPORTED_WEBHOOK_FLOWS` | `Payments`, `Refunds` | `Payments` | Refunds are declared `NotSupported`, `Execute` returns `FlowNotSupported`, and `get_webhook_event_type` maps no refund event |

Because the default `ConnectorValidation::validate_connector_against_payment_request` checks the requested capture method against `supported_capture_methods`, a Coinbase payment with `capture_method: manual` is now rejected up front with `NotSupported`, instead of being accepted and failing at capture time.

**Why `SequentialAutomatic` is kept** (the issue suggests removing it): `should_initiate_capture_flow` only chains a capture call when the attempt is `Authorized`, and only for Paybox. Coinbase's status mapping never produces `Authorized` (`New` → `AuthenticationPending`, `Completed`/`Resolved` → `Charged`), so for Coinbase `SequentialAutomatic` behaves exactly like `Automatic` and works today. Removing it would start rejecting requests that currently succeed. This matches `cryptopay`, which declares the same pair.

### Additional Changes

## Motivation and Context

Fixes #14277.

## How did you test it?

Added unit tests in `crates/hyperswitch_connectors/src/connectors/coinbase.rs` that call `validate_connector_against_payment_request` for each capture method and check `get_supported_webhook_flows`.

```
cargo test -p hyperswitch_connectors --features v1,frm,payouts,revenue_recovery,dummy_connector --lib connectors::coinbase::tests
```

**Tests against the old declarations** (tests only):

```
test connectors::coinbase::tests::accepts_automatic_and_sequential_automatic_capture ... ok
test connectors::coinbase::tests::declares_only_payments_webhook_flow ... FAILED
test connectors::coinbase::tests::rejects_manual_capture_because_capture_flow_is_not_supported ... FAILED
  left: Some([Payments, Refunds])
 right: Some([Payments])
assertion failed: !validate_capture_method(CaptureMethod::Manual)
test result: FAILED. 1 passed; 2 failed; 0 ignored; 0 measured; 10 filtered out
```

**With the fix:**

```
test connectors::coinbase::tests::accepts_automatic_and_sequential_automatic_capture ... ok
test connectors::coinbase::tests::declares_only_payments_webhook_flow ... ok
test connectors::coinbase::tests::rejects_manual_capture_because_capture_flow_is_not_supported ... ok
test result: ok. 3 passed; 0 failed; 0 ignored; 0 measured; 10 filtered out
```

Not tested against a live Coinbase account: the change only affects declarations and request validation, and no Coinbase API call changes.

## Checklist


abhisheksharma2411 (commented):
Verified both removals against the connector rather than the description, and both are correct — this is removing capabilities the code cannot actually perform.

**Manual capture.** `impl ConnectorIntegration<Capture, …> for Coinbase` at L349 has a `build_request` that returns `Err(FlowNotSupported)` outright. So a merchant configuring manual capture on Coinbase got an accepted configuration and a runtime failure at capture time — after the customer had already paid. Moving that rejection to request validation is the right trade: in payments, failing at configuration beats failing mid-transaction.

Keeping `SequentialAutomatic` with the reasoning spelled out (a charge never reports `Authorized`, so no follow-up capture is ever attempted) is the detail that makes this safe to merge — that's the one that would have been easy to remove along with `Manual` and quietly break working integrations.

**Refunds webhook.** Also correct, and doubly so:

```
ConnectorIntegration<Execute, RefundsData, …>::build_request  ->  Err(FlowNotSupported)
get_webhook_event_type  ->  PaymentIntentSuccess | PaymentActionRequired
                            | PaymentIntentProcessing | EventNotSupported
```

The refund *flow* is unimplemented, and the webhook mapper cannot emit a single refund event — there is no `IncomingWebhookEvent::Refund*` anywhere in the connector. So `EventClass::Refunds` was advertising a webhook class that can never fire.

**This looks like a small pattern rather than a one-off.** Scanning for connectors whose `Capture` impl returns `FlowNotSupported` *and* which still name `CaptureMethod::Manual`:

```
cashtocode
coinbase     <- this PR
tokenio
zen
```

`cashtocode` has `Manual` sitting in its `supported_capture_methods` vector next to `Automatic` and `SequentialAutomatic`, same shape as Coinbase before this change. I haven't checked whether each of those three has the same runtime outcome — a couple may route capture differently — but they're worth a look, and the `validate_capture_method` test added here is exactly the guard that would pin them once checked. Happy to file an issue listing them if that's useful rather than expanding this PR.

The test is the right shape: asserting through `validate_connector_against_payment_request` rather than reading the constant means a future edit to the supported list has to confront it.

PR #14294, opened 2026-09-17
fix: preserve payment authentication results and Worldpay ECI

### LINKED ISSUE #14293: [BUG] Payment responses lose original 3DS authentication results
External authentication exposes only a summary status: EMV `A` (attempted) and `N` (failed) both appear as `failed`. Worldpay XML also parses `ThreeDSecureResult` but drops it from card, Apple Pay and Google Pay responses.

Expected: preserve the original authentication status/result in payment confirm and retrieve responses, including later polling updates and responses without an authorization code.

Longsueynin:
Payment responses lose authentication information: external 3DS collapses distinct EMV statuses into a summary, and Worldpay XML drops the authentication result and ECI returned for cards, Apple Pay and Google Pay. Preserve that information across confirm, retrieve, and subsequent polling updates, including responses without an authorization code.

- Expose the stored EMV status as `external_authentication_details.trans_status`, retaining the existing summary status.
- Return Worldpay's result description and ECI as `authentication_data.three_d_secure_result` and `authentication_data.eci` under `payment_method_data.card`, `payment_method_data.wallet.apple_pay`, or `payment_method_data.wallet.google_pay`.
- Read ECI from `ThreeDSecureResult/eci`, falling back to `payment/eci`. Preserve leading zeros. An ECI-only response produces only `{"eci":"…"}`; it does not synthesize a successful 3DS result, protocol version, or transaction ID.
- Preserve previously reported authentication fields when a later connector response omits them; newly reported fields take precedence.
- Mask the new wallet authentication fields with `Secret<serde_json::Value>` in diagnostic output. Only outcome metadata is mapped; no cryptograms or authentication credentials are exposed.

This includes Google Pay PAN_ONLY and wallet authentication results returned without a separate Hyperswitch external authentication flow. `external_authentication_details` may still be null in that case: connector-reported wallet authentication data does not imply a separate EMV 3DS flow occurred. The change does not initiate or alter wallet/3DS authentication.

The API additions are optional and require no database migration. Existing wallet records remain readable; wallet response shapes gain `authentication_data: null` when no result is available.

Closes #14293.

Validation: six focused tests pass on the branch rebased onto `origin/main` (`00fe936154`): three Worldpay XML mapping tests, the router storage/polling regression, wallet response serialization and diagnostic masking, and preservation of the original EMV status. Ran `cargo test --locked -p router -p hyperswitch_connectors -p api_models --lib` with filters `three_ds`, `stored_wallet_authentication_is_exposed_in_payment_response`, and `external_authentication_preserves_emv_trans_status`. `cargo +nightly fmt --all --check` and `git diff --check` also pass.


abhisheksharma2411 (commented):
The regression this fixes is the kind that only shows up in production — an authentication result that exists at confirm and is gone by the time anyone reads the payment back. The test is the right shape for it:

```rust
// Sync/capture can report an auth code without repeating the 3DS result.
let connector = json!({method: {"auth_code": "123456"}});
let updated = update_additional_payment_data_with_connector_response_pm_data(Some(stored), Some(connector))
```

asserting the stored `authentication_data` survives a *later* connector response that doesn't mention it, across Card, Apple Pay and Google Pay. That's exactly the failure mode — the second response wins and silently drops a field it never carried — and covering all three methods matters because they store under different pointers.

`if auth_code.is_none() && authentication_data.is_none()` returning early is a good detail too: an authentication failure legitimately has no auth code, so the alternative would have been emitting an empty container on every such response.

**I checked what actually lands in the field, since it's untyped**, and on this connector it's narrow:

```rust
let authentication_data = payment_data.three_d_secure_result
    .map(|description| serde_json::json!({ "three_d_secure_result": description }));
```

— a human-readable description (`"Authentication offered but not used"`, `"Authentication unavailable"`, `"Authentication rejected"`). No CAVV, no ECI, no DS transaction id. So nothing sensitive is being newly exposed here, which was my first question given this ends up on the payment response.

**That's worth writing down somewhere, because the type won't enforce it.** `authentication_data: Option<serde_json::Value>` on a response-exposed struct is an open container, and the next connector to implement this will copy Worldpay's shape as the precedent. A 3DS result is exactly the place where a cardholder authentication value could plausibly get added "because the connector returned it" — and CAVV/ECI on a payment-retrieve response is a different conversation from a status description. A doc comment on the field saying what belongs in it (connector-reported authentication *outcome*, not authentication *credentials*) costs one line and outlives the reviewer.

**One wire-format confirmation worth making explicit in the PR body:** `WalletAdditionalDataForCard` carries no `skip_serializing_if` on any field, so the new key will appear as `"authentication_data": null` on every response for that struct, not only when populated. That's consistent with its existing siblings (`auth_code`, `issuer_name`, `issuer_country` all behave the same), so it's house style rather than a new problem — but it is a visible response change for integrators diffing payloads, and "no database migration" in the description may read as "no client-visible change" when the response body does gain a key.

`trans_status` on `ExternalAuthenticationDetailsResponse` being an enum rather than a string is the right call — that one has a fixed EMV domain and deserves the type.

XyneSpaces:
[should-fix] The new `authentication_data` field in `crates/api_models/src/payments.rs` and `crates/api_models/src/payments/additional_info.rs` holds raw connector 3DS/auth payloads as `serde_json::Value`. These payloads can contain sensitive authentication details; wrap the field in `Secret<serde_json::Value>` or a typed PII wrapper to keep it out of logs and serialized diagnostics.

```suggestion
pub authentication_data: Option<Secret<serde_json::Value>>,
```

PR #14299, opened 2026-09-18
feat(connector): [Nuvei] enable UCS 3DS Authenticate and PostAuthenticate legs

shuklatushar226:
Depends on juspay/hyperswitch-prism#2324

## Why

Nuvei's 3DS authentication legs (PreAuthenticate / Authenticate / PostAuthenticate) are implemented in UCS
(juspay/hyperswitch-prism#2324) but Hyperswitch cannot reach them: `ConnectorSpecifications::is_authentication_flow_required`
and `is_post_authentication_flow_required` fall through to the trait defaults in
`crates/hyperswitch_interfaces/src/api.rs:467` and `:471`, both of which return `false`. Hyperswitch therefore never
invokes the Authenticate or PostAuthenticate step for Nuvei, and the UCS implementation is dead code from HS.

This PR adds the two overrides to the existing `ConnectorSpecifications` impl block in
`crates/hyperswitch_connectors/src/connectors/nuvei.rs`, using `cybersource.rs:2546` / `:2570` as the reference shape.
+37 lines, one file, nothing else touched.

| Change | Gate fn (file:line) | Why unreachable | Flows unblocked |
|---|---|---|---|
| `P-__hs__-01` — override `is_authentication_flow_required` for the 3DS card flows | `crates/hyperswitch_connectors/src/connectors/nuvei.rs:1818` (impl block at `:1796`); default at `crates/hyperswitch_interfaces/src/api.rs:467` | The trait default returns `false` and `nuvei.rs` did not override it, so the Authenticate leg is never invoked | ThreeDS (PreAuthenticate, Authenticate) |
| `P-__hs__-02` — override `is_post_authentication_flow_required` | `crates/hyperswitch_connectors/src/connectors/nuvei.rs:1835` (impl block at `:1796`); default at `crates/hyperswitch_interfaces/src/api.rs:471` | The trait default returns `false`, so the PostAuthenticate leg is never invoked | ThreeDS (PostAuthenticate) |

## Known limitation — read before approving (B068 / RV-011, S1)

> [!WARNING]
> **These two overrides are currently no-ops on the Direct path.**
>
> Nuvei is still listed in `default_imp_for_authenticate_steps!` (`crates/hyperswitch_connectors/src/default_implementations.rs:2996`)
> and `default_imp_for_post_authenticate_steps!` (`:3164`). Both newly reachable legs therefore resolve to the
> macro-generated **empty** impl, whose trait-default `build_request`
> (`crates/hyperswitch_interfaces/src/api.rs:243-253`) increments `metrics::UNIMPLEMENTED_FLOW` and returns `Ok(None)`.
>
> The plan items called for removing those two macro entries. That was deliberately **not** done, and the reason is
> recorded in the run's `code/11-__hs__.json .deviations[]`:
>
> - removing an entry without providing a real `ConnectorIntegration<Authenticate, …>` / `<PostAuthenticate, …>` impl
>   **does not compile**; and
> - providing real impls means authoring a full HS-native transformer pair, which contradicts the plan item's own
>   stated scope (*"Minimum change: nothing else in nuvei.rs is touched"*).
> - `crates/hyperswitch_connectors/src/connectors/paysafe.rs:1385` is the in-tree precedent for the
>   gate-override-only shape.
>
> **Related rollout risk:** `nuvei` is **not** listed in `ucs_only_connectors`
> (UCS `config/development.toml:1610`). On a Direct-path 3DS card attempt the two new branches fire into the empty
> impl — a silent no-op leg rather than an error. Whoever enables Nuvei 3DS must either add `nuvei` to
> `ucs_only_connectors` so the UCS path serves these legs, or land real HS-native impls and drop the two macro entries.
>
> The asymmetry is worth naming: the **PreAuthenticate** leg was wired correctly (real impl at `nuvei.rs:1017`, absent
> from the pre-auth macro list). Only Authenticate and PostAuthenticate are affected.

Two further S0 defects in the UCS composite 3DS loop (an unbounded `/initPayment.do` loop and a double charge on the
challenge path) are documented in juspay/hyperswitch-prism#2324. They are latent today only because a third defect
(`session_token` hard-coded `None`) makes composite 3DS abort before the first HTTP call. **Do not enable Nuvei 3DS
end to end until those are fixed.**

## Merge order

1. juspay/hyperswitch-prism#2324 merged and UCS deployed.
2. This PR.

## Required human checks

- [ ] Link an open `juspay/hyperswitch` issue to this PR — CI job "Verify PR contains one or more linked issues" fails without one; GRACE opens none
- [ ] Decide the B068 follow-up: add `nuvei` to `ucs_only_connectors`, or land real HS-native Authenticate / PostAuthenticate impls and remove the two `default_implementations.rs` macro entries
- [ ] Confirm the UCS-side 3DS blockers (B062, B063, B064) are resolved before routing real 3DS traffic

## Validation

| Command | Exit |
|---|---|
| `cargo +nightly fmt --all` (in the run worktree) | 0 |
| `cargo check -p hyperswitch_connectors --features v1,frm,payouts` | 0 |
| `cargo check -p router --features v1` | not run — no `crates/router` path is dirty |

E2E through this change (HS → UCS → connector): **ThreeDS `E2E_STATUS=FAILED`** — `hs cases failed`; one three-hop
transcript exists (`ThreeDS.hs_rest.happy.01`) but the unit did not pass end to end. Full per-flow status and
evidence in juspay/hyperswitch-prism#2324.


XyneSpaces:
[blocking] This PR enables Nuvei's Authenticate and PostAuthenticate 3DS legs on the Direct path, but Nuvei still uses the default implementations (`default_imp_for_authenticate_steps!` / `default_imp_for_post_authenticate_steps!`) and is not registered in `ucs_only_connectors`. On the Direct path the legs will silently resolve to the empty macro-generated impls that increment `UNIMPLEMENTED_FLOW` and return `Ok(None)`, so 3DS card transactions will appear to succeed while skipping real authentication.

Either add Nuvei to `ucs_only_connectors` (in development.toml, sandbox.toml, production.toml, env_specific.toml, integration_test.toml, and config.example.toml) so the new legs route through UCS, or remove it from the default authenticate/post_authenticate macro lists and provide real HS-native implementations before enabling `is_authentication_flow_required` / `is_post_authentication_flow_required`.

XyneSpaces:
[should-fix] `crates/hyperswitch_connectors/src/connectors/nuvei.rs:1818` — `is_authentication_flow_required` override is a no-op on the Direct path because Nuvei is still listed in `default_imp_for_authenticate_steps!` (`default_implementations.rs:2996`). Reachable legs hit the empty macro impl and return `Ok(None)`, so the 3DS Authenticate leg will not be invoked as intended.

Remove Nuvei from the default macro or ensure the explicit override actually overrides the default behavior.

[should-fix] `crates/hyperswitch_connectors/src/connectors/nuvei.rs:1835` — same structural issue for `is_post_authentication_flow_required` vs `default_imp_for_post_authenticate_steps!` (`default_implementations.rs:3164`).

PR #14305, opened 2026-09-18
fix(authentication): added highest common supported version for 3ds server

sahkal:
## Type of Change

- [x] Refactoring

## Description

added highest common supported version for 3ds server

## Motivation and Context

Currently we are taking the highest supported version of the ACS but for juspayThreeDs case we should take the highest supported version by the server not the acs

## How did you test it?

tested Juspay3ds locally

## Checklist


deepanshu-iiitu (dismissed):

XyneSpaces:
[blocking] `message_version` is silently set to `None` when `highest_common_supported_version` is absent in the UAS eligibility response (`crates/hyperswitch_connectors/src/connectors/unified_authentication_service/transformers.rs:452-454`). The missing value only surfaces later as an opaque `get_required_value("message_version")` failure deep in the authentication flow, making root-cause debugging difficult and delaying the failure.

Fail fast with a descriptive error at the source, or fall back to `maximum_supported_3ds_version` when the new field is absent so existing responses remain usable.

XyneSpaces:
[should-fix] `crates/hyperswitch_connectors/src/connectors/unified_authentication_service/transformers.rs:452`

`message_version` now comes solely from the new optional `highest_common_supported_version` field. If the upstream eligibility service has not started populating it, pre-authentication flows will run with `message_version = None` and likely fail downstream 3DS handling. Keep `maximum_supported_3ds_version` as a fallback for rollout safety:

```rust
message_version: three_ds_eligibility_response
    .as_ref()
    .and_then(|response| {
        response
            .highest_common_supported_version
            .clone()
            .or(maximum_supported_3ds_version.clone())
    }),
```

XyneSpaces:
[should-fix] `crates/hyperswitch_connectors/src/connectors/unified_authentication_service/transformers.rs:452` now takes `message_version` only from `highest_common_supported_version` with no fallback.

For 3DS eligibility responses that still populate `maximum_supported_3ds_version` but not the new server-common field (external/non-Juspay3DS servers or older deployments), `message_version` becomes `None` where it previously held the max ACS version. This can break downstream 3DS method/challenge construction for those responses.

```suggestion
message_version: three_ds_eligibility_response
    .as_ref()
    .and_then(|response| response.highest_common_supported_version.clone())
    .or(maximum_supported_3ds_version.clone()),
```

XyneSpaces:
[should-fix] `crates/hyperswitch_connectors/src/connectors/unified_authentication_service/transformers.rs:453` — `message_version` is now taken only from `highest_common_supported_version` with no fallback to the previous `maximum_supported_3ds_version`. When the new field is absent the version becomes `None`, regressing all non-`juspayThreeDs` flows.

Preserve a fallback chain: use `highest_common_supported_version` if present, else fall back to `maximum_supported_3ds_version`.

XyneSpaces:
:rotating_light: **Potential silent regression in 3DS message version sourcing** (file:line reference below; inline comments could not be posted by this automation):

- `crates/hyperswitch_connectors/src/connectors/unified_authentication_service/transformers.rs:452`

`message_version` is now sourced only from the optional `highest_common_supported_version` field with no fallback. When the upstream UAS eligibility response omits this field, `message_version` becomes `None` and the failure is deferred to a downstream `get_required_value("message_version")`, producing an opaque error far from the root cause. Either fail fast here with a descriptive error, or preserve `maximum_supported_3ds_version` as a fallback so non-juspayThreeDs / older responses continue to work.

bsayak03 (dismissed):

bsayak03 (approved):

deepanshu-iiitu (approved):

Issue #173, opened 2022-12-19
feat: Implement Payment Verify for connectors


dracarys18:
# Description
Implement Verify flow for the connectors. Verify flow makes sure if the card is valid or not for future payments like mandate etc.

# Connectors to implement
- [ ] Authorizedotnet
- [ ] Checkout
- [ ] ACI
- [ ] #1068
- [x] #1045

PR #14311, opened 2026-09-18
refactor(deja): move the pin to deja main and port the seams onto the miss arm

maverox:
> [!WARNING]
> **Draft — the pin is temporary and must move before review.**
>
> The nine deja revisions currently point at `159aadc`, the head of deja's `work/teardown-safe-correlation-reader` ([juspay/deja#171](https://github.com/juspay/deja/pull/171)), not at a commit on deja main. Every manifest says so in a comment. Bumping back to a main commit once #171 merges is a **precondition for review**, not a follow-up to it.
>
> It is pinned by revision rather than `branch = "..."` deliberately: a branch pin re-resolves on any lock refresh, so the deja underneath would move without a manifest edit — and deja records its own sha into every tape it writes, which makes a moving pin a moving experiment.

## Type of Change

- [x] Refactoring
- [x] Dependency updates

## Description

**Replaces #14202**, which is closed. Same intent, cut fresh against `main` rather than rebased: that branch sat 309 files behind, and every rebase conflict turned out to be one comment in two phrasings, because the shortened versions had already landed via #14201.

Two commits, and they are not separable.

**The pin.** Nine crates move together from `ade8fd16c1` to deja main `8d87c19`. All nine, because two revisions of deja in one build means two copies of the recording statics, and a recorder writing into the copy nobody reads fails silently rather than erroring. The check that matters is that the old revision has zero occurrences afterwards, lock included.

For the record, since #14202's pin has been described loosely: `d587865` was **not** wrong or divergent — `ade8fd16c1` is an ancestor of it, so that PR would have moved the build six commits forward on the same lineage. It is simply 41 commits stale now.

What the bump brings in beyond `ade8fd16c1`: per-process tape identity at ingest, address-based call pairing, the run list served without its scorecards, replay-run scheduling, and six review-driven fixes on top. The one most visible in a replay report is that a re-keyed call is now a consequence only of a divergence that happened **before** it, where it used to be demoted by any divergence anywhere in the correlation.

**The seams.** The new pin removes `MissPolicy` and `dispatch_async_or_miss`, so the four sites that used them cannot compile unchanged. The port is what the bump requires, not dressing on it.

A boundary used to *declare in advance* what a miss would do. It now *returns what it did*, and the observation records the outcome rather than the declaration. The two used to be the same thing, because a declared miss value always produced one; they came apart once a miss arm could decline.

- **Superposition** returns `Synthesized(Err(NotFound))` — the error the caller already survives, so the DB-to-default fallback runs and the correlation continues. The message is built from the miss alone, which is the property the closure signature exists to enforce: same query, same value, every run.
- **The gRPC transport** keeps its two arms distinct, which is the part that would have been easy to lose. Transport present, a miss may be a genuine novel call, so that arm returns `NoValue`. Transport deliberately absent, there is nothing to run and a miss establishes nothing about the candidate, so that arm still calls `fail_stop_absent_executor` and keeps its own reason. Collapsing both into `NoValue` would have compiled, and would have replaced a specific operator-facing message with a generic one.
- **`router_env`'s id-generation lookup** loses the declared field. Behaviour unchanged: no miss value is offered, so the caller falls through to minting a fresh uuid — precisely a value the recording never held.
- **The http boundary** restores the extension a response body is captured from. Without it, capturing a reconstructed response reports "body not captured" instead of the body just rebuilt, so `capture(reconstruct(v))` does not equal `v`. Nothing re-captures on a replay hit today, so this is inert at runtime; the codec should not depend on that staying true.

**The in-memory cache degrades instead of dying.** `get_val` and `exists` answer a Substitute miss with "not in cache" rather than fail-stopping. A candidate under review adds a cache read — that is what a change is — and the fail-stop default answers the first one by unwinding, which actix does not contain, so the worker writes nothing and the whole correlation scores `500 vs 200`. Measured: one added read took **45 of 52 correlations** that way and left a single correlation matched. One regression censoring every other signal is worse than no run, because it looks like a result.

`None` is honest here in a way a fabricated value is not — it asserts only that this key is not in the process-local cache, which is true under replay, and the fallback to Redis or the database is separately instrumented. Egress keeps the fail-stop default deliberately: synthesizing an http or grpc response would claim a third party answered when none did. The miss is not swallowed either — the lookup emits its blocking novel-call divergence before the miss arm is reached, so the scorecard still shows the added read; it simply localises to the subtree that depended on it.

## Not in this PR

**The collections facade is now included**, which is why the pin points at a branch. `common_utils::collections` gives `HashMap`/`HashSet` a correlation-derived hash seed so a set's iteration order matches between recording and replay. It does not compile against deja main: `Default for CorrelationHasher` runs on every `HashMap::new()`, destructors included, and `current_correlation_id` reads its thread-local with a plain `with`/`borrow` — which aborts the process there rather than failing a request. deja#171 adds the fallible twin it calls.

Nothing calls the facade yet. It is the mechanism, not the adoption. The target is `payment_methods_enabled`, built at `crates/router/src/core/payment_methods/cards.rs:5215` by iterating nested `HashMap`s and pushing into `Vec`s — ordering there is 26 of 27 blocking body mismatches on the run we care about. The construction path is verified; the causal link to those 26 is strongly indicated and not yet proven against a failing scorecard, which is what a build carrying this is for.

**A miss arm on `ReplayCodec`** — juspay/deja#169. `reconstruct` takes a recorded value, so the trait covers only the hit half and every site open-codes its miss elsewhere. The gRPC transport here is the clearest case: it cannot use `codec = <Path>` at all, because that is an attribute on `#[deja::instrument]`-decorated functions and the transport is a `tower::Service` impl. `GrpcResultEnvelope` **is** its codec; it is just supplied as two loose closures rather than declared, and its miss arm is an inline branch. Worth fixing in deja rather than worked around here.

## How did you test it?

By compiling, with the feature sets that actually reach the changed files:

- `external_services` at `--features v1,deja,superposition`
- `router_env` at `--features deja`
- `router` at `--features v1,deja,release`
- `storage_impl` at `--features v1,deja`
- `common_utils` at `--features deja` — green verified capable of failing first, since the seeded hasher sits behind `#[cfg(feature = "deja")]`

All clean.

**The superposition check was verified capable of failing before its green was believed.** `superposition.rs` sits behind `#[cfg(feature = "superposition")]`, so a run without that feature compiles none of the ported file and reports success anyway — that happened on the first attempt at this work. A `compile_error!` inserted at the top of the file fired under this feature set, which is what makes the green mean something.

**The lock diff is accounted for line by line.** Beyond the pin itself, three lines moved: `siphasher 1.0.1` gains a consumer (deja's hash seeding), `prost-build` rebinds `heck` 0.5.0 → 0.4.1, and `libloading` rebinds `windows-targets` 0.53.5 → 0.48.5. No package was added or removed — every one of those versions was already in `main`'s lock, and both rebound crates are build-time or platform shims. They are named here because a lock diff nobody accounts for is how an unrelated dependency moves under a pin bump unnoticed.

## Additional Changes


XyneSpaces:
[blocking] This PR pins the `deja` dependency to a commit on the unmerged branch `work/teardown-safe-correlation-reader` (juspay/deja#171) instead of a commit on `main` or a released tag. Merging this would make hyperswitch builds depend on unstable, review-in-progress code that may be force-pushed or deleted.

Pin `deja` to a commit on `main` (or a release tag) before merging. If the upstream branch is required temporarily, document the merge blocker and the revert plan in the PR body.

XyneSpaces:
[blocking] `crates/common_utils/Cargo.toml:35` (and `diesel_models`, `external_services`, `redis_interface`, `router`, `router_env`) — `deja` is pinned to branch `work/teardown-safe-correlation-reader` / rev `159aadc`, not a main commit. Branch-based git dependencies are non-deterministic and can be deleted. The PR body itself says this temporary pin must move before review.

Pin `deja` to a tag or `main` commit before merging.

[should-fix] `crates/storage_impl/src/redis/cache.rs:318` / `:377` — `Cache::get_val` and `Cache::exists` now declare `on_miss = None`/`false`, degrading replay cache misses instead of fail-stopping. Confirm this substitution cannot leak into production behavior outside replay, or keep the fail-stop behavior for non-replay callers.

XyneSpaces:
:rotating_light: **Temporary deja pin to unreleased branch** (file:line references below; inline comments could not be posted by this automation):

- `crates/common_utils/Cargo.toml:35`
- (and the same `deja` / `deja-core` pin in `diesel_models`, `external_services`, `redis_interface`, `router`, `router_env`, `storage_impl` manifests)

The `deja` dependency is pinned to the `work/teardown-safe-correlation-reader` branch from juspay/deja#171 and marked TEMPORARY. Merging would ship unreleased upstream code that may be rebased or deleted. Wait for juspay/deja#171 to land on main, then bump every reference to a stable main revision and remove the TEMPORARY PIN comments.

:warning: `crates/common_utils/src/collections.rs:84` — `keys_from_correlation` derives HashMap/HashSet seeds via `std::collections::hash_map::DefaultHasher` (fixed keys). If a correlation id can ever be influenced by attacker-controlled input, this removes per-process randomization and exposes hash-flooding DoS risk. Guarantee correlation ids are internally generated and never derived from untrusted data, or pass them through a keyed one-way function before seeding `DejaBuildHasher`.

maverox (commented):
Reviewed the seam ports, the collections facade and the pin. The structure is right — collapsing the second dispatch function plus a declared policy into one `reconstruct` that answers both arms removes a class of bug where the declaration and the thunk drift apart, and `superposition.rs` building its miss message from `miss.method` alone is exactly the determinism property that makes a synthesized value safe.

Four things, one of which is a cross-repo interaction that landed a few hours ago and is not the author's doing.

## 1. The `blocking` claim is stale as of deja #173, and it is load-bearing

Both `cache.rs` comments say the lookup emits a **blocking** novel-call divergence before `on_miss` is reached. That was true when written. It is not true against deja main as of today.

The chain, checked rather than reasoned:

- `on_miss = <expr>` expands to `Reconstructed::Synthesized(expr)` — `deja-derive/src/instrument.rs:474`
- `Reconstructed::Synthesized(_) => SubstituteOutcome::Synthesized` — `deja-runtime/src/lib.rs:4143`
- the ledger now grades a novel row `("novel", stopped_at(obs))`, and `stopped_at` is `outcome == SubstituteOutcome::Stopped` — `deja-orchestrator/src/divergence/ledger.rs:564` and `:188`

So a novel call answered by `on_miss` is `Synthesized`, never `Stopped`, and therefore scores **non-blocking**.

The argument in these comments survives intact — the miss is still scored, it is still a novel row, it still shows on the scorecard, and it still localises instead of taking the correlation down. Only the adjective is wrong. I am flagging it because it is precisely the sentence a future reader would use to justify the next `on_miss`, and the reasoning would inherit a false premise.

**Not yours to chase across repos:** deja's own comment carries the identical wording at `deja-derive/src/instrument.rs:137` ("the blocking NovelCall divergence is emitted by the lookup"). Same root cause, and it is mine — #173 changed the rule and left that sentence behind. I will fix the deja side; flagging so the two do not get corrected in opposite directions.

## 2. `CorrelationHasher` merges as unreachable code

The facade is wired in now (`lib.rs:7`), which makes this live rather than hypothetical. `CorrelationHasher` is defined, documented at length, unit-tested and exported — and referenced from nowhere. The aliases default to `RandomState`, and nothing in the tree sets `S = CorrelationHasher`.

Its own doc says it "is the piece that makes an import-only migration possible". An import-only migration is `use common_utils::collections::HashMap`, which resolves to `S = RandomState` and never reaches it. So the module documents two incompatible designs: explicit per-site naming via `seeded_map`, which the aliases and the module header describe, and implicit correlation-derived seeding, which `CorrelationHasher` describes and which nothing can currently reach.

One of the two has to give before this merges. Either the aliases default to `CorrelationHasher` under the `deja` feature — which makes phase 1 a real behaviour change rather than the byte-for-byte no-op the module header promises, deliberately — or the doc is overclaiming and it is an escape hatch for a future call site. Both are defensible; shipping it undecided means the next reader takes the doc at its word.

## 3. `request_id` is the one site the old policy's own doc named

Removing the declaration is forced — `ReplayLookup` no longer has the field — so this is not a choice the PR made. But the removed `MissPolicy::FailStop` documented itself as right "wherever no value is honest — egress, and anything whose absence would silently condition the rest of the correlation on a value the recording never held", and the new comment states that a miss now falls through to minting a fresh uuid that the rest of the correlation is keyed on. The comment is honest about it, which I appreciate; the question is whether that is the outcome anyone wants.

This seam takes the `Option`-returning `try_replay_with_context` path rather than the macro, so there is no `Reconstructed` here and no way to signal `Stopped` — which is also why `on_miss` is not available to it as written. Worth deciding explicitly: is a deterministically synthesized id derived from the query the right answer here, the way `superposition.rs` derives its message from `miss.method` two files over? A minted uuid is the one value guaranteed not to be the recorded one.

Not a blocker, and possibly out of scope for this PR — but it should be a decision rather than a consequence.

## 4. Duplicate attribute

`collections.rs:252-253` carries `#[cfg(feature = "deja")]` twice on `pub use correlation::CorrelationHasher;`. Harmless — both must hold, so it compiles the same — just noise.

## Pin

`159aadc` is on deja main, so the pin is valid. Worth knowing for whenever it next moves: `deja`, `deja-runtime` and `deja-kernel` all gained `serde_json/preserve_order` after that commit, and two of those ship in the router. The next bump therefore changes how the recorder serialises — recorded artifacts keep the key order a payload had instead of sorting it. That is what the deja change was for, but it means tapes either side of that bump differ in key order, so it is a behaviour change rather than housekeeping.

maverox (commented):
Anchoring the four findings from my summary review to the lines they apply to.

crates/storage_impl/src/redis/cache.rs
```diff
@@ -295,6 +295,27 @@ impl Cache {
...
+    //
+    // `None` is honest here in a way a fabricated value never is. It asserts only
+    // that this key is not in the process-local cache, which IS true under replay
+    // — the moka is cold and correlation-namespaced — and the caller's fallback
+    // to Redis or the database is separately instrumented, so the work that
+    // follows the miss is scored on its own boundaries rather than hidden behind
+    // a substituted value. Egress keeps the fail-stop default: synthesizing an
+    // http/grpc response would claim a third party answered when none did.
+    //
+    // The miss is not swallowed. The lookup emits its blocking novel-call
```
maverox, line 326: **`blocking` is stale as of deja #173** (merged today).

`on_miss = None` expands to `Reconstructed::Synthesized(None)` (`instrument.rs:474`), which maps to `SubstituteOutcome::Synthesized` (`runtime/lib.rs:4143`). The ledger now grades a novel row `("novel", stopped_at(obs))` where `stopped_at` is `outcome == Stopped` (`ledger.rs:564`, `:188`). So this call is `Synthesized`, never `Stopped`, and scores **non-blocking**.

The rest of the sentence holds — the miss is still scored, still a novel row, still localises. Only the adjective is wrong, and it is the sentence someone will quote to justify the next `on_miss`.

deja's own comment has the same wording at `deja-derive/src/instrument.rs:137`; that one is mine and I will fix it.

crates/storage_impl/src/redis/cache.rs
```diff
@@ -328,6 +350,20 @@ impl Cache {
...
+    // in this method's own type. "Not in cache" is TRUE under replay — the moka
+    // is cold and correlation-namespaced — and a `false` sends the caller down
+    // the fallback path, which is instrumented on its own boundaries.
+    //
+    // Without it, this sibling still takes the correlation down. `get_val`'s
+    // declaration covers reads made THROUGH `get_val`, so a candidate that adds
+    // an existence check fail-stops exactly the way an added read did before
+    // that declaration existed: the same incident, one method over.
+    //
+    // As there, the miss is not swallowed — the lookup emits its blocking
```
maverox, line 375: Same `blocking` claim as line 315 — `on_miss = false` is `Synthesized` too, so this one is non-blocking as well. Both want the same correction.

crates/common_utils/src/collections.rs
```diff
@@ -0,0 +1,259 @@
...
+/// The hasher `std` uses. Named here so a signature can spell the default
+/// without importing from `std::collections::hash_map`.
+pub use std::collections::hash_map::RandomState;
+
+/// `std::collections::HashMap`, re-exported so changing an import is a no-op.
+///
+/// The `S` parameter defaults to [`RandomState`], so `HashMap::new()`,
+/// `.collect()` and every other `std` affordance behave exactly as before.
+/// Pass a seed from [`seeded_map`] to get a reproducible iteration order.
+pub type HashMap<K, V, S = RandomState> = std::collections::HashMap<K, V, S>;
```
maverox: This is what makes `CorrelationHasher` below unreachable.

The alias defaults to `RandomState`, so an import-only migration — `use common_utils::collections::HashMap` — resolves to std behaviour and never reaches the correlation hasher. But `CorrelationHasher`'s own doc says it "is the piece that makes an import-only migration possible".

Two incompatible designs in one module: explicit per-site naming via `seeded_map`, which this line and the module header describe, and implicit correlation-derived seeding, which nothing can currently reach. Either this defaults to `CorrelationHasher` under the `deja` feature — making phase 1 a deliberate behaviour change rather than the byte-for-byte no-op the header promises — or that doc is overclaiming. Worth settling before it merges; the module is wired into `lib.rs` now, so this is live.

crates/common_utils/src/collections.rs
```diff
@@ -0,0 +1,259 @@
...
+        I: IntoIterator<Item = T>,
+    {
+        let mut set = seeded_set(name);
+        set.extend(items);
+        set
+    }
+}
+
+#[cfg(feature = "deja")]
+#[cfg(feature = "deja")]
```
maverox: `#[cfg(feature = "deja")]` twice. Compiles the same — both must hold — just noise.

crates/router_env/src/request_id.rs
```diff
@@ -148,10 +148,12 @@ pub(crate) mod boundary {
             return None;
         }
         let recorded = hook.try_replay_with_context(deja::ReplayLookup {
-            // `FailStop`: with no `on_miss` the caller mints a fresh uuid —
-            // a value the recording never held, and one the rest of the
-            // correlation is then keyed on.
-            miss_policy: deja::MissPolicy::FailStop,
+            // This boundary offers no miss value, so the caller falls through to
```
maverox, line 151: Removing the declaration is forced (`ReplayLookup` no longer has the field), so this is not a choice the PR made — but it lands on the one site the old policy's doc named explicitly.

`MissPolicy::FailStop` documented itself as right "wherever no value is honest — egress, and anything whose absence would silently condition the rest of the correlation on a value the recording never held." This comment now states that a miss mints a fresh uuid the rest of the correlation is keyed on. I appreciate that it says so plainly rather than quietly.

This seam takes the `Option`-returning `try_replay_with_context` path rather than the macro, so there is no `Reconstructed` and no way to signal `Stopped` — which is also why `on_miss` is not available here as written.

Worth an explicit decision: should this synthesize an id **deterministically from the query**, the way `superposition.rs` builds its miss message from `miss.method` alone? A freshly minted uuid is the one value guaranteed not to be the recorded one. Possibly out of scope for this PR, but better as a decision than a consequence.

crates/router/src/core/webhooks/outgoing_v2.rs
```diff
@@ -153,6 +151,10 @@ pub(crate) async fn create_event_and_trigger_outgoing_webhook(
     let cloned_processor_merchant_id = processor_merchant_id.clone();
     // Using a tokio spawn here and not arbiter because not all caller of this function
     // may have an actix arbiter
+    #[allow(
+        clippy::disallowed_methods,
+        reason = "lone detached spawn: it carries the request span so its boundaries stay correlated, but it has no racing sibling to be paired against. An identity of its own needs a named-fork API from deja: fork_span() hardcodes one span name, and a host cannot replicate spawn_fork without capture_current/scope_snapshot"
+    )]
```
maverox: why cant we make the rule such that it forgives where .in_current_span() is used

check and fix all instacne this is too verbose

crates/router/src/core/payment_methods.rs
```diff
@@ -472,7 +470,7 @@ pub async fn render_pm_collect_link(
             };
             Ok(services::ApplicationResponse::GenericLinkForm(Box::new(
                 GenericLinks {
-                    allowed_domains: HashSet::from([]),
+                    allowed_domains: Default::default(),
```
maverox: shouldn't we use a from impl like the hunk you deleted?

crates/common_utils/src/collections.rs
```diff
@@ -0,0 +1,240 @@
...
+        };
+        (draw(0), draw(1))
+    }
+
+    impl Default for CorrelationHasher {
+        fn default() -> Self {
+            // `try_` because this runs from arbitrary code, destructors
+            // included, and a plain `with` on a destroyed thread-local panics
+            // inside a drop, which aborts the process.
+            let (k0, k1) = match deja::try_current_correlation_id() {
```
maverox: Finding 1: no test goes through this `Default` inside an entered correlation. Forcing the arm below to always derive, or never derive, or dropping the runtime-disabled check at `:110`, leaves the full `common_utils --features deja` suite green (0 tests fail for each). The review summary lists four tests that would catch it, including the producer-side one for `request_id.rs:1091`.

crates/common_utils/src/collections.rs
```diff
@@ -0,0 +1,240 @@
...
+    /// `FromIterator`, where no call site could pass anything. Outside a
+    /// correlation the keys are random, which keeps hash-flooding resistance.
+    ///
+    /// The keys are fixed when the collection is BUILT, not when it is read. So
+    /// one built with no correlation in scope, a settings table deserialized at
+    /// startup for instance, keeps one random order for the life of the
+    /// process. That matters only where the order leaves the collection: deja
+    /// reads a JSON object by key, so what is exposed is a list or a string
+    /// built by iterating it. Nothing here addresses that. Rebuilding the
+    /// collection inside the request, by collecting it, gives it the request's
```
maverox: Finding 4: this gives the request's keys, but not a reproducible order. With one fixed pair of keys, 30 elements inserted in 200 shuffled orders iterate in 190 distinct orders. A set collected from a startup-built set inherits that set's random iteration order as its insertion order. Suggest saying keys, not order, and pointing at sorting where the order leaves the collection.

.clippy.toml
```diff
@@ -10,6 +10,7 @@ allow-unwrap-in-tests = true
 # with a reason. "It is only telemetry" and "it is only a test" are not exceptions:
 # a raw read in either place is the one the next person copies onto a request path.
 disallowed-methods = [
+  { path = "tracing::Instrument::in_current_span", reason = "a spawned future sharing the caller span is separable only by scheduler order; name the fork with .instrument(tracing::debug_span!(\"...\")) or allow with a reason" },
```
maverox: Finding 2: `tracing_futures::Instrument` is a separate trait with its own `in_current_span`, so this path doesn't reach it. Eight production spawns go through it unflagged, including the four outgoing-webhook spawns in `router/src/utils.rs`. Worth adding `tracing_futures::Instrument::in_current_span` here, whatever shape the verbosity thread settles on.

Cargo.lock
```diff
@@ -7541,7 +7544,7 @@ version = "0.14.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "343d3bd7056eda839b03204e68deff7d1b13aba7af2b2fd16890697274262ee7"
 dependencies = [
- "heck 0.5.0",
+ "heck 0.4.1",
```
maverox: Finding 3: this edge was `heck 0.5.0` at the merge base. With these manifests, cargo itself resolves 0.5.0, so this came from the lock edit rather than resolution. The same applies to `libloading`'s `windows-targets 0.48.5` at `:5627` (was 0.53.5). Both 0.5.0 and 0.53.5 are still in the lock for other crates, which is why a version-line count misses it.

maverox (commented):
Review of `f163b5227e` (merge base `0defa221a3`). No runtime regression found. There are four findings: missing coverage on the mechanism this PR introduces, a lint blind spot, a lock regression with a two-line fix, and a doc claim about order that doesn't hold, with the places where that matters. Anything not listed below was checked and holds; what was run is at the end.

## Findings, most severe first

**1. The hash-key guard is untested in both directions (confirmed by mutation).** `CorrelationHasher::default()` (`common_utils/src/collections.rs:92-99`, guard at `:109-111`) decides for every map and set in the workspace whether the keys are derived from the correlation or drawn at random. I ran each change below against the full `cargo test -p common_utils --features deja --lib` suite. Baseline is 98 pass / 2 fail, and both failures are pre-existing, in files this PR does not touch.
- guard forced to always derive: 0 tests fail
- guard forced to never derive: 0 tests fail
- `!deja::runtime_mode_is_disabled() &&` removed: 0 tests fail
- map `Debug` rendered as `HashMap({..})` instead of forwarding: 0 tests fail

Every correlation test builds `CorrelationHasher { k0, k1 }` directly or calls the pure `sampled()`. None goes through `Default` inside an entered correlation, which is the path every real map takes. The producer side is untested too: nothing asserts that record mode pushes `Skip` for a request the sampler declines (`router_env/src/request_id.rs:1091`), which is the property the guard's doc at `:113-118` depends on. Precondition for harm: a later change makes that push conditional, or refactors the guard. The first gives predictable keys to unsampled traffic on a recording pod. The second quietly turns the facade off. Either one ships green today. Tests that would catch each (the runtime hook is a one-shot `OnceLock`, so each hook state needs its own test binary):
- replay hook installed (empty table at `deja::POLICY_VERSION`): two default-built maps inside one correlation render identically;
- no hook, and separately a record hook with `set_recording_decision(id, false)`: they render differently;
- a `router_env` test driving `RequestIdMiddleware` in record mode with a sampler returning `false`, asserting `deja::recording_decision(<request id>) == Some(RecordDecision::Skip)` inside the handler;
- `format!("{:?}")` of a facade map equals std's built with the same hasher keys. SQL bind comments and logs render maps through it.

**2. The fork-naming lint does not see `tracing_futures::Instrument` (confirmed).** `.clippy.toml` bans `tracing::Instrument::in_current_span`. `tracing-futures` 0.2.5 declares its own `Instrument` trait with a method of the same name, which is a different path. Four router files import it, and 8 production spawns go unflagged with no allow: `payment_confirm.rs:1524`, `payment_response.rs:92` and `:3265`, `utils.rs:1273`, `:1363`, `:1451` and `:1521` (the outgoing-webhook spawns), and `utils/currency.rs:229` (silenced anyway by that file's allow). They are v1 code, and the stable-toolchain job runs `just clippy` under `-D warnings` and is green on this tip, so they are linted and not flagged. The tally on the tip is 36 `.in_current_span()` call sites: 26 carry an allow, 8 go through `tracing_futures`, and 2 are string literals in a test. Whatever shape the lint takes after the verbosity thread, it should also name `tracing_futures::Instrument::in_current_span` (or those imports should move to `tracing::Instrument`).

**3. `Cargo.lock` moves two dependency edges backwards (confirmed).** Compared per package against the merge base:
- `prost-build 0.14.3`: `heck 0.5.0` → `0.4.1`
- `libloading 0.8.8`: `windows-targets 0.53.5` → `0.48.5`

Both come from `a92acdb8c1`. Every package version involved still exists in the lock, because other crates use them, so a count of changed `version =` lines cannot see this. Cargo would not produce it: with these manifests and the base lock, `cargo metadata --offline` resolves `heck 0.5.0` and `windows-targets 0.53.5`, and the tip lock with only those two lines restored passes `cargo metadata --locked`. Runtime impact is likely nil (Windows-only, and build-time case conversion for prost codegen), but it would land on main as a silent downgrade.

**4. Same keys give the same order only if the insertion order is also the same (confirmed by measurement), and six response surfaces inherit an order built at startup.** Using hashbrown with one fixed pair of keys, I inserted the same 30 elements in 200 shuffled orders and got 190 distinct iteration orders (200 of 200 at 50 elements). Equal keys reproduce the order only when the insertion sequence repeats. That qualifies the doc at `collections.rs:54-56`. Rebuilding a collection inside the request does give it the request's keys, but it does not give a reproducible order when the source iterates randomly, and a set built at startup is exactly that case. The same applies wherever a std map is collected into the facade (the redis reads, for example): the result is re-keyed, but its order is still random. This is not a regression, since these orders were random before. It is the documented limit, now located:
- `payment_methods/cards.rs:4032-4039`: `get_banks` clones the configured bank set (`Clone` keeps the startup hasher) and collects it into `bank_name`;
- `configs/settings.rs:1063`: `get_all_keys` collects `self.key.values()` into the JWKS array, even though `get_signing_key` just above sorts the key ids;
- `payment_methods/cards.rs:7361-7378`: list-countries-currencies returns the configured `currency` set as an array;
- `payment_methods/validator.rs:132` and `payout_link.rs:178`: the default `enabled_payment_methods` sets the order of a list written to the DB row and into the payout-link page's JSON string;
- traced but not re-read by me: the feature-matrix `supported_currencies` and the EPS bank `options` in the required-fields response.

A reordered array in a response body is scored as an order-only difference, so the DB write and the HTML string are the two that would score as real divergences. Using fixed keys when there is no correlation would fix these only where deserialization inserts in a fixed order. Sorting at the point where the order leaves fixes them regardless.

## Earlier threads, on this tip
- `storage_impl/src/redis/cache.rs:326` and `:375` still say "blocking novel-call divergence", so those two threads are still open.
- `payment_methods.rs:473`: yes. The three `HashSet::from([])` → `Default::default()` rewrites in `34a4b82f0d` were forced while the facade was a type alias, because std defines `From<[T; N]>` only for `RandomState`. The wrapper now has its own (`collections/wrapper.rs:328`), so the original form can come back unchanged.
- `collections.rs:59` (unreachable `CorrelationHasher`) no longer applies: `DefaultHashBuilder` is `CorrelationHasher` under `deja` (`:30`). The duplicate `cfg` is gone.

## Nits
- The `.clippy.toml` comment above `disallowed-types` still calls the facade types "aliases".
- `a_millisecond_reading_round_trips_through_the_codec` goes through `serde_json::to_string`/`from_str`, while `SerdeCodec` captures through `serde_json::Value` (`result_serialize` / `from_value`). The two are equivalent over the tested range, but the test doesn't exercise the codec its name claims.
- Only `new` and `with_capacity` are banned as fully qualified std constructor paths. `std::collections::HashMap::from(..)` / `::default()` / `::from_iter(..)` pass both lints. None survives on this tip outside a named allow.

## Verified and holding
- **Pin:** 10 declarations in 8 manifests, one deja rev (`8c6d3e4a78`) in the lock. Merges cleanly with current main, and main's new commits add no map or set use and no unallowed spawn.
- **OpenAPI:** v1 and v2 regenerated at this tip with a sentinel in place are byte-identical to the checked-in specs, which match the base's. CI's spec job at this sha also committed nothing.
- **Wrapper forwarding:** `Default`, `Clone`, `Debug` (forwards the `Formatter`, so `{:#?}` survives and there is no `HashMap(..)` wrapper text), `PartialEq`/`Eq`, `FromIterator`, both `Extend`s, the three `IntoIterator`s, `Index`, `From` in both directions, `Serialize`, and `Deserialize` (via `S::default()`) all delegate to std with std's bounds. No crate in the workspace, and not `hyperswitch_masking`, implements a trait on std's map or set that the wrapper could silently miss.
- **Order on the wire:** in a build where deja is disabled at runtime, keys are still random per instance, as before. Nothing that was stably ordered becomes unstable. The std-map boundary sites don't leak order into anything replay compares: the UCS request map is canonicalized by `wire_canon` (repeated fields sorted), local protos decode to JSON objects, and Kafka produce is not a boundary. The redis reads collected into the facade keep a random order (see finding 4); I did not trace whether that order leaves the map.
- **Exemptions:** the stated reasons are true. `common_enums` and `router_env` would form a cycle, `router_derive` and `smithy` are proc-macros, and the other six have no `common_utils` dependency.
- **Hash flooding:** recording uses the configured `id_reuse_strategy` (default `ignore_incoming`), so a client cannot choose the correlation that keys a recorded request. Only replay forces reuse of incoming ids.
- **Merging one map into another under shared keys:** checked whether the quadratic copy behaviour that `RandomState`'s per-instance keys exist to prevent comes back. It does not with hashbrown: linear from 1k to 3.6M entries, at most about 1.5× the cost of distinct keys.
- **OpenSearch args:** a pure function of the request. `filters` is a `Vec` built in a fixed order, and nothing order-dependent is formatted.

## Tests run
`cargo test --locked -p router --features v1,deja` was run per target on both this tip and the merge base, and the two sides were compared by failing test name. Across all 22 targets the failure sets are identical, so the PR introduces no failure. Most of those tests fail on both sides at `AppState` construction (Superposition config is read relative to `crates/router`) or on missing connector credentials, and exercise nothing. Run from the repository root instead, `health_check` (a full `AppState` boot against Postgres and Redis) and four more lib tests pass on both sides. `router_env` passes 6/6 on both. `common_utils --features deja` has the same two failures on both sides, and the 13 tests this PR adds all pass.

Pre-existing on main, not this PR: `router/tests/deja_jwt_novel_token_fail_stops.rs:27` hardcodes `policy_version: 1`. deja implements 2 at both the old and the new pin, so the hook refuses the table and the test panics before its assertion. The fix is `policy_version: deja::POLICY_VERSION`. Also, `api_keys::tests::test_hashing_and_verification` races the `payment_attempt` tests' `set_var` under parallel threads.

The five red checks on this tip are all non-code. Spell check is red on main too, and none of its 51 flagged words is in a line this PR adds. Linked-issue and `S-test-ready` are label policy, and the two consolidated Cypress results depend on the label check.

**Verdict:** merge once the lock fix is pushed, the guard tests from finding 1 land, and the doc sentence from finding 4 is corrected. Finding 2, the six sites in finding 4 and the nits can ride along or follow.

crates/router/src/core/payment_methods.rs
maverox: Restored in 5ab697c0d8, at all three sites in this file: `HashSet::from([])`, with the v1-gated import it needs. The rewrite to `Default::default()` came from the first commit, when the facade was an alias and `From<[T; N]>` existed only for std's hasher. The facade is now a type of its own with that impl, so the original spelling compiles again.

One thing a `From` impl cannot do here, checked by trying it: the four other places the migration writes `.into_iter().collect()` each turn a `std::collections::HashMap` (from prost, redis, or a crate that does not use the facade) into ours. The wrapper's `From<std::HashMap<K, V, S>>` is generic over `S` and carries it through unchanged, so `.into()` only converts between maps with the same hasher. Those four cross from `RandomState` to `CorrelationHasher`, which is a rehash: every entry has to be re-inserted under the new keys, and `collect()` is the shortest way to say that. A `From` that rehashed would have to be a separate impl, `From<HashMap<K, V, RandomState>> for HashMap<K, V, CorrelationHasher>`, and that overlaps the generic one when `S = CorrelationHasher`, so the compiler rejects the pair. Left as `collect()`.

maverox: Correction: the facade has since moved to #14368, so the commit is 2fed44f096 on that chain, not 5ab697c0d8. Same change.

crates/router/src/core/webhooks/outgoing_v2.rs
maverox: Done in fad4b3cc6a. The lint no longer targets `in_current_span`. It targets the raw spawn entry points (`tokio::spawn`, `JoinSet::spawn`, plus `spawn_local` / `Handle::spawn` / `Runtime::spawn` for the future). Spawns go through `router_env::spawn` / `spawn_in_set`, which accept only a `tracing::instrument::Instrumented<F>`. So `.in_current_span()` passes without comment, and a bare future fails to compile.

All 26 per-site allows are removed. One allow remains on each helper, plus one on a selenium harness in test_utils, which has no router_env dependency.

PR #8669, opened 2025-07-17
feat(connector): [FISERV] Added PayPal Redirect Payment Method

bsayak03:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

Added PayPal Redirect Payment Flow. First we Create an Order and then in the return_url, customer gets redirected. After that we Capture the funds.

Note: We are getting a missing fields error from the connector, possibly because this particular payment method hasn't been enabled from the connector dashboard.

### Additional Changes

- [ ] This PR modifies the API contract
- [ ] This PR modifies the database schema
- [ ] This PR modifies application configuration/environment variables

## Motivation and Context

Ref: https://developer.fiserv.com/product/CommerceHub/api/?type=post&path=/checkouts/v1/orders&branch=main&version=1.25.1000

## How did you test it?

Payments Create Paypal Redirect

cURL: 

```
curl --location 'http://localhost:8080/payments' \
--header 'Content-Type: application/json' \
--header 'Accept: application/json' \
--header 'api-key: dev_JYKTXJwwDWQuGPW0WjUKFXMJmJP468CKJE4U6qddT8LddLXpll9t8e0yWSLczCVm' \
--header 'Cookie: PHPSESSID=0b47db9d7de94c37b6b272087a9f2fa7' \
--data-raw '{
    "amount": 700,
    "all_keys_required": true,
    "currency": "EUR",
    "connector": [
        "fiserv"
    ],
    "confirm": true,
    "payment_link": false,
    "capture_on": "2029-09-10T10:11:12Z",
    "amount_to_capture": 700,
    "name": "John Doe",
    "email": "dg@example.com",
    "phone": "999999999",
    "phone_country_code": "+1",
    "description": "Its my first payment request",
    "authentication_type": "three_ds",
    "return_url": "https://google.com",
    "payment_method": "wallet",
    "payment_method_type": "paypal",
    "payment_method_data": {
        "wallet": {
            "paypal_redirect": {}
        }
    },
    "browser_info": {
        "color_depth": 24,
        "java_enabled": true,
        "java_script_enabled": true,
        "language": "en-GB",
        "screen_height": 720,
        "screen_width": 1280,
        "time_zone": -330,
        "ip_address": "208.127.127.193",
        "accept_header": "text/html,application/xhtml+xml,application/xml;q=0.9,image/webp,image/apng,*/*;q=0.8",
[...]

PR #14320, opened 2026-09-20
fix(connector): [FISERV] deserialize checkout responses into the checkout variant

### LINKED ISSUE #14321: [BUG] [FISERV] checkout responses deserialize as charge responses, dropping the redirect URL
### Bug Description

`FiservPaymentsResponse` (`crates/hyperswitch_connectors/src/connectors/fiserv/transformers.rs`) is an untagged enum:

```rust
#[serde(untagged)]
pub enum FiservPaymentsResponse {
    Charges(FiservChargesResponse),
    Checkout(FiservCheckoutResponse),
}
```

`FiservChargesResponse` requires only `gatewayResponse` and `paymentReceipt`. `FiservCheckoutResponse` requires those two plus `interactions`, with an optional `order`. Serde ignores unknown fields, so the charges variant matches every checkout response and the `Checkout` variant is never constructed.

### Expected Behavior

A response carrying `interactions` deserializes as `FiservCheckoutResponse`, so `redirection_data` carries `interactions.actions.url` and `connector_metadata.order_id` carries `order.orderId`.

### Actual Behavior

Every response deserializes as `Charges`, and the branch that reads the redirect URL is unreachable:

```rust
FiservPaymentsResponse::Charges(res)  => (res.gateway_response.clone(), None, None),
FiservPaymentsResponse::Checkout(res) => (
    res.gateway_response.clone(),
    Some(res.interactions.actions.url.clone()),
    res.order.as_ref().map(|o| o.order_id.clone()),
),
```

`redirection_data` stays `None` and `connector_metadata.order_id` is null for every redirect payment, so the customer is never sent to the checkout page.

The enum and the redirect flow that depends on it were introduced in the same commit, b0b71935ca (#8669), so this path has not worked since.

### Steps To Reproduce

Deserialize a checkout-shaped body into `FiservPaymentsResponse` and inspect the variant:

```json
{
  "gatewayResponse": {
    "transactionState": "CREATED",
    "transactionProcessingDetails": { "orderId": "ORD-1", "transactionId": "TXN-1" }
  },
  "paymentReceipt": { "approvedAmount": { "total": 12.34, "currency": "USD" } },
  "interactions": {
    "actions": { "type": "REDIRECT", "url": "https://checkout.fiserv.test/redirect/abc" }
  }
}
```

The result is `FiservPaymentsResponse::Charges`, not `Checkout`. A unit test covering both directions is in #14320.

### Context For The Bug

Found by scanning the connector crate for untagged enums whose earlier variant has a required-field set that is a subset of a later variant's fields.

A related question that a live Fiserv response would settle: `FiservOrderStatus` carries `rename_all = "UPPERCASE"`, so the only accepted token for `order.orderStatus` is `PAYERACTIONREQUIRED`, without underscores. If Fiserv sends `PAYER_ACTION_REQUIRED`, `order` fails to deserialize and takes the whole `Checkout` variant with it.

### Environment

Building and running locally.

1. Operating System or Linux distribution: macOS 15 (aarch64)
2. Rust version (output of `rustc --version`): `rustc 1.98.1 (48a229cea 2026-09-01)`
3. App version: tree at 329f7d7d3d3d178c9be1efa08ec39c3eaaf9fc0a

LudwigJMarx:
## Type of Change

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

`FiservPaymentsResponse` is an untagged enum whose first variant, `FiservChargesResponse`, requires only `gatewayResponse` and `paymentReceipt`. A checkout response carries both of those plus `interactions` and `order`, and serde ignores unknown fields, so the charges variant matches every checkout response and `Checkout` is never constructed.

The redirect URL is read only from that variant:

```rust
FiservPaymentsResponse::Charges(res)  => (res.gateway_response.clone(), None, None),
FiservPaymentsResponse::Checkout(res) => (
    res.gateway_response.clone(),
    Some(res.interactions.actions.url.clone()),
    res.order.as_ref().map(|o| o.order_id.clone()),
),
```

So `redirection_data` stays `None` and `connector_metadata.order_id` is null for every redirect payment, and the customer is never sent to the checkout page. The enum and the redirect flow that depends on it were introduced in the same commit, b0b71935ca (#8669), so this path has not worked since.

Listing `Checkout` first resolves it without a discriminator: that variant additionally requires `interactions`, which a charge response does not carry, so the two shapes no longer overlap in either direction. The response carries no tag field, so ordering is the available mechanism.

### Additional Changes

## Motivation and Context

Found by scanning the connector crate for untagged enums whose earlier variant has a required-field set that is a subset of a later variant's fields.

One open question for someone with Fiserv access. `FiservOrderStatus` carries `rename_all = "UPPERCASE"`, so the only token accepted for `order.orderStatus` is `PAYERACTIONREQUIRED`, with no underscores:

```
serde_json::to_string(&FiservOrderStatus::PayerActionRequired) == "\"PAYERACTIONREQUIRED\""
```

If Fiserv sends `PAYER_ACTION_REQUIRED`, `order` fails to deserialize and takes the whole `Checkout` variant with it, which puts the response back on the charges shape and makes this fix ineffective for those payloads. The test payload therefore omits `order`, which is optional. If the wire value does carry underscores, that enum needs `rename_all = "SCREAMING_SNAKE_CASE"`; I left that out of this PR because I cannot verify the value against a live Fiserv response.

## How did you test it?

Unit tests in `fiserv/transformers.rs` covering both directions: a response carrying `interactions` must deserialize as `Checkout`, and a response without it must still deserialize as `Charges`.

Before the change:

```
running 2 tests
test connectors::fiserv::transformers::tests::charges_response_is_parsed_as_a_charge ... ok
test connectors::fiserv::transformers::tests::checkout_response_is_not_parsed_as_a_charge ... FAILED

---- connectors::fiserv::transformers::tests::checkout_response_is_not_parsed_as_a_charge stdout ----
panicked at crates/hyperswitch_connectors/src/connectors/fiserv/transformers.rs:1402:9:
a response carrying `interactions` must deserialize as Checkout, otherwise the redirect URL is dropped

test result: FAILED. 1 passed; 1 failed; 0 ignored; 0 measured; 10 filtered out
```

After the change, the full crate test suite:

```
running 12 tests
test result: ok. 12 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
```

Command used:

```
cargo test -p hyperswitch_connectors --lib --features v1,payouts,frm,revenue_recovery
```

Environment: rustc 1.98.1, cargo 1.98.1, macOS aarch64, tree at 329f7d7d3d.

## Checklist

Closes #14321


XyneSpaces:
**Pre-existing issue (not introduced by this PR):** `FiservOrderStatus` uses `#[serde(rename_all = "UPPERCASE")]`, which maps `PayerActionRequired` to `PAYERACTIONREQUIRED`; a real `PAYER_ACTION_REQUIRED` checkout payload therefore cannot deserialize into the newly prioritized `Checkout` variant and can still fall back to `Charges`, dropping the redirect URL. Please add an explicit `#[serde(rename = "PAYER_ACTION_REQUIRED")]` and include `orderStatus` in the regression fixture.

LudwigJMarx:
Confirmed, and measured rather than assumed. Adding a realistic `order` object to the checkout fixture turns the test red again on the previous commit:

```
test connectors::fiserv::transformers::tests::checkout_response_is_not_parsed_as_a_charge ... FAILED
test result: FAILED. 1 passed; 1 failed; 0 ignored; 0 measured; 10 filtered out
```

`intent` has the same defect. `FiservIntent` accepts only `AUTHORIZE`, so an order created with a capture intent breaks the `Checkout` variant exactly the same way.

Both fields are never read. `FiservOrderStatus` occurs twice in the tree, at its declaration and at its own definition, and the response-side `intent` has no reader either. The consumer takes `order_id` and nothing else. A required field that nothing reads can only reject a response, so d2cae00657 drops both, together with the now unused enum, and the fixture carries `"orderStatus": "PAYER_ACTION_REQUIRED"` and `"intent": "CAPTURE"` as you asked.

I left out `rename = "PAYER_ACTION_REQUIRED"` on purpose. It fixes one of the two fields and pins a value I cannot verify from here:

- The published Fiserv models (`Fiserv/adobe-commerce-plugin`, `Lib/CommerceHub/Model`) cover `/payments/v1/charges` only, with no `orderStatus` and no `interactions`.
- b0b71935ca (#8669), which introduced this flow, records a failed call rather than a successful checkout response, so there is no captured payload in the project either.
- The spelling is likely: the Commerce Hub checkout wraps a PayPal order, PayPal's Orders API uses `PAYER_ACTION_REQUIRED`, and this repository's own paypal connector models that status with `SCREAMING_SNAKE_CASE`.

Likely is not measured, and with the fields gone the deserialization no longer depends on getting it right. If someone has a live Commerce Hub checkout response, that settles whether these fields are worth modelling again with the correct tokens, and I will add them back with the real values.

Test state on the branch:

```
running 12 tests
test result: ok. 12 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
```

PR #14323, opened 2026-09-20
feat(revenue_recovery): [chargebee] populate card info from payment_method_details for wallet payment methods

### LINKED ISSUE #14322: [FEATURE] [Chargebee] consume payment_method_details for wallet payment methods in revenue recovery webhooks
### Feature Description

In the Chargebee revenue recovery webhook flow, `content.transaction.payment_method_details` is a JSON string that Chargebee sends alongside the transaction. Today we only look at it when `content.transaction.payment_method` is one of `card`, `unionpay` or `south_korean_cards`. For every other payment method the string is discarded and `card_info` on `RevenueRecoveryAttemptData` is left as `AdditionalCardInfo::default()`.

Chargebee also populates `payment_method_details` with the underlying card object for wallet payment methods such as `apple_pay`, `google_pay` and `paypal_express_checkout`, since those are frequently funded by a card. Discarding it means we lose `card_network` and `card_isin` for those attempts, which in turn degrades the recovery decisioning that relies on card metadata.

There is a second problem with the current handling: when the payment method *is* a card and the details string fails to deserialize, we return `ConnectorError::WebhookBodyDecodingFailed` and the whole webhook is rejected, even though the card metadata is only supplementary information.

### Possible Implementation

Stop keying the parsing off the Chargebee payment method and instead attempt the card deserialization on whatever `payment_method_details` contains:

- If the string deserializes into the card shape, populate `card_info` with `card_network` and `card_isin`, regardless of whether the payment method is `card`, `apple_pay`, `google_pay` or `paypal_express_checkout`.
- If it does not deserialize (e.g. a PayPal transaction funded by a bank account), leave `card_info` at its default instead of failing the webhook.

`payment_method_sub_type` should keep its current semantics: wallets continue to resolve to their own sub type (`ApplePay`, `GooglePay`, `Paypal`, ...) and only card payment methods derive the sub type from the card's `funding_type`.

NISHANTH1221:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

Card metadata on a Chargebee revenue recovery attempt can arrive from two places, and until now we read only a narrow slice of one of them:

- `content.transaction.payment_method_details`, a JSON string, which was parsed **only** when `content.transaction.payment_method` was `card`, `unionpay` or `south_korean_cards`;
- `content.card`, a sibling of `transaction` / `invoice` / `customer` in the webhook body, which was **not modelled at all** and so was silently discarded by serde.

Neither is guaranteed to be present on a given webhook, and the string is not guaranteed to hold card data when it is present. This PR makes the transformation take whatever is actually available and populate the rest as `None`, instead of erroring or discarding.

**Parsing is no longer keyed off the payment method.** `ChargebeeCardPaymentMethodDetails::parse_card_details` attempts the card shape against whatever `payment_method_details` contains and returns an `Option`. Chargebee populates it with the underlying card object for wallets such as `apple_pay`, `google_pay` and `paypal_express_checkout`, which are commonly card funded, so those now yield card metadata too. A payload that is not card shaped no longer fails the webhook with `ConnectorError::WebhookBodyDecodingFailed`.

**`content.card` is now read as a fallback.** When the string is absent, or present but not card shaped, the card object is used instead. When neither is usable, `card_info` is `AdditionalCardInfo::default()`.

```rust
let card_details = item.content.transaction.payment_method_details
    .as_deref()
    .and_then(ChargebeeCardPaymentMethodDetails::parse_card_details)
    .or(item.content.card);
```

The string is preferred because it is per transaction, whereas `content.card` describes the customer's stored payment source and may not be the card this attempt used.

**Both shapes are accepted.** `payment_method_details` names the network `brand`; the `card` resource names it `card_type`. A `#[serde(alias = "card_type")]` on `brand` lets one struct read either.

**Fields are individually optional.** `ChargebeeCardDetails` previously required `funding_type`, `brand` and `iin` together, so a partial card object failed to deserialize as a whole — relevant for Apple Pay and Google Pay, which may expose a network without a usable IIN since the PAN is a DPAN. Each field is now independent, and the set read has grown from two to six:

| `AdditionalCardInfo` | Chargebee |
| --- | --- |
| `card_network` | `brand` / `card_type` |
| `card_isin` | `iin` |
| `last4` | `last4` |
| `card_exp_month` | `expiry_month`, zero padded |
| `card_exp_year` | `expiry_year` |
| `funding_source` | `funding_type` (`credit` / `debit` / `prepaid`; `not_known` and `not_applicable` map to `None`) |

**Unsupported payment methods fall back to card rather than erroring.** `payment_method_type` uses `enums::PaymentMethod::try_from(..).unwrap_or(enums::PaymentMethod::Card)` and `payment_method_sub_type` ends in `.unwrap_or(enums::PaymentMethodType::Card)`. The `TryFrom<ChargebeeTransactionPaymentMethod> for enums::PaymentMethod` impl is unchanged and still reports `ConnectorError::NotSupported`; the recovery transform absorbs it so an unmapped payment method degrades the attempt rather than dropping the webhook.

`payment_method_sub_type` otherwise keeps its semantics: `payment_method_sub_type()` is consulted first, so wallets resolve to `ApplePay` / `GooglePay` / `Paypal` rather than being reported as `Credit` or `Debit`. Only card payment methods, for which it returns `None`, derive the sub type from `funding_type`.

`ChargebeePaymentMethodDetails` is dropped, since card details are now an `Option<ChargebeeCardDetails>`.

### Additional Changes

## Motivation and Context

Closes #14322.

Dropping this metadata costs the recovery decisioning that depends on card details, and it was being dropped in three separate ways: wallets were never parsed, a partial card object was discarded whole, and `content.card` was never modelled. Failing the entire webhook because the string did not parse, or because Chargebee sent an unmapped payment method, is disproportionate — the recovery attempt is still actionable, and a rejected webhook loses it rather than degrading it.

## How did you test it?

Exercised through the billing connector webhook endpoint. Source resolution:

| `payment_method_details` | `content.card` | `card_info` |
| --- | --- | --- |
| card shaped | either | from the string |
| present, not card shaped | present | from `content.card` |
| present, not card shaped | absent | all `None` |
| absent | present | from `content.card` |
| absent | absent | all `None` |

Payment method mapping:

| `payment_method` | `payment_method_type` | `payment_method_sub_type` |
| --- | --- | --- |
| `card` | `Card` | from `funding_type`, else `Card` |
| `apple_pay` / `google_pay` | `Wallet` | `ApplePay` / `GooglePay` |
| `paypal_express_checkout` | `Wallet` | `Paypal` |
| unmapped (`Other`) | `Card` | from `funding_type`, else `Card` |

Verified locally:

```
cargo +nightly fmt --all
just clippy       # v1 features, passed
just clippy_v2    # v2 + revenue_recovery features, passed
```

## Checklist


swangi-kumari (dismissed):

XyneSpaces:
[should-fix] Two related robustness issues in the Chargebee revenue-recovery webhook parser (`crates/hyperswitch_connectors/src/connectors/chargebee/transformers.rs`):

1. `ChargebeeCardPaymentMethodDetails::parse_card_details` (around the new line 602) silently turns JSON parse failures into `None` via `.ok()`. If Chargebee ever sends malformed card details, we lose the failure signal and fall back to no card info. Propagate the error or at least attach it as context before falling back.

2. `PaymentMethod::try_from(chargebee_payment_method).unwrap_or(enums::PaymentMethod::Card)` (around line 852) defaults an unknown payment method to `Card`. This can misclassify non-card methods (e.g. wallets/bank debits) as cards in recovery records. Use an explicit unsupported/unknown mapping or return an error instead of defaulting to `Card`.

swangi-kumari (approved):

aniketburman014 (approved):

PR #14327, opened 2026-09-21
fix(connector): omit empty CVV in Authorize.net setup

### LINKED ISSUE #14326: [BUG] Authorize.net setup sends cardCode when vaulted CVV is absent
## Bug Description

The Authorize.net setup-mandate adapter unconditionally includes `cardCode` when creating a customer payment profile, even when a saved card retrieved from the Hyperswitch vault has no CVV.

Authorize.net permits this field to be omitted, but a supplied value must contain 3–4 numeric digits. As a result, vault-only registration on another connector account fails at schema validation instead of allowing the gateway to evaluate a registration without CVV.

## Expected Behavior

When CVV is unavailable, omit `cardCode` from the Authorize.net customer-payment-profile request. Preserve a valid, supplied CVV.

This should allow registration to proceed subject to Authorize.net validation mode, merchant settings, and processor rules; omission does not guarantee approval.

## Actual Behavior

A zero-dollar setup using a saved-card token reaches the requested Authorize.net connector account, but the sandbox returns `E00003`, identifying the `cardCode` element as invalid and reporting a pattern-constraint failure. Registration fails.

No CVV was supplied by the calling application. The observed response establishes invalid-field serialization; it does not establish that Authorize.net requires CVV when the field is omitted.

## Steps To Reproduce

1. Configure two Authorize.net merchant connector accounts under the same Hyperswitch merchant/profile, using sandbox credentials.
2. Register and vault a test card for a customer through connector account A.
3. Retrieve a saved-card payment token for that same customer.
4. Create a zero-dollar payment with `payment_type: "setup_mandate"`, `confirm: false`, the same customer/profile, and `setup_future_usage: "off_session"`.
5. Confirm it using the saved-card token, `payment_method: "card"`, explicit single-account routing to connector account B, and valid `mandate_data.customer_acceptance` representing existing offline consent. Do not supply PAN, CVV, or connector A's processor mandate/token.
6. Observe the setup result and the Authorize.net `E00003` error concerning `cardCode`.

The reproduction uses two connector IDs pointing to the same sandbox merchant. The intended use case includes different gateway merchant accounts; the sandbox reproduction does not establish portability between different merchants.

## Context For The Bug

An operator needs to correct a connector selection for a future payment, using the same customer's already-vaulted card. This requires creating a registration on the target gateway account rather than copying a processor token between merchant accounts.

CVV is not retained for later use. Requiring it again prevents unattended registration even where the gateway allows registration without it. Existing customer consent is available; this is not a request to fabricate consent, retain CVV, or bypass gateway verification.

### Source evidence

In [v1.126.0, the setup adapter](https://github.com/juspay/hyperswitch/blob/v1.126.0/crates/hyperswitch_connectors/src/connectors/authorizedotnet/transformers.rs#L483-L490) sets:

```rust
card_code: Some(ccard.card_cvc.clone()),
```

The [field is already optional and skips serialization for None](https://github.com/juspay/hyperswitch/blob/v1.126.0/crates/hyperswitch_connectors/src/connectors/authorizedotnet/transformers.rs#L112-L118). An empty value wrapped in `Some` is not omitted.

### Authoritative documentation

- [Authorize.net live API schema](https://api.authorize.net/xml/v1/schema/AnetApiSchema.xsd): `creditCardType.cardCode` has `minOccurs="0"`; its type is numeric with minimum length 3 and maximum length 4.
- [Authorize.net Customer Profiles API](https://developer.authorize.net/api/reference/features/customer-profiles.html).
- [Authorize.net CIM guide](https://www.authorize.net/content/dam/anet-redesign/documents/CIM_SOAP_guide.pdf): documents cardCode as optional, used for validation, and not stored in the customer profile.
- [PCI SSC guidance on not retaining card verification codes](https://www.pcisecuritystandards.org/faqs/1280/).

## Environment

- Runtime reproduction: Hyperswitch router v1.121.0, Authorize.net sandbox.
- Latest release inspected: [v1.126.0](https://github.com/juspay/hyperswitch/releases/tag/v1.126.0), which retains the unconditional assignment above.
- The latest release has been source-inspected, not runtime-tested for this report.

## Suggested Fix and Regression Coverage

Omit `cardCode` when no genuine CVV is available; preserve supplied valid CVV. Please cover vault-backed setup without CVV, empty internal CVV serialization, and supplied-CVV preservation. An integration check should distinguish schema rejection from merchant-policy rejection.

Issue search did not find a matching report for `authorizedotnet cardCode`.

atali:
## Type of Change

- [x] Bugfix

## Description

Authorize.net setup-mandate requests currently serialize an empty `cardCode` when the vault-backed card has no CVV. Authorize.net allows the field to be absent, but rejects an empty value because its schema requires 3–4 numeric digits when supplied.

Omit `cardCode` only when the internal CVV is empty. Preserve nonempty values unchanged, including leading zeroes, so existing gateway validation still applies. The change is limited to customer payment profile creation; payment authorization, wallet handling, and validation-mode selection are unchanged.

### Additional Changes

## Motivation and Context

Closes #14326.

A customer’s vaulted card may need a new registration on another connector account, with existing customer consent and without retrieving PAN into the calling application or storing CVV. This fix removes the invalid-field serialization failure; the target gateway’s merchant and processor rules still determine whether registration is accepted without CVV.

References:
- [Authorize.net API schema](https://api.authorize.net/xml/v1/schema/AnetApiSchema.xsd): `cardCode` is optional (`minOccurs="0"`), with numeric length 3–4 when supplied.
- [Customer Profiles API](https://developer.authorize.net/api/reference/features/customer-profiles.html).
- [CIM guide](https://www.authorize.net/content/dam/anet-redesign/documents/CIM_SOAP_guide.pdf): cardCode is optional, used for validation, and not stored in the profile.

## How did you test it?

Tests exercise the actual setup request transformer and JSON serialization:
- An unavailable/empty CVV omits the key entirely, in both test and live validation modes.
- Three- and four-digit CVVs, including leading zeroes, are preserved.
- Nonempty malformed values are not silently discarded.
- Customer profile ID, card number, expiration, and validation mode remain intact.

- Before the fix: the empty-CVV regression test fails because `cardCode` is present; the two preservation tests pass.
- `cargo test -p hyperswitch_connectors --features v1,frm,payouts --lib`: **13 passed**, including all three new regression tests.
- `cargo +nightly fmt --all -- --check`: passed.
- `git diff --check`: passed.
- `cargo clippy -p hyperswitch_connectors --features v1,frm,payouts --lib --tests`: completed successfully; no warnings in the changed connector. Existing `common_utils` and dependency future-compatibility warnings remain.

The connector crate requires `frm,payouts` alongside `v1` on the tested upstream revision; a `v1`-only invocation fails on unrelated feature-gated imports.

No live gateway request or deployment was performed for this PR. Sandbox end-to-end validation remains separate from the request-serialization regression coverage.

## Checklist

- [x] I formatted the code with the nightly formatter
- [x] I addressed lints thrown by `cargo clippy` in the changed code


adashantiai:
any news ?

PR #14332, opened 2026-09-21
fix(superposition): validate proxy org and workspace headers against configured scope

### LINKED ISSUE #14331: [BUG] Superposition proxy does not enforce configured org/workspace scope
### Bug Description

The Superposition config proxy (`/v1/superposition/*`) does not apply the scope checks it was intended to apply.

1. `extract_proxy_headers` in `crates/router/src/core/superposition_proxy.rs` reads the `x-org-id` and `x-workspace` headers and checks only that they are present. It does not compare them against the Superposition scope the deployment is configured for.

2. `ListAuditLogsQuery::execute` accepts caller-supplied `dimension[...]` filters but performs no validation on them, unlike `ListContextsQuery::execute`, which calls `require_superposition_context` and `validate_superposition_params`.

The Superposition org/workspace is a separate namespace from the Hyperswitch `organization_id` carried in the JWT, so these headers cannot be checked against a JWT claim. The configured `[superposition]` scope is the trusted source for them.

### Expected Behavior

- `x-org-id` and `x-workspace` are compared against the configured Superposition scope, and a mismatch is rejected.
- Listing endpoints that accept caller-supplied dimension filters require at least one scoping dimension and validate it against the caller's JWT.

### Actual Behavior

- The header values are accepted as supplied and forwarded to the Superposition SDK.
- `/v1/superposition/audit` applies neither `require_superposition_context` nor `validate_superposition_params`.

### Steps To Reproduce

Read `crates/router/src/core/superposition_proxy.rs`:

1. `extract_proxy_headers` returns the header values after a presence check only; no comparison against configured state occurs.
2. `ListAuditLogsQuery::execute` takes `_auth` and calls neither scope validator, whereas `ListContextsQuery::execute` calls both.

### Context For The Bug

The headers were added for the Superposition dashboard embed. The dashboard supplies the correct deployment values, and Superposition trusts the router's server-side token by design, so the router is the component responsible for enforcing scope on these endpoints.

NISHANTH1221:
## Type of Change

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

Constrains the Superposition config proxy to the tenant the router is configured for, and brings the audit-log endpoint in line with the scope checks the other listing endpoints already perform.

**`extract_proxy_headers` validates the proxy headers against the configured scope**

`x-org-id` and `x-workspace` were previously checked only for presence and then forwarded to the Superposition SDK verbatim, while the router authenticates with its own server-side token. They now have to match `[superposition] org_id` / `workspace_id`; a mismatch returns `403 AccessForbidden` naming whichever header was rejected. Missing headers still return `400`.

These headers name a Superposition org/workspace, which is a different namespace from the Hyperswitch `organization_id` in the JWT, so there is no claim to compare them against. Configuration is the trusted source. `SuperpositionClient` already stored this scope and already injects it on the internal write path, so the proxy path now reads it back through two new accessors.

The function now takes `&AppState` and resolves the client internally.

**`ListAuditLogsQuery::execute` validates caller scope**

It now calls `require_superposition_context` and `validate_superposition_params`, the same two guards `ListContextsQuery::execute` already used. Previously it took `_auth` and ran neither, so a request carrying no `dimension[...]` filter was forwarded unfiltered. The dimension parameters were already plumbed through to the SDK in `into_input`; only the validation was missing.

### Additional Changes

No new configuration keys. `[superposition] org_id` and `workspace_id` already exist in `config/development.toml`, `config/docker_compose.toml` and `config/config.example.toml`; this change starts enforcing them on the proxy path.

## Motivation and Context

The proxy endpoints were meant to validate the supplied Superposition org/workspace before proxying, so that the router is not an open proxy onto any tenant reachable with its server-side token. That validation was not implemented when the headers were added for the dashboard embed. `/v1/superposition/audit` was additionally the only listing endpoint accepting caller-supplied dimension filters without validating them.

Closes #14331

## How did you test it?

Build gates on both feature sets:

```
cargo +nightly fmt --all
just clippy       # exit 0
just clippy_v2    # exit 0
```

Behaviour to verify against a running router with the Superposition dashboard embed:

1. A proxy call whose `x-org-id` / `x-workspace` match the configured scope succeeds as before.
2. A proxy call with either header altered returns `403` naming the rejected header.
3. `GET /v1/superposition/audit` without a `dimension[...]` parameter is rejected; with a `dimension[...]` matching the caller's JWT it returns that caller's records.

## Checklist


sahkal (dismissed):

crates/router/src/core/superposition_proxy.rs
```diff
@@ -171,30 +172,47 @@ fn map_superposition_err(
...
+        req.headers()
+            .get(name)
+            .and_then(|value| value.to_str().ok())
+            .map(String::from)
+            .ok_or_else(|| {
+                HttpResponse::BadRequest().json(serde_json::json!({
+                    "error": { "message": format!("missing required header: {name}") }
+                }))
+            })
+    };
```
ThisIsMani: Why a closure here?

crates/router/src/core/superposition_proxy.rs
```diff
@@ -171,30 +172,47 @@ fn map_superposition_err(
...
-        .headers()
-        .get("x-workspace")
-        .and_then(|v| v.to_str().ok())
-        .map(String::from)
-        .ok_or_else(|| {
-            HttpResponse::BadRequest().json(serde_json::json!({
-                "error": { "message": "missing required header: x-workspace" }
-            }))
-        })?;
+    let org_id = header("x-org-id")?;
```
ThisIsMani: Let's name this superposition_org_id, it's causing confusion with our org_id.

crates/router/src/core/superposition_proxy.rs
```diff
@@ -171,30 +172,47 @@ fn map_superposition_err(
...
+    let workspace_id = header("x-workspace")?;
+
+    let org_matches = org_id == superposition_client.configured_org_id();
+    let workspace_matches = workspace_id == superposition_client.configured_workspace_id();
+
+    if !org_matches || !workspace_matches {
+        let resource = match (org_matches, workspace_matches) {
+            (false, false) => "superposition org and workspace",
+            (false, true) => "superposition org",
+            _ => "superposition workspace",
```
ThisIsMani: Remove if and have (true, true) here.

crates/router/src/core/superposition_proxy.rs
```diff
@@ -171,30 +172,47 @@ fn map_superposition_err(
...
-                "error": { "message": "missing required header: x-org-id" }
-            }))
-        })?;
+/// Extract the `x-org-id` and `x-workspace` headers and check them against the
+/// configured Superposition scope. These name a Superposition org/workspace, a
+/// different namespace from the JWT's `organization_id`, so config is the only
+/// trusted source to compare against.
+pub fn extract_proxy_headers(
+    req: &HttpRequest,
+    state: &AppState,
```
ThisIsMani: Can we take superposition client instead?

crates/router/src/core/superposition_proxy.rs
NISHANTH1221: Moved it to a separate function

crates/router/src/core/superposition_proxy.rs
NISHANTH1221: Done

crates/router/src/core/superposition_proxy.rs
NISHANTH1221: Done

crates/router/src/core/superposition_proxy.rs
NISHANTH1221: Inverted the match

ThisIsMani (approved):

sahkal (approved):

sauraww (approved):

ShankarSinghC (approved):

PR #14333, opened 2026-09-21
fix(payments): sum connector latency across retry attempts

prasunna09:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

### Additional Changes

## Checklist


abhisheksharma2411 (commented):
This is a real bug and the fix is correct — the reasoning in the two comments is the part I'd have had to reconstruct myself, so thanks for writing it down rather than leaving it implicit.

I traced the premise rather than taking it on faith, and it holds: `do_retry` takes `router_data` **by value**, hands it to `modify_trackers`, and then rebinds a brand-new one out of `call_connector_service_prerequisites`. So the returned value really is a fresh `RouterData` per attempt and nothing accumulates today — only the final attempt's `external_latency` reaches `payments.rs:1329`, and everything before it silently becomes Hyperswitch time in the API event.

A few things I checked so nobody has to repeat them:

- **All reassignment sites are covered.** `router_data` is only rebound twice inside `do_gsm_actions` (the step-up call and the auto-retry loop), and both are immediately followed by an `accumulate_external_latency`. The single write-back before `Ok(router_data)` is the right shape — one exit, one assignment, and the `GsmDecision::DoDefault => break` path still passes through it.
- **The `v1` gate is correct.** There's only one `do_gsm_actions` and it's already `#[cfg(feature = "v1")]`; there's no v2 twin carrying the same gap.
- **Payouts does *not* have the same bug.** `do_gsm_single_connector_actions` / `do_gsm_multiple_connector_actions` in `payouts.rs` retry the same way, so I expected a matching defect — but nothing in the payouts path ever reads `external_latency`, so there's nothing to accumulate. Worth saying explicitly, because "the sibling retry helper has the same shape" is the obvious follow-up question and the answer is no.

Two things I'd like to see before this merges, neither of them about the logic:

**1. A unit test on `accumulate_external_latency`.** It's a free function with no async, no state and no connector — the cheapest possible test — and it encodes a decision that isn't self-evident: `None + Some(x) = Some(x)`, i.e. an attempt that never recorded a latency is treated as absent rather than as zero. That's the right call, but it's exactly the kind of thing a later refactor flips to `unwrap_or(0)` without noticing. `crates/router/src/core/payments/helpers.rs` already carries a `#[cfg(feature = "v1")] #[cfg(test)] mod tests`, so the pattern is there:

```rust
#[cfg(feature = "v1")]
#[cfg(test)]
mod tests {
    use super::*;

    #[test]
    fn accumulates_external_latency_across_attempts() {
        assert_eq!(accumulate_external_latency(None, None), None);
        assert_eq!(accumulate_external_latency(None, Some(120)), Some(120));
        assert_eq!(accumulate_external_latency(Some(120), None), Some(120));
        assert_eq!(accumulate_external_latency(Some(120), Some(80)), Some(200));
    }
}
```

**2. The PR description is empty.** No Motivation/Context, no test plan, and every checklist box unticked including the type of change. The behaviour change here is invisible in logs and lands in billing-adjacent telemetry — the reasoning you already wrote in the code comments would carry over almost verbatim and save a reviewer the trace I just did.

One judgement call worth stating out loud so it's a decision and not an accident: summing assumes attempts are sequential. They are here, so the sum is the honest number for "how long did this payment spend waiting on connectors". If retries ever go concurrent, summing would overstate the wall-clock share and this helper is where that assumption lives — which is another small argument for the test.

Logic-wise I have no objections.

crates/router/src/core/payments/retry.rs
```diff
@@ -102,6 +111,12 @@ where
         false
     };
 
+    // Total outbound-call time for this payment, summed across every attempt. Each retry builds a
+    // fresh `RouterData` whose `external_latency` starts at `None`, so without accumulating here
+    // only the final attempt would survive into the API event — and `latency - hs_latency` would
+    // then bill every earlier attempt's connector wait to Hyperswitch.
```
kashif-m: Make this concise

crates/router/src/core/payments/retry.rs
```diff
@@ -231,13 +249,23 @@ where
...
                     retries = retries.map(|i| i - 1);
                 }
                 storage_enums::GsmDecision::DoDefault => break,
             }
             initial_gsm = None;
         }
     }
+
+    // Handed back once, at the single exit: the whole payment's outbound-call time rather than
+    // just the last attempt's.
```
kashif-m: Make it concise

hrithikesh026 (approved):

kashif-m (approved):

PR #14325, opened 2026-09-21
feat(connector): add complete Stripe Checkout support

854771076:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

Adds opt-in Stripe hosted Checkout support for one-time payments and native Stripe subscriptions while preserving the existing PaymentIntent flow as the default.

The subscription path validates the configured Stripe price, amount, and currency; binds a successfully confirmed Checkout payment to the native Stripe subscription; and supports customer creation, pause, resume, cancellation, invoice renewals, and failed-payment recovery through Stripe Billing.

Webhook handling accepts rotated Stripe signatures, enforces timestamp tolerance, and applies idempotent, monotonic subscription and invoice state transitions. The change also includes the database migrations required by the updated invoice and subscription models, including transaction-safe concurrent indexes.

This replaces #14324 and includes the requested flow-level success, connector-failure, and edge-case coverage.

### Additional Changes

- [x] This PR modifies the API contract
- [x] This PR modifies the database schema
- [ ] This PR modifies application configuration/environment variables

API changes:
- `api-reference/v1/openapi_spec_v1.json`
- `api-reference/v2/openapi_spec_v2.json`

Database migrations:
- `migrations/2026-09-16-120000_add_subscription_connector_id_index`
- `migrations/2026-09-18-120000_add_invoice_billing_period_end`
- `migrations/2026-09-18-120100_add_invoice_billing_period_end_index`

## Motivation and Context

Merchants need a Stripe-hosted payment flow that can initiate a native subscription and keep Hyperswitch synchronized with recurring invoices and lifecycle changes.

Closes #14250.

## How did you test it?

- `cargo check -p subscriptions --features v1,fred`
- `cargo check -p router --no-default-features --features v1,fred`
- `cargo test -p subscriptions --features v1,fred --lib` (11 passed)
- `cargo test -p hyperswitch_connectors --features v1,frm,payouts --lib checkout` (12 passed)
- `cargo test -p hyperswitch_connectors --features v1,frm,payouts --lib subscription_tests` (14 passed)
- `cargo test -p hyperswitch_connectors --features v1,frm,payouts --lib webhook_signature_tests` (4 passed)
- `cargo fmt --all -- --check`
- `diesel migration run --locked-schema`
- `diesel migration redo --locked-schema --all`
[...]

PR #14336, opened 2026-09-21
feat(connector): add complete Stripe Checkout support

### LINKED ISSUE #14250: feat(connector): support Stripe hosted Checkout
### Feature request

Add an opt-in Stripe hosted Checkout flow while preserving the existing direct PaymentIntent integration as the default.

The hosted flow should create and synchronize Checkout Sessions, support success/cancel redirects and promotion-code configuration, map Checkout webhooks, and preserve monotonic payment state when Checkout Session and PaymentIntent events arrive out of order.

A corresponding implementation is available in PR #14249.

854771076:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

Adds opt-in Stripe hosted Checkout support for one-time payments and native Stripe subscriptions while preserving the existing PaymentIntent flow as the default.

The subscription path validates the configured Stripe price, amount, currency, and Stripe price identifier; binds a confirmed Checkout payment to the native Stripe subscription; and supports customer creation, pause, resume, cancellation, invoice renewals, and failed-payment recovery through Stripe Billing.

Webhook handling accepts rotated Stripe signatures, enforces timestamp tolerance, and applies idempotent, monotonic subscription and invoice state transitions. Required invoice and subscription database migrations, rollback statements, and concurrent indexes are included.

This is a fresh PR from an audited branch with a single commit. It supersedes #14325.

### Additional Changes

API changes:
- `api-reference/v1/openapi_spec_v1.json`
- `api-reference/v2/openapi_spec_v2.json`

Database migrations:
- `migrations/2026-09-16-120000_add_subscription_connector_id_index`
- `migrations/2026-09-18-120000_add_invoice_billing_period_end`
- `migrations/2026-09-18-120100_add_invoice_billing_period_end_index`

## Motivation and Context

Merchants need a Stripe-hosted payment flow that can initiate a native subscription and keep Hyperswitch synchronized with recurring invoices and lifecycle changes.

Closes #14250.

## How did you test it?

- `cargo check -p subscriptions --features v1,fred`
- `cargo check -p router --no-default-features --features v1,fred`
- `cargo test -p subscriptions --features v1,fred --lib` (11 passed)
- `cargo test -p hyperswitch_connectors --features v1,frm,payouts --lib checkout` (12 passed)
- `cargo test -p hyperswitch_connectors --features v1,frm,payouts --lib subscription_tests` (14 passed)
- `cargo test -p hyperswitch_connectors --features v1,frm,payouts --lib webhook_signature_tests` (4 passed)
- `git diff --check`
- OpenAPI JSON parsing and component-reference validation for V1 and V2
- `diesel migration run --locked-schema`
- `diesel migration redo --locked-schema --all`
- Migration compatibility validation (0 breaking changes, 0 warnings)

## Checklist

Modified Rust sources were checked with the available stable `rustfmt`; the local environment did not have the nightly toolchain installed.


XyneSpaces:
[should-fix] `converge_stripe_invoice` in `crates/subscriptions/src/webhooks.rs:122` caps retries at 3 and then returns `Ok(current)`, silently completing the webhook when the conditional update never converged. This can record a payment state that never actually matched the invoice.

```suggestion
// After exhausting retries, propagate an error (or route to reconciliation)
// instead of returning Ok(current).
```

XyneSpaces:
[should-fix] `converge_subscription_from_stripe_invoice` in `crates/subscriptions/src/webhooks.rs:164` caps retries and then returns `Ok(())` when convergence fails, losing the conflict signal. The webhook will look successful even though subscription state was never reconciled.

```suggestion
// Return an error (or enqueue for reconciliation) once retries are exhausted.
```

XyneSpaces:
[should-fix] The catch-all `_ => false` in `should_apply_stripe_invoice_status` at `crates/subscriptions/src/webhooks.rs:32` ignores unrecognized status transitions. A new Stripe status will be silently skipped instead of logged or rejected.

```suggestion
// Return an error for unexpected status transitions, or at least log them.
```

XyneSpaces:
[should-fix] The catch-all `_ => None` in `subscription_status_from_renewal_invoice` at `crates/subscriptions/src/webhooks.rs:64` silently drops unknown invoice/subscription status pairs, making it impossible to detect new Stripe statuses.

```suggestion
// Return a WebhookError for unrecognized status pairs instead of defaulting to None.
```

XyneSpaces:
[should-fix] The added `_ => None` arm in `extract_connector_metadata` at `crates/router/src/core/payments/operations/payment_response.rs:2766` ignores any new `PaymentsResponseData` variant that carries metadata. Future variants will silently lose their metadata.

```suggestion
// Explicitly handle each variant, or return an error when an unexpected variant is seen.
```

XyneSpaces:
[should-fix] In `crates/hyperswitch_connectors/src/connectors/stripe.rs:1048`, when the PSync response id starts with `cs_` the code mutates `connector_metadata` but skips assigning `integrity_object`. Integrity checks are therefore not populated for checkout-session PSync responses.

```suggestion
// Populate integrity_object for checkout-session PSync responses as well,
// or document why it is intentionally omitted.
```

XyneSpaces:
[blocking] Stripe Checkout amount integrity uses `amount_subtotal`

`crates/hyperswitch_connectors/src/connectors/stripe.rs` (around lines 1031 and 1274) builds the authorize/sync integrity object from `response.amount_subtotal`. The field actually charged by Stripe is `amount_total` (subtotal excludes tax, shipping, and discounts), so the integrity object will mismatch the captured amount whenever those adjustments apply. Use `amount_total` for the integrity object.

[blocking] PSync `cs_*` fallback drops `integrity_object` and overwrites `connector_metadata`

In the same file (around line 1082), when a `cs_*` ID resolves via `PaymentIntentSyncResponse`, the `if x.starts_with(STRIPE_CHECKOUT_SESSION_ID_PREFIX)` branch replaces `connector_metadata` with a minimal `{"stripe_checkout_session_id": x}` object and never assigns `router_data.request.integrity_object`. Preserve existing metadata and set `integrity_object` from the PaymentIntent amount in this branch as well.

[should-fix] New `schema.rs` columns lack a Diesel migration

`crates/diesel_models/src/schema.rs` adds `billing_period_end` to `invoice` and `last_applied_billing_period_end` to `subscription`, but no `up.sql`/`down.sql` migration is present in the diff. Existing databases will not have these columns at runtime.

XyneSpaces:
## Review findings for PR #14336

**PR:** [#14336 feat(connector): add complete Stripe Checkout subscription flows](https://github.com/juspay/hyperswitch/pull/14336)  
**Head:** `dd415714c521f86887f0160bad30e8bd0531edf6`

> **Note:** Inline review comments could not be posted because the current tool set does not include a pull-request review-comment writer. Findings below reference the diff lines they apply to.

---

### :rotating_light: Silent suppression of webhook HMAC verification errors

**File:** `crates/hyperswitch_connectors/src/connectors/stripebilling.rs`  
**Location:** `StripeBillingWebhookVerifier::verify_signature`, ~line 76

```rust
Ok(signatures.chunks_exact(32).any(|signature| {
    common_utils::crypto::HmacSha256
        .verify_signature(secret, signature, message)
        .unwrap_or(false)
}))
```

`verify_signature` returns `CustomResult<bool, CryptoError>`. `.unwrap_or(false)` collapses every crypto failure (malformed signature, library error, bad secret length) into “not valid,” making tampered payloads and operational errors indistinguishable.

**Fix:** propagate the crypto error unless at least one chunk verifies successfully:

```suggestion
signatures
    .chunks_exact(32)
    .try_fold(false, |matched, signature| {
        if matched {
            Ok(true)
        } else {
            common_utils::crypto::HmacSha256
                .verify_signature(secret, signature, message)
        }
    })
```



---

### :rotating_light: Subscription-create response silently defaults to USD / zero amount

**File:** `crates/hyperswitch_connectors/src/connectors/stripebilling/transformers.rs`  
**Location:** `TryFrom<ResponseRouterData<…SubscriptionCreate…>>`, ~lines 276–285

```rust
let currency_code = invoice_details
    .as_ref()
    .map(|invoice| invoice.currency_code)
    .unwrap_or(common_enums::Currency::USD);

let total_amount = invoice_details
    .as_ref()
    .map(|invoice| invoice.total)
    .unwrap_or(MinorUnit::new(0));
```

When `latest_invoice` is not expanded (e.g., `collection_method=send_invoice`), the connector stores `USD` and `0` instead of failing. This corrupts the subscription ledger.

**Fix:** do not default. Return `ConnectorError::MissingRequiredField { field_name: "latest_invoice" }` when the invoice is absent.



---

### :warning: Currency parse failures discard the original cause

**File:** `crates/hyperswitch_connectors/src/connectors/stripebilling/transformers.rs`  
**Location:** `parse_currency`, ~line 124

```rust
fn parse_currency(value: &str) -> Result<common_enums::Currency, errors::ConnectorError> {
    common_enums::Currency::from_str(&value.to_uppercase())
        .map_err(|_| errors::ConnectorError::ResponseDeserializationFailed)
}
```

The original parse error is discarded, so production logs cannot show which unsupported currency string was received.

**Fix:** preserve the cause:

```suggestion
common_enums::Currency::from_str(&value.to_uppercase())
    .change_context(errors::ConnectorError::ResponseDeserializationFailed)
    .attach_printable(format!("unsupported currency from Stripe Billing: {value}"))
```



---

### :warning: Price/estimate responses default missing `unit_amount` to zero

**File:** `crates/hyperswitch_connectors/src/connectors/stripebilling/transformers.rs`  
**Locations:** `TryFrom<ResponseRouterData<…GetSubscriptionItemPrices…>>` (~line 138) and `TryFrom<ResponseRouterData<…GetSubscriptionEstimate…>>` (~line 178)

```rust
amount: price.unit_amount.unwrap_or(MinorUnit::new(0)),
// and
let amount = item.response.unit_amount.unwrap_or(MinorUnit::new(0));
```

A missing `unit_amount` becomes `0`, silently corrupting pricing responses.

**Fix:** fail with `MissingRequiredField` instead of defaulting to zero.

```suggestion
let amount = item
    .response
    .unit_amount
    .ok_or(errors::ConnectorError::MissingRequiredField {
        field_name: "unit_amount".into(),
    })?;
```



---

**Summary:** 2 critical, 2 warning. No `unwrap()`/`expect()`/`panic!()` in non-test code, no `TODO`/`FIXME`, and no missing `#[instrument(skip_all)]` on new router async functions were detected in the diff.

PR #14338, opened 2026-09-21
chore(deja): gate the fork-naming convention with a clippy lint

maverox:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [x] CI/CD

## Description

#14201 named eighteen concurrent forks so that a transposed pair stops being
scored as two behaviour changes. Nothing stopped the nineteenth landing with
`.in_current_span()` again.

`crates/router/tests/detached_spawn_inventory.rs` cannot catch it — it accepts
`.in_current_span()` and `.instrument(...)` equally — and no CI job runs tests
(the `nextest` lines in `CI-push.yml` are commented out, and the only real
`cargo test` calls live in `.github/workflows/archive/`). `just clippy` does use
`--all-targets`, so that test file is *compiled* in CI, which is the trap: it
looks covered. Compiling an assertion does not run it.

So the convention becomes a `disallowed-methods` entry — the same shape the
clock and entropy seams already use. A spawn either names its fork or carries an
`#[allow]` saying why it does not, and a reviewer sees the choice.

**The path must be the canonical `tracing::Instrument::in_current_span`.** Every
file here imports the trait as `tracing_futures::Instrument`, and an entry
spelled that way resolves to nothing. Clippy does not warn about an unresolvable
`disallowed-methods` path, so that version would have shipped looking like a
working guard while enforcing nothing.

### The 26 sites

| group | n | why it is allowed |
|---|---|---|
| process-lifetime | 14 | drainer, both scheduler loops, the router's scheduler binary, `receiver_for_error`, the redis and storage pools — spawned outside any request, so there is no correlation to lose and no sibling to be transposed with |
| test-only | 1 | `router/tests/utils.rs` |
| lone detached spawn | 11 | webhooks, refunds, disputes, payment methods, user lineage — they carry the request span so their boundaries stay correlated, but have no racing sibling to be paired against |

The eleven are the interesting ones. Giving them an identity of their own needs
a named-fork API from deja: `fork_span()` hardcodes a single span name, so every
fork in a correlation shares one path and is separable only by a creation-order
`fork_seq`; and a host cannot replicate `spawn_fork` itself, because
`capture_current` / `scope_snapshot` are not exported. If that API lands, those
eleven `#[allow]`s are the conversion list.

### Why the count is a union of four configurations

The two redis backends lint **different files** — `redis_rs.rs:180` under
redis-rs, `fred.rs:107` under fred — and `payment_methods.rs` and
`outgoing_v2.rs` only compile under v2. A single-config sweep undercounts, and
since `clippy_v2` runs with `-D warnings`, an undercount is a red build rather
than a warning.

### Additional Changes

## Motivation and Context

Keeps the fork-naming fix from #14201 from silently regressing, and gives a
deterministic answer to "are all spawns covered?" that grep cannot — grep counts
comments and cannot see feature-gated code.

## How did you test it?

All four CI configurations, unpiped, exit code taken from the command itself:

| gate | exit | diagnostics |
|---|---|---|
| `just clippy` | 0 | 0 |
| `just clippy fred` | 0 | 0 |
| `just clippy_v2` | 0 | 0 |
| `just clippy_v2 fred` | 0 | 0 |

The diff is pure insertion: `.clippy.toml` plus 26 allows across 21 files, no
lines removed.

## Checklist


maverox:
Superseded by #14311 — the same commit now sits there (`e79576f25e`), which is where this belonged. This PR was opened against `main` only because #14201's branch had been deleted on merge and I created a new one instead of asking which existing PR to use.

Verified on #14311's tip after rebasing onto `ce00e1c704`: `just clippy`, `just clippy fred`, `just clippy_v2`, `just clippy_v2 fred` — all exit 0, zero diagnostics.

PR #14344, opened 2026-09-21
feat: add merchant alert read permission

VenuMadhav2541:
## Type of Change

- [x] New feature

## Description

Adds the `Alert` resource and its read-only `MerchantAlertRead` permission. The existing `internal_admin` and `internal_view_only` roles receive the `AlertsView` permission group. The group is hidden from permission listings and cannot be assigned to a custom role, so ordinary roles do not gain alert access.

The alerts backend uses the existing `POST /user/internal/authorize` endpoint with the Control Center JWT and `MerchantAlertRead`. This PR adds no alert-specific authorization endpoint or role.

## Motivation and Context

Control Center is moving the Merchant Alerts screen into its shell. Hyperswitch must decide whether each JWT may read its data, while keeping alert read access with the existing internal roles.

## How did you test it?

- `cargo check -p router --features olap --offline -q`
- Called the existing authorizer against the local router with three local JWTs: internal view and internal admin returned HTTP 200 with an empty body; a normal user returned HTTP 403.

```sh
curl -i -X POST http://127.0.0.1:8080/user/internal/authorize \
  -H 'Content-Type: application/json' \
  --data '{"token":"<Control Center JWT>","permission":"MerchantAlertRead"}'
```

## Checklist

- [x] I formatted the code


XyneSpaces:
[should-fix] `get_alert_read_access` in `crates/router/src/core/user.rs:221` is missing `#[instrument(skip_all)]`, so this new async core flow won't be traced consistently with the surrounding user-core functions.

```rust
#[instrument(skip_all)]
pub async fn get_alert_read_access(
    state: SessionState,
    user_from_token: auth::UserFromToken,
) -> UserResponse<serde_json::Value> {
```

crates/router/src/services/authorization/permissions.rs
```diff
@@ -158,6 +162,7 @@ pub fn get_resource_name(resource: Resource, entity_type: EntityType) -> Option<
         (Resource::ReconRule, _) => Some("Recon Rules"),
         (Resource::SuperpositionConfig, _) => Some("Superposition Configs"),
         (Resource::Offers, _) => Some("Offers"),
+        (Resource::Alert, _) => None,
```
ThisIsMani, line 165: Should be Alerts

crates/router/src/services/authorization/permissions.rs
```diff
@@ -119,6 +119,10 @@ generate_permissions! {
             scopes: [Read, Write],
             entities: [Profile]
         },
+        Alert: {
+            scopes: [Read],
```
ThisIsMani: Add Write also for this.

crates/router/src/services/authorization/permission_groups.rs
```diff
@@ -202,6 +205,7 @@ impl PermissionGroupExt for PermissionGroup {
             | Self::ReconTransactionsManage
             | Self::ReconRulesView
             | Self::ReconRulesManage => RoleProductCategory::Recon,
+            Self::AlertsView => RoleProductCategory::Dashboard,
```
ThisIsMani: Should be part of Orchestration.

ThisIsMani: Changed both alert permission groups to `RoleProductCategory::Orchestration` in daa98dc0f0 so they are not inherited by Recon through the shared Dashboard category.

crates/router/src/services/authorization/permissions.rs
ThisIsMani: Added `Alert` write scope and an `AlertsManage` group in daa98dc0f0. Internal admin receives view/manage; internal view-only remains read-only.

crates/router/src/services/authorization/permissions.rs
ThisIsMani: Checked the Control Center flow: the custom-role screen consumes `GET /user/role/parent/list`, and `get_descriptions_for_groups` drops the entire parent when `get_resource_name` returns `None`. Clone Connector is hidden through this exact mechanism (`Resource::CloneConnector => None`), with the legacy listing also hidden by its `None` group/parent descriptions. Alerts are internal-role-only and rejected by `validate_role_groups`, so keeping this as `None` avoids showing users a permission they cannot assign.

ThisIsMani (approved):

ShankarSinghC (approved):

PR #14347, opened 2026-09-21
feat(connector): [Shift4] route external 3DS, MIT via NTID and setup mandate through UCS

shuklatushar226:
## Summary
Hyperswitch side of the Shift4 UCS connector work: juspay/hyperswitch-prism#2340. Without these changes the new UCS flows cannot be reached from HS.

| Change | Unblocks |
|---|---|
| `Connector::Shift4` moved to the true arm of `is_separate_authentication_supported` | external 3DS |
| Shift4 3DS settle Authorize reads the `tok_…` token from the redirect response and sends it as `authentication_data.threeds_server_transaction_id`; Authorize gate continues inline when no redirect is returned | 3DS |
| `shift4` added to `network_transaction_id_supported_connectors` (development, sandbox, integration_test) | MIT via NTID (RepeatPayment) |
| `should_call_connector_customer` for Shift4 when `setup_future_usage = off_session` | SetupMandate / mandate CIT |
| Dispute webhook object reference prefers `connector_transaction_id` over `connector_dispute_id` | dispute webhooks |
| Cypress: Shift4 mandate/MIT cases un-skipped, SetupMandate expects 200, `shift4` in `AUTH_CODE`, no auth_code assertion while `requires_customer_action` | test coverage |

## Review notes
- **`gateway.rs` change is not Shift4-specific**: it flips id precedence in `event_reference_to_object_ref` for every connector's dispute webhooks. Needs a reviewer's call on whether to gate it per connector.
- **Shift4 redirect-token lift** reads `token` from the redirect query/payload and only accepts values starting with `tok_`; the value is client-supplied (shopper browser return), so it is untrusted input passed to UCS.
- IncomingWebhook is not proven end to end from HS: HS Cypress has no Shift4 webhook fixture.
- Excluded on purpose: local GRACE evidence recording and local port-rewrite changes to `cypress.config.js`, `support/e2e.js`, `support/redirectionHandler.js`.

## Testing
HS → UCS → Shift4 sandbox Cypress runs passed for 3DS, PSync, Capture, Void, Refund/RSync, SetupMandate and RepeatPayment (evidence in the UCS PR).


XyneSpaces:
[should-fix] The change in `crates/router/src/core/webhooks/gateway.rs` makes dispute object-reference resolution prefer `connector_transaction_id` over `connector_dispute_id` globally. For connectors whose dispute webhook lookup still relies on `connector_dispute_id`, this changes the payment lookup key and can make valid dispute webhooks resolve to the wrong object or fail. Scope the precedence change to Shift4 (or preserve the existing connector-specific behavior) and add coverage for the affected webhook paths.

XyneSpaces:
[blocking] Dispute webhook reference precedence changed for every connector

`crates/router/src/core/webhooks/gateway.rs` (around line 941) now prefers `connector_transaction_id` over `connector_dispute_id` for all connectors, not just Shift4. Connectors that publish only `connector_dispute_id` in dispute webhooks will fail object-reference matching. Gate this behavior to Shift4 only, or restore the previous default.

[blocking] Shift4 3DS token is read from untrusted redirect input without validation

`crates/router/src/core/unified_connector_service/transformers.rs` (around line 915) extracts the `token` query/body parameter from the shopper redirect and forwards it to UCS as `threeds_server_transaction_id`. Only the `tok_` prefix is checked; there is no length/format validation, URL decoding, or signature verification. Validate the token before passing it upstream, and fail with a clear 4xx/connector error when it is missing or malformed instead of silently proceeding with `None`.

XyneSpaces:
Two things to tighten up before this lands:

1. **Config coverage for Shift4 NTID support** – the PR adds `shift4` to `[network_transaction_id_supported_connectors]` in `integration_test.toml`, `sandbox.toml`, and `development.toml`, but `config.example.toml` and `production.toml` are left unchanged. If NTID-based MIT is meant to work for local or production deployments, those files need the same entry.

2. **Unrelated dispute webhook reference change** – `crates/router/src/core/webhooks/gateway.rs` now prefers `connector_transaction_id` over `connector_dispute_id` when resolving a dispute webhook’s object reference. This doesn’t look related to the Shift4 feature described. Please confirm it’s intentional and tested; if it’s an orthogonal fix, consider splitting it into a separate PR.

XyneSpaces:
[blocking] Please scope the new `.connector_transaction_id.or(dispute.connector_dispute_id)` precedence in `crates/router/src/core/webhooks/gateway.rs` (added lines 943–945) to the Shift4 path, or prove that this is a deliberate global contract change. For connectors that populate both fields with different meanings, this can associate an incoming dispute webhook with the wrong payment/dispute record.

PR #13928, opened 2026-08-31
test(cypress): BlockImplicitCustomerCreation for adyen

HS-QUALIBOT:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [x] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

This PR adds Cypress E2E test coverage for the **Block Implicit Customer Creation** feature. A new spec file (`57-BlockImplicitCustomerCreation.cy.js`) is added with 3 test contexts covering the happy path, negative case (blocking via superposition config), and edge case (resetting the block config). Config keys `BlockImplicitCustomerCreationAllowed` and `BlockImplicitCustomerCreationBlocked` are added to `Commons.js` and `Adyen.js`, and the `BLOCK_IMPLICIT_CUSTOMER_CREATION` connector list is registered in `Utils.js` for the adyen connector.

### Additional Changes

- [ ] This PR modifies the API contract
- [ ] This PR modifies the database schema
- [ ] This PR modifies application configuration/environment variables

## Motivation and Context

This coverage is required for the feature introduced in [PR #13561](https://github.com/juspay/hyperswitch/pull/13561), which introduces a superposition config for blocking implicit customer creation. This config is scoped at org level, defaulting to false. Payment creation now uses this configuration instead of PM Modular eligibility when deciding whether a missing customer can be created implicitly. The tests verify that payments succeed when implicit creation is allowed, fail with `HE_02` when blocked, and succeed again after the block is reset.

Closes #13927

Related to #13561

## How did you test it?

Full regression suite was executed against the changed connector(s). All `RUNNER_RESULT` blocks from the QA pipeline are included verbatim below.

### Changed-spec verification — `adyen`

```
RUNNER_RESULT:
  Connector: adyen
  SpecFile: cypress/e2e/spec/Payment/57-BlockImplicitCustomerCreation.cy.js
  PrereqsUsed: spec/Payment/01-,02-,03-
  TotalTests: 9
  Passed: 9
  Failed: 0
  Skipped: 0
  OverallStatus: PASS
  FlakeyTests:
    - TestName: reset block config, create payment with non-existent customer (expect 200)
      Note: First run crashed (Electron renderer memory exhaustion). Second run passed (298s). Recommend hardening polling approach.
```

### Summary

| Connector | Specs Run | Passed | Failed | Skipped | Status |
|---|---|---|---|---|---|
[...]

PR #14351, opened 2026-09-22
fix(ci): pass superposition credentials to cypress test jobs

### LINKED ISSUE #14352: Superposition-gated Cypress specs silently skip all tests in CI
## Problem
PR #13928 introduced Cypress specs gated on Superposition config overrides (e.g. `54-BlockImplicitCustomerCreation.cy.js` for adyen), but the Cypress CI workflow never set the `SUPERPOSITION_*` environment variables. `cypress-tests/cypress/utils/State.js` seeds `globalState` from those env vars, so in CI `superpositionBaseUrl` / `superpositionSecret` / `superpositionAuthToken` stay undefined, the specs' `before` hooks set `specShouldSkip = true`, and **every test in the spec is silently skipped** — the suite reports green while testing nothing.

A second silent failure mode hides underneath: even without the before-hook skip, `createSuperpositionOverrides` in `cypress-tests/cypress/support/commands.js` logs "Superposition credentials not set ... skipping config set" and returns when the credentials are absent.

## Affected specs
- `cypress-tests/cypress/e2e/spec/Payment/42-RequiresCVV.cy.js`
- `cypress-tests/cypress/e2e/spec/Payment/54-BlockImplicitCustomerCreation.cy.js`
- `cypress-tests/cypress/e2e/spec/Payment/54-ConnectorAgnosticMandates.cy.js`

The jobs already run the `superposition-demo` service on `localhost:8081` and seed it (`Seed Superposition` step, org `localorg` / workspace `dev`), and the router polls it (`config/development.toml` `[superposition]`, with `ROUTER__SUPERPOSITION__POLLING_INTERVAL=10` in CI) — only the Cypress-side credentials were never wired through.

## Fix
Pass the credentials (mirroring `config/development.toml` `[superposition]`) to the Cypress runs. Note the org/workspace must be `localorg`/`dev` — `commands.js` otherwise defaults to `hyperswitch`, and overrides would land in a context the router never polls.

vasanthp-jus:
Fixes #14352

## Summary
- PR #13928 added superposition-gated Cypress specs, but `.github/workflows/cypress-tests-runner.yml` never set the `SUPERPOSITION_*` env vars. `cypress-tests/cypress/utils/State.js` seeds `globalState` from those env vars, so in CI they stay unset, the specs' `before` hooks set `specShouldSkip`, and **every test in those specs silently skips** — the suite stays green while testing nothing.
- Affected specs: `42-RequiresCVV.cy.js`, `54-BlockImplicitCustomerCreation.cy.js`, `54-ConnectorAgnosticMandates.cy.js` (adyen runs in the `mock-payments-batch-2` MITM job, which has the superposition service and seed step but never passed the credentials to the test run).
- Fix: declare the credentials in the workflow-level `env:` block so every Cypress job that runs against the `superposition-demo` service at `localhost:8081` inherits them, mirroring `config/development.toml` `[superposition]`.

## Details
- Values must be non-empty (`State.js` / `commands.js` guard on them). The demo image does not enforce auth, but the token keeps parity with the router config. It is the same public dev token already committed in `config/development.toml`.
- `SUPERPOSITION_ORG_ID=localorg` / `SUPERPOSITION_WORKSPACE_ID=dev` are required: `commands.js` defaults them to `hyperswitch`, and overrides created in the wrong org/workspace would land in a context the router (which polls `localorg`/`dev`) never reads.
- Also removes the second silent failure mode: `createSuperpositionOverrides` logs "skipping config set" and returns when credentials are absent.

## Testing
- [ ] CI on this branch: the `Run MITM mock Cypress tests (mock-payments-batch-2)` job should now **execute** the adyen cases of `54-BlockImplicitCustomerCreation.cy.js` instead of skipping them (look for `Superposition config set` in the logs / mochawesome report).


XyneSpaces:
[blocking] Please remove the plaintext `SUPERPOSITION_SECRET` and `SUPERPOSITION_AUTH_TOKEN` values added in `.github/workflows/cypress-tests-runner.yml` (added lines 45–46) and reference repository/org secrets instead. These values are exposed to anyone with repository history access and should be rotated if they are active.

vasanthp-jus:
**Addressing review feedback: removed the plaintext `SUPERPOSITION_SECRET` / `SUPERPOSITION_AUTH_TOKEN` values** (8d89cb37d2).

Rather than pinning the values or adding repo secrets, the `Fetch keys` step now reads them from `config/development.toml` — the same file the router reads its Superposition config from — and exports them via `GITHUB_ENV`, following the pattern already used in that step for `CYPRESS_ADMINAPIKEY` (`secrets.admin_api_key`). This:

- keeps a single source of truth (the Cypress-side credentials can never drift from what the router actually polls — same `endpoint` / `token` / `org_id` / `workspace_id`),
- leaves nothing credential-looking in the workflow file,
- needs no new repo secrets for what is an unauthenticated local demo service.

On rotation: the value is the token of the `superposition-demo` container (ghcr.io/juspay/superposition-demo:0.102.0) that CI runs on localhost:8081. The demo image does not enforce auth — the seed step talks to it with only `x-org-id`/`x-workspace` headers — and the same value is already committed at `config/development.toml` `superposition.token`. It is not an active credential, so there is nothing to rotate; the workflow no longer carries it regardless.

